PR #1472, opened 2023-05-04
Reverts changes for publishing symbols package

### LINKED ISSUE #1455: Publishing Symbols package to Nuget Symbol Server fails validation
During the release, publishing of symbols package to Nuget symbols server fails with the below error. 

The symbol package Microsoft.DataApiBuilder 0.6.14 failed validation because of the following reason(s):
•	The uploaded symbols package contains pdb(s) for a corresponding dll(s) not found in the nuget package.

Your symbol package was not published on NuGet Gallery and is not available for consumption.

Investigate and come up with a fix for successfully pushing it to Nuget.org's symbol server

severussundar:
## Why make this change?

- Closes #1455 
- Symbols package fails validation at the Nuget.org symbol server with the following error
>  The uploaded symbols package contains pdb(s) for a corresponding dll(s) not found in the nuget package.

## What is this change?

- Changes for publishing the symbols package are reverted. As a result, symbols package is not generated anymore.
- The PDB files are included in the primary NuGet package. So, developers still have the option of debugging through the source code. 

## How was this tested?

- [x] Build pipeline produces only the primary NuGet package. Symbols package is not generated.
- [x] Health status - All the health status items comes up as valid in the NuGet generated in the build pipeline.
image

## Sample Request(s)

- N/A

### DIFF
--- a/.pipelines/build-pipelines.yml
+++ b/.pipelines/build-pipelines.yml
@@ -211,7 +211,7 @@ steps:
   inputs:
     command: 'push'
     feedsToUse: 'select'
-    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg'
+    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg;$(Build.ArtifactStagingDirectory)/**/*.snupkg'
     nuGetFeedType: 'internal'
     publishVstsFeed: CosmosDB/DataApiBuilder
     versioningScheme: 'off'
@@ -222,7 +222,7 @@ steps:
   condition: and(succeeded(), eq(variables.isReleaseBuild, 'true'))
   inputs:
     command: push
-    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg'
+    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg;$(Build.ArtifactStagingDirectory)/**/*.snupkg'
     nuGetFeedType: external
     publishFeedCredentials: Microsoft.DataApiBuilder
 
--- a/.pipelines/build-pipelines.yml
+++ b/.pipelines/build-pipelines.yml
@@ -211,7 +211,7 @@ steps:
   inputs:
     command: 'push'
     feedsToUse: 'select'
-    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg;$(Build.ArtifactStagingDirectory)/**/*.snupkg'
+    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg;!$(Build.ArtifactStagingDirectory)/**/*.symbols.nupkg'
     nuGetFeedType: 'internal'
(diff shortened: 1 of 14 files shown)


Aniruddh25 (changes requested):
Looking for testing

seantleonard:
Making this change disables source link and you mentioned that this adds the pdb's to the binaries. Is this the only option in order to not display a commit number when running `dab --version` ? Can that CLI call perhaps drop the commit number when displaying the version? 

Your first justification for this change makes sense, we'd want symbols package to pass validation.

.pipelines/build-pipelines.yml
```diff
@@ -211,7 +211,7 @@ steps:
   inputs:
     command: 'push'
     feedsToUse: 'select'
-    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg'
+    packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg;!$(Build.ArtifactStagingDirectory)/**/*.symbols.nupkg'
```
abhishekkumams, line 214: i thought `.symbols.nupkg` files invalid and are not generated anymore

severussundar: remnant of git revert. removed it

src/Service.GraphQLBuilder/Azure.DataApiBuilder.Service.GraphQLBuilder.csproj
```diff
@@ -5,13 +5,6 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <OutputPath>$(BaseOutputPath)\engine</OutputPath>
-    <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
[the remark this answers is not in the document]
severussundar: Only the changes related to publishing the symbols package is reverted now. Deterministic builds are still enabled. 

With these changes, the health status of the NuGet (that is generated in the build pipeline) is valid

image

severussundar:
> Making this change disables source link and you mentioned that this adds the pdb's to the binaries. Is this the only option in order to not display a commit number when running `dab --version` ? Can that CLI call perhaps drop the commit number when displaying the version?
> 
> Your first justification for this change makes sense, we'd want symbols package to pass validation.



> Making this change disables source link and you mentioned that this adds the pdb's to the binaries. Is this the only option in order to not display a commit number when running `dab --version` ? Can that CLI call perhaps drop the commit number when displaying the version?
> 
> Your first justification for this change makes sense, we'd want symbols package to pass validation.

I agree. Only the changes related to publishing the symbols package is reverted now. Deterministic builds are still enabled. I'll pick up the changes for displaying the version number without the commit id hash in a follow-up PR

Aniruddh25 (approved):
LGTM, thanks for the follow up!

abhishekkumams (approved):

PR #1473, opened 2023-05-06
Support `time` data type in MsSql

### LINKED ISSUE #1401: [Enhancement] Support `time` data type in MsSql

Aniruddh25:
## Why make this change?

- Closes #1401 

## What is this change?

- `time` data type is detected as `TimeSpan` in C#. Add respective additional case for this type where ever types are handled. 

## Additional Fixes

- In GraphQL, while parsing a values of `datetime`, `datetime2`, they should be treated as UTC, otherwise HotChocolate tries to convert into UTC and max values stored in the db then cause overflow exceptions.
- Use `InvariantInfo` format provider while parsing

## How was this tested?

- [X] Integration Tests - in `SupportedType`
- [X] Unit Tests - `StoredProcedureBuilderTests`

## Sample Request(s)

image


src/Service/Parsers/EdmModelBuilder.cs
```diff
@@ -150,6 +102,42 @@ SourceDefinition sourceDefinition
             return this;
         }
 
+        /// <summary>
+        /// Given the system type, returns the corresponding primitive type kind.
+        /// </summary>
+        /// <param name="columnSystemType">Type of the column.</param>
+        /// <returns>EdmPrimitiveTypeKind</returns>
+        /// <exception cref="ArgumentException">Throws when the column</exception>
+        private static EdmPrimitiveTypeKind GetEdmPrimitiveTypeFromSystemType(Type columnSystemType)
```
ayush3797, line 111: Suggestion: we can move this function to the DbTypeHelper class (could be renamed to a more generic TypeHelper class as already done by Sean in his PR for openapi doc).

src/Service.Tests/GraphQLBuilder/Sql/StoredProcedureBuilderTests.cs
```diff
@@ -53,6 +53,7 @@ public class StoredProcedureBuilderTests
         [DataRow(typeof(DateTime), DATETIME_TYPE, "12/31/2030 12:00:00 AM", false, DisplayName = "DateTime")]
         [DataRow(typeof(DateTime), DATETIME_TYPE, "12/31/2030 12000 AM", true, DisplayName = "DateTime")]
         [DataRow(typeof(DateTimeOffset), DATETIME_TYPE, "11/19/2012 10:57:11 AM -08:00", false, DisplayName = "DateTimeOffset")]
+        [DataRow(typeof(TimeSpan), TIMESPAN_TYPE, "10:57:11.0000", false, DisplayName = "TimeSpan")]
```
severussundar, line 56: HotChocolate supports string literals only in ISO-8601 duration whereas SQL Server supports the following two additional formats (apart from the ones in ISO-8601)
`hhAM[PM]` and `hh AM[PM]`

[HotChocolate .NET Scalar Docs](https://chillicream.com/docs/hotchocolate/v13/defining-a-schema/scalars#datetime-type)
[Sql Server's time datatype Docs](https://learn.microsoft.com/en-us/sql/t-sql/data-types/time-transact-sql?view=sql-server-ver16)

Could be worth validating the other string formats and adding test cases

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -66,7 +66,7 @@ CREATE TABLE books(
 
 CREATE TABLE players(
     id int IDENTITY(5001, 1) PRIMARY KEY,
-    name varchar(max) NOT NULL,
+    [name] varchar(max) NOT NULL,
```
seantleonard, line 69: brackets because name is a reserved keyword?

src/Service/Services/ResolverMiddleware.cs
```diff
@@ -200,8 +201,9 @@ private static object PreParseLeaf(IMiddlewareContext context, string leafJson)
             {
                 ByteType => byte.Parse(leafJson),
                 SingleType => Single.Parse(leafJson),
-                DateTimeType => DateTimeOffset.Parse(leafJson),
+                DateTimeType => DateTimeOffset.Parse(leafJson, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
```
seantleonard, line 204: Your PR description notes:
> In GraphQL, while parsing a values of datetime, datetime2, they should be treated as UTC, otherwise HotChocolate tries to convert into UTC and max values stored in the db then cause overflow exceptions.

The first part i understand: Any datetime input provided by the user is assumed to be in UTC, which means an offset is required if they want to refer to their own local time.

```csharp
string dateString;
DateTimeOffset offsetDate;

dateString = "05/01/2008 6:00:00";

// Assume time is UTC
offsetDate = DateTimeOffset.Parse(dateString, null, DateTimeStyles.AssumeUniversal);
Console.WriteLine(offsetDate.ToString());   // Displays 5/1/2008 6:00:00 AM +00:00
```

So when `DateTimeOffset.Parse` is executed here, an offset +00:00 is then added [docs](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.datetimestyles?view=net-6.0#fields)
> If no time zone is specified in the parsed string, the string is assumed to denote a UTC. 

I'm not following the last part:
> HotChocolate tries to convert into UTC and max values stored in the db then cause overflow exceptions.

seantleonard (commented):
Few questions

src/Core/Services/TypeHelper.cs
```diff
@@ -29,6 +29,7 @@ public static class TypeHelper
             [typeof(char)] = DbType.StringFixedLength,
             [typeof(Guid)] = DbType.Guid,
             [typeof(byte[])] = DbType.Binary,
+            [typeof(TimeSpan)] = DbType.Time,
```
seantleonard, line 32: This PR seems to have additional Logic handling DateTime and DateTimeOffset, would it be appropriate (or necessary) to add:
```csharp
            [typeof(DateTime)] = DbType.DateTime,
            [typeof(DateTimeOffset)] = DbType.DateTimeOffset,
            [typeof(TimeSpan)] = DbType.Time,
```

src/Core/Services/TypeHelper.cs
```diff
@@ -43,6 +44,7 @@ public static class TypeHelper
             [typeof(bool?)] = DbType.Boolean,
             [typeof(char?)] = DbType.StringFixedLength,
             [typeof(Guid?)] = DbType.Guid,
+            [typeof(TimeSpan?)] = DbType.Time,
```
seantleonard, line 47: My PR has a helper you can mirror so that adding nullable types to this dictionary isn't necessary: https://github.com/Azure/data-api-builder/blob/04319b4da04fd1a1590ccc67f345b17522aa7177/src/Core/Services/TypeHelper.cs#L173-L203

src/Service.Tests/DatabaseSchema-MsSql.sql
abhishekkumams: yes.

abhishekkumams:
Some of the Tests are failing due to the following Assertion failure
as graphql is converting `1999-01-08T10:23:00` to `1999-01-08T10:23:00.000Z`, due to which the value returned by graphQL and the value stored in the DB do not match.

**Reason**:
When a  DateTime  value is serialized in GraphQL, it is typically represented in the ISO 8601 format, which includes the date, time, and timezone information. 
  
In the example provided above, the value "1999-01-08T10:23:00" is serialized as "1999-01-08T10:23:00.000Z". The ".000Z" represents the milliseconds and the "Z" indicates that the datetime is in UTC (Coordinated Universal Time) timezone. 
 
This default behavior ensures consistency in representing datetime values across different systems and ensures that timezone information is not lost during serialization and deserialization. 
 
If we want to customize the serialization format of the  DateTime  type in your GraphQL schema, we can define a custom scalar type and specify your desired serialization logic. However, it's important to note that changing the default serialization format may introduce compatibility issues with other systems consuming your GraphQL API.

abhishekkumams:
image

image

So, the difference is in date_types, datetime_types, datetime2_types, smalldatetime_types and timespan_types

abhishekkumams:
Here are a few examples of compatibility issues that could occur when customizing the serialization format of the  DateTime  type in GraphQL: 
 
1. Client Expectations: If you have clients consuming your GraphQL API that rely on the default ISO 8601 format for  DateTime  values, changing the serialization format could break their expectations. Clients may have logic or parsers in place that assume the default format, and altering it could lead to parsing errors or incorrect handling of datetime values. 
 
2. Third-Party Integrations: If your GraphQL API interacts with other systems or third-party integrations that expect datetime values in the default ISO 8601 format, changing the serialization format could cause compatibility issues. The receiving systems may not be able to parse or handle the modified format correctly, leading to data inconsistencies or errors. 
 
3. Tooling and Libraries: Many GraphQL client libraries and tools are built with the assumption of the default serialization format for datetime values. If you deviate from the standard format, you may encounter issues with these tools. For example, client-side code generators or query builders may not be able to handle the modified format, resulting in code generation errors or runtime issues. 
 
4. Data Exchange: If you share your GraphQL schema or data with other developers or organizations, they may expect datetime values to follow the default serialization format. Changing the format could lead to confusion or difficulties when exchanging data, as the receiving party may not be aware of the modification. 
 
It's important to consider these compatibility issues and potential impacts before making changes to the default serialization format of  DateTime  values in your GraphQL schema. If customization is necessary, it's recommended to communicate the changes clearly to all relevant stakeholders and ensure compatibility with existing clients and integrations.

abhishekkumams:
Notes after discussion:
1. rename timespan_type column to time_type.
2. find an equivalent C# type and hotchocolate for time.

NOTE:
In C#, starting from .NET 6.0, there is a TimeOnly structure available in the System.TimeOnly namespace. TimeOnly represents a time of day without a date component, similar to the SQL Server time data type.

TimeOnly provides properties like Hour, Minute, Second, and Millisecond to access the individual components of the time. It also supports various methods for comparing, formatting, and performing operations on time values.

src/Service.Tests/GraphQLBuilder/Sql/StoredProcedureBuilderTests.cs
abhishekkumams: added a test for both.

src/Service/Parsers/EdmModelBuilder.cs
```diff
@@ -150,6 +102,42 @@ SourceDefinition sourceDefinition
...
+                "Int32" => EdmPrimitiveTypeKind.Int32,
+                "Int64" => EdmPrimitiveTypeKind.Int64,
+                "Single" => EdmPrimitiveTypeKind.Single,
+                "Double" => EdmPrimitiveTypeKind.Double,
+                "Decimal" => EdmPrimitiveTypeKind.Decimal,
+                "Boolean" => EdmPrimitiveTypeKind.Boolean,
+                "DateTime" => EdmPrimitiveTypeKind.DateTimeOffset,
+                "DateTimeOffset" => EdmPrimitiveTypeKind.DateTimeOffset,
+                "Date" => EdmPrimitiveTypeKind.Date,
+                "TimeSpan" => EdmPrimitiveTypeKind.TimeOfDay,
```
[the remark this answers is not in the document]
Aniruddh25: Because Edm.Duration has the `days` component whereas `TimeOfDay` only has the time component.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -360,10 +361,12 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
                 "Double" => double.Parse(param),
                 "Decimal" => decimal.Parse(param),
                 "Boolean" => bool.Parse(param),
-                "DateTime" => DateTimeOffset.Parse(param),
+                "DateTime" => DateTime.Parse(param, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
                 "DateTimeOffset" => DateTimeOffset.Parse(param),
                 "Date" => DateOnly.Parse(param),
                 "Guid" => Guid.Parse(param),
+                "TimeOnly" => TimeOnly.Parse(param),
+                "TimeSpan" => TimeOnly.Parse(param),
```
seantleonard, line 369: This needs a test to confirm behavior: TimeSpan can be a negative number https://learn.microsoft.com/en-us/dotnet/api/system.timespan?view=net-7.0#:~:text=A%20TimeSpan%20value%20can,a%20time%20of%20day.

so if a param value is set to `-10.20:30:40.0500000` won't this fail? There is some historical precedent with entity framework and sql time type the relies on the user to provide a valid value within 0-24 hours, but I would rather not introduce that ambiguity here. 

Check out this doc to see if there are any pointers: https://learn.microsoft.com/en-us/dotnet/standard/datetime/how-to-use-dateonly-timeonly#work-with-timespan-and-datetime

src/Service.Tests/GraphQLBuilder/Sql/StoredProcedureBuilderTests.cs
```diff
@@ -54,6 +54,7 @@ public class StoredProcedureBuilderTests
         [DataRow(typeof(DateTime), DATETIME_TYPE, "12/31/2030 12:00:00 AM", false, DisplayName = "DateTime")]
         [DataRow(typeof(DateTime), DATETIME_TYPE, "12/31/2030 12000 AM", true, DisplayName = "DateTime")]
         [DataRow(typeof(DateTimeOffset), DATETIME_TYPE, "11/19/2012 10:57:11 AM -08:00", false, DisplayName = "DateTimeOffset")]
+        [DataRow(typeof(TimeOnly), LOCALTIME_TYPE, "10:57:11.0000", false, DisplayName = "LocalTime")]
```
seantleonard, line 57: need examples of negative value to see behavior as well as a time over 24 hours i.e. 30 hours

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/MySqlGQLSupportedTypesTests.cs
```diff
@@ -57,6 +57,8 @@ protected override bool IsSupportedType(string type)
             return type switch
             {
                 GUID_TYPE => false,
+                DATETIMEOFFSET_TYPE => false,
+                TIME_TYPE => false,
```
seantleonard, line 61: i wonder if this should be an allow list so that arbitrary values don't accidentally pass a test because a value like "MYCUSTOMTYPE" would result in 'true' here, wouldn't it?

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -262,6 +281,26 @@ public async Task InsertIntoTypeColumn(string type, string value)
             await ResetDbStateAsync();
         }
 
+        [DataTestMethod]
+        [DataRow(TIME_TYPE, "\"32:59:59.9999999\"")]
+        [DataRow(TIME_TYPE, "\"22:67:59.9999999\"")]
```
seantleonard, line 286: add comment for what you are validating. This may address one scenario i mentioned in another comment: including a value representing a time over 24 hours and attempt to have DAB convert it to TimeOnly. SHould also try a negative value.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -509,7 +539,67 @@ private static void CompareDateTimeResults(string actual, string expected)
...
+                // Adjusting to universal, since DateTime doesn't account for TimeZone
+                DateTime expectedDateTimeUniversal = DateTime.Parse(expectedDateTime, CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal);
+                DateTime actualDateTimeUniversal = DateTime.Parse(actualDateTime, CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal);
+                Assert.AreEqual(expectedDateTimeUniversal, actualDateTimeUniversal);
+            }
+        }
+
+        /// <summary>
+        /// Required due to different format between mysql datetimeoffset and HotChocolate datetime
+        /// result
```
seantleonard, line 551: can you elaborate here?

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -509,7 +539,67 @@ private static void CompareDateTimeResults(string actual, string expected)
...
+            string expectedTimeString = expectedJsonDoc.RootElement.GetProperty(fieldName).ToString();
+
+            // handles cases when one of the values is null
+            if (string.IsNullOrEmpty(actualTimeString) || string.IsNullOrEmpty(expectedTimeString))
+            {
+                Assert.AreEqual(expectedTimeString, actualTimeString);
+            }
+            else
+            {
+                TimeOnly actualTime = TimeOnly.Parse(actualTimeString, CultureInfo.InvariantCulture);
```
seantleonard, line 600: can you add comments why you chose certain `CultureInfo.` values? What kind of documentation do we need to add to Dab docs to clarify what timezone/lack of timezone / culture may be expected in GraphQL results and request payloads?

seantleonard (changes requested):
need some more tests and explanation for why certain culture/universal time values are used when parsing dates. Also need some more clarity on Converting between TimeSpan and TimeOnly.

src/Service.Tests/GraphQLBuilder/Sql/StoredProcedureBuilderTests.cs
abhishekkumams: Those tests are already there in the GraphQLSupportedTestBase.cs

Aniruddh25 (commented):
Thanks for investigating and finding solution to use NodaTime! Left few suggestions.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
abhishekkumams: SqlClient currently is not recognizing type TimeOnly, the only way to use TimeOnly is to Parse it. That's why I have added Both, But the parsing will only happen into TimeOnly.

I have added a test for negative values as well.

src/Core/Services/TypeHelper.cs
```diff
@@ -151,6 +180,15 @@ public static JsonDataType GetJsonDataTypeFromSystemType(Type type)
         /// <returns>DbType for the given system type. Null when no mapping exists.</returns>
         public static DbType? GetDbTypeFromSystemType(Type systemType)
         {
+            // Get the underlying type argument if the 'systemType' argument is a nullable type.
+            Type? nullableUnderlyingType = Nullable.GetUnderlyingType(systemType);
```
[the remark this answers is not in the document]
seantleonard: A nullable value type like `int? myInt` is represented by `Nullable<T>` -> `Nullable<int>`. So `Type? systemType` wouldn't apply.

`Nullable.GetUnderlyingType(Type systemType)` would resolve `int` from the passed in `int?` /  `Nullable<int>`.

Then the question switches to nullable reference types like `string?`. Per [csharplang discussion](https://github.com/dotnet/csharplang/discussions/3003#discussioncomment-129629), `string?` is represented in runtime as `string` with additional type metadata:
> Because nullable reference type does not have a standalone runtime representation. When using reflection, it's represented by NullableContextAttribute on the target member.

Whether we ever resolve a nullable value/reference type from the database is another question. I ran the type tests locally from 'main' and removed all the nullable value types from `_systemTypeToDbTypeMap`, and didn't come across failures in tests containing the string 'type'. I don't recall why the nullable types were in that collection to begin with. Because they were present and I didn't want to risk regressions in #1568 with regards to type handling, i added the helper replicated here for pulling the underlying type out of `Nullable<T>` /nullable value types. I also had written tests to cover nullable value types if they ever occurred: https://github.com/Azure/data-api-builder/blob/60e38a2a3c2cdb250ad71582becd69d9388ffa73/src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs#L108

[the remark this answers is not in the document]
Aniruddh25: I get it now.

src/Directory.Packages.props
```diff
@@ -19,7 +20,7 @@
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="6.0.14" />
     <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.20.0" />
     <!--When updating Microsoft.Data.SqlClient, update license URL in scripts/notice-generation.ps1-->
-    <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.0.1" />
+    <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.1.1" />
```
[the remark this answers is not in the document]
abhishekkumams: sql 5.1.1 was necessary for .Net 6.0 TimeOnly, i was getting error without it, https://erikej.github.io/dotnet/sqlclient/2022/11/17/sqlclient-dateonly.html

src/Core/Services/TypeHelper.cs
abhishekkumams: Not Adding this hard mapping, since System.DateTime is used for both SqldbType.DateTime and SqldbType.DateTime2. So no change required here as seperate logic exist which does the mapping and also Hotchocolate only has Hotchocolate.Types.DateTime for SqlDbType.DateTime/DateTime2/DateTimeOffset.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -509,7 +544,67 @@ private static void CompareDateTimeResults(string actual, string expected)
...
+        /// Required due to different format between mysql datetimeoffset and HotChocolate datetime
+        /// result
+        /// </summary>
+        private static void CompareDateTimeOffsetResults(string actual, string expected)
+        {
+            string fieldName = "datetimeoffset_types";
+
+            using JsonDocument actualJsonDoc = JsonDocument.Parse(actual);
+            using JsonDocument expectedJsonDoc = JsonDocument.Parse(expected);
+
```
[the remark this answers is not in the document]
abhishekkumams: added tests for the same.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/PostgreSqlGQLSupportedTypesTests.cs
```diff
@@ -54,7 +54,8 @@ protected override bool IsSupportedType(string type)
             return type switch
             {
                 BYTE_TYPE => false,
-                DATETIME_NONUTC_TYPE => false,
+                DATETIMEOFFSET_TYPE => false,
+                TIME_TYPE => false,
```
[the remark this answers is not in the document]
abhishekkumams: yep created

src/Service/Services/ResolverMiddleware.cs
abhishekkumams: **updated Description**. By using  `DateTimeStyles.AssumeUniversal` , wecan parse DateTime strings without requiring an explicit time zone indicator, as long as the input is in the expected UTC format. It doesn't try to convert it, it just assumes the time provided is already in the UTC format.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/MySqlGQLSupportedTypesTests.cs
abhishekkumams: If an arbitrary type is specified here and it passes, it means that db support that type, the main reason for adding these was to skip testing dbs for certain types.

src/Core/Services/TypeHelper.cs
```diff
@@ -108,14 +100,51 @@ public static class TypeHelper
...
+                "Int64" => EdmPrimitiveTypeKind.Int64,
+                "Single" => EdmPrimitiveTypeKind.Single,
+                "Double" => EdmPrimitiveTypeKind.Double,
+                "Decimal" => EdmPrimitiveTypeKind.Decimal,
+                "Boolean" => EdmPrimitiveTypeKind.Boolean,
+                "DateTime" => EdmPrimitiveTypeKind.DateTimeOffset,
+                "DateTimeOffset" => EdmPrimitiveTypeKind.DateTimeOffset,
+                "Date" => EdmPrimitiveTypeKind.Date,
+                "TimeOnly" => EdmPrimitiveTypeKind.TimeOfDay,
+                "TimeSpan" => EdmPrimitiveTypeKind.TimeOfDay,
```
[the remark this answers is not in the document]
abhishekkumams: Removed TimeOnly, because when fetching Schema, it is recognizing the Time column as TimeSpan, usually TimeSpan should be mapped with EdmType.Duration, but here we using TimeOfDay because it's for the Time and not Duration.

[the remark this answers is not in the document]
abhishekkumams: I added tests for OData as well.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
seantleonard: > SqlClient currently is not recognizing type TimeOnly,

This works with SqlClient 5.1.1 though right?

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -271,8 +324,10 @@ public async Task InsertIntoTypeColumn(string type, string value)
         [DataRow(FLOAT_TYPE, -3.33)]
         [DataRow(DECIMAL_TYPE, 1222222.00000929292)]
         [DataRow(BOOLEAN_TYPE, true)]
-        [DataRow(DATETIME_NONUTC_TYPE, "1999-01-08 10:23:54+8:00")]
+        [DataRow(DATETIMEOFFSET_TYPE, "1999-01-08 10:23:54+8:00")]
         [DataRow(DATETIME_TYPE, "1999-01-08 10:23:54")]
+        [DataRow(TIME_TYPE, "\"23:59:59.9999999\"")]
+        [DataRow(TIME_TYPE, "null")]
```
seantleonard, line 330: need test description: what is being tested and what are you validating

src/Core/Services/TypeHelper.cs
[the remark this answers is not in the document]
seantleonard: > Removed TimeOnly, because when fetching Schema, it is recognizing the Time column as TimeSpan,

When would we resolve a SQL column of tsql type `time` to .NET `TimeOnly` as indicated is possible by https://erikej.github.io/dotnet/sqlclient/2022/11/17/sqlclient-dateonly.html

Since we map SqlDbType.Time and DbType.Time to .NET `TimeOnly` we may want to keep `"TimeOnly" => EdmPrimitiveTypeKind.TimeOfDay,` in case we update to support reading db type time to .net TimeOnly type. or if there is a code path today that does that , that we didn't catch here.

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
```diff
@@ -99,12 +99,12 @@ private static IEnumerable<object[]> GetTestData_SupportedSystemTypesMapToJsonVa
...
     [DataRow(typeof(Guid?))]
+    [DataRow(typeof(TimeOnly?))]
     [DataRow(typeof(TimeSpan?))]
-    [DataRow(typeof(DateTime?))]
-    [DataRow(typeof(DateTimeOffset?))]
     [DataTestMethod]
     public void ResolveUnderlyingTypeForNullableValueType(Type nullableType)
     {
         Assert.AreNotEqual(notExpected: JsonDataType.Undefined, actual: TypeHelper.GetJsonDataTypeFromSystemType(nullableType));
+        Assert.IsNotNull(TypeHelper.GetDbTypeFromSystemType(nullableType));
```
seantleonard, line 108: why is this assert needed? `TypeHelper.GetDbTypeFromSystemType` never returns null.

src/Core/Services/TypeHelper.cs
Aniruddh25: Can you please add this reason as a comment here? It seems like a valid suggestion but has an equally valid reasoning to not do it, so people looking at the code should know the reasing :)

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -621,7 +637,15 @@ private static void ValidateArrayResults(JsonDocument actualJsonDoc, JsonDocumen
                 actualElement.TryGetProperty(fieldName, out JsonElement actualValue);
                 expectedElement.TryGetProperty(fieldName, out JsonElement expectedValue);
 
-                if (fieldName.StartsWith(DATETIME_TYPE.ToLower()))
+                if (fieldName.StartsWith(DATETIMEOFFSET_TYPE.ToLower()))
+                {
+                    DateTimeOffset actualDateTimeOffset = DateTimeOffset.Parse(actualValue.ToString(), DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal);
```
Aniruddh25, line 642: The actual comparison of the string values in this ValidateArrayResults and the above Compare* functions seem to be same. Can we move into helper functions?

Aniruddh25 (commented):
Few more questions

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -361,8 +361,8 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
                 "Double" => double.Parse(param),
                 "Decimal" => decimal.Parse(param),
                 "Boolean" => bool.Parse(param),
-                "DateTime" => DateTime.Parse(param, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
-                "DateTimeOffset" => DateTimeOffset.Parse(param),
+                "DateTime" => DateTimeOffset.Parse(param, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
```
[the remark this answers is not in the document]
Aniruddh25: nvm I found the reason in old PR from Gledis.

src/Core/Services/TypeHelper.cs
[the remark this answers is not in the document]
abhishekkumams: yep, makes sense.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
abhishekkumams: Sql 5.1.1 supports TimeOnly, but the problem we have is the `SqlDataReader` which we use to parse the Schema of tables. in that the SystemType automatically comes as TimeSpan, instead of TimeOnly for SqlDbType.Time

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -161,37 +167,59 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
         [DataRow(DATETIME_TYPE, "neq", "\'1999-01-08 10:23:54\'", "\"1999-01-08 10:23:54\"", "!=")]
         [DataRow(DATETIME_TYPE, "eq", "\'1999-01-08 10:23:54\'", "\"1999-01-08 10:23:54\"", "=")]
         public async Task QueryTypeColumnFilterAndOrderByDateTime(string type, string filterOperator, string sqlValue, string gqlValue, string queryOperator)
+        {
+            // In MySQL, the DATETIME data type supports a range from '1000-01-01 00:00:00.0000000' to '9999-12-31 23:59:59.0000000'
+            if (DatabaseEngine is TestCategory.MYSQL && sqlValue is "\'9999-12-31 23:59:59.9999999\'")
+            {
+                sqlValue = "\'9999-12-31 23:59:59.0000000\'";
+                gqlValue = "\"9999-12-31 23:59:59.0000000\"";
+            }
```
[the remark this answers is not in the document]
abhishekkumams: We don't have it. For MySql, if the value is `9999-12-31 23:59:59.9999999`
we get the error stating "Datetime function: datetime field overflow"

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -161,37 +167,59 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
+
+            await QueryTypeColumnFilterAndOrderBy(type, filterOperator, sqlValue, gqlValue, queryOperator);
+        }
+
+        /// <summary>
+        /// Test case for LocalTime filters.
```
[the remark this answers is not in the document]
abhishekkumams: added test

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -239,8 +243,10 @@ public static ObjectType UnderlyingGraphQLEntityType(IType type)
                     SINGLE_TYPE => new(SINGLE_TYPE, new SingleType().ParseValue(float.Parse(defaultValueFromConfig))),
                     FLOAT_TYPE => new(FLOAT_TYPE, new FloatValueNode(double.Parse(defaultValueFromConfig))),
                     DECIMAL_TYPE => new(DECIMAL_TYPE, new FloatValueNode(decimal.Parse(defaultValueFromConfig))),
-                    DATETIME_TYPE => new(DATETIME_TYPE, new DateTimeType().ParseResult(DateTime.Parse(defaultValueFromConfig))),
+                    DATETIME_TYPE => new(DATETIME_TYPE, new DateTimeType().ParseResult(
+                        DateTime.Parse(defaultValueFromConfig, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal))),
                     BYTEARRAY_TYPE => new(BYTEARRAY_TYPE, new ByteArrayType().ParseValue(Convert.FromBase64String(defaultValueFromConfig))),
+                    LOCALTIME_TYPE => new(LOCALTIME_TYPE, new LocalTimeType().ParseResult(LocalTimePattern.ExtendedIso.Parse(defaultValueFromConfig).Value)),
```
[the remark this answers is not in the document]
abhishekkumams: Currently, through graphQL if we supply, let's say `10:23:14.999999999` (9 decimal places), in the DB the value inserted will be  `10:23:15.0000000` (7 decimal places)

abhishekkumams (approved):

src/Core/Services/TypeHelper.cs
```diff
@@ -17,6 +17,10 @@ public static class TypeHelper
     {
         /// <summary>
         /// Maps .NET Framework types to DbType enum
+        /// Not Adding this hard mapping for System.DateTime/DateTimeOffset,
+        /// since System.DateTime is used for both SqldbType.DateTime and SqldbType.DateTime2.
+        /// So no change required here as seperate logic exist which does the mapping
```
Aniruddh25, line 22:
```suggestion
        /// So no change required here as separate logic exist which does the mapping
```

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -637,35 +662,43 @@ private static void ValidateArrayResults(JsonDocument actualJsonDoc, JsonDocumen
...
+                AssertOnFields(fieldName, actualValue.ToString(), expectedValue.ToString());
+            }
+        }
+
+        /// <summary>
+        /// Compare given fields from actual and expected json.
+        /// </summary>
+        private static void AssertOnFields(string field, string actualElement, string expectedElement)
+        {
+            if (field.StartsWith(DATETIMEOFFSET_TYPE.ToLower()))
```
Aniruddh25, line 674: Does it have to be `StartsWith` or `Equals`?

Aniruddh25:
Approved but since I authored it initially, can't represent it in github. Can consider @abhishekkumams approval on my behalf.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
abhishekkumams: StartsWith because the field here is the name of the column, ex: datetime_types, while the const variable here just represent the type, i.e, datetime.

src/Core/Services/TypeHelper.cs
```diff
@@ -16,6 +17,10 @@ public static class TypeHelper
     {
         /// <summary>
         /// Maps .NET Framework types to DbType enum
+        /// Not Adding this hard mapping for System.DateTime/DateTimeOffset,
+        /// since System.DateTime is used for both SqldbType.DateTime and SqldbType.DateTime2.
+        /// So no change required here as separate logic exist which does the mapping
```
seantleonard, line 22: nit: the comment has no context from the PR Comment thread. Mapping of `System.DateTime`, `DateTimeOffset`, 
and `System.DateTime` to what? 

Also SqlDbType is not relevant in this mapping dictionary is it?

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
ayush3797: Why this change?

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
seantleonard: No comment was ever added

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -262,6 +323,32 @@ public async Task InsertIntoTypeColumn(string type, string value)
...
+        /// Test case for invalid time, such as negative values or hours>24 or minutes/seconds>60.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(TIME_TYPE, "\"32:59:59.9999999\"")]
+        [DataRow(TIME_TYPE, "\"22:67:59.9999999\"")]
+        [DataRow(TIME_TYPE, "\"14:12:99.9999999\"")]
+        [DataRow(TIME_TYPE, "\"-22:67:59.9999999\"")]
+        [DataRow(TIME_TYPE, "\"22:-67:59.9999999\"")]
+        [DataRow(TIME_TYPE, "\"22:67:59.-9999999\"")]
+        public async Task InsertInvalidIntoTypeColumn(string type, string value)
```
ayush3797, line 336: The function name needs to be reviewed.

src/Service.GraphQLBuilder/GraphQLUtils.cs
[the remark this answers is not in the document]
seantleonard: the loss of precision and rounding needs to be documented, please write that up in a doc

seantleonard (approved):

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
abhishekkumams: Updated description on Additional Fix in PR description. Sorry for the confusion.

src/Core/Services/TypeHelper.cs
abhishekkumams: yep, not SqlDbType but DbType

PR #1492, opened 2023-05-15
Fix schema path in test files

abhishekkumams:
## Why make this change?

- To update the files to use the correct reference of schema file.

## What is this change?

- removing usage of Azure storage links.
- updating it to be consistent with the other test files.

### DIFF
--- a/src/Service.Tests/dab-config.MsSql.json
+++ b/src/Service.Tests/dab-config.MsSql.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://dataapibuilder.azureedge.net/schemas/vmajor.minor.patch-beta/dab.draft.schema.json",
+  "$schema": "../../../schemas/dab.draft.schema.json",
   "data-source": {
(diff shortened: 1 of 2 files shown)


src/Service.Tests/dab-config.MsSql.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://dataapibuilder.azureedge.net/schemas/vmajor.minor.patch-beta/dab.draft.schema.json",
+  "$schema": "../../../schemas/dab.draft.schema.json",
```
severussundar, line 2:
```suggestion
  "$schema": "../../schemas/dab.draft.schema.json",
```

severussundar: Please update the same in `dab-config.MySql.json` as well

severussundar (approved):

Aniruddh25 (approved):

PR #1493, opened 2023-05-15
Fix merge config file not available issue

abhishekkumams:
## What is the issue?
- This issue occurs specifically when the environment variable is set and `environmentBasedConfigFile` is not present and the user tries to do `dab start`. It will search for the merged file and fail directly if the merged file is not present. Ideally it should not and rather look for the default file.
- Even though the merging was not happening and the method `TryMergeConfigsIfAvailable` was returning false, the out file was not empty and it was containing the name of expected mergeConfigFile.
 
## Why make this change?
  - the merge config file name is generated and set as the config to be used even if the merged config file is not available.
  - This will make sure merge config file name is not used if it is not available

## What is this change?

- Moving the line which creates the merged file name inside the validation check for existence of the base and environment config.

## How was this tested?

- [x] Unit Tests

### DIFF
--- a/src/Cli.Tests/UtilsTests.cs
+++ b/src/Cli.Tests/UtilsTests.cs
@@ -239,7 +239,7 @@ namespace Cli.Tests
         /// 1. Environment value is set.
         /// 2. Both Base and envBased config file is present.
         /// In all other cases, the TryMergeConfigsIfAvailable method should return false
-        /// and out string for the mergedConfigFile should be empty.
+        /// and out param for the mergedConfigFile should be null.
         /// </summary>
         [DataTestMethod]
         [DataRow("", false, false, null, false, DisplayName = "If environment value is not set, merged config file is not generated.")]
--- a/src/Cli.Tests/UtilsTests.cs
+++ b/src/Cli.Tests/UtilsTests.cs
@@ -222,7 +222,7 @@ namespace Cli.Tests
             Environment.SetEnvironmentVariable(RUNTIME_ENVIRONMENT_VAR_NAME, "Test");
             File.WriteAllText("dab-config.json", BASE_CONFIG);
             File.WriteAllText("dab-config.Test.json", ENV_BASED_CONFIG);
-            if (TryMergeConfigsIfAvailable(out string mergedConfig))
+            if (TryMergeConfigsIfAvailable(out string? mergedConfig))
             {
                 Assert.AreEqual(mergedConfig, "dab-config.Test.merged.json");
                 Assert.IsTrue(File.Exists(mergedConfig));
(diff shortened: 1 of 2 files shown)


src/Cli.Tests/UtilsTests.cs
```diff
@@ -234,6 +234,64 @@ public void TestMergeConfig()
...
+        [DataRow("", false, true, "", false, DisplayName = "If environment value is not set, merged config file is not generated.")]
+        [DataRow("", true, false, "", false, DisplayName = "If environment value is not set, merged config file is not generated.")]
+        [DataRow("", true, true, "", false, DisplayName = "If environment value is not set, merged config file is not generated.")]
+        [DataRow("Test", false, false, "", false, DisplayName = "Environment value set but base config not available, merged config file is not generated.")]
+        [DataRow("Test", false, true, "", false, DisplayName = "Environment value set but base config not available, merged config file is not generated.")]
+        [DataRow("Test", true, false, "", false, DisplayName = "Environment value set but env based config not available, merged config file is not generated.")]
+        [DataRow("Test", true, true, "dab-config.Test.merged.json", true, DisplayName = "Environment value set and both base and envConfig available, merged config file is generated.")]
+        public void TestMergeConfigAvailability(
```
Aniruddh25, line 253: Was there no other test testing these scenarios indirectly ie End 2 End?

Aniruddh25 (approved):
LGTM, thanks for catching this scenario!

src/Cli.Tests/UtilsTests.cs
abhishekkumams: Nope, Last time the test was focused more the merge capabilities.

src/Cli/Utils.cs
```diff
@@ -898,10 +898,10 @@ public static bool TryMergeConfigsIfAvailable(out string mergedConfigFile)
             {
                 string baseConfigFile = RuntimeConfigPath.DefaultName;
                 string environmentBasedConfigFile = RuntimeConfigPath.GetFileName(environmentValue, considerOverrides: false);
-                mergedConfigFile = RuntimeConfigPath.GetMergedFileNameForEnvironment(CONFIGFILE_NAME, environmentValue);
 
                 if (DoesFileExistInCurrentDirectory(baseConfigFile) && !string.IsNullOrEmpty(environmentBasedConfigFile))
                 {
+                    mergedConfigFile = RuntimeConfigPath.GetMergedFileNameForEnvironment(CONFIGFILE_NAME, environmentValue);
```
severussundar, line 904: Might be helpful to the user to display the base config, override config file and the final merged config file names.

severussundar, line 904: This line could be moved after the line where the merge function is called.
`Merge(baseConfigJson, overrideConfigJson)`

This way, the merged config file name is set only after the json strings are merged successfully.

src/Cli/Utils.cs
```diff
@@ -898,17 +898,18 @@ public static bool TryMergeConfigsIfAvailable(out string mergedConfigFile)
...
                 if (DoesFileExistInCurrentDirectory(baseConfigFile) && !string.IsNullOrEmpty(environmentBasedConfigFile))
                 {
                     try
                     {
                         string baseConfigJson = File.ReadAllText(baseConfigFile);
                         string overrideConfigJson = File.ReadAllText(environmentBasedConfigFile);
+                        _logger.LogInformation($"Merging {baseConfigFile} and {environmentBasedConfigFile}");
                         string mergedConfigJson = Merge(baseConfigJson, overrideConfigJson);
-
+                        mergedConfigFile = RuntimeConfigPath.GetMergedFileNameForEnvironment(CONFIGFILE_NAME, environmentValue);
```
seantleonard, line 910: The update here is that `mergedConfigFile` will be the default config file when a merged file is not created? is this variable now not representative of what its name describes? 

 `TryMergeConfigsIfAvailable(out string mergedConfigFile)` will need to be updated based on your PR description comment by checking that the out parameter is null when the return value is false? and an out param attribute `[NotNullWhen(true)]`

src/Cli/Utils.cs
```diff
@@ -898,17 +898,18 @@ public static bool TryMergeConfigsIfAvailable(out string mergedConfigFile)
...
                         File.WriteAllText(mergedConfigFile, mergedConfigJson);
+                        _logger.LogInformation($"Generated merged config file: {mergedConfigFile}");
```
seantleonard, line 912: is it possible to log the full file path? This would be a good place to start with that enhancement, and other locations can follow in future PRs.

src/Cli.Tests/UtilsTests.cs
```diff
@@ -234,6 +234,64 @@ public void TestMergeConfig()
...
+            string environmentValue,
+            bool isBaseConfigPresent,
+            bool isEnvironmentBasedConfigPresent,
+            string expectedMergedConfigFileName,
+            bool expectedIsMergedConfigAvailable)
+        {
+            Environment.SetEnvironmentVariable(RUNTIME_ENVIRONMENT_VAR_NAME, environmentValue);
+            string baseConfig = "dab-config.json";
+            string envBasedConfig = "dab-config.Test.json";
+            if (File.Exists(baseConfig))
```
seantleonard, line 263: Can you wrap these file operations in a comment describing that they are setting up the test scenario?

src/Cli.Tests/UtilsTests.cs
```diff
@@ -234,6 +234,64 @@ public void TestMergeConfig()
...
+        [DataRow("Test", true, false, "", false, DisplayName = "Environment value set but env based config not available, merged config file is not generated.")]
+        [DataRow("Test", true, true, "dab-config.Test.merged.json", true, DisplayName = "Environment value set and both base and envConfig available, merged config file is generated.")]
+        public void TestMergeConfigAvailability(
+            string environmentValue,
+            bool isBaseConfigPresent,
+            bool isEnvironmentBasedConfigPresent,
+            string expectedMergedConfigFileName,
+            bool expectedIsMergedConfigAvailable)
+        {
+            Environment.SetEnvironmentVariable(RUNTIME_ENVIRONMENT_VAR_NAME, environmentValue);
```
seantleonard, line 260: Does the CLI act differently if the environment variable KEY is not present? this function call sets the key but may leave the value empty, which is slightly different than the key not existing.

src/Cli.Tests/UtilsTests.cs
```diff
@@ -234,6 +234,64 @@ public void TestMergeConfig()
...
+            if (isEnvironmentBasedConfigPresent)
+            {
+                if (!File.Exists(envBasedConfig))
+                {
+                    File.Create(envBasedConfig).Close();
+                    File.WriteAllText(envBasedConfig, "{}");
+                }
+            }
+
+            Assert.AreEqual(expectedIsMergedConfigAvailable, TryMergeConfigsIfAvailable(out string mergedConfigFile));
```
seantleonard, line 291: mentioned in another comment, but would we want a test to ensure that `TryMergeConfigsIfAvailable()` never gives the out param a value AND return false?

seantleonard (dismissed):
some considerations

src/Cli/Utils.cs
abhishekkumams: updated

src/Cli.Tests/UtilsTests.cs
abhishekkumams: Thanks, Added tests for the same.

src/Cli.Tests/UtilsTests.cs
abhishekkumams: The Data Test row all the scenarios are testing that when it returns false, the out param is null.

severussundar (approved):
Thanks for spotting this and pushing a fix

Issue #1494, opened 2023-05-15
Scenarios for rest path/methods not validated for entities


ayush3797:
Encountered various scenarios where we should be failing the runtime startup during config validation but we are not.
If the rest settings are configured for an entity, then:

1. The value for "path" property cannot contain any reserved characters like the restPath or the graphqlPath.
2. "path" property cannot have the same value for 2 different entities.
3. "path" property cannot be empty.

PR #1402, opened 2023-04-03
Overhaul of config system

aaronpowell:
## Why make this change?

The original config system has a lot of areas where it doesn't provide concrete types for what we need, due to the flexible nature of the JSON file, and instead forces the code to check at runtime (is the property a bool or an object?) and then have branching logic based on that.

This results in duplication of code, as well as hidden assumptions for what the "expanded" version of shorthand properties (by shorthand properties I mean properties like `"rest": false` on the `runtime`).

With this change we reduce the config C# type system to only handle the fully expanded objects, resulting in a clearer structure for what the types are between JSON and C#, reduce the amount of type casting that needs to be done when checking the config, and unblocking at ability to tackle more complex config features, such as #722.

## What is this change?

- Full rewrite of the config type system and "loader" process
  - This includes a set of custom `JsonConverter` classes for parts of the config serializer/deserializer, so that we can expand the config and apply the defaults as expected. The result of this is that if you deserialize then serialize the resulting config can look different as we _always_ serialize to the expanded version
- Leveraging a file system abstraction for easier testing
  - Unit tests can generate a mock `RuntimeConfig` and write it to a mock file system, meaning they are a lot more isolated from each other
  - Integration tests can load the "base" `RuntimeConfig` for the DB and then write out a config (with test-specific features) to the mock file system, which is then given to the test runner to use, rather than writing to disk (although some tests will still write to disk)
- Refactored engine to use the new types
- Introduced [Snapshot Testing](https://jestjs.io/docs/en/snapshot-testing/) via [Snapshooter](https://github.com/SwissLife-OSS/snapshooter) to make it easier to see the broad impact of refactors to underlying object structures
- Refactored the CLI to be a bit more compartmentalised in the code, relying less on callbacks, making for easier testing of specific commands

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

PR #1496, opened 2023-05-15
Filling gaps in config validation for rest settings for entities

ayush3797:
## Why make this change?
There were a lot of validation failure cases that were not in place for runtime config validation and hence were not getting caught as mentioned in the issue: https://github.com/Azure/data-api-builder/issues/1494. This PR addresses the same. 

## What is this change?
Added validations for the below mentioned cases for REST settings for entity:
`path`:
1. it cannot have the same value for 2 different entities,
2. It cannot be empty.
3. It cannot contain any reserved characters

## How was this tested?

- [X] Unit Tests


src/Config/Entity.cs
```diff
@@ -594,17 +600,17 @@ public record RestStoredProcedureEntitySettings([property: JsonPropertyName("met
...
 
     /// <summary>
     /// Describes the GraphQL settings specific to an entity.
     /// </summary>
     /// <param name="Type">Defines the name of the GraphQL type.
     /// Can be a string or Singular-Plural type.
     /// If string, a default plural route will be added as per the rules at
-    /// </param>
     /// <seealso cref="<https://engdic.org/singular-and-plural-noun-rules-definitions-examples/"/>
+    /// </param>
```
seantleonard, line 613: Does IntelliSense still detect `<seealso />` be nested inside `<param></param>`?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1747,140 @@ string[] linkingTargetFields
...
+        /// <summary>
+        /// Test to validate that when multiple entities have the same custom rest path configured, we throw an exception.
+        /// </summary>
+        /// <param name="exceptionExpected">Whether an exception is expected as a result of test run.</param>
+        /// <param name="restPathForFirstEntity">Custom rest path to be configured for the first entity.</param>
+        /// <param name="restPathForSecondEntity">Custom rest path to be configured for the second entity.</param>
+        /// <param name="expectedExceptionMessage">The expected exception message.</param>
+        [DataTestMethod]
+        [DataRow(false, "restPathA", "restPathB", DisplayName = "Unique rest paths configured for entities pass config validation.")]
+        [DataRow(false, true, false, DisplayName = "Rest path configured as boolean values for an entities fails config validation.")]
```
seantleonard, line 1845:
```csharp
[DataRow(false, true, false, DisplayName = "Rest path configured as boolean values for an entities fails config validation.")]
```
This test doesn't belong here since it isn't validating behavior of two entities using the same REST path value.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1747,140 @@ string[] linkingTargetFields
...
+        [DataRow(true, "restPath", "restPath", "Multiple entities found with same rest path: restPath.",
+            DisplayName = "Duplicate rest paths configures for entities fail config validation.")]
+        [DataRow(true, "", "restPathB", "Entity: EntityA has an empty rest path.",
+            DisplayName = "Empty rest path configured for an entitiy fails config validation.")]
+        [DataRow(true, "restPathA", null, "Entity: EntityB has a null rest path. Accepted value types are: string, boolean.",
+            DisplayName = "NULL rest path configured for an entitiy fails config validation.")]
+        [DataRow(true, "restPathA", 1, $"Entity: EntityB has rest path specified with incorrect data type. Accepted data types are: string, boolean.",
+            DisplayName = "Rest path configured as integer for an entitiy fails config validation.")]
```
seantleonard, line 1853: Do these tests belong here? It seems like these tests would be simpler or achieve the same results by only testing with a single entity and assigning various fault inducing values. e.g. `restPathForFirstEntity` value `restPathA` doesn't affect the outcome of these tests, only the value you provide for `restPathForSecondEntity`

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -239,25 +239,48 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(IDict
...
             {
                 Entity entity = entityCollection[entityName];
+                if (entity.Rest is not null)
+                {
+                    JsonElement restJsonElement = JsonSerializer.SerializeToElement(entity.Rest);
+
+                    // We do the validation for rest path only if the 'rest' property maps to a json object.
+                    if (restJsonElement.ValueKind is JsonValueKind.Object)
+                    {
+                        // Since 'path' is an optional property, we skip validation if its absent.
```
seantleonard, line 271: Defining REST 'path' indicates that the developer doesn't want the top-level entity name to be the rest path value. Wouldn't we also need to populate `restPathsForEntities` with top-level entity names when:
(entity.REST.enabled == true AND entity.path is not specified) 
so those conflicts can be avoided?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +306,128 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+        /// <summary>
+        /// Helper method to validate that the rest path is correctly configured for the entity.
+        /// The rest path can only be a boolean value or a string.
+        /// If configured as a string, it should not be null/empty and should not conflict with the rest path
+        /// configured for any other entity.
+        /// </summary>
+        /// <param name="entityName">Name of the entity.</param>
+        /// <param name="restPathElement">The rest path element for the entity.</param>
+        /// <param name="restPathsForEntities">Set of unique rest paths configured for the entities in the config.</param>
+        /// <exception cref="DataApiBuilderException"></exception>
```
seantleonard, line 318: need this description completed.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +306,128 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+        private static void ValidateRestPathForEntity(string entityName, JsonElement restPathElement, HashSet<string> restPathsForEntities)
+        {
+            if (restPathElement.ValueKind is JsonValueKind.Null)
```
seantleonard, line 321: this is when:
```json
{
    "path": null
}
```
and not when the path property is completely left out of the config right? so if it is explicitly left out, then we would somehow resolve the path value to be null, will we?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +306,128 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
         }
 
+        /// <summary>
+        /// Helper method to validate that the rest path is correctly configured for the entity.
+        /// The rest path can only be a boolean value or a string.
+        /// If configured as a string, it should not be null/empty and should not conflict with the rest path
+        /// configured for any other entity.
+        /// </summary>
+        /// <param name="entityName">Name of the entity.</param>
+        /// <param name="restPathElement">The rest path element for the entity.</param>
```
seantleonard, line 316: perhaps a rename from `restPathElement` to a more generic name because this object can be a boolean, which does not represent a rest path, it just means rest is enabled/disabled for the entity.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +306,128 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+                    $"Accepted data types are: string, boolean.",
+                    statusCode: HttpStatusCode.ServiceUnavailable,
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError
+                    );
+            }
+
+            if (restPathElement.ValueKind is JsonValueKind.String)
+            {
+                string path = restPathElement.ToString();
+                if (string.IsNullOrEmpty(path))
```
seantleonard, line 346: would we want to check for whitespace here as well?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +306,128 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+                {
+                    // Presence of multiple entities having the same rest path configured causes conflict.
+                    throw new DataApiBuilderException(
+                        message: $"Multiple entities found with same rest {RestEntitySettings.PROPERTY_PATH}: {path}.",
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError
+                        );
+                }
+
+                restPathsForEntities.Add(path);
```
seantleonard, line 366:
```suggestion
                if (!restPathsForEntities.Add(path))
                {
                    // Presence of multiple entities having the same rest path configured causes conflict.
                    throw new DataApiBuilderException(
                        message: $"Multiple entities found with same rest {RestEntitySettings.PROPERTY_PATH}: {path}.",
                        statusCode: HttpStatusCode.ServiceUnavailable,
                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError
                        );
                }
```

possible simplification

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +306,128 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+        }
+
+        /// <summary>
+        /// Helper method to validate that the Rest methods are correctly configured for the entity.
+        /// Rest methods can only be configured for stored procedures as an array of valid REST operations.
+        /// </summary>
+        /// <param name="entityName">Name of the entity.</param>
+        /// <param name="restMethodsElement">Rest methods element configured for the entity.</param>
+        /// <param name="entity">Entity object.</param>
+        /// <exception cref="DataApiBuilderException">Throws exception whenever a validation fails.</exception>
```
seantleonard, line 377: "Throws exception whenever a validation fails." should be more specific because an exception inherently means that validation failed, we'd want to answer "for what validation did this function fail, thus resulting in the exception"

seantleonard (changes requested):
Thank you for adding these validations. Few areas where tests could be organized differently to fit the test description and additional consideration of top-level entity name needed for validating the rest path value.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: But even in that case,  `"path": true`. So, either we change the feature name in config and here or this will be fine as is. I have mentioned in the comments as well, what are the expected value types the restPath can take.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: > and not when the path property is completely left out of the config right?

Right. No, we would not resolve to path as null if it is explicitly left out. In that case we won't even do the validations related to `path`.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: > Defining REST 'path' indicates that the developer doesn't want the top-level entity name to be the rest path value

Not always. It can just be set as `path: true`. But yes, valid point. We need to account for entity names as well.

src/Config/Entity.cs
ayush3797: yes

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Wow, I was looking for something like `TryAdd`. Wasn't aware that `Add` does it for us here. Thanks!

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Done.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -239,25 +239,70 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(IDict
...
             }
         }
 
         /// <summary>
-        /// Check whether the entity name defined in runtime config only contains
-        /// characters allowed for GraphQL names.
-        /// Does not perform validation for entities which do not
+        /// Check whether the entity configuration defined in runtime config only contains characters allowed for GraphQL names
+        /// and other validations related to rest path and methods configured for the entity.
+        /// The GraphQL validation is not performed for entities which do not
```
Aniruddh25, line 251: Similar to GraphQL, is the REST validation not performed as well if entity.rest = false?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -239,25 +239,70 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(IDict
...
-        /// <param name="runtimeConfig"></param>
-        public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entityCollection)
+        /// <param name="runtimeConfig">The runtime configuration.</param>
+        public static void ValidateEntityConfiguration(RuntimeConfig runtimeConfig)
         {
-            foreach (string entityName in entityCollection.Keys)
+            // Stores the unique rest paths configured for different entities present in the config.
+            HashSet<string> restPathsForEntities = new();
+
+            foreach (KeyValuePair<string, Entity> Entity in runtimeConfig.Entities)
```
Aniruddh25, line 261: nit: calling the pair as Entity seems inappropriate since the value is the Entity. Also, the name of this pair variable should be camelCased since its a function local variable.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -239,25 +239,70 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(IDict
...
             {
-                Entity entity = entityCollection[entityName];
+                string entityName = Entity.Key;
+                Entity entity = Entity.Value;
+
+                // If no custom rest path is defined for the entity, we default it to the entityName.
+                string pathForEntity = entityName;
+                if (entity.Rest is not null)
```
Aniruddh25, line 268: I wonder with the overhaul of config system PR, this code might need to be modified.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -239,25 +239,70 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(IDict
...
+                    {
+                        throw new DataApiBuilderException(
+                        message: $"The 'rest' property for entity: {entityName} can only be a boolean value or a json object.",
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError
+                        );
+                    }
+                }
+
+                if (!restPathsForEntities.Add(pathForEntity))
```
Aniruddh25, line 297: What if for EntityA rest is set to false, and for EntityB the path is set to EntityA, will it allow such a config or will it fail? What should be the desired behavior?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +328,120 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
             }
         }
 
+        /// <summary>
+        /// Helper method to get the rest path for the entity if it is correctly configured.
+        /// The rest path can only be a boolean value or a string.
```
Aniruddh25, line 333: Im not sure if path can be a boolean value. The parent `rest` can be but not the path. What does it mean if its a boolean?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +328,120 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+        }
+
+        /// <summary>
+        /// Helper method to validate that the Rest methods are correctly configured for the entity.
+        /// Rest methods can only be configured for stored procedures as an array of valid REST operations.
+        /// </summary>
+        /// <param name="entityName">Name of the entity.</param>
+        /// <param name="restMethodsElement">Rest methods element configured for the entity.</param>
+        /// <param name="entity">Entity object.</param>
+        /// <exception cref="DataApiBuilderException">Throws exception whenever a the rest methods are configured for a non-stored procedure entity or
```
Aniruddh25, line 390:
```suggestion
        /// <exception cref="DataApiBuilderException">Throws exception whenever the rest methods are configured for a non-stored procedure entity or
```

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -239,25 +239,70 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(IDict
...
+        /// Check whether the entity configuration defined in runtime config only contains characters allowed for GraphQL names
+        /// and other validations related to rest path and methods configured for the entity.
+        /// The GraphQL validation is not performed for entities which do not
         /// have GraphQL configuration: when entity.GraphQL == false or null.
         /// </summary>
         /// <seealso cref="https://spec.graphql.org/October2021/#Name"/>
-        /// <param name="runtimeConfig"></param>
-        public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entityCollection)
+        /// <param name="runtimeConfig">The runtime configuration.</param>
+        public static void ValidateEntityConfiguration(RuntimeConfig runtimeConfig)
```
Aniruddh25, line 256: Do you need the complete runtime config as input to this function? Seems like only the entity name value pairs is what you need, so the previous input Dictionary seems to not required to change.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +328,120 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+        /// </summary>
+        /// <param name="entityName">Name of the entity.</param>
+        /// <param name="restMethodsElement">Rest methods element configured for the entity.</param>
+        /// <param name="entity">Entity object.</param>
+        /// <exception cref="DataApiBuilderException">Throws exception whenever a the rest methods are configured for a non-stored procedure entity or
+        /// contain an unexpected value.</exception>
+        private static void ValidateRestMethodsForEntity(string entityName, JsonElement restMethodsElement, Entity entity)
+        {
+            // This is needed to correctly populate the source type for the entity.
+            entity.TryPopulateSourceFields();
```
Aniruddh25, line 395: Shouldn't this population be done before calling this validation workflow? What happens if in Production mode since we wont do any of these validations, who will be calling TryPopulateSourceFields?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +328,120 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+            if (entity.ObjectType is not SourceType.StoredProcedure)
```
Aniruddh25, line 396: Do we allow `"methods" : null` for non-stored procedure entities? should this verify first that restMethodsElement is not of JsonValueKind.Null?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +328,120 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
+            {
+                // The rest property 'methods' can only be present for stored procedures.
+                throw new DataApiBuilderException(
+                    message: $"The rest property '{RestStoredProcedureEntitySettings.PROPERTY_METHODS}' present for entity: {entityName} " +
+                    $"of type: {entity.ObjectType} is only valid for type: {SourceType.StoredProcedure}.",
```
Aniruddh25, line 401:
```suggestion
                    $"of type: {entity.ObjectType}, but is only valid for type: {SourceType.StoredProcedure}.",
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1755,189 @@ string[] linkingTargetFields
...
+        /// <summary>
+        /// Test to validate that when multiple entities have the same custom rest path configured, we throw an exception.
+        /// </summary>
+        /// <param name="exceptionExpected">Whether an exception is expected as a result of test run.</param>
+        /// <param name="restPathForEntity">Custom rest path to be configured for the first entity.</param>
+        /// <param name="restPathForSecondEntity">Custom rest path to be configured for the second entity.</param>
+        /// <param name="expectedExceptionMessage">The expected exception message.</param>
+        [DataTestMethod]
+        [DataRow(true, "", "Entity: EntityA has an empty rest path.",
+            DisplayName = "Empty rest path configured for an entitiy fails config validation.")]
```
Aniruddh25, line 1854:
```suggestion
            DisplayName = "Empty rest path configured for an entity fails config validation.")]
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1755,189 @@ string[] linkingTargetFields
...
+        [DataRow(true, null, "Entity: EntityA has a null rest path. Accepted value types are: string, boolean.",
+            DisplayName = "NULL rest path configured for an entitiy fails config validation.")]
```
Aniruddh25, line 1856:
```suggestion
            DisplayName = "NULL rest path configured for an entity fails config validation.")]
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1755,189 @@ string[] linkingTargetFields
...
+        [DataRow(true, 1, $"Entity: EntityA has rest path specified with incorrect data type. Accepted data types are: string, boolean.",
+            DisplayName = "Rest path configured as integer for an entitiy fails config validation.")]
```
Aniruddh25, line 1858:
```suggestion
            DisplayName = "Rest path configured as integer for an entity fails config validation.")]
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1755,189 @@ string[] linkingTargetFields
...
+        public void ValidateRestPathForEntityInConfig(
```
Aniruddh25, line 1859: any test cases with no exception? if not, we dont need the boolean argument exceptionExpected

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1755,189 @@ string[] linkingTargetFields
...
+        /// <summary>
+        /// Test to validate that when multiple entities have the same custom rest path configured, we throw an exception.
+        /// </summary>
+        /// <param name="exceptionExpected">Whether an exception is expected as a result of test run.</param>
+        /// <param name="restPathForFirstEntity">Custom rest path to be configured for the first entity.</param>
+        /// <param name="restPathForSecondEntity">Custom rest path to be configured for the second entity.</param>
+        /// <param name="expectedExceptionMessage">The expected exception message.</param>
+        [DataTestMethod]
+        [DataRow(false, "restPathA", "restPathB", DisplayName = "Unique rest paths configured for entities pass config validation.")]
+        [DataRow(false, true, false, DisplayName = "Rest path configured as boolean values for an entities pass config validation.")]
```
Aniruddh25, line 1901: Can you please double check our official configuration file docs/json schema if we allow boolean for path. I thought it was only allowed for `rest`

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1747,5 +1755,189 @@ string[] linkingTargetFields
...
+        [DataRow(true, "restPath", "restPath", "The rest path: restPath specified for entity: EntityB is already used by another entity.",
```
Aniruddh25, line 1902: Can you please add a test for the scenario where `rest` is false for `EntityA` and for `EntityB` the path is `EntityA`

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: You're correct. I checked in the draft schema as well. Will update this.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: You're correct. Will do the required change.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Thanks for pointing, added test without exception.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: This method is called by `ValidateEntityConfiguration` which is executed only when we are operating in development mode.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Not relevant anymore as I misinterpreted that the rest path can be a boolean. It can only be a string.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: I do see a lot of entity configuration in the config file where the rest.path is set as `true` when it should have been a string. I think that's how I got the impression that the rest.path can be a bool/string. So this now becomes a breaking change to not support boolean values for rest.path. Need to discuss this further.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Properties like runtimeConfig.RestGlobalSettings.Enabled and runtimeConfig.GraphQLGlobalSettings.Enabled are needed. I can instead use the existing dictionary parameter but then I would need two different methods(and loops) for GQL and REST validations. So, I thought I'll change to use runtimeConfig which provides flexibility for any potential future dependencies on values in the config outside the entity's configuration.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Yeah it shouldn't be allowed to specify this as null (if it is specified at all- we're fine if the methods property is omitted). The check will be done only for storedprocedure entities. For non-sp entities, we'll have the above exception.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Done. The validation will pass in this case. The test case is added in this [commit](https://github.com/Azure/data-api-builder/pull/1496/commits/04c92078af497cba866261838964e07ea8346a86).

src/Config/Entity.cs
aaronpowell, line 1: This file will be impacted by #1402 as the type structure has changed and how we deserialize has changed.

src/Service.Tests/Authorization/AuthorizationHelpers.cs
```diff
@@ -137,6 +129,25 @@ public static AuthorizationResolver InitAuthorizationResolver(RuntimeConfig runt
             return runtimeConfig;
         }
 
+        /// <summary>
+        /// Helper method to create runtime settings with given authentication provider.
+        /// </summary>
+        /// <param name="authProvider">Authentication provider.</param>
+        /// <returns>Created runtime settings.</returns>
+        public static Dictionary<GlobalSettingsType, object> CreateRuntimeSettings(string authProvider = "AppService")
```
aaronpowell, line 137: #1402 will see this revisited as we now have a type rather than a dictionary for these settings.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -239,25 +243,102 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(IDict
...
-        /// <param name="runtimeConfig"></param>
-        public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entityCollection)
+        /// <param name="runtimeConfig">The runtime configuration.</param>
+        public static void ValidateEntityConfiguration(RuntimeConfig runtimeConfig)
         {
-            foreach (string entityName in entityCollection.Keys)
+            // Stores the unique rest paths configured for different entities present in the config.
+            HashSet<string> restPathsForEntities = new();
+
+            foreach ((string entityName, Entity entity) in runtimeConfig.Entities)
```
aaronpowell, line 265: With #1402 we should be able to simplify this loop greatly as we no longer need to do our own deserialization of the objects.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -283,6 +364,119 @@ public static void ValidateEntityNamesInConfig(Dictionary<string, Entity> entity
...
 
+        /// <summary>
+        /// Helper method to get the rest path/rest enabled properties for the entity if it is correctly configured.
+        /// The rest path should not be null/empty and should not conflict with the rest path
+        /// configured for any other entity.
+        /// </summary>
+        /// <param name="entityName">Name of the entity.</param>
+        /// <param name="restPathElement">The rest path element for the entity.</param>
+        /// <exception cref="DataApiBuilderException">Throws exception when rest path contains an unexpected value.</exception>
+        private static Tuple<bool,string> ValidateAndGetRestSettingsForEntity(string entityName, JsonElement restPathElement)
```
aaronpowell, line 375: As with the `foreach` loop above, this should be simplified by #1402.

src/Service/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -424,48 +424,20 @@ private static string GetEntityPath(Entity entity, string entityName)
                 return string.Empty;
             }
 
-            // otherwise we have to convert each part of the Rest property we want into correct objects
-            // they are json element so this means deserializing at each step with case insensitivity
-            JsonSerializerOptions options = RuntimeConfig.SerializerOptions;
             JsonElement restConfigElement = (JsonElement)entity.Rest;
-            if (entity.ObjectType is SourceType.StoredProcedure)
+            if (restConfigElement.TryGetProperty(RestEntitySettings.PROPERTY_PATH, out JsonElement path))
```
aaronpowell, line 428: With #1402 we won't be working with `JsonElement` so this should be simplified.

aaronpowell (changes requested):
I think this PR might be greatly impacted by #1402 as the validation is limited by the fact that we end up with `JsonElement` being passed around in the deserialized object rather than a single type setup.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1326,13 +1338,15 @@ private static void ValidateExceptionForDuplicateQueriesDueToEntityDefinitions(S
         /// <summary>
         /// Method to create a sample entity with GraphQL enabled,
         /// with given source and relationship Info.
+        /// Rest is disabled by default, unless specified otherwise.
```
seantleonard, line 1341: "Rest is disabled by default", only for this test and not in the context of the runtime config schema?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -229,25 +232,96 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(Runti
...
+        private static void ValidateAndGetRestSettingsForEntity(string entityName, string pathForEntity)
+        {
+            if (string.IsNullOrEmpty(pathForEntity))
+            {
+                // The rest 'path' cannot be empty.
+                throw new DataApiBuilderException(
+                    message: $"The rest path for entity: {entityName} cannot be empty.",
+                    statusCode: HttpStatusCode.ServiceUnavailable,
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError
+                    );
```
seantleonard, line 312: Would this ever be triggered when the user explicitly does not include a rest path? e.g. they don't want to customize the rest path.

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -36,13 +36,13 @@ public class RuntimeConfigValidator : IConfigValidator
...
         private static readonly Regex _invalidClaimCharsRgx = new(_invalidClaimChars, RegexOptions.Compiled);
 
-        // "Reserved characters as defined in RFC3986 are not allowed to be present in the
-        // REST/GraphQL custom path because they are not acceptable to be present in URIs.
-        // " Refer here: https://www.rfc-editor.org/rfc/rfc3986#page-12.
-        private static readonly string _invalidPathChars = @"[\.:\?#/\[\]@!$&'()\*\+,;=]+";
+        // Reserved characters as defined in RFC3986 are not allowed to be present in the
+        // REST/GraphQL URI path because they are not acceptable to be present in URIs.
+        // Refer here: https://www.rfc-editor.org/rfc/rfc3986#page-12.
+        private static readonly string _invalidURIComponentChars = @"[\.:\?#/\[\]@!$&'()\*\+,;=]+";
```
seantleonard, line 42:
```suggestion
        private static readonly string _invalidUriCharacters = @"[\.:\?#/\[\]@!$&'()\*\+,;=]+";
```

Since these reserved characters can't be included in URI's, this regex would be appropriate for any URI checking?

src/Service/Configurations/RuntimeConfigValidator.cs
seantleonard: > > Defining REST 'path' indicates that the developer doesn't want the top-level entity name to be the rest path value
> 
> Not always. It can just be set as `path: true`. But yes, valid point. We need to account for entity names as well.

Do we actually allow `path: true` ? After reviewing the config refactor PR, we double checked the runtime config schema to see that 'true' wasn't valid.

https://github.com/Azure/data-api-builder/blob/aa5c8a33bfa4213a87afc0b5a7a31dc2db0cf954/schemas/dab.draft.schema.json#L278-L317

src/Service.Tests/GraphQLBuilder/Sql/SchemaConverterTests.cs
```diff
@@ -741,7 +741,7 @@ public static Entity GenerateEmptyEntity(string entityName)
         {
             return new Entity(
                 Source: new($"{SCHEMA_NAME}.{TABLE_NAME}", EntitySourceType.Table, null, null),
-                Rest: new(EntityRestOptions.DEFAULT_SUPPORTED_VERBS),
+                Rest: new(new SupportedHttpVerb[] { }),
```
seantleonard, line 744: empty?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1841,5 +1855,202 @@ string[] linkingTargetFields
...
+        /// Rest methods can only be configured for stored procedures as an array of valid REST operations. 
+        /// </summary>
+        /// <param name="sourceType">The source type of the entity.</param>
+        /// <param name="methods">Value of the rest methods property configured for the entity.</param>
+        /// <param name="exceptionExpected">Boolean value representing whether an exception is expected or not.</param>
+        /// <param name="expectedErrorMessage">Expected error message when an exception is expected for the test run.</param>
+        [DataTestMethod]
+        /*[DataRow(EntitySourceType.Table, "[\"get\"]", true,
+            $"The rest property '{Entity.PROPERTY_METHODS}' is present for entity: HybridEntity of type: Table, but is only valid for type: StoredProcedure.",
+            DisplayName = "Rest methods specified for non-storedprocedure entity fail config validation.")]*/
```
seantleonard, line 1870: commented out code can be removed.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1841,5 +1855,202 @@ string[] linkingTargetFields
...
+            {
+                DataApiBuilderException ex =
+                    Assert.ThrowsException<DataApiBuilderException>(() => RuntimeConfigValidator.ValidateEntityConfiguration(runtimeConfig));
+                Assert.AreEqual(expectedErrorMessage, ex.Message);
+                Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
+                Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ConfigValidationError, ex.SubStatusCode);
+            }
+            else
+            {
+                RuntimeConfigValidator.ValidateEntityConfiguration(runtimeConfig);
```
seantleonard, line 1940: Should there be an assert here? This being false would be unexpected and may still fail the test. However, would the logging of why the test failed give enough information to troubleshoot? e.g. `Assert(expected, actual, failureMessage).`

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1841,5 +1855,202 @@ string[] linkingTargetFields
...
+            }
+        }
+
+        /// <summary>
+        /// Test to validate that when multiple entities have the same custom rest path configured, we throw an exception.
+        /// </summary>
+        /// <param name="exceptionExpected">Whether an exception is expected as a result of test run.</param>
+        /// <param name="restPathForEntity">Custom rest path to be configured for the first entity.</param>
+        /// <param name="restPathForSecondEntity">Custom rest path to be configured for the second entity.</param>
+        /// <param name="expectedExceptionMessage">The expected exception message.</param>
```
seantleonard, line 1950: Summary looks to be inadvertently copy/pasted from a different test. needs revision

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1841,5 +1855,202 @@ string[] linkingTargetFields
...
+        [DataRow(false, "restPathA", "restPathB", DisplayName = "Unique rest paths configured for entities pass config validation.")]
+        [DataRow(true, "restPath", "restPath", "The rest path: restPath specified for entity: EntityB is already used by another entity.",
+            DisplayName = "Duplicate rest paths configures for entities fail config validation.")]
+        [DataRow(true, null, "EntityA", "The rest path: EntityA specified for entity: EntityB is already used by another entity.",
+            DisplayName = "Rest path for an entity configured as the name of another entity fails config validation.")]
+        public void ValidateUniqueRestPathsForEntitiesInConfig(
+            bool exceptionExpected,
+            string restPathForFirstEntity,
+            string restPathForSecondEntity,
+            string expectedExceptionMessage = "")
```
seantleonard, line 2011: Maybe covered in a different pre-existing test: would the validation method raise an exception when the rest path of entityB matches the top level entity name of entityA?

i.e.
toplevel entity name: "entityA"
restpath: "entityA"

toplevel entity name: "entityB"
restpath: "entityA"

seantleonard (changes requested):
Some questions and requested revisions.

src/Config/ObjectModel/Entity.cs
```diff
@@ -20,4 +20,8 @@ public record Entity(
     EntityRestOptions Rest,
     EntityPermission[] Permissions,
     Dictionary<string, string>? Mappings,
-    Dictionary<string, EntityRelationship>? Relationships);
+    Dictionary<string, EntityRelationship>? Relationships)
+{
+    public const string PROPERTY_PATH = "path";
```
Aniruddh25, line 25: these are properties of `EntityRestOptions`, can't we do `nameof(EntityRestOptions.Path)` `nameof(EntityRestOptions.Methods)`?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -229,25 +232,96 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(Runti
...
+                else if (entity.GraphQL.Enabled)
                 {
                     ValidateNameRequirements(entity.GraphQL.Singular);
                     ValidateNameRequirements(entity.GraphQL.Plural);
                 }
             }
         }
 
+        /// <summary>
+        /// Helper method to get the rest path/rest enabled properties for the entity if it is correctly configured.
```
Aniruddh25, line 296: The summary no longer represents accurately what the function is doing

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -229,25 +232,96 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(Runti
...
             {
-                if (entity.GraphQL.Enabled)
+                if (runtimeConfig.Runtime.Rest.Enabled)
+                {
+                    // By default we assume Rest is enabled for the entity.
+                    bool isRestEnabledForEntity = true;
+
+                    // If no custom rest path is defined for the entity, we default it to the entityName.
+                    string pathForEntity = entityName;
+                    if (entity.Rest is not null)
```
Aniruddh25, line 256: can we add the check for enabled directly in this if:

 `if (entity.Rest is not null && entity.Rest.Enabled)`

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -229,25 +232,96 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(Runti
...
+                if (!runtimeConfig.Runtime.GraphQL.Enabled)
+                {
+                    continue;
+                }
+
+                if (entity.GraphQL is null)
+                {
+                    continue;
+                }
+                else if (entity.GraphQL.Enabled)
```
Aniruddh25, line 287: instead of doing 2 continues, can we simplify this into a single if statement?

`if (runtimeConfig.Runtime.GraphQL.Enabled && entity.GraphQL is not null && entity.GraphQL.Enabled)` ?

src/Service/Configurations/RuntimeConfigValidator.cs
```diff
@@ -229,25 +232,96 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(Runti
...
+                    // By default we assume Rest is enabled for the entity.
+                    bool isRestEnabledForEntity = true;
+
+                    // If no custom rest path is defined for the entity, we default it to the entityName.
+                    string pathForEntity = entityName;
+                    if (entity.Rest is not null)
+                    {
+                        pathForEntity = entity.Rest.Path is not null ? entity.Rest.Path.TrimStart('/') : entityName;
+                        isRestEnabledForEntity = entity.Rest.Enabled;
+                        SupportedHttpVerb[] supportedOperations = entity.Rest.Methods;
```
Aniruddh25, line 260: is this local variable `supportedOperations` used anywhere in validations? Am I missing something?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1841,5 +1855,202 @@ string[] linkingTargetFields
...
+            bool exceptionExpected,
+            string restPathForEntity,
+            string expectedExceptionMessage = "")
+        {
+            Dictionary<string, Entity> entityMap = new();
+            Entity sampleEntity = GetSampleEntityUsingSourceAndRelationshipMap(
+                source: "TEST_SOURCEA",
+                relationshipMap: null,
+                graphQLDetails: null,
+                restDetails: new(new SupportedHttpVerb[] { }, restPathForEntity, true)
```
Aniruddh25, line 1965: Can we test if the entity name itself contains an invalid character and when that name becomes the restPath for the entity, would we give a validation error?

src/Service.Tests/GraphQLBuilder/Sql/SchemaConverterTests.cs
ayush3797: Yeah, for table, rest methods should ideally be empty because they are meant for stored procedures.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Yeah we can. A correction here, the condition would be:

`if (!runtimeConfig.Runtime.GraphQL.Enabled || entity.GraphQL is null || !entity.GraphQL.Enabled)`

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Even if Rest is disabled we still need to perform the validations related to path done by the method `ValidateRestPathSettingsForEntity`. So, we cannot club the two conditions into one if.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Yeah correct it isn't.

src/Config/ObjectModel/Entity.cs
ayush3797: The properties in `EntityRestOptions` start with uppercase P/M. We need lowercased characters in the property names. Using `nameof(EntityRestOptions.Path).ToLower()` makes the expression non-constant. So, we cannot do it. Also, we cannot rename the property from Path->path as the property name has to start with a capital letter - that's a naming rule.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Yes.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Yeah

src/Service.Tests/GraphQLBuilder/Sql/SchemaConverterTests.cs
ayush3797: We don't allow anything other than string as a value for `rest.path`. Any other datatype would fail deserialization.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Going to use ignore on this test right now, because with Aaron's PR, this will pass even for tables/views - when it shouldn't. Once we have that change in, we can un-ignore this.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: `ValidateEntityConfiguration` method is of void return type just like other methods in runtimeconfigvalidator `ValidatePermissionsInConfig`.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: So the rest path will always be there - either it would be the top level entity name or the custom rest path. The validation will be performed in either cases.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Already covered here:

https://github.com/Azure/data-api-builder/blob/f0aba3cd06f8ed74bd0f85f8ac7d7cd11ea019a6/src/Service.Tests/Unittests/ConfigValidationUnitTests.cs#L2005

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Sure.

ayush3797: Added

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -226,25 +229,85 @@ public static void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(Runti
...
+                    message: $"The rest path for entity: {entityName} cannot be empty.",
+                    statusCode: HttpStatusCode.ServiceUnavailable,
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError
+                    );
+            }
+
+            if (_invalidUriCharactersRgx.IsMatch(pathForEntity))
+            {
+                throw new DataApiBuilderException(
+                    message: $"The rest path: {pathForEntity} for entity: {entityName} contains one or more reserved characters.",
```
seantleonard, line 304: nit: does http spec call 'reserved' characters 'reserved' or 'restricted'? subtle difference but aligning to spec would be more informative to the end user.

src/Service/Configurations/RuntimeConfigValidator.cs
seantleonard: Can you add a comment in the code for "Even if Rest is disabled we still need to perform the validations related to path done by the method" and add an explanation "why?"

seantleonard (approved):
lgtm. Just one nit from me and Ani's comments remain.

src/Service/Configurations/RuntimeConfigValidator.cs
Aniruddh25: why would we need to do those path validations if rest is disabled?

src/Config/ObjectModel/Entity.cs
Aniruddh25: Where is `PROPERTY_PATH` string used? I couldn't find it when I searched. Is it for error messaging only? Do we really need a `const` string? 

Also, if at all we still need it, wouldn't `EntityRestOptions` a better class to have these defined?

ayush3797: Moved the properties to the entityrestoptions class

src/Core/Configurations/RuntimeConfigValidator.cs
ayush3797: Yeah they call it reserved characters: https://www.rfc-editor.org/rfc/rfc3986#section-2.2

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Yeah here: https://github.com/Azure/data-api-builder/issues/1570

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: I thought this is what we discussed, Ani. The user might disable rest for an entity, but the paths for entities have to be unique irrespective of that. And similarly, the path must not contain reserved characters or be empty.

src/Config/ObjectModel/EntityRestOptions.cs
```diff
@@ -18,5 +18,8 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 [JsonConverter(typeof(EntityRestOptionsConverter))]
 public record EntityRestOptions(SupportedHttpVerb[] Methods, string? Path = null, bool Enabled = true)
 {
+    public const string PROPERTY_PATH = "path";
+    public const string PROPERTY_METHODS = "methods";
+    public const string PROPERTY_ENABLED = "enabled";
```
seantleonard, line 23: Is there an alternative approach to manually creating properties for each record property? i.e. similar to what we had in the past, [using jsonpropertyname attribute](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#:~:text=You%20may%20want,to%20the%20parameter.):
```csharp
/// <summary>
/// Person record type
/// </summary>
/// <param name="FirstName">First Name</param>
/// <param name="LastName">Last Name</param>
/// <remarks>
/// The person type is a positional record containing the
/// properties for the first and last name. Those properties
/// map to the JSON elements "firstName" and "lastName" when
/// serialized or deserialized.
/// </remarks>
public record Person([property: JsonPropertyName("firstName")] string FirstName, 
    [property: JsonPropertyName("lastName")] string LastName);
```

seantleonard (commented):
question about alternative approach to creating record type properties just for lowercase string

src/Service/Configurations/RuntimeConfigValidator.cs
Aniruddh25: Its been a long time since we discussed this, so I don't remember what the reasoning behind this decision was. At this point, it seems unnecessary to validate if the path of an entity contains reserved characters/or is empty if the entity is disabled for rest. Do you remember the reasoning?

src/Config/ObjectModel/EntityRestOptions.cs
ayush3797: This approach helped in deserializing and to help know the mapping from `firstName` in json to `FirstName` in object. Essentially we cannot refer to the jsonpropertyname anywhere. Like in EntityRestOptionsConverter.cs class since it is an attribute.

Aniruddh25 (approved):
LGTM, after resolving remaining comments.

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: The exact reasoning I am unable to recall but one argument in favour of validating is - we are validating that two entities do not have duplicate paths even if rest is disabled on both of them. Now suppose if the path had reserved characters, then it would give the impression from the exception that it was configured correctly- just that two entities had the same path configured lead to the exception, which is not the case.

Otherwise, yes. I don't feel and felt like we should be doing validations when rest is disabled. And I think the duplicate path validation should also be done only when rest is enabled on both of the entities.

Aniruddh25: With your latest changes, isn't duplicate path validation only happening when rest is enabled for both of the entities?
line 260 `if (isRestEnabledForEntity && !restPathsForEntities.Add(pathForEntity))` here: https://github.com/Azure/data-api-builder/blob/381a893774282c9ccb061f75b318a60e782fbb2a/src/Core/Configurations/RuntimeConfigValidator.cs#L260C22-L260C22

seantleonard (approved):

src/Service/Configurations/RuntimeConfigValidator.cs
ayush3797: Doing only when rest is enabled now.

PR #1497, opened 2023-05-16
Add support for .env file

### LINKED ISSUE #1374: Add support for providing environment var/values in a `.env` file read by `dotenv.net`
### Discussed in https://github.com/Azure/data-api-builder/discussions/1361

<div type='discussions-op-text'>

<sup>Originally posted by **glaucia86** March 25, 2023</sup>

> Hi! 
> 
> I would like to propose that we can make use of the .env file in the `dab.config.json` file.
> As it already happens with the files generated by the SWA CLI: `staticwebapp.database.config.json` (example: **[HERE](https://github.com/Azure-Samples/dab-swa-library-demo/blob/main/swa-db-connections/staticwebapp.database.config.json#L8)**)
> 
> Because the connection string is very exposed, where it has extremely sensitive data such as: login, password and database name.
> 

</div>

This issue is to add support for `.env` so that I can just put my sensitive data into the .env, if I don't need the complexity of having multiple configuration files. 

This is separate from the current `@env()` feature where we need to set environment variables on the system.

abhishekkumams:
## Why make this change?

- Closes #1374 
  - To allow users maintain a local `.env` to manage environment variables during local development.

## What is this change?

- Using the package [DotNetEnv](https://www.nuget.org/packages/DotNetEnv/)
- Now, users can maintain a local file `.env` to manage environment variables instead of setting environment variables in the system.
- If `.env` file is used, it will take precedence over system environment variables.

## How was this tested?

- [x] Unit Tests

## NOTE:
- This feature is to enhance local development experience, and it works when used via DAB CLI.

## QnA

1. Where should the `.env` file be created?
-> For `.env file` to be considered, it should be present in the same directory where the dab commands are executed.

2. Does it overwrite the existing environment variables?
-> No, it doesn't update any existing variables. It just gives precedence to the variable defined in the file over the system defined variable in case it is present in both places.


src/Cli.Tests/EndToEndTests.cs
```diff
@@ -631,6 +631,75 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+    /// <summary>
+    /// Test to verify that .env is supported by our CLI.
+    /// Below test is covering the following scenarios:
+    /// 1. Setting up environment variable in the system is consumed correctly.
+    /// 2. .env file if available will use the variable values from the .env file.
+    /// 3. If same variable is defined in both system and .env file, .env file takes precedence.
+    /// 4. If a variable is not present in the .env file then the system defined variable would be used, if defined.
+    /// </summary>
+    [TestMethod]
+    public void TestSupportForEnvironmentVariableFile()
```
aaronpowell, line 643: Looking through this test, it looks like we're testing a few different things here, first that we can read from `System.Environment` accurately, and then another that uses `DotEnv`.

We should split the tests into multiple smaller tests to it's easier to maintain.

seantleonard:
PR description: 
> If .env file is not used, it will take precedence over system environment variables.

should the line read: "If .env file **is** used" ?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -631,6 +631,75 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+                    ""hostingEnvValue"": ""PHOENIX_TEST""
+                }"),
+            JObject.Parse(resolvedJson)));
+
+        // Removing the .env file it will then use the value of system environment variable.
+        Environment.SetEnvironmentVariable(envVariableName, "TEST");
+        File.Delete("test.env");
+        DotNetEnv.Env.Load("test.env");
+        Assert.AreEqual("TEST", Environment.GetEnvironmentVariable(envVariableName));
+
```
seantleonard, line 700:
```suggestion
```

src/Cli/Program.cs
```diff
@@ -29,6 +29,8 @@ public static int Main(string[] args)
                 }
             );
 
+            DotNetEnv.Env.TraversePath().Load();
```
seantleonard, line 32: Usage of this library in the current implementation, overwrites environment variables on the system yes?
The PR description:

> Now, users can maintain a local file .env to manage environment variables instead of setting environment variables in the system.
> If .env file is not used, it will take precedence over system environment variables.

should clarify that users no longer manually have to modify env vars, because this package does it for them.
Saying that there are "precedence" rules seems like we are saying existing environment variables are not touched, but if the env var is defined in the .env file, it will be overwritten.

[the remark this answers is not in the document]
seantleonard: Do we need to handle cases where the folder traversal results in 'access denied?' so we can gracefully handle and display such an error? This concern is also applicable to the case where a dev provides a full path to the file.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -631,6 +631,75 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+        Assert.IsTrue(JToken.DeepEquals(
+            JObject.Parse(
+                @"{
+                    ""envValue"": ""DEVELOPMENT"",
+                    ""hostingEnvValue"": ""PHOENIX_TEST""
+                }"),
+            JObject.Parse(resolvedJson)));
+
+        // Removing the .env file it will then use the value of system environment variable.
+        Environment.SetEnvironmentVariable(envVariableName, "TEST");
```
seantleonard, line 696: We'll need to document that usage of .env file overwrites existing env vars.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -631,6 +631,75 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+        File.Delete("test.env");
+        DotNetEnv.Env.Load("test.env");
```
seantleonard, line 698: Does this line gracefully handle/hide a 'file not found' error?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -631,6 +631,75 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+
+        // If a variable is not present in the .env file then the system defined variable would be used if defined.
+        Environment.SetEnvironmentVariable("HOSTING_TEST_ENVIRONMENT", "PHOENIX_TEST");
+        resolvedJson = RuntimeConfigPath.ParseConfigJsonAndReplaceEnvVariables(
+            @"{
+                ""envValue"": ""@env('DAB_TEST_ENVIRONMENT')"",
+                ""hostingEnvValue"": ""@env('HOSTING_TEST_ENVIRONMENT')""
+                }"
+        );
+        Assert.IsNotNull(resolvedJson);
```
seantleonard, line 686: if `RuntimeConfigPath.ParseConfigJsonAndReplaceEnvVariables()` fails, would the test execution even reach this assert? e.g. an exception may be thrown for (all?) errors resulting in the call to the function mentioned?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -631,6 +631,75 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
 
     }
 
+    /// <summary>
```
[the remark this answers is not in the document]
seantleonard: especially given how the newly introduced package overwrites environment variables.

seantleonard (changes requested):
Some questions

src/Cli.Tests/EndToEndTests.cs
aaronpowell: No, it wouldn't as there's nothing to catch the exception.

Just as an aside - this will need to be refactored with #1402 as we don't directly expose that method.

src/Cli/Program.cs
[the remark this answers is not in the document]
aaronpowell: > Do we need to handle cases where the folder traversal results in 'access denied?' so we can gracefully handle and display such an error? This concern is also applicable to the case where a dev provides a full path to the file.

That is something I would expect `DotEnv` to handle for us and not something you'd need to handle in your own code.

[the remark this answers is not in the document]
seantleonard: I figured as much. After poking around their code, I don't think `DotEnv` handles all file IO errors, an example being Get: https://github.com/tonerdo/dotnet-env/blob/3dbee19770951a75785d398d02d7eed92261d0fe/src/DotNetEnv/Env.cs File IO not wrapped in try/catch so whatever access errors that may occur are bubbled up to our code.

abhishekkumams: we do not update the environment variables defined in the system.

src/Cli.Tests/EndToEndTests.cs
abhishekkumams: done

src/Cli/Program.cs
```diff
@@ -29,6 +29,8 @@ public static int Main(string[] args)
...
+
```
Aniruddh25, line 33: Did we get ACK and clarify the precedence requirements of .env over system environment variables?

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,119 @@
...
+        // Test environment variable is correctly resolved in the config file
+        string? resolvedJson = RuntimeConfigPath.ParseConfigJsonAndReplaceEnvVariables(jsonWithEnvVariable);
+        Assert.IsNotNull(resolvedJson);
+        Assert.IsTrue(JToken.DeepEquals(JObject.Parse(@"{""envValue"": ""TEST""}"), JObject.Parse(resolvedJson)));
+
+        // removing Environment variable from the System
+        Environment.SetEnvironmentVariable(envVariableName, null);
+        Assert.ThrowsException<DataApiBuilderException>(() =>
+            RuntimeConfigPath.ParseConfigJsonAndReplaceEnvVariables(jsonWithEnvVariable),
+            $"Environmental Variable, {envVariableName}, not found.");
```
aaronpowell, line 37: I'd create a separate test for this scenario because if an earlier `Assert` fails then you won't get this far, but it could also be failing for different reasons, which you won't know about.

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,119 @@
...
+        // Creating environment variable file
+        File.Create("test.env").Close();
+        File.WriteAllText("test.env", $"{envVariableName}=DEVELOPMENT");
+        DotNetEnv.Env.Load("test.env");
+
+        // Test environment variable is picked up from the .env file and is correctly resolved in the config file.
+        Assert.AreEqual("DEVELOPMENT", Environment.GetEnvironmentVariable(envVariableName));
+        string? resolvedJson = RuntimeConfigPath.ParseConfigJsonAndReplaceEnvVariables(jsonWithEnvVariable);
+        Assert.IsNotNull(resolvedJson);
+        Assert.IsTrue(JToken.DeepEquals(JObject.Parse(@"{""envValue"": ""DEVELOPMENT""}"), JObject.Parse(resolvedJson)));
```
aaronpowell, line 64:
```suggestion
        Assert.IsTrue(JToken.DeepEquals(JObject.Parse(@"{""envValue"": ""DEVELOPMENT""}"), JObject.Parse(resolvedJson)), "JSON resolved with environment variable correctly");
```

aaronpowell (changes requested):

src/Cli/Program.cs
abhishekkumams: Yes

src/Cli.Tests/EndToEndTests.cs
abhishekkumams: If file is not found, no error is thrown and it uses the system environment variables. This file is just an addition, not having this file is not an issue.

src/Cli/Program.cs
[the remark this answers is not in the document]
Aniruddh25: Wouldnt we need to add try catch within our code for such cases? @abhishekkumams, Is this comment resolved?

Aniruddh25:
As per @severussundar and discussion, I think we should add support for .env file name provided from user as well. Could be a follow up PR to support that option.

> I think we should consider getting the exact file name of the .env file from the user

Aniruddh25 (changes requested):
Thanks for clarifying the behavior, Waiting for few unanswered questions.

abhishekkumams:
> As per @severussundar and discussion, I think we should add support for .env file name provided from user as well. Could be a follow up PR to support that option.
> 
> > I think we should consider getting the exact file name of the .env file from the user

Agree, we can incorporate that based on the request as a follow up Pr.

src/Cli/Program.cs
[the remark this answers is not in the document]
abhishekkumams: Thanks everyone for the concern. We have moved out of traversePath and now using exact path (DotNetEnv.Env.Load()), which will not throw any IO exception as it is only using `File.Exist()`. Since, no exception would be thrown we do not need to wrap it around try/catch.

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        Environment.SetEnvironmentVariable(TEST_ENV_VARIABLE, "TEST");
+        Assert.IsFalse(File.Exists(".env"));
+        DotNetEnv.Env.Load(); // No error is thrown
+        Assert.AreEqual("TEST", Environment.GetEnvironmentVariable(TEST_ENV_VARIABLE));
+    }
+
+    /// <summary>
+    /// Test to verify that if the environment variables are not resolved correctly, runtime engine will not start.
+    /// Here, in the first scenario, engine failed to start because the variable defined in the environment file
+    /// is typed incorrectly and do not match with the one present in the config.
```
seantleonard, line 123:
```suggestion
    /// Here, in the first scenario, engine fails to start because the variable defined in the environment file
    /// is typed incorrectly and does not match the one present in the config.
```
nit

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        string? output = process.StandardOutput.ReadToEnd();
+        Assert.IsNotNull(output);
+
+        if (isFailure)
+        {
+            // Failed to resolve the environment variables in the config.
+            Assert.IsFalse(output.Contains("Starting the runtime engine..."));
+            Assert.IsTrue(output.Contains("Error: Failed due to: Environmental Variable, CONN_STRING, not found."));
+            Assert.IsTrue(output.Contains("Error: Failed to start the engine."));
```
seantleonard, line 158: I would argue that you should also assert that these two lines of console output are not present in the success scenario to directly validate your expectations, instead of relying on the presence of the string "starting the runtime engine"

seantleonard (dismissed):
some additional suggestions.

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        Assert.IsNotNull(resolvedJson);
+        Assert.IsTrue(JToken.DeepEquals(
+            JObject.Parse(
+                @"{
+                    ""envValue"": ""DEVELOPMENT"",
+                    ""hostingEnvValue"": ""PHOENIX_TEST""
+                }"),
+            JObject.Parse(resolvedJson)), "JSON resolved with environment variable correctly");
+
+        // Removing the .env file it will then use the value of system environment variable.
```
ayush3797, line 104: This comment looks misplaced. Did you forget to remove the file?

src/Cli/Utils.cs
```diff
@@ -608,13 +610,20 @@ public static bool TryGetRoleAndOperationFromPermission(IEnumerable<string> perm
...
+
+            try
+            {
+                // Tries to read the config and resolve environment variables.
+                runtimeConfigJson = GetRuntimeConfigJsonString(configFile);
+            }
+            catch (Exception e)
             {
-                _logger.LogError($"Failed to read the config file: {configFile}.");
+                _logger.LogError("Failed due to: {exceptionMessage}", e.Message);
```
ayush3797, line 622: Might be worth mentioning "What failed?"

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        Assert.AreEqual("DEVELOPMENT", Environment.GetEnvironmentVariable(TEST_ENV_VARIABLE));
+        string? resolvedJson = RuntimeConfigPath.ParseConfigJsonAndReplaceEnvVariables(jsonWithEnvVariable);
+        Assert.IsNotNull(resolvedJson);
+        Assert.IsTrue(JToken.DeepEquals(
+            JObject.Parse(@"{""envValue"": ""DEVELOPMENT""}"),
+            JObject.Parse(resolvedJson)), "JSON resolved with environment variable correctly");
+    }
+
+    /// <summary>
+    /// This test setups a environment variable in the system and also creates a .env file containing
```
ayush3797, line 71:
```suggestion
    /// This test setups an environment variable in the system and also creates a .env file containing
```

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// same variable with different value to show the value stored in .env file is given
```
ayush3797, line 72:
```suggestion
    /// the same variable with different value. In such a case, the value stored for variable in .env file is given
```

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// precedence over the system variable.
```
ayush3797, line 73:
```suggestion
    /// precedence over the value specified in the system.
```

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        Environment.SetEnvironmentVariable(TEST_ENV_VARIABLE, "TEST");
+        Assert.IsFalse(File.Exists(".env"));
+        DotNetEnv.Env.Load(); // No error is thrown
+        Assert.AreEqual("TEST", Environment.GetEnvironmentVariable(TEST_ENV_VARIABLE));
+    }
+
+    /// <summary>
+    /// Test to verify that if the environment variables are not resolved correctly, runtime engine will not start.
+    /// Here, in the first scenario, engine failed to start because the variable defined in the environment file
```
ayush3797, line 122:
```suggestion
    /// Here, in the first scenario, engine fails to start because the variable defined in the environment file
```

src/Cli.Tests/EnvironmentTests.cs
ayush3797, line 123:
```suggestion
    /// is typed incorrectly and does not match with the one present in the config.
```

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    public void TestSystemEnvironmentVariableIsUsedInAbsenceOfEnvironmentFile()
+    {
+        Environment.SetEnvironmentVariable(TEST_ENV_VARIABLE, "TEST");
+        Assert.IsFalse(File.Exists(".env"));
+        DotNetEnv.Env.Load(); // No error is thrown
+        Assert.AreEqual("TEST", Environment.GetEnvironmentVariable(TEST_ENV_VARIABLE));
+    }
+
+    /// <summary>
+    /// Test to verify that if the environment variables are not resolved correctly, runtime engine will not start.
```
ayush3797, line 121:
```suggestion
    /// Test to verify that if the environment variables are not resolved correctly, the runtime engine will not start.
```

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+                    ""hostingEnvValue"": ""PHOENIX_TEST""
+                }"),
+            JObject.Parse(resolvedJson)), "JSON resolved with environment variable correctly");
+
+        // Removing the .env file it will then use the value of system environment variable.
+
+    }
+
+    /// <summary>
+    /// Test to verify that no error is thrown if .env file is not present, and existing system variables is used.
```
ayush3797, line 109:
```suggestion
    /// Test to verify that no error is thrown if .env file is not present, and existing system variables are used.
```

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// correctly consumed.
+    /// For this test there were no existing environment variables. The values are picked up
+    /// directly from the `.env` file.
+    /// </summary>
+    [TestMethod]
+    public void TestEnvironmentFileIsConsumedCorrectly()
+    {
+        string jsonWithEnvVariable = @"{""envValue"": ""@env('DAB_TEST_ENVIRONMENT')""}";
+
+        // No environment File, No environment variable set in the system
```
ayush3797, line 53:
```suggestion
        // No environment File, no environment variable set in the system
```

ayush3797 (approved):
Some nits. LGTM otherwise!

src/Cli.Tests/EnvironmentTests.cs
abhishekkumams: sure

src/Cli/Utils.cs
abhishekkumams: it woukld be printing the same thing twice, the method called in the condition already logs the exception

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -99,13 +99,22 @@ public void TestPrecedenceOfEnvironmentFileOverExistingVariables()
...
+
+    /// <summary>
+    /// Test to verify that if .env file is not present then existing system variables will be used.
+    /// </summary>
+    [TestMethod]
+    public void TestSystemEnvironmentVariableIsUsedInAbsenceOfEnvironmentFile()
+    {
+        Environment.SetEnvironmentVariable(TEST_ENV_VARIABLE, "TEST");
+        Assert.IsFalse(File.Exists("test.env"));
+        DotNetEnv.Env.Load("test.env"); // No error is thrown
```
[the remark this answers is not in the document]
Aniruddh25: How was this resolved? I dont think we should call DotNetEnv.Env.Load in the test. Since that would mean we are testing this externally - not the actual usage of Load in DAB. Do we need to add such a test at all?

Aniruddh25 (changes requested):
Not sure why we are testing DotNetEnv.Env.Load externally.

abhishekkumams:
> Not sure why we are testing DotNetEnv.Env.Load externally.

We have tests that verifies DotNetEnv.Env.Load internally. [Line:125(Environment)](https://github.com/Azure/data-api-builder/blob/029ad357bd83b547d1c32f38f8f0ae4ae8adfe9c/src/Cli.Tests/EnvironmentTests.cs#L125)

we have some external tests as well to just verify the method `ParseConfigJsonAndReplaceEnvVariables` as calling this directly won't load the variables from .env file.

Loading from .env file happens at the initialization step of DAB CLI, so that every cli command is synced with the environment variables.

Hope this addresses your concerns?

Aniruddh25 (approved):

src/Cli.Tests/EnvironmentTests.cs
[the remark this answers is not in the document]
abhishekkumams: We have tests that verifies DotNetEnv.Env.Load internally. [Line:125(Environment)](https://github.com/Azure/data-api-builder/blob/029ad357bd83b547d1c32f38f8f0ae4ae8adfe9c/src/Cli.Tests/EnvironmentTests.cs#L125)

we have some external tests as well to just verify the method ParseConfigJsonAndReplaceEnvVariables as calling this directly won't load the variables from .env file.

Loading from .env file happens at the initialization step of DAB CLI, so that every cli command is synced with the environment variables.

Aniruddh25:
> > Not sure why we are testing DotNetEnv.Env.Load externally.
> 
> We have tests that verifies DotNetEnv.Env.Load internally. [Line:125(Environment)](https://github.com/Azure/data-api-builder/blob/029ad357bd83b547d1c32f38f8f0ae4ae8adfe9c/src/Cli.Tests/EnvironmentTests.cs#L125)
> 
> we have some external tests as well to just verify the method `ParseConfigJsonAndReplaceEnvVariables` as calling this directly won't load the variables from .env file.
> 
> Loading from .env file happens at the initialization step of DAB CLI, so that every cli command is synced with the environment variables.
> 
> Hope this addresses your concerns?

Thanks for the clarifications, could you comment the additional scenarios which this test is handling.

Issue #1420, opened 2023-04-06
[Bug]: Wrong NextLink if using with Azure Static Web Apps


sebader:
### What happened?

Not sure if this is the right place to address this but...:

When using DAB with Azure Static Web Apps' `/data-api/` the `nextLink` is not usable as it does not get rewritten to reflect the frontend url:

![image](https://user-images.githubusercontent.com/11445087/230365252-1e624439-9818-4fcb-bf6c-e1810570e8ee.png)

![image](https://user-images.githubusercontent.com/11445087/230365295-8ea51bb6-3549-4d5e-adf6-7a6a7ebc0627.png)

### Version

Microsoft.DataApiBuilder 0.6.13+32fc03f4fe439d02aef454f10cc58e50d0d29f79

### What database are you using?

Azure SQL

### What hosting model are you using?

Static Web Apps (SWA)

### Which API approach are you accessing DAB through?

REST

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

seantleonard:
Thank you for raising this issue, @sebader. We are looking into this now.

PR #1506, opened 2023-05-18
Add support for adding base-route for Runtime

ayush3797:
## Why make this change?
SWA or any other front end can interact with DAB using a path-based routing mechanism, where they may use a path (`base-route`) to redirect requests to DAB (refer to the bug encountered: https://github.com/Azure/data-api-builder/issues/1420). This `base-route` gets stripped off when the request lands at DAB. Consequently, DAB is not even aware of anything like a `base-route` being used by the front end.
This poses a problem in generation of the `nextLink`, where the `nextLink` which DAB generates, does not contain the `base-route` and consequently when the front end receives the next request, it does not find the base route in the request URL and hence, has no idea where to redirect the request to.

Using the newly added option to configure base route for rest requests via the config (CLI option - `--runtime.base-route`), the front-end like SWA can make DAB aware that they are using a base route to redirect requests to DAB. DAB can then accordingly account for the base route while generating the `nextLink` for paginated responses.
## What is this change?

1. Added feature to specify `base-route` in global runtime settings in the config which can be via CLI as well using `--runtime.base-route `option in the `dab init` command. By default, it will be initialized as null. When the base-route is a non-empty string, it will be included in the generated nextLink. This feature is only supported when `authentication provider == StaticWebApps`. We throw an exception when base-route is configured for non-SWA providers.
2. Extended functionality of `Utils.IsApiPathValid(string? apiPath, ApiType apiType)` method to  `Utils.IsUriComponentValid(string? uriComponent)` because similar validations are required for api properties: `rest.path`/ `runtime.base-route`/ `graphql.path`, all of which are part of the request URI.
3. Renamed `RuntimeConfigValidator.ValidateRestPathForRelationalDbs()` to `RuntimeConfigValidator.ValidateRestURI()`.
4. Renamed `RuntimeConfigValidator.ValidateGraphQLPath()` to `RuntimeConfigValidator.ValidateGraphQLURI()` so that we follow same convention between rest and gql.

## Q&A

1. Will this require updating an already created configuration file? 
No, the existing config file will work perfectly fine. The base-route is totally an optional feature.
2. Will a default value be implicitly used if the new base path config property is not set? 
The default value for the base-route is null. This essentially means that base-route will not come into the picture.
3. What is the default value and does the value differ between dev/prod swa/non-swa?
The default value will be same irrespective of dev/prod/swa/non-swa environment and equal to null.

So to say, the **change is not a breaking change and we are totally backwards compatible**.

## How was this tested?
- [x] Integration Tests - Was added to `ConfigurationTests.cs` to assert that the generated nextLink contains the base-route.
- [x] Unit tests - were added to `ConfigValidationUnitTests.cs` to confirm that we throw proper exceptions when runtime base-route is formatted incorrectly (contains reserved characters or doesn't start with a '/') or when base-route is configured for non-SWA authentication providers.


seantleonard:
Can you add comment about backwards compatibility in pr description:
- Will this require updating an already created configuration file? 
- Will a default value be implicitly used if the new base path config property is not set? 
- What is the default value and does the value differ between dev/prod swa/non-swa?

ayush3797:
> Can you add comment about backwards compatibility in pr description:
> 
> * Will this require updating an already created configuration file?
> * Will a default value be implicitly used if the new base path config property is not set?
> * What is the default value and does the value differ between dev/prod swa/non-swa?

Done

seantleonard:
> "The default value for the base-route = "" (empty string). This essentially means that base-route will not come into the picture."

I interpret the above statement to mean that user intervention (config modification) is required to mitigate the bug mentioned in #1420. While this is not 'breaking' in the scope of engine startup failures, it would be 'breaking' in the sense that the config must be changed for the `nextLink` to work as expected in SWA.

Let me know if/what I'm missing.

ayush3797:
> > "The default value for the base-route = "" (empty string). This essentially means that base-route will not come into the picture."
> 
> I interpret the above statement to mean that user intervention (config modification) is required to mitigate the bug mentioned in #1420. While this is not 'breaking' in the scope of engine startup failures, it would be 'breaking' in the sense that the config must be changed for the `nextLink` to work as expected in SWA.
> 
> Let me know if/what I'm missing.

Yes, that's the expectation and I think the SWA side is aware of this, right @aniruddh25?

seantleonard:
> > > "The default value for the base-route = "" (empty string). This essentially means that base-route will not come into the picture."
> > 
> > 
> > I interpret the above statement to mean that user intervention (config modification) is required to mitigate the bug mentioned in #1420. While this is not 'breaking' in the scope of engine startup failures, it would be 'breaking' in the sense that the config must be changed for the `nextLink` to work as expected in SWA.
> > Let me know if/what I'm missing.
> 
> Yes, that's the expectation and I think the SWA side is aware of this, right @Aniruddh25?

They are aware of the behavior, though may not be aware of how the fix works or whether it is breaking.

ayush3797:
> > > > "The default value for the base-route = "" (empty string). This essentially means that base-route will not come into the picture."
> 
> > > 
> 
> > > 
> 
> > > I interpret the above statement to mean that user intervention (config modification) is required to mitigate the bug mentioned in #1420. While this is not 'breaking' in the scope of engine startup failures, it would be 'breaking' in the sense that the config must be changed for the `nextLink` to work as expected in SWA.
> 
> > > Let me know if/what I'm missing.
> 
> > 
> 
> > Yes, that's the expectation and I think the SWA side is aware of this, right @Aniruddh25?
> 
> 
> 
> They are aware of the behavior, though may not be aware of how the fix works or whether it is breaking.

I'm not very clear about what you mean here. The base route is something which they use on top of DAB. So it makes sense for them to let DAB know what's that base route? And that will be done via the config file.

src/Cli/ConfigGenerator.cs
```diff
@@ -96,7 +97,16 @@ public static bool TryCreateRuntimeConfig(InitOptions options, out string runtim
                             "it does not support REST yet.");
                     }
 
+                    // If the option --rest.base-route is specified for cosmosdb_nosql, log a warning because
+                    // rest is not supported for cosmosdb_nosql yet.
+                    if (!string.Empty.Equals(restBaseRoute))
+                    {
+                        _logger.LogWarning("Configuration option --rest.base-route is not honored for cosmosdb_nosql since " +
+                            "it does not support REST yet.");
```
aaronpowell, line 105:
```suggestion
                        _logger.LogWarning("Configuration option --rest.base-route is not honored for cosmosdb_nosql since it does not support REST.");
```

No need to split this over multiple lines.

Also dropped `yet` as that implies it's something we will end up supporting but there's no guarantee that we are going to support it.

src/Cli/ConfigGenerator.cs
```diff
@@ -127,7 +137,10 @@ public static bool TryCreateRuntimeConfig(InitOptions options, out string runtim
                 return false;
             }
 
-            if (!IsApiPathValid(restPath, ApiType.REST) || !IsApiPathValid(options.GraphQLPath, ApiType.GraphQL))
+            if (!IsURIComponentValid(restPath, ApiType.REST, ApiSettings.PROPERTY_NAME_PATH) ||
+                !IsURIComponentValid(options.GraphQLPath, ApiType.GraphQL, ApiSettings.PROPERTY_NAME_PATH) ||
+                !IsURIComponentValid(restBaseRoute, ApiType.REST, ApiSettings.PROPERTY_NAME_BASE_ROUTE))
+
```
aaronpowell, line 143: I wonder if we'd be better splitting this into three separate `if` blocks so that we can provide better diagnostic information as to why something failed to pass.

src/Cli/Utils.cs
```diff
@@ -324,9 +330,9 @@ public static bool TryParseMappingDictionary(IEnumerable<string> mappingList, ou
 
             // If restPath is null, it implies we are dealing with cosmosdb_nosql,
             // which only supports graphql.
-            if (restPath is not null)
+            if (restPath is not null && restBaseRoute is not null)
             {
-                defaultGlobalSettings.Add(GlobalSettingsType.Rest, new RestGlobalSettings(Enabled: restEnabled, Path: restPath));
+                defaultGlobalSettings.Add(GlobalSettingsType.Rest, new RestGlobalSettings(Enabled: restEnabled, Path: restPath, BaseRoute: restBaseRoute));
```
aaronpowell, line 335: We'll have a merge conflict with #1402 as the structure of the config we store this in has changed. It'll be a new property on `RestRuntimeOptions` and we'll have to set it appropriately there.

src/Config/GlobalSettings.cs
aaronpowell, line 1: With #1402 this will be invalid, instead the property will go to `RestRuntimeOptions`.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1086,6 +1086,46 @@ public void TestConnectionStringEnvVarHasHighestPrecedence()
...
+            using (TestServer server = new(Program.CreateWebHostBuilder(args)))
+            using (HttpClient client = server.CreateClient())
+            {
+                string requestPath = "/api/MappedBookmarks";
+                HttpMethod httpMethod = SqlTestHelper.ConvertRestMethodToHttpMethod(RestMethod.Get);
+                HttpRequestMessage request = new(httpMethod, requestPath);
+
+                HttpResponseMessage response = await client.SendAsync(request);
+                string responseBody = await response.Content.ReadAsStringAsync();
+                Assert.IsTrue(response.StatusCode is HttpStatusCode.OK);
```
aaronpowell, line 1114:
```suggestion
                Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
```

Let's use `AreEqual` rather than `IsTrue` with an assert so it's more obvious when it fails.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1086,6 +1086,46 @@ public void TestConnectionStringEnvVarHasHighestPrecedence()
...
+
+                JsonElement responseElement = JsonSerializer.Deserialize<JsonElement>(responseBody);
+                JsonElement responseValue = responseElement.GetProperty(SqlTestHelper.jsonResultTopLevelKey);
+                string nextLink = responseElement.GetProperty("nextLink").ToString();
+
+                // Assert that we got an array response with length equal to the maximum allowed records in a paginated response.
+                Assert.IsTrue(responseValue.ValueKind is JsonValueKind.Array);
+                Assert.IsTrue(responseValue.GetArrayLength() == 100);
```
aaronpowell, line 1122:
```suggestion
                Assert.AreEqual(100, responseValue.GetArrayLength());
```

Let's use `AreEqual` rather than `IsTrue` with an assert so it's more obvious when it fails.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1086,6 +1086,46 @@ public void TestConnectionStringEnvVarHasHighestPrecedence()
...
+                HttpResponseMessage response = await client.SendAsync(request);
+                string responseBody = await response.Content.ReadAsStringAsync();
+                Assert.IsTrue(response.StatusCode is HttpStatusCode.OK);
+
+                JsonElement responseElement = JsonSerializer.Deserialize<JsonElement>(responseBody);
+                JsonElement responseValue = responseElement.GetProperty(SqlTestHelper.jsonResultTopLevelKey);
+                string nextLink = responseElement.GetProperty("nextLink").ToString();
+
+                // Assert that we got an array response with length equal to the maximum allowed records in a paginated response.
+                Assert.IsTrue(responseValue.ValueKind is JsonValueKind.Array);
```
aaronpowell, line 1121:
```suggestion
                Assert.AreEqual(JsonValueKind.Array, responseValue.ValueKind);
```

Let's use `AreEqual` rather than `IsTrue` with an assert so it's more obvious when it fails.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1086,6 +1086,46 @@ public void TestConnectionStringEnvVarHasHighestPrecedence()
...
+                Assert.IsTrue(responseValue.GetArrayLength() == 100);
+
+                // Assert that the nextLink contains the rest base-route just before the request path.
+                Assert.IsTrue(nextLink.Contains(restBaseRoute + requestPath));
```
aaronpowell, line 1125:
```suggestion
                StringAssert.Contains(nextLink, restBaseRoute + requestPath);
```

Let's use `StringAssert` here which is more optimised for asserting on strings and will give us better diagnostics on test failure.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1086,6 +1086,46 @@ public void TestConnectionStringEnvVarHasHighestPrecedence()
...
+            {
+                string requestPath = "/api/MappedBookmarks";
+                HttpMethod httpMethod = SqlTestHelper.ConvertRestMethodToHttpMethod(RestMethod.Get);
+                HttpRequestMessage request = new(httpMethod, requestPath);
+
+                HttpResponseMessage response = await client.SendAsync(request);
+                string responseBody = await response.Content.ReadAsStringAsync();
+                Assert.IsTrue(response.StatusCode is HttpStatusCode.OK);
+
+                JsonElement responseElement = JsonSerializer.Deserialize<JsonElement>(responseBody);
```
aaronpowell, line 1116: Should we create a C# type for the response? We know what it'd look like so would that make sense, rather than us having to walk JSON structures in the test?

src/Service.Tests/TestHelper.cs
```diff
@@ -240,26 +240,39 @@ public static void AddMissingEntitiesToConfig(RuntimeConfig config, string entit
...
         /// <summary>
         /// Utility method that reads the config file for a given database type and constructs a
-        /// new config file with changes just in the host mode section.
+        /// new config file with custom changes as specified in the method parameters.
         /// </summary>
         /// <param name="configFileName">Name of the new config file to be constructed</param>
         /// <param name="hostModeType">HostMode for the engine</param>
         /// <param name="databaseType">Database type</param>
-        public static void ConstructNewConfigWithSpecifiedHostMode(string configFileName, HostModeType hostModeType, string databaseType)
+        public static void ConstructNewConfigWithCustomSettings(
```
aaronpowell, line 248: Just noting that #1402 will require a revisit of this file due to the structure changes of the config types.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1438,7 +1436,7 @@ string[] linkingTargetFields
             Dictionary<GlobalSettingsType, object> settings = new()
             {
                 { GlobalSettingsType.GraphQL, JsonSerializer.SerializeToElement(new GraphQLGlobalSettings(Path: graphQLPathPrefix)) },
-                { GlobalSettingsType.Rest, JsonSerializer.SerializeToElement(new RestGlobalSettings(){ Path = restPathPrefix }) }
+                { GlobalSettingsType.Rest, JsonSerializer.SerializeToElement(new RestGlobalSettings(){ Path = restPathPrefix, BaseRoute = "/rest-route" }) }
```
aaronpowell, line 1439: Just noting that this will be impacted by #1402

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -10,7 +10,8 @@
   "runtime": {
     "rest": {
       "enabled": true,
-      "path": "/api"
+      "path": "/api",
+      "base-route": ""
```
aaronpowell, line 14: Is this change (and on the MySQL and PostgreSQL files) the result of running the `config-generator` scripts? We need to ensure that that is the way these files are generated so that we avoid situations where they get out of sync and can cause problems on the build server (which runs them).

src/Service/Configurations/RuntimeConfigProvider.cs
```diff
@@ -58,6 +58,11 @@ public virtual string RestPath
             get { return RuntimeConfiguration is not null ? RuntimeConfiguration.RestGlobalSettings.Path : string.Empty; }
         }
 
+        public virtual string RestBaseRoute
+        {
+            get { return RuntimeConfiguration is not null ? RuntimeConfiguration.RestGlobalSettings.BaseRoute : string.Empty; }
+        }
+
```
aaronpowell, line 65: Just noting that #1402 will impact this, primarily because these properties aren't shadowed in the types like this, you always go back to the `RuntimeConfig` type to read it.

src/Service/Configurations/RuntimeConfigProvider.cs
```diff
@@ -58,6 +58,11 @@ public virtual string RestPath
             get { return RuntimeConfiguration is not null ? RuntimeConfiguration.RestGlobalSettings.Path : string.Empty; }
         }
 
+        public virtual string RestBaseRoute
```
aaronpowell, line 61: Is there a reason we wouldn't have this as `string?` - should `null` be allowed, as that's valid.

src/Service/Resolvers/SqlQueryEngine.cs
```diff
@@ -213,6 +214,39 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
             // nextLink is the URL needed to get the next page of records using the same query options
             // with $after base64 encoded for opaqueness
             string path = UriHelper.GetEncodedUrl(_httpContextAccessor.HttpContext!.Request).Split('?')[0];
+
+            // If the base route is not empty, we need to insert it into the URI before the rest path.
+            string baseRoute = _runtimeConfigProvider.RestBaseRoute;
```
aaronpowell, line 219: With #1402 we'll go `runtimeConfig.Runtime.Rest.BaseRoute` rather than a shadowed property off the provider.

src/Service/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -108,6 +108,15 @@ public interface ISqlMetadataProvider
...
         bool TryGetEntityNameFromPath(string entityPathName, [NotNullWhen(true)] out string? entityName);
 
+        /// <summary>
+        /// Try to obtain the path of the Entity that has the provided entityName. If It
+        /// exists save in out param, and return true, otherwise return false.
+        /// </summary>
+        /// <param name="entityName">Name of the associated entity.</param>
+        /// <param name="entityPathName">Entity's path as seen in a request.</param>
+        /// <returns>True if exists, otherwise false.</returns>
+        bool TryGetEntityPathFromName(string entityName, [NotNullWhen(true)] out string? entityPathName);
```
aaronpowell, line 118: You could avoid having to implement this on the Cosmos provider by using a [default interface implementation](https://devblogs.microsoft.com/dotnet/default-implementations-in-interfaces/).

```suggestion
        bool TryGetEntityPathFromName(string entityName, [NotNullWhen(true)] out string? entityPathName) => throw new NotImplementedException();
```

aaronpowell (changes requested):
I've gone through the PR in light of the changes coming from #1402 and have called out some places that will be impacted (but not every place).

There are also a few general changes suggested, particularly around the tests.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -119,7 +119,7 @@ public void TestInitForCosmosDBPostgreSql()
     public void TestInitializingRestAndGraphQLGlobalSettings()
     {
         string[] args = { "init", "-c", TEST_RUNTIME_CONFIG_FILE, "--database-type", "mssql", "--rest.path", "/rest-api",
-                          "--rest.disabled", "--graphql.path", "/graphql-api" };
+                          "--rest.base-route", "/rest-route", "--rest.disabled", "--graphql.path", "/graphql-api" };
```
Aniruddh25, line 122: There might have been some misunderstanding, here's some clarification:

- the `base-route` feature is applicable to the entire runtime - both graphql and rest. So, it needs to be under the runtime section, since when accessed from SWA its `/data-api/rest/*` and `/data-api/graphql`
- Today, SWA does not allow to customize this path, and plans to send us the base route through the configuration endpoint V2 [parameters](https://github.com/Azure/data-api-builder/blob/2e8d1bb8b7d83bbe2a90472c2ccdc8acb87fa5ec/src/Service/Controllers/ConfigurationController.cs#L130). As such it should be honored by the engine.
- SWA will send it as part of system configuration so it stays customer hidden and SWA-CLI will not provide a way today to customize it, may be it will in future, so we can let ability to customize it through DAB CLI

src/Service/Resolvers/SqlQueryEngine.cs
```diff
@@ -213,6 +215,40 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
             // nextLink is the URL needed to get the next page of records using the same query options
             // with $after base64 encoded for opaqueness
             string path = UriHelper.GetEncodedUrl(_httpContextAccessor.HttpContext!.Request).Split('?')[0];
+
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            // If the base route is not empty, we need to insert it into the URI before the rest path.
+            string? baseRoute = runtimeConfig.Runtime.Rest.BaseRoute;
+            if (!string.IsNullOrEmpty(baseRoute))
+            {
+                _sqlMetadataProvider.TryGetEntityPathFromName(context.EntityName, out string? pathNameForEntity);
```
Aniruddh25, line 224: Can we move this up in the `if` condition so that we proceed only when we get a `pathNameForEntity`

src/Service/Resolvers/SqlQueryEngine.cs
```diff
@@ -213,6 +215,40 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+
+                // Path is of the form ....restPath/pathNameForEntity. We want to insert the base route before the restPath,
+                // hence we remove pathNameForEntity and restPath from the path and append them back after appending the base route to the path.
+                // Initialize lastIdxOfPathToBeIncluded by removing the pathNameForEntity from the path.
+                int lastIdxOfPathToBeIncluded = path.Length - pathNameForEntity!.Length - 1;
```
Aniruddh25, line 229: Performing this for each request seems computationally expensive. Can we optimize in any way?

src/Service/Resolvers/SqlQueryEngine.cs
```diff
@@ -213,6 +215,40 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+                lastIdxOfPathToBeIncluded = lastIdxOfPathToBeIncluded - runtimeConfig.Runtime.Rest.Path.Length;
+
+                // Initialize pathWithBaseRoute by removing restPath and pathNameForEntity from the path.
+                // Finally, it will be of the form: .../baseRoute/restPath/pathNameForEntity.
+                StringBuilder pathWithBaseRoute = new(path.Substring(0, lastIdxOfPathToBeIncluded + 1));
+
+                // Append base route.
+                pathWithBaseRoute.Append(baseRoute);
+                // Append rest path.
+                pathWithBaseRoute.Append($"{runtimeConfig.Runtime.Rest.Path}");
```
Aniruddh25, line 246: is runtimeConfig.Runtime.Rest.Path expected to contain a forward slash?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1426,73 +1426,72 @@ string[] linkingTargetFields
         /// <summary>
         /// Tests whether the API path prefix is well formed or not.
         /// </summary>
-        /// <param name="apiPathPrefix">API path prefix</param>
+        /// <param name="apiPath">API path prefix</param>
         /// <param name="expectedErrorMessage">Expected error message in case an exception is thrown.</param>
         /// <param name="IsPathContainingReservedCharacters">Boolean indicating if path prefix contains reserved characters.</param>
         /// <param name="expectError">Exception expected</param>
-        // @"[\.:\?#/\[\]@!$&'()\*\+,;=]+";
```
Aniruddh25, line 1433: Are these the reserved characters?

src/Service/Resolvers/SqlQueryEngine.cs
ayush3797: Yes, that's correct. If it doesn't, we will fail during the config validation stage.

src/Config/ObjectModel/ApiType.cs
```diff
@@ -9,5 +9,7 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 public enum ApiType
 {
     REST,
-    GraphQL
+    GraphQL,
+    // This is required to indicate features common between REST/GraphQL APIs.
+    Both
```
Aniruddh25, line 14: Looks like apiType `both` is really only used to print out error messages.
in RuntimeConfigValidator:
e.g.:
"{apiType} {apiProperty} cannot be null or empty."
The apiType name `Both` seems inappropriate for the error message.

In future, if there are other APIs, `Both` will be restrictive. Perhaps, `All` makes more sense?

Aniruddh25:
nit: Update the description to reflect its `runtime` section's property

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Yes

Aniruddh25 (changes requested):
Waiting on resolving conversations before approval.

src/Service/Resolvers/SqlQueryEngine.cs
ayush3797: I'm afraid not, will give it some more thought.

src/Service/Configurations/RuntimeConfigProvider.cs
ayush3797: Outdated code.

src/Service.Tests/dab-config.MsSql.json
ayush3797: Outdated code again. But to answer your question, yes I ran the scripts locally and this was generated via that. The config file generated had a large diff because of ctrl characters i guess but this was the only change that had to be incorporated in.

src/Cli/ConfigGenerator.cs
ayush3797: This will already be logged here: https://github.com/Azure/data-api-builder/blob/b818d6fa239ebb80ed71b9bfa335924c4da8d819/src/Cli/Utils.cs#L213.

src/Service.Tests/Configuration/ConfigurationTests.cs
ayush3797: I don't think we should create it as yet. If the use cases increase we can create a c# type. Do you see other use cases currently as well? I might be missing..

src/Service/Resolvers/SqlQueryEngine.cs
ayush3797: I am not sure what you mean here. The pathNameForEntity will either be the entity's name or entity's rest path. In any case, if the base route is non-empty (and obviously non-null), we would perform the same logic.

src/Cli/ConfigGenerator.cs
ayush3797: removed code.

src/Service/Resolvers/SqlQueryEngine.cs
ayush3797: Due to the dynamic nature of uri, I don't think we can kind of precalculate it instead of doing it on a per request basis.

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronpowell: It's more looking at the complexity of the following `Assert` statements and that we need to have the test aware of the structure of the JSON payload its working with. While it's not inherently problematic, I always ask the question when we have to walk JSON structures in tests.

src/Config/ObjectModel/RuntimeOptions.cs
```diff
@@ -3,4 +3,9 @@
 
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
-public record RuntimeOptions(RestRuntimeOptions Rest, GraphQLRuntimeOptions GraphQL, HostOptions Host);
+public record RuntimeOptions(RestRuntimeOptions Rest, GraphQLRuntimeOptions GraphQL, HostOptions Host, string? BaseRoute = null)
+{
+    public const string JSON_PROPERTY_NAME = "runtime";
+    public const string PROPERTY_NAME_PATH = "path";
+    public const string PROPERTY_NAME_BASE_ROUTE = "base-route";
```
aaronpowell, line 10: I have reservations about this as it means we're starting to bleed out the structure of the JSON config to other parts of the codebase.

We shouldn't need to say "this is what the JSON looks like", that should only be the concern of the serializer/deserializer.

I see where this is being used, downstream in the validator to push into the potential exception messages. This is a risk factor created by really generic methods (such as those in the validator), and I wonder if we can't achieve the same with a `nameof` call against the relevant property, especailly as `runtime` and `path` aren't properties of `RuntimeOptions` so we're kind of shoving stuff in places that they don't directly link to.

ayush3797: I see what you're saying. @aniruddh25 also had a similar comment on the other PR: https://github.com/Azure/data-api-builder/pull/1496#discussion_r1253854635. You can have a look there.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -208,6 +210,40 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+                StringBuilder pathWithBaseRoute = new(path.Substring(0, lastIdxOfPathToBeIncluded + 1));
+
+                // Append base route.
+                pathWithBaseRoute.Append(baseRoute);
+                // Append rest path.
+                pathWithBaseRoute.Append($"{runtimeConfig.Runtime.Rest.Path}");
+                // Append path name for entity.
+                pathWithBaseRoute.Append($"/{pathNameForEntity}");
+                path = pathWithBaseRoute.ToString();
+            }
```
seantleonard, line 245: If you are trying to just prefix `/<rest-base-path>/<entityName>` with `/<base-path>`, you can do this an easier/less verbose way by:
1. Accessing `_httpContextAccessor.HttpContext!.Request.Path` --> `/api/myEntity` to build your `pathWithBaseRoute`
2. Reference implementation of UriHelper.GetEncodedUrl() used on line 212. Because you don't shouldn't modify the httpcontext directly with your new path value. (that would need more vetting as an option).

Below you would just substitute out request.Path with your generated `pathWithBaseRoute` and strip off the query string.

```csharp
HttpRequest request = _httpContextAccessor.HttpContext!.Request;
string newpath = BuildAbsolute(request.Scheme, request.Host, request.PathBase, pathWithBaseRoute, request.QueryString).Split('?')[0];
```
image

src/Service/Resolvers/SqlQueryEngine.cs
Aniruddh25:
```
if (!string.IsNullOrEmpty(baseRoute) && _sqlMetadataProvider.TryGetEntityPathFromName(context.EntityName, out string? pathNameForEntity))
{
 ...
} 
```
I mean the above code, this way you dont need to use null forgiving operator further below the code for `pathNameForEntity`

seantleonard:
Static Web Apps does not let the user customize the 'base-path' and [statically sets it to](https://learn.microsoft.com/en-us/azure/static-web-apps/database-overview#endpoint-location) `/data-api`. So requests to dab entities will substitue look like:

```https
<host:port>/data-api/{dab-global-rest/graphql-path}/{entity}
```

Since nextLink is broken in StaticWebApps, what would be the effect of automatically setting the base-path property you introduce to `/data-api` when Auth provider is `StaticWebApps`? (because that provider is the only one supported/and synonymous with StaticWebApp deployment)

src/Core/Resolvers/SqlQueryEngine.cs
seantleonard: Or you may even be able to achieve all of this on line 212, and just provide a value for request.pathBase argument with the config set value.

src/Cli/Utils.cs
```diff
@@ -186,26 +186,26 @@ public static bool TryParseMappingDictionary(IEnumerable<string> mappingList, ou
         /// <summary>
         /// Returns true if the api path contains any reserved characters like "[\.:\?#/\[\]@!$&'()\*\+,;=]+"
         /// </summary>
-        /// <param name="apiPath">path prefix for rest/graphql apis</param>
+        /// <param name="uriComponent">Path prefix/base route for REST/GraphQL APIs.</param>
         /// <param name="apiType">Either REST or GraphQL</param>
-        public static bool IsApiPathValid(string? apiPath, ApiType apiType)
+        public static bool IsURIComponentValid(string? uriComponent, ApiType apiType, string apiProperty)
```
seantleonard, line 191:
```suggestion
        public static bool IsUriComponentValid(string? uriComponent, ApiType apiType, string apiProperty)
```
nit: not necessary to have uppercase URI.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -208,6 +210,40 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
             // nextLink is the URL needed to get the next page of records using the same query options
             // with $after base64 encoded for opaqueness
             string path = UriHelper.GetEncodedUrl(_httpContextAccessor.HttpContext!.Request).Split('?')[0];
+
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            // If the base route is not empty, we need to insert it into the URI before the rest path.
```
Aniruddh25, line 215: Instead of inserting the base-route with substring etc, can we construct the full path ourselves since we know all the components anyway? 

E.g. Can we use this: https://devio.wordpress.com/2009/10/19/get-absolut-url-of-asp-net-application/ ? 

What will be more efficient and accurate to do?

Can we move this logic into a small helper function?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -273,8 +271,11 @@ public static void ValidateGlobalEndpointRouteConfig(RuntimeConfig runtimeConfig
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
             }
 
-            ValidateRestPathForRelationalDbs(runtimeConfig);
-            ValidateGraphQLPath(runtimeConfig);
+            string? runtimeBaseRoute = runtimeConfig.Runtime.BaseRoute;
+            ValidateURIComponent(runtimeBaseRoute, RuntimeOptions.PROPERTY_NAME_BASE_ROUTE, ApiType.All);
+
+            ValidateRestURI(runtimeConfig);
+            ValidateGraphQLURI(runtimeConfig);
```
seantleonard, line 278: To avoid having to created extra properties on `RuntimeOptions`, and passing around property names and Api types, maybe this spot could be refactored to a more generic `IsValidUri(uri)`, if false, raise an exception and statically create the message with "REST path invalid" etc. 

These are essentially the same. Could you add any cosmos checks prior to validating the rest uri?
- ValidateRestURI(runtimeConfig);
- ValidateGraphQLURI(runtimeConfig);

Any exceptions thrown in validation helpers like ValidateRestUri can instead be logged with ILogger, the whole point being that the error is surfaced to the user on the console.

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -107,6 +107,16 @@ public interface ISqlMetadataProvider
...
+        /// <summary>
+        /// Try to obtain the path of the Entity that has the provided entityName. If It
+        /// exists save in out param, and return true, otherwise return false.
+        /// </summary>
+        /// <param name="entityName">Name of the associated entity.</param>
+        /// <param name="entityPathName">Entity's path as seen in a request.</param>
+        /// <returns>True if exists, otherwise false.</returns>
+        bool TryGetEntityPathFromName(string entityName, [NotNullWhen(true)] out string? entityPathName)
+            => throw new NotImplementedException();
+
```
seantleonard, line 119: may not need this pending simplification uri creation with base route in sqlqueryengine

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -189,6 +191,12 @@ public virtual bool TryGetEntityNameFromPath(string entityPathName, [NotNullWhen
             return EntityPathToEntityName.TryGetValue(entityPathName, out entityName);
         }
 
+        /// <inheritdoc/>
+        public virtual bool TryGetEntityPathFromName(string entityName, [NotNullWhen(true)] out string? entityPathName)
+        {
+            return EntityNameToEntityPath.TryGetValue(entityName, out entityPathName);
+        }
+
```
seantleonard, line 199: see other comments about whether this will be necessary.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1410,73 +1410,72 @@ string[] linkingTargetFields
...
         [DataRow("/ api_path", null, false, ApiType.GraphQL, false,
-            DisplayName = "API path prefix containing space at the start and underscore in between.")]
-        [DataRow("/", null, false, ApiType.GraphQL, false,
-            DisplayName = "API path containing only a forward slash.")]
-        public void ValidateApiPathIsWellFormed(
-            string apiPathPrefix,
+            DisplayName = "GraphQL path prefix containing space at the start and underscore in between.")]
+        [DataRow("/", null, false, ApiType.REST, false,
+            DisplayName = "REST path containing only a forward slash.")]
+        public void ValidateApiURIsAreWellFormed(
```
seantleonard, line 1470: Depending on whether any refactors occur based on other comments, a test on `ValidateGlobalEndpointRouteConfig()` instead of the individual `Validate<Rest/GraphQL/Any>Uri()` methods, may be more effective. Then you can utilize ILogger to check error message output.

src/Config/ObjectModel/RuntimeOptions.cs
seantleonard: I posted a comment elsewhere with suggestions on how to avoid adding these string properties.

seantleonard (changes requested):
Some suggestions on simplification

src/Core/Resolvers/SqlQueryEngine.cs
ayush3797: Woww, that simplifies things a lot. Thanks for suggesting Sean!

src/Core/Resolvers/SqlQueryEngine.cs
ayush3797: With Sean's suggestion, this is taken care of.

src/Core/Resolvers/SqlQueryEngine.cs
ayush3797: Incorporated this!

src/Service/Resolvers/SqlQueryEngine.cs
ayush3797: Incorporated https://github.com/Azure/data-api-builder/pull/1506#discussion_r1254974637

src/Service/Resolvers/SqlQueryEngine.cs
ayush3797: Oh okay, got it. Don't need this code now though.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -61,6 +61,8 @@ public abstract class SqlMetadataProvider<ConnectionT, DataAdapterT, CommandT> :
 
         private Dictionary<string, string> EntityPathToEntityName { get; } = new();
 
+        private Dictionary<string, string> EntityNameToEntityPath { get; } = new();
```
Aniruddh25, line 64: If this is no longer populated, do we still need the member?

Aniruddh25 (approved):
LGTM after resolving any pending suggestions, thanks for the patience on this one!

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Oh no, we don't. Must've missed removing this. Thanks!

src/Core/Configurations/RuntimeConfigValidator.cs
ayush3797: Incorporated.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Since the tests are specifically for testing if the gql and rest uris are well formed, I would say, lets keep it this way.

src/Config/ObjectModel/RuntimeOptions.cs
ayush3797: I have made the possible reduction in passing of these constants as arguments but if we totally want to remove these constants, we need to use string literals.

src/Service.Tests/Configuration/ConfigurationTests.cs
ayush3797: Will take this into consideration in a follow up issue.

abhishekkumams (approved):
LGTM, thanks for adding this support.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -307,75 +320,76 @@ public static void ValidateRestPathForRelationalDbs(RuntimeConfig runtimeConfig)
...
-                {
-                    errorMessage = INVALID_REST_PATH_WITH_RESERVED_CHAR_ERR_MSG;
-                }
-
-                throw new DataApiBuilderException(
-                    message: errorMessage,
-                    statusCode: HttpStatusCode.ServiceUnavailable,
-                    subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
-            }
+            return _invalidApiPathCharsRgx.IsMatch(uriComponent);
```
seantleonard, line 392: now that Uri validation is generic, should the name `_invalidApiPathCharsRgx` be more generic?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1199,6 +1199,47 @@ public void TestConnectionStringEnvVarHasHighestPrecedence()
             }
         }
 
+        /// <summary>
+        /// Test to validate that when an entity which will return a paginated response is queried, and a custom runtime base route is configured in the runtime configuration,
+        /// then the generated nextLink in the response would contain the rest base-route just before the rest path. For the subsequent query, the rest base-route will be trimmed
+        /// by the upstream before the request lands at DAB.
```
seantleonard, line 1205: > For the subsequent query, the rest base-route will be trimmed by the upstream before the request lands at DAB.

I'm not sure the relevance or what you mean by this, did you intend to have a second request and assert that the nextLink is not present?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1410,73 +1410,72 @@ string[] linkingTargetFields
...
+            string apiPath,
             string expectedErrorMessage,
             bool pathContainsReservedCharacters,
             ApiType apiType,
             bool expectError)
         {
-            ValidateRestAndGraphQLPathIsWellFormed(
-                apiPathPrefix,
+            ValidateRestAndGraphQLURIIsWellFormed(
```
seantleonard, line 1477: Are there any other callers of your validation logic (`ValidateRestAndGraphQLURIIsWellFormed()`)? I don't follow why a separate function is needed here

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1525,31 +1524,25 @@ string[] linkingTargetFields
...
-                    RuntimeConfigValidator.ValidateRestPathForRelationalDbs(configuration));
-
                     if (pathContainsReservedCharacters)
                     {
-                        expectedErrorMessage = RuntimeConfigValidator.INVALID_REST_PATH_WITH_RESERVED_CHAR_ERR_MSG;
-                    }
-                    else
-                    {
-                        expectedErrorMessage = "REST " + expectedErrorMessage;
+                        expectedErrorMessage =
```
seantleonard, line 1529: Why the recalculation of the `expectedErrorMessage` if a value is provided in the DataRow? Would an assert on whether the provided expected error message work instead?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1493,7 +1492,7 @@ string[] linkingTargetFields
         /// <param name="pathContainsReservedCharacters">Boolean indicating if path prefix contains reserved characters.</param>
         /// <param name="expectError">Exception expected</param>
         /// <param name="apiType">Either REST or GraphQL</param>
-        private static void ValidateRestAndGraphQLPathIsWellFormed(
+        private static void ValidateRestAndGraphQLURIIsWellFormed(
```
seantleonard, line 1495: The many conditionals in this method seem to complicate what you are trying to test. Is there any cleaner way to accomplish the asserts on the error messages you are expecting? It looks like `ApiType` isn't necessary.

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json",
   "data-source": {
     "database-type": "mssql",
-    "connection-string": "Server=tcp:127.0.0.1,1433;Persist Security Info=False;User ID=sa;Password=REPLACEME;MultipleActiveResultSets=False;Connection Timeout=5;",
+    "connection-string": "Server=tcp:127.0.0.1,1433;Persist Security Info=False;Trusted_Connection=True;TrustServerCertificate=True;MultipleActiveResultSets=False;Connection Timeout=5;",
```
seantleonard, line 5: I don't think this change was intended? `TrustServerCertificate=True` may be fine when testing locally, but ideally shouldn't be a setting we encourage in our sample files. This setting will also be overridden in mssqlqyueryexecutor

seantleonard (changes requested):
Some questions/updates requested based on latest changes.

src/Config/ObjectModel/RuntimeOptions.cs
seantleonard: I'm still on the fence about whether explicit constants are necessary. Was the motivation to have fewer places to change a value in case we ever change the JSON property name? As any changes to property names would be breaking changes, that scenario seems unlikely.

aaronpowell: I'm with @seantleonard here, I feel like it's tackling a hypothetical scenario that, if it came to be, would have more cascading problems than mistakenly changing a string value.

src/Service.Tests/dab-config.MsSql.json
ayush3797: yeah thanks

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: This wasn't a change from this PR. But yeah, there are no other callers and this function can be removed. Thanks!

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Again, not a change from this PR. But a value was not provided for every datarow in the test- specifically for the ones that contain reserved characters. And that's why the calculation. I will update so that the test does not have to deal with this. Thanks!

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Simplified!

src/Service.Tests/Configuration/ConfigurationTests.cs
ayush3797: No, this is just an FYI that even though the nextLink generated contains the base-route, it will not be there in the subsequent request uri.

src/Config/ObjectModel/RuntimeOptions.cs
ayush3797: Yeah, agree with you two on this. Thanks, incorporated this.

src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithoutIEnumerables.verified.txt
```diff
@@ -25,7 +25,8 @@
         Provider: StaticWebApps
       },
       Mode: Production
-    }
+    },
+    BaseRoute: /base-route
```
severussundar, line 29: This new element needs to be added to the schema file - not necessarily in the same PR

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -307,75 +320,76 @@ public static void ValidateRestPathForRelationalDbs(RuntimeConfig runtimeConfig)
...
+        /// any reserved characters. In case the URI component is not well formed the exception message containing
+        /// the reason for ill-formed URI component is returned. Else we return an empty string.
         /// </summary>
-        /// <param name="apiPath">path prefix for rest/graphql apis</param>
-        /// <param name="apiType">Either REST or GraphQL</param>
-        /// <exception cref="DataApiBuilderException">Thrown if the path is null/empty or it doesn't start with a preceding /.</exception>
-        private static void ValidateApiPath(string apiPath, ApiType apiType)
+        /// <param name="uriComponent">path prefix/base route for rest/graphql apis</param>
+        /// <returns>Exception message when the URI component is not well formed.</returns>
+        private static string ValidateAndGetUriComponentExceptionMsg(string? uriComponent)
```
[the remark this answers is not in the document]
ayush3797: done, thanks. Looks better.

src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithoutIEnumerables.verified.txt
ayush3797: Thanks, added.

seantleonard: @ayush3797 looks like the updated schema didn't make it in latest changes.

ayush3797: oops,missed that in git add!

seantleonard (approved):
LGTM!

src/Service.Tests/Configuration/ConfigurationTests.cs
Aniruddh25: I just realized, this assumption may NOT be always true in case DAB is not consumed by SWA. 
```
/// For the subsequent query, the rest base-route will be trimmed
/// by the upstream before the request lands at DAB.
```

What happens if DAB is stand alone hosted, and a base-route gets configured to some NON-default value e.g. the request is sent to:

`https://localhost:5000/base/api/EntityName`

how do we identify the EntityName? Do we error out?
if we are sending the `nextLink` with the complete route, we should be able to handle both the scenarios - a route reaching to us with base-route is still in the path(for non-SWA usecases) as well as a route where base-route is NOT in the path(for SWA use-cases).

Aniruddh25 (changes requested):
Question around how base-route is handled for non-SWA/frontend usecases when it is NOT stripped off before request reaches DAB.

src/Config/ObjectModel/RuntimeOptions.cs
```diff
@@ -3,4 +3,8 @@
 
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
-public record RuntimeOptions(RestRuntimeOptions Rest, GraphQLRuntimeOptions GraphQL, HostOptions Host);
+public record RuntimeOptions(RestRuntimeOptions Rest, GraphQLRuntimeOptions GraphQL, HostOptions Host, string? BaseRoute = null)
+{
+    public const string JSON_PROPERTY_NAME = "runtime";
+    public const string PROPERTY_NAME_BASE_ROUTE = "base-route";
```
Aniruddh25, line 9: Does this affect OpenAPI document generation? Do we prepend the entity routes with base-route information?

seantleonard: We'd need to append the base-route, if present, to the server property of the OpenApi doc:
https://github.com/Azure/data-api-builder/blob/bda0d6551f8f75fc7b98f888bdfce1f3dd317378/src/Core/Services/OpenAPI/OpenApiDocumentor.cs#L129-L132

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: Good point. I think we need to scope this feature scenario identical to how the SWA experience would work.

1. Using SWA, a client sends a request to `/data-api/*` where `*` is `<rest-api-path>/<entity_name>
    - SWA forwards requests to DAB but does not include `/data-api` in the request, which resulted in the issue being submitted in the first place.
    - This PR allows the nextLink to contain the base-url defined, which when followed by a client, is correctly resolved by dab because it is routed correctly to SWA, and then SWA drops that base-url from the route values that are passed to dab
2. [Your concern] Reverse proxy or Local(non-swa) environment with base route configured: nextLink would contain the base url and following that link would result in an error since DAB doesn't know how to distinguish the base-url from the rest of the route.

This PR should address number 1, but in it's current state does not prevent the error from number 2 from occurring.

Aniruddh25: That's right For #2, we should do one of the following: 
- Prevent this use case altogether for local environment which means - not allowing configuration through dab CLI and restricting its usage to SWA ( which is guaranteed to strip off the base-route before sending it to DAB)
OR 
- Prevent the error from occurring if we enable this for non-swa use cases.

ayush3797:
As per discussion with Ani over call:

1. We would do the validation that the base-route can be non-empty only for SWA.
2. We won't document/include this base-route option in the schema. This feature is only for system configuration and not for customers.

src/Config/ObjectModel/RuntimeOptions.cs
ayush3797: Thanks, added.

src/Service.Tests/Configuration/ConfigurationTests.cs
ayush3797: With the changes per this: https://github.com/Azure/data-api-builder/pull/1506#issuecomment-1635335758 incorporated, the base-route will only be coming into the picture when DAB is accessed via SWA.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2051,5 +1995,55 @@ string[] linkingTargetFields
...
+        [DataRow("/base-route", "AppService", true, "Runtime base-route is not used for non-Static Web Apps authentication providers.",
+            DisplayName = "Runtime base-route specified for non-Static Web Apps authentication provider - AppService.")]
+        [DataRow("/base+?route", "StaticWebApps", true, $"Runtime base-route {RuntimeConfigValidator.URI_COMPONENT_WITH_RESERVED_CHARS_ERR_MSG}",
+            DisplayName = "Runtime base-route specified for Static Web Apps authentication provider containing reserved characters +?")]
+        [DataRow("/base%&#route", "StaticWebApps", true, $"Runtime base-route {RuntimeConfigValidator.URI_COMPONENT_WITH_RESERVED_CHARS_ERR_MSG}",
+            DisplayName = "Runtime base-route specified for Static Web Apps authentication provider containing reserved characters %&#")]
+        [DataRow("base-route", "StaticWebApps", true, $"Runtime base-route should start with a '/'.",
+            DisplayName = "Runtime base-route specified for Static Web Apps authentication provider not starting with '/'")]
+        public void ValidateRuntimeBaseRouteSettings(
```
seantleonard, line 2017: May be worth having two additional tests to show that BaseRoute being null is handled properly :
1. base-route value is `null` and provider is non-EasyAuth ~ "otherProvider"
2. base-route value is `null` and provider is StaticWebApps

src/Cli/ConfigGenerator.cs
```diff
@@ -149,24 +149,34 @@ public static bool TryCreateRuntimeConfig(InitOptions options, RuntimeConfigLoad
...
             {
-                _logger.LogError($"Runtime {RuntimeOptions.PROPERTY_NAME_BASE_ROUTE} {RuntimeConfigValidator.URI_COMPONENT_WITH_RESERVED_CHARS_ERR_MSG}");
+                _logger.LogError($"Runtime base-route {RuntimeConfigValidator.URI_COMPONENT_WITH_RESERVED_CHARS_ERR_MSG}");
                 return false;
             }
 
+            if (runtimeBaseRoute is not null)
+            {
+                bool isEasyAuthProvider = Enum.TryParse(options.AuthenticationProvider, ignoreCase: true, out EasyAuthType easyAuthMode);
+                if (!isEasyAuthProvider || isEasyAuthProvider && easyAuthMode is not EasyAuthType.StaticWebApps)
```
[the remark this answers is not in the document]
ayush3797: correct thanks

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: There is no difference in between the two cases. Essentially because we cannot differentiate between the scenarios when base-route is configured as null vs when base-route is not specified at all. So, when the auth provider is not Static Web Apps, and base-route is defined as null, we don't throw an exception.

src/Cli/ConfigGenerator.cs
```diff
@@ -147,11 +149,33 @@ public static bool TryCreateRuntimeConfig(InitOptions options, RuntimeConfigLoad
...
+            {
+                _logger.LogError($"Runtime base-route {RuntimeConfigValidator.URI_COMPONENT_WITH_RESERVED_CHARS_ERR_MSG}");
+                return false;
+            }
+
+            if (runtimeBaseRoute is not null)
+            {
+                if (!Enum.TryParse(options.AuthenticationProvider, ignoreCase: true, out EasyAuthType easyAuthMode) || easyAuthMode is not EasyAuthType.StaticWebApps)
+                {
+                    _logger.LogError($"Runtime base-route can only be specified when authentication provider is Static Web Apps.");
```
seantleonard, line 174:
```suggestion
                    _logger.LogError($"Runtime base-route can only be specified when the authentication provider is Static Web Apps.");
```

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -720,4 +721,32 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+    /// <summary>
+    /// Test to verify that when base-route is configured, the runtime config is only successfully generated when the
+    /// authentication provider is Static Web Apps.
+    /// </summary>
+    /// <param name="authProvider">Authentication provider specified for the runtime.</param>
+    [DataTestMethod]
+    [DataRow("StaticWebApps")]
+    [DataRow("AppService")]
+    [DataRow("AzureAD")]
+    public void TestBaseRouteIsConfigurablForSWA(string authProvider)
```
seantleonard, line 734:
```suggestion
    public void TestBaseRouteIsConfigurableForSWA(string authProvider)
```

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -720,4 +721,32 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+    {
+        string[] initArgs = { "init", "-c", TEST_RUNTIME_CONFIG_FILE, "--host-mode", "development", "--database-type", "mssql",
+            "--connection-string", "localhost:5000", "--auth.provider", authProvider, "--runtime.base-route", "base-route" };
+        Program.Execute(initArgs, _cliLogger!, _fileSystem!, _runtimeConfigLoader!);
+
+        if (!Enum.TryParse(authProvider, ignoreCase: true, out EasyAuthType easyAuthProvider)
```
seantleonard, line 740: The Enum.TryParse logic already occurs in the CLI, so having that here wouldn't really be testing CLI functionality. You could achieve the desired result by having your test cases pass in a "isErrorExpected" argument.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -720,4 +721,32 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+
+    /// <summary>
+    /// Test to verify that when base-route is configured, the runtime config is only successfully generated when the
+    /// authentication provider is Static Web Apps.
+    /// </summary>
+    /// <param name="authProvider">Authentication provider specified for the runtime.</param>
+    [DataTestMethod]
+    [DataRow("StaticWebApps")]
+    [DataRow("AppService")]
+    [DataRow("AzureAD")]
```
seantleonard, line 733: For non -Easy Auth providers, you'll need to supply values for `--auth.issuer` and `--auth.audience` to the CLI otherwise you get the errors:
```
Error: Authentication providers other than EasyAuth and Simulator require both Audience and Issuer.
Error: Could not generate config file.
```

While this test uses the presence of the generated config file to confirm behavior, it doesn't necessarily validate that any failures were because of the base-route configuration value provided which are indicated with your added messages:

```
Error: Runtime base-route can only be specified when authentication provider is Static Web Apps.
Error: Could not generate config file.
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
seantleonard: Only reason i bring this up is to add coverage for the case that Ani is fixing in #1580. 

Your point is valid that validation only occurs if a value is actually provided. Wouldn't a null case here validate that statement?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -105,7 +105,7 @@ public void TestInitForCosmosDBPostgreSql()
     public void TestInitializingRestAndGraphQLGlobalSettings()
     {
         string[] args = { "init", "-c", TEST_RUNTIME_CONFIG_FILE, "--database-type", "mssql", "--rest.path", "/rest-api",
-                          "--rest.disabled", "--graphql.path", "/graphql-api" };
+                          "--runtime.base-route", "/base-route", "--rest.disabled", "--graphql.path", "/graphql-api" };
```
seantleonard, line 108: This argument seems out of place for testing/validation since this test is checking rest/graphql global settings behavior.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -241,7 +242,7 @@ public void TestAddEntityWithoutIEnumerable()
     public Task TestConfigGeneratedAfterAddingEntityWithoutIEnumerables()
     {
         string[] initArgs = { "init", "-c", TEST_RUNTIME_CONFIG_FILE, "--database-type", "mssql", "--connection-string", "localhost:5000",
-            "--set-session-context", "true" };
+            "--set-session-context", "true", "--runtime.base-route", "/base-route" };
```
seantleonard, line 245: The `"--runtime.base-route` is out of place here since this test isn't validating that behavior.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -720,4 +721,32 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+        if (!Enum.TryParse(authProvider, ignoreCase: true, out EasyAuthType easyAuthProvider)
+            || easyAuthProvider is not EasyAuthType.StaticWebApps)
+        {
+            Assert.IsFalse(_runtimeConfigLoader!.TryLoadConfig(TEST_RUNTIME_CONFIG_FILE, out RuntimeConfig? runtimeConfig));
+            Assert.IsNull(runtimeConfig);
+        }
+        else
+        {
+            Assert.IsTrue(_runtimeConfigLoader!.TryLoadConfig(TEST_RUNTIME_CONFIG_FILE, out RuntimeConfig? runtimeConfig));
+            Assert.IsNotNull(runtimeConfig);
```
seantleonard, line 749: You could relocate your assert from the rest/graphqlglobalsettings test here:
`Assert.AreEqual("/base-route", runtimeConfig.Runtime.BaseRoute);`

seantleonard (commented):
Feedback on latest updates.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: But when we see that base-route is null in runtimeconfigvalidator, we can't tell if it is null because of it being absent in the config or being explicitly specified as null. We can do something like this - we can mandate that if the base-route is specified, it has to be a non-empty string.

src/Cli.Tests/EndToEndTests.cs
seantleonard: [DataRow("AzureAD")] is still a valid and necessary test for coverage, isn't it? Just need additional logic to handle adding the extra params to the CLI call in that situation.

ayush3797: I have this in progress, needed some more refactor for it.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
seantleonard: > "we can't tell if it is null because of it being absent in the config or being explicitly specified as null" 

That part is fine. 

Without a null test case, do we have coverage that explicitly validates that "no unexpected errors or validations occur if `base-path` is null?"

> we can mandate that if the base-route is specified, it has to be a non-empty string.

Probably would want to go further than just empty string and disallow a whitespace only base-route. You may want to update `if (!string.IsNullOrEmpty(runtimeBaseRoute))` to `if (!string.IsNullOrWhitespace(runtimeBaseRoute))` in runtimeconfigvalidator and add a test to confirm behavior.

ayush3797: Well actually, we should continue with the current scenario. We should be allowed to specify null/whitespace base-route. Even though right now only SWA will use this base-route feature with value "/data-api" but we shouldn't put any restriction. We would just not append base-route in the uri if it is null/whitespace.

src/Cli.Tests/EndToEndTests.cs
ayush3797: done

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -359,75 +381,72 @@ public static void ValidateRestPathForRelationalDbs(RuntimeConfig runtimeConfig)
...
-        /// <param name="apiType">Either REST or GraphQL</param>
-        /// <exception cref="DataApiBuilderException">Thrown if the path is null/empty or it doesn't start with a preceding /.</exception>
-        private static void ValidateApiPath(string apiPath, ApiType apiType)
+        /// <param name="uriComponent">path prefix/base route for rest/graphql apis</param>
+        /// <returns>false when the URI component is not well formed.</returns>
+        private static bool TryValidateUriComponent(string? uriComponent, out string exceptionMessageSuffix)
         {
-            if (string.IsNullOrEmpty(apiPath))
+            exceptionMessageSuffix = string.Empty;
+            ;
```
seantleonard, line 419:
```suggestion
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
seantleonard: > Well actually, we should continue with the current scenario. We should be allowed to specify null/whitespace base-route.

I agree that we should not append the base route in the url if base route is null. 

However, if the base route is whitespace, that implies that the end user intentionally provided a value for that property. If we allow that we should add a comment for 'why?' and also add public documentation that whitespace only will be ignored, even if it doesn't cause a validation failure.
- uri spec allows whitespace in url but the whitespace values will be escaped by browsers/and uri helper functions. Though a whitespace only base url is technically not forbidden by spec.. my sentiment earlier for why i initially suggested it shouldn't be allowed is that it doesn't make sense to me as a valid use case. If we want to disallow it in the future, we can revisit in a future pr.

src/Config/ObjectModel/AuthenticationOptions.cs
```diff
@@ -36,4 +36,10 @@ public record AuthenticationOptions(string Provider, JwtOptions? Jwt)
     /// </summary>
     /// <returns>True if the provider is enabled for JWT, otherwise false.</returns>
     public bool IsJwtConfiguredIdentityProvider() => !IsEasyAuthAuthenticationProvider() && !IsAuthenticationSimulatorEnabled();
+
+    /// <summary>
+    /// A shorthand method to determine whether Static Web Apps is configured for the current authentication provider.
+    /// </summary>
+    /// <returns>True if the provider is enabled for JWT, otherwise false.</returns>
```
Aniruddh25, line 43:
```suggestion
    /// <returns>True if the provider is enabled for StaticWebApps authentication, otherwise false.</returns>
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2051,5 +1995,65 @@ string[] linkingTargetFields
...
+        /// <param name="runtimeBaseRoute">Value of runtime base-route as specified in config.</param>
+        /// <param name="authenticationProvider">The authentication provider configured.</param>
+        /// <param name="isExceptionExpected">Whether an exception is expected as a result of test run.</param>
+        /// <param name="expectedExceptionMessage">The expected exception message.</param>
+        [DataTestMethod]
+        [DataRow("/base-route", "StaticWebApps", false, DisplayName = "Runtime base-route correctly configured for Static Web Apps.")]
+        [DataRow(null, "StaticWebApps", false,
+            DisplayName = "Runtime base-route specified as null for Static Web Apps authentication provider passes config validation.")]
+        [DataRow(null, "AppService", false,
+            DisplayName = "Runtime base-route specified as null for Static Web Apps authentication provider passes config validation.")]
```
Aniruddh25, line 2012: nit: should AppService really pass?

Aniruddh25 (approved):

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Let me correct myself. Base-route should not be appended if it is null. And if the user intentionally provides it as  string containing whitespaces (i.e. `/      `, we should throw an exception as this makes the url invalid.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: Yes, null base-route is ignored for every auth provider and hence it will pass for AppService.

seantleonard (approved):

PR #1525, opened 2023-05-30
Fixing bug with only __typename in the selection set

### LINKED ISSUE #1463: Selection set without entity fields returns an error response for GraphQL query and mutation
For the below query and mutations, an error response is returned

Query:

```graphql
{
    book_by_pk(id: 1){
         __typename
     }
}
```
pk query bug

Mutation:

```graphql
mutation {
     updatebook(id: 1, item: {title: "Another great book"}){
        __typename  
     }
}
```
Mutation bug

Root Cause: 
The SELECT query is constructed without any column. This leads to failure at the database layer.

Since, there are no entity fields in the selection set, the flow should not go to the database layer.

severussundar:
## Why make this change?

-  Closes #1463 
  - Queries and Mutations with only _typename in the selection set fails at the database layer because there are no columns to choose in the database query. 
  ```graphql
       {
          book_by_pk(id: 1){
               __typename
           }
       }
     
```

pk query bug

## What is this change?

- During the construction of `SqlQueryStructure`, if it is observed that there are no columns selected, then one of the primary keys is added so that the eventual database query constructed contains a valid column name in the SELECT statement.  

## How was this tested?

- [X] Integration Tests
- [X] Manual Tests

### CODE REVIEW

File: src/Service/Resolvers/Sql Query Structures/SqlQueryStructure.cs [hunk 1]
```diff
@@ -395,19 +395,19 @@ private List<OrderByColumn> PrimaryKeyAsOrderByColumns()
...
                 {
                     _limit++;
                 }
             }
 
+            // If there are no columns, add the primary key column
+            // to prevent failures when executing the database query.
+            if (!Columns.Any())
+            {
+                AddColumn(PrimaryKey()[0]);
```
Aniruddh25, line 408: what if there are multiple primary key columns? Are we choosing only the first? 

For pagination, it may make sense to still hit the database by creating a query so that it gets the hasNextPage field correctly.
But for the cases when there is no database column to project or no dependency on database results why do we even try to hit the database?

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -387,6 +387,228 @@ public virtual async Task MultipleResultJoinQuery()
...
+            string graphQLQueryName = "executeGetBooks";
+            string graphQLQuery = @"{
+                executeGetBooks{
+                    __typename
+                }     
+            }";
+
+            string expected = @"
+                [
+      {
```
Aniruddh25, line 545: What decides how many __typename items should be returned back as part of the response array?

Aniruddh25 (changes requested):
Waiting for few answers, perhaps the optimization of not hitting the db can be a followup PR

aaronpowell:
> Waiting for few answers, perhaps the optimization of not hitting the db can be a followup PR

I would have expected that to be the better approach, since you don't actually need any data back from the DB, _but_ I can see there being a problem when you've done a query like:

```graphql
query {
  books {
    items {
      __typename
    }
  }
}
```

That should return a `items` array that has _n_ `__typename` properties in it, where _n_ is the count of records in the query (up to 100).

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
severussundar: There would be no change in the number of items returned in the response when the selection set contains entity fields vs just __typename field. 

- For instance, let's say a policy is defined - `@item.id ne 1`. This would be honored and the response will not contain an entry `__typename: <actual_type>` for id corresponding to 1. 

- For a query such as this, 
```graphql
{
  book_by_pk (id: 11111){
    __typename
  }

}

```
where id `11111` is not present in the table, the response will be the following:

```graphql
{
  "data": {
    "book_by_pk": null
  }
}
```
which is the same when the selection set contains entity fields such as `id`, `title`, etc.



Essentially, the number of items in the response array will be the same except that it would contain just the type names.

src/Service/Resolvers/Sql Query Structures/SqlQueryStructure.cs
severussundar: Yes, when the PK is composite, only the first key/column is chosen. Having just one valid column in the `select` query makes it run successfully. I have added a test case for this scenario where the Pk is composite. 


There are only very few cases where it is possible to completely avoid hitting the database. 
Only in this case (and similar one for stored procedure),
```graphql
{
  books{
    __typename
  }
}
```
it might be possible to avoid hitting the database. 


In all other cases, such as these, the result depends on the items present in the database, policies defined etc.

```graphql

  {
       books{
         items{
           __typename
         }
       }
   }

```
- In the above query with `filter` or `first` defined would necessitate querying the database.

- Incase of nested queries, where some of the selection sets contain entity fields while the rest contain only `__typename` such as this
 
```graphql
{
  publishers {
    items {
      id
      books {
        items {
          __typename
        }
      }
    }
  }
}
```
also necessitates that we query the database.

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -387,6 +387,209 @@ public virtual async Task MultipleResultJoinQuery()
...
+        ""__typename"": ""GetBooks""
```
Aniruddh25, line 546: Instead of hard coding this element "GetBooks" the required number of times, can't we generate such an expected string programmatically?

Aniruddh25 (approved):
LGTM after 1 optimization to generate expected output

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
severussundar: Thanks for the suggestion! I have updated the test to dynamically generate this response string.

src/Service.Tests/SqlTests/SqlTestHelper.cs
```diff
@@ -305,6 +306,29 @@ public static SqlException CreateSqlException(int number, string message = "")
             return e;
         }
 
+        /// <summary>
+        /// Helper method to construct GraphQL responses when only __typename is queried
+        /// </summary>
+        /// <param name="typename">A json string of the format { __typename : entity_typename }  </param>
+        /// <param name="times">Number of times to repeat typename in the response</param>
+        /// <returns>An string representation of an array of typename json strings</returns>
```
Aniruddh25, line 314:
```suggestion
        /// <returns>A string representation of an array of typename json strings</returns>
```

Aniruddh25 (approved):
LGTM. Thanks for addressing all comments.

ayush3797 (approved):
LGTM!

PR #1529, opened 2023-06-08
Excluding null values from the generated config

aaronpowell:
Enabling a setting on the serializer that will exclude `null` when it's serializing the runtime config file.


src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -28,10 +28,7 @@
       },
       "authentication": {
         "provider": "StaticWebApps",
-        "jwt": {
-          "audience": null,
-          "issuer": null
-        }
+        "jwt": {}
```
Aniruddh25, line 31: Why is the `jwt` object added if its anyway empty?

src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -40,9 +37,7 @@
     "Planet": {
   "source": {
     "object": "graphqldb.planet",
-    "type": "table",
-    "parameters": null,
-    "key-fields": null
+    "type": "table"
```
Aniruddh25, line 40: this is interesting, the `type` should not be here since its cosmosdb_nosql, besides `table` doesnt make sense.

src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -105,48 +84,28 @@
...
+          "policy": {}
         },
         {
           "action": "read",
-          "fields": null,
-          "policy": {
-            "request": null,
-            "database": null
-          }
+          "policy": {}
```
Aniruddh25, line 91: policy is not supported for cosmosdb_nosql at this point. Can we skip writing the policy as well? Can we define it nullable so it doesnt write to the file?

aaronpowell: We'd be better off making `Policy` on `EntityAction` a nullable property so we can null it out, rather than nulling its properties.

I'll see what the complexity/cascading impacts of that is.

src/Service.Tests/dab-config.CosmosDb_NoSql.json
aaronpowell: Could we get away with making a new `EntitySourceType` of `Collection` that is for CosmosDB NoSQL?

I worry that it's so heavily relied on as a non-nullable property in SQL that we'd spend a lot of time having to either use `!` to ignore the null check, or writing null checks for minimal value.

This is one of the use cases that surfaced when doing the new config engine that there are very tiny discrepancies between how SQL and Cosmos work and that we've kind of just lived with having to do constant type checking.

src/Service.Tests/dab-config.CosmosDb_NoSql.json
aaronpowell: Made `Policy` nullable and that has cleaned up the config file even more.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -124,6 +125,7 @@ public static JsonSerializerOptions GetSerializationOptions()
             PropertyNamingPolicy = new HyphenatedNamingPolicy(),
             ReadCommentHandling = JsonCommentHandling.Skip,
             WriteIndented = true,
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
```
seantleonard, line 128: Cool that this is available. Essentially resulting in null properties not being written to the config?

src/Config/Converters/EntityRestOptionsConverter.cs
```diff
@@ -100,7 +100,12 @@ public override void Write(Utf8JsonWriter writer, EntityRestOptions value, JsonS
     {
         writer.WriteStartObject();
         writer.WriteBoolean("enabled", value.Enabled);
-        writer.WriteString("path", value.Path);
+
+        if (value.Path is not null)
+        {
+            writer.WriteString("path", value.Path);
+        }
```
seantleonard, line 107: Is it possible to have this property written by default? and if null, abide by the JsonSerializer option `DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull` to not write out the property?

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -19,4006 +19,2989 @@
...
-              }
-            }
-          ]
+  "source": {
+    "object": "publishers",
+    "type": "table"
+  },
+  "graphql": {
+    "enabled": true,
+    "operation": null,
```
seantleonard, line 41: Is this another property that can be dropped if null?

src/Service.Tests/dab-config.MySql.json
```diff
@@ -23,2149 +23,2102 @@
...
         {
-          "role": "authenticated",
-          "actions": [
-            {
-              "action": "*",
-              "policy": {}
-            }
-          ]
+          "action": "*",
+          "policy": {}
```
Aniruddh25, line 2113: Looks like `"policy": { }`  needs to be removed from these configs as well

src/Service.Tests/dab-config.MySql.json
```diff
@@ -23,2149 +23,2102 @@
...
+      ]
     }
+  ],
+  "mappings": {
+    "id": "bkid",
+    "bkname": "name"
   }
 }
+  }
+}
```
Aniruddh25, line 2124:
```suggestion
}

```

src/Service.Tests/dab-config.CosmosDb_NoSql.json
Aniruddh25: If it was made nullable, why do we write the empty object to the config file. Cant we skip writing it altogether?

seantleonard (commented):
some suggestions/question

src/Config/Converters/EntityRestOptionsConverter.cs
aaronpowell: That's a good question, this might not be needed. I'll verify and remove if it isn't.

src/Config/RuntimeConfigLoader.cs
aaronpowell: Yep! I didn't know that that existed, made this change a whole lot simpler 😁

src/Service.Tests/dab-config.CosmosDb_NoSql.json
aaronpowell: Looks like I've missed a place where it's being initialized still. I'll clean that up.

src/Service.Tests/dab-config.MsSql.json
aaronpowell: Going back to your previous comment about writing nulls, looks like you have to explicitly skip it if you're doing your own converter. The `EntityGraphQLConverter` is causing that.

src/Service.Tests/dab-config.MySql.json
aaronpowell: Given this file is auto-generated, manually adding new lines won't be a good idea.

src/Config/Converters/EntityRestOptionsConverter.cs
aaronpowell: It'll be needed as we are owning how the properties end up in the `writer` in this case, so we can still write nulls, even if the `DefaultIgnoreCondition` says not to.

I've changed this so that we respect what the `DefaultIgnoreCondition` is regarding `null`, so that if we change it in the future the code should adapt.

aaronpowell:
There's a lot of snapshots that are updated with the change to Policy, which is why there seems to be so many files - and this makes it more obvious where the change impacts across our testing too.

src/Service.Tests/dab-config.MySql.json
```diff
@@ -23,2149 +23,1922 @@
...
```
Aniruddh25, line 1944:
```suggestion
}

```

src/Service.Tests/dab-config.CosmosDb_NoSql.json
Aniruddh25: #1530 addressed this

src/Cli/ConfigGenerator.cs
```diff
@@ -920,7 +922,7 @@ public static bool TryStartEngineWithOptions(StartOptions options, RuntimeConfig
             loader.UpdateBaseConfigFileName(runtimeConfigFile);
 
             // Validates that config file has data and follows the correct json schema
-            if (!loader.TryLoadKnownConfig(out RuntimeConfig? deserializedRuntimeConfig))
+            if (!loader.TryLoadConfig(runtimeConfigFile, out RuntimeConfig? deserializedRuntimeConfig))
```
Aniruddh25, line 925: This seems to be an incorrect merge

seantleonard: Fixed!

Aniruddh25 (approved):
LGTM

src/Cli/ConfigGenerator.cs
```diff
@@ -934,7 +936,7 @@ public static bool TryStartEngineWithOptions(StartOptions options, RuntimeConfig
 
             /// This will add arguments to start the runtime engine with the config file.
             List<string> args = new()
-            { "--ConfigFileName", runtimeConfigFile };
+            { "--" + nameof(RuntimeConfigLoader.CONFIGFILE_NAME), runtimeConfigFile };
```
seantleonard, line 939: This doesn't seem right, should be either `-c` `--config` 
image

src/Cli.Tests/UtilsTests.cs
```diff
@@ -249,13 +249,22 @@ public void TestMergeConfig()
         fileSystem.AddFile("dab-config.Test.json", new MockFileData(ENV_BASED_CONFIG));
 
         RuntimeConfigLoader loader = new(fileSystem);
+        bool old = RuntimeConfigLoader.CheckPrecedenceForConfigInEngine;
```
seantleonard, line 252: I lost our previous comment thread on this. We save the old value and set it to true for this specific test. However, if other tests change this flag without reverting the value back to how it was prior to test execution, does that mean those tests may not be deterministic?

seantleonard (approved):
Two questions, but looks good!

src/Cli.Tests/UtilsTests.cs
aaronpowell: It's mostly a safety thing that we're catching it here, since it's a static value that we're working with the value will be impacted by whatever test was run before it, and we'll then impact all other tests with the following line.

Although, I thought I removed that property...

src/Cli/ConfigGenerator.cs
aaronpowell: Yes, and I thought I fixed that in the branch for #1402, so I'll check this branch still works and is up to date.

aaronpowell: Checked against main, this I think is wrong in `main` too... but shouldn't tests fail?

aaronpowell:
This PR is reintroducing some stuff that it shouldn't, I'll have to check all the .NET changes again before it can be merged

src/Cli/ConfigGenerator.cs
aaronpowell: It is an incorrect value but it's not that it should be `config` it should be `ConfigFileName` to match https://github.com/Azure/data-api-builder/blob/2e8d1bb8b7d83bbe2a90472c2ccdc8acb87fa5ec/src/Service/Startup.cs#L61

This change is the result of a mistaken merge with the branch #1402, so I'll revert.

src/Cli.Tests/UtilsTests.cs
Aniruddh25: I thought so too that you had removed this property in #1402. Maybe an incorrect merge?

src/Cli/ConfigGenerator.cs
Aniruddh25: > This doesn't seem right, should be either `-c` `--config` image

The args for starting cli are `-c`, `--config` using `dab start`. 
However, the args for running the engine directly used `--ConfigFileName` so we should continue to support both to avoid regression

src/Cli.Tests/UtilsTests.cs
aaronpowell: Yeah, something stuffed up with the merge I reckon, so I reverted it from the PR too.

src/Cli/ConfigGenerator.cs
aaronpowell: We do support both, but in the case where this line of code is called it's only launching the service using information provided from the CLI.

PR #1530, opened 2023-06-09
Making the EntitySource.Type nullable

aaronpowell:
When it is a CosmosDB_NoSQL config, the entity 'type' is not important as there are no types in CosmosDB_NoSQL, so we don't want anything there.

Combining this with #1529 will see the value removed from the config entirely

### DIFF
--- a/src/Service.Tests/dab-config.PostgreSql.json
+++ b/src/Service.Tests/dab-config.PostgreSql.json
@@ -17,9 +17,7 @@
     },
     "host": {
       "cors": {
-        "origins": [
-          "http://localhost:5000"
-        ],
+        "origins": ["http://localhost:5000"],
         "allow-credentials": false
       },
       "authentication": {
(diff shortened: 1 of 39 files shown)


Aniruddh25 (approved):

ayush3797 (approved):

src/Service.Tests/dab-config.PostgreSql.json
```diff
@@ -34,3485 +34,3485 @@
...
-          "cardinality": "many",
-          "target.entity": "Book",
-          "source.fields": [],
-          "target.fields": [],
-          "linking.object": null,
-          "linking.source.fields": [],
-          "linking.target.fields": []
-        }
-      }
+  "source": {
```
abhishekkumams, line 37: incorrect indentation

abhishekkumams: probably no change is required in this file. maybe just revert the identation change

abhishekkumams (approved):

src/Service.Tests/dab-config.PostgreSql.json
aaronpowell: Because of how the config generation happens indentation is problematic, and the indentation on the right is actually "correct"

aaronpowell: There are some minor stylistic changes and it seems very dependent on which formatter you use 😅

PR #1541, opened 2023-06-21
fix init command with environment variable

### LINKED ISSUE #1527: [Bug]: DAB init doesn't honor DAB_ENVIRONMENT variable
### What happened?

when `dab init` is called with environment variable `DAB_ENVIRONMENT=Development`. It should create a file with name `dab-config.Development.json`. Currently, it's only creating file dab-config.json

Although it does error out if dab-config.Development.json is already present, but if not present it should create this file.
image

### Version

running from main

Azure SQL

Local (including CLI)

_No response_

### Relevant log output

```Text
PS C:\myDir\dab-testing\latest-test> dab init --database-type mssql
Information: Microsoft.DataApiBuilder 1.0.0+e1bf19cc334b19534a30ab5f70274bba4bd81e24
Information: Environment variable DAB_ENVIRONMENT is Development
Unable to find config file: dab-config.Development.json does not exist.
Found config file: dab-config.json.
Error: Config file: dab-config.json already exists. Please provide a different name or remove the existing config file.
Error: Could not generate config file.
```

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #1527 
  - dab init doesn't create file with the environment Value, and always create default config if not specified explicitly.
  - For ex: if DAB_ENVIRONMENT is set to Development, it should create dab-config.Development.json, which is not the case.

## What is this change?

- Updating the logic to create config file giving precedence to environment value, also making sure error is thrown if the file is not present already.

## How was this tested?

- [X] Unit Tests

## Sample Request(s)

`dab init --database-type mssql`
above command will create:
1. DAB_ENVIRONMENT = null
dab-config.json
2. DAB_ENVIRONMENT = Development
dab-config.Development.json


src/Cli/ConfigGenerator.cs
```diff
@@ -32,24 +32,35 @@ public class ConfigGenerator
         /// </summary>
         public static bool TryGenerateConfig(InitOptions options)
         {
-            if (!TryGetConfigFileBasedOnCliPrecedence(options.Config, out string runtimeConfigFile))
+            string runtimeConfigFile = RuntimeConfigPath.DefaultName;
+            string? environmentValue = Environment.GetEnvironmentVariable(RuntimeConfigPath.RUNTIME_ENVIRONMENT_VAR_NAME);
```
Aniruddh25, line 36: Can we GetEnvironmentVariable in the else?
This saves us 1 call if user already provides a config file

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// <summary>
+    /// Tests that environment config file is successfully generated when that file is not already present.
+    /// When environment variable is not set, it should generate the default config file.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow(true, false, "Test", "dab-config.Test.json", DisplayName = "Failed to generate the config file when environment config file is present.")]
+    [DataRow(false, true, "Test", "dab-config.Test.json", DisplayName = "Successfully generated the config file when environment config file is not present.")]
+    [DataRow(false, true, "", "dab-config.json", DisplayName = "Successfully generated the config file when environment config file is not present and environment variable is not set.")]
```
Aniruddh25, line 51: Isnt the last DataRow already being tested by other tests:

`Successfully generated the config file when environment config file is not present and environment variable is not set.`

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+    public void TryGenerateConfig_UsingEnvironmentVariable(
+        bool isConfigFilePresentAlready,
+        bool isConfigGenerationSuccessful,
+        string environmentValue,
+        string configFileName)
+    {
+        Environment.SetEnvironmentVariable(RUNTIME_ENVIRONMENT_VAR_NAME, environmentValue);
+        HandleConfigFileCreationAndDeletion(configFileName, isConfigFilePresentAlready);
```
Aniruddh25, line 59: What would happen if this test and the above test run in parallel? would the file creation/deletion interfere with each other?

src/Cli/ConfigGenerator.cs
```diff
@@ -32,24 +32,35 @@ public class ConfigGenerator
...
+            if (!string.IsNullOrWhiteSpace(options.Config))
+            {
+                _logger.LogInformation("Generating user provided config file: {configFileName}", options.Config);
```
seantleonard, line 39:
```suggestion
                _logger.LogInformation("Generating user provided config file path: {configFileName}", options.Config);
```

either add path or name.. otherwise it reads as if the user is providing the actual file

src/Cli/ConfigGenerator.cs
```diff
@@ -32,24 +32,35 @@ public class ConfigGenerator
...
-                    "Please provide a different name or remove the existing config file.");
+                _logger.LogError("Config file: {configFileName} already exists. Please provide a different name or remove the existing config file.", runtimeConfigFile);
                 return false;
             }
 
             // Creating a new json file with runtime configuration
             if (!TryCreateRuntimeConfig(options, out string runtimeConfigJson))
             {
-                _logger.LogError($"Failed to create the runtime config file.");
+                _logger.LogError("Failed to create the runtime config file.");
```
seantleonard, line 63: This error message seems inaccurate. There was a failure creating the runtime config with the provided options.

Failure to create the runtime config file is already logged in `Utils.WriteJsonContentToFile(fileName, configJson)`
Additionally, that same utils function may benefit from logging "successfully created config file {configFileName}" where the file name is the FULL path, which would help in debugging.

src/Cli/ConfigGenerator.cs
```diff
@@ -32,24 +32,35 @@ public class ConfigGenerator
...
-                _logger.LogInformation($"Creating a new config file: {runtimeConfigFile}");
+                _logger.LogInformation("Generating default config file: {config}", runtimeConfigFile);
             }
 
             // File existence checked to avoid overwriting the existing configuration.
             if (File.Exists(runtimeConfigFile))
             {
-                _logger.LogError($"Config file: {runtimeConfigFile} already exists. " +
-                    "Please provide a different name or remove the existing config file.");
+                _logger.LogError("Config file: {configFileName} already exists. Please provide a different name or remove the existing config file.", runtimeConfigFile);
```
seantleonard, line 56: This would also benefit from logging the full path of the file name to assist with debugging. If dab cli installed via Nuget, it may not be immediately obvious where the config file is saved if user doesn't provide a full path.

src/Config/RuntimeConfigPath.cs
```diff
@@ -233,6 +233,11 @@ private static string GetOverriddenName(string fileName)
             return $"{fileName}.overrides{CONFIG_EXTENSION}";
         }
 
+        public static string GetEnvironmentFileName(string fileName, string environmentValue)
```
seantleonard, line 236: in a method comment, can you note that fileName must not include an extension? i.e. not `myConfig.json`

src/Cli.Tests/TestHelper.cs
```diff
@@ -1133,5 +1133,25 @@ public static void SetupTestLoggerForCLI()
...
+            return new(
+                databaseType: DatabaseType.mssql,
+                connectionString: "testconnectionstring",
+                cosmosNoSqlDatabase: null,
+                cosmosNoSqlContainer: null,
+                graphQLSchemaPath: null,
+                setSessionContext: true,
+                hostMode: HostModeType.Development,
+                corsOrigin: new List<string>(),
+                authenticationProvider: EasyAuthType.StaticWebApps.ToString(),
```
seantleonard, line 1153: Do other sample configs also set the authentication provider to `StaticWebApps`?

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// <summary>
+    /// Tests that user provided config file is successfully generated when that file is not already present.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow(true, false, DisplayName = "Failed to generate config file when user provided config file is present.")]
+    [DataRow(false, true, DisplayName = "Successfully generated config file when user provided config file is not present.")]
+    public void TryGenerateConfig_WithUserProvidedConfig(
+        bool isConfigFilePresentAlready,
```
seantleonard, line 28:
```suggestion
        bool isConfigFilePresent,
```

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        string environmentValue,
+        string configFileName)
+    {
+        Environment.SetEnvironmentVariable(RUNTIME_ENVIRONMENT_VAR_NAME, environmentValue);
+        HandleConfigFileCreationAndDeletion(configFileName, isConfigFilePresentAlready);
+        Assert.AreEqual(isConfigFilePresentAlready, File.Exists(configFileName));
+
+        InitOptions options = CreateBasicInitOptionsForMsSqlWithConfig();
+
+        Assert.AreEqual(isConfigGenerationSuccessful, ConfigGenerator.TryGenerateConfig(options));
```
seantleonard, line 64: To explicitly validate that `ConfigGenerator.TryGenerateConfig(options)` failed because the specified file already exists, would you be able to validate that the ilogger logged ~~"Config file: {configFileName} already exists."

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+    [DataRow(true, false, "Test", "dab-config.Test.json", DisplayName = "Failed to generate the config file when environment config file is present.")]
+    [DataRow(false, true, "Test", "dab-config.Test.json", DisplayName = "Successfully generated the config file when environment config file is not present.")]
+    [DataRow(false, true, "", "dab-config.json", DisplayName = "Successfully generated the config file when environment config file is not present and environment variable is not set.")]
+    public void TryGenerateConfig_UsingEnvironmentVariable(
+        bool isConfigFilePresentAlready,
+        bool isConfigGenerationSuccessful,
+        string environmentValue,
+        string configFileName)
+    {
+        Environment.SetEnvironmentVariable(RUNTIME_ENVIRONMENT_VAR_NAME, environmentValue);
```
seantleonard, line 58: Can you remind me whether `RUNTIME_ENVIRONMENT_VAR_NAME` is guaranteed to be set in production / customer environments, and set by which mechanism? If we can't guarantee the present of the key `DAB_ENVIRONMENT`, are there tests covering scenarios where the env var key is not even present (and not just empty)?

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+    public static void CleanUp()
+    {
+        if (File.Exists("dab-config.Test.json"))
+        {
+            File.Delete("dab-config.Test.json");
+        }
+
+        if (File.Exists("dab-config.json"))
+        {
+            File.Delete("dab-config.json");
```
seantleonard, line 100: You don't need the existence check in `File.Delete(file)`: "If the file to be deleted does not exist, no exception is thrown." https://learn.microsoft.com/en-us/dotnet/api/system.io.file.delete?view=net-7.0#remarks

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            Assert.AreEqual(isConfigGenerationSuccessful, File.Exists(configFileName));
+        }
+    }
+
+    /// <summary>
+    /// This method handles the creation and deletion of a configuration file.
+    /// </summary>
+    private static void HandleConfigFileCreationAndDeletion(string configFilePath, bool configFilePresent)
+    {
+        if (File.Exists(configFilePath) && !configFilePresent)
```
seantleonard, line 76: This could be simplified given the behavior of File.Delete mentioned in another comment.

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            File.Delete(configFilePath);
+        }
+        else if (!File.Exists(configFilePath) && configFilePresent)
+        {
+            File.Create(configFilePath).Dispose();
+        }
```
seantleonard, line 83: Could this be simplified given that File.Create overwrites the file if it already exists?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -18,7 +18,12 @@ public static void Setup(TestContext context)
     {
         if (!File.Exists(TEST_SCHEMA_FILE))
         {
-            File.Create(TEST_SCHEMA_FILE);
+            File.Create(TEST_SCHEMA_FILE).Dispose();
+        }
+
+        if (File.Exists(TEST_RUNTIME_CONFIG_FILE))
+        {
+            File.Delete(TEST_RUNTIME_CONFIG_FILE);
```
seantleonard, line 26: Could this be simplified given File.Delete behavior? also, is it desireable to create the file during every classInitialize because pre-existing file may lead to unintended side effects?

seantleonard (changes requested):
Some questions and suggestions for simplification

src/Cli.Tests/TestHelper.cs
abhishekkumams: Yep, all the test configs use `StaticWebApps`

src/Cli/ConfigGenerator.cs
abhishekkumams: we need the value before that since we are validating the condition in elseif

src/Cli.Tests/ConfigGeneratorTests.cs
abhishekkumams: yep, right. Thanks for flagging.

src/Cli/ConfigGenerator.cs
abhishekkumams: using the Utils method for logs, and removing extra logs here.

src/Cli.Tests/ConfigGeneratorTests.cs
abhishekkumams: NO, it's guaranteed to be set, and it can be set by directly logging into the console of the production system. Added a test to cover that

src/Cli.Tests/ConfigGeneratorTests.cs
abhishekkumams: added validation for logger message

src/Cli.Tests/ConfigGeneratorTests.cs
abhishekkumams: In C#, by default, the  TestMethod  under a  TestClass  does not run in parallel. Each test method is executed sequentially, one after another. This default behavior ensures that each test is executed in isolation and avoids any potential interference between test cases. 
Although we can parallelize it by using `[Parallelize]` attribute.

seantleonard (approved):
Thanks for fixing! also for addressing comment, very clean. Appreciate the ability to see full file path in the logs, will help us/PMs/customers troubleshoot if necessary!

src/Cli/ConfigGenerator.cs
Aniruddh25: We could write the else as 

```
else
{
envValue =  GetEnvironmentVariable()
 if (envValue)
}
```

abhishekkumams: but we have another else as well, which should only execute when the above two conditions are not met.

src/Cli/ConfigGenerator.cs
```diff
@@ -36,23 +36,38 @@ public class ConfigGenerator
...
+                    _logger.LogInformation("Generating default config file: {config}", Path.GetFullPath(runtimeConfigFile));
+                }
             }
 
             // File existence checked to avoid overwriting the existing configuration.
             if (fileSystem.File.Exists(runtimeConfigFile))
             {
-                _logger.LogError("Config file: {runtimeConfigFile} already exists. Please provide a different name or remove the existing config file.", runtimeConfigFile);
+                _logger.LogError("Config file: {runtimeConfigFile} already exists. Please provide a different name or remove the existing config file.",
+                    Path.GetFullPath(runtimeConfigFile));
```
Aniruddh25, line 64: What if Path.GetFullPath throws exception?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -230,6 +230,15 @@ private static string GetOverriddenName(string fileName)
         return $"{fileName}.overrides{CONFIG_EXTENSION}";
     }
 
+    /// <summary>
+    /// Generates the name of the file based on environment value.
+    /// NOTE: Input File name should not contain extension
```
Aniruddh25, line 235: What if it contains the file extension name? How would the end user know NOT to provide the extension? 

Can we be flexible? Check for json extension at the end, if not present add it?

Aniruddh25 (approved):
LGTM, after fixing some corner cases.

src/Cli/ConfigGenerator.cs
aaronpowell: We should be using `fileSystem.Path` not `Path` directly to ensure we're working against the right file system.

aaronpowell (dismissed):

src/Config/RuntimeConfigLoader.cs
abhishekkumams: This method is not invoked by user. It's an internal method which is invoked when a default scenario is triggered.
The comment is intended for developers who would may want to reuse this method for other purposes.

PR #1548, opened 2023-06-26
Fix REST behavior for Stored Procedures when methods property is absent in the config file 

### LINKED ISSUE #1504: [Bug]: adding the `rest` object in an entity, remove the default value for the "method" property
### What happened?

I have an entity based on a stored procedure 

```json
"GetBusNearPosition": {
      "source": {
        "object": "dbo.GetBusNearPosition",
        "type": "stored-procedure",
        "parameters": {
          "latitude": "0",
          "longitude": "0",
          "radius": 500
        }
      },      
      "permissions": [
        {
          "role": "anonymous",
          "actions": [ "*" ]
        }
      ],     
    }
```

and by default the entity is available, using REST, via POST at `/api/GetBusNearPosition` 

If I add the `rest` object to configure the route:

```json
   "rest": {           
        "path": "/bus-near-position"        
   }
```

the default support for the POST verb is gone. If I try to call the endpoint:

```http
POST http://localhost:5000/api/bus-near-position
```

I get a 405 error:

```text
HTTP/1.1 405 Method Not Allowed
Connection: close
Content-Type: application/json; charset=utf-8
Date: Tue, 16 May 2023 21:37:14 GMT
Server: Kestrel
Transfer-Encoding: chunked
x-ms-correlation-id: 6b3db163-426f-454d-8eb1-79cb839d1078

{
  "error": {
    "code": "BadRequest",
    "message": "This operation is not supported.",
    "status": 405
  }
}
```

In order to make everything working again I need to add the `methods` property to the `rest` object:

```json
   "rest": {           
        "methods": [ "post" ]
        "path": "/bus-near-position"        
   }
```

With this everything works again.

It seems that the property `methods` is considered empty if there is a `rest` object that doesn't explicitly contain `methods`. The expectation is that, if no `methods` is present, the default, for a Stored Procedure, is `post`.

Tested both on 0.6.14 and 0.7.5

### Version

Microsoft.DataApiBuilder 0.7.5+2f938a6dfe4b90c26d9e24352b38280cb6e66f15

Azure SQL

Local (including CLI)

REST

### Relevant log output

```Text
Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 POST http://localhost:5000/api/bus-near-position application/json 66
      Request finished HTTP/1.1 POST http://localhost:5000/api/bus-near-position application/json 66 - 307 0 - 3.7979ms
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 POST https://localhost:5001/api/bus-near-position application/json 66
dbug: Azure.DataApiBuilder.Service.AuthenticationHelpers.ClientRoleHeaderAuthenticationMiddleware[0]
      3a72d9e5-e0f7-40d8-b5fe-a9af88d05387: Request authentication state: Anonymous.
dbug: Azure.DataApiBuilder.Service.AuthenticationHelpers.ClientRoleHeaderAuthenticationMiddleware[0]
      3a72d9e5-e0f7-40d8-b5fe-a9af88d05387: The request will be executed in the context of Anonymous role
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'Azure.DataApiBuilder.Service.Controllers.RestController.Insert (Azure.DataApiBuilder.Service)'
info: Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker[3]
      Route matched with {action = "Insert", controller = "Rest"}. Executing controller action with signature System.Threading.Tasks.Task`1[Microsoft.AspNetCore.Mvc.IActionResult] Insert(System.String) on controller Azure.DataApiBuilder.Service.Controllers.RestController (Azure.DataApiBuilder.Service).
fail: Azure.DataApiBuilder.Service.Controllers.RestController[0]
      3a72d9e5-e0f7-40d8-b5fe-a9af88d05387: This operation is not supported.
fail: Azure.DataApiBuilder.Service.Controllers.RestController[0]
      3a72d9e5-e0f7-40d8-b5fe-a9af88d05387:    at Azure.DataApiBuilder.Service.Services.RestService.ExecuteAsync(String entityName, Operation operationType, String primaryKeyRoute)
         at Azure.DataApiBuilder.Service.Controllers.RestController.HandleOperation(String route, Operation operationType)
info: Microsoft.AspNetCore.Mvc.Infrastructure.SystemTextJsonResultExecutor[1]
      Executing JsonResult, writing value of type '<>f__AnonymousType0`1[[<>f__AnonymousType1`3[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Int32, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], Azure.DataApiBuilder.Service, Version=0.7.5.0, Culture=neutral, PublicKeyToken=null]]'.
info: Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker[2]
      Executed action Azure.DataApiBuilder.Service.Controllers.RestController.Insert (Azure.DataApiBuilder.Service) in 42.8858ms
      Executed endpoint 'Azure.DataApiBuilder.Service.Controllers.RestController.Insert (Azure.DataApiBuilder.Service)'
      Request finished HTTP/1.1 POST https://localhost:5001/api/bus-near-position application/json 66 - 405 - application/json;+charset=utf-8 82.6701ms
```

### Code of Conduct

severussundar:
## Why make this change?

- Closes #1504 
  - For stored procedure entities, when the `methods` field of the `rest` section is absent in the config file, POST requests fail.
> Sample config:

```json
   "rest": {           
        "path": "bus-near-position"        
   }
```
  - Unless otherwise configured, by default, REST POST requests should be enabled for stored procedures. 
  
## What is this change?

- `EntityRestOptions`: `DEFAULT_SUPPORTED_VERBS_FOR_SP` is added to represent the default HTTP methods supported for stored procedures 
- `EntityRestOptionsConverter`: Logic is modified to always populate the REST Methods as an empty array when absent in the config file. This is because, at this point, the converter cannot be aware of the entity type and hence, cannot populate different default values for tables/views vs stored procedures. The right values are populated later during the deserialization process.
-  `RuntimeEntities`:  `ProcessRestDefaults` method is modified to populate the right set of supported HTTP methods based on the entity type. 
- `TestSPRestDefaultsForManuallyConstructedConfigs`: Integration tests for validating the REST API behavior for stored procedures for various scenarios are added.
 
## How was this tested?

- [x] Integration Tests

**Manual Tests:**
- [x] REST section is completely absent in the config file. 
    > POST requests should be supported on the default endpoint (/api/{entity's name})
- [x]  REST section contains only `enabled` field: True 
    > POST requests should be supported on the default endpoint (/api/{entity's name})
- [x] REST section contains only `enabled` field : False
    > REST APIs should be disabled for this entity.
- [x] REST section contains only `path` element
    > POST requests should be supported on the specified endpoint (/api/{path})
- [x] REST section contains only `methods` element
    > Specified list of HTTP actions should be supported on the default endpoint (/api/{entity's name})
- [x] REST section contains both `methods` and `path` element
    > Specified list of HTTP actions should be supported on the specified endpoint (/api/{path})


src/Config/ObjectModel/RuntimeEntities.cs
```diff
@@ -121,9 +121,28 @@ private static Entity ProcessRestDefaults(Entity nameCorrectedEntity)
...
+                        with
+                                    { Rest = new EntityRestOptions(EntityRestOptions.DEFAULT_SUPPORTED_VERBS_FOR_SP) }
+                    :
+                    nameCorrectedEntity
+                        with
+                    { Rest = new EntityRestOptions(EntityRestOptions.DEFAULT_SUPPORTED_VERBS) };
+        }
+        else
+        {
+            if (nameCorrectedEntity.Rest.Methods.Length == 0)
```
seantleonard, line 135: Add a comment that calls out that we only care about this extra check when length == 0 because the `EntityRestOptionsConverter` properly parses any user specified methods.

src/Cli.Tests/Snapshots/UpdateEntityTests.TestUpdateRestAndGraphQLSettingsForStoredProcedures_c9b12fe27be53878.verified.txt
```diff
@@ -40,6 +40,9 @@
           Operation: Mutation
         },
         Rest: {
+          Methods: [
```
severussundar, line 43: When `Enabled` property is set to `false`, the REST APIs are disabled for the entity (tested this manually). However, when writing to the config json, the `methods` and `path` fields are written. This is the same case for `graphql` field too. 

Clean-up of this can be taken up in a subsequent PR

abhishekkumams (approved):
Thanks for fixing it. just one nit.

src/Config/ObjectModel/RuntimeEntities.cs
```diff
@@ -121,9 +121,28 @@ private static Entity ProcessRestDefaults(Entity nameCorrectedEntity)
         // If no Rest node was provided in the config, set it with the default state of enabled for all verbs
         if (nameCorrectedEntity.Rest is null)
         {
-            nameCorrectedEntity = nameCorrectedEntity
-                with
-            { Rest = new EntityRestOptions(EntityRestOptions.DEFAULT_SUPPORTED_VERBS) };
+            nameCorrectedEntity = (nameCorrectedEntity.Source.Type is EntitySourceType.StoredProcedure) ?
```
abhishekkumams, line 124: we can simplify this logic for better readability.
```
    if (nameCorrectedEntity.Rest is null)
    {
        var defaultMethods = (nameCorrectedEntity.Source.Type is EntitySourceType.StoredProcedure) ?
            EntityRestOptions.DEFAULT_SUPPORTED_VERBS_FOR_SP :
            EntityRestOptions.DEFAULT_SUPPORTED_VERBS;

        nameCorrectedEntity = nameCorrectedEntity with { Rest = new EntityRestOptions(defaultMethods) };
    }
    else if (nameCorrectedEntity.Rest.Methods.Length == 0)
    {
        var defaultMethods = (nameCorrectedEntity.Source.Type is EntitySourceType.StoredProcedure) ?
            EntityRestOptions.DEFAULT_SUPPORTED_VERBS_FOR_SP :
            EntityRestOptions.DEFAULT_SUPPORTED_VERBS;

        nameCorrectedEntity = nameCorrectedEntity with
        {
            Rest = new EntityRestOptions(
                Methods: defaultMethods,
                Path: nameCorrectedEntity.Rest.Path,
                Enabled: nameCorrectedEntity.Rest.Enabled
            )
        };
    }
    ```

Comment (severussundar) [reply src/Config/ObjectModel/RuntimeEntities.cs:124]: When `nameCorrectedEntity.Rest.Methods.Length == 0` is true, we want to populate the rest methods only for stored procedures (for tables/views, when rest is enabled, we support all 5 request types).

But, I see your point regarding readability, updated the logic.

File: src/Config/Converters/EntityRestOptionsConverter.cs [hunk 4]
```diff
@@ -14,7 +14,7 @@ internal class EntityRestOptionsConverter : JsonConverter<EntityRestOptions>
     {
         if (reader.TokenType == JsonTokenType.StartObject)
         {
-            EntityRestOptions restOptions = new(Methods: EntityRestOptions.DEFAULT_SUPPORTED_VERBS, Path: null, Enabled: true);
+            EntityRestOptions restOptions = new(Methods: Array.Empty<SupportedHttpVerb>(), Path: null, Enabled: true);
```
ayush3797, line 17: Per specs, the `Methods` property should be nullable. We might need to take this into consideration maybe in a subsequent PR.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -89,6 +89,395 @@ public class ConfigurationTests
...
+              }
+        }";
+
+        /// <summary>
+        /// A config file with a SP entity for which REST APIs are disabled.
+        /// This config string is used for validating that none of the REST methods are enabled.
+        /// </summary>
+        public const string SP_CONFIG_WITH_REST_DISABLED = @"
+            {
+              ""$schema"": ""https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json"",
```
Aniruddh25, line 331: Hard coding different variations of the config file for the various test cases doesn't seem scalable. 
Can we do better? Only parts of the config are varying, can we keep a common const string, then build new configs for the parts that are varying?

Aniruddh25 (changes requested):
Thank you for the extensive manual testing, looking forward to more optimal way to have automated test.

src/Config/Converters/EntityRestOptionsConverter.cs
aaronpowell: The non-nullability is probably more a result of my misunderstanding of what the role of `Methods` was than anything else.

It can probably be made nullable, it'd then just be a matter of finding the impact downstream in the codebase and having the appropriate `null` checks in place.

severussundar: I've created [this issue](https://github.com/Azure/data-api-builder/issues/1571) to track this refactoring.

src/Service.Tests/TestHelper.cs
```diff
@@ -114,7 +115,10 @@ public static RuntimeConfig AddMissingEntitiesToConfig(RuntimeConfig config, str
         public const string SAMPLE_SCHEMA_DATA_SOURCE = SCHEMA_PROPERTY + "," + @"
             ""data-source"": {
               ""database-type"": ""mssql"",
-              ""connection-string"": ""testconnectionstring""
+              ""connection-string"": ""testconnectionstring"",
+              ""options"": {
```
severussundar, line 119: CLI, explicitly writes out true/false depending on whether session-context is enabled or not. So, adding this

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: Updated to generate the configs dynamically

src/Config/ObjectModel/RuntimeEntities.cs
```diff
@@ -121,9 +121,21 @@ private static Entity ProcessRestDefaults(Entity nameCorrectedEntity)
...
+                        with
+                                    { Rest = new EntityRestOptions(EntityRestOptions.DEFAULT_SUPPORTED_VERBS_FOR_SP) }
+                    :
+                    nameCorrectedEntity
+                        with
+                    { Rest = new EntityRestOptions(EntityRestOptions.DEFAULT_SUPPORTED_VERBS) };
+        }
+        else if (nameCorrectedEntity.Source.Type is EntitySourceType.StoredProcedure && nameCorrectedEntity.Rest.Methods.Length == 0)
+        {
+            // REST Method field is relevant only for stored procedures. For entities backed by tables and views, all the methods are supported.
```
seantleonard, line 135: It would probably be more accurate to say that tables and views support all HTTP verbs by design (unless permissions config determines otherwise). For stored procedures, we still support all HTTP verbs but POST is assigned by default unless otherwise specified.

seantleonard: essentially the use of the word 'support' in the comment can be updated to a word that better reflects the intended design.

src/Config/ObjectModel/EntityRestOptions.cs
```diff
@@ -19,4 +19,5 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 public record EntityRestOptions(SupportedHttpVerb[] Methods, string? Path = null, bool Enabled = true)
 {
     public static readonly SupportedHttpVerb[] DEFAULT_SUPPORTED_VERBS = new[] { SupportedHttpVerb.Get, SupportedHttpVerb.Post, SupportedHttpVerb.Put, SupportedHttpVerb.Patch, SupportedHttpVerb.Delete };
+    public static readonly SupportedHttpVerb[] DEFAULT_SUPPORTED_VERBS_FOR_SP = new[] { SupportedHttpVerb.Post };
```
seantleonard, line 22: nit: do we want to have this named "supported"?

seantleonard (commented):
Few questions. Would be worth identifying path forward based on Ani's comment: not using hard coded string config file values. Now that we have the new config system, you can create configs programmatically and in your case, only change the entity under test.

src/Config/ObjectModel/EntityRestOptions.cs
severussundar: updated

src/Config/ObjectModel/RuntimeEntities.cs
severussundar: updated the comment

severussundar:
> Few questions. Would be worth identifying path forward based on Ani's comment: not using hard coded string config file values. Now that we have the new config system, you can create configs programmatically and in your case, only change the entity under test.

I considered using `InitMinimalRuntimeConfig` helper method in `ConfigurationTests` for generating the config files. There are few reasons why I decided not to use that and chose to go with hard coded config files. 

a)  This bug was identified by constructing the config file by hand and not through CLI.
b)  This PR modifies the deserialization logic for the rest element in the config file. A direct and cleaner way to test this would be to pass the various hard coded strings that are possible and see if the deserialization works as expected. This ensures that the serialization logic does not interfere with the way we would like our test config files to be (programmatically it is not possible to generate all the cases that are possible when editing the config by hand).
For example: The scenario of not having `methods` in the config file at all. 
Even if an Entity object (backed by a SP) is created by populating the rest methods as an empty string, finally, in the config file, it gets written out as an array with a POST element.

Considering the way this particular bug was identified; to have an exact representation of that scenario in the tests - hard coding the config file would be the way to go IMHO.

There was scope to remove some duplication by re-using the common parts of the hard coded string across all the scenarios - I have incorporated this.

Aniruddh25 (approved):
LGTM, thanks for removing as much duplication as possible, yet preserving to test the bug scenario.

PR #1549, opened 2023-06-26
Cleanup Test Project Configuration

### LINKED ISSUE #1547: Cleanup sample config files for REST on entities to NOT support Boolean path
In #1402 support was added for the REST config of an entity to be:

```json
"MyEntity": {
  "rest": {
    "path": true
  }
}
```

This wasn't previously supported so the schema file needs to be updated to support it.

## Context

Reopening as I encountered that my local config, and config files present in the test project have `"path": true` used, which results in an engine startup failure.

Schema file is accurate to only support 'string' as you two mention: https://github.com/Azure/data-api-builder/blob/90229412a221cb0caaefd450fca57debd4bc9bf9/schemas/dab.draft.schema.json#L286C15-L288

I'll submit a PR to fix the files in the test project (so we don't increase file count here). 

I confirming whether the CLI tool properly adheres to the schema requirement so we don't have to tag this as a breaking change.

_Originally posted by @seantleonard in https://github.com/Azure/data-api-builder/pull/1402#discussion_r1238990524_

seantleonard:
## Why make this change?

- Closes #1547 which highlights incorrect value used in the runtime config for the test project. An entity's rest path value can't be a boolean as set in `dab-config.MsSql.json` where `"path": true`. 

Schema file is accurate to only support 'string': https://github.com/Azure/data-api-builder/blob/90229412a221cb0caaefd450fca57debd4bc9bf9/schemas/dab.draft.schema.json#L286C15-L288

DAB CLI does not create config files with the `path` property set to a boolean.

## What is this change?

- Removes instance of `"path": true`.

## How was this tested?
Pass existing tests. Setting `"path": true` in runtime config will result in a startup/config parsing error after #1402 is merged.

- [x] Integration Tests
- [x] Unit Tests
- [x] Validate CLI does not create invalid configs for both `dab add` and `dab update`

### CLI Test 1
```bash
.\Microsoft.DataApiBuilder update MyEntity --rest false
```
Output: `"rest": false`

### CLI Test 2
```bash
.\Microsoft.DataApiBuilder update MyEntity --rest pathvalue
```
Output: 
```json
      "rest": {
        "path": "/commodities"
      }
```

### CLI Test 3
```bash
.\Microsoft.DataApiBuilder update MyEntity --rest "basicPath"
```
Output:
```json
      "rest": {
        "path": "/basicPath"
      }
```

### CLI Test 4
```bash
.\Microsoft.DataApiBuilder update MyEntity --rest True
```
Output: `"rest": true`


aaronpowell:
Was the file generated from the `config-generator` or was it hand-edited? The latter approach can be inconsistent with CI which generates those files.

abhishekkumams (approved):
thanks for the cleanup!

Aniruddh25 (approved):

Aniruddh25 (approved):

seantleonard:
These changes no longer apply as #1402 and #1529 made the relavant changes.

PR #1551, opened 2023-06-27
OpenAPI - Distinguish between stored procedure parameters and result set columns

### LINKED ISSUE #1509: [Bug]: Inconsistent Example Schema for Request Body of Stored-Procedure in Swagger UI
### What happened?

Example schema for request body shown for stored-procedure is not consistent.

To reproduce:
1. Use the dab-config.MsSql.json to start the engine.
2. open the swagger UI through the endpoint : https://localhost:5001/swagger

Inconsistencies noted:
1. Empty schema for request Body shown for **/InsertBook** (missing parameter info)
image

2. Incorrect request body schema shown for **/CountBook**,  (should be empty json, as it doesn’t need params)
image
 
3. Incorrect request body schema shown for **/UpdateBookTitle**,  (publisher_id is not required in the request payload.)
image
Similar issue with **/GetAuthorsHistoryByFirstName** and **/InsertAndDisplayAllBooksUnderGivenPublisher**

### Version

0.7.5

Azure SQL

Local (including CLI)

REST

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
## Why make this change?

- Closes #1509

## What is this change?

- The OpenApiDocumentor service now correctly distinguishes between input parameters and result set columns when documenting a stored procedures REST endpoint. Previously, only the result set columns were resolved. As a result, the OpenApi document that is generated now shows input parameters as fields to be used in the request bodies of POST, PUT, and PATCH.  (GET and DELETE have no request bodies).

## How was this tested?

- [x] Integration Tests - Added tests with additional stored procedures in our sql deployment script which demonstrates how parameters and output result set columns (and their JSON data types) are resolved and included in the open api doc.

## OpenAPI document example 
An example OpenAPI document which helps visualize how the parsing works. Irrelevant fields were removed.
```json
{
    "openapi": "3.0.1",
    "paths": {
        "/sp1": {
            "get": {"..."},
            "post": {
                "tags": [
                    "sp1"
                ],
                "description": "Executes a stored procedure.",
                "requestBody": {
                    "content": {
                        "application/json": {
                            "schema": {
                                "$ref": "#/components/schemas/sp1_sp_request"
                            }
                        }
                    },
                    "required": true
                },
                "responses": {
                    "201": {
                        "description": "Created",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "properties": {
                                        "value": {
                                            "type": "array",
                                            "items": {
                                                "$ref": "#/components/schemas/sp1_sp_response"
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    },
                    "400": {
                        "description": "BadRequest"
                    }
                }
            }
		}
    },
    "components": {
        "schemas": {
            "sp1_sp_request": {
                "type": "object",
                "properties": {
                    "inputName": {
                        "type": "string",
                        "format": ""
                    }
                }
            },
            "sp1_sp_response": {
                "type": "object",
                "properties": {
                    "outputName": {
                        "type": "string",
                        "format": ""
                    }
                }
            }
        }
    }
}
```


src/Service.Tests/OpenApiDocumentor/StoredProcedureGeneration.cs
```diff
@@ -0,0 +1,346 @@
...
+                // auto-resolves the reference to a concrete OpenApiSchema object. However, to avoid testing OpenAPI.NET functionality,
+                // this test looks directly at the OpenAPI document's "components" property to validate presence
+                // and composition of the generated schema component.
+                Assert.IsTrue(_openApiDocument.Components.Schemas.ContainsKey($"{entityName}{OpenApiDocumentor.SP_REQUEST_SUFFIX}"), message: "Unexpected absence of schema component definition.");
+                Dictionary<string, OpenApiSchema> schemaComponentProperties = new(GetSchemaComponentProperties($"{entityName}{OpenApiDocumentor.SP_REQUEST_SUFFIX}"));
+
+                // Validate that the generated properties do not outnumber the count of expected columns.
+                Assert.AreEqual(expectedParameters.Length, schemaComponentProperties.Count, message: "The number of generated parameters is not expected.");
+
+                // Validate column presence and accurate column JSON type.
```
Aniruddh25, line 179: was column here intended to be a parameter?

src/Service.Tests/OpenApiDocumentor/StoredProcedureGeneration.cs
```diff
@@ -0,0 +1,346 @@
...
+
+            // Validate the correct schema component is referenced in the response body.
+            // The OpenApiResponses dictionary key represents the integer value of the HttpStatusCode,
+            // which is returned when using Enum.ToString("D").
+            // The "D" format specified "displays the enumeration entry as an integer value in the shortest representation possible."
+            OpenApiReference reference = GetResponseBodyReference(responses[HttpStatusCode.OK.ToString("D")]);
+            Assert.AreEqual(expected: $"{entityName}{OpenApiDocumentor.SP_RESPONSE_SUFFIX}", actual: reference.Id, message: SCHEMA_REF_ID_ERROR);
+            Assert.AreEqual(expected: $"{SCHEMA_REF_PREFIX}{entityName}{OpenApiDocumentor.SP_RESPONSE_SUFFIX}", actual: reference.ReferenceV3);
+
+            // It is possible to get the schema component from a OpenApiResponse object because OpenAPI.NET functionality
```
Aniruddh25, line 217:
```suggestion
            // It is possible to get the schema component from an OpenApiResponse object because OpenAPI.NET functionality
```

src/Service.Tests/OpenApiDocumentor/StoredProcedureGeneration.cs
```diff
@@ -0,0 +1,346 @@
...
+            // can also validate Reference.Id 'sp1_sp_reponse' to ensure correct mapping.
+            // Though, the fact that Properties is populated correctly, means the OpenApi.Net mechanisms
+            // populated Properties with the contents of the referenced schema component.
+            OpenApiResponses responses = GetOperationResponses(entityName, OperationType.Get);
+
+            // Validate the correct schema component is referenced in the response body.
+            // The OpenApiResponses dictionary key represents the integer value of the HttpStatusCode,
+            // which is returned when using Enum.ToString("D").
+            // The "D" format specified "displays the enumeration entry as an integer value in the shortest representation possible."
+            OpenApiReference reference = GetResponseBodyReference(responses[HttpStatusCode.OK.ToString("D")]);
```
Aniruddh25, line 213: Once the OpenApiReference is obtained, looks like both the previous and this test perform same kind of assertions. Can we move the assertions into a helper?

Aniruddh25 (approved):
Looks good! Left few minor suggestions.

src/Service.Tests/OpenApiDocumentor/StoredProcedureGeneration.cs
seantleonard: yep, for that section of code. Updating.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -831,7 +864,45 @@ private static OpenApiMediaType CreateResponseContainer(string responseObjectSch
...
+            Dictionary<string, OpenApiSchema> properties = new();
+
+            foreach (string parameter in fields.Keys)
+            {
+                string typeMetadata = TypeHelper.SystemTypeToJsonDataType(fields[parameter].SystemType).ToString().ToLower();
+
+                properties.Add(parameter, new OpenApiSchema()
+                {
+                    Type = typeMetadata,
+                    Format = string.Empty
```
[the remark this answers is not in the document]
seantleonard: Currently not using/populating this property. It is optional per open api spec.I'll actually remove this from being expliticly set to string.empty. https://spec.openapis.org/oas/v3.0.3#:~:text=Primitives%20have%20an%20optional%20modifier%20property%3A%20format.
> Primitives have an optional modifier property: format.

abhishekkumams:
you mentioned in PR description that `GET  and DELETE won't have request bodies`, but they can have input params, right? So, why not request bodies?

src/Service.Tests/OpenApiDocumentor/StoredProcedureGeneration.cs
seantleonard: Moved into a common helper. Renamed error messages and helper variables from column/parameters to "Properties" to reflect the changes. No functional changes though. Result is reduced duplicate code, slightly more generic error messages. There are distinct tests for validating request body versus response body so any failures in either will directly correlate to whether a validation issue occurred for parameters versus columns.

seantleonard:
> you mentioned in PR description that `GET and DELETE won't have request bodies`, but they can have input params, right? So, why not request bodies?

Request bodies in those two request types are not convention per HTTP spec:
1. GET https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.1-6
2. DELETE  https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.5-6

Relevant text with my annotation in square brackets:
>  A client SHOULD NOT generate content in a [GET/DELETE] request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported.

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -530,3 +532,10 @@ EXEC('CREATE FUNCTION dbo.revenuesPredicate(@accessible_role varchar(20))
 -- Adding a security policy which would restrict access to the rows in revenues table for
 -- SELECT,UPDATE,DELETE operations using the filter predicate dbo.revenuesPredicate.
 EXEC('CREATE SECURITY POLICY dbo.revenuesSecPolicy ADD FILTER PREDICATE dbo.revenuesPredicate(accessible_role) ON dbo.revenues;');
+
+-- SP with one input parameter with different name than the column name present in the first result set.
```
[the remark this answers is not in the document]
seantleonard: I ended up using one of the stored procedures already present and removed the two i added as that SP accomplishes what i wanted to test.

src/Service.Tests/OpenApiDocumentor/StoredProcedureGeneration.cs
```diff
@@ -36,9 +36,10 @@ public class StoredProcedureGeneration
         private const string SCHEMA_REF_ID_ERROR = "Unexpected schema reference id.";
         private static OpenApiDocument _openApiDocument;
         private static RuntimeEntities _runtimeEntities;
+
         /// <summary>
         /// Bootstraps test server once using a single runtime config file so
-        /// each test need not boot the entire server to generate an openapi description doc.
+        /// each test need not boot the entire server to generate an  description doc.
```
Aniruddh25, line 42:
```suggestion
        /// each test need not boot the entire server to generate an description doc.
```

Aniruddh25 (approved):

src/Directory.Packages.props
```diff
@@ -43,4 +44,4 @@
     <PackageVersion Include="Verify.MsTest" Version="20.1.0" />
     <PackageVersion Include="Verify.DiffPlex" Version="2.2.1" />
   </ItemGroup>
-</Project>
+</Project>
```
severussundar, line 47: nit: EOL needed

severussundar (approved):
LGTM

src/Directory.Packages.props
seantleonard: Is that required? I'm use any modifications added by `dotnet format` prior to pushing commits.

severussundar: Github is flagging a warning, hence, I added that comment.

ayush3797:
Nit: The comment here: https://github.com/Azure/data-api-builder/blob/73716e70134ed6ab09fb482746d135019528c202/src/Core/Services/OpenAPI/OpenApiDocumentor.cs#L169 is repeated at line 178.

seantleonard:
> It might not be relevant for this PR but I think we default stored procedures to atleast support POST operation if no method is defined. This doesn't seem to be the case here:
> 
> https://github.com/Azure/data-api-builder/blob/73716e70134ed6ab09fb482746d135019528c202/src/Core/Services/OpenAPI/OpenApiDocumentor.cs#L456

Good catch. The sp default methods are added (POST) when `RuntimeEntities` is created and `ProcessRestDefaults()` is executed. This also looks to be artifacts of old config system where the collection may be null. The nullable annotation and null check are not necessary.

src/Directory.Packages.props
seantleonard: fixed

PR #1558, opened 2023-06-29
Update pipelines to support `-rc` for dab versions

### LINKED ISSUE #1545: update pipelines to accommodate `-rc`
we need to make changes so that our pipeline knows when we are publishing a release candidate vs a stable version.

abhishekkumams:
## Why make this change?

- Closes #1545
- This will allow us to make pre-releases, with dab version containing `-rc` suffix for release candidates.

## What is this change?

- updating bash script in the pipeline that sets the `dabVersion` to add `-rc` suffix in case  it's not a stable release, to identify as a release candidate.

## How was this tested?

- By running pipelines and checking pipeline logs.


.pipelines/build-pipelines.yml
```diff
@@ -38,15 +38,25 @@ steps:
 # We cannot set this in variables section above because $(isNugetRelease) 
 # is not available at pipeline compilation time.
 - bash: |
-    echo ISNUGETRELEASE = $ISNUGETRELEASE
+    echo IsNugetRelease = $ISNUGETRELEASE
+    echo IsReleaseCandidate = $ISRELEASECANDIDATE
```
Aniruddh25, line 42: will these be 2 new variables to the pipeline - ISRELEASECANDIDATE
RELEASE_CANDIDATE_NUMBER?

.pipelines/build-pipelines.yml
```diff
@@ -38,15 +38,25 @@ steps:
...
+    echo RELEASE_CANDIDATE_NUMBER = $RELEASE_CANDIDATE_NUMBER
```
Aniruddh25, line 43: Any way we can autoassign RELEASE_CANDIDATE_NUMBER?

.pipelines/build-pipelines.yml
abhishekkumams: Yes

Aniruddh25 (approved):
LGTM, Thanks for these changes!

.pipelines/build-pipelines.yml
abhishekkumams: As of now, I couldn't find any direct method to do so. Although there are other complex approaches that requires parsing etc. I'll create an issue and implement it if I find a better approach.

.pipelines/build-pipelines.yml
abhishekkumams: update to add only 1 new variable.

.pipelines/build-pipelines.yml
```diff
@@ -39,15 +39,23 @@ steps:
...
 # is not available at pipeline compilation time.
 - bash: |
-    echo ISNUGETRELEASE = $ISNUGETRELEASE
+    echo IsNugetRelease = $ISNUGETRELEASE
+    echo IsReleaseCandidate = $ISRELEASECANDIDATE
+    dabVersion=$(major).$(minor).$(patch)
     if [ "$ISNUGETRELEASE" = "true" ]
     then
-      dabVersion=$(major).$(minor).$(patch)
+      if [ "$ISRELEASECANDIDATE" = "true" ]
```
severussundar, line 47: nit: Re-naming $ISRELEASECANDIDATE to something like `isPreRelease` would make it easier to follow (intuitively distinguishes pre-release vs release)

abhishekkumams: I did it specifically because we are calling it release candiate version, with `-rc` tag. Also, when I looked for differences between the two, I found out the main differentiating factor is that the `preRelease` version may contain bugs and may not have all feature implemented. It refers to the version still under development.
While, the `releaseCandidate` version refers to a version of software that is considered to be near completion and ready for release. It has undergone rigorous testing and is expected to be stable, but may still undergo some final adjustments before the official release.

So, `releaseCandidate` seems like a suitable choice here.

Issue #1546, opened 2023-06-26
Refactor tests to have shared setup logic


aaronpowell:
With the changes in #1402 there is repeated code for how we setup the mock file system environment across the CLI tests.

We should refactor the repeated code into a helper method so that we are more consistent in how we setup the tests.

## Context

Yep, I'll make a separate task for that as I think there's a bit of "general cleanup" we can do once this is merged to have some more shared code and things like that, but it's more "nice to have" than "must have".

_Originally posted by @aaronpowell in https://github.com/Azure/data-api-builder/pull/1402#discussion_r1239136401_

PR #1564, opened 2023-07-03
Tidy up of test logic that can be shared.

aaronpowell:
This is addressing #1546 by cleaning up some of the CLI tests that had shared setup logic.

The setup logic has been pulled into dedicated supporting classes that can be used within the tests.

I deliberately didn't use the existing `TestHelper.cs` class to avoid it becoming a dumping ground with no clear indicator of what tests use what.


src/Cli.Tests/TestLoggerSupport.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Cli.Tests;
+
+internal static class TestLoggerSupport
+{
+    public static ILoggerFactory ProvisionLoggerFactory() =>
```
abhishekkumams, line 8: Maybe we could move it to the TestHelper class in Cli.Tests ?

aaronpowell: I noted in the PR description that it was a deliberate decision to not put it in that class. When doing the changes for #1402 I had many conflicts on these "helper" classes as they can be iterated on frequently. This isn't just isolated to test projects either.

So as a result I'll always try to introduce a new class which is specific for the "feature", even if it only has a single method in it, as it become clearer where code will exist when scanning through a repo.

abhishekkumams (approved):

src/Cli.Tests/FileSystemUtils.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public static MockFileSystem ProvisionMockFileSystem()
+    {
+        MockFileSystem fileSystem = new();
+
+        fileSystem.AddFile(
+            fileSystem.Path.Combine(
+                fileSystem.Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) ?? "",
+                "dab.draft.schema.json"),
+            new MockFileData("{ \"additionalProperties\": {\"version\": \"https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json\"} }"));
```
seantleonard, line 19: Why is this needed? I see it was there before, would be nice to have a comment why we add the additional property here

Aniruddh25 (approved):

src/Cli.Tests/FileSystemUtils.cs
aaronpowell: I knew there was a reason but I didn't remember what that reason was, so I deleted it and ran the tests!

The reason is that this method gets called: https://github.com/Azure/data-api-builder/blob/main/src/Config/RuntimeConfigLoader.cs#L255

If we haven't pushed the file into the mock file system, it'll get a `FileNotFoundException` which traces back to https://github.com/Azure/data-api-builder/blob/main/src/Cli/ConfigGenerator.cs#L146

It use to not be a problem as we were doing disk IO but now that we aren't (it's a mock disk) we need to ensure that files we expect to be on disk _are_ on disk.

I'll add a comment to it.

PR #1442, opened 2023-04-18
Introducing parameters' DbType before passing on to the database - MsSql

ayush3797:
## Why make this change?
Fixes https://github.com/Azure/data-api-builder/issues/475
Most column types will accept a string null and the db will automatically infer our intention, but for varbinary in MsSql the db complains that it cannot convert varchar to varbinary. For more details about the issue refer: https://stackoverflow.com/questions/29254690/why-does-dbnull-value-require-a-proper-sqldbtype

## What is this change?
Changing how we form the `parameters` as we pass them to the database. Previously we used to only pass the parameter's value. Now, we would also pass the parameter's `DbType` along with the value as suggested in the stack overflow article. 

**Notes:**
1. We would only be passing the DbType for MsSql.
2. Usually, we haven't seen any need to pass the `DbType` for Sql Server data types other than `varbinary`. However, we will still be supplying the DbType for all other parameter's for which we are confirmed about the 1:1 mapping from SystemType to DbType. Sql server datatypes like `datetime`, `datetime2`, `datetimeoffset` are all treated as `datetime` by graphql, so there is no 1:1 mapping for datetime datatypes. So we are skipping supplying the DbType for these.
3. As explained in this stack overflow article [here](https://stackoverflow.com/questions/6868300/net-guid-uppercase-string-format), the GUID when converted to string, is output in lowercase. That's why when the query to verify the MsSql test `InsertOneInSupportedTypes` is written, we have explicitly cast `guid_types` to lower case.

## Additional change
1. The `real` sql server datatype was incorrectly mapped to `long` dotnet type. It has been correctly mapped to `decimal` now.

## Issue with DateTime/DateTimeOffset Sql Server data types
For a mutation request like this, we get a response:
image

and the parameters are generated as:
image

the parameter generated is of type `DateTimeOffset` (as evident from the value highlighted) but the underlying Sql Server datatype for the `datetime_types` field is `datetime`. Even though we supplied a value `1999-01-08 09:20:00` in the mutation, it became `08-01-1999 09:20:00 +05:30`, because that's how graphql treats datetime/datetimoffset/date types. Then this implicit conversion from a value of type DateTimeOffset to DateTime cannot happen throwing an exception with message : `Failed to convert parameter value from a DateTimeOffset to a DateTime.` Hence we cannot add a mapping from `typeof(DateTime) = DbType.DateTime`.

[...]

PR #1568, opened 2023-07-06
OpenAPI - Improved resolving of database types to json data types

### LINKED ISSUE #1508: [Bug]: Undefined Types shown in Example Schema for request body in Swagger
### What happened?

Example schema shown in the request body is very useful to know the parameters and it’s types, but for this particular case I see it's not useful and confusing for some composite dataTypes such as datetime.
I am not sure if I should enter a string or some other type.
image

This is the underlying entity in the database:
```
CREATE TABLE type_table(
    byte_types tinyint,
    short_types smallint,
    int_types int,
    long_types bigint,
    string_types varchar(max),
    single_types real,
    float_types float,
    decimal_types decimal(38, 19),
    boolean_types bit,
    date_types date,
    datetime_types datetime,
    datetime2_types datetime2,
    datetimeoffset_types datetimeoffset,
    smalldatetime_types smalldatetime,
    bytearray_types varbinary(max),
    guid_types uniqueidentifier DEFAULT newid()
);
```

### Version

0.7.5

Azure SQL

Local (including CLI)

REST

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
## Why make this change?

- Closes #1508 

## What is this change?

- improves the mapping between SqlDbType <=> CLR/.NET Framwork (System) types <=> JsonDataType so that OpenApi document generation resolves the appropriate value types in REST endpoint result field value types and input (request body) field value types.

- To view OpenApi document output, browse to the endpoint /swagger
### Before (Swagger view with default values based on resolved JSON data type)
```json
{
  "value": [
    {
      "typeid": 0,
      "byte_types": "string",
      "short_types": 0,
      "int_types": 0,
      "long_types": 0,
      "string_types": "string",
      "single_types": 0,
      "float_types": 0,
      "decimal_types": 0,
      "boolean_types": true,
      "date_types": "Unknown Type: undefined",
      "datetime_types": "Unknown Type: undefined",
      "datetime2_types": "Unknown Type: undefined",
      "datetimeoffset_types": "Unknown Type: undefined",
      "smalldatetime_types": "Unknown Type: undefined",
      "bytearray_types": "string",
      "guid_types": "string"
    }
  ]
}
```
### After (Swagger view with default values based on resolved JSON data type)
```json
{
  "value": [
    {
      "typeid": 0,
      "byte_types": "string",
      "short_types": 0,
      "int_types": 0,
      "long_types": 0,
      "string_types": "string",
      "single_types": 0,
      "float_types": 0,
      "decimal_types": 0,
      "boolean_types": true,
      "date_types": "string",
      "datetime_types": "string",
      "datetime2_types": "string",
      "datetimeoffset_types": "string",
      "smalldatetime_types": "string",
      "bytearray_types": "string",
      "guid_types": "string"
    }
  ]
}
```

## After: OpenApi document json
```json
"components": {
        "schemas": {
            "SupportedType": {
                "type": "object",
                "properties": {
                    "typeid": {
                        "type": "number",
                        "format": ""
                    },
                    "byte_types": {
                        "type": "string",
                        "format": ""
                    },
                    "short_types": {
                        "type": "number",
                        "format": ""
                    },
                    "int_types": {
                        "type": "number",
                        "format": ""
                    },
                    "long_types": {
                        "type": "number",
                        "format": ""
                    },
                    "string_types": {
                        "type": "string",
                        "format": ""
                    },
                    "single_types": {
                        "type": "number",
                        "format": ""
                    },
                    "float_types": {
                        "type": "number",
                        "format": ""
                    },
                    "decimal_types": {
                        "type": "number",
                        "format": ""
                    },
                    "boolean_types": {
                        "type": "boolean",
                        "format": ""
                    },
                    "date_types": {
                        "type": "string",
                        "format": ""
                    },
                    "datetime_types": {
                        "type": "string",
                        "format": ""
                    },
                    "datetime2_types": {
                        "type": "string",
                        "format": ""
                    },
                    "datetimeoffset_types": {
                        "type": "string",
                        "format": ""
                    },
                    "smalldatetime_types": {
                        "type": "string",
                        "format": ""
                    },
                    "bytearray_types": {
                        "type": "string",
                        "format": ""
                    },
                    "guid_types": {
                        "type": "string",
                        "format": ""
                    }
                }
            }
        }
    }
```
## How was this tested?

- [X] Unit Tests


src/Core/Models/SqlTypeConstants.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Service.Models
```
aaronpowell, line 4: Let's start using [file-scoped namespaces](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces) to reduce the bloat in files.

Easy refactor in VS, just put a `;` on the end of line 4 and it'll fix the rest up for you.

src/Core/Models/SqlTypeConstants.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// String literal representation of SQL Server data types that are returned by Microsoft.Data.SqlClient.
+    /// </summary>
+    /// <seealso cref="https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/tests/PerformanceTests/Config/Constants.cs"/>
+    public class SqlTypeConstants
```
aaronpowell, line 10: This can be a `static` class as it only has static members right?

```suggestion
    public static class SqlTypeConstants
```

src/Core/Services/TypeHelper.cs
```diff
@@ -96,26 +91,117 @@ public static class TypeHelper
...
+        /// however SqlDbType.Binary now exists.
+        /// </summary>
+        /// <param name="dbTypeName">String value sourced from the DATA_TYPE column in the Procedure Parameters or Columns
+        /// schema collections.</param>
+        /// <seealso cref="https://learn.microsoft.com/dotnet/framework/data/adonet/sql-server-schema-collections#columns"/>
+        /// <seealso cref="https://learn.microsoft.com/dotnet/framework/data/adonet/sql-server-schema-collections#procedure-parameters"/>
+        /// <exception>Failed type conversion.</exception>"
+        public static Type GetSystemTypeFromSqlDbType(string sqlDbTypeName)
+        {
+            if (Enum.TryParse(enumType: typeof(SqlDbType), value: sqlDbTypeName, ignoreCase: true, out object? result) && result is not null)
```
aaronpowell, line 191: You can save the cast on line 193 with `Enum.TryParse<SqlDbType>`.

aaronpowell: You don't need to do the `null` test on `result` as `Enum.TryParse` has `NotNullWhen(true)` on the `out` parameter.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1423,7 +1422,7 @@ public async Task OpenApi_EntityLevelRestEndpoint()
             using TestServer server = new(Program.CreateWebHostBuilder(args));
             using HttpClient client = server.CreateClient();
             // Setup and send GET request
-            HttpRequestMessage readOpenApiDocumentRequest = new(HttpMethod.Get, $"{RestRuntimeOptions.DEFAULT_PATH}/{OpenApiDocumentor.OPENAPI_ROUTE}");
+            HttpRequestMessage readOpenApiDocumentRequest = new(HttpMethod.Get, $"{RestRuntimeOptions.DEFAULT_PATH}/{Core.Services.OpenAPI.OpenApiDocumentor.OPENAPI_ROUTE}");
```
aaronpowell, line 1425: Why inline the namespace?

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Collections.Generic;
+using System.Data;
+using Azure.DataApiBuilder.Core.Services;
+using Azure.DataApiBuilder.Core.Services.OpenAPI;
+using Azure.DataApiBuilder.Service.Exceptions;
+using Azure.DataApiBuilder.Service.Models;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Azure.DataApiBuilder.Service.Tests.OpenApiDocumentor
```
aaronpowell, line 13: Let's use [file-scoped namespaces](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces)

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        private const string SQLDBTYPE_RESOLUTION_ERROR = "failed to resolve to SqlDbType.";
+        private const string JSONDATATYPE_RESOLUTION_ERROR = "(when supported) should map to a system type and associated JsonDataType.";
+        private const string DBTYPE_RESOLUTION_ERROR = "(when supported) should map to a system type and associated DbType.";
+
+        /// <summary>
+        /// Validates that all DAB supported CLR types (system types) map to a defined JSON value type.
+        /// A DAB supported CLR type is a CLR type mapped from a database value type.
+        /// </summary>
+        [TestMethod]
+        public void SupportedSystemTypesMapToJsonValueType()
```
aaronpowell, line 31: Rather than using a `foreach` loop in the test, you should use a data-drive test and provide the collection [as describe here](https://learn.microsoft.com/en-us/visualstudio/test/how-to-create-a-data-driven-unit-test?view=vs-2022#member-data-driven-test).

This makes it more obvious when a particular test fails, as they will be named more appropriately, and simplifies debugging, as you can be more targeted on debugging only the failing test.

ayush3797:
Could you attach a screenshot of how the entity (which is attached in the issue) looks like now in swagger UI?

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
seantleonard: Thanks for the suggestion. I updated this to a DynamicData test and the tests are all distinct with readable test names. Makes it way easier to troubleshoot and identify which test case is running as you pointed out.

seantleonard:
> Could you attach a screenshot of how the entity (which is attached in the issue) looks like now in swagger UI?

Added before and after to the PR description

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -34,57 +32,12 @@ public override string GetDefaultSchemaName()
         }
 
         /// <summary>
-        /// Takes a string version of an MS SQL data type and returns its .NET common language runtime (CLR) counterpart
-        /// As per https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings
+        /// Takes a string version of an SQL Server data type and returns its .NET common language runtime (CLR) counterpart
```
Aniruddh25, line 35: Could be Azure SQL DB too, so SQL Server might be too specific.

src/Core/Services/TypeHelper.cs
```diff
@@ -2,120 +2,202 @@
...
+/// <summary>
+/// Type mapping helpers to convert between SQL Server types, .NET Framework types, and Json value types.
+/// </summary>
+/// <seealso cref="https://learn.microsoft.com/dotnet/framework/data/adonet/sql-server-data-type-mappings"/>
+public static class TypeHelper
 {
     /// <summary>
-    /// Helper class used to resolve CLR Type to the associated DbType or JsonDataType
+    /// Maps .NET Framework types to DbType enum
+    /// Unnecessary to add nullable types because
```
Aniruddh25, line 19: nit: incomplete reasoning? why not adding nullable types?

src/Core/Services/TypeHelper.cs
```diff
@@ -2,120 +2,202 @@
...
-        /// </summary>
-        /// <param name="systemType">The system type for which the DbType is to be determined.</param>
-        /// <returns>DbType for the given system type.</returns>
-        public static DbType? GetDbTypeFromSystemType(Type systemType)
+    /// <summary>
+    /// Maps SqlDbType enum to .NET Framework type (System type).
+    /// </summary>
+    private static readonly Dictionary<SqlDbType, Type> _sqlDbTypeToType = new()
+    {
+        { SqlDbType.BigInt,typeof(long) },
```
aaronpowell, line 103: The way this dictionary is defined is different to the ones above (they're using indexer assignment rather than providing an enumable of key/value pairs).

Should we have one pattern used consistently?

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronpowell, line 1: Do we know why this file has had a huge amount of changes? Is it because you changed it to a file-scoped namespace?

Might be worth reverting that in this file so it's easier to see what changes you've added, as we already have a mix of file scoped and non-file scoped, so we should only adopt it progressively when it's clearer that we might not "hide" changes.

aaronpowell (approved):

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: reverted.

src/Core/Services/TypeHelper.cs
seantleonard: Updated for consistency to be defined by providing an enumerable of key/value pairs.

src/Core/Services/TypeHelper.cs
seantleonard: updated comment to apply to correct dictionary. Not this one (can be a follow up PR since changing this one would be a separate concern.)
also added unit test and explanation for not adding nullable types to the dictionary.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
seantleonard: https://docs.microsoft.com/dotnet/framework/data/adonet/sql-server-data-type-mappings

Since there doesn't seem to be Azure SQL data types explicitly that differentiate from the defined Sql Server data types , I'm updating the comment to reflect that Sql Server data types _also apply to Azure SQL DB_

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: My changes now accurately show in the github diff tool

src/Core/Services/TypeHelper.cs
Aniruddh25:
```suggestion
        { SqlDbType.BigInt, typeof(long) },
```

src/Core/Models/SqlTypeConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+            { "tinyint", true },              // SqlDbType.TinyInt
+            { "varbinary", true },            // SqlDbType.VarBinary
+            { "varchar", true },              // SqlDbType.VarChar
+            { "sql_variant", false },         // SqlDbType.Variant (unsupported)
+            { "xml", false },                 // SqlDbType.Xml (unsupported)
+            { "date", true },                 // SqlDbType.Date
+            { "time", true },                 // SqlDbType.Time
+            { "datetime2", true },            // SqlDbType.DateTime2
+            { "datetimeoffset", true },       // SqlDbType.DateTimeOffset
+            { "", false },                    // SqlDbType.Udt and SqlDbType.Structured provided by SQL as empty strings (unsupported)
```
Aniruddh25, line 50: How about geography, spatial? Should we add this dictionary to TypeHelper?

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            Type resolvedType = TypeHelper.GetSystemTypeFromSqlDbType(sqlDataTypeLiteral);
+            Assert.AreEqual(true, isSupportedSqlDataType, ERROR_PREFIX + $"{{{sqlDataTypeLiteral}}} " + SQLDBTYPE_RESOLUTION_ERROR);
+
+            JsonDataType resolvedJsonType = TypeHelper.GetJsonDataTypeFromSystemType(resolvedType);
+            Assert.AreEqual(isSupportedSqlDataType, resolvedJsonType != JsonDataType.Undefined, ERROR_PREFIX + $"{{{sqlDataTypeLiteral}}} " + JSONDATATYPE_RESOLUTION_ERROR);
+        }
+        catch (DataApiBuilderException)
+        {
+            Assert.AreEqual(false, isSupportedSqlDataType, ERROR_PREFIX + $"{{{sqlDataTypeLiteral}}} " + SQLDBTYPE_RESOLUTION_ERROR);
```
Aniruddh25, line 43: The error here if this fails would be `The SqlDbType x failed to resolve to SqlDbType`. Should this instead be `The SqlDbType x failed to resolve to SystemType` ?

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
```diff
@@ -0,0 +1,102 @@
...
+    private const string SQLDBTYPE_RESOLUTION_ERROR = "failed to resolve to SqlDbType.";
+    private const string JSONDATATYPE_RESOLUTION_ERROR = "(when supported) should map to a system type and associated JsonDataType.";
+    private const string DBTYPE_RESOLUTION_ERROR = "(when supported) should map to a system type and associated DbType.";
+
+    /// <summary>
+    /// Validates that all DAB supported CLR types (system types) map to a defined JSON value type.
+    /// A DAB supported CLR type is a CLR type mapped from a database value type.
+    /// </summary>
+    [TestMethod]
+    [DynamicData(nameof(GetTestData_SupportedSystemTypesMapToJsonValueType), DynamicDataSourceType.Method)]
```
Aniruddh25, line 30: Like this DynamicData approach!

src/Core/Services/TypeHelper.cs
seantleonard: This is from an older commit. i addressed Aaron's feedback and made this consistent with how other dictionaries in this file are created.

```csharp
// old
{ SqlDbType.BigInt,typeof(long) },

//new
[SqlDbType.BigInt] = typeof(long),
```

Aniruddh25 (approved):
LGTM, Thanks for refactoring our type system!

Aniruddh25:
Does `0` as the JsonType value in the `After` correspond to `Number`?

src/Core/Services/TypeHelper.cs
```diff
@@ -29,6 +35,7 @@ public static class TypeHelper
             [typeof(char)] = DbType.StringFixedLength,
             [typeof(Guid)] = DbType.Guid,
             [typeof(byte[])] = DbType.Binary,
+            [typeof(DateTimeOffset)] = DbType.DateTimeOffset,
```
Aniruddh25, line 38: I wonder this was leftout as per this PR:
https://github.com/Azure/data-api-builder/pull/1442
for a reason.

Probably, we should NOT add it in this PR to avoid regression unless we know for sure - although its interesting to see all tests are passing.

Aniruddh25 (changes requested):
Left a question for DateTimeOffset to DbType mapping.

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
seantleonard: Good catch, this should read as "The input string value x failed to resolve to <SqlDbType || System Type || JsonDataType>" to be more accurate.

src/Core/Services/TypeHelper.cs
seantleonard: GraphQL Supported types tests failed when I included system types DateTime, DateTime2, and TimeSpan here. Those look to be handled by #1473  I reran those tests locally and they passed when i removed the entry you pointed out. So i'm removing to avoid any possible regression.

seantleonard:
> Does `0` as the JsonType value in the `After` correspond to `Number`?

My bad, that was actually a default value (thanks for catching). A default value resolved by swagger UI based on the json data type. I added extra example of the raw openapi schema doc which shows the actual json data type resolved as a result of this change. (this requires extra click throughs in swagger to 'view schema').

src/Service.Tests/OpenApiDocumentor/CLRtoJsonValueTypeUnitTests.cs
seantleonard: updated

src/Core/Models/SqlTypeConstants.cs
seantleonard: updated that issue with task to investigate newest preview of sqlclient which references spatial types (Microsoft.SqlServer.Types.dll)

src/Core/Services/TypeHelper.cs
seantleonard: removed. Running tests in ci/cd to confirm

Aniruddh25 (approved):
LGTM, thanks again for refactoring our Type system.

src/Core/Models/SqlTypeConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// SqlDbType names ordered by corresponding SqlDbType.
+    /// Keys are lower case to match formatting of SQL Server INFORMATION_SCHEMA DATA_TYPE column value.
+    /// Sourced directly from internal/private SmiMetaData.cs in Microsoft.Data.SqlClient
+    /// Value indicates whether DAB engine supports the SqlDbType.
+    /// </summary>
+    /// <seealso cref="https://github.com/dotnet/SqlClient/blob/2b31810ce69b88d707450e2059ee8fbde63f774f/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Server/SmiMetaData.cs#L637-L674"/>
+    public static readonly Dictionary<string, bool> SupportedSqlDbTypes = new()
+    {
+            { "bigint", true },               // SqlDbType.BigInt
```
abhishekkumams, line 21: the indentation seems off, looks like an extra tab from line#21 to line#51

src/Core/Models/SqlTypeConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+            { "binary", true },               // SqlDbType.Binary
+            { "bit", true },                  // SqlDbType.Bit
+            { "char", true },                 // SqlDbType.Char
+            { "datetime", true },             // SqlDbType.DateTime
+            { "decimal", true },              // SqlDbType.Decimal
+            { "float", true },                // SqlDbType.Float
```
abhishekkumams, line 27: what happens when a sqlDataType value is not present in this SupportedSqlDbTypes dictionary? do we throw any exception?

src/Core/Models/SqlTypeConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+            { "varbinary", true },            // SqlDbType.VarBinary
+            { "varchar", true },              // SqlDbType.VarChar
+            { "sql_variant", false },         // SqlDbType.Variant (unsupported)
+            { "xml", false },                 // SqlDbType.Xml (unsupported)
+            { "date", true },                 // SqlDbType.Date
+            { "time", true },                 // SqlDbType.Time
+            { "datetime2", true },            // SqlDbType.DateTime2
+            { "datetimeoffset", true },       // SqlDbType.DateTimeOffset
+            { "", false },                    // SqlDbType.Udt and SqlDbType.Structured provided by SQL as empty strings (unsupported)
+        };
```
abhishekkumams, line 51: I see a you have created an issue to include more SqlDbTypes, but do we have a test case where the sqlDataTypeLiteral is a value that is not present in the SupportedSqlDbTypes dictionary?

seantleonard: Just added a test for this in another data row. Unexpected SqlDbType values provided by SQL which aren't in SQL documentation properly resolve to `JsonDataType.Undefined`

src/Core/Models/SqlTypeConstants.cs
seantleonard: thanks, fixed!

PR #1574, opened 2023-07-08
Adding constructor to RunTimeConfigProvider so that we have option to…

rohkhann:
… initialize the RuntimeConfigProvider with runtimeconfig if being created dynamically.

## Why make this change?

  Need ability to provide a runtimeconfig into runtimeconfigprovider if being initialized dynamically without a .json file.

## What is this change?

  Added extra constructor. 
  Made RuntimeConfigLoader nullable as for the case where there is no file to load for, it is not valid. 

## How was this tested?

- [ ] Integration Tests
yes
- [ ] Unit Tests
yes

### DIFF
--- a/src/Core/Configurations/RuntimeConfigProvider.cs
+++ b/src/Core/Configurations/RuntimeConfigProvider.cs
@@ -40,14 +40,19 @@ public class RuntimeConfigProvider
     /// </summary>
     public string? ManagedIdentityAccessToken { get; private set; }
 
-    private readonly RuntimeConfigLoader _runtimeConfigLoader;
+    private readonly RuntimeConfigLoader? _runtimeConfigLoader;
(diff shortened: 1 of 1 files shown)


aaronpowell (changes requested):
I don't understand the scenario that is trying to be tackled here, as there is no usage of this change added to the codebase.

There is intended to be a clear separation between responsibilities, that a `RuntimeConfigProvider` is the thing that is responsible for _providing access to the `RuntimeConfig`_ and it shouldn't ever receive one from somewhere else.

Changing this design means that the hierachy of access is changed and we could see an introduction of unintended side effects.

rohkhann:
Closing pull request. going to follow design discussed to create Interfact for runtimeconfigloader and provide different clients ability to implement it as applicable.

PR #1580, opened 2023-07-14
[BugFix] Handle configs which have missing options for MsSql

Aniruddh25:
## Why make this change?

- With the recent refactor #1402, if `options` property is not specified in the config, after deserialization the `Options` dictionary of `DataSource` object is `null`  
- This leads to a null dereference when trying to `GetTypedOptions(MsSqlOptions)` [here](https://github.com/Azure/data-api-builder/blob/bda0d6551f8f75fc7b98f888bdfce1f3dd317378/src/Core/Resolvers/MsSqlQueryExecutor.cs#L76C70-L76C70). 

## What is this change?

- Make `Options` a nullable property since its not required as per the schema.
- Handle all the instances of Options considering it could be null.

## How was this tested?

- Manually running dab.
- Automated tests:
      1. Removed the `options: set-session-context` from the initial config used in the CLI tests to verify deserialization results in a null  `Options` dictionary as demonstrated by the Snapshots
      2. Modified tests which create a custom config and start the engine by setting `Options` to null and verifying those continue to pass.  
Note: The checked in configuration file for MsSql sets the session context to true for integration tests of row level security feature. So, did not modify that file even though options is optional for MsSql.

### DIFF
--- a/src/Config/ObjectModel/DataSource.cs
+++ b/src/Config/ObjectModel/DataSource.cs
@@ -12,40 +12,62 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 /// </summary>
 /// <param name="DatabaseType">Type of database to use.</param>
 /// <param name="ConnectionString">Connection string to access the database.</param>
-/// <param name="Options">Custom options for the specific database. If there are no options this will be empty.</param>
-public record DataSource(DatabaseType DatabaseType, string ConnectionString, Dictionary<string, JsonElement> Options)
+/// <param name="Options">Custom options for the specific database. If there are no options, this could be null.</param>
+public record DataSource(DatabaseType DatabaseType, string ConnectionString, Dictionary<string, JsonElement>? Options)
 {
     /// <summary>
     /// Converts the <c>Options</c> dictionary into a typed options object.
+    /// May return null if the dictionary is null.
     /// </summary>
     /// <typeparam name="TOptionType">The strongly typed object for Options.</typeparam>
     /// <returns>The strongly typed representation of Options.</returns>
     /// <exception cref="NotSupportedException">Thrown when the provided <c>TOptionType</c> is not supported for parsing.</exception>
-    public TOptionType GetTypedOptions<TOptionType>() where TOptionType : IDataSourceOptions
+    public TOptionType? GetTypedOptions<TOptionType>() where TOptionType : IDataSourceOptions
     {
         HyphenatedNamingPolicy namingPolicy = new();
 
         if (typeof(TOptionType).IsAssignableFrom(typeof(CosmosDbNoSQLDataSourceOptions)))
         {
-            return (TOptionType)(object)new CosmosDbNoSQLDataSourceOptions(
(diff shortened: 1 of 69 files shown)


src/Service.Tests/TestHelper.cs
```diff
@@ -115,10 +115,7 @@ public static RuntimeConfig AddMissingEntitiesToConfig(RuntimeConfig config, str
         public const string SAMPLE_SCHEMA_DATA_SOURCE = SCHEMA_PROPERTY + "," + @"
             ""data-source"": {
               ""database-type"": ""mssql"",
-              ""connection-string"": ""testconnectionstring"",
-              ""options"": {
-                ""set-session-context"": true
-                }
```
seantleonard, line 121: removing options here to cover non-MsSql databases where these options will be null to demonstrate that null dereference no longer occurs?

seantleonard (commented):
Thanks for adding. Pending addressing ci/cd build error

src/Config/ObjectModel/DataSource.cs
```diff
@@ -12,40 +12,62 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
...
 {
     /// <summary>
     /// Converts the <c>Options</c> dictionary into a typed options object.
+    /// May return null if the dictionary is null.
     /// </summary>
     /// <typeparam name="TOptionType">The strongly typed object for Options.</typeparam>
     /// <returns>The strongly typed representation of Options.</returns>
     /// <exception cref="NotSupportedException">Thrown when the provided <c>TOptionType</c> is not supported for parsing.</exception>
-    public TOptionType GetTypedOptions<TOptionType>() where TOptionType : IDataSourceOptions
+    public TOptionType? GetTypedOptions<TOptionType>() where TOptionType : IDataSourceOptions
```
aaronpowell, line 25: Rather than returning `null`, should we return the default for the requested options?

Aniruddh25: I am actually returning the default already. But since that could be null, VS complains if I dont make the return type nullable

abhishekkumams (approved):

src/Config/ObjectModel/DataSource.cs
aaronpowell: I meant constructing an object with default values, not returning `default`, which will return `null` since the options here are reference types not value types (if we wanted them to be value types they'd need to be `record struct` instead of `record`).

Aniruddh25: We dont know what the default values for `CosmosDbNoSQLDataSourceOptions` should be, those are mandatory options. But if they are not specified, we need to throw an exception. To identify if they were specified or not by the customer, null check is better than trying to check for default values that we might assign to the object. 
For CosmosDbOptions, it will also be harder to distinguish between default values assigned by us v/s what the customer is providing.

I let MsSqlOptions to have default values if not specified.

src/Service.Tests/TestHelper.cs
Aniruddh25: to check for null dereference, modified the tests that start the engine and actually execute some queries. Removing here, to simulate deserialization results in options being null.

ayush3797 (approved):
LGTM!

PR #1587, opened 2023-07-17
Rohkhann/run time config constructor

rohkhann:
## Why make this change?

Different users of dataapi builder may have different ways in which they want to load in runtimeconfig. For example: There is a scenario where one might want to create the runtimeconfig object dynamically without having a runtime json ready.

## What is this change?

In this pr, we create an abstract class BaseRunTimeConfigLoader that has certain common methods that all derived classes can use. The main TryLoadconfig method will be implemented differently by different classes. The RunTimeConfigProvider by dab loads config from the filesystem. Depending on use case this can be done differently by different consumers.
- Also made some name simplifications in test files.

## How was this tested?

- [x] Integration Tests
yes
- [x] Unit Tests
yes


src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <returns>True if the config was loaded, otherwise false.</returns>
+        public abstract bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config);
+
+        /// <summary>
+        /// Parses a JSON string into a <c>RuntimeConfig</c> object
+        /// </summary>
+        /// <param name="json">JSON that represents the config file.</param>
+        /// <param name="config">The parsed config, or null if it parsed unsuccessfully.</param>
+        /// <returns>True if the config was parsed, otherwise false.</returns>
+        public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeConfig? config, ILogger? logger = null, string? connectionString = null)
```
aaronpowell, line 35: Doesn't this make the assumption that all implementations of a loader are going to do JSON deserialization?

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
aaronpowell, line 1: This should be at the root of the `Config` project, same level as `RuntimeConfigLoader`.

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
```diff
@@ -0,0 +1,98 @@
...
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.Converters;
+using Azure.DataApiBuilder.Config.NamingPolicies;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.DataApiBuilder.Config.ObjectModel
```
aaronpowell, line 11: Use a file-scoped namespace, putting a `;` at the end of the line will refactor everything.

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    public abstract class BaseRunTimeConfigLoader
```
aaronpowell, line 13:
```suggestion
    public abstract class RuntimeConfigLoader
```

We should call this `RuntimeConfigLoader` and the current implementation `FileSystemRuntimeConfigLoader` so it's clear what the implementation does.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -27,12 +23,11 @@ namespace Azure.DataApiBuilder.Config;
 /// which allows for mocking of the file system in tests, providing a way to run the test
 /// in isolation of other tests or the actual file system.
 /// </remarks>
-public class RuntimeConfigLoader
+public class RuntimeConfigLoader : BaseRunTimeConfigLoader
```
aaronpowell, line 26:
```suggestion
public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
```

This would be dependent on the other change that I've suggested.

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
rohkhann: This seemed more like a utlity method as it didnt operate on any class variables to Load a runtime config object from a json.  The same code can be used by all implementators of this base class. For the scenario where you dynamically initialize the runtimeconfig you could have a case where you are doing this only for the first time then serializing to a json storing it in a cache then on next call loading it in and it would be useful to have the tryparseconfig.

aaronpowell (changes requested):
Added a few notes around naming conventions.

While I've provided them as suggestions on the GitHub code review, it'd be better to do the refactor in VS as then it'll update all references.

Also, the file names will need to be updated to match the new type names.

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
aaronpowell: I get that we should push the method up to the base class, since it's something that we'll need to implement for all the different providers. But my understanding was that the provider you are looking at won't be providing JSON so you're going to override this anyway, so we'd be best to have this as just an abstract method and we have one implementation that is doing JSON parsing (the current implementation) and another that is doing your parsing.

src/Config/RuntimeConfigLoader.cs
rohkhann: Resolved.

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
rohkhann: Fixed.

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
rohkhann: fixed.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -2,72 +2,30 @@
...
+    public RuntimeConfigLoader(string? connectionString = null)
     {
-        _fileSystem = fileSystem;
-        _baseConfigFileName = baseConfigFileName;
         _connectionString = connectionString;
     }
 
     /// <summary>
-    /// Load the runtime config from the specified path.
+    /// Returns RunTimeConfig. Needs to be implemented by the derived class.
```
aaronpowell, line 24:
```suggestion
    /// Returns RuntimeConfig. Needs to be implemented by the derived class.
```

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -40,12 +40,12 @@ public class RuntimeConfigProvider
...
     public string? ManagedIdentityAccessToken { get; private set; }
 
-    private readonly RuntimeConfigLoader _runtimeConfigLoader;
+    private readonly FileSystemRuntimeConfigLoader _runtimeConfigLoader;
     private RuntimeConfig? _runtimeConfig;
 
     public string RuntimeConfigFileName => _runtimeConfigLoader.ConfigFileName;
 
-    public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
+    public RuntimeConfigProvider(FileSystemRuntimeConfigLoader runtimeConfigLoader)
```
aaronpowell, line 48: Shouldn't that take a `RuntimeConfigLoader` now, so it's the base class?

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
rohkhann: We would be providing json possibily on subsequent calls if we decide to store a json of the runtimeconfig and store it in local cache. I thought the tryparseconfig method was very generic and pretty much what is to be expected with json parsing and hence kept the implementation there.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -130,7 +130,7 @@ public bool TryGetLoadedConfig([NotNullWhen(true)] out RuntimeConfig? runtimeCon
             throw new ArgumentException($"'{nameof(configuration)}' cannot be null or empty.", nameof(configuration));
         }
 
-        if (RuntimeConfigLoader.TryParseConfig(
+        if (FileSystemRuntimeConfigLoader.TryParseConfig(
```
aaronpowell, line 133: Should this be calling the instance method rather than the static method from the `_runtimeConfigLoader` field?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -182,7 +182,7 @@ public async Task<bool> Initialize(string jsonConfig, string? graphQLSchema, str
 
         IsLateConfigured = true;
 
-        if (RuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig? runtimeConfig))
+        if (FileSystemRuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig? runtimeConfig))
```
aaronpowell, line 185: Should this be calling the instance method rather than the static method from the `_runtimeConfigLoader` field?

src/Config/ObjectModel/BaseRunTimeConfigLoader.cs
rohkhann: fixed.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: this is fixed. missed this earlier with some merge conflict fixes.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: fixed.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: fixed.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -0,0 +1,254 @@
...
+            Console.WriteLine($"Unable to find config file: {fileName} does not exist.");
+            return false;
+        }
+    }
+
+    /// <summary>
+    /// This method reads the dab.draft.schema.json which contains the link for online published
+    /// schema for dab, based on the version of dab being used to generate the runtime config.
+    /// </summary>
+    public string GetPublishedDraftSchemaLink()
```
aaronpowell, line 195: Not needed in this PR, but we'll probably want to move this as an abstract method on the base class, as other implementations are likely to need to get the schema link. The implementation _might_ also be valid in the base class, but we can evaluate that when the time comes.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -2,72 +2,30 @@
...
+    public RuntimeConfigLoader(string? connectionString = null)
     {
-        _fileSystem = fileSystem;
-        _baseConfigFileName = baseConfigFileName;
         _connectionString = connectionString;
     }
 
     /// <summary>
-    /// Load the runtime config from the specified path.
+    /// Returns RuntimeConfig. Needs to be implemented by the derived class.
```
aaronpowell, line 24: nit: You don't need to state that it needs implementing in the derived class, that's implied by the `abstract` keyword.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -116,6 +74,9 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
         return true;
     }
 
+    /// <summary>
+    /// Get Serializer options for the config file.
+    /// </summary>
     public static JsonSerializerOptions GetSerializationOptions()
```
aaronpowell, line 80: Since we don't use the JSON serializer in the base class, we should lift that up to the `FileSystemRuntimeConfigProvider`.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -40,14 +40,13 @@ public class RuntimeConfigProvider
     /// </summary>
     public string? ManagedIdentityAccessToken { get; private set; }
 
-    private readonly RuntimeConfigLoader _runtimeConfigLoader;
-    private RuntimeConfig? _runtimeConfig;
+    public RuntimeConfigLoader ConfigLoader { get; private set; }
```
aaronpowell, line 43: What's the reason that this is public? There shouldn't be a need for anything that uses the Provider to directly access the Loader.

src/Service/Startup.cs
```diff
@@ -58,12 +58,12 @@ public Startup(IConfiguration configuration, ILogger<Startup> logger)
...
         {
-            string configFileName = Configuration.GetValue<string>("ConfigFileName", RuntimeConfigLoader.DEFAULT_CONFIG_FILE_NAME);
+            string configFileName = Configuration.GetValue<string>("ConfigFileName", FileSystemRuntimeConfigLoader.DEFAULT_CONFIG_FILE_NAME);
             string? connectionString = Configuration.GetValue<string?>(
-                RuntimeConfigLoader.RUNTIME_ENV_CONNECTION_STRING.Replace(RuntimeConfigLoader.ENVIRONMENT_PREFIX, ""),
+                FileSystemRuntimeConfigLoader.RUNTIME_ENV_CONNECTION_STRING.Replace(FileSystemRuntimeConfigLoader.ENVIRONMENT_PREFIX, ""),
                 null);
             IFileSystem fileSystem = new FileSystem();
-            RuntimeConfigLoader configLoader = new(fileSystem, configFileName, connectionString);
+            FileSystemRuntimeConfigLoader configLoader = new(fileSystem, configFileName, connectionString);
```
aaronpowell, line 66: As a note, we'll need to think how this section is implemented, as at the moment it's very heavily tied to the Loader implementation, but if we want to support others, we're going to need a way to specify _which_ Loader implementation to use.

I did try and do this in the past, but there was a problem that we need to access the Provider elsewhere in the codebase (seen in line 287 for example) and we can't resolve it from the DI container as the container isn't built yet.

src/Service/Startup.cs
```diff
@@ -287,13 +287,15 @@ private void AddGraphQLService(IServiceCollection services)
         public void Configure(IApplicationBuilder app, IWebHostEnvironment env, RuntimeConfigProvider runtimeConfigProvider)
         {
             bool isRuntimeReady = false;
+            FileSystemRuntimeConfigLoader fileSystemRuntimeConfigLoader = (FileSystemRuntimeConfigLoader)runtimeConfigProvider.ConfigLoader;
```
aaronpowell, line 290: Can we create a task that notes we'll need to clean this up in the future since there will be instances where we won't have the concept of a **filename** to use in the log message on line 298/exception on 309, so the message will need to be refactored/reworked to provide contextually accurate information.

aaronpowell (approved):
Generally speaking this looks good.

Added a small nit in there and some notes on where we've got areas that need to be reworked in the future phases of this workstream.

src/Config/FileSystemRuntimeConfigLoader.cs
rohkhann: Makes sense. i will move it as part of this pr as this seems like a very valid thing to have in the base class.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: There was code in startup that depends on showing error logs based on where the runtimeconfig filepath is. I did not want to alter that behaviour. Since Runtimeconfigprovider now has a generic runtimeconfigloader and not a filesystemloader, there was no way to get the filepath without getting the loader.

src/Service/Startup.cs
rohkhann: Sure I will create a task to clean this up. https://github.com/Azure/data-api-builder/issues/1591#issue-1810334013

src/Config/RuntimeConfigLoader.cs
rohkhann: We use the Json serializer options in the TryParseConfig method in the base class.

src/Service/Startup.cs
rohkhann: Yeah this makes sense.

src/Config/RuntimeConfigLoader.cs
rohkhann: fixed.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -116,6 +80,9 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
         return true;
     }
 
+    /// <summary>
+    /// Get Serializer options for the config file.
+    /// </summary>
```
seantleonard, line 85: Can you clarify why the base class has `TryParseConfig()`? Wouldn't config parsing of JSON occur in the fileSystemRuntimeconfigloader?

src/Service/Startup.cs
seantleonard: will need a task for this one as well

seantleonard (commented):
thanks for contributing these changes! just a couple comments

src/Config/RuntimeConfigLoader.cs
rohkhann: Had conversation with Aaron on this above. Overall, this is because this function seems more like a utility function -> given a runtime json can you give me the corresponding runtimeconfig and doesnt seem a specific use case for filesystemruntimeconfigloader. For other loaders as well, say the one that dynamically creates a runtimeconfig, it could be useful to in the first iteration of the call create the runtimeconfig dynamically and then store the runtimeconfig json on a cache as a json and on all subsequent calls pull the json from the cache and deserialize using this utility method.

src/Service/Startup.cs
rohkhann: https://github.com/Azure/data-api-builder/issues/1593. added here.

seantleonard (approved):
Thanks for adding clarification to my questions

src/Core/Configurations/RuntimeConfigProvider.cs
aaronpowell: Yeah I did end up seeing it (and commented on it later in the PR), I was just lazy and didn't delete this comment 😅

Issue #1584, opened 2023-07-17
[Bug]: UPSERT doesn't work if there is a calculated/generated column - PgSql


ayush3797:
On similar lines to option A discussed for MsSql in https://github.com/Azure/data-api-builder/issues/1453, need to investigate and implement a similar solution for PgSql.

ayush3797:
Postgres supports generated columns just like MySql. As specified here in the [Pg docs](https://www.postgresql.org/docs/current/ddl-generated-columns.html), `PostgreSQL currently implements only stored generated columns` and not VIRTUAL ones. We can determine if a column is generated using this query:

```sql
SELECT
    attname AS column_name
FROM
    pg_attribute
WHERE
    attrelid = 'table_schema.table_name'::regclass
    AND attgenerated = 's';
```

Issue #1583, opened 2023-07-17
[Bug]: UPSERT doesn't work if there is a calculated column - Mysql


ayush3797:
On similar lines to option A discussed for MsSql in https://github.com/Azure/data-api-builder/issues/1453, need to implement a similar solution for MySql.

ayush3797:
**For tables in MySql**, _generated columns (also known as computed columns)_ are not updatable. Refer here: [Generated columns in MySql](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html#:~:text=CREATE%20VIEW%20v1%20AS%20SELECT%20%2A%2C%20CONCAT%20%28first_name%2C%27,AS%20full_name%20FROM%20t1%3B%20SELECT%20full_name%20FROM%20v1%3B) to get more idea about generated columns. Generated columns can be VIRTUAL/STORED. 

- VIRTUAL: Column values are not stored, but are evaluated when rows are read, immediately after any BEFORE triggers. A virtual column takes no storage.
- STORED: Column values are evaluated and stored when rows are inserted or updated. A stored column does require storage space and can be indexed.

I came across two solutions to collect metadata about generate columns:

1. Use the below sql:
```sql
SELECT COLUMN_NAME,EXTRA
FROM INFORMATION_SCHEMA.COLUMNS
WHERE TABLE_SCHEMA = 'mysql'
  AND TABLE_NAME = 'todos'
  AND COLUMN_DEFAULT IS NULL
  AND EXTRA = 'VIRTUAL GENERATED';
```
But this solution only works for virtual generated columns and not the stored ones.

2. Use the below sql to get all the generated columns which have generation_expression != NULL.
```sql
select column_name,generation_expression from information_schema.columns 
where table_schema='mysql' and table_name = 'todos';
```
This solution works for both STORED/VIRTUAL generated columns. Will go ahead with this.

ayush3797:
While I was working on implementing this solution for MySql, I encountered a limitation of MySql. Currently if you see, upsert queries for MySql is generated as:

```sql
SET @LU_0 := 0;
SET @LU_1 := 0;
SET @LU_2 := 0;
SET @LU_3 := 0;
SET @LU_4 := 0;
[...]

Issue #1453, opened 2023-04-25
[Bug]: UPSERT doesn't work if there is a calculated column - MsSql


yorek:
### What happened?

If there is a calculated column in a table, for example:

```sql
create table [dbo].[todos]
(
  [id] [uniqueidentifier] not null,
  [title] [nvarchar](1000) not null,
  [completed] [bit] not null,
  [custom] [nvarchar](max) null,	
  [owner_id] [varchar](128) not null,
  [created_on] datetime2(7) not null,
) 
go
alter table dbo.todos
add created_on_date as cast(created_on as date)
go
```

then when doing an UPSERT the operation will fail with the following error:

```json
{
  "error": {
    "code": "DatabaseOperationFailed",
    "message": "The column \u0022created_on_date\u0022 cannot be modified because it is either a computed column or is the result of a UNION operator.",
    "status": 500
  }
}
```

the generated SQL, in fact, should *not* do any operation on the calculated colum, but instead it tries to update/insert it:

```
UPDATE [dbo].[todos] WITH(UPDLOCK) SET [dbo].[todos].[title] = @param2, [dbo].[todos].[completed] = @param3, [dbo].[todos].[owner_id] = @param4, [dbo].[todos].[custom] = @param5, [dbo].[todos].[created_on] = @param6, [dbo].[todos].[position] = @param7, [dbo].[todos].[created_on_date] = @param8 OUTPUT Inserted.[id] AS [tid], Inserted.[title] AS [title], Inserted.[completed] AS [isDone], Inserted.[custom] AS [custom], Inserted.[owner_id] AS [visibility], Inserted.[created_on] AS [created_on], Inserted.[position] AS [position], Inserted.[created_on_date] AS [created_on_date] WHERE [dbo].[todos].[id] = @param1 AND ([owner_id] = @param0) 

IF @@ROWCOUNT = 0 BEGIN; 
INSERT INTO [dbo].[todos] ([id], [title], [completed], [owner_id]) OUTPUT Inserted.[id] AS [tid], Inserted.[title] AS [title], Inserted.[completed] AS [isDone], Inserted.[custom] AS [custom], Inserted.[owner_id] AS [visibility], Inserted.[created_on] AS [created_on], Inserted.[position] AS [position], Inserted.[created_on_date] AS [created_on_date] VALUES (@param1, @param2, @param3, @param4)END
```

### Version

0.6.13

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

[...]

PR #1596, opened 2023-07-20
Preventing update/insert of read-only fields in a table by user

ayush3797:
## Why make this change?
There are certain type of columns in databases which cannot be provided a value (not even null value) for by the user. Eg. in MsSql tables, we cannot provide values for columns that are computed base on other columns in the table or have a datatype of timestamp/rowversion. Similarly for PgSql,/MySql we cannot provide a  value for generated columns. This PR prevents DAB from allowing user to attempt an update/insert of such columns. The way DAB will do this is by collecting metadata about columns during the startup phase and populate the `IsReadOnly` property of the `ColumnDefinition` of each column in the table. More details about the specifics can be found on the linked issues for MsSql(https://github.com/Azure/data-api-builder/issues/1453), PgSql(https://github.com/Azure/data-api-builder/issues/1584), MySql(https://github.com/Azure/data-api-builder/issues/1583).

**NOTE:** The columns which are autogenerated are also considered as read-only columns irrespective of the database flavor (MsSql/PgSql/MySql).

## What is this change?
- Added a new boolean property `IsReadOnly` to `ColumnDefinition` class which if true, will indicate that the column is read-only, i.e. it cannot be provided a value for (via INSERT/UPDATE).
- Added a new method `GetQueryToGetReadOnlyColumns()` to `IQueryBuilder.cs` class which will have its overridden implementation in Pg/My/MsSql. This method will return the query that will fetch the metadata about the whether a column is read-only.
- Added validations to `RequestValidator.cs` class which will ensure that no read-only column is present the request body either for update (via `Upsert`/`UpsertIncremental`/`Update`/`UpdateIncremental`) or insert operations.
- Added similar validations to `SqlInsertQueryStructure.cs`/`SqlUpdateQueryStructure.cs`- classes as the request validation stage is not a part of the codeflow for GQL requests.
- Via GQL, value for a read-only field cannot be provided. This will be enabled by adding the `AutoGeneratedDirectiveType` when a column is read-only. Previous this directive was only added for autogenerated fields (which is a subset of read-only fields).

## Implemented Behaviors:
1. `When a read-only field is included in the request body for mutation:` This is a bad request as we cannot provide values for read-only fields. We will throw an exception no matter what operation is being executed (`PUT/PATCH/POST via REST). For create/update via GQL`, we would see a warning in the BCP UI for read-only fields as well that the field is not valid for mutation (just like we see currently for autogen fields).
2. `When a read-only field is excluded from the request body for mutation`: This is a valid scenario. In this case (`for PUT/PATCH operations via REST`), we won't NULL out the field's value but let the responsibility on the database to deal with it. For other operations, we really don't need to do anything as the field is not present in the request body.

## How was this tested?

- [x] Integration Tests


src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -252,7 +253,17 @@ CREATE TABLE mappedbookmarks
...
-) 
+)
+
+create table books_sold
+(
+  [id] int PRIMARY KEY not null,
+  [book_name] varchar(50),
+  [row_version] timestamp,
+  [copies_sold] int default 0,
+  [last_sold_on] datetime2(7),
```
Aniruddh25, line 264: `last_sold_on` doesnt have a default value, so it will be nullified. So `last_sold_on_date` will also be null. 
 Which is the same effect if it were NOT a read only field. Can we set a default value to `last_sold_on` to say CURRENT_TIMESTAMP to test a non-null value gets assigned to `last_sold_on_date`. 

 At the same time, we already have a table `stocks_price` which has `instant` column with a default value of CURRENT_TIMESTAMP. This means we can add a column to that table to test the same functionality. Trying to reduce the number of additional tables we need for testing.

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -252,7 +253,17 @@ CREATE TABLE mappedbookmarks
...
 	id int IDENTITY(1,1) PRIMARY KEY,
 	bkname nvarchar(50) NOT NULL
-) 
+)
+
+create table books_sold
+(
+  [id] int PRIMARY KEY not null,
+  [book_name] varchar(50),
+  [row_version] timestamp,
```
Aniruddh25, line 262: Do we even support timestamp datatype? Given than `time`datatype is not checked in yet. What does this `timestamp` type get mapped to?

src/Service.Tests/DatabaseSchema-PostgreSql.sql
```diff
@@ -217,6 +218,14 @@ CREATE TABLE mappedbookmarks(
     bkname text NOT NULL
 );
 
+CREATE TABLE books_sold (
+  id int PRIMARY KEY,
+  book_name TEXT,
+  copies_sold INT DEFAULT 0,
+  last_sold_on TIMESTAMP,
```
Aniruddh25, line 225: Is there no datetime2 equivalent for Postgres? Also, how about `rowversion` ?

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -261,6 +273,67 @@ public async Task TestCleanup()
...
+        public async Task InsertOneWithTimestampFieldInRequestBody()
+        {
+            // Validate that a BadRequest exception is thrown when a timestamp field (here 'row_version') is included in request body.
+            string requestBody = @"
+            {
+                ""id"": 2,
+                ""book_name"": ""Another Awesome Book"",
+                ""copies_sold"": 100,
+                ""last_sold_on"": null,
+                ""row_version"": null
```
Aniruddh25, line 320: Can we test assigning a non-null valid value?

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
```diff
@@ -299,6 +299,53 @@ public virtual async Task PatchOne_Update_Test()
...
+                    primaryKeyRoute: expectedLocationHeader,
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PatchOneUpdateWithComputedFieldMissingFromRequestBody"),
+                    operationType: EntityActionOperation.UpsertIncremental,
+                    requestBody: requestBody,
+                    expectedStatusCode: HttpStatusCode.OK
+                );
+
+            // Validate successful execution of a PATCH insert when a computed field (here 'last_sold_on_update')
```
Aniruddh25, line 328:
```suggestion
            // Validate successful execution of a PATCH insert when a computed field (here 'last_sold_on_date')
```

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -190,6 +215,100 @@ public async Task PatchOneViewBadRequestTest()
...
+                    primaryKeyRoute: expectedLocationHeader,
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PatchOneUpdateWithTimestampFieldMissingFromRequestBody"),
+                    operationType: EntityActionOperation.UpsertIncremental,
+                    requestBody: requestBody,
+                    expectedStatusCode: HttpStatusCode.OK
+                );
+
+            // Validate successful execution of a PATCH insert when a timestamp field (here 'row_version')
```
Aniruddh25, line 244: Here it could be `last_sold_on_date` too right?

src/Service.Tests/SqlTests/SqlTestHelper.cs
```diff
@@ -96,6 +96,26 @@ public static void TestForErrorInGraphQLResponse(string response, string message
...
+        /// <summary>
+        /// Helper method to preprocess response to unescape unicode characters and remove all carriage returns/new lines.
+        /// </summary>
+        /// <param name="response">Raw response received from the API.</param>
+        /// <returns>Processed response containing unescaped unicode characters without new lines/carriage returns.</returns>
+        private static string PreprocessResponse(string response)
+        {
+            // Quote(") has to be treated differently than other unicode characters
+            // as it has to be added with a preceding backslash.
+            response = Regex.Replace(response, @"\\u0022", @"\\""");
```
Aniruddh25, line 108: Better to use StringBuilder when doing replace performance wise.

src/Service.Tests/SqlTests/SqlTestHelper.cs
```diff
@@ -181,16 +201,7 @@ public static void TestForErrorInGraphQLResponse(string response, string message
...
-                // Quote(") has to be treated differently than other unicode characters
-                // as it has to be added with a preceding backslash.
-                responseBody = Regex.Replace(responseBody, @"\\u0022", @"\\""");
-
-                // Remove all carriage returns and new lines from the response body.
-                responseBody = Regex.Replace(responseBody, @"\\n|\\r", "");
-
-                // Convert the escaped characters into their unescaped form.
-                responseBody = Regex.Unescape(responseBody);
+                responseBody = PreprocessResponse(responseBody);
```
Aniruddh25, line 204: Any reason for moving this into a function now? Its good, but just curious...

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1001,6 +1003,43 @@ private void InitODataParser()
             PopulateColumnDefinitionWithHasDefault(
                 sourceDefinition,
                 columnsInTable);
+
+            if (GetDatabaseType() is not DatabaseType.MySQL && entity is not null && entity.Source.Type is EntitySourceType.Table)
```
Aniruddh25, line 1007: Is this restricted only for tables? Cant we infer this for views?

Aniruddh25 (changes requested):
Few questions and suggestions...

Aniruddh25 (changes requested):
Few questions and suggestions...

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -256,5 +256,16 @@ public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName)
...
             return query;
         }
+
+        /// <inheritdoc/>
+        public string BuildQueryToGetReadOnlyColumns()
+        {
+            string query = "select ifsc.column_name from sys.columns as sc, information_schema.columns as ifsc " +
+                "where (sc.is_computed = 1 or ifsc.data_type = 'timestamp') " +
+                "and sc.object_id = object_id(@param0+'.'+@param1) and ifsc.table_name = @param1 " +
+                "and ifsc.table_schema = @param0 and ifsc.column_name = sc.name;";
```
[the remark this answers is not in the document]
ayush3797: sure, makes sense.

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
ayush3797: thanks (this was really silly!)

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
ayush3797: Yes, we have a test for last_sold_on_date in PatchApiTestBase as it is a computed column and computed columns are read-only both for Pg and MsSql. Thus, that's a common test.

src/Service.Tests/SqlTests/SqlTestHelper.cs
ayush3797: We cannot do regex replace on StringBuilders.

src/Service.Tests/SqlTests/SqlTestHelper.cs
ayush3797: `Currently GQL does not have any tests (but REST has) which needs this replacement logic on the exception message. In future, GQL can also utilize this method and looks cleaner now.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: No, for views the metadata is not populated correctly.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -985,6 +986,7 @@ private void InitODataParser()
                 {
                     IsNullable = (bool)columnInfoFromAdapter["AllowDBNull"],
                     IsAutoGenerated = (bool)columnInfoFromAdapter["IsAutoIncrement"],
+                    IsReadOnly = (bool)columnInfoFromAdapter["IsAutoIncrement"],
```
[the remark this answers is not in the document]
ayush3797: IsAutoIncrement is just a special case of read only. Computed/timestamp being other ones. There is no specific mapping for read-only columns but if the column is auto-gen, then we can straight away start with IsReadOnly  = IsAutoIncrement = true.

src/Core/Services/RequestValidator.cs
```diff
@@ -303,6 +303,14 @@ public static void ValidatePrimaryKeyRouteAndQueryStringInURL(EntityActionOperat
                     continue;
                 }
 
+                if (insertRequestCtx.FieldValuePairsInBody.ContainsKey(exposedName!) && sourceDefinition.Columns[column.Key].IsReadOnly)
```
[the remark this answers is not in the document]
ayush3797: yeah same thing. Updated.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1001,6 +1003,43 @@ private void InitODataParser()
...
+            Dictionary<string, DbConnectionParam> parameters = new()
+            {
+                { $"{BaseQueryStructure.PARAM_NAME_PREFIX}param0", new(schemaName, DbType.String) },
+                { $"{BaseQueryStructure.PARAM_NAME_PREFIX}param1", new(tableName, DbType.String) }
+            };
+
+            JsonArray? resultArray = await QueryExecutor.ExecuteQueryAsync(
+                sqltext: queryToGetReadOnlyColumns,
+                parameters: parameters!,
+                dataReaderHandler: QueryExecutor.GetJsonArrayAsync);
```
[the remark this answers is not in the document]
ayush3797: I think the approach you suggested would be better. Won't require json parsing. Thanks, I like this suggestion and would implement that.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: For a view like this:

```tsql
-- Create base table
CREATE TABLE Employees (
    EmployeeID INT PRIMARY KEY,
    FirstName NVARCHAR(50),
    LastName NVARCHAR(50)
);

-- Insert sample data
INSERT INTO Employees (EmployeeID, FirstName, LastName)
VALUES (1, 'John', 'Doe'),
       (2, 'Jane', 'Smith');

-- Create view
CREATE VIEW EmployeeView AS
SELECT
    EmployeeID,
    FirstName,
    LastName,
    CONCAT(FirstName, ' ', LastName) AS FullName
FROM
    Employees;
```

ayush3797: Where you would expect EmployeeView.FullName to be a computed column, the metadata comes out to be as:
image

src/Service.Tests/DatabaseSchema-PostgreSql.sql
ayush3797: This here: https://www.postgresql.org/docs/9.1/datatype-datetime.html suggests the equivalent of datatime is timestamp(n).There might be. Although here, we are not actually concerned with the datatype. It can be time/integer/string etc. We are concerned only with a column computed based on another column.

RowVersion or an equivalent is not present in PgSql.

src/Service.Tests/DatabaseSchema-MsSql.sql
ayush3797: `time` relates to time. `timestamp` is a column related to versioning.
https://learn.microsoft.com/en-us/sql/relational-databases/xml/for-xml-support-for-the-timestamp-data-type?view=sql-server-ver16

ayush3797: From the above article:

`In the FOR XML transformation, timestamp type values are treated as varbinary(8) data and will always be base 64 encoded. The XSD or XDR schema, if requested, reflects this type.`

src/Service.Tests/DatabaseSchema-MsSql.sql
ayush3797: @Aniruddh25 Even though that table can ideally be used, because of this bug: https://github.com/Azure/data-api-builder/pull/1636, we cannot use that right now unless we wait for that PR to be merged first.

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -2478,6 +2478,39 @@
...
+      },
+      "rest": {
+        "enabled": true,
+        "methods": [
+          "get",
+          "post",
+          "put",
+          "patch",
+          "delete"
+        ]
```
seantleonard, line 2501: without this, does startup fail? I know Shyam has a PR out to fix this, but if it doesn't break, you should not include the methods section for a table.

src/Service.Tests/dab-config.PostgreSql.json
```diff
@@ -2037,6 +2037,39 @@
...
+      },
+      "rest": {
+        "enabled": true,
+        "methods": [
+          "get",
+          "post",
+          "put",
+          "patch",
+          "delete"
+        ]
```
seantleonard, line 2060: same comment from MSSQL about not including this section 'methods' if it doesn't cause issues.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -261,6 +273,67 @@ public async Task TestCleanup()
 
         #endregion
 
+        /// <summary>
+        /// Test to validate successful execution of a request when a timestamp field is missing from the request body.
```
seantleonard, line 277: nit: i don't think the focus of the comment and testname should be on "timestamp". The current naming implies that there is something special about the timestamp field, when the test should be validating the behavior of readonly fields. 

Comment should describe why the missing field is significant and the effect you expect the missing field to have on DAB when Dab evaluates the request.

seantleonard: The field 'row_version' doesn't look to be auto-generated/calculated. So including this field would be fine too.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
seantleonard: The example you provide in the .sql file for timestamp https://learn.microsoft.com/en-us/sql/relational-databases/xml/for-xml-support-for-the-timestamp-data-type?view=sql-server-ver16 shows an insert using value null for the timestamp field. so why it would not be okay here.. i don't understand.

src/Service.Tests/SqlTests/RestApiTests/Insert/MySqlInsertApiTests.cs
```diff
@@ -182,6 +182,20 @@ public void InsertOneInViewBadRequestTest()
             throw new NotImplementedException();
         }
 
+        [TestMethod]
+        [Ignore]
+        public override Task InsertOneWithComputedFieldMissingInRequestBody()
```
seantleonard, line 187: Why is mysql excluded?

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
```diff
@@ -299,6 +299,53 @@ public virtual async Task PatchOne_Update_Test()
                 );
         }
 
+        /// <summary>
+        /// Test to validate successful execution of a request when a computed field is missing from the request body.
+        /// </summary>
+        [TestMethod]
+        public virtual async Task PatchOneWithComputedFieldMissingFromRequestBody()
+        {
+            // Validate successful execution of a PATCH update when a computed field (here 'last_sold_on_update')
```
seantleonard, line 308:
```suggestion
            // Validate successful execution of a PATCH update when a computed field (here 'last_sold_on_date')
```

seantleonard (changes requested):

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -807,35 +807,45 @@ public string ConstructPrimaryKeyRoute(RestRequestContext context, Dictionary<st
...
+        /// Helper method to populate all the params from the Rest request's primary key into the paramaters dictionary.
+        /// An entry is added only for those parameters which actually map to a backing column in the table/view.
+        /// </summary>
+        /// <param name="parameters">Parameters dictionary to be populated.</param>
+        /// <param name="context">Rest request's context</param>
+        private static void PopulateParamsFromRequest(Dictionary<string, object?> parameters, Dictionary<string, object?> fieldValuePairs)
+        {
+            foreach ((string field, object? value) in fieldValuePairs)
+            {
+                parameters.Add(field, value);
```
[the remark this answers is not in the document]
ayush3797: Will do this in a different PR (where duplicates are actually possible - the PR for allowing extraneous fields).

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -256,5 +256,16 @@ public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName)
...
             return query;
         }
+
+        /// <inheritdoc/>
+        public string BuildQueryToGetReadOnlyColumns(string schemaParamName, string tableParamName)
+        {
+            string query = "SELECT ifsc.column_name from sys.columns as sc INNER JOIN information_schema.columns as ifsc " +
+                "ON (sc.is_computed = 1 or ifsc.data_type = 'timestamp') " +
+                $"AND sc.object_id = object_id({schemaParamName}+'.'+{tableParamName}) and ifsc.table_name = {tableParamName} " +
+                $"AND ifsc.table_schema = {schemaParamName} and ifsc.column_name = sc.name;";
```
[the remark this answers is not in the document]
ayush3797: yes

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1001,6 +1003,49 @@ private void InitODataParser()
             PopulateColumnDefinitionWithHasDefault(
                 sourceDefinition,
                 columnsInTable);
+
+            if (GetDatabaseType() is not DatabaseType.MySQL && entity is not null && entity.Source.Type is EntitySourceType.Table)
+            {
```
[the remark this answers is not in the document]
ayush3797: https://github.com/Azure/data-api-builder/issues/1583#issuecomment-1643450629

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
ayush3797: row_version is a timetstamp field which is read-only is MsSql. I have added details in the PR desc and the comment here in the linked issue for MsSql: https://github.com/Azure/data-api-builder/issues/1453#issuecomment-1637775760

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -62,7 +62,7 @@ public static class SchemaConverter
                     directives.Add(new DirectiveNode(PrimaryKeyDirectiveType.DirectiveName, new ArgumentNode("databaseType", column.SystemType.Name)));
                 }
 
-                if (databaseObject.SourceType is not EntitySourceType.StoredProcedure && column.IsAutoGenerated)
+                if (databaseObject.SourceType is not EntitySourceType.StoredProcedure && column.IsReadOnly)
```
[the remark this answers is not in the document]
ayush3797: column.IsAutoGenerated is a subcase of column.IsReadOnly

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
ayush3797: image
We can't insert explicit values into timetstamp columns.

image
The example you linked seems to work but the value: `null` does not take effect even in that case.

src/Service.Tests/SqlTests/RestApiTests/Insert/MySqlInsertApiTests.cs
ayush3797: Because we don't collect metadata about read only fields for MySql. And hence the term ComputedFieldMissing/included does not make sense right now.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1001,6 +1003,49 @@ private void InitODataParser()
...
+                return;
+            }
+
+            foreach (string readOnlyField in readOnlyFields)
+            {
+                if (sourceDefinition.Columns.TryGetValue(readOnlyField, out ColumnDefinition? columnDefinition))
+                {
+                    // Mark the column as read-only.
+                    columnDefinition.IsReadOnly = true;
```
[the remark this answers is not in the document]
ayush3797: Yes the first one is possible. For a computed field, isautogen = false but isreadonly=true. Same goes for a timestamp field in MsSql.

src/Core/Services/RequestValidator.cs
```diff
@@ -369,6 +377,14 @@ public static void ValidatePrimaryKeyRouteAndQueryStringInURL(EntityActionOperat
                     continue;
                 }
 
+                if (upsertRequestCtx.FieldValuePairsInBody.ContainsKey(exposedName!) && column.Value.IsReadOnly)
+                {
+                    throw new DataApiBuilderException(
+                            message: $"Field '{exposedName}' provided in request body cannot be assigned a value.",
```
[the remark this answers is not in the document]
ayush3797: No, the customer should not be including the field in the request. 
`null` is still a value? If it causes confusion, I can change the messaging.
Yes, it will be affected by the request-body-strict PR.  This PR goes in first and then the request-body-PR has to be accordingly accounted for.

src/Core/Resolvers/PostgresQueryBuilder.cs
```diff
@@ -224,5 +224,13 @@ public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName)
         {
             throw new NotImplementedException();
         }
+
+        /// <inheritdoc/>
+        public string BuildQueryToGetReadOnlyColumns(string schemaParamName, string tableParamName)
+        {
+            string query = "SELECT attname AS column_name FROM pg_attribute " +
```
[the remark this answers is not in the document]
ayush3797: as far as I checked, even a newly created user is able to execute this query and get same results..

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
seantleonard: > The timestamp syntax is deprecated. This feature will be removed in a future version of Microsoft SQL Server. Avoid using this feature in new development work, and plan to modify applications that currently use this feature.
[MsLearn](https://learn.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver16#:~:text=The%20timestamp%20syntax%20is%20deprecated.%20This%20feature%20will%20be%20removed%20in%20a%20future%20version%20of%20Microsoft%20SQL%20Server.%20Avoid%20using%20this%20feature%20in%20new%20development%20work%2C%20and%20plan%20to%20modify%20applications%20that%20currently%20use%20this%20feature.)

Should not use `rowversion` for this pr to avoid introducing deprecated features.

src/Service.Tests/DatabaseSchema-MsSql.sql
seantleonard: nit: the linked article is specific to using `FOR XML` transformation. For explicit details about how `timestamp` is treated, you can reference the `rowversion` documentation. https://learn.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver16

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
```diff
@@ -299,6 +299,53 @@ public virtual async Task PatchOne_Update_Test()
...
+                    expectedStatusCode: HttpStatusCode.OK
+                );
+
+            // Validate successful execution of a PATCH insert when a computed field (here 'last_sold_on_date')
+            // is missing from the request body.
+            requestBody = @"
+            {
+                ""book_name"": ""New book"",
+                ""copies_sold"": 50
+            }";
```
seantleonard, line 334: What's the significance of including/excluding `last_sold_on` in the PATCH update versus insert?

`last_sold_on_date` is the auto generated column.

Should be described in the test summary.

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
```diff
@@ -412,6 +412,53 @@ public virtual async Task PutOne_Nulled_Test()
...
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PutOneUpdateWithComputedFieldMissingFromRequestBody"),
+                    operationType: EntityActionOperation.Upsert,
+                    requestBody: requestBody,
+                    expectedStatusCode: HttpStatusCode.OK
+                );
+
+            // Validate successful execution of a PUT insert when a computed field (here 'last_sold_on_update')
+            // is missing from the request body.
```
seantleonard, line 442: how are you validating that DAB doesn't attempt to null out the computed field `last_sold_on_update` when that column does not have a default value and is nulled to begin with? (essentially, Ani's comment on another file).

src/Core/Services/RequestValidator.cs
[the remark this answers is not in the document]
seantleonard: Yes because `null` can be used to change a column's value. 

The messaging should reflect how a user can fix the request. Perhaps mentioning that the column should not be included in this request.

Aniruddh25 (changes requested):
Still few unresolved conversations from previous review, and a few more

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
ayush3797: The screenshot is from main branch. On my branch, it will be 400 yes.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1001,6 +1003,52 @@ private void InitODataParser()
             PopulateColumnDefinitionWithHasDefault(
                 sourceDefinition,
                 columnsInTable);
+
+            if (GetDatabaseType() is not DatabaseType.MySQL && entity is not null && entity.Source.Type is EntitySourceType.Table)
+            {
+                // For MySql, the way an update/insert query is generated, we cannot skip updating a read-only field without
+                // being unable to return it in the result. So, if we want to skip updating the field, then we cannot return it in the
+                // result. Hence, skipping this for MySql.
+                // This metadata is only populated for tables because the tables being queries to collect the metadata provide
```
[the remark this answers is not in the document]
ayush3797: Added the required changes to MySqlQueryBuilder. Will add tests post merging of the datetime pr.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1398,6 +1446,34 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+        {
+            // Extract all the rows in the current Result Set of DbDataReader.
+            DbResultSet readOnlyFieldRowsWithProperties =
+                await QueryExecutor.ExtractResultSetFromDbDataReader(reader);
+
+            List<string> readOnlyFields = new();
+
+            foreach (DbResultSetRow readOnlyFieldRowWithProperties in readOnlyFieldRowsWithProperties.Rows)
+            {
+                Dictionary<string, object?> foreignKeyInfo = readOnlyFieldRowWithProperties.Columns;
```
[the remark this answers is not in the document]
ayush3797: done

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -99,9 +99,9 @@ public abstract class BaseSqlQueryStructure : BaseQueryStructure
             //result of adding (SourceDefinition.Columns - MutationFields) to UpdateOperations
             foreach (string leftoverColumn in leftoverSchemaColumns)
             {
-                // If the left over column is autogenerated
+                // If the left over column is a read-only column,
                 // then no need to add it with a null value.
-                if (sourceDefinition.Columns[leftoverColumn].IsAutoGenerated)
```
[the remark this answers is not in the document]
ayush3797: No, it does not cover all the cases. Some eg.:

https://github.com/Azure/data-api-builder/blob/96c880d16685231cfc45d53c45cbff80bea827d4/src/Core/Resolvers/MySqlQueryBuilder.cs#L268

https://github.com/Azure/data-api-builder/blob/5e3874e6a77179523d270a0358efdcbbea9762fa/src/Core/Resolvers/MySqlQueryBuilder.cs#L296

where LAST_INSERT_ID() should be used only for an auto-gen column.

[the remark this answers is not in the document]
ayush3797: And there are use cases in other PRs as well like the PR for mutations with triggers.

src/Service.Tests/DatabaseSchema-MsSql.sql
ayush3797: Proceeding with this table only since the PR: https://github.com/Azure/data-api-builder/pull/1636 is still left with some decision making first.

src/Core/Services/RequestValidator.cs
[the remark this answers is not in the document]
ayush3797: Updated.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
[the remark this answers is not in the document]
ayush3797: Essentially, just like autogen fields, read-only fields cannot be included in the fields to be mutated in the mutation.

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
ayush3797: I have updated the column `last_sold_on` to have a default value. So the values for both `last_sold_on` and `last_sold_on_date` would be non-null after insertion.

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
ayush3797: There is no difference. Just validating the changes for insert/update via PATCH. Updated the comment and the request body for the update test.

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -2443,6 +2443,33 @@
...
+      "graphql": {
+        "enabled": true,
+        "type": {
+          "singular": "books_sold",
+          "plural": "books_solds"
+        }
+      },
+      "rest": {
+        "enabled": true,
+        "methods": []
```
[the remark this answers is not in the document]
ayush3797: nope removed.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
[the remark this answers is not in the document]
ayush3797: Yes. A computed column or a timestamp/rowversion column is not autogenerated but is read only.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
ayush3797: Added test for non-null value. For Sean's comment: https://github.com/Azure/data-api-builder/pull/1596#discussion_r1300331376.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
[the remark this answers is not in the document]
ayush3797: https://github.com/Azure/data-api-builder/pull/1596#discussion_r1308230767

src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -179,18 +177,25 @@ public override string BuildForeignKeyInfoQuery(int numberOfParameters)
...
-            string sets = String.Join(";\n", columns.Select((x, index) => $"SET {"@LU_" + index.ToString()} := 0"));
+            SourceDefinition sourceDefinition = structure.GetUnderlyingSourceDefinition();
+            List<string> columns = structure.AllColumns();
 
-            // Fetch the value to local variables
-            string updates = String.Join(", ", columns.Select((x, index) =>
-                $"{QuoteIdentifier(x)} = (SELECT {"@LU_" + index.ToString()} := {QuoteIdentifier(x)})"));
+            // Create local variables to store the updatable columns.
+            string sets = String.Join(";\n",
+                columns.Where(col => !sourceDefinition.Columns[col].IsReadOnly || sourceDefinition.Columns[col].IsAutoGenerated).
```
seantleonard, line 187: `.` should be on the next line -> `.Select()`

src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -179,18 +177,25 @@ public override string BuildForeignKeyInfoQuery(int numberOfParameters)
         /// <summary>
         /// Makes the query segments to store PK during an update
         /// </summary>
-        private (string, string, string) MakeStoreUpdatePK(List<string> columns,
-                                                           List<LabelledColumn> outputColumns)
+        private (string, string, string) MakeStoreUpdatePK(BaseSqlQueryStructure structure, List<LabelledColumn> outputColumns)
```
seantleonard, line 180: Comment needs to be updated to include description of what the return tuple represents.

src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -323,6 +333,14 @@ private static string GetMySQLDefaultValue(ColumnDefinition column)
             return defaultValue;
         }
 
+        /// <inheritdoc/>
+        public string BuildQueryToGetReadOnlyColumns(string schemaParamName, string tableParamName)
+        {
+            string query = "select column_name as column_name from information_schema.columns " +
+                $"where table_schema = {schemaParamName} and table_name = {tableParamName} and generation_expression!= '';";
```
seantleonard, line 340:
```suggestion
                $"where table_schema = {schemaParamName} and table_name = {tableParamName} and generation_expression != '';";
```

src/Core/Resolvers/MySqlQueryBuilder.cs
seantleonard: Comment should also address:
- why this is necessary for MySQL only
- example of a built segment for sets, updates, select

src/Core/Services/MetadataProviders/MySqlMetadataProvider.cs
```diff
@@ -26,6 +27,15 @@ public class MySqlMetadataProvider : SqlMetadataProvider<MySqlConnection, MySqlD
...
             : base(runtimeConfigProvider, queryExecutor, sqlQueryBuilder, logger)
         {
+            try
+            {
+                using MySqlConnection conn = new(ConnectionString);
+                _databaseName = conn.Database;
+            }
+            catch
+            {
+                logger.LogWarning("Could not determine database name from the connection string. Defaulting it to 'mysql'.");
```
seantleonard, line 37:
```suggestion
                logger.LogWarning("Could not determine database name from the connection string. The default database name 'mysql' will be used.");
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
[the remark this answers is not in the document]
seantleonard: > A computed column or a timestamp/rowversion column is not autogenerated but is read only.

Such a distiction is artificially made by DAB.. right? SQL Server metadata will be provided by `columnInfoFromAdapter["IsAutoIncrement"],` which is used for both `IsReadOnly` and `IsAutoGenerated`. I don't see how there could ever be a circumstance where `IsReadOnly` and `IsAutoGenerated` will be different.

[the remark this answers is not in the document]
seantleonard: Perhaps a comment in the code to note that the `IsReadOnly` field will be overridden later by `PopulateColumnDefinitionsWithReadOnlyFlag()` and is only set here because `IsReadOnly` is not nullable.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -256,5 +256,16 @@ public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName)
                             "WHERE is_hidden is not NULL AND is_hidden = 0";
             return query;
         }
+
+        /// <inheritdoc/>
```
seantleonard, line 260: The implementation here is specific to SQL Server. Instead of `<inheritdoc/>`, describe what is considered a read-only column in sql server: `timestamp` / `is_computed`. And noting (If i recall from a previous screenshot you provided) the `timestamp` field has the trait `is_computed=0`.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1398,6 +1448,33 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
             return pairToFkDefinition;
         }
 
+        /// <summary>
+        /// Helper method to get all the read-only fields name in a table by processing the DbDataReader instance
+        /// which contains the name of all the fields - one field per DbResult row.
+        /// </summary>
+        /// <param name="reader">The DbDataReader.</param>
+        /// <param name="args">Arguments to this function. This parameter is unused in this method.
+        /// This is added so that the method conforms with the Fucnc delegate's signature.</param>
```
seantleonard, line 1457:
```suggestion
        /// This is added so that the method conforms with the Func delegate's signature.</param>
```

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -460,6 +471,8 @@ SET IDENTITY_INSERT authors_history OFF
 INSERT INTO revenues(id, category, revenue, accessible_role) VALUES (1, 'Book', 5000, 'Anonymous'), (2, 'Comics', 10000, 'Anonymous'),
 (3, 'Journals', 20000, 'Authenticated'), (4, 'Series', 40000, 'Authenticated');
 
+INSERT INTO books_sold(id,book_name,last_sold_on) values(1,'Awesome Book',GETDATE());
```
seantleonard, line 474: When using `GETDATE()` how do you distinguish whether a default value was used? e.g. the default is `CURRENT_TIMESTAMP`. Wouldn't the values be the same?

src/Service.Tests/DatabaseSchema-PostgreSql.sql
```diff
@@ -336,6 +345,7 @@ INSERT INTO journals(id, journalname, color, ownername) VALUES (1, 'Journal1', '
 
 INSERT INTO aow("NoteNum", "DetailAssessmentAndPlanning", "WagingWar", "StrategicAttack") VALUES (1, 'chapter one notes: ', 'chapter two notes: ', 'chapter three notes: ');
 INSERT INTO sales(id, item_name, subtotal, tax) VALUES (1, 'Watch', 249.00, 20.59), (2, 'Montior', 120.50, 11.12);
+INSERT INTO books_sold(id,book_name,last_sold_on) values(1, 'Awesome Book', CURRENT_DATE);
```
seantleonard, line 348: Should MS/MY/PG sql all be consisten and use a literal date versus a shortcut like `CURRENT_DATE`?

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -2443,6 +2443,33 @@
...
+      },
+      "permissions": [
+        {
+          "role": "anonymous",
+          "actions": [
+            {
+              "action": "*"
```
seantleonard, line 2467: This doesn't look right. If you want all actions, it is a one liner: `"actions": ["*"]`

src/Core/Resolvers/MySqlQueryBuilder.cs
Aniruddh25: Function name should be made more intuitive e.g. `MakeQuerySegmentsForUpdate`

src/Core/Resolvers/IQueryBuilder.cs
```diff
@@ -64,6 +64,15 @@ public interface IQueryBuilder
...
         public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName);
 
+        /// <summary>
+        /// Returns the query to get the read-only columns present in the table.
+        /// </summary>
+        /// <param name="schemaOrDatabaseParamName">Param name of the schema/database.</param>
+        /// <param name="tableParamName">Param name of the table.</param>
+        /// <exception cref="NotImplementedException">Thrown when class implementing the interface has not provided
+        /// an overridden implementation of the method.</exception>
+        string BuildQueryToGetReadOnlyColumns(string schemaOrDatabaseParamName, string tableParamName);
```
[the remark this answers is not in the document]
ayush3797: I just stuck to the placements that we have now. Even for the queries that collect metadata, they are placed in {*}QueryBuilder classes. And yeah I think QueryBuilder is still a better option, since it deals with constructing the query. MetadataProviders should be providing metadata and should not have to deal with query generation/construction.

src/Core/Resolvers/MySqlQueryBuilder.cs
ayush3797: Yep, I'll make those additional changes.

ayush3797: > why this is necessary for MySQL only

Do you mean why this function is called only for MySql or why are we not including the read-only column in the set/update/select clause? I am assuming the latter.

src/Service.Tests/dab-config.MsSql.json
ayush3797: I think this is how the config is generated using the config generating command. I just copy pasted from what it actually looks like when config is generated. And that's why you'll see there are other instances of this same format within this same file.

src/Service.Tests/DatabaseSchema-MsSql.sql
ayush3797: Yes, correct. However, I don't think we are concerned with distinguishing between the values. I update the default value to constant datetime.

src/Service.Tests/DatabaseSchema-PostgreSql.sql
ayush3797: yeah, makes sense. updated.

src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -177,20 +175,47 @@ public override string BuildForeignKeyInfoQuery(int numberOfParameters)
...
-            string sets = String.Join(";\n", columns.Select((x, index) => $"SET {"@LU_" + index.ToString()} := 0"));
-
-            // Fetch the value to local variables
-            string updates = String.Join(", ", columns.Select((x, index) =>
-                $"{QuoteIdentifier(x)} = (SELECT {"@LU_" + index.ToString()} := {QuoteIdentifier(x)})"));
-
-            // Select local variables and mapping to original column name
-            string select = String.Join(", ", outputColumns.Select((x, index) => $"{"@LU_" + index.ToString()} AS {QuoteIdentifier(x.Label)}"));
+            SourceDefinition sourceDefinition = structure.GetUnderlyingSourceDefinition();
+            List<string> columns = structure.AllColumns();
```
seantleonard, line 192: This may be an issue where you are resolving all columns for the subsequent SELECT. The requestor may not have read permissions for all columns or the requestor may have listed only a subset of columns in the GraphQL selection set or REST query string ($select).

src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -177,20 +175,47 @@ public override string BuildForeignKeyInfoQuery(int numberOfParameters)
...
+
+            // Create local variables to store the updatable columns.
+            string sets = String.Join(";\n",
+                columns.Where(col => !sourceDefinition.Columns[col].IsReadOnly || sourceDefinition.Columns[col].IsAutoGenerated)
+                .Select((x, index) => $"SET {"@LU_" + index.ToString()} := 0"));
```
seantleonard, line 197: Can you use a more descriptive name than `x`?

src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -323,6 +353,14 @@ private static string GetMySQLDefaultValue(ColumnDefinition column)
             return defaultValue;
         }
 
+        /// <inheritdoc/>
+        public string BuildQueryToGetReadOnlyColumns(string schemaParamName, string tableParamName)
+        {
+            string query = "select column_name as column_name from information_schema.columns " +
```
seantleonard, line 359:
```suggestion
            string query = "select column_name from information_schema.columns " +
```

Is `column_name as column_name` necessary since you aren't changing the name of the returned column?

src/Service.Tests/DatabaseSchema-PostgreSql.sql
seantleonard: *updated for table creation when setting default value for last_sold_on

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
```diff
@@ -412,6 +412,53 @@ public virtual async Task PutOne_Nulled_Test()
...
+            await SetupAndRunRestApiTest(
+                    primaryKeyRoute: expectedLocationHeader,
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PutOneUpdateWithComputedFieldMissingFromRequestBody"),
+                    operationType: EntityActionOperation.Upsert,
+                    requestBody: requestBody,
+                    expectedStatusCode: HttpStatusCode.OK
+                );
+
```
seantleonard, line 440: The query `PutOneUpdateWithComputedFieldMissingFromRequestBody` doesn't actually verify that the computed field `last_sold_on_update` isn't touched (Nulled out because of exclusion from request body/ PUT semaintics.)

In your PR description, you state  _"When a read-only field is excluded from the request body for mutation, we won't NULL out the field's value but let the responsibility on the database to deal with it"_. When you say _"let the responsibility on the database to deal with it"_, is that what is happening here? the database would return an error if DAB erroneously included the computed field `last_sold_on_update` in the update query (even when user excludes the field from the GraphQL/REST body)?

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
```diff
@@ -299,6 +299,54 @@ public virtual async Task PatchOne_Update_Test()
                 );
         }
 
+        /// <summary>
+        /// Test to validate successful execution of a request when a computed field is missing from the request body.
+        /// </summary>
```
seantleonard, line 304: Can you add context to _how_ successful execution validates what you are testing? A `PATCH` request shouldn't touch fields not listed in the request body by definition. Your validation SQL query also does not query the `last_sold_on_date` field to check for value changes. 

An explanation like 'this test would fail when DAB includes the `last_sold_on_date` field in the PATCH request, resulting in an exception being returned by the database due to an attempt to modify a read-only field.'

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
```diff
@@ -299,6 +299,54 @@ public virtual async Task PatchOne_Update_Test()
...
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PatchOneUpdateWithComputedFieldMissingFromRequestBody"),
+                    operationType: EntityActionOperation.UpsertIncremental,
+                    requestBody: requestBody,
+                    expectedStatusCode: HttpStatusCode.OK
+                );
+
+            // Validate successful execution of a PATCH insert when a computed field (here 'last_sold_on_date')
+            // is missing from the request body. In such a case we don't try to put NULL value for such a field while performing
```
seantleonard, line 329: _In such a case we don't try to put NULL value for such a field while performing_ 

The sql query doesn't validate the state of the `last_sold_on_date` column in this test. A request resulting in a query like:
```sql
 UPDATE books_sold
  SET last_sold_on = null where id = 1
```
both the `last_sold_on` and `last_sold_on_date` would be updated to null.

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -190,6 +215,100 @@ public async Task PatchOneViewBadRequestTest()
             await base.PatchOneViewBadRequestTest(expectedErrorMessage);
         }
 
+        /// <summary>
+        /// Test to validate successful execution of a request when a rowversion field is missing from the request body.
```
seantleonard, line 219: _successful execution_: no error thrown because `rowversion` was excluded from the request, right?

src/Service.Tests/SqlTests/RestApiTests/Insert/InsertApiTestBase.cs
```diff
@@ -87,6 +87,35 @@ public virtual async Task InsertOneWithByteArrayTypeAsNull()
...
         }
 
+        /// <summary>
+        /// Test to validate successful execution of a request when a computed field is missing from the request body.
+        /// In such a case, we skip inserting the field. 
+        /// </summary>
+        [TestMethod]
+        public virtual async Task InsertOneWithComputedFieldMissingInRequestBody()
+        {
+            // Validate successful execution of a POST request when a computed field (here 'last_sold_on_update')
```
seantleonard, line 97:
```suggestion
            // Validate successful execution of a POST request when a computed field (here 'last_sold_on_date')
```

seantleonard (changes requested):
Clarification on how tests validate behavior -> would be good to be explicit in comments across tests. 

Question about how post mutation select uses `TableDefinition.AllColumns()` which may result in not honoring read permissions or may not honor GraphQL selection set fields / REST request querystring `$select` fields.

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -253,7 +254,17 @@ CREATE TABLE mappedbookmarks
 (
 	id int IDENTITY(1,1) PRIMARY KEY,
 	bkname nvarchar(50) NOT NULL
-) 
+)
+
+create table books_sold
+(
+  [id] int PRIMARY KEY not null,
```
Aniruddh25, line 261: nit: 4 spaces at the beginning

src/Service.Tests/dab-config.MsSql.json
Aniruddh25: Yes, this is how the new config system creates it. The one line is accepted as well while deserializing but when we serialize this is the new way to write it.

Aniruddh25:
> Clarification on how tests validate behavior -> would be good to be explicit in comments across tests. 

Question about how post mutation select uses `TableDefinition.AllColumns()` which may result in not honoring read permissions or may not honor GraphQL selection set fields / REST request querystring `$select` fields. 

I think Shyam's PR should handle that separately.

Aniruddh25 (approved):
LGTM, few nits. Thanks!

src/Core/Resolvers/MySqlQueryBuilder.cs
ayush3797: Not a new change that is introduced in this PR. As Ani mentioned: https://github.com/Azure/data-api-builder/pull/1596#issuecomment-1712358895, it is a TODO with @severussundar.

src/Core/Resolvers/MySqlQueryBuilder.cs
ayush3797: Yes, its needed. By default, with no alias, the name is captialized as `COLUMN_NAME`. So, I aliased it to `column_name`.

src/Core/Resolvers/MySqlQueryBuilder.cs
ayush3797: Clarified and added example.

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
ayush3797: > When you say "let the responsibility on the database to deal with it", is that what is happening here? the database would return an error if DAB erroneously included the computed field last_sold_on_update in the update query (even when user excludes the field from the GraphQL/REST body)?

Correct. The thing that the PUT request executed successfully on a table with read-only field confirms that we did not attempt to update the value of the read-only field.

ayush3797: I'll add to the comments.

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
ayush3797: right, clarified in the comment added.

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
ayush3797: I updated the sql query to validate the state of last_sold_on_date and last_sold_on columns as well.

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
ayush3797: added explanation to tests.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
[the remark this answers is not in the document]
ayush3797: added

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
ayush3797: Added those for MsSql/PgSql. For MySql, I think there is a bug as to how the default values are parsed. Opened an issue here: https://github.com/Azure/data-api-builder/issues/1696.

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
```diff
@@ -412,6 +412,55 @@ public virtual async Task PutOne_Nulled_Test()
...
+        /// <summary>
+        /// Test to validate successful execution of a request when a computed field is missing from the request body.
+        /// In such a case, we don't attempt to NULL out computed field (as per PUT semantics) but instead skip updating/inserting the field. 
+        /// </summary>
+        [TestMethod]
+        public virtual async Task PutOneWithComputedFieldMissingFromRequestBody()
+        {
+            // Validate successful execution of a PUT update when a computed field (here 'last_sold_on_update')
+            // is missing from the request body. Successful execution of the PUT request confirms that we did not
+            // attempt to NULL out the 'last_sold_on_update' field.
```
seantleonard, line 424: I'm confused still. `PutOneUpdateWithComputedFieldMissingFromRequestBody` has this query which checks for a record where the read-only column is null, which contradicts your code comment:
```csharp
            {
                "PutOneUpdateWithComputedFieldMissingFromRequestBody",
                $"SELECT * FROM { _tableWithReadOnlyFields } " +
                $"WHERE [id] = 1 AND [book_name] = 'New book' AND [copies_sold] = 101 AND [last_sold_on] is NULL AND [last_sold_on_date] is NULL " +
            },
```

Shouldn't the read-only column be prepopulated with a value and the validation query should be validating that the column has that value (and not NULL)?

seantleonard: It looks like you have already made the adjustments for the PUT insert test which proceeds this put-update test.

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
```diff
@@ -233,6 +259,109 @@ public async Task PutOneInViewBadRequest()
...
+        /// <summary>
+        /// Test to validate successful execution of a request when a rowversion field is missing from the request body.
+        /// In such a case, we don't attempt to NULL out rowversion field (as per PUT semantics) but instead skip updating/inserting the field. 
+        /// </summary>
+        [TestMethod]
+        public async Task PutOneWithRowversionFieldMissingFromRequestBody()
+        {
+            // Validate successful execution of a PUT update when a rowversion field (here 'row_version')
+            // is missing from the request body. Successful execution of the PUT request confirms that we did not
+            // attempt to NULL out the 'row_version' field.
```
seantleonard, line 271: `PutOneUpdateWithRowversionFieldMissingFromRequestBody` does not include rowversion in the verification query.. so there's no validating whether row_version is touched.

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
```diff
@@ -233,6 +259,109 @@ public async Task PutOneInViewBadRequest()
...
+                ""last_sold_on"": null
+            }";
+
+            expectedLocationHeader = $"id/2";
+
+            await SetupAndRunRestApiTest(
+                    primaryKeyRoute: expectedLocationHeader,
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PutOneInsertWithRowversionFieldMissingFromRequestBody"),
```
seantleonard, line 306: `PutOneInsertWithRowversionFieldMissingFromRequestBody` does not include rowversion in the verification query.. so there's no validating whether row_version is touched.

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
```diff
@@ -299,6 +299,54 @@ public virtual async Task PatchOne_Update_Test()
...
+                ""book_name"": ""New book"",
+                ""copies_sold"": 50
+            }";
+            string expectedLocationHeader = $"id/1";
+
+            await SetupAndRunRestApiTest(
+                    primaryKeyRoute: expectedLocationHeader,
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PatchOneUpdateWithComputedFieldMissingFromRequestBody"),
```
seantleonard, line 322: `PatchOneUpdateWithComputedFieldMissingFromRequestBody` should be updated accordingly to validate that computed field wasn't nulled out.

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -242,7 +243,8 @@ public async Task PatchOneWithRowversionFieldMissingFromRequestBody()
                 );
 
             // Validate successful execution of a PATCH insert when a rowversion field (here 'row_version')
-            // is missing from the request body.
+            // is missing from the request body.Successful execution of the PATCH request confirms that we did not
```
seantleonard, line 246:
```suggestion
            // is missing from the request body. Successful execution of the PATCH request confirms that we did not
```

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -190,6 +216,101 @@ public async Task PatchOneViewBadRequestTest()
...
+                ""copies_sold"": 100,
+                ""last_sold_on"": null
+            }";
+            string expectedLocationHeader = $"id/1";
+
+            await SetupAndRunRestApiTest(
+                    primaryKeyRoute: expectedLocationHeader,
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PatchOneUpdateWithRowversionFieldMissingFromRequestBody"),
```
seantleonard, line 239: `PatchOneUpdateWithRowversionFieldMissingFromRequestBody` doesn't validate the value of `row_version`

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -190,6 +216,101 @@ public async Task PatchOneViewBadRequestTest()
...
+                ""last_sold_on"": null
+            }";
+
+            expectedLocationHeader = $"id/2";
+
+            await SetupAndRunRestApiTest(
+                    primaryKeyRoute: expectedLocationHeader,
+                    queryString: null,
+                    entityNameOrPath: _entityWithReadOnlyFields,
+                    sqlQuery: GetQuery("PatchOneInsertWithRowversionFieldMissingFromRequestBody"),
```
seantleonard, line 261: `PatchOneInsertWithRowversionFieldMissingFromRequestBody` doesn't check the row_version field.
The comment for this particular test is still unclear about how you're validating that the patch request did not attempt to null out `row_version` field. Stating "successful execution of X confirms" doesn't really capture what I'm asking. You'll still need to answer **why** succesful execution indicates DAB didn't attempt to null out the field. Let's say someone updates the code to actually modify the value/attempt to modify the value: how would this test fail?

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -261,6 +273,68 @@ public async Task TestCleanup()
...
+        /// <summary>
+        /// Test to validate successful execution of a request when a rowversion field is missing from the request body.
+        /// In such a case, we skip inserting the field. 
+        /// </summary>
+        [TestMethod]
+        public async Task InsertOneWithRowversionFieldMissingFromRequestBody()
+        {
+            // Validate successful execution of a POST request when a rowversion field (here 'row_version')
+            // is missing from the request body. Successful execution of the POST request confirms that we did not
+            // attempt to provide a value for the 'row_version' field.
```
seantleonard, line 285: should expand on this comment: "if `row_version` is correctly excluded from the request body, and dab internals attempt to provide a value for the `row_version` field, X error would occur"

seantleonard (commented):
Circled back to a few areas where the validation query for a test doesn't look to actually validate what you are checking for. Some tests were updated to accommodate, but there are still some leftovers to address.

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
ayush3797: > Shouldn't the read-only column be prepopulated with a value and the validation query should be validating that the column has that value (and not NULL)?

No, it should be NULL. You are referencing a PUT update which does not include last_sold_on in the request body. So that field would be nulled out by DAB. Accordingly the database will null out the last_sold_on_date field as well.

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
ayush3797: We can't know the value of row_version field beforehand. The best we can do is a NOT NULL check. Will incorporate that.

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
ayush3797: https://github.com/Azure/data-api-builder/pull/1596#discussion_r1322397055

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
ayush3797: https://github.com/Azure/data-api-builder/pull/1596#discussion_r1322397055

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
ayush3797: https://github.com/Azure/data-api-builder/pull/1596#discussion_r1322397055

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
ayush3797: We cannot provide any custom value to the last_sold_on field in the request body, because the test is common between the 3 databases, and for MySql there is this bug: https://github.com/Azure/data-api-builder/issues/1696. So, we have to provide a default value of `CURRENT_TIMESTAMP` for the `last_sold_on` field. This value would change everytime the test is re-run and the database is reset.

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
ayush3797: https://github.com/Azure/data-api-builder/pull/1596#discussion_r1322397055

src/Service.Tests/SqlTests/RestApiTests/Put/PutApiTestBase.cs
ayush3797: I updated the test to have a value supplied for `last_sold_on` field. DAB would update this field and the database would automatically update the `last_sold_on_date` field

src/Service.Tests/SqlTests/RestApiTests/Patch/PatchApiTestBase.cs
seantleonard: This `PatchOneUpdateWithComputedFieldMissingFromRequestBody` query would now need the NULL check that you added for other tests on the field: `last_sold_on_update`

seantleonard (approved):
Thank you for addressing comments and implementing this feature! I added one last commit to address a comment that is consistent with the other changes you were making for the sake of time.

Issue #1604, opened 2023-07-24
[Bug] When database name is empty it generates wrong sql query


rohkhann:
For databases like datamart that are mssql, databasename is emtpy from the connection string. 

In sql metadataprovider, during initialize async the tableprefix is creating using databasename:
            selectCommand.CommandText
                = $"SELECT * FROM {tablePrefix}.{SqlQueryBuilder.QuoteIdentifier(tableName)}";

if databasename is empty then the tableprefix generated is [].[dbo].[books] for a sample table books with dbo schema which results in sql query issue.

PR #1605, opened 2023-07-24
When databasename is empty it should not be added to prefix.

rohkhann:
## Why make this change?

  Fixes issue mentioned in : https://github.com/Azure/data-api-builder/issues/1604

## What is this change?

Added check to see if databasename is null or empty then we will not add it to the table prefix. 
## How was this tested?

- [ ] Integration Tests
existing integration tests test this path of code
- [ ] Unit Tests
Added new test called checkTablePrefix. Chose to go with internalsvisibleto route as the main goal was to test gettableprefix method and ROI on moqing out components of the public method initializeasync to do the test was low.


aaronpowell:
Are we able to add tests for this?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1150,12 +1150,24 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
...
+                // determine schemaName for prefix.
                 schemaName = SqlQueryBuilder.QuoteIdentifier(schemaName);
+            }
+
+            if (!string.IsNullOrEmpty(databaseName))
+            {
+                // determine databaseName for prefix.
+                databaseName = SqlQueryBuilder.QuoteIdentifier(databaseName);
+                tablePrefix.Append(databaseName);
                 tablePrefix.Append($".{schemaName}");
```
Aniruddh25, line 1165: if schemaName is empty and db is not empty, this will be `[db].`, as opposed to just `[db]` that was there previously...

rohkhann: Yes am making this fix and adding appropriate test cases.

rohkhann: Should be fixed now.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1150,11 +1150,27 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
 
         private string GetTablePrefix(string databaseName, string schemaName)
         {
-            StringBuilder tablePrefix = new(SqlQueryBuilder.QuoteIdentifier(databaseName));
-            if (!string.IsNullOrEmpty(schemaName))
+            StringBuilder tablePrefix = new();
+
+            if (!string.IsNullOrEmpty(databaseName))
```
seantleonard, line 1155: When both databaseName and schemaName are empty, this method will turn an empty string. The callers will append this empty string to 2 places, one of which: 
```csharp
string tablePrefix = GetTablePrefix(conn.Database, schemaName);
            selectCommand.CommandText
                = $"SELECT * FROM {tablePrefix}.{SqlQueryBuilder.QuoteIdentifier(tableName)}";
```

The references will need to be updated too.

seantleonard: I guess for the purposes of this pr, database name will always be populated.

Aniruddh25: This is a good point, what happens if both databaseName/schemaName are empty? 
the prefix will end up being empty, so either 
 1. the caller should check for emptiness of prefix before appending the `.` to {SqlQueryBuilder.QuoteIdentifier(tableName)} OR 
 2. this function should do the check and append the `.` to the tablePrefix before returning.

rohkhann: Fixed above in latest iteration.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1140,21 +1140,38 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
             };
 
             string tablePrefix = GetTablePrefix(conn.Database, schemaName);
+            string queryprefix = string.IsNullOrEmpty(tablePrefix) ? string.Empty : $"{tablePrefix}.";
```
Aniruddh25, line 1143:
```suggestion
            string queryPrefix = string.IsNullOrEmpty(tablePrefix) ? string.Empty : $"{tablePrefix}.";
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1140,21 +1140,38 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
...
             selectCommand.CommandText
-                = $"SELECT * FROM {tablePrefix}.{SqlQueryBuilder.QuoteIdentifier(tableName)}";
+                = $"SELECT * FROM {queryprefix}{SqlQueryBuilder.QuoteIdentifier(tableName)}";
```
Aniruddh25, line 1145:
```suggestion
                = $"SELECT * FROM {queryPrefix}{SqlQueryBuilder.QuoteIdentifier(tableName)}";
```

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -83,6 +85,31 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
...
+            RuntimeConfig baseConfigFromDisk = SqlTestHelper.SetupRuntimeConfig();
+            RuntimeConfigProvider runtimeConfigProvider = TestHelper.GenerateInMemoryRuntimeConfigProvider(baseConfigFromDisk);
+
+            ILogger<ISqlMetadataProvider> sqlMetadataLogger = new Mock<ILogger<ISqlMetadataProvider>>().Object;
+            Mock<IQueryExecutor> queryExecutor = new();
+            IQueryBuilder queryBuilder = new MsSqlQueryBuilder();
+
+            SqlMetadataProvider<SqlConnection, SqlDataAdapter, SqlCommand> provider = new MsSqlMetadataProvider(runtimeConfigProvider, queryExecutor.Object, queryBuilder, sqlMetadataLogger);
+            string tableprefix = provider.GetTablePrefix(databaseName, schemaName);
+            Assert.IsTrue(tableprefix.Equals(expectedPrefix));
```
Aniruddh25, line 110: nit: prefer using `Assert.AreEqual`
to help easily identify what the values are in case of failures.

Aniruddh25 (approved):
LGTM, left few nits.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -5,6 +5,7 @@
 using System.Data.Common;
 using System.Diagnostics.CodeAnalysis;
 using System.Net;
+using System.Runtime.CompilerServices;
```
Aniruddh25, line 8: Please fix PR description - remove unnecessary info from the description template, add reason why chose to do `InternalsVisibleTo`.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Do we want to go a step ahead and check for `whitespace` string as well (string.isNullOrWhiteSpace())?

ayush3797 (approved):
LGTM!

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rohkhann: fixed.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
rohkhann: fixed.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rohkhann: fixed description.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rohkhann: I dont think GetTableprefix would ever return a whitespace? databasename is taken from the connection string and then the schemaname is also validated in input.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -83,6 +85,31 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
             await CheckExceptionForBadConnectionStringHelperAsync(DatabaseEngine, connectionString);
         }
 
+        /// <summary>
+        /// <code>Do: </code> Tests with different combinations of connection string to ensure
+        /// tableprefix generated is correct.
```
seantleonard, line 90:
```suggestion
        /// tableprefix generated is correctly.
```
nit

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -83,6 +85,31 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
...
+        /// <code>Check: </code> Making sure table prefix matches expected prefix.
+        /// </summary>
+        [TestMethod]
```
seantleonard, line 93:
```suggestion
        [DataTestMethod]
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1140,21 +1140,38 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
...
+                    tablePrefix.Append($".{schemaName}");
+                }
+            }
+            else if (!string.IsNullOrEmpty(schemaName))
             {
+                // Determine schemaName for prefix.
                 schemaName = SqlQueryBuilder.QuoteIdentifier(schemaName);
-                tablePrefix.Append($".{schemaName}");
+                // Database name is empty we just need the schema name.
+                tablePrefix.Append(schemaName);
```
seantleonard, line 1174: any way to simplify this code since it's slightly duplicated except for adding the period?

seantleonard (approved):

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rohkhann: I could create a seperate function that does the tablePrefix append -> code on lines 1172 and 1165 but it seems easier to understand this way.

seantleonard (approved):

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rohkhann: Will resolve based on offline discussion.

Issue #1570, opened 2023-07-07
Need to unignore test once behavior of stored procedures is fixed


ayush3797:
### What happened?

Currently, we are ignoring this test: https://github.com/Azure/data-api-builder/blob/ca3c4448106d67bb2bca7370f38c0c30857c2095/src/Service.Tests/Unittests/ConfigValidationUnitTests.cs#L1858 because the behavior of rest.methods property is not functioning correctly. The methods property is only relevant for stored-procedures but it is also getting populated for tables/views. Will have to un-ignore the test once behavior is fixed.

ayush3797:
FYI @severussundar @Aniruddh25.

severussundar:
For tables and views, the CLI writes out all the 5 HTTP verbs in `Methods` field. 

This particular test tries to populate the methods field with just `GET` for a table (possibly to simulate the scenario of editing the config by hand?).  Even if the config is hand-edited to have only `GET`, all the 5 HTTP actions are successfully enabled which is the right behavior as per design.

The test expects an error to be thrown when deserializing a config which has `Methods` array populated for tables/views - which needs to thought through again. IMO, the current behavior - irrespective of the value present in the `Methods` field in the config file, enabling of all 5 HTTP actions is in-line with the design decision.

severussundar:
At the moment, when an entity backed by a table or view is created, CLI writes out all the 5 REST methods in the config file. Since, the `methods` property does not play any role for tables and views (When tables/views are exposed for REST, all 5 HTTP actions are enabled by design. Only for stored-procedures, there is option of configuring a set of HTTP verbs that can be enabled), it need not be written out to the config file for tables and views. [This issue](https://github.com/Azure/data-api-builder/issues/1571) tracks the effort for the same. 

At the moment, If the config file is hand-edited to add the methods property manually, it is not honored and all the 5 HTTP actions are supported (as noted in my previous comment).  

Currently, presence of extra fields does not lead to errors.
For example, For stored procedures, adding `key-fields` in the `source` section does not lead to errors during deserialization and validation. At the same time, the addition of `key-fields` does not play any role/have any effect since it is applicable only for views. Same is the case with adding `parameters` in the `source` section for tables/views. Doesn't lead to errors, at the same time has no effect as it relevant only for stroed procedures. 

A similar approach could be followed here
1. CLI will not write out the `methods` field for tables and views to the config file. 
2. CLI does not provide a way to declare `methods` for tables and views - this validation already exists.
[...]

PR #1607, opened 2023-07-25
Skip writing the HTTP actions to config file and log warning when REST Methods are configured for tables/views

### LINKED ISSUE #1571: Make the REST Methods property nullable in EntityRestOptions class
As per specs, the `Methods` property can be null. So, there is scope to make the `Methods` property of the `Rest` field nullable.

severussundar:
## Why make this change?

- Closes #1571 #1570 
- At the moment, CLI writes out all the 5 HTTP methods to the config file when adding an entity backed by table/view. 
image

- However, for tables/views, the `Methods` property does not play any role and can be skipped writing to the config file. 
  
## What is this change?

- `EntityRestOptions`: `Methods` property is made nullable. References where it was initialized with an empty array is updated to `null`.  Appropriate `null` checks are added. 
- `EntityRestOptionsConverter`: Serialization logic is updated to write to the config file only when the `Methods` within REST element is not empty 
- `RuntimeConfigValidator.ValidatePresenceOfExtraRestProperties()`: Helper method to log a warning message when `Methods` property is added manually to an entity backed by table/view.
- HTTP Actions are removed from the snapshot files.
-  `dab-config.*` files - `Methods` property is removed from the reference config files

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests
- [x] Manual Tests

## Sample Request(s)

- CLI: REST actions are not written to the config file
`dab add books --source books --permissions "anonymous:*"`

Config Json created:
```json
"entities": {
    "books": {
  "source": {
    "object": "books",
    "type": "table"
  },
  "graphql": {
    "enabled": true,
    "type": {
      "singular": "books",
      "plural": "books"
    }
  },
  "rest": {
    "enabled": true
  },
  "permissions": [
    {
      "role": "anonymous",
      "actions": [
        {
          "action": "*"
        }
      ]
    }
  ]
}
  }
```
- When the config file is hand-edited to add `methods` property to a table/view, engine logs a warning message during start-up

Config:
```json
"Book": {
      "source": {
        "object": "books",
        "type": "table"
      },
      "rest": {
        "enabled": true,
        "methods": [
          "get",
          "post"
        ]
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "book",
          "plural": "books"
        }
      },
```
Warning Logged:

image


seantleonard (commented):
Some questions.

src/Config/Converters/EntityRestOptionsConverter.cs
```diff
@@ -108,13 +108,21 @@ public override void Write(Utf8JsonWriter writer, EntityRestOptions value, JsonS
...
+            {
+                writer.WriteStringValue(JsonSerializer.SerializeToElement(method, options).GetString());
+            }
+
+            writer.WriteEndArray();
+        }
+        else if (value.Methods is null && options.DefaultIgnoreCondition != JsonIgnoreCondition.WhenWritingNull)
+        {
+            writer.WriteNull("methods");
         }
```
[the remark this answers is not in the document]
severussundar: Removed the condition.

We do set `DefaultIgnoreCondition` as `JsonIgnoreCondition.WhenWritingNull` currently. Even if we change it to something else in the future, it would be better to not write anything to the config file than writing `null` values.

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -232,7 +232,8 @@ public void SetEntityPermissionMap(RuntimeConfig runtimeConfig)
                 bool isStoredProcedureEntity = entity.Source.Type is EntitySourceType.StoredProcedure;
                 if (isStoredProcedureEntity)
                 {
-                    SupportedHttpVerb[] methods = entity.Rest.Methods;
+                    // An entity backed by a stored procedure is guaranteed to have REST methods populated.
```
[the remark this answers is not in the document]
severussundar: During deserialization, the REST methods is always populated - even if the config file does not contain rest or methods fields.

However, to get away with using null forgiving operator, I've added a null check and initialized `methods` as an empty array when REST methods is null.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -451,7 +451,8 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
             if (dbObject.SourceType == EntitySourceType.StoredProcedure)
             {
                 Entity entityTest = _runtimeConfig.Entities[entityName];
-                List<SupportedHttpVerb> spRestMethods = entityTest.Rest.Methods.ToList();
+
+                List<SupportedHttpVerb>? spRestMethods = (entityTest.Rest.Methods is not null) ? entityTest.Rest.Methods.ToList() : null;
```
ayush3797, line 455: Do we need to check for nullability of entity.rest.methods even for stored procedures? As per this code block, it doesn't look like we need to: https://github.com/Azure/data-api-builder/blob/83e300fa86f528e49a7deff09aa8605e10f39be6/src/Config/ObjectModel/RuntimeEntities.cs#L134

src/Core/Services/RestService.cs
```diff
@@ -334,7 +334,7 @@ private bool TryGetStoredProcedureRESTVerbs(string entityName, [NotNullWhen(true
             {
                 if (runtimeConfig.Entities.TryGetValue(entityName, out Entity? entity))
                 {
-                    SupportedHttpVerb[] methods = entity.Rest.Methods;
+                    SupportedHttpVerb[] methods = (entity.Rest.Methods is not null) ? entity.Rest.Methods : Array.Empty<SupportedHttpVerb>();
```
ayush3797, line 337: Same question as https://github.com/Azure/data-api-builder/pull/1607/files#r1277992470

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1260,7 +1259,7 @@ private static RuntimeConfig BuildTestRuntimeConfig(EntityPermission[] permissio
         {
             Entity sampleEntity = new(
                 Source: new(entityName, EntitySourceType.Table, null, null),
-                Rest: new(Array.Empty<SupportedHttpVerb>()),
+                Rest: new(Enabled: true),
```
ayush3797, line 1262: Since Enabled = true is default, can skip providing a value here.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1343,7 +1343,7 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
                 GetConnectionStringFromEnvironmentConfig(environment: TestCategory.MSSQL), Options: null);
             Entity viewEntity = new(
                 Source: new("books_view_all", EntitySourceType.Table, null, null),
-                Rest: new(EntityRestOptions.DEFAULT_SUPPORTED_VERBS),
+                Rest: new(Enabled: true),
```
ayush3797, line 1346: Can skip providing value here as well.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -267,6 +269,21 @@ public static void ValidateEntityConfiguration(RuntimeConfig runtimeConfig)
...
+        /// <summary>
+        /// Helper method to validate and let users know whether insignificant properties are present in the REST field.
+        /// Currently, it checks for the presence of Methods property when the entity type is table/view and logs a warning.
+        /// Methods property plays a role only in case of stored procedures.
+        /// </summary>
+        /// <param name="entity">Entity object for which validation is performed</param>
+        /// <param name="entityName">Name of the entity</param>
+        private void ValidatePresenceOfExtraRestProperties(Entity entity, string entityName)
+        {
+            if (entity.Source.Type != EntitySourceType.StoredProcedure && entity.Rest.Methods is not null && entity.Rest.Methods.Length > 0)
```
ayush3797, line 281: Super nit: Either use `is not` or `!=` for both the comparisons. Looks better :)

ayush3797: Since this method is called from only one place, do we want to create a method at all?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1778,88 +1785,64 @@ string[] linkingTargetFields
         }
 
         /// <summary>
-        /// Test to validate that the rest methods are correctly configured for entities in the config.
-        /// Rest methods can only be configured for stored procedures as an array of valid REST operations. 
+        /// Validates that a warning is logged when REST methods are configured for tables and views.
```
ayush3797, line 1788: And that it works fine for stored procedures!

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1914,17 +1897,23 @@ string[] linkingTargetFields
                 Entities: new(entityMap)
             );
 
+            MockFileSystem fileSystem = new();
+            FileSystemRuntimeConfigLoader loader = new(fileSystem);
+            RuntimeConfigProvider provider = new(loader);
+            RuntimeConfigValidator configValidator = new(provider, fileSystem, new Mock<ILogger<RuntimeConfigValidator>>().Object);
+            Mock<ISqlMetadataProvider> _sqlMetadataProvider = new();
```
ayush3797, line 1904: This variable has no references.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1982,17 +1971,23 @@ string[] linkingTargetFields
                 Entities: new(entityMap)
             );
 
+            MockFileSystem fileSystem = new();
```
ayush3797, line 1974: This block of code is common among few tests. Can we scope it out into another helper which will return the instance of RuntimeConfigValidator that we need?

src/Service.Tests/TestHelper.cs
```diff
@@ -74,7 +74,7 @@ public static RuntimeConfig AddMissingEntitiesToConfig(RuntimeConfig config, str
             Entity entity = new(
                 Source: new(entityName, EntitySourceType.Table, null, null),
                 GraphQL: new(entityKey, entityKey.Pluralize()),
-                Rest: new(Array.Empty<SupportedHttpVerb>()),
+                Rest: new(Enabled: true),
```
ayush3797, line 77: same suggestion

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -232,8 +232,7 @@ public void SetEntityPermissionMap(RuntimeConfig runtimeConfig)
                 bool isStoredProcedureEntity = entity.Source.Type is EntitySourceType.StoredProcedure;
                 if (isStoredProcedureEntity)
                 {
-                    SupportedHttpVerb[] methods = entity.Rest.Methods;
-
+                    SupportedHttpVerb[] methods = (entity.Rest.Methods is not null) ? entity.Rest.Methods : Array.Empty<SupportedHttpVerb>();
```
ayush3797, line 235: https://github.com/Azure/data-api-builder/pull/1607/files#r1277992470

src/Config/ObjectModel/EntityRestOptions.cs
```diff
@@ -16,7 +16,7 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 /// <param name="Methods">The HTTP verbs that are supported for this entity.</param>
```
ayush3797, line 16: As an addition to the param summary, it would help to add that Methods only hold relevance for stored procedures. Won't cause confusion to someone new looking at the code about the purpose of this property.

ayush3797:
Some questions and suggestions around the same. Looks good so far!

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
severussundar: Right, while it is true that Rest Methods is guaranteed to be populated with the current logic, a null check acts as an extra guard - if the deserialization logic were to change in the future in which the Rest Methods could not be populated.

IMO, adding a null check does no harm, plus brings additional defence

src/Core/Services/RestService.cs
severussundar: Please refer to my response in that comment.

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
severussundar: Writing Enabled:true makes it slightly more readable - without opening the REST object definition, it is possible to understand that REST endpoint is enabled for this entity.

So, I've followed this convention throughout.

src/Core/Configurations/RuntimeConfigValidator.cs
severussundar: My thought process behind introducing a separate method is to group all validations related to the presence of extra/insignificant properties in a single place - At the moment, only one field is being validated but more such validations can be added in the future.

Also, having a separate method makes it re-usable in another place in the future if needed

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
severussundar: Thanks for the suggestion. Introduced a helper method.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
severussundar: Good catch, removed

src/Core/Services/RestService.cs
Aniruddh25, line 337: Shouldn't the methods array here be populated with POST - the default method for SP as per the function comment?

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
Aniruddh25, line 455: Shouldn't we populate this list spRestMethods with POST?

src/Core/Authorization/AuthorizationResolver.cs
Aniruddh25, line 235: Same - shouldn't this methods array be populated with POST?

Aniruddh25 (commented):
Looks good so far, left few comments.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -1778,88 +1757,64 @@ string[] linkingTargetFields
...
-                DataApiBuilderException ex =
-                    Assert.ThrowsException<DataApiBuilderException>(() => RuntimeConfigValidator.ValidateEntityConfiguration(runtimeConfig));
-                Assert.AreEqual(expectedErrorMessage, ex.Message);
-                Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
-                Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ConfigValidationError, ex.SubStatusCode);
-            }
-            else
+            MockFileSystem fileSystem = new();
+            FileSystemRuntimeConfigLoader loader = new(fileSystem);
+            RuntimeConfigProvider provider = new(loader);
```
Aniruddh25, line 1799: nit: use the newly added helper IntiatilizeRuntimeConfigObject private function here?

src/Core/Services/RestService.cs
severussundar: Thanks for the suggestion. Updated the logic to populate withPOST when the REST APIs are enabled for the entity

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
severussundar: updated

src/Core/Authorization/AuthorizationResolver.cs
severussundar: updated

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
severussundar: Since, the log message is being validated - both the log level and the message logged, we need the instance of the mock logger object. The helper method directly returns the `RuntimeConfigProvider` (which is what is needed in all the other tests).  So, I've not used the helper method for this test.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -451,7 +451,16 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
             if (dbObject.SourceType == EntitySourceType.StoredProcedure)
             {
                 Entity entityTest = _runtimeConfig.Entities[entityName];
-                List<SupportedHttpVerb> spRestMethods = entityTest.Rest.Methods.ToList();
+
+                List<SupportedHttpVerb>? spRestMethods;
+                if (entityTest.Rest.Methods is not null)
```
seantleonard, line 456: nit: i don't think i had a reason to continue using the name `entityTest`. You can just revise it to be `entity`.

Aniruddh25 (approved):
Thanks for noticing this bug and fixing it for 0.8!

src/Config/Converters/EntityRestOptionsConverter.cs
```diff
@@ -108,13 +108,17 @@ public override void Write(Utf8JsonWriter writer, EntityRestOptions value, JsonS
             writer.WriteNull("path");
         }
 
-        writer.WriteStartArray("methods");
-        foreach (SupportedHttpVerb method in value.Methods)
+        if (value.Methods is not null)
```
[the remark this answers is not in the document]
severussundar: Tried different combinations but unable to hit this case. However, it seems logical not to write an empty array to the config file. So, I've updated the condition to check for length > 0

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -232,7 +232,15 @@ public void SetEntityPermissionMap(RuntimeConfig runtimeConfig)
...
-                    SupportedHttpVerb[] methods = entity.Rest.Methods;
+                    SupportedHttpVerb[] methods;
+                    if (entity.Rest.Methods is not null)
+                    {
+                        methods = entity.Rest.Methods;
+                    }
+                    else
+                    {
+                        methods = (entity.Rest.Enabled) ? new SupportedHttpVerb[] { SupportedHttpVerb.Post } : Array.Empty<SupportedHttpVerb>();
+                    }
```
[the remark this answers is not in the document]
severussundar: 1. True, `entity.Rest.Methods` will get processed during config deserialization. With the current logic, `entity.Rest.Methods` will never come out to be `null`.

The code branch where `entity.Rest.Methods` can be `null` is added
a) To avoid using null forgiving operator
b) To safeguard against exceptions in the future, - if the deserialization logic should change in a way that the entity.Rest.Methods could come to be `null`.


2) Having an entity with entity.Rest disabled while having rest methods populated does not break. In `SqlMetadataProvider`, the logic for populating `EntityPathToEntityName` takes care of checking if the rest is enabled before mapping a certain path to an entity. 

When a REST request is made, `EntityPathToEntityName` is checked to see if the `path` in the request is a known path (maps to any of the configured entities or not). If an entity is not exposed through REST endpoint, this data structure would not contain a mapping for the `path` and would lead to failure of the request (404 not found).

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -267,6 +269,21 @@ public static void ValidateEntityConfiguration(RuntimeConfig runtimeConfig)
...
         }
 
+        /// <summary>
+        /// Helper method to validate and let users know whether insignificant properties are present in the REST field.
+        /// Currently, it checks for the presence of Methods property when the entity type is table/view and logs a warning.
+        /// Methods property plays a role only in case of stored procedures.
+        /// </summary>
+        /// <param name="entity">Entity object for which validation is performed</param>
+        /// <param name="entityName">Name of the entity</param>
+        private void ValidatePresenceOfExtraRestProperties(Entity entity, string entityName)
```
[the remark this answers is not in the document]
severussundar: updated

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -450,8 +450,17 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
...
             if (dbObject.SourceType == EntitySourceType.StoredProcedure)
             {
-                Entity entityTest = _runtimeConfig.Entities[entityName];
-                List<SupportedHttpVerb> spRestMethods = entityTest.Rest.Methods.ToList();
+                Entity entity = _runtimeConfig.Entities[entityName];
+
+                List<SupportedHttpVerb>? spRestMethods;
+                if (entity.Rest.Methods is not null)
+                {
+                    spRestMethods = entity.Rest.Methods.ToList();
```
ayush3797, line 458: Is it possibe for rest to be disabled here?

ayush3797 (approved):
Just one question. LGTM! Thanks for fixing this!

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
severussundar: No. The caller of this method checks if rest is enabled for the entity before calling this method.

The check happens at `OpenApiDocumentor.BuildPaths()` : L 180 -186

PR #1626, opened 2023-08-10
Fix Location Header in POST Response header

### LINKED ISSUE #1613: [Bug] Unable to perform GET request using Location Header returned for POST requests
POST [Specs](https://www.rfc-editor.org/rfc/rfc7231#section-4.3.3) specify that a Location header should be returned when a POST request results in the creation of a new item.

The Location Header returned by DAB is incorrect. Performing a GET request on it results in the following error.

Location Header returned at the moment: `https://localhost:5001/api/Book//id/5009`

GET request on this URI results in the following error

```json
{
    "error": {
        "code": "BadRequest",
        "message": "Support for url template with implicit primary key field names is not yet added.",
        "status": 400
    }
}
```
Correct value to be returned: `https://localhost:5001/api/Book/id/5009` 

For Stored Procedures:

Location Header returned: `https://localhost:5001/api/CountBooks//CountBooks`
Correct value to be returned: `https://localhost:5001/api/CountBooks/`

severussundar:
## Why make this change?

- Closes #1613 
  - According to [HTTP POST Specs](https://www.rfc-editor.org/rfc/rfc9110#name-post), when a POST request results in the creation of a new item, then a Location header field should be returned in the response header.
  > If one or more resources has been created on the origin server as a result of successfully processing a POST request, the origin server SHOULD send a [201 (Created)](https://www.rfc-editor.org/rfc/rfc9110#status.201) response containing a [Location](https://www.rfc-editor.org/rfc/rfc9110#field.location) header field that provides an identifier for the primary resource created ([Section 10.2.2](https://www.rfc-editor.org/rfc/rfc9110#field.location)) and a representation that describes the status of the request while referring to the new resource(s).
  - The Location header returned in the response headers were incorrect
> Tables/View: `https://localhost:5001/api/Book//id/5009`
> Stored Procedures: `https://localhost:5001/api/CountBooks//CountBooks`

  - When `base-route` is configured in the config file, the Location returned did not account for it. So, a subsequent GET request performed using the Location is unsuccessful.

## What is this change?

- `RestController`: Logic to construct the `Location` header is updated to accommodate the `base-route` field
- `RestService`: Helper method to extract the configured base-route is added
- `ConfigurationTests`: Integration tests to validate the Location field when a) base-route is configured b) base-route is absent for tables and stored procedures are added.

## How was this tested?

- [x] Integration Tests
- [x] Manual Tests

## Sample Request(s)

- Table and base-route is not configured
image

- Stored Procedure and base-route is not configured
image

- Table and base-route is configured. Configured base-route: `/data-api`
image

- Stored Procedure and base-route is configured. Configured base-route: `/data-api`
image


src/Service/Controllers/RestController.cs
```diff
@@ -224,16 +224,21 @@ public static JsonResult ErrorResponse(string code, string message, HttpStatusCo
...
                     // then update the Location of the created result to this value.
                     CreatedResult createdResult = (result as CreatedResult)!;
-                    string location = UriHelper.GetEncodedUrl(HttpContext.Request) + "/" + createdResult.Location;
-                    createdResult.Location = location;
+                    string baseURL = UriHelper.BuildAbsolute(
+                                        scheme: HttpContext.Request.Scheme,
+                                        host: HttpContext.Request.Host,
+                                        pathBase: _restService.GetBaseRouteFromConfig(),
+                                        path: HttpContext.Request.Path);
+                    createdResult.Location = baseURL.EndsWith('/') ? baseURL + createdResult.Location : baseURL + "/" + createdResult.Location;
```
Aniruddh25, line 241: Is it not deterministic if the baseURL will contain a trailing '/'? What if the entity name contains '/'?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,177 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+                    {
+                        Content = JsonContent.Create(requestBodyElement)
+                    };
+                }
+
+                HttpResponseMessage response = await client.SendAsync(request);
+                Assert.AreEqual(HttpStatusCode.Created, response.StatusCode);
+
+                string locationHeader = response.Headers.Location.AbsoluteUri;
+                Assert.IsTrue(locationHeader.StartsWith(expectedLocationHeader));
```
Aniruddh25, line 1494: Why check for StartsWith as opposed to Equals?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,177 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+
+                string path = response.Headers.Location.AbsolutePath;
+                string completeUrl = "http://localhost" + path.Substring(baseRoute.Length - 1);
```
Aniruddh25, line 1497: Can you please add the reason why completeUrl should skip baseRoute in code comments here?

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -297,7 +297,7 @@ public async Task InsertOneInViewBadRequestTest()
                 operationType: EntityActionOperation.Execute,
                 requestBody: requestBody,
                 expectedStatusCode: HttpStatusCode.Created,
-                expectedLocationHeader: _integrationProcedureInsertOneAndDisplay_EntityName,
+                expectedLocationHeader: string.Empty,
```
Aniruddh25, line 300: We already have some location header checks. Arent those sufficient? I remember our base tests had been modified to simulate real client server request responses as well.. Cant we send the followup response in these integration tests? 

Do we really need to add additional tests to the ConfigurationTests?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,177 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+                configuration = InitMinimalRuntimeConfig(dataSource, graphqlOptions, restRuntimeOptions, entity, entityName);
+            }
+            else
+            {
+                configuration = InitMinimalRuntimeConfig(dataSource, graphqlOptions, restRuntimeOptions);
+            }
+
+            const string CUSTOM_CONFIG = "custom-config.json";
+
+            AuthenticationOptions AuthenticationOptions = new(Provider: EasyAuthType.StaticWebApps.ToString(), null);
```
Aniruddh25, line 1458: Can this be another DataRow in the test above? Looks like this test has lot of similar code than the previous...

Aniruddh25 (changes requested):
Thanks for finding this bug and fixing it! Left few questions..

src/Core/Services/RestService.cs
```diff
@@ -395,6 +395,22 @@ public bool TryGetRestRouteFromConfig([NotNullWhen(true)] out string? configured
...
+        /// Helper method to extract the configured base route
+        /// </summary>
+        public string GetBaseRouteFromConfig()
+        {
+            if (_runtimeConfigProvider.TryGetConfig(out RuntimeConfig? config)
+                && config is not null
+                && config.Runtime is not null
+                && config.Runtime.BaseRoute is not null)
+            {
+                return config.Runtime.BaseRoute;
```
ayush3797, line 408: An empty base-route i.e. `"/"` is allowed. But in such a case it is not included in the url. We should only return base-route when base-route is of the form "/xyz". You can be assured that the base-route always starts with forward slash as we have a validation for that in place.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,177 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
             }
         }
 
+        /// <summary>
+        /// Validates the Location header field returned as part of POST requests.
+        /// </summary>
+        /// <param name="entityType">Type of the entity</param>
+        /// <param name="requestPath">Request path for performing POST API requests on the entity</param>
+        /// <returns></returns>
```
ayush3797, line 1339: nit: remove the returns tag.

ayush3797, line 1339:
```suggestion
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,177 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+            string[] args = new[]
+            {
+                $"--ConfigFileName={CUSTOM_CONFIG}"
+            };
+
+            using (TestServer server = new(Program.CreateWebHostBuilder(args)))
+            using (HttpClient client = server.CreateClient())
+            {
+                HttpMethod httpMethod = SqlTestHelper.ConvertRestMethodToHttpMethod(SupportedHttpVerb.Post);
+                HttpRequestMessage request = new(httpMethod, requestPath);
```
ayush3797, line 1384: For tests performing mutations, we reset the database state after performing the mutation operation. However, here it would be difficult to reset the state unless we do something like we do with RestApiTestBase. You might consider performing a subsequent delete operation after performing the GET operation in this test. This way, the test would not change the state of the database.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,177 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+        }
+
+        /// <summary>
+        /// Validates the Location header field returned as part of POST requests
+        /// </summary>
+        /// <param name="entityType">Type of the entity</param>
+        /// <param name="requestPath">Request path for performing POST API requests on the entity</param>
+        /// <param name="baseRoute">Configured base route</param>
+        /// <param name="expectedLocationHeader">Expected value for Location field in the response header</param>
+        /// <returns></returns>
```
ayush3797, line 1420:
```suggestion
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,177 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+        [DataTestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        [DataRow(EntitySourceType.Table, "/api/Book", "/data-api", "http://localhost/data-api/api/Book/id/")]
+        [DataRow(EntitySourceType.StoredProcedure, "/api/GetBooks", "/data-api", "http://localhost/data-api/api/GetBooks")]
```
ayush3797, line 1424: Missing displaynames.

src/Service.Tests/Configuration/ConfigurationTests.cs
ayush3797: Yes, because when the request lands at DAB, the base-route would be stripped off from the url.

src/Service/Controllers/RestController.cs
ayush3797: To add here: Entity name cannot contain '/'.

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: `base-route` is used by SWA for redirection only. Any request to `/data-api` (or the configured base-route) gets redirected to phoenix clusters by SWA. However, all the APIs are hosted in DAB containers at the `/api`. 
So, when DAB is not integrated with SWA, to be able to perform a GET request, we need to remove the `base-route` from the URL.

src/Core/Services/RestService.cs
severussundar: Thanks for bringing up this edge case. I tested with just `/` as the base-route. The `Location` header is getting constructed with the right URI. 

The baseURL is constructed using `https://learn.microsoft.com/dotnet/api/microsoft.aspnetcore.http.extensions.urihelper.buildabsolute` and it takes care of dealing with such strings I suppose.

For my understanding, it seems like we shouldn't be allowing just `/` in the base-route. Could you please explain the thought process behind why we chose allow.

ayush3797: This is because '//' still is valid in a url i.e. a url can contain en empty component.

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: When `base-route` is configured, the Location header in the repsonse will be of the form `/<base-route>/api/EntityName/id/<PKs>`. However, `base-route` is applicable only in SWA-DAB integrated scenario. When DAB engine is run independently, all the APIs are hosted on the `/api` endpoint. So, we cannot directly perform a GET request on the Location header received in the response (like we do in the test where `base-route` is absent).

To test whether the returned `base-route` returned in the response is correct, I am asserting it against the `expectedLocationHeader`. Since, the PK for the `Book` entity is auto-generated, it is not possible to know the exact `id` beforehand. So,  `expectedLocationHeader` string does not have the exact id (is of the form `.../id/`). `Equals` would not work here and hence decided to go with `startsWith`. Also, to validate that the id returned in the response is valid, another GET request is performed after removing the base-route from the URL and checking that it results in an OK response (200).

src/Core/Services/RestService.cs
ayush3797: Looks like this needs to be fixed: https://github.com/Azure/data-api-builder/blob/adb2be7f6ea2726d29da284fc935dc11a3741885/src/Core/Services/RestService.cs#L366. We would want to strip off leading '/' from the `route` before this check is performed.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
severussundar: The current checks we have validates only the PK part of the Location header. The entire field was not validated.
a) Adding tests in Configuration Tests lets us modify the config to add a custom base route which gets used only in that test. We need this for the scenario where base-route is configured.
b) In the existing integration tests, we have a common validation method for all Insert API tests. The current check validates only the PK part of the location header and not the entire field. Performing a GET follow-up request there would mean that the GET action should be enabled for the entities (and we need to ensure that entities that are added in the future would also have GET APIs enabled). Or introduce another flag in the tests to essentially indicate if a follow up GET request needs to be performed.

Considering these, I've added new tests in ConfigurationTests

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: I started down that route but noticed that there were lots of branching logic needed in the tests.

Here are the places where branching logic was necessary.

a) Entity creation for tables vs stored procedures
b) Custom config file creation when base-route is configured vs not configured
c) Request body creation for tables vs stored-procedures 
d) Validations based on whether base-route is configured vs not configured.

Overall, having too many branching in a single test made the test cluttered. So, breaking them up into two different tests reduced the clutter and (hopefully) made it more intuitive to follow.

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: I touched upon this in an earlier comment also. This was one of the reasons for me to choose an entity whose PK is auto generated. As part of this test, a single new book item is created whose id cannot be known beforehand. So, we can be sure that the generated id for this item could not have been used in any other test for read, update, delete etc. AFAIK, all the integration tests (except for the ones in ConfigurationTests) has a test setup phase where the database state is reset to scratch by executing the .sql scripts. Creation of two new items in this entity shouldn't interfere with other tests.

ayush3797: I see. So unless I run other tests which reset the database, this tests will continue to insert new records everytime it is run?

src/Core/Services/RestService.cs
seantleonard: @ayush3797 `GetRouteAfterPathBase(string route)` expects the `route` argument to not have a starting slash '/'

Can you clarify what you are suggesting should be fixed?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1331,6 +1331,179 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+        /// <param name="entityType">Type of the entity</param>
+        /// <param name="requestPath">Request path for performing POST API requests on the entity</param>
+        [DataTestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        [DataRow(EntitySourceType.Table, "/api/Book", DisplayName = "Location Header validation - Table, Base Route not configured")]
+        [DataRow(EntitySourceType.StoredProcedure, "/api/GetBooks", DisplayName = "Location Header validation - Stored Procedures, Base Route not configured")]
+        public async Task ValidateLocationHeaderFieldForPostRequests(EntitySourceType entityType, string requestPath)
+        {
+
+            GraphQLRuntimeOptions graphqlOptions = new(Enabled: true);
```
seantleonard, line 1346: Why set graphql to enabled? can disable it and not have extra code defining an entity's graphQL metadata. That slightly distracts from the test because setting that metadata makes me think it is relevant in testing.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: is there a structure like AbsoluteUri that can be contructed with the value in your location header? then you can more efficiently(not writing a custom algo) check any remaining components of the uri.

src/Service/Controllers/RestController.cs
```diff
@@ -224,16 +224,21 @@ public static JsonResult ErrorResponse(string code, string message, HttpStatusCo
 
                 if (result is CreatedResult)
                 {
-                    // Location is made up of two parts, the first being constructed
-                    // from the HttpRequest found in the HttpContext. The other part
-                    // is the primary key route, which has already been saved in the
+                    // Location is made up of three parts, the first being constructed
+                    // from the HttpRequest found in the HttpContext. The second part being the
+                    // base route configured in the config file.
+                    // The third part is the primary key route, which has already been saved in the
```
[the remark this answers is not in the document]
seantleonard: So the reason the locatation header logic is in Restcontroller is because you have no access to the runtime config (base path value) in SQLmutation engine? 

Did you consider keeping the location header population in sqlmutation engine, and just using this place in RestController to prepend the base path, if configured?

seantleonard (commented):
Some suggestions for test layout.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -253,13 +253,13 @@ public class SqlMutationEngine : IMutationEngine
                     {
                         using (JsonDocument jsonDocument = JsonDocument.Parse(resultArray.ToJsonString()))
                         {
-                            return new CreatedResult(location: context.EntityName, OkMutationResponse(jsonDocument.RootElement.Clone()).Value);
+                            return new CreatedResult(location: string.Empty, OkMutationResponse(jsonDocument.RootElement.Clone()).Value);
```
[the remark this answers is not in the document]
severussundar: For SP, the Location is of the form `../api/<SP-Entity-Name>`

Since the baseURL already has the `Entity Name`, setting the entity name again in location results in incorrect Location header value --> `../api/CountBooks/CountBooks`. To prevent this, location needs to be set as an empty string.

src/Service/Controllers/RestController.cs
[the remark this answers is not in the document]
severussundar: 1) Empty string is set only in case of stored procedures. For other entity types, non-empty value is returned. I've described the reason for returning an empty string for stored procedures in an earlier comment. I would say the comment is accurate. 

2) For non-stored procedures, the location is set with the PK of the newly created item. This is constructed in `SqlMutationEngine.ConstructPirmaryKeyRoute()`. The PK of the newly created item is accessible here, and hence, that calculation takes place here.
To extract the base-route, access to Runtime config is necessary, and therefore that logic is written in `RestService`. In `RestController`, all the three components - baseURL, base-route, and PK route are assembled to form the entire location header before returning the response.

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: It is possible to get the individual components of the URI (we get the complete URI from the response header) by assigning it to URI class and writing a little bit of custom logic to identify the base-route, Entity name etc. But, introducing a custom logic cannot be avoided.

My thought process behind testing the location header this way are
a) The intention of returning a Location header is to facilitate performing a GET request to fetch the details of the newly created item. So, I had added a validation that performs a GET request using the location header.
b) Since the test runs DAB independently (not integrated with SWA), a separate validation is needed for base-route (a direct GET request cannot be performed on the location header). Base route is validated by asserting against expected values. 

IMHO, this does not introduce a long running test (just one additional GET request) - which should be fine given the nature of the field being validated.

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: Agreed. I've disabled graphql

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: Thanks for the suggestion. I've added a DELETE request to cleanup the newly created item

Aniruddh25 (approved):

src/Core/Services/RestService.cs
```diff
@@ -395,6 +395,20 @@ public bool TryGetRestRouteFromConfig([NotNullWhen(true)] out string? configured
             return false;
         }
 
+        /// <summary>
+        /// Helper method to extract the configured base route
+        /// </summary>
+        public string GetBaseRouteFromConfig()
+        {
+            if (_runtimeConfigProvider.TryGetConfig(out RuntimeConfig? config)
+                && config.Runtime.BaseRoute is not null)
```
Aniruddh25, line 404: Are we not checking for config.Runtime is not null because it can never be?

src/Core/Services/RestService.cs
ayush3797: That we need to strip off all the leading '/' at line 365 i.e. TrimStart('/').

src/Core/Services/RestService.cs
severussundar: Yeah, since `Runtime` is a non-nullable property of `RuntimeConfig`, that check is skipped.

src/Core/Services/RestService.cs
severussundar: Added the logic to remove leading slashes

src/Service/Controllers/RestController.cs
```diff
@@ -224,16 +224,21 @@ public static JsonResult ErrorResponse(string code, string message, HttpStatusCo
...
                     // Location of the created result. So we form the entire location
-                    // from appending the primary key route  already stored in the
+                    // from appending the base route and the primary key route  already stored in the
                     // created result to the url constructed from the HttpRequest. We
                     // then update the Location of the created result to this value.
                     CreatedResult createdResult = (result as CreatedResult)!;
-                    string location = UriHelper.GetEncodedUrl(HttpContext.Request) + "/" + createdResult.Location;
-                    createdResult.Location = location;
+                    string baseURL = UriHelper.BuildAbsolute(
```
seantleonard, line 236: naming this 'baseURL' seems odd since we have a value called "BaseRoute" and 'base' implies that there is more URL to follow. Would 'locationUrl' make more sense? or something like that?

src/Service/Controllers/RestController.cs
```diff
@@ -224,16 +224,21 @@ public static JsonResult ErrorResponse(string code, string message, HttpStatusCo
...
+                                        scheme: HttpContext.Request.Scheme,
+                                        host: HttpContext.Request.Host,
+                                        pathBase: _restService.GetBaseRouteFromConfig(),
+                                        path: HttpContext.Request.Path);
+                    createdResult.Location = baseURL.EndsWith('/') ? baseURL + createdResult.Location : baseURL + "/" + createdResult.Location;
                     result = createdResult;
```
seantleonard, line 242: Instead of string concatenation and ternary conditional operator `? :`, could you use the Microsoft.AspNetCore.Http `PathString.Add(PathString)` method? perhaps that will accommodate checking for trailing slashses. https://learn.microsoft.com/dotnet/api/microsoft.aspnetcore.http.pathstring.add?view=aspnetcore-6.0#microsoft-aspnetcore-http-pathstring-add(microsoft-aspnetcore-http-pathstring)

src/Core/Services/RestService.cs
```diff
@@ -372,6 +372,13 @@ public string GetRouteAfterPathBase(string route)
             // The RuntimeConfigProvider enforces the expectation that the configured REST path starts with a
             // forward slash '/'.
             configuredRestPathBase = configuredRestPathBase.Substring(1);
+
+            // Consider a typical URL of the form: <scheme>://<domain-name>/<path_to_the_resource>.
+            // The variable route contains the value of <path-to-resource> section.   
+            // Since, <scheme>://<domain-name>//<path_to_the_resource> is a valid URL, leading slashses are removed before comparing it with
+            // incoming request's path.
+            route = route.TrimStart('/');
```
seantleonard, line 380: if `//api/<path_to_the_resource>` is a valid url, why are we modifying the user entered URL here? An invalid URL should route to HTTP 404 Not Found. By trimming this URL we are assuming (perhaps incorrectly) what the user intended url should be.

Consider the double slash with different placement in the URL: `https://localhost:5001/api//Book`. That is an invalid route for accessing book, because the name of the entity is blank `//`, and Book would be interpreted as the primary key. This example is handled by preexisting code in the return statement at the end of this method `return route.Substring(configuredRestPathBase.Length).TrimStart('/');` I think this should probably be removed too and instead result in 404.

seantleonard: TrimStart in the return statement was intended to only remove a slash if only 1 existed, not >1.

src/Service.Tests/Unittests/RestServiceUnitTests.cs
```diff
@@ -70,7 +70,6 @@ public class RestServiceUnitTests
         /// <param name="route">The route to be parsed.</param>
         /// <param name="path">An invalid path for the given route.</param>
         [DataTestMethod]
-        [DataRow("/foo/bar", "foo")]
```
seantleonard, line 73: I would argue that this test case is correct to have in "ErrorForInvalidRouteAndPathToParseTest" because it implies that the URL entered into the browser or api client was `https://localhost:5000//api/entity`.

`GetRouteAfterPathBase()` requires a URL with no starting slashes and the route value calculated by the RestController in `        public async Task<IActionResult> Find(string route)` will remove the first slash  leaving `/api/entity` as the route.

src/Service.Tests/Unittests/RestServiceUnitTests.cs
```diff
@@ -40,7 +40,7 @@ public class RestServiceUnitTests
         /// <param name="expectedPrimaryKeyRoute">The primary key route we
         /// expect to parse from route.</param>
         [DataTestMethod]
-        [DataRow("rest-api/Book/id/1", "/rest-api", "Book", "id/1")]
+        [DataRow("/rest-api/Book/id/1", "/rest-api", "Book", "id/1")]
```
seantleonard, line 43: see comment on other test's modification: this should be considered an invalid route.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1389,6 +1389,194 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
             }
         }
 
+        /// <summary>
+        /// Validates the Location header field returned as part of POST requests.
```
seantleonard, line 1393: Can you be more detailed? The test displaynames mention "Base Route not configured". This comment should explain the significance of that and how it affects the expected/not expected results.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1389,6 +1389,194 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+                        Content = JsonContent.Create(requestBodyElement)
+                    };
+                }
+
+                HttpResponseMessage response = await client.SendAsync(request);
+
+                // Location header field is expected only when POST request results in the creation of a new item
+                Assert.AreEqual(HttpStatusCode.Created, response.StatusCode);
+
+                string locationHeader = response.Headers.Location.AbsoluteUri;
```
seantleonard, line 1461: Some of my other comments call out the fact that we call `TrimStart('/')` in our RestService code to modify the route. Depending on whether we remove that to let invalid routes fail, I think this test would benefit from explicitly comparing against the actualRoute returned in the location header, and an expectedRoute provided by the DataRows of the test. That would also help self-describe what you are validating without reading through the entirety of the test's logic.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1389,6 +1389,194 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+            }
+        }
+
+        /// <summary>
+        /// Validates the Location header field returned as part of POST requests
+        /// </summary>
+        /// <param name="entityType">Type of the entity</param>
+        /// <param name="requestPath">Request path for performing POST API requests on the entity</param>
+        /// <param name="baseRoute">Configured base route</param>
+        /// <param name="expectedLocationHeader">Expected value for Location field in the response header</param>
```
seantleonard, line 1483: This was discussed in a different thread, but would be good to call out in this param comment that the expectedLocationHeader excludes the primary key value from the header because we check using `string.startswith()`

seantleonard (changes requested):
Concern with trimming >1 consecutive slashes from the URL and treating that as valid url. while //api/Book/id/1 is a "valid URL" , it is not a valid way to access the book entity. That is an invalid path which should result in either a 404 not found or 400 bad request.

seantleonard:
Please ignore the first review as it is a duplicate cause by VScode GitHub PR extension.

src/Service/Controllers/RestController.cs
severussundar: Thanks for the suggestion. 

`PathString` expects the string to start with `/`, failing which it throws an exception. The `createdResult.Location` that we create does not have a leading `/`. To be able to use `PathString`, we need to add a leading slash to the PK part (createdResult.Location). AFAICT, manual check/addition cannot be avoided completely. 

Considering this, the current implementation should be fine - performs a check and introduces a `/` only when necessary

src/Service/Controllers/RestController.cs
severussundar: Renamed

src/Core/Services/RestService.cs
severussundar: Removed logics related to trimming leading slashes. Since this discussion is tangential to the intention of this PR, opened [this](https://github.com/Azure/data-api-builder/issues/1640) for tracking the same. Based on the conclusion there, a follow-up PR can take care of those changes.

src/Service.Tests/Unittests/RestServiceUnitTests.cs
severussundar: added back

src/Service.Tests/Unittests/RestServiceUnitTests.cs
severussundar: added back. Created [this](https://github.com/Azure/data-api-builder/issues/1640) issue for discussions related to consecutive slashes.

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: Removed logic for trimming `/`. URL entered by the user is not modified. Created [this](https://github.com/Azure/data-api-builder/issues/1640) issue for discussion on how to deal with consecutive slashes. At the moment, the existing logic is retained as is. 

I do think performing a subsequent GET request would be the best way to validate the correctness of the PK part of the Location header.

src/Core/Services/RestService.cs
seantleonard: just as a follow to this thread, the leading slashes should not be removed because even though `//` is a valid URL, it is not a valid dab URL -> should result in a bad request or 404 not found.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1389,6 +1389,206 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Validates the Location header field returned for a POST request when it results in a 201 response. The idea behind returning
+        /// a Location header is to provide a URL against which a GET request can be performed to fetch the details of the new item.
+        /// Base Route is configured in the config file used for this test. So, it is expected that the Location header returned will contain the base-route.
+        /// This test performs a POST request, and checks if it results in a 201 response. If so, the test validates the correctness of the Location header in two steps.
+        /// Since, base-route has significance only in the SWA-DAB integrated scenario and this test is executed against DAB running independently,
```
seantleonard, line 1486:
```suggestion
        /// Since base-route has significance only in the SWA-DAB integrated scenario and this test is executed against DAB running independently,
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1389,6 +1389,206 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+        /// the expectedLocationHeader excludes the PK. Because of this, the actual location header is validated by checking if it starts with the expectedLocationHeader.</param>
+        [DataTestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        [DataRow(EntitySourceType.Table, "/api/Book", "/data-api", "http://localhost/data-api/api/Book/id/", DisplayName = "Location Header validation - Table, Base Route configured")]
+        [DataRow(EntitySourceType.StoredProcedure, "/api/GetBooks", "/data-api", "http://localhost/data-api/api/GetBooks", DisplayName = "Location Header validation - Stored Procedure, Base Route configured")]
+        public async Task ValidateLocationHeaderWhenBaseRouteIsConfigured(EntitySourceType entityType,
+                                                                          string requestPath,
+                                                                          string baseRoute,
+                                                                          string expectedLocationHeader)
```
seantleonard, line 1503: first param should be bumped to next line and then indentation is consistent with other tests that have params.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1389,6 +1389,206 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+
+                string locationHeader = response.Headers.Location.AbsoluteUri;
+                Assert.IsTrue(locationHeader.StartsWith(expectedLocationHeader));
+
+                // The URL to perform the GET request is constructed by skipping the base-route.
+                // Base Route field is applicable only in SWA-DAB integrated scenario. When DAB engine is run independently, all the
+                // APIs are hosted on /api. But, the returned Location header in this test will contain the configured base-route. So, this needs to be
+                // removed before performing a subsequent GET request.
+                string path = response.Headers.Location.AbsolutePath;
+                string completeUrl = "http://localhost" + path.Substring(baseRoute.Length - 1);
```
seantleonard, line 1576: You don't need to explicitly set localhost here as that is implied when creating a httprequestmessage.

seantleonard (approved):
just a last few suggestions before merging.

src/Core/Services/RestService.cs
severussundar: The existing logic is retained as is. The logic for removing leading slashes are not removed.

Issue #452, opened 2022-05-27
Table with triggers error out on UPDATE


yorek:
If a table has a trigger, the following error is generated:

```
Route matched with {action = "Post", controller = "GraphQL"}. Executing controller action with signature System.Threading.Tasks.Task`1[System.Text.Json.JsonElement] PostAsync() on controller Azure.DataGateway.Service.Controllers.GraphQLController (Azure.DataGateway.Service).
UPDATE [s001].[todos] SET [s001].[todos].[title] = @param1 OUTPUT Inserted.[id] WHERE [s001].[todos].[id] = @param0;
Microsoft.Data.SqlClient.SqlException (0x80131904): The target table 's001.todos' of the DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause.
```

ayush3797:
I spent some time doing exploration on this issue and will be documenting my findings here.

1. We can use `OUTPUT` clause if a trigger is not defined.

**Demonstration using T-SQL:**

```tsql
-- Creating a table TWOT (table without trigger) and fetching data using OUTPUT clause 
-- works fine since no trigger is created (and enabled) for the table.
DROP TABLE IF EXISTS TWOT;
create Table TWOT(
id int PRIMARY KEY,
title varchar(50)
);
INSERT INTO TWOT(id,title) values(1,'book');
update TWOT set title='DAB' output inserted.id,inserted.title where id=1;
```

2. `OUTPUT` clause used on a table with trigger configured errors out - if not used with `INTO` clause.

**Demonstration using T-SQL:**
```tsql
DROP TABLE IF EXISTS TWT;
create Table TWT(
id int PRIMARY KEY,
title varchar(50)
);
INSERT INTO TWT(id,title) values(1,'book');
CREATE OR ALTER TRIGGER TWT_AFTER_TRIGGER ON TWT AFTER UPDATE AS
BEGIN
    UPDATE TWT
    SET title = 'triggerExec'
    FROM TWT
    INNER JOIN deleted ON TWT.id = deleted.id;
END;
update TWT set title='DAB' output inserted.id,inserted.title where id=1;
```
3. `OUTPUT` clause used on a table with trigger configured works fine if used with `INTO` clause.

**Demonstration using T-SQL:**
```tsql
DROP TABLE IF EXISTS TWT;
create Table TWT(
id int PRIMARY KEY,
title varchar(50)
);
INSERT INTO TWT(id,title) values(1,'book');
[...]

PR #1630, opened 2023-08-13
Mutations on table with triggers - MsSql

ayush3797:
## Why make this change?
When there is a DML trigger configured on a table, we cannot use OUTPUT clause to return data. When an insert DML trigger is configured, OUTPUT clause cannot be used to return data because of the insert operation. When an update DML trigger is configured, OUTPUT clause cannot be used to return data because of the update operation. In such cases, we need to use a subsequent select statement to get the data after the mutation.

## Details
**What would handling of different kind of mutations look like when a trigger is enabled on a table?**

1. **DELETE:** We don't need to return any data for deletions, so nothing will change for deletions.

2. **UPDATE:** For updates, we will cease to use the output clause. Instead, we will do a subsequent select query on the table to get the data. Since PK would already be provided to us we can easily do a select query. Not to mention, we can do this only when we determine that there is indeed a trigger enabled on the table. If not, we can continue with the same query that we generate in present day.

3. **INSERT:** When it comes to insertions, we can perform insertions on two kind of tables:
     1. **Tables with non-auto generated primary keys:** When we have tables with non-autogen PK, we can insert values of non-autogen PK fields into a temporary table (we might not know value of every PK before hand because PK can have a default value as well) and then perform an inner join for the subsequent select on the actual table and the temporary table based on the values of these PK fields.
     2. **Tables with auto-gen primary keys:** Tables with an autogenerated PK column fall into this category.. For tables with autogen PK, we would not be aware of the PK beforehand and so we cannot directly perform the subsequent select query. To get the autogen primary key, we can use `SCOPE_IDENTITY()` method provided by Sql Server.. Once we get the autogen PK, in addition to having an inner join for non-autogen PKs as discussed in the previous approach, we can add an additional `WHERE` predicate `where autogen_PK_field = SCOPE_IDENTITY()`. Let us discuss more about this approach and the concerns:

**Delving deeper into [Approach#3](https://github.com/Azure/data-api-builder/issues/452#issuecomment-1639811191):**
We can use `OUTPUT` clause coupled with `INTO` clause to store the values for the primary keys in a temporary table. And then do a subsequent select query on the table to get the data using those primary keys. In the temporary table, we insert only those columns from the PK which are not autogenerated. Why? Because we cannot insert a value for an autogenerated column even in a temporary table.  Hence in the select query, we will add an additional WHERE predicate (alongwith the join) for the autogenerated column in the PK. It is to be noted that **MsSql supports only one IDENTITY/autogenerated column per table**.
We can choose to go with this approach only if we determine that there are any triggers enabled on the table.

Two assumptions which I am making here that I want to put ahead (btw, I don't think those are valid use cases) are:

1. **What if primary key column has a datatype of timestamp?** This concern arise because we cannot insert value for `timestamp` column (which is what the approach includes - inserting value via `INTO` clause). Well, `rowversion`/`timestamp` is intended for versioning and tracking changes within a table, rather than serving as a primary key. Also, by design, primary key values should not be modified after they have been assigned to a record. So having a PK with timestamp datatype does not seem to be a valid use case to me.

2. **What if the trigger updates the values of the primary keys** and the subsequent select query we are going to perform returns no data because the values of the PK has been changed in the table? This concern arises because technically, it is possible for triggers to update the value of PK columns. However, this is generally not recommended and can lead to undesirable consequences. Changing the value of a primary key can result in data integrity issues, as it may cause conflicts or break referential integrity constraints if other tables are referencing the primary key. It can also lead to confusion and difficulty in identifying records uniquely.

## What is this change?

1. Added properties `IsUpdateDMLTriggerEnabled` and `IsInsertDMLTriggerEnabled` to `SourceDefinition` class which if true will indicate that there is a corresponding DML trigger enabled on the table.
2. Added method `SqlMetadataProvider.PopulateTriggerMetadataForTable` whose overridden implementation in `MsSqlMetadataProvider` will populate the properties mentioned in (1) with appropriate metadata values.
3. Refactored how queries are built in the `SqlInsertStructure`, `SqlUpsertQueryStructure`, `SqlUpdateStructure` so that when trigger is configured for the mutation operation, a subsequent select query is used instead of the `OUTPUT` clause to return data.
4. In case of insertions in table with auto-gen PK, we get the latest value of the auto-gen column using `SCOPE_IDENTITY() ` function provided by sql server. Reference: https://learn.microsoft.com/en-us/sql/t-sql/functions/scope-identity-transact-sql?view=sql-server-ver16

## How was this tested?

- [x] Integration Tests - Added mutations tests to Rest/GQL performing inserts/updates/upserts on tables with insert/update DML trigger enabled. Tests are added for tables with autogen/non-autogen PK as the queries differ in the two cases.


zebleck:
Would be amazing if this could get fixed.

src/Core/Resolvers/IQueryBuilder.cs
```diff
@@ -64,6 +64,14 @@ public interface IQueryBuilder
         /// <returns></returns>
         public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName);
 
+        /// <summary>
+        /// Builds the query to determine the number of enabled triggers on a database table.
+        /// Needed only for MsSql.
+        /// </summary>
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        string GetQueryToGetEnabledTriggers() => throw new NotImplementedException();
```
seantleonard, line 73: I agree that you are 'building' a query here, however this query is explicitly related to metadata fetching. In my opinion, this would be better suited for the metadata provider, since there would presumably be database engine specific query text. Also, why add method that isn't used?

seantleonard:
In your PR description you state _"because of the {mutation/insert/update} operation."_

It would be helpful to what about the operation renders the OUTPUT clause incorrect

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -47,6 +47,8 @@ DROP TABLE IF EXISTS graphql_incompatible;
 DROP TABLE IF EXISTS GQLmappings;
 DROP TABLE IF EXISTS bookmarks;
 DROP TABLE IF EXISTS mappedbookmarks;
+DROP TABLE IF EXISTS fte_data;
+DROP TABLE IF EXISTS intern_data;
```
seantleonard, line 51: do you need to drop the triggers before SQL will let you drop the tables?

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
```diff
@@ -233,6 +251,81 @@ public async Task PutOneInViewBadRequest()
             await base.PutOneInViewBadRequest(expectedErrorMessage);
         }
 
+        /// <summary>
+        /// Test to validate that even when an update DML trigger is enabled on a table, we still return the
+        /// latest data as it is present after the trigger gets executed. Whenever an update DML trigger is enabled,
+        /// we use a subsequent SELECT query to get the data instead of using OUTPUT clause.
+        /// </summary>
```
seantleonard, line 258: It would be helpful to identify how the validation is done: ensured that even though salary was provided as -9, the response contains 0 due to the effect of the update trigger.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -302,6 +314,57 @@ public async Task InsertOneInViewBadRequestTest()
...
+        {
+            // Validate that insert operation succeeds when an insert DML trigger is enabled for a table
+            // with autogenerated primary key. The trigger updates the salary to max(0,min(100,salary)).
+            string requestBody = @"
+            {
+                ""name"": ""Joel"",
+                ""salary"": 102
+            }";
+
+            string expectedLocationHeader = $"id/{STARTING_ID_FOR_TEST_INSERTS}/u_id/2";
```
seantleonard, line 333: This test should also validate only a single record is updated when in the FTE table you have multiple records with `u_id = 2` becuase the definition in the table is `u_id int DEFAULT 2`. so for example:
id: 1 u_id=2 name=bob
id: 2 u_id=2 name janet

an update to pk (id:1, u_id2) does not also update the second record due. This validates your business logic where you add non-autogen PKs to a temp table and then join with the scope_identity value. (maybe this would only apply as a concern to the insert test.)

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -190,6 +208,78 @@ public async Task PatchOneViewBadRequestTest()
...
+        {
+            // Validate that PATCH operation (resulting in update) succeeds when an update DML trigger is configured for a table
+            // with autogenerated primary key. The trigger updates the salary to max(0,min(150,salary)).
+            string requestBody = @"
+            {
+                ""salary"": -9
+            }";
+
+            await SetupAndRunRestApiTest(
+                primaryKeyRoute: "id/1/u_id/2",
```
seantleonard, line 227: does this prevent updating multiple columns when more than one record has `u_id == 2`?

src/Core/Resolvers/MsSqlQueryBuilder.cs
seantleonard, line 1: It's a little difficult to visualize what these new queries will look like. it's unclear if examples of what you are generating for upsert/insert/update for auto/non-auto PK can be seen with your in-depth explanation here?: https://github.com/Azure/data-api-builder/issues/452#issuecomment-1639811191

It would be helpful to have visual of what each type of query would look like when built.

seantleonard (changes requested):
Before approving, it would be helpful to get some clarifying comments where suggested.

It would also help to have additional examples of what your newly built queries look like when triggers are involved.

For tests, it would be helpful to be more descriptive in what you are validating. (your descriptions for your ignoring extraneous fields in rest body is a good example of descriptive comments about validation).

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -349,6 +349,18 @@ private void LogPrimaryKeys()
         /// </summary>
         public abstract Type SqlToCLRType(string sqlType);
 
+        /// <summary>
+        /// Helper method to populate metadata about whether insert/update DML triggers are enabled for a table.
```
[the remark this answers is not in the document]
ayush3797: There's a difference between exist and enabled. There may be triggers that exist on the table but are disabled.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -40,5 +45,35 @@ public override Type SqlToCLRType(string sqlType)
         {
             return TypeHelper.GetSystemTypeFromSqlDbType(sqlType);
         }
+
+        public override async Task PopulateTriggerMetadataForTable(string schemaName, string tableName, SourceDefinition sourceDefinition)
+        {
+            string queryToGetNumberOfEnabledTriggers = SqlQueryBuilder.GetQueryToGetEnabledTriggers();
```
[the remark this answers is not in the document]
ayush3797: Thanks for the suggestion! I started with the count of triggers, but turns out we need to only use SELECT instead of OUTPUT when a trigger is enabled. Forgot to update the name.

src/Service.Tests/DatabaseSchema-MsSql.sql
ayush3797: No, we don't.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -40,5 +45,35 @@ public override Type SqlToCLRType(string sqlType)
...
+            Dictionary<string, DbConnectionParam> parameters = new()
+            {
+                { $"{BaseQueryStructure.PARAM_NAME_PREFIX}param0", new(schemaName, DbType.String) },
+                { $"{BaseQueryStructure.PARAM_NAME_PREFIX}param1", new(tableName, DbType.String) }
+            };
+
+            JsonArray? resultArray = await QueryExecutor.ExecuteQueryAsync(
+                sqltext: queryToGetNumberOfEnabledTriggers,
+                parameters: parameters!,
```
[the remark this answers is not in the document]
ayush3797: Its not. Thanks.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -63,27 +65,144 @@ public string Build(SqlQueryStructure structure)
...
+                // If there is a column in the PK which is autogenerated, we cannot and will not insert it into the temporary table.
+                // Hence in the select query, we will add an additional WHERE predicate (alongwith the join).
+                // for the autogenerated column in the PK.
+                // It is to be noted that MsSql supports only one IDENTITY/autogenerated column per table.
+
+                string tempTableName = $"{QuoteIdentifier(structure.DatabaseObject.SchemaName)}.{QuoteIdentifier($"#{structure.DatabaseObject.Name}_T")}";
+                (string autoGenPKColumn, List<string> nonAutoGenPKColumns) = GetSegregatedPKColumns(sourceDefinition);
+                if (nonAutoGenPKColumns.Count > 0)
+                {
+                    // Create temporary table containing zero rows and all the non-autogenerated columns present in the PK.
```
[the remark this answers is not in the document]
ayush3797: right

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -63,27 +65,144 @@ public string Build(SqlQueryStructure structure)
...
+                // Build the subsequent select query to return the inserted data.
+                StringBuilder subsequentSelect = new($"SELECT {MakeOutputColumns(structure.OutputColumns, tableName)} FROM {tableName} ");
+
+                if (nonAutoGenPKColumns.Count > 0)
+                {
+                    // We will perform inner join on the basis of all the non-autogenerated columns in the PK.
+                    string joinPredicates = string.Join(
+                        "AND ",
+                        nonAutoGenPKColumns.Select(pk => $"{tableName}.{QuoteIdentifier(pk)} = {tempTableName}.{QuoteIdentifier(pk)}"));
+                    subsequentSelect.Append($"INNER JOIN {tempTableName} ON {joinPredicates} ");
```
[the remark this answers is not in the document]
ayush3797: > (i see the next block of code appends an additional WHERE clause using scope_identity() so maybe that answer my question.

yes, this will help uniquely get a record.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -151,17 +287,50 @@ public string Build(SqlUpsertQueryStructure structure)
...
+
+                    insertQuery.Append($"OUTPUT {columnsToBeReturned}");
+                }
+                // If an insert trigger is enabled but there was no update trigger enabled,
+                // we need to generated columns to be returned without 'Inserted' prefix.
+                else if (!isUpdateTriggerEnabled)
+                {
+                    // This is again just an optimisation. If update trigger was enabled, then the columnsToBeReturned would
+                    // have already been created without any prefix.
+                    columnsToBeReturned = MakeOutputColumns(structure.OutputColumns, string.Empty);
```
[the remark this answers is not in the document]
ayush3797: Would still need to change the method signature because we need table_name (a string) in prefix here:
https://github.com/Azure/data-api-builder/blob/c3e1c32c0d12f53717a5bd9c01a4497ab52e6b38/src/Core/Resolvers/MsSqlQueryBuilder.cs#L127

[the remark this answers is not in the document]
ayush3797: Removing the None value from the enum

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -63,27 +65,144 @@ public string Build(SqlQueryStructure structure)
...
+                MakeOutputColumns(structure.OutputColumns, isUpdateTriggerEnabled ? "" : OutputQualifier.Inserted.ToString());
+
+            StringBuilder updateQuery = new($"UPDATE {tableName} SET {Build(structure.UpdateOperations, ", ")} ");
 
-            return $"UPDATE {QuoteIdentifier(structure.DatabaseObject.SchemaName)}.{QuoteIdentifier(structure.DatabaseObject.Name)} " +
-                    $"SET {Build(structure.UpdateOperations, ", ")} " +
-                    $"OUTPUT {MakeOutputColumns(structure.OutputColumns, OutputQualifier.Inserted)} " +
-                    $"WHERE {predicates};";
+            // If a trigger is enabled on the entity, we cannot use OUTPUT clause to return the record.
+            // In such a case, we will use a subsequent select query to get the record.
```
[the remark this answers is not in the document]
ayush3797: `because` won't be the right word here since its not the reason we are switching to using SELECT clause instead of OUTPUT clause.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
ayush3797: This is an insert test. Also this table has a composite PK (id,u_id) which uniquely identifies a record. To make an update, we need to provide value for every PK field. So, how can mutiple records get updated?

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
ayush3797: Not clear with the question. What do we need to prevent here? To uniquely identify a record, we are supplying the value for each of the PK field.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -130,19 +252,33 @@ public string Build(SqlUpsertQueryStructure structure)
             // Query to update record (if there exists one for given PK).
             StringBuilder updateQuery = new(
                 $"IF @ROWS_TO_UPDATE = 1 " +
+                $"BEGIN " +
```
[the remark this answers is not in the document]
ayush3797: This is used to execute multiple statements in an IF/ELSE block. Reference: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/if-else-transact-sql?view=sql-server-ver16#arguments

> To define a statement block, use the control-of-flow keywords BEGIN and END.

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
ayush3797: Added calrification to comment.

src/Core/Resolvers/IQueryBuilder.cs
```diff
@@ -64,6 +64,12 @@ public interface IQueryBuilder
         /// <returns></returns>
         public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName);
 
+        /// <summary>
+        /// Builds the query to determine the number of enabled triggers on a database table.
+        /// Needed only for MsSql.
+        /// </summary>
+        string BuildFetchEnabledTriggersQuery() => throw new NotImplementedException();
```
abhishekkumams, line 71: can be renamed to `BuildMsSqlFetchEnabledTriggersCountQuery()`
It will make it clear that the method is specific to MsSQL Server, and that it returns a SQL query to fetch the count of enabled triggers on a database table.

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
```diff
@@ -198,6 +198,24 @@ public class MsSqlPutApiTests : PutApiTestBase
                 $"SELECT [id], [revenue], [category], [accessible_role] FROM { _tableWithSecurityPolicy } " +
                 $"WHERE [id] = 1 AND [revenue] = 2000 AND [category] = 'anime' AND [accessible_role] = 'Anonymous' " +
                 $"FOR JSON PATH, INCLUDE_NULL_VALUES, WITHOUT_ARRAY_WRAPPER"
+            },
+            {
+                "PutOneUpdateInTableWithAutoGenPKAndTrigger",
```
abhishekkumams, line 203: `PutOneInsertInTableWithAutoGenPKAndTrigger`, this case?

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
```diff
@@ -233,6 +251,81 @@ public async Task PutOneInViewBadRequest()
...
 
+        /// <summary>
+        /// Test to validate that even when an update DML trigger is enabled on a table, we still return the
+        /// latest data as it is present after the trigger gets executed. Whenever an update DML trigger is enabled,
+        /// we use a subsequent SELECT query to get the data instead of using OUTPUT clause.
+        /// </summary>
+        [TestMethod]
+        public async Task PutOneUpdateInTableWithUpdateTrigger()
+        {
+            // Validate that PUT operation (resulting in update) succeeds when an update DML trigger is enabled for a table
```
abhishekkumams, line 262: This can be two different tests, would be easier to debug incase of failures.

src/Core/Resolvers/IQueryBuilder.cs
seantleonard: DB specific method names aren't appropriate for the generic interface. At some point PG / MySQL may need to utilize this. Even if not, a notImplementedException or other more verbose messaging and graceful error handling can be performed.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -295,15 +299,15 @@ public string Build(SqlUpsertQueryStructure structure)
                 {
                     if (isUpdateTriggerEnabled)
                     {
-                        // This is just an optimisation. If update trigger was not enabled, then the columnsToBeReturned would
-                        // have already been created with Inserted prefix.
+                        // This is just an optimisation.If update trigger is not enabled, then this build method has already created
```
seantleonard, line 302:
```suggestion
                        // This is just an optimisation. If update trigger is not enabled, then this build method has already created
```

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -63,27 +65,147 @@ public string Build(SqlQueryStructure structure)
...
+            if (!isInsertDMLTriggerEnabled)
+            {
+                // When there is no DML trigger enabled on the table for insert operation, we can use OUTPUT clause to return the data.
+                insertQuery.Append($"INSERT INTO {tableName} ({insertColumns}) OUTPUT " +
+                    $"{MakeOutputColumns(structure.OutputColumns, OutputQualifier.Inserted.ToString())} ");
+                insertQuery.Append(values);
+            }
+            else
+            {
+                // When a DML trigger for INSERT operation is enabled on the table, its a bit tricky to get the inserted data.
```
seantleonard, line 93:
```suggestion
                // When a DML trigger for INSERT operation is enabled on the table, it's a bit tricky to get the inserted data.
```

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -63,27 +65,147 @@ public string Build(SqlQueryStructure structure)
...
+                // We need to insert the values for all the non-autogenerated columns in the PK into a temporary table.
```
seantleonard, line 94:
```suggestion
                // We need to insert the values for all the non-autogenerated PK columns into a temporary table.
```

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -63,27 +65,147 @@ public string Build(SqlQueryStructure structure)
...
+                // Finally this temporary table will be used to do a subsequent select on the actual table where we would join the
+                // actual table and the temporary table based on the values of the non-autogenerated PK columns.
+                // If there is a column in the PK which is autogenerated, we cannot and will not insert it into the temporary table.
+                // Hence in the select query, we will add an additional WHERE predicate (alongwith the join).
```
seantleonard, line 98: additional WHERE predicate for the SELECT on the actual table? that would be a helpful clarification

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -302,6 +314,58 @@ public async Task InsertOneInViewBadRequestTest()
             );
         }
 
+        /// <summary>
+        /// Test to validate that even when an insert DML trigger is enabled on a table, we still return the
+        /// latest data as it is present after the trigger gets executed. Whenever an insert DML trigger is enabled,
+        /// we use a subsequent SELECT query to get the data instead of using OUTPUT clause. To validate that the data is returned
+        /// as it is after the trigger is executed, we use the values which are updated by the trigger in the WHERE predicates of the verifying sql query.
```
seantleonard, line 321: i don't remember if i asked for this pr, but would be helpful to describe why both autogen/non-autogen pk tests are needed. e.g. specific branching logic in mssqlquerybuilder that handles each case?

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
seantleonard: ok seems i posted on wrong test

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -190,6 +208,79 @@ public async Task PatchOneViewBadRequestTest()
             await base.PatchOneViewBadRequestTest(expectedErrorMessage);
         }
 
+        /// <summary>
+        /// Test to validate that even when an update DML trigger is enabled on a table, we still return the
+        /// latest data as it is present after the trigger gets executed. Whenever an update DML trigger is enabled,
+        /// we use a subsequent SELECT query to get the data instead of using OUTPUT clause. To validate that the data is returned
+        /// as it, is after the trigger is executed, we use the values which are updated by the trigger in the WHERE predicates of the verifying sql query.
```
seantleonard, line 215:
```suggestion
        /// after the trigger is executed, we use the new values (written by the trigger) in the WHERE predicates of the verifying sql query.
```
phrasing awkward, i think this captures intent?

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
seantleonard: i think i cleared up  my understand by re-reading mssqlquerybuilder implementation/new comments you added. I was confused about the implications of having auto-gen pk in an added where clause on the main table and the non-autogen pk's existing in the temp table.

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -190,6 +208,79 @@ public async Task PatchOneViewBadRequestTest()
...
+                sqlQuery: GetQuery("PatchOneUpdateInTableWithNonAutoGenPKAndTrigger"),
+                operationType: EntityActionOperation.UpsertIncremental,
+                requestBody: requestBody,
+                expectedStatusCode: HttpStatusCode.OK
+            );
+        }
+
+        /// <summary>
+        /// Test to validate that even when an insert DML trigger is enabled on a table, we still return the
+        /// latest data as it is present after the trigger gets executed. Whenever an insert DML trigger is enabled,
```
seantleonard, line 257:
```suggestion
        /// latest data (values changed by trigger execution). Whenever an insert DML trigger is enabled,
```

seantleonard (commented):
Added more feedback. Thanks for updating so far, i think the most helpful thing now would be to get a few examples of the queries you generate with you business logic in mssqlquerybuilder. i dont' think the issue RFC captures all these examples.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -151,17 +291,51 @@ public string Build(SqlUpsertQueryStructure structure)
...
                 // Query to insert record (if there exists none for given PK).
-                StringBuilder insertQuery = new($"INSERT INTO {tableName} ({insertColumns}) OUTPUT {outputColumns}");
-
+                StringBuilder insertQuery = new($"INSERT INTO {tableName} ({insertColumns}) ");
+
+                bool isInsertTriggerEnabled = sourceDefinition.IsInsertDMLTriggerEnabled;
+                // We can only use OUTPUT clause to return inserted data when there is no trigger enabled on the entity.
+                if (!isInsertTriggerEnabled)
+                {
+                    if (isUpdateTriggerEnabled)
```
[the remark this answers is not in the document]
ayush3797: Yes we can. The tables `FteData` and `InternData` used for testing have both of these triggers enabled.

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
ayush3797: You cannot Insert via PUT in a table with autogen PK.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -151,17 +291,51 @@ public string Build(SqlUpsertQueryStructure structure)
...
+                    {
+                        // This is just an optimisation.If update trigger is not enabled, then this build method has already created
```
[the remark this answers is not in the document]
ayush3797: not enabled.

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
ayush3797: The intent of the test is to validate behavior of PUT update on tables. Autogen/Non-autogen are just two different cases. I think we should keep this as is.

severussundar (approved):
LGTM! Kudos for the detailed explanation about different approaches investigated/considered in the issue comments and the PR description.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -63,27 +65,170 @@ public string Build(SqlQueryStructure structure)
...
+            // If a trigger is enabled on the entity, we cannot use OUTPUT clause to return the record.
+            // In such a case, we will use a subsequent select query to get the record. By the time the subsequent select executes,
+            // the trigger would have already executed and we get the data as it is present in the table.
+            if (isUpdateTriggerEnabled)
+            {
+                updateQuery.Append($"WHERE {predicates};");
+                updateQuery.Append($"SELECT {columnsToBeReturned} FROM {tableName} WHERE {predicates};");
+                /* An example final update query on a table with update trigger enabled would look like:
+                 *  UPDATE [dbo].[table] SET [dbo].[table].[field1] = @param3
+                 *  WHERE ([field3] != @param0) AND [dbo].[table].[id] = @param1 AND [dbo].[table].[nonautogen_id] = @param2;
```
seantleonard, line 220: `([field3] != @param0)` is an example of a predicate representing the DB Policy?

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -206,6 +206,100 @@ public async Task TestStoredProcedureMutationForDeletion(string dbQueryToVerifyD
...
+                    createInternData(item: { id: 4, name: ""Tommy"", salary: 45 }) {
+                        id
+                        months
+                        name
+                        salary
+                    }
+                }
+            ";
+
+            JsonElement actual = await ExecuteGraphQLRequestAsync(graphQLMutation, graphQLMutationName, isAuthenticated: true, new() { { "title", "My New Book" }, { "publisher_id", 1234 } });
```
seantleonard, line 228: How does this test past as-is when variables unrelated to `createInternData` are provided?

`{ "title", "My New Book" }, { "publisher_id", 1234 }`

seantleonard: this is present for the next handfull of tests. If it is as simple as removing, can you follow up with why their presence didn't fail the tests?

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -206,6 +206,100 @@ public async Task TestStoredProcedureMutationForDeletion(string dbQueryToVerifyD
             Assert.AreEqual(updatedResult.RootElement.GetProperty("maxId").GetInt64(), 13);
         }
 
+        public async Task InsertMutationOnTableWithTriggerWithNonAutoGenPK(string dbQuery)
+        {
+            // Validate that insert operation succeeds when an insert DML trigger is enabled for a table with non-autogenerated primary key.
+            // For non-autogenerated PK fields, the values are inserted into a temporary table and then an INNER JOIN is performed
+            // based on those fields' values from the actual table and the temporary table.
+            // Trigger behavior: The trigger updates the salary to max(0,min(30,salary)).
+            // We would not need to specify the value of the PK column 'months' as it has a default value of 2.
```
seantleonard, line 215: You didn't need to explain how the backend logic works. A more helpful comment that describes this test's validation would cover:
- given input `item: { id: 4, name: ""Tommy"", salary: 45 }` 
- this test proves that the selection set would return: `id: 4, months: 2, name: Tommy, salary: 30`
- Because the selection set shows `salary:30` that proves the insert trigger executed and the selection set included the value modified by the trigger.

seantleonard: You can extrapolate this type of test description to the other related tests below.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -378,6 +390,65 @@ public async Task InsertOneInViewBadRequestTest()
...
+        [TestMethod]
+        public async Task InsertOneInTableWithInsertTrigger()
+        {
+            // Validate that insert operation succeeds when an insert DML trigger is enabled for a table
+            // with autogenerated primary key. For table containing an autogenerated PK field (id), and non-autogenerated PK field (u_id),
+            // the values of the non-autogenerated PK fields are inserted into a temporary table and then an INNER JOIN is performed
+            // based on those fields' values from the actual table and the temporary table. For the autogenerated PK field, an additional
+            // WHERE predicate is added to fetch the unique record by extracting the value of the autogenerated PK field using the
+            // SCOPE_IDENTITY() method provided by Sql Server.
+            // Trigger behavior: The trigger updates the salary to max(0,min(100,salary)).
```
seantleonard, line 408: Refer to comment https://github.com/Azure/data-api-builder/pull/1630#discussion_r1325210147

src/Service.Tests/SqlTests/RestApiTests/Patch/MsSqlPatchApiTests.cs
```diff
@@ -313,6 +331,79 @@ public virtual async Task PatchOneWithRowversionFieldInRequestBody()
...
+        /// <summary>
+        /// Test to validate that even when an update DML trigger is enabled on a table, we still return the
+        /// latest data as it is present after the trigger gets executed. Whenever an update DML trigger is enabled,
+        /// we use a subsequent SELECT query to get the data instead of using OUTPUT clause. To validate that the data is returned
+        /// after the trigger is executed, we use the new values (written by the trigger) in the WHERE predicates of the verifying sql query.
+        /// </summary>
+        [TestMethod]
+        public async Task PatchOneUpdateInTableWithUpdateTrigger()
+        {
+            // Validate that PATCH operation (resulting in update) succeeds when an update DML trigger is configured for a table
```
seantleonard, line 343: Refer to https://github.com/Azure/data-api-builder/pull/1630#discussion_r1325210147

src/Service.Tests/SqlTests/RestApiTests/Put/MsSqlPutApiTests.cs
```diff
@@ -364,6 +382,83 @@ public virtual async Task PutOneWithRowversionFieldInRequestBody()
...
                 );
         }
+        /// <summary>
+        /// Test to validate that even when an update DML trigger is enabled on a table, we still return the
+        /// latest data as it is present after the trigger gets executed. Whenever an update DML trigger is enabled,
+        /// we use a subsequent SELECT query to get the data instead of using OUTPUT clause. To validate that the data is returned
+        /// as it is after the trigger is executed, we use the values which are updated by the trigger in the WHERE predicates of the verifying sql query.
+        /// </summary>
+        [TestMethod]
+        public async Task PutOneUpdateInTableWithUpdateTrigger()
```
seantleonard, line 392: refer to https://github.com/Azure/data-api-builder/pull/1630#discussion_r1325210147

seantleonard (changes requested):
I wrote some feedback on what a more helpful test description would look like. Instead of just saying 'validate that request succeeds' and what happens in the engine, you should explain the significance of the expected results. e.g. The returned Selection set field X has value Y because trigger Z changed the value.

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
ayush3797: The variables were not referenced anywhere in the gql mutation and hence their absence/presence does not make any difference. Thanks for catching!

src/Core/Resolvers/MsSqlQueryBuilder.cs
ayush3797: Right!

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
ayush3797: Thanks for the suggestions! I have updated the comments/summaries.

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -206,6 +206,100 @@ public async Task TestStoredProcedureMutationForDeletion(string dbQueryToVerifyD
...
+        public async Task UpdateMutationOnTableWithTriggerWithNonAutoGenPK(string dbQuery)
+        {
+            // Given input item { salary: 100 }, this test verifies that the selection would return salary = 50.
+            // Thus confirming that we return the data being updated by the trigger where,
+            // the trigger behavior is that it updates the salary to max(0,min(50,salary)).
+            string graphQLMutationName = "updateInternData";
+            string graphQLMutation = @"
+                mutation{
+                    updateInternData(id: 1, months: 3, item: { salary: 100 }) {
+                        id
```
seantleonard, line 265: The dbQuery passed in has the following predicates:
```sql
                WHERE [table0].[id] = 1
                    AND [table0].[u_id] = 2
                    AND [table0].[salary] = 0
```

how does this test pass?

seantleonard: nvm. looked at wrong query.

seantleonard (approved):
Thanks for bearing with me and addressing feedback. The code comments are very helpful and appreciated and will be useful whenever we read this code in the future

ayush3797:
> Thanks for bearing with me and addressing feedback. The code comments are very helpful and appreciated and will be useful whenever we read this code in the future

Happy to learn and get better :)

PR #1636, opened 2023-08-16
Handling datetime types

ayush3797:
## Why make this change?
Currently, we are incorrectly handling the datetime datatypes. A snippet from `BaseSqlQueryStructure` class:

image

As evident from the highlighted portion, even for `DateTime` .NET type, we are parsing it with `DateTimeOffset` which is wrong. Just for an example, if you execute a GET request on the `stocks_price` table which has `(categoryid,pieceid,instant)` as PK, you would not get any result back even when the record exists for that PK in the table. Why? Because even though instant is a `datetime` column, it gets parsed as `datetimeoffset` column. This is shown in the next 2 images.

image
image

Also, the existing tests (like the ones fixed in `MsSqlGraphQLPaginationTests` and `GraphQLSupportedTypesTestsBase`) were wrongly written, in which values in the range of `datetime2` were supplied for `datetime` column.

## What is this change?
1. Correctly return value using `DateTime` field of the `DateTimeOffset` parsed value in `BaseSqlQueryStructure` class,
2. Populate the DbType of the datetime parameters when they are sent to the database so that the database knows the exact type of the parameter. Eg. by default, if you send a `datetime2` param without a DbType, it is assumed to be a `datetime` and an exception is thrown incorrectly by the database. This is done using the map `TypeHelper._timeSqlDbTypeToDbType`.
It should be noted 4 sql server types - `date`, `smalldatetime`, `datetime`, `datetime2` map to the same .NET system type `System.DateTime`. The existing `TypeHelper._systemTypeToDbTypeMap` would not be of any use to us because it has mapping from system type to DbType and all the above 3 sql server types have the same system type. So, we need to rely on the actual mapping from sql server type to DbType which is:
 `datetime -> DbType.DateTime, smalldatetime -> DbType.DateTime, datetime2 -> DbType.DateTime2`
`date -> DbType.Date`

3. Fix the existing tests.

## How was this tested?

- [x] Integration Tests - Added test for getting single record by PK having a `datetime` column as a PK column to `FindApiTestBase`.
- [x] Unit Tests - Added filter/orderby and insert tests to `GraphQLSupportedTypesTestsBase.QueryTypeColumnFilterAndOrderByDateTime` and `GraphQLSupportedTypesTestsBase.InsertIntoTypeColumn`for datetime types - `datetime2`, `smalldatetime`, `date`.

## Sample Request(s)

- Example REST request to demonstrate modifications
image

- Example GQL request to demonstrate modifications
image


src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -360,7 +360,7 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
                 "Double" => double.Parse(param),
                 "Decimal" => decimal.Parse(param),
                 "Boolean" => bool.Parse(param),
-                "DateTime" => DateTimeOffset.Parse(param),
+                "DateTime" => DateTime.Parse(param),
```
abhishekkumams, line 363: This is a bit of a tricky one. `DateTime` type is handled by HotChocolate as `DateTimeOffset`. Also, our db column type detection method gives `DateTime` for all `Date`, `DateTime`, and `DateTimeOffset`. Meaning that while the user can define non `DateTimeOffset` columns, they will be interpreted as `DateTimeOffset` and the gaps will be filled with 00:00:00 time and local offset.

Here is the Initial PR: https://github.com/Azure/data-api-builder/commit/63d5ef24e3bb90057fed015caa1d81f222de232a

src/Service.Tests/SqlTests/GraphQLPaginationTests/MsSqlGraphQLPaginationTests.cs
```diff
@@ -43,7 +43,7 @@ public static async Task SetupAsync(TestContext context)
             DisplayName = "Test after token for decimal values.")]
         [DataRow("boolean_types", "false", "true", 2, 4,
             DisplayName = "Test after token for boolean values.")]
-        [DataRow("date_types", "\"0001-01-01\"",
+        [DataRow("date_types", "\"1753-01-01\"",
```
abhishekkumams, line 46: why this change? The range of SQL Date data type is from January 1, 0001, through December 31, 9999.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -167,10 +167,10 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
         [DataRow(DATETIME_TYPE, "gt", "\'1999-01-08\'", "\"1999-01-08\"", " > ")]
         [DataRow(DATETIME_TYPE, "gte", "\'1999-01-08\'", "\"1999-01-08\"", " >= ")]
-        [DataRow(DATETIME_TYPE, "lt", "\'0001-01-01\'", "\"0001-01-01\"", " < ")]
-        [DataRow(DATETIME_TYPE, "lte", "\'0001-01-01\'", "\"0001-01-01\"", " <= ")]
-        [DataRow(DATETIME_TYPE, "neq", "\'0001-01-01\'", "\"0001-01-01\"", "!=")]
-        [DataRow(DATETIME_TYPE, "eq", "\'0001-01-01\'", "\"0001-01-01T01:01:01\"", "=")]
+        [DataRow(DATETIME_TYPE, "lt", "\'1753-01-01\'", "\"1753-01-01\"", " < ")]
+        [DataRow(DATETIME_TYPE, "lte", "\'1753-01-01\'", "\"1753-01-01\"", " <= ")]
+        [DataRow(DATETIME_TYPE, "neq", "\'1753-01-01\'", "\"1753-01-01\"", "!=")]
+        [DataRow(DATETIME_TYPE, "eq", "\'1753-01-01\'", "\"1753-01-01\"", "=")]
```
abhishekkumams, line 173: can you add tests for DateTime2 as well

src/Service.Tests/SqlTests/GraphQLPaginationTests/MsSqlGraphQLPaginationTests.cs
ayush3797: Yeah, corrected. Thanks!

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
ayush3797: done

abhishekkumams:
please check once that filter queries with GraphQL are working for dateTime/dateTime2 and DateTimeOffset

ayush3797:
> please check once that filter queries with GraphQL are working for dateTime/dateTime2 and DateTimeOffset

Added tests for `smalldatetime`, `date`, `datetime2` types which are specific to MsSql. For datetime, datetimeoffset, the tests are already added present in test `GraphQLSupportedTypesTestBase.QueryTypeColumnFilterAndOrderByDateTime`.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -171,40 +133,66 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
+            DisplayName = "datetime type filter and orderby test with neq operator and specific value '1999-01-08 10:23:00'.")]
+        [DataRow(DATETIME_TYPE, "eq", "'1999-01-08 10:23:00'", "\"1999-01-08 10:23:00\"", "=",
+            DisplayName = "datetime type filter and orderby test with eq operator and specific value '1999-01-08 10:23:00'.")]
+        [DataRow(DATETIME_TYPE, "lte", "'2079-06-06'", "\"2079-06-06\"", " <= ",
+            DisplayName = "datetime type filter and orderby test with lte operator and specific value '2079-06-06'.")]
+        [DataRow(DATETIME_TYPE, "neq", "'1999-01-08 10:23:54'", "\"1999-01-08 10:23:54\"", "!=",
+            DisplayName = "datetime type filter and orderby test with neq operator and specific value '1999-01-08 10:23:54'.")]
+        [DataRow(DATETIMEOFFSET_TYPE, "neq", "'1999-01-08 10:23:54.9999999-14:00'", "\"1999-01-08 10:23:54.9999999-14:00\"", "!=",
+            DisplayName = "datetimeoffset type filter and orderby test with neq operator")]
+        [DataRow(DATETIMEOFFSET_TYPE, "eq", "'1999-01-08 10:23:54.9999999-14:00'", "\"1999-01-08 10:23:54.9999999-14:00\"", "=",
```
Aniruddh25, line 164: Is there a test testing max value for DATETIMEOFFSET_TYPE?

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -289,19 +277,25 @@ public async Task TestTimeTypePrecisionCheck(string gqlValue, int count)
...
+        [DataRow(BYTEARRAY_TYPE, "null")]
         [DataRow(TIME_TYPE, "\"23:59:59.9999999\"")]
         [DataRow(TIME_TYPE, "\"23:59:59\"")]
         [DataRow(TIME_TYPE, "\"23:59:59.9\"")]
         [DataRow(TIME_TYPE, "\"23:59\"")]
         [DataRow(TIME_TYPE, "null")]
-        [DataRow(BYTEARRAY_TYPE, "\"V2hhdGNodSBkb2luZyBkZWNvZGluZyBvdXIgdGVzdCBiYXNlNjQgc3RyaW5ncz8=\"")]
-        [DataRow(BYTEARRAY_TYPE, "null")]
+        [DataRow(DATETIME_TYPE, "\"1999-01-08 09:20:00\"")]
+        [DataRow(DATETIME_TYPE, "\"1999-01-08\"")]
```
Aniruddh25, line 289: Is there a test for max value of DATETIME type insertion?

Aniruddh25 (changes requested):
Thanks for identifying this and picking it up! Questions on some missing tests..

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -361,8 +360,8 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
                 "Double" => double.Parse(param),
                 "Decimal" => decimal.Parse(param),
                 "Boolean" => bool.Parse(param),
-                "DateTime" => DateTimeOffset.Parse(param, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
-                "DateTimeOffset" => DateTimeOffset.Parse(param, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
+                "DateTime" => DateTime.Parse(param),
+                "DateTimeOffset" => DateTimeOffset.Parse(param),
```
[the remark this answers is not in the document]
ayush3797: No, while feeding the value to the database, we don't need it.

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -1209,6 +1209,14 @@
         "enabled": true
       },
       "permissions": [
+        {
```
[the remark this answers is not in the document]
ayush3797: There isn't.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/PostgreSqlGQLSupportedTypesTests.cs
```diff
@@ -54,6 +54,9 @@ protected override bool IsSupportedType(string type)
             return type switch
             {
                 BYTE_TYPE => false,
+                DATE_TYPE => false,
```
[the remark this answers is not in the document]
ayush3797: Created issues for MySql/PgSql here: https://github.com/Azure/data-api-builder/issues/1643, https://github.com/Azure/data-api-builder/issues/1642

config-generators/mysql-commands.txt
```diff
@@ -87,7 +87,7 @@ update BookWebsitePlacement --config "dab-config.MySql.json" --permissions "auth
 update BookWebsitePlacement --config "dab-config.MySql.json" --permissions "authenticated:delete" --fields.include "*" --policy-database "@claims.userId eq @item.id"
 update Author --config "dab-config.MySql.json"  --permissions "authenticated:create,read,update,delete" --rest true --graphql true
 update WebsiteUser --config "dab-config.MySql.json" --permissions "authenticated:create,read,delete,update" --rest false --graphql "websiteUser:websiteUsers"
-update stocks_price --config "dab-config.MySql.json" --permissions "authenticated:create,read,delete,update" --rest false
+update stocks_price --config "dab-config.MySql.json" --permissions "authenticated:create,read,delete,update" --rest true
```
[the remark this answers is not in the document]
ayush3797: This change is to remove any differences between the 3 dbs. MySql already had read permissions for anonymous role. No all 3 dbs have same permissions for anonymous role.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -16,64 +16,26 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.GraphQLSupportedTypesTests
...
+        [DataRow(STRING_TYPE, 4, DisplayName = "Query by PK test selecting only string_types.")]
+        [DataRow(BOOLEAN_TYPE, 1, DisplayName = "Query by PK test selecting only boolean_types.")]
+        [DataRow(DATETIME_TYPE, 2, DisplayName = "Query by PK test selecting only datetime_types.")]
+        [DataRow(DATETIME2_TYPE, 3, DisplayName = "Query by PK test selecting only datetime2_types.")]
+        [DataRow(DATETIMEOFFSET_TYPE, 4, DisplayName = "Query by PK test selecting only datetimeoffset_types.")]
+        [DataRow(SMALLDATETIME_TYPE, 4, DisplayName = "Query by PK test selecting only smalldatetime_types.")]
+        [DataRow(DATE_TYPE, 1, DisplayName = "Query by PK test selecting only date_types.")]
+        [DataRow(TIME_TYPE, 2, DisplayName = "Query by PK test selecting only time_types.")]
+        [DataRow(BYTEARRAY_TYPE, 3, DisplayName = "Query by PK test selecting only bytearray_types.")]
+        [DataRow(GUID_TYPE, 4, DisplayName = "Query by PK test selecting only guid_types.")]
```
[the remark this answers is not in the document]
ayush3797: I see. Added those back.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
ayush3797: added

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
ayush3797: added now

abhishekkumams:
is it working with 'Z' notation for datetime in filter queries for graphQL?

ayush3797:
> is it working with 'Z' notation for datetime in filter queries for graphQL?

When DbType for these columns are populated in the parameters sent to the database (which is necessary to provide for MsSql to avoid unexpected behaviors) and when you want to put filters on datetime, datetime2, date, smalldatetime, datetimeoffset, you need to provide the value for the column which is present in the database.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
abhishekkumams: Is there tests for the same?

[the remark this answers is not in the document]
ayush3797: Yes.

Aniruddh25:
> > is it working with 'Z' notation for datetime in filter queries for graphQL?

When DbType for datetime, datetime2, date, smalldatetime, datetimeoffset columns are populated in the parameters sent to the database (which is necessary to provide for MsSql to avoid unexpected behaviors) and when you want to put filters on these columns you need to provide the value for the column which is present in the database i.e. the Z notation would not work. 

When passing the filter value without the 'Z' notation through HotChocolate, will HotChocolate detect it as DateTimeType or string?

ayush3797:
> > When DbType for datetime, datetime2, date, smalldatetime, datetimeoffset columns are populated in the parameters sent to the database (which is necessary to provide for MsSql to avoid unexpected behaviors) and when you want to put filters on these columns you need to provide the value for the column which is present in the database i.e. the Z notation would not work.
> 
> When passing the filter value without the 'Z' notation through HotChocolate, will HotChocolate detect it as DateTimeType or string?

Does not matter how you pass it, with or without Z, the way we parse it, it will always take it up as a string.

image

Aniruddh25:
> > > When DbType for datetime, datetime2, date, smalldatetime, datetimeoffset columns are populated in the parameters sent to the database (which is necessary to provide for MsSql to avoid unexpected behaviors) and when you want to put filters on these columns you need to provide the value for the column which is present in the database i.e. the Z notation would not work.
> > 
> > 
> > When passing the filter value without the 'Z' notation through HotChocolate, will HotChocolate detect it as DateTimeType or string?
> 
> Does not matter how you pass it, with or without Z, the way we parse it, it will always take it up as a string.
> 
> image

Gotcha, will passing value with 'Z' throw an error now unless the datatype is datetimeoffset on the db? Is this a breaking change now?

ayush3797:
> > > > When DbType for datetime, datetime2, date, smalldatetime, datetimeoffset columns are populated in the parameters sent to the database (which is necessary to provide for MsSql to avoid unexpected behaviors) and when you want to put filters on these columns you need to provide the value for the column which is present in the database i.e. the Z notation would not work.
> > > 
> > > 
> > > When passing the filter value without the 'Z' notation through HotChocolate, will HotChocolate detect it as DateTimeType or string?
> > 
> > 
> > Does not matter how you pass it, with or without Z, the way we parse it, it will always take it up as a string.
> > image
> 
> Gotcha, will passing value with 'Z' on the filter e.g. '1999-01-08 10:23:00Z' for a column with datetime(2) type throw an error? will it throw error if the datatype is datetimeoffset on the db? Is this a breaking change?

It won't throw an error in the case of datetime2 but would not return any records because of loss of precision in the Z notation.
For datetimeoffset, yes it will throw an error.

abhishekkumams:
Can you update the docs

> > > > > When DbType for datetime, datetime2, date, smalldatetime, datetimeoffset columns are populated in the parameters sent to the database (which is necessary to provide for MsSql to avoid unexpected behaviors) and when you want to put filters on these columns you need to provide the value for the column which is present in the database i.e. the Z notation would not work.
> > > > 
> > > > 
> > > > When passing the filter value without the 'Z' notation through HotChocolate, will HotChocolate detect it as DateTimeType or string?
> > > 
> > > 
> > > Does not matter how you pass it, with or without Z, the way we parse it, it will always take it up as a string.
> > > image
> > 
> > 
> > Gotcha, will passing value with 'Z' on the filter e.g. '1999-01-08 10:23:00Z' for a column with datetime(2) type throw an error? will it throw error if the datatype is datetimeoffset on the db? Is this a breaking change?
> 
> It won't throw an error in the case of datetime2 but would not return any records because of loss of precision in the Z notation. For datetimeoffset, yes it will throw an error.

can you update the docs, I have created a PR which contains limitations/comments for every supported type in Different DBs, you can add the details so that users are aware about this. mentioning the values to be used in filters vs the values we get from HotChocolate.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
ayush3797: This is not required when feeding the values to the database. Hotchocolate converts date/time to UTC while returning the parsed result but not for values that are feeded for parameters.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -361,8 +360,8 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
                 "Double" => double.Parse(param),
                 "Decimal" => decimal.Parse(param),
                 "Boolean" => bool.Parse(param),
-                "DateTime" => DateTimeOffset.Parse(param, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
-                "DateTimeOffset" => DateTimeOffset.Parse(param, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
+                "DateTime" => DateTime.Parse(param),
```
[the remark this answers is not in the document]
ayush3797: This is the bug. Why would we parse `DateTime` value with `DateTimeOffset` at the first place? This was wrong all along.
For Abhishek's thread - `DateTime type is handled by HotChocolate as DateTimeOffset` - Hotchocolate converts date/time to UTC while returning the parsed result but not for values that are feeded for parameters.

config-generators/mysql-commands.txt
```diff
@@ -87,7 +87,7 @@ update BookWebsitePlacement --config "dab-config.MySql.json" --permissions "auth
 update BookWebsitePlacement --config "dab-config.MySql.json" --permissions "authenticated:delete" --fields.include "*" --policy-database "@claims.userId eq @item.id"
 update Author --config "dab-config.MySql.json"  --permissions "authenticated:create,read,update,delete" --rest true --graphql true
 update WebsiteUser --config "dab-config.MySql.json" --permissions "authenticated:create,read,delete,update" --rest false --graphql "websiteUser:websiteUsers"
-update stocks_price --config "dab-config.MySql.json" --permissions "authenticated:create,read,delete,update" --rest false
```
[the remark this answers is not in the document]
ayush3797: No tests were present for this entity before. I added one test. Rest was disabled for this entity and hence I had to enable it.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -40,5 +43,20 @@ public override Type SqlToCLRType(string sqlType)
         {
             return TypeHelper.GetSystemTypeFromSqlDbType(sqlType);
         }
+
+        /// <inheritdoc/>
+        public override DbType? GetDbTypeForDateTimeTypes(string sqlDateTimeType)
```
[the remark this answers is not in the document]
ayush3797: updated

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -307,14 +307,20 @@ private void LogPrimaryKeys()
                 // row["DATA_TYPE"] has value type string so a direct cast to System.Type is not supported.
                 // See https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings
                 Type systemType = SqlToCLRType((string)row["DATA_TYPE"]);
+                ParameterDefinition paramDefinition = new()
+                {
+                    SystemType = systemType,
+                    DbType = TypeHelper.GetDbTypeFromSystemType(systemType)
+                };
+
+                DbType? dbTypeForMsSqlDateTimeType = GetDbTypeForDateTimeTypes((string)row["DATA_TYPE"]);
```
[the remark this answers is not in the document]
ayush3797: updated.

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -1209,6 +1209,14 @@
...
+          "role": "anonymous",
+          "actions": [
+            {
+              "action": "read"
+            }
+          ]
+        },
```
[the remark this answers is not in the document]
ayush3797: Is this required at this point in time since all tests are passing/ are not impacted - meaning they don't really have a dependency on this configuration?

src/Service.Tests/SqlTests/RestApiTests/RestApiTestBase.cs
```diff
@@ -57,6 +57,8 @@ public abstract class RestApiTestBase : SqlTestBase
         protected static readonly string _Composite_NonAutoGenPK_EntityPath = "commodities";
         protected static readonly string _integrationTypeEntity = "SupportedType";
         protected static readonly string _integrationTypeTable = "type_table";
+        protected static readonly string _entityWithDateTimePK = "stocks_price";
+        protected static readonly string _tableWithDateTimePK = "stocks_price";
```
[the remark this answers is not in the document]
ayush3797: updated

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
abhishekkumams: I added it because when inserting Hotchocolate was converting it to local timezone.
image

can you check if after your changes something like this is not happening. Even if it's not happening, i feel it's good to have it explicitly in the code for better clarity that we want it to assume it's in UTC and not to apply any conversion.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
abhishekkumams: I added it because when inserting Hotchocolate was converting it to local timezone.
image

can you check if after your changes something like this is not happening. Even if it's not happening, i feel it's good to have it explicitly in the code for better clarity that we want it to assume it's in UTC and not to apply any conversion.

ayush3797:
Putting this PR on hold since after Abhishek's PR : https://github.com/Azure/data-api-builder/pull/1473, the get by pk for REST is working fine. Moreover, since we want to continue to support the Z notation, for filters, insertions and read operations, we would still need to continue parsing datetime, datetime2, smalldatetime (which all map to System.DateTime .NET type) as DateTimeOffset because when we parse these values as DateTime, we cannot support Z notation without adding the offset in the local time.

For a create mutation like this:

image

parsing with DateTime gives:

image

So an offset of local time (+5:30 in IST 10:23->15:53) is added when a value is supplied with 'Z' notation.


The same does not happen when we use DateTimeOffset to parse the value:
image.

So, bottom line is if we want to continue to support Z notation, we have to continue parsing all the datetime values using DateTimeOffset because when we parse using DateTime the value is parsed by adding offset for locale if a value is supplied with Z, which is not what we want.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
ayush3797: Confirmed. Not happening.

ayush3797:
## UPDATE:
We can continue parsing datetime values using `DateTimeOffset` class and thus continue the support for z notation. However, after parsing the datetime value, we will return the `DateTime` field of the `DateTimeOffset` object i.e. `parsedDateTimeOffsetObj.DateTime`. This field does not take into account the offset in the input datetime value. So, it does not matter if the z notation is present in the datetime value or not. It will always be ignored. _**So, we can have the same parsing logic for both GQL and REST.**_

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
ayush3797: No, this is not happening. Thanks for pointing out.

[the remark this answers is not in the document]
seantleonard: Can this be a test? that way we don't need to remember to manually validate in the future

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -26,6 +27,8 @@ public class MsSqlMetadataProvider :
         {
         }
 
+        public static readonly HashSet<string> DateTimeTypes = new() { "date", "smalldatetime", "datetime", "datetime2", "datetimeoffset" };
```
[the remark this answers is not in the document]
seantleonard: @ayush2797, is this still used? i don't see any references.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -382,9 +375,24 @@ private void LogPrimaryKeys()
...
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Takes a string version of a sql date/time type and returns its corresponding DbType.
+        /// Returns null when the provided sql type is not a date/time type.
         /// </summary>
-        public abstract DbType DatetimeSqlTypeToDbype(string sqlType);
+        /// <param name="sqlDbTypeName">Name of the SqlDb type.</param>
+        public virtual DbType? SqlDbTypeToDbType(string sqlDbTypeName) => null;
```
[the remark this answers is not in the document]
seantleonard: I see it's usage is specific to mssql. Perhaps a more generic name then because it doesn't really make sense to have this in pgsql and mysql. I only say that because `SqlDbType` is specific to mssql.

Maybe `ResolveDbType(string s)` then when it comes to PG and MySQL we can find the mapping between their specific types and the `DbType` enum.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -173,40 +186,68 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
+        [DataRow(DATETIME2_TYPE, "neq", "'9999-12-31 23:59:59'", "\"9999-12-31 23:59:59\"", "!=",
+            DisplayName = "datetime2 type filter and orderby test with neq operator")]
         public async Task QueryTypeColumnFilterAndOrderByDateTime(string type, string filterOperator, string sqlValue, string gqlValue, string queryOperator)
         {
             // In MySQL, the DATETIME data type supports a range from '1000-01-01 00:00:00.0000000' to '9999-12-31 23:59:59.0000000'
-            if (DatabaseEngine is TestCategory.MYSQL && sqlValue is "\'9999-12-31 23:59:59.9999999\'")
+            if (DatabaseEngine is TestCategory.MYSQL && sqlValue is "'9999-12-31 23:59:59.9999999'")
             {
-                sqlValue = "\'9999-12-31 23:59:59.0000000\'";
+                sqlValue = "'9999-12-31 23:59:59.0000000'";
```
seantleonard, line 250: What change allowed you to removed the `\` escape? was it just unnecessary to begin with?

seantleonard:
> So an offset of local time (+5:30 in IST 10:23->15:53) is added when a value is supplied with 'Z' notation.

Doesn't Z zulu notation imply UTC? So the code which creates a new DateTime object must be passing in a param that says to store the date as local time, which is why you're seeing the conversion. maybe this is hitting ResolverMiddleware.cs:
`LocalTimeType => LocalTimePattern.ExtendedIso.Parse(leafJson).Value` which is hotchocolates usage of nodatime.

I'm just a little confused trying to piece together the context throughout this PR's change history.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -173,40 +186,68 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
+        [DataRow(DATETIME_TYPE, "eq", "'1753-01-01'", "\"1753-01-01\"", "=",
+            DisplayName = "datetime type filter and orderby test with eq operator and specific value '1753-01-01'.")]
+        [DataRow(DATETIME_TYPE, "gt", "'1999-01-08 10:23:00'", "\"1999-01-08 10:23:00\"", " > ",
+            DisplayName = "datetime type filter and orderby test with gt operator and specific value '1999-01-08 10:23:00'.")]
+        [DataRow(DATETIME_TYPE, "gte", "'1999-01-08 10:23:00'", "\"1999-01-08 10:23:00\"", " >= ",
+            DisplayName = "datetime type filter and orderby test with gte operator and specific value '1999-01-08 10:23:00'.")]
+        [DataRow(DATETIME_TYPE, "lt", "'9999-12-31 23:59:59'", "\"9999-12-31 23:59:59\"", " < ",
+            DisplayName = "datetime type filter and orderby test with lt operator and specific value '9999-12-31 23:59:59'.")]
+        [DataRow(DATETIME_TYPE, "lte", "'9999-12-31 23:59:59'", "\"9999-12-31 23:59:59\"", " <= ",
+            DisplayName = "datetime type filter and orderby test with lte operator and specific value '9999-12-31 23:59:59'.")]
```
seantleonard, line 206: are Z notation tests needed? lots of discussion of that in PR chat history.

seantleonard (changes requested):
some questions and suggestions

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
ayush3797: Yeah, was unnecessary.

src/Core/Services/TypeHelper.cs
```diff
@@ -111,13 +111,13 @@ public static class TypeHelper
             [SqlDbType.VarChar] = typeof(string)
         };
 
-        private static Dictionary<string, DbType> _timeSqlDbTypeToDbType = new()
+        public static Dictionary<SqlDbType, DbType> _sqlDbTypeToDbType = new()
```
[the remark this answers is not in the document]
ayush3797: Yeah, made it private thanks.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1319,18 +1331,11 @@ internal string GetTablePrefix(string databaseName, string schemaName)
                     }
 
                     columnDefinition.DbType = TypeHelper.GetDbTypeFromSystemType(columnDefinition.SystemType);
+                    DbType? dbTypeForMsSqlDateTimeType = SqlDbTypeToDbType((string)columnInfo["DATA_TYPE"]);
 
-                    if (GetDatabaseType() is DatabaseType.MSSQL)
+                    if (dbTypeForMsSqlDateTimeType is not null)
```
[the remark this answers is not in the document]
ayush3797: Added a check for MsSql. For other 2 db types, if the method is called, we will throw an unimplemented exception.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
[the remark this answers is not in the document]
ayush3797: > because SqlDbType is specific to mssql.

Why does PgSql, MySql not qualify for `sql`?

src/Core/Services/TypeHelper.cs
```diff
@@ -241,22 +241,16 @@ public static Type GetSystemTypeFromSqlDbType(string sqlDbTypeName)
...
             {
-                return value;
+                return dbType;
             }
 
-            throw new DataApiBuilderException(
-                message: $"Unexpected datetime data type: {sqlDbTypeName} encountered.",
-                statusCode: HttpStatusCode.ServiceUnavailable,
-                subStatusCode: DataApiBuilderException.SubStatusCodes.UnexpectedError);
+            return null;
```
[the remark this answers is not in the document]
ayush3797: This null should not be hit ideally now. But if it does, it means there are more mssql datatypes  which map to DateTime/DateTimeOffset system types other than date,datetime,datetime2,smalldatetime,datetimeoffset. I have covered all the datatypes as of now, but if this code block is hit, probably MsSql started supported a new datatype which we don't currently have a mapping for. Returning null instead of exception prevents DAB from handling that unexpected case.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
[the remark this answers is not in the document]
ayush3797: We already have tests validating insertions for datetimeoffset, datetime types now.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1319,18 +1331,11 @@ internal string GetTablePrefix(string databaseName, string schemaName)
                     }
 
                     columnDefinition.DbType = TypeHelper.GetDbTypeFromSystemType(columnDefinition.SystemType);
+                    DbType? dbTypeForMsSqlDateTimeType = SqlDbTypeToDbType((string)columnInfo["DATA_TYPE"]);
```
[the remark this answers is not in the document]
ayush3797: 1. yes
2. Don't need to unnecessarily add mappings for every data type. Almost all of the mssql data types have direct mappings for DbType except datetime,datetimeoffset ones. These mappings also hold true for other db types. For mssql datetime/datetimeoffset ones, will rename methods,variable names to be more indicative.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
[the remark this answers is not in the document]
seantleonard: `SqlDbType` refers to SQL Server specific type: https://learn.microsoft.com/dotnet/api/system.data.sqldbtype?view=net-6.0

>Specifies SQL Server-specific data type of a field, property, for use in a [SqlParameter](https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlparameter?view=net-6.0).

I don't want to use naming that conflicts with pre-existing naming scheme for enums provided by dotnet/ relevant ADO.NET Data Provider. 

PgSql/MySql will not use `SqlDbType`. For example, PG (Npgsql) uses the following table for data type mappings which include usage of the enum `NpgsqlDbType ` -> `DbType` https://www.npgsql.org/doc/types/basic.html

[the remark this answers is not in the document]
ayush3797: I see. Makes sense!

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
ayush3797: Yep, makes sense to have them. The relevant code changes in this PR modifies the parsing logic for insertions though and not filters/order bys. Added tests for insertions thus.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -342,15 +342,18 @@ public async Task TestTimeTypePrecisionCheck(string gqlValue, int count)
...
         [DataRow(SMALLDATETIME_TYPE, "\"1900-01-01\"")]
         [DataRow(SMALLDATETIME_TYPE, "\"2079-06-06\"")]
         [DataRow(DATETIME2_TYPE, "\"0001-01-01 00:00:00.0000000\"")]
         [DataRow(DATETIME2_TYPE, "\"9999-12-31 23:59:59.9999999\"")]
+        [DataRow(DATETIME2_TYPE, "\"9999-12-31 23:59:59.999Z\"")]
         [DataRow(DATE_TYPE, "\"0001-01-01\"")]
         [DataRow(DATE_TYPE, "\"9999-12-31\"")]
         [DataRow(DATETIMEOFFSET_TYPE, "\"0001-01-01 00:00:00.0000000\"")]
         [DataRow(DATETIMEOFFSET_TYPE, "\"9999-12-31 23:59:59.9999999\"")]
+        [DataRow(DATETIMEOFFSET_TYPE, "\"9999-12-31 23:59:59.9999999Z\"")]
```
seantleonard, line 356: I see in the chat history that `9999-12-31T23:59:59.9999999Z` with `T` in the date is something users can provide. Is that a format that needs to be tested here?

seantleonard (approved):
last question. Thanks for making all the updates you have made to address feedback!

src/Core/Services/TypeHelper.cs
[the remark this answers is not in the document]
Aniruddh25: By returning null now, aren't we silently hiding the case? Would it be better to throw an exception either here or in the caller to identify whenever in future a new date time type is added?

Aniruddh25 (approved):
Looks good to merge after resolving the renames and other nits, thank you for chasing this down!

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -173,40 +186,68 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
+        [DataRow(DATETIME_TYPE, "gte", "'1999-01-08 10:23:00'", "\"1999-01-08 10:23:00\"", " >= ",
+            DisplayName = "datetime type filter and orderby test with gte operator and specific value '1999-01-08 10:23:00'.")]
+        [DataRow(DATETIME_TYPE, "lt", "'9999-12-31 23:59:59'", "\"9999-12-31 23:59:59\"", " < ",
+            DisplayName = "datetime type filter and orderby test with lt operator and specific value '9999-12-31 23:59:59'.")]
+        [DataRow(DATETIME_TYPE, "lte", "'9999-12-31 23:59:59'", "\"9999-12-31 23:59:59\"", " <= ",
+            DisplayName = "datetime type filter and orderby test with lte operator and specific value '9999-12-31 23:59:59'.")]
+        [DataRow(DATETIME_TYPE, "neq", "'1999-01-08 10:23:00'", "\"1999-01-08 10:23:00\"", "!=",
+            DisplayName = "datetime type filter and orderby test with neq operator and specific value '1999-01-08 10:23:00'.")]
+        [DataRow(DATETIME_TYPE, "eq", "'1999-01-08 10:23:00'", "\"1999-01-08 10:23:00\"", "=",
+            DisplayName = "datetime type filter and orderby test with eq operator and specific value '1999-01-08 10:23:00'.")]
```
Aniruddh25, line 210: Are the `datetime`/`datetime2`/`smalldatetime`/`date` types in SQL still mapped to `DateTimeOffset` type of HotChocolate in the GraphQL schema?

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
```diff
@@ -20,7 +20,10 @@ public static class SupportedTypes
         public const string DECIMAL_TYPE = "Decimal";
         public const string STRING_TYPE = "String";
         public const string BOOLEAN_TYPE = "Boolean";
+        public const string DATE_TYPE = "Date";
```
Aniruddh25, line 23: The comment of this function says: "The type names mentioned here are HotChocolate Scalar built in types." I dont think DateTime2 is a supported type in HotChocolate, may need to modify the comment to reflect what this class means..

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -173,40 +186,68 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
...
-        [DataRow(DATETIME_TYPE, "eq", "\'1999-01-08 10:23:54.9999999-14:00\'", "\"1999-01-08 10:23:54.9999999-14:00\"", "=")]
-        [DataRow(DATETIMEOFFSET_TYPE, "neq", "\'1999-01-08 10:23:54.9999999-14:00\'", "\"1999-01-08 10:23:54.9999999-14:00\"", "!=")]
-        [DataRow(DATETIMEOFFSET_TYPE, "eq", "\'1999-01-08 10:23:54.9999999-14:00\'", "\"1999-01-08 10:23:54.9999999-14:00\"", "=")]
-        [DataRow(DATETIME_TYPE, "gt", "\'1999-01-08 10:22:00\'", "\"1999-01-08 10:22:00\"", " > ")]
-        [DataRow(DATETIME_TYPE, "gte", "\'1999-01-08 10:23:54\'", "\"1999-01-08 10:23:54\"", " >= ")]
-        [DataRow(DATETIME_TYPE, "lt", "\'2079-06-06\'", "\"2079-06-06\"", " < ")]
-        [DataRow(DATETIME_TYPE, "lte", "\'2079-06-06\'", "\"2079-06-06\"", " <= ")]
-        [DataRow(DATETIME_TYPE, "neq", "\'1999-01-08 10:23:54\'", "\"1999-01-08 10:23:54\"", "!=")]
-        [DataRow(DATETIME_TYPE, "eq", "\'1999-01-08 10:23:54\'", "\"1999-01-08 10:23:54\"", "=")]
+        [DataRow(DATETIME_TYPE, "eq", "'1999-01-08'", "\"1999-01-08\"", " = ",
```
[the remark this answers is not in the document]
ayush3797: The tests are not deleted. Just added display names for the tests.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
ayush3797: Yes

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -79,5 +79,21 @@ public override async Task PopulateTriggerMetadataForTable(string entityName, st
                 }
             }
         }
+
+        /// <inheritdoc/>
+        public override DbType? ResolveDbType(string sqlDbTypeName)
```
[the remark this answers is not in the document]
ayush3797: I think having ResolveDbType in SqlMetaDataProvider will allow us to have overridden implementations for getting the `DbType` from the pgsql/mysql/(probably dwsql?) data type if there is a need in future. This will help us avoid having to write specific hardcoded handlers in the TypeHelper class for each of the database flavors.

src/Core/Services/TypeHelper.cs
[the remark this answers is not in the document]
ayush3797: Yeah, I think that makes sense.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
ayush3797: Makes sense. Can add tests to be sure.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -325,15 +325,32 @@ private void LogPrimaryKeys()
...
-                    new()
+                string sqlType = (string)row["DATA_TYPE"];
+                Type systemType = SqlToCLRType(sqlType);
+                ParameterDefinition paramDefinition = new()
+                {
+                    SystemType = systemType,
+                    DbType = TypeHelper.GetDbTypeFromSystemType(systemType)
+                };
+
+                if (GetDatabaseType() is DatabaseType.MSSQL &&
```
Aniruddh25, line 336: This is a base class implementation of `FillSchemaForStoredProcedureAsync`. If there is certain portion of the code that really is only meant for a particular type of db, I think this function should get a overridden definition `FillSchemaForStoredProcedureAsync`. 
The parts of code that are common for all dbs can still live in the base class - so the overridden function calls the parts that are common and does additional stuff where needed. Thats why we have the derived classes.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1315,6 +1339,24 @@ internal string GetTablePrefix(string databaseName, string schemaName)
                     {
                         columnDefinition.DefaultValue = columnInfo["COLUMN_DEFAULT"];
                     }
+
+                    columnDefinition.DbType = TypeHelper.GetDbTypeFromSystemType(columnDefinition.SystemType);
+                    if (GetDatabaseType() is DatabaseType.MSSQL &&
```
Aniruddh25, line 1344: Same comment - why do specific things for MsSql in a base class? Can we have an overridden derived class implementation?

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -79,5 +79,24 @@ public override async Task PopulateTriggerMetadataForTable(string entityName, st
...
+                // Hence we cannot directly determine the DbType from the system type.
+                // However, to make sure that the database correctly interprets these datatypes, it is necessary to correctly
+                // populate the DbTypes.
+                return TypeHelper.TryGetDbTypeFromSqlDbDateTimeType(sqlDbType, out dbType);
+            }
+
+            // This code should never be hit because every sqlDbTypeName must have a corresponding sqlDbType.
+            // However, when a new data type is introduced in MsSql which maps to .NET type of DateTime, this code block
+            // will be hit. Returning false instead of throwing an exception in that case prevents the engine from crashing.
+            dbType = 0;
```
Aniruddh25, line 98: What does dbType 0 correspond to?

Aniruddh25 (dismissed):
Fix base/derived class implementation

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
```diff
@@ -26,4 +26,15 @@ public static class SupportedTypes
         public const string LOCALTIME_TYPE = "LocalTime";
         public const string TIME_TYPE = "Time";
     }
+
+    /// <summary>
+    /// Class representing the sql data types supported by DAB which do not have a
+    /// corresponding identical representation in HotChocolate scalar built in types.
+    /// </summary>
+    public static class SupportedSqlTypes
```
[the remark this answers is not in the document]
seantleonard: and an updated comment.

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
```diff
@@ -8,7 +8,7 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder.GraphQLTypes
     /// The type names mentioned here are Hotchocolate scalar built in types.
     /// The corresponding SQL type name may be different for e.g. UUID maps to Guid as the SQL type.
     /// </summary>
-    public static class SupportedTypes
+    public static class SupportedHotChocolateTypes
```
seantleonard, line 11: Updated filename to reflect this name would be beneficial for discoverability. With that being said, do you think it would be worth splitting these two classes into separate files?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -337,7 +338,15 @@ private void LogPrimaryKeys()
...
                     // Thus to determine the actual dbtype, we use the underlying MsSql type instead of the .NET type.
-                    paramDefinition.DbType = ResolveDbType((string)row["DATA_TYPE"]);
+                    DbType dbType;
+                    if (TryResolveDbType(sqlType, out dbType))
+                    {
+                        paramDefinition.DbType = dbType;
+                    }
+                    else
+                    {
+                        _logger.LogWarning("Could not determine DbType for SqlDb type of {sqlType}", sqlType);
```
seantleonard, line 348: What are the implications of this warning from the user's perspective:
- does the user need to act?
- will the user expect to see other errors because of this?
- will the user rarely, if ever, see this warning printed?

seantleonard (commented):
needs Ani's comments addressed

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -325,15 +325,32 @@ private void LogPrimaryKeys()
...
                     {
-                        SystemType = systemType,
-                        DbType = TypeHelper.GetDbTypeFromSystemType(systemType)
+                        _logger.LogWarning("Could not determine DbType for SqlDb type of {sqlType}", sqlType);
                     }
-                );
+                }
+
+                // Add to parameters dictionary without the leading @ sign
+                storedProcedureDefinition.Parameters.TryAdd(((string)row["PARAMETER_NAME"])[1..], paramDefinition);
```
aaronburtle, line 353: There should not be any duplicates here, so do we need to use `tryAdd`? If there is the possibility of duplicates and you want to avoid an exception here, right now this will ignore any attempts to add duplicates to the Stored Producedure Parameters, is that the intended behavior, that first should take precedent?

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
aaronburtle, line 11: I am confused on what this class represents. Are these all of the support hot chocolate types and also only the types that are exclusive to SQL? But then in addition to that, those types listed as GraphQL Types, while supported in SQL are potentially mapped to a different SQL type than is the type name we see in Hot Chocolate?

I find this layout a bit confusing and not sure how easy it will be to extend and re-use given that confusion.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: I think, the use of TryAdd() is something that Ani has been suggesting to avoid any unforeseen exceptions. Duplicate stored procedure params, I don't think is something possible though but refraining to change this at this moment (Since this is not a change from this PR)

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
ayush3797: Renamed the class/added comment.

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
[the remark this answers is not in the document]
ayush3797: Renamed the class/added comment.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -381,6 +398,14 @@ public virtual Task PopulateTriggerMetadataForTable(string entityName, string sc
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Takes a string version of a sql date/time type and returns its corresponding DbType.
+        /// </summary>
+        /// <param name="sqlDbTypeName">Name of the sqlDbType.<</param>
+        /// <param name="dbType">DbType of the parameter corresponding to its sqlDbTypeName.</param>
+        /// <returns>Returns true when the given sqlDbTypeName datetime type is supported by DAB and resolve it to its corresponding DbType, else false.</returns>
+        public virtual bool TryResolveDbType(string sqlDbTypeName, [NotNullWhen(true)] out DbType dbType) => throw new NotImplementedException();
```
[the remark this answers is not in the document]
ayush3797: done

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Sure, we can have that. But then we cannot throw exception in the default implementation of the method and would have to instead return null.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
ayush3797: image

We don't care here though since we are returning false in which case we don't use the `out` value

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Again, this would only be hit when a new data type is added in MsSql which maps to .NET type of `DateTime`. The user cannot do anything regarding this. This would involve a code change to add a mapping for the new data type.

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
ayush3797: I think `SupportedTypes` make sense at the moment. If the number of classes get too big, we can seperate them into different files.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
aaronburtle: > Sure, we can have that. But then we cannot throw exception in the default implementation of the method and would have to instead return null. Since currently we are throwing an exception in the default implementation of the method, we need to ensure that we only call the method for MsSql and hence this check at L1344.

Can you explain more about why this can not remain void if we have the different behavior for MsSql split into an overriding function in the MsSql class? I am not following why we cant just have the MsSql specific code attached to a function in MsSql that overrides this and calls a shared helper that executes the code shared between them.

ayush3797: Keeping the function void would require to write two different functions for ParameterDefinition and ColumnDefinition to populate the DbType. What I am doing now is that the function will remain of return type boolean but I have moved to mssql specific logic to the mssql class.

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
```diff
@@ -20,10 +20,21 @@ public static class SupportedTypes
...
         public const string TIME_TYPE = "Time";
     }
+
+    /// <summary>
+    /// Class representing the sql datetime types supported by DAB which all map to the same .NET type of DateTime
+    /// and Hotchocolate scalar type of DateTime.
+    /// </summary>
+    public static class SupportedDateTimeTypes
+    {
+        public const string DATETIME_TYPE = "DateTime";
```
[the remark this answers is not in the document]
ayush3797: Moved this back to the SupportedHotChocolateTypes to be more clear. The casing is important for HotChocolate.

src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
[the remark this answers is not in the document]
ayush3797: Yes, the build is failing on my local.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Done.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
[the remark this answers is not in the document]
ayush3797: For now, moved the method to MsSqlMetadataProvider.

src/Core/Services/TypeHelper.cs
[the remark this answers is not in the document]
ayush3797: We will log a warning to just know that an additional datatype was added to MsSql which maps to .NET type of datetime.

seantleonard (approved):

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -80,5 +85,156 @@ public override async Task PopulateTriggerMetadataForTable(string entityName, st
...
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
+            }
+
+            // Each row in the procedureParams DataTable corresponds to a single parameter
+            DataTable parameterMetadata = await conn.GetSchemaAsync(collectionName: "ProcedureParameters", restrictionValues: procedureRestrictions);
+
+            // For each row/parameter, add an entry to StoredProcedureDefinition.Parameters dictionary
+            foreach (DataRow row in parameterMetadata.Rows)
+            {
+                // row["DATA_TYPE"] has value type string so a direct cast to System.Type is not supported.
```
aaronburtle, line 161: nit: code duplication with the base class function

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -80,5 +85,156 @@ public override async Task PopulateTriggerMetadataForTable(string entityName, st
...
+                }
+
+                // Add to parameters dictionary without the leading @ sign
+                storedProcedureDefinition.Parameters.TryAdd(((string)row["PARAMETER_NAME"])[1..], paramDefinition);
+            }
+
+            // Loop through parameters specified in config, throw error if not found in schema
+            // else set runtime config defined default values.
+            // Note: we defer type checking of parameters specified in config until request time
+            Dictionary<string, object>? configParameters = procedureEntity.Source.Parameters;
```
aaronburtle, line 189: nit: code duplication with the base class

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -80,5 +85,156 @@ public override async Task PopulateTriggerMetadataForTable(string entityName, st
...
+            DataTable allColumnsInTable)
+        {
+            foreach (DataRow columnInfo in allColumnsInTable.Rows)
+            {
+                string columnName = (string)columnInfo["COLUMN_NAME"];
+                bool hasDefault =
+                    Type.GetTypeCode(columnInfo["COLUMN_DEFAULT"].GetType()) != TypeCode.DBNull;
+                if (sourceDefinition.Columns.TryGetValue(columnName, out ColumnDefinition? columnDefinition))
+                {
+                    columnDefinition.HasDefault = hasDefault;
```
aaronburtle, line 101: nit: code duplication with base class

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -80,5 +85,156 @@ public override async Task PopulateTriggerMetadataForTable(string entityName, st
...
+
+        /// <inheritdoc/>
+        protected override async Task FillSchemaForStoredProcedureAsync(
+            Entity procedureEntity,
+            string entityName,
+            string schemaName,
+            string storedProcedureSourceName,
+            StoredProcedureDefinition storedProcedureDefinition)
+        {
+            using DbConnection conn = new SqlConnection();
```
aaronburtle, line 132: this code is quite similar to the code in the base class for this same function, is there anyway we can have this in a form factor so that the similar code can be shared?

aaronburtle (approved):
Looks good just a few nits before merging if possible.

abhishekkumams (approved):

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
ayush3797: Even though the code is similar, we want to not have mssql specific logic in the base class code. Also, if I move back the code to the base class, we would need to have conditional for datatime types and then call the overridden method for populating DbType. For database types other than MsSql, calling the overridden method again and again won't make sense.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
ayush3797: https://github.com/Azure/data-api-builder/pull/1636#discussion_r1416766521

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
ayush3797: https://github.com/Azure/data-api-builder/pull/1636#discussion_r1416766521

seantleonard:
@microsoft-github-policy-service rerun

ayush3797:
Ignoring the failing test for DwSql **FindByDateTimePKTest()**. @rohkhann will take a look. Not a blocker.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
aaronburtle: I dont think this refactor needs to combine MsSql specific code, will follow up in the issue created.

Issue #1638, opened 2023-08-16
RFC: Multiple data sources


JerryNixon:
# Summary
Allow a single DAB instance to interact with multiple backend data sources. 

# Motivation
This design enables DAB to address more sophisticated customers.

# Old behavior
Currently, DAB supports only a single `data source` and a single `dab-config.json` file. 

## Barriers to multiple data sources: 
1. The `data-source` property in the config is of type `object`, not `array`. 
2. The `data-source` property in the config applies to every `entity`.  

# New behavior
Support the use of multiple `dab-config.json` files, each corresponding to a different `data-source`, to merge at runtime. 

⭐ When DAB starts, the GLOBAL and CHILD files will be merged. 

## Design goals:
1. A single `dab-config.json` file must always be viable in DAB. 
1. We will not impose arbitrary limits on `data-source` or `entities` at this time. 
1. Developers must easily understand which file has an error. 
1. `entity` names across all files must be unique to support future features. 

## Sample global file
````json
{
  "$schema": "https://github.com/Azure/data-api-builder/releases/download/v0.7.6/dab.draft.schema.json",
  "data-source-files": [
     "dab-config.mssql-hr.json",
     "dab-config.cosmosdb-sales.json",
     "dab-config.mysql-inventory.json",
     "dab-config.postgres-shipping.json",
   ],
  ...
}
````

## Global file characteristics
1. The global file must have `runtime` properties.
1. The global file's `data-source` property is optional. 
1. The global file's `data-source-files` property is optional. 
1. Paths in `data-source-files` may include Unicode characters. 

## New `data-source-files` property in JSON schema.
````json
{
  "$schema": "https://json-schema.org/draft-07/schema",
  "data-source-files": { // new property
      "type": "array",
      "description": "Child file paths",
      "items": {
        "type": "string"
      },
      "default": []
  }
}
````

This new property, of type `string[]`, is optional. If the `data-source-files` property is `[]` or omitted, the `data-source` property is required. Otherwise, the `data-source` property is optional. 

## Path translation
When filled, the `data-source-files` property contains _paths_ to child files. These paths can include dots `.` and slashes `/` & `\`, indicating the location of the child file relative to the global config file. 

If the global config file is at `c:\web\api\dab-config.json`, then:

|Path in Config|File Location|
|-|-
|`dab-config.02.json`|`c:\web\api\dab-config.02.json`
|`./dab-config.03.json`|`c:\web\api\dab-config.03.json`
[...]

PR #1639, opened 2023-08-18
Runtime config changes for multiple database support

rohkhann:
## Why make this change? 

1. In this pr, RuntimeConfig object is extended to support multiple database scenario. Along with this all the class objects are respectively extended to be able to take multiple database input.
2. Consequently, there are downstream changes done in the different executors. Currently each executor will use the default datasource to run their queries. 
3. In subsequent pr's through Named dependency injection, the executors will be able to choose which db to act against.
4. Please see RFC: https://github.com/Azure/data-api-builder/issues/1638

## What is this change?

- This change is meant to be part of a series of changes that will enable multiple database support on data api builder. 
- This first change ensures compatibility of runtimeconfig with multiple database approach. 
- This change does not unlock any new scenario as of now. (there will be further pr's that will build on this change to fully support multiple database scenario). It just extends runtimeconfig while ensuring that the single datasource scenario works as expected.

## How was this tested?
- Existing integration tests cover all execution paths to test for backword compatibility. 
- Ran dab locally as well to test that it works.
- Existing unit tests cover all execution paths to test for backword compatibility.


src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +105,57 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            {
+                foreach (KeyValuePair<string, string> pair in datasourceNameToConnectionString)
+                {
+                    DataSource ds = config.DatasourceNameToDataSource[pair.Key];
+                    ds = ds with { ConnectionString = pair.Value };
+                    config.DatasourceNameToDataSource[pair.Key] = ds;
+                }
+            }
+        }
+        catch (JsonException ex)
```
ayush3797, line 137: This function replicates the functionality of the `TryParseConfig` function except that it missed the adding application name for telemetry for MsSQL. Was that intentional?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +105,57 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
         return true;
     }
 
+    /// <summary>
+    /// Parses a JSON string into a <c>RuntimeConfig</c> object for multiple database scenario.
+    /// </summary>
+    /// <param name="json">JSON that represents the config file.Json expected to contain Datasources dictionary.</param>
+    /// <param name="config">The parsed config, or null if it parsed unsuccessfully.</param>
+    /// <returns>True if the config was parsed, otherwise false.</returns>
+    public static bool TryParseConfigMultipleDatabase(string json, [NotNullWhen(true)] out RuntimeConfig? config, ILogger? logger = null, Dictionary<string, string>? datasourceNameToConnectionString = null)
```
ayush3797, line 114: This function replicates the functionality of the `TryParseConfig` function except that it missed the adding application name for telemetry for MsSQL. Was that intentional?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +105,57 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            config = JsonSerializer.Deserialize<RuntimeConfig>(json, options);
+
+            if (config is null)
+            {
+                return false;
+            }
+
+            if (datasourceNameToConnectionString != null)
+            {
+                foreach (KeyValuePair<string, string> pair in datasourceNameToConnectionString)
```
ayush3797, line 129: To enhance readability:
```suggestion
                foreach ((string dataSourceName, string connString) in datasourceNameToConnectionString)
```

src/Config/RuntimeConfigLoader.cs
```diff
@@ -107,7 +167,8 @@ public static JsonSerializerOptions GetSerializationOptions()
             PropertyNamingPolicy = new HyphenatedNamingPolicy(),
             ReadCommentHandling = JsonCommentHandling.Skip,
             WriteIndented = true,
-            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+            IncludeFields = true,
```
ayush3797, line 171: What does this do?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -114,6 +114,7 @@ public bool TryGetLoadedConfig([NotNullWhen(true)] out RuntimeConfig? runtimeCon
     /// This initialization method is used when the configuration is sent to the ConfigurationController
     /// in the form of a string instead of reading the configuration from a configuration file.
     /// This method assumes the connection string is provided as part of the configuration.
+    /// Initialzed the first database within the datasource list.
```
ayush3797, line 117:
```suggestion
    /// Initialize the first database within the datasource list.
```

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -96,13 +96,16 @@ public void ValidateConfig()
...
-            // Connection string can't be null or empty
-            if (string.IsNullOrWhiteSpace(runtimeConfig.DataSource.ConnectionString))
+            foreach (DataSource dataSource in runtimeConfig.DatasourceNameToDataSource.Values)
             {
-                throw new DataApiBuilderException(
-                    message: DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE,
-                    statusCode: HttpStatusCode.ServiceUnavailable,
-                    subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
+                // Connection string can't be null or empty
+                if (string.IsNullOrEmpty(dataSource.ConnectionString))
```
ayush3797, line 102: Will the data source names be uniquely generated by us in a multi-db scenario? If so, I assume they are for internal use to DAB only and should not be included in the exception message. I am thinking about the scenario where we have a validation error in one of the config files, and just by looking at the exception, we want to know which config file has the error.

src/Config/RuntimeConfigLoader.cs
rohkhann: This allows fields to be serialized. We converted runtime config to have explicit fields

src/Core/Configurations/RuntimeConfigValidator.cs
rohkhann: They will be generated by us internally so the customer will not be aware of them and ideally we will tie this to the filename when we do the merge. However, its a little tricky - We will have to validate this in the cli when we do the merge if that makes sense. To the core system there is just a single runtimeconfig file not multiple. Hence will do this validation, when I do the cli changes.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -71,7 +71,16 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
                 updatedConnectionString = GetConnectionStringWithApplicationName(updatedConnectionString);
             }
 
-            config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+            if (datasourceName == null || datasourceName == config.DefaultDBName)
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Config/RuntimeConfigLoader.cs
rohkhann: good catch. fixed.

src/Config/RuntimeConfigLoader.cs
rohkhann: fixed.

src/Core/Configurations/RuntimeConfigValidator.cs
ayush3797: I see. Thanks for the clarification.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -34,13 +34,13 @@ public class MsSqlQueryExecutor : QueryExecutor<SqlConnection>
         /// The managed identity Access Token string obtained
         /// from the configuration controller.
         /// </summary>
-        private readonly string? _accessTokenFromController;
+        private readonly Dictionary<string, string?> _accessTokenFromController;
 
         /// <summary>
-        /// The MsSql specific connection string builder.
+        /// The MySql specific connection string builder.
```
ayush3797, line 40: MySql?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -86,17 +101,22 @@ public override SqlConnectionStringBuilder ConnectionStringBuilder
...
-        public override async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection conn)
+        public override async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection conn, string? datasourceName = null)
         {
+            if (string.IsNullOrEmpty(datasourceName))
+            {
+                datasourceName = ConfigProvider.GetConfig().DefaultDBName;
+            }
             // Only attempt to get the access token if the connection string is in the appropriate format
-            if (_attemptToSetAccessToken)
+            // using default for first db - maintaining backward compatibility for single db scenario.
```
ayush3797, line 111: The comment makes more sense before line 108?

src/Core/Resolvers/MySqlQueryExecutor.cs
```diff
@@ -52,20 +53,36 @@ public override MySqlConnectionStringBuilder ConnectionStringBuilder
...
         {
+            _attemptToSetAccessToken = new Dictionary<string, bool>();
             _accessTokenFromController = runtimeConfigProvider.ManagedIdentityAccessToken;
-            _attemptToSetAccessToken =
-                ShouldManagedIdentityAccessBeAttempted();
+            IEnumerable<KeyValuePair<string, DataSource>> mysqldbs = runtimeConfigProvider.GetConfig().DatasourceNameToDataSource.Where(x => x.Value.DatabaseType == DatabaseType.MySQL);
 
-            // Force always allow user variables
-            ConnectionStringBuilder.AllowUserVariables = true;
+            foreach (KeyValuePair<string, DataSource> dataSourcePair in mysqldbs)
```
ayush3797, line 63: Similar suggestion: https://github.com/Azure/data-api-builder/pull/1639/files#r1302063470

src/Core/Resolvers/PostgreSqlExecutor.cs
```diff
@@ -53,17 +54,32 @@ public override NpgsqlConnectionStringBuilder ConnectionStringBuilder
...
                   httpContextAccessor)
         {
+            IEnumerable<KeyValuePair<string, DataSource>> mysqldbs = runtimeConfigProvider.GetConfig().DatasourceNameToDataSource.Where(x => x.Value.DatabaseType == DatabaseType.PostgreSQL);
+            _attemptToSetAccessToken = new Dictionary<string, bool>();
             _accessTokenFromController = runtimeConfigProvider.ManagedIdentityAccessToken;
-            _attemptToSetAccessToken =
-                ShouldManagedIdentityAccessBeAttempted();
 
-            if (runtimeConfigProvider.IsLateConfigured)
+            foreach (KeyValuePair<string, DataSource> dataSourcePair in mysqldbs)
```
ayush3797, line 64: https://github.com/Azure/data-api-builder/pull/1639/files#r1302063470

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -61,22 +61,37 @@ public override SqlConnectionStringBuilder ConnectionStringBuilder
...
                   httpContextAccessor)
         {
             RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
+            IEnumerable<KeyValuePair<string, DataSource>> mssqldbs = runtimeConfig.DatasourceNameToDataSource.Where(x => x.Value.DatabaseType == DatabaseType.MSSQL);
+            _attemptToSetAccessToken = new Dictionary<string, bool>();
+            _isSessionContextEnabled = new Dictionary<string, bool>();
+            _accessTokenFromController = runtimeConfigProvider.ManagedIdentityAccessToken;
 
-            if (runtimeConfigProvider.IsLateConfigured)
+            foreach (KeyValuePair<string, DataSource> dataSourcePair in mssqldbs)
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/PostgreSqlExecutor.cs
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -200,7 +210,7 @@ public virtual void PopulateDbTypeForParameter(KeyValuePair<string, DbConnection
         }
 
         /// <inheritdoc />
-        public virtual async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection conn)
+        public virtual async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection conn, string? datasourceName = null)
```
ayush3797, line 213: Can we use one of `dbName`/`dataSourceName` everywhere? Prompts me to think if they mean different things since we are in initial stages of this change.

src/Core/Resolvers/PostgreSqlExecutor.cs
```diff
@@ -28,23 +29,23 @@ public class PostgreSqlQueryExecutor : QueryExecutor<NpgsqlConnection>
...
         /// from the configuration controller.
         /// </summary>
-        private readonly string? _accessTokenFromController;
+        private readonly Dictionary<string, string?> _accessTokenFromController;
 
         public DefaultAzureCredential AzureCredential { get; set; } = new();
 
         /// <summary>
-        /// The PostgreSql specific connection string builder.
+        /// The MySql specific connection string builder.
```
ayush3797, line 37: PostgreSql?

ayush3797:
Looks good. Still left with some files to review. Left some nits.

src/Core/Resolvers/PostgreSqlExecutor.cs
rohkhann: fixed.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -221,8 +226,12 @@ public async Task TestRetryPolicySuccessfullyExecutingQueryAfterNAttempts()
             Mock<MsSqlQueryExecutor> queryExecutor
                 = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object);
 
-            queryExecutor.Setup(x => x.ConnectionStringBuilder).Returns(
-                new SqlConnectionStringBuilder(provider.GetConfig().DataSource.ConnectionString));
+            Dictionary<string, DbConnectionStringBuilder> dictionary = new()
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/MsSqlQueryExecutor.cs
rohkhann: fixed.

src/Cli.Tests/ModuleInitializer.cs
```diff
@@ -21,6 +21,12 @@ public static void Init()
         VerifierSettings.IgnoreMember<DataSource>(dataSource => dataSource.ConnectionString);
         // Ignore the JSON schema path as that's unimportant from a test standpoint.
         VerifierSettings.IgnoreMember<RuntimeConfig>(config => config.Schema);
+        // Ignore the databaseNameToDataSource array as its not exposed to customer.
+        VerifierSettings.IgnoreMember<RuntimeConfig>(config => config.DatasourceNameToDataSource);
```
Aniruddh25, line 25: nit: rename to `DataSourceNameToDataSource`

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +105,65 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            if (config is null)
+            {
+                return false;
+            }
+
+            if (datasourceNameToConnectionString != null)
+            {
+                foreach ((string dataSourceName, string connectionString) in datasourceNameToConnectionString)
+                {
+                    DataSource ds = config.DatasourceNameToDataSource[dataSourceName];
```
Aniruddh25, line 131: Are we certain that the keys from the input dictionary will be the same in the DatasourceNameToDataSource member?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +105,65 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            config = JsonSerializer.Deserialize<RuntimeConfig>(json, options);
+
+            if (config is null)
+            {
+                return false;
+            }
+
+            if (datasourceNameToConnectionString != null)
+            {
+                foreach ((string dataSourceName, string connectionString) in datasourceNameToConnectionString)
```
Aniruddh25, line 129: When its a single database scenario, why cant the same function still be useful? This loop just run once, right? Couldn't you pass in a dictionary of the single datasource name to connection string as an argument?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -36,9 +36,9 @@ public class RuntimeConfigProvider
     public bool IsLateConfigured { get; set; }
 
     /// <summary>
-    /// The access token representing a Managed Identity to connect to the database.
+    /// The access tokens representing a Managed Identity to connect to the database.
```
Aniruddh25, line 39: What is the key?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -212,8 +213,13 @@ private async Task<bool> InvokeConfigLoadedHandlersAsync()
         return results.All(x => x);
     }
 
-    private static RuntimeConfig HandleCosmosNoSqlConfiguration(string? schema, RuntimeConfig runtimeConfig, string connectionString)
+    private static RuntimeConfig HandleCosmosNoSqlConfiguration(string? schema, RuntimeConfig runtimeConfig, string connectionString, string? databaseName = null)
     {
+        if (string.IsNullOrEmpty(databaseName))
```
Aniruddh25, line 218: Lets keep this argument name as dataSourceName since this is different than the databasename in the connection string

Aniruddh25 (changes requested):
Posting review comments so far..

src/Cli.Tests/ModuleInitializer.cs
rohkhann: fixed.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -71,7 +71,16 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
-            config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+            if (datasourceName == null || string.Equals(datasourceName, config.DefaultDataSourceName, StringComparison.OrdinalIgnoreCase))
+            {
+                // single database scenario - default db is set.
+                config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+                config.DatasourceNameToDataSource[config.DefaultDataSourceName] = config.DataSource;
+            }
+            else
+            {
+                config.DatasourceNameToDataSource[datasourceName] = config.DatasourceNameToDataSource[datasourceName] with { ConnectionString = updatedConnectionString };
```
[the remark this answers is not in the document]
rohkhann: We will not be able to set the value, as this is in the tryparse config case. But the tryget makes sense and we should throw an exception here if it doesnt exist. will add that.

src/Config/RuntimeConfigLoader.cs
rohkhann: You could, ive just created a separate method to maintain backward compatibility.

src/Config/RuntimeConfigLoader.cs
rohkhann: Yes they are supposed to be. Ive added a conditon to validate this. DataSourceNameToDatasource would be a created dictionary part of the config that is tightly couples to the entityNametodatasource dict. Hence whenever the connection strings are added they will have to be added to the correct datasources in the datasourcenametodatasource dict only.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: Updated comment.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: fixed.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -6,12 +6,60 @@
...
+    public string DefaultDataSourceName;
+
+    [JsonIgnore]
+    public Dictionary<string, DataSource> DatasourceNameToDataSource { get; set; }
+
+    [JsonIgnore]
+    public Dictionary<string, string> EntityNameToDataSourceName { get; set; }
+
+    [JsonConstructor]
+    public RuntimeConfig(string Schema, DataSource DataSource, RuntimeOptions Runtime, RuntimeEntities Entities)
```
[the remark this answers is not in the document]
rohkhann: this would be a breaking change for anyone using parameter names to reference this constructor: 
image. Hence want to avoid it for this one.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -71,7 +71,24 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
                 updatedConnectionString = GetConnectionStringWithApplicationName(updatedConnectionString);
             }
 
-            config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+            if (dataSourceName == null || string.Equals(dataSourceName, config.DefaultDataSourceName, StringComparison.OrdinalIgnoreCase))
```
aaronpowell, line 74:
```suggestion
            if (dataSourceName is null null || string.Equals(dataSourceName, config.DefaultDataSourceName, StringComparison.OrdinalIgnoreCase))
```

Keeping consistent with our other code.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -71,7 +71,24 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            {
+                // single database scenario - default db is set.
+                config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+                config.DataSourceNameToDataSource[config.DefaultDataSourceName] = config.DataSource;
+            }
+            else
+            {
+                config.DataSourceNameToDataSource.TryGetValue(config.DefaultDataSourceName, out DataSource? ds);
+                if (ds != null)
```
aaronpowell, line 83:
```suggestion
                if (ds is not null)
```

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +113,72 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+        try
+        {
+            config = JsonSerializer.Deserialize<RuntimeConfig>(json, options);
+
+            if (config is null)
+            {
+                return false;
+            }
+
+            if (datasourceNameToConnectionString != null)
```
aaronpowell, line 135:
```suggestion
            if (datasourceNameToConnectionString is not null)
```

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +113,72 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            {
+                foreach ((string dataSourceName, string connectionString) in datasourceNameToConnectionString)
+                {
+                    if (config.DataSourceNameToDataSource.ContainsKey(dataSourceName))
```
aaronpowell, line 139: You should use `TryGetValue` here as it saves having to access the indexer on line 141.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -96,13 +96,16 @@ public void ValidateConfig()
...
+            foreach (DataSource dataSource in runtimeConfig.DataSourceNameToDataSource.Values)
             {
-                throw new DataApiBuilderException(
-                    message: DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE,
-                    statusCode: HttpStatusCode.ServiceUnavailable,
-                    subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
+                // Connection string can't be null or empty
+                if (string.IsNullOrEmpty(dataSource.ConnectionString))
+                {
+                    throw new DataApiBuilderException(
```
seantleonard, line 104: Why change from `IsNullOrWhiteSpace()`? is there a valid scenario where connectionstring is all whitespace?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -390,19 +396,22 @@ public static void ValidateGlobalEndpointRouteConfig(RuntimeConfig runtimeConfig
         /// <param name="runtimeConfig"></param>
         public static void ValidateRestURI(RuntimeConfig runtimeConfig)
         {
-            // CosmosDB_NoSQL does not support rest. No need to do any validations.
-            if (runtimeConfig.DataSource.DatabaseType is DatabaseType.CosmosDB_NoSQL)
+            foreach (DataSource dataSource in runtimeConfig.DataSourceNameToDataSource.Values)
```
seantleonard, line 399: This foreach loop doesn't make sense here. For example, if your datasources collection has 5 entries, you'd be validating the same `restPath` 5 times. It looks like the only valid check to include in the foreach would be to skip REST validation if any of the datasource DBtypes is Cosmos. Even then, we'd probably want a separate validation to ensure all database sources share the same dbtype, wouldn't we?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +113,72 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            {
+                logger.LogError(ex, errorMessage);
+            }
+
+            config = null;
+            return false;
+        }
+
+        return true;
+    }
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -137,16 +139,16 @@ public class CosmosQueryEngine : IQueryEngine
...
         {
             // TODO: fixme we have multiple rounds of serialization/deserialization JsomDocument/JObject
             // TODO: add support for nesting
             // TODO: add support for join query against another container
             // TODO: add support for TOP and Order-by push-down
 
             CosmosQueryStructure structure = new(context, parameters, _metadataStoreProvider, _authorizationResolver, _gQLFilterParser);
-
-            Container container = _clientProvider.Client.GetDatabase(structure.Database).GetContainer(structure.Container);
+            CosmosClient client = string.IsNullOrEmpty(dataSourceName) ? _clientProvider.Clients[_clientProvider._defaultDataSourceName] : _clientProvider.Clients[dataSourceName];
```
seantleonard, line 150: I'm not too familiar with CosmosClient SDK, but making copies of Singleton CosmosClient objects seems like it could be problematic. At the end of this method, the CosmosClient object is disposed and I'm not sure if that would have negative consequences. (CosmosClient might not be disposed here, you should use the `using` keyword here regardless)

seantleonard (commented):
I still have to review tests.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -6,12 +6,60 @@
...
+    }
+
+    public RuntimeConfig(string Schema, Dictionary<string, DataSource> dataSourceDict, Dictionary<string, string> entityNameToDataSourceDict, RuntimeOptions Runtime, RuntimeEntities Entities)
+    {
+        this.Schema = Schema;
+        this.DatasourceNameToDataSource = dataSourceDict;
+        this.EntityNameToDataSourceName = entityNameToDataSourceDict;
+        this.Runtime = Runtime;
+        this.Entities = Entities;
+        KeyValuePair<string, DataSource> keyValuePair = dataSourceDict.First();
```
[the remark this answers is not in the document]
rohkhann: Im confused here on what to do. The default db name should not even be applicable in a multiple database scenario and hence it technically does not matter what you set it to. Maybe can take an input here from constructor and enforce that if the dictionary is not null then the default db name should also not be null?

src/Core/Configurations/RuntimeConfigProvider.cs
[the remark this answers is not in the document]
rohkhann: fixed in previous iteration.

src/Config/ObjectModel/RuntimeConfig.cs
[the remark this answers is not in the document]
rohkhann: Merged to one constructor. Added description above.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +105,65 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            {
+                foreach ((string dataSourceName, string connectionString) in datasourceNameToConnectionString)
+                {
+                    DataSource ds = config.DatasourceNameToDataSource[dataSourceName];
+                    string updatedConnectionString = connectionString;
+
+                    // Add Application Name for telemetry for MsSQL
+                    if (ds.DatabaseType is DatabaseType.MSSQL)
+                    {
+                       updatedConnectionString= GetConnectionStringWithApplicationName(connectionString);
```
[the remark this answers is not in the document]
rohkhann: fixed in previous iteration.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -96,6 +105,65 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+        try
+        {
+            config = JsonSerializer.Deserialize<RuntimeConfig>(json, options);
+
+            if (config is null)
+            {
+                return false;
+            }
+
+            if (datasourceNameToConnectionString != null)
```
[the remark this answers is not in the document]
rohkhann: fixed in previous iteration.

src/Core/Configurations/RuntimeConfigValidator.cs
rohkhann: Fixed. Was rewriting this code and typically always used isnullorempty. moved back to isnullorwhitespace.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -61,22 +61,35 @@ public override SqlConnectionStringBuilder ConnectionStringBuilder
...
             }
 
-            MsSqlOptions? msSqlOptions = runtimeConfig.DataSource.GetTypedOptions<MsSqlOptions>();
-            _isSessionContextEnabled = msSqlOptions is null ? false : msSqlOptions.SetSessionContext;
-            _accessTokenFromController = runtimeConfigProvider.ManagedIdentityAccessToken;
-            _attemptToSetAccessToken = ShouldManagedIdentityAccessBeAttempted();
+            if (!_accessTokenFromController.ContainsKey(runtimeConfig.DefaultDataSourceName))
+            {
+                _accessTokenFromController[runtimeConfig.DefaultDataSourceName] = null;
+            }
```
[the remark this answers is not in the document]
rohkhann: Based on the logic in line 118, If this value is null we get the default access token. However, i agree that this is not the best way to do it. Changed the logic using a tryGetValue method.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -50,9 +50,9 @@ public override SqlConnectionStringBuilder ConnectionStringBuilder
         /// </summary>
         private AccessToken? _defaultAccessToken;
 
-        private bool _attemptToSetAccessToken;
+        private Dictionary<string, bool> _attemptToSetAccessToken;
 
-        private bool _isSessionContextEnabled;
+        private Dictionary<string, bool> _isSessionContextEnabled;
```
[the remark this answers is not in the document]
rohkhann: Added comments

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -61,7 +61,7 @@ public class SqlQueryEngine : IQueryEngine
         /// </summary>
         /// <param name="context">HotChocolate Request Pipeline context containing request metadata</param>
         /// <param name="parameters">GraphQL Query Parameters from schema retrieved from ResolverMiddleware.GetParametersFromSchemaAndQueryFields()</param>
-        public async Task<Tuple<JsonDocument?, IMetadata?>> ExecuteAsync(IMiddlewareContext context, IDictionary<string, object?> parameters)
+        public async Task<Tuple<JsonDocument?, IMetadata?>> ExecuteAsync(IMiddlewareContext context, IDictionary<string, object?> parameters, string dataSourceName = "")
```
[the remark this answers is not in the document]
rohkhann: Yes. in following pr's the datasource name will actually be passed in enabling it to be used over the default db name.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -65,12 +64,23 @@ public class QueryExecutor<TConnection> : IQueryExecutor
...
-            List<string>? args = null)
+            List<string>? args = null,
+            string dataSourceName = "")
         {
+            if (string.IsNullOrEmpty(dataSourceName))
+            {
+                dataSourceName = ConfigProvider.GetConfig().DefaultDataSourceName;
+            }
+
+            if (!ConnectionStringBuilders.ContainsKey(dataSourceName))
```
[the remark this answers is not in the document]
rohkhann: Since this parameter is passed in through a function call it could definitely be invalid. In our logic, we would retrieve the entity name from the entityNametoDbName dict and then pass it in to this function. However, if there is some logic error and we pass in the wrong name (for example, we send in some other variable due to a code error instead of one retrieved from entityNametodbName), this should throw exception. Even test methods in future that may want to test against this method could always by mistake provide wrong input.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -65,12 +64,23 @@ public class QueryExecutor<TConnection> : IQueryExecutor
...
+            {
+                throw new Exception("Query execution failed. Could not find datasource to execute query against");
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -34,13 +34,13 @@ public class MsSqlQueryExecutor : QueryExecutor<SqlConnection>
         /// The managed identity Access Token string obtained
         /// from the configuration controller.
         /// </summary>
-        private readonly string? _accessTokenFromController;
+        private readonly Dictionary<string, string?> _accessTokenFromController;
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/IQueryExecutor.cs
```diff
@@ -22,6 +22,7 @@ public interface IQueryExecutor
         /// <param name="parameters">The parameters used to execute the SQL text.</param>
         /// <param name="dataReaderHandler">The function to invoke to handle the results
         /// in the DbDataReader obtained after executing the query.</param>
+        /// <param name="dataSourceName">dataSourceName against which to run query.</param>
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -6,12 +6,60 @@
...
+        this.EntityNameToDataSourceName = new Dictionary<string, string>();
+
+        this.DefaultDBName = Guid.NewGuid().ToString();
+        this.DatasourceNameToDataSource.Add(this.DefaultDBName, this.DataSource);
+
+        if (Entities != null)
+        {
+            foreach (KeyValuePair<string, Entity> entity in Entities)
+            {
+                EntityNameToDataSourceName.Add(entity.Key, DefaultDBName);
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Config/ObjectModel/RuntimeConfig.cs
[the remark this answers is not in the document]
rohkhann: discussed offline. resovling.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -65,12 +67,23 @@ public class QueryExecutor<TConnection> : IQueryExecutor
             IDictionary<string, DbConnectionParam> parameters,
             Func<DbDataReader, List<string>?, Task<TResult>>? dataReaderHandler,
             HttpContext? httpContext = null,
-            List<string>? args = null)
+            List<string>? args = null,
+            string dataSourceName = "")
         {
+            if (string.IsNullOrEmpty(dataSourceName))
```
vadeveka, line 73: Will this condition and providing empty string as default be removed in the next PR i.e. all the interfaces will know the exact datasource to use?

src/Config/RuntimeConfigLoader.cs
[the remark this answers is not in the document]
rohkhann: discussed offline. throwing exception now.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -249,13 +258,18 @@ private static RuntimeConfig HandleCosmosNoSqlConfiguration(string? schema, Runt
             options[namingPolicy.ConvertName(nameof(CosmosDbNoSQLDataSourceOptions.Database))] = JsonSerializer.SerializeToElement(database);
         }
 
+        dataSource = dataSource with { Options = options, ConnectionString = connectionString };
+
         // Update the connection string in the parsed config with the one that was provided to the controller
```
Aniruddh25, line 263: This comment should be moved up, since the connectionstring replacement happens on line 261 now.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -390,12 +396,13 @@ public static void ValidateGlobalEndpointRouteConfig(RuntimeConfig runtimeConfig
         /// <param name="runtimeConfig"></param>
         public static void ValidateRestURI(RuntimeConfig runtimeConfig)
         {
-            // CosmosDB_NoSQL does not support rest. No need to do any validations.
-            if (runtimeConfig.DataSource.DatabaseType is DatabaseType.CosmosDB_NoSQL)
+            if (runtimeConfig.DataSourceNameToDataSource.Values.Any(x => x.DatabaseType is DatabaseType.CosmosDB_NoSQL))
             {
+                // if we have a cosmos db even in multiple db scenario - no rest support.
```
Aniruddh25, line 401: If the other dbs are SQL, why would we want to restrict REST support for those?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -61,7 +61,8 @@ public class SqlQueryEngine : IQueryEngine
         /// </summary>
         /// <param name="context">HotChocolate Request Pipeline context containing request metadata</param>
         /// <param name="parameters">GraphQL Query Parameters from schema retrieved from ResolverMiddleware.GetParametersFromSchemaAndQueryFields()</param>
-        public async Task<Tuple<JsonDocument?, IMetadata?>> ExecuteAsync(IMiddlewareContext context, IDictionary<string, object?> parameters)
+        /// <param name="dataSourceName">Name of datasource for which to set access token. Default dbName taken from config if null</param>
```
Aniruddh25, line 64: Can this be null or simply an empty string?

src/Core/Resolvers/CosmosQueryEngine.cs
rohkhann: Umm, we are making multiple copies here right? If there are multiple db's we will register multiple unique instances which have their own connection strings.

src/Config/RuntimeConfigLoader.cs
rohkhann: fixed after offline discussion.

Aniruddh25 (changes requested):
Comments so far..

src/Config/Converters/RuntimeConfigExcludePropertiesConverter.cs
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the MIT License.
+
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.ObjectModel;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+
+internal class RuntimeConfigConditionalConverter : JsonConverter<RuntimeConfig>
```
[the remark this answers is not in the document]
vadeveka: Can you elaborate on the runtimeConfing specifics being used here? I did not notice any RunTimeConfig specific code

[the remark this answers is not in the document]
rohkhann: image
I think maybe the changes were not pushed. Also when you go to git, make sure to select all commits option and not specific one. i think it defaults to the commit this comment was on. Let me know if more questions.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -101,20 +185,30 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
     /// </summary>
     public static JsonSerializerOptions GetSerializationOptions()
     {
+        List<String> propertiesToExcludeForSerialization = new()
+        {
+            "DefaultDataSourceName",
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: fixed.

src/Core/Configurations/RuntimeConfigValidator.cs
rohkhann: In general are we allowing any REST experience for multiple datasources? i thought not and hence there should be no rest enabled for this scenario?

src/Core/Resolvers/SqlQueryEngine.cs
rohkhann: Ive default made it empty. It should not be null. in the future pr this will always be passed in from the callers of queryExecutor who need to be aware of the database to execute the query against.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -71,7 +71,91 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
+            if (dataSourceName is null || string.Equals(dataSourceName, config.DefaultDataSourceName, StringComparison.OrdinalIgnoreCase))
+            {
+                // single database scenario - default db is set.
+                config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+                config.DataSourceNameToDataSource[config.DefaultDataSourceName] = config.DataSource;
+            }
+            else
+            {
+                config.DataSourceNameToDataSource.TryGetValue(config.DefaultDataSourceName, out DataSource? ds);
+                if (ds is not null)
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -71,7 +71,91 @@ public static bool TryParseConfig(string json, [NotNullWhen(true)] out RuntimeCo
...
-            config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+            if (dataSourceName is null || string.Equals(dataSourceName, config.DefaultDataSourceName, StringComparison.OrdinalIgnoreCase))
+            {
+                // single database scenario - default db is set.
+                config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+                config.DataSourceNameToDataSource[config.DefaultDataSourceName] = config.DataSource;
+            }
+            else
+            {
+                config.DataSourceNameToDataSource.TryGetValue(config.DefaultDataSourceName, out DataSource? ds);
```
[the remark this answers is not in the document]
rohkhann: outdated file.

src/Config/ObjectModel/RuntimeConfig.cs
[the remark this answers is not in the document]
rohkhann: This constructor no longer exists.

src/Config/Converters/RuntimeConfigExcludePropertiesConverter.cs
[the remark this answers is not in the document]
rohkhann: discussed offline resolving.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -327,16 +332,18 @@ public static OkObjectResult OkResponse(JsonElement jsonResult)
         // Unlike a normal query, result from database may not be JSON. Instead we treat output as SqlMutationEngine does (extract by row).
         // As such, this could feasibly be moved to the mutation engine. 
         // </summary>
-        private async Task<JsonDocument?> ExecuteAsync(SqlExecuteStructure structure)
+        private async Task<JsonDocument?> ExecuteAsync(SqlExecuteStructure structure, string dataSourceName = "")
```
vadeveka, line 335: Do we need this defaulting to empty string here or other upstream method invocations if it is also defaulted downstream in executors?

src/Core/Resolvers/PostgreSqlExecutor.cs
```diff
@@ -27,24 +28,29 @@ public class PostgreSqlQueryExecutor : QueryExecutor<NpgsqlConnection>
         /// <summary>
         /// The managed identity Access Token string obtained
         /// from the configuration controller.
+        /// Key: datasource name, Value: access token for this datasource.
         /// </summary>
-        private readonly string? _accessTokenFromController;
+        private readonly Dictionary<string, string?> _accessTokensFromConfiguration;
```
vadeveka, line 33: NIT: Would it be simpler/efficient to have a single map with complex value rather than 3 different maps?

src/Config/NamingPolicies/HyphenatedNamingPolicy.cs
```diff
@@ -26,11 +25,6 @@ public sealed class HyphenatedNamingPolicy : JsonNamingPolicy
     /// <inheritdoc />
     public override string ConvertName(string name)
```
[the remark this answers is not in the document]
rohkhann: We would still need this class as other converters would use this to retrieve the hyphenatednamingpolicy.

src/Core/Resolvers/SqlQueryEngine.cs
rohkhann: This will be changed in future pr's If i dont default this here, then have to change the calls which I did not want to do as of now. This will be removed in the next pr.

src/Core/Resolvers/PostgreSqlExecutor.cs
rohkhann: Discussed offline. Good to have this will put in a task and move to later pr: https://github.com/Azure/data-api-builder/issues/1651

src/Config/RuntimeConfigLoader.cs
```diff
@@ -183,21 +142,28 @@ public static bool TryParseConfigMultipleDatabase(string json, [NotNullWhen(true
...
     /// <param name="replaceEnvVar">Whether to replace environment variable with value or not while deserializing.
     /// By default, no replacement happens.</param>
     public static JsonSerializerOptions GetSerializationOptions(bool replaceEnvVar = false)
     {
+        List<String> propertiesToExcludeForSerialization = new()
+        {
+            nameof(RuntimeConfig.DefaultDataSourceName),
+            nameof(RuntimeConfig.DataSourceNameToDataSource),
+            nameof(RuntimeConfig.EntityNameToDataSourceName)
+        };
```
seantleonard, line 158: this just essentially excludes fields users don't need to know or care about that are explicit to multi database scenario?

rohkhann: Yes!

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -74,7 +77,7 @@ public class QueryExecutor<TConnection> : IQueryExecutor
 
             if (!ConnectionStringBuilders.ContainsKey(dataSourceName))
             {
-                throw new Exception("Query execution failed. Could not find datasource to execute query against");
+                throw new DataApiBuilderException("Query execution failed. Could not find datasource to execute query against", HttpStatusCode.InternalServerError, DataApiBuilderException.SubStatusCodes.DataSourceNotFound);
```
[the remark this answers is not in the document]
rohkhann: Yes it would be. I changed this to badrequest.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -50,8 +52,14 @@ public class MsSqlQueryExecutor : QueryExecutor<SqlConnection>
         /// </summary>
         private AccessToken? _defaultAccessToken;
 
+        /// <summary>
+        /// DatasourceName to boolean value indicating if access token should be set for db.
+        /// </summary>
```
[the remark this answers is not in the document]
rohkhann: Fixed name.

seantleonard (approved):
Thank you for driving and implementing this new capability!

src/Config/RuntimeConfigLoader.cs
```diff
@@ -42,18 +43,24 @@ public RuntimeConfigLoader(string? connectionString = null)
     public abstract string GetPublishedDraftSchemaLink();
 
     /// <summary>
-    /// Parses a JSON string into a <c>RuntimeConfig</c> object
+    /// Parses a JSON string into a <c>RuntimeConfig</c> object for single database scenario.
     /// </summary>
     /// <param name="json">JSON that represents the config file.</param>
     /// <param name="config">The parsed config, or null if it parsed unsuccessfully.</param>
     /// <returns>True if the config was parsed, otherwise false.</returns>
     /// <param name="replaceEnvVar">Whether to replace environment variable with its
```
Aniruddh25, line 51: nit: the order of params in the summary here doesnt align with the actual order below

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2176,7 +2178,7 @@ private static JsonContent GetJsonContentForCosmosConfigRequest(string endpoint,
                     // exception. To prevent this, CosmosClientProvider parses the token and retrieves the "exp" property
                     // from the token, if it's not valid, then we will throw an exception from our code before it
                     // initiating a client. Uses a valid fake JWT access token for testing purposes.
-                    RuntimeConfig overrides = new(null, new DataSource(DatabaseType.CosmosDB_NoSQL, "AccountEndpoint=https://localhost:8081/;", new()), null, null);
+                    RuntimeConfig overrides = new(null, new DataSource(DatabaseType.CosmosDB_NoSQL, "AccountEndpoint=https://localhost:8081/;", new()), null, new(new Dictionary<string, Entity>()));
```
Aniruddh25, line 2181: nit: it will be easier to understand if this argument names are provided here.

src/Core/Resolvers/CosmosClientProvider.cs
```diff
@@ -49,38 +51,49 @@ private void InitializeClient(RuntimeConfig? configuration)
...
+
+                    ParseCosmosConnectionString(dataSource.ConnectionString);
+
+                    if (!string.IsNullOrEmpty(_accountKey))
+                    {
+                        client = new CosmosClientBuilder(dataSource.ConnectionString).WithContentResponseOnWrite(true)
+                            .WithApplicationName(userAgent)
+                            .Build();
+                    }
+                    else if (_accessToken.ContainsKey(dataSourceName))
```
Aniruddh25, line 82: Shouldn't this be NOT contain?

src/Core/Resolvers/CosmosClientProvider.cs
```diff
@@ -49,38 +51,49 @@ private void InitializeClient(RuntimeConfig? configuration)
...
+                    {
+                        TokenCredential servicePrincipal = new AADTokenCredential(_accessToken[dataSourceName]!);
+                        client = new CosmosClient(_accountEndpoint, servicePrincipal, options);
+                    }
+
+                    Clients.Add(dataSourceName, client);
                 }
             }
+
+            _defaultDataSourceName = configuration.DefaultDataSourceName;
```
Aniruddh25, line 96: why do we need this private member? 
Also, why do we need the private members - _accountKey, _accountEndpoint?
Looking at code for multiple scenario, seems like at the end of this loop those private members will be set to the value of the last datasource. Is that what we want? Or do we need those values to be  default datasource members values?

src/Config/RuntimeConfigLoader.cs
rohkhann: fixed.

src/Core/Resolvers/SqlQueryEngine.cs
rohkhann: fixed.

src/Core/Resolvers/IQueryEngine.cs
```diff
@@ -20,7 +20,7 @@ public interface IQueryEngine
         /// <returns>
         /// returns the json result and a metadata object required to resolve the Json.
         /// </returns>
-        public Task<Tuple<JsonDocument?, IMetadata?>> ExecuteAsync(IMiddlewareContext context, IDictionary<string, object?> parameters);
+        public Task<Tuple<JsonDocument?, IMetadata?>> ExecuteAsync(IMiddlewareContext context, IDictionary<string, object?> parameters, string dataSourceName = "");
```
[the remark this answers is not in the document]
rohkhann: Default value must be a runtime constant right? string.empty will throw error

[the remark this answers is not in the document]
rohkhann: discussed offline. will resolve.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -121,7 +121,8 @@ public void CheckCommentParsingInConfigFile()
                                             ""set-session-context"": true
                                         },
                                     ""connection-string"": ""Server=tcp:127.0.0.1,1433;Persist Security Info=False;Trusted_Connection=True;TrustServerCertificate=True;MultipleActiveResultSets=False;Connection Timeout=5;""
-                                    }
+                                    },
```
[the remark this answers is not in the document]
rohkhann: We changed entities to be a non-nullable type on the constructor and hence this is required. Otherwise it takes entities as null and then when it does the iteration it will throw null reference exception.

src/Core/Resolvers/CosmosClientProvider.cs
rohkhann: Removed default on class object. will be retrieved from runtime config like others. Fixed error related to accoundkey and accountendpoint pulling. We do not need those class objects now.

src/Core/Resolvers/CosmosClientProvider.cs
rohkhann: correct. fixed.

src/Service.Tests/Configuration/ConfigurationTests.cs
rohkhann: fixed.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -215,8 +217,13 @@ private async Task<bool> InvokeConfigLoadedHandlersAsync()
         return results.All(x => x);
     }
 
-    private static RuntimeConfig HandleCosmosNoSqlConfiguration(string? schema, RuntimeConfig runtimeConfig, string connectionString)
+    private static RuntimeConfig HandleCosmosNoSqlConfiguration(string? schema, RuntimeConfig runtimeConfig, string connectionString, string? dataSourceName = null)
```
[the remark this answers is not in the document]
rohkhann: made it "" everywhere.

src/Core/Resolvers/PostgreSqlExecutor.cs
```diff
@@ -160,7 +184,7 @@ private bool IsDefaultAccessTokenValid()
                 // has a valid connection string without a password in it
                 if (firstAttemptAtDefaultAccessToken)
                 {
-                    _attemptToSetAccessToken = false;
+                    _dataSourceAccessTokenUsage[ConfigProvider.GetConfig().DefaultDataSourceName] = false;
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -61,22 +69,30 @@ public override SqlConnectionStringBuilder ConnectionStringBuilder
...
+                {
+                    builder.Encrypt = SqlConnectionEncryptOption.Mandatory;
+                    builder.TrustServerCertificate = false;
+                }
 
-            MsSqlOptions? msSqlOptions = runtimeConfig.DataSource.GetTypedOptions<MsSqlOptions>();
-            _isSessionContextEnabled = msSqlOptions is null ? false : msSqlOptions.SetSessionContext;
-            _accessTokenFromController = runtimeConfigProvider.ManagedIdentityAccessToken;
-            _attemptToSetAccessToken = ShouldManagedIdentityAccessBeAttempted();
+                ConnectionStringBuilders.Add(dataSourceName, builder);
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -187,7 +200,7 @@ public class QueryExecutor<TConnection> : IQueryExecutor
         }
 
         /// <inheritdoc />
-        public virtual string GetSessionParamsQuery(HttpContext? httpContext, IDictionary<string, DbConnectionParam> parameters)
+        public virtual string GetSessionParamsQuery(HttpContext? httpContext, IDictionary<string, DbConnectionParam> parameters, string? dataSourceName = null)
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -200,7 +213,7 @@ public virtual void PopulateDbTypeForParameter(KeyValuePair<string, DbConnection
         }
 
         /// <inheritdoc />
-        public virtual async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection conn)
+        public virtual async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection conn, string? dataSourceName = null)
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -34,17 +34,19 @@ public class QueryExecutor<TConnection> : IQueryExecutor
 
         private AsyncRetryPolicy _retryPolicy;
 
-        public virtual DbConnectionStringBuilder ConnectionStringBuilder { get; set; }
+        /// <summary>
+        /// Dictionary that stores DbName to its corresponding connection string builder.
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -327,16 +332,18 @@ public static OkObjectResult OkResponse(JsonElement jsonResult)
...
 
             JsonArray? resultArray =
                 await _queryExecutor.ExecuteQueryAsync(
-                    queryString,
-                    structure.Parameters,
-                    _queryExecutor.GetJsonArrayAsync,
-                    _httpContextAccessor.HttpContext!);
+                    sqltext: queryString,
+                    parameters: structure.Parameters,
+                    dataReaderHandler: _queryExecutor.GetJsonArrayAsync,
```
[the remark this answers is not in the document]
rohkhann: This will be handled in next pr.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -390,12 +396,13 @@ public static void ValidateGlobalEndpointRouteConfig(RuntimeConfig runtimeConfig
         /// <param name="runtimeConfig"></param>
         public static void ValidateRestURI(RuntimeConfig runtimeConfig)
         {
-            // CosmosDB_NoSQL does not support rest. No need to do any validations.
-            if (runtimeConfig.DataSource.DatabaseType is DatabaseType.CosmosDB_NoSQL)
+            if (runtimeConfig.DataSourceNameToDataSource.Values.All(x => x.DatabaseType is DatabaseType.CosmosDB_NoSQL))
             {
+                // if all db's are cosmos no rest support.
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -66,22 +73,52 @@ public RuntimeConfigLoader(string? connectionString = null)
...
-                updatedConnectionString = GetConnectionStringWithApplicationName(updatedConnectionString);
+                datasourceNameToConnectionString = new Dictionary<string, string>();
             }
 
-            config = config with { DataSource = config.DataSource with { ConnectionString = updatedConnectionString } };
+            // add to dictionary if datasourceName is present (will either be the default or the one provided)
+            datasourceNameToConnectionString.TryAdd(dataSourceName, updatedConnectionString);
+
+            // iterate over dictionary and update runtime config with connection strings.
+            foreach ((string dataSourceKey, string connectionValue) in datasourceNameToConnectionString)
```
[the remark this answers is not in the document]
rohkhann: refactored.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -1,23 +1,214 @@
...
+    /// <summary>
+    /// Add entity to the EntityNameToDataSourceName dictionary.
+    /// </summary>
+    /// <param name="entityName">EntityName</param>
+    /// <param name="dataSourceName">DatasourceName.</param>
+    /// <returns>True indicating success, False indicating failure.</returns>
+    /// <exception cref="DataApiBuilderException">Not found exception if key is not found.</exception>
+    public bool TryAddEntity(string entityName, string dataSourceName)
+    {
+        CheckDataSourceNamePresent(dataSourceName);
```
Aniruddh25, line 149: Functions starting with Try e.g. here TryAddEntity usually shouldnt throw exceptions. Either change the name to AddEntity or avoid throwing exception, just return false.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -1,23 +1,214 @@
...
+        // get default serializer options if none provided.
+        jsonSerializerOptions = jsonSerializerOptions ?? RuntimeConfigLoader.GetSerializationOptions();
+        return JsonSerializer.Serialize(this, jsonSerializerOptions);
+    }
+
+    private void CheckDataSourceNamePresent(string dataSourceName)
+    {
+        if (!_dataSourceNameToDataSource.ContainsKey(dataSourceName))
+        {
+            throw new DataApiBuilderException($"{nameof(dataSourceName)} could not be found within the config", HttpStatusCode.BadRequest, DataApiBuilderException.SubStatusCodes.DataSourceNotFound);
```
Aniruddh25, line 203: shouldn't we log the actual value of dataSourceName that is not found as well? So this is debuggable?

Aniruddh25 (approved):
Thanks for working on this, and your patience while it took time for us to review! LGTM after last bit of comments are resolved.

PR #1641, opened 2023-08-18
Replace environment variable only for dab `start`

### LINKED ISSUE #1632: [Bug]: Enviroment Variables are expanded/inlined in the configuration file
### What happened?

I have created a new configuration file using the following DAB CLI command:

```
dab init --database-type mssql --host-mode development --connection-string "@env('MSSQL')"
```

and I have a `.env` file that contains the environment variable with the connection string:

```
MSSQL='<Connection String Here>'
```

If I now add a entity via DAB CLI again:

```
dab add Todos --source dbo.todos --permissions "anonymous:*" --rest todos
```

Now the configuration file is updated and look like the following, where the environment variable reference is replaced with its value:

```
"data-source": {
    "database-type": "mssql",
    "connection-string": "\u003CConnection String Here\u003E",
    "options": {
      "set-session-context": false
    }
  },
``` 

This is wrong and potentially a security concern as in the config file there will now be the full connection string, with login and password details.

### Version

Microsoft.DataApiBuilder 0.8.44-rc+ebe4ef414deaf7670b579fe160b872c6b682c114

Azure SQL

Local (including CLI)

REST

### Relevant log output

_No response_

### Code of Conduct

Aniruddh25:
## Why make this change?

- Closes #1632 
- it's an unintended side effect of an optimisation pathway was added to the deserializer in #1402.
Previously, deserializing the JSON config and replacing environment variables were done as two separate parses of the config JSON blob. First the JSON would be read from disk and then we'd create a JSON reader that would walk all the tokens in the file, we'd check the token to see if it matched the regex for an environment variable, we'd replace it and then return the updated JSON string. Next this would be given to the deserializer and it would be turned into the .NET object that we would use throughout the engine. This meant that we'd end up walking the AST for the JSON twice.
- With #1402, when we are doing deserialization we would also do the environment variable replacement, meaning that we'd only walk the JSON AST once, thus improving performance.
- There is a downside to this - the CLI doesn't need to do environment variable replacements (since we don't need to actually connect to the DB or anything, except for start, but that doesn't write out an updated config).

## What is this change?

- Introduces a flag replaceEnvVar to determine whether to do the environment variable replacement in the JsonConverters when being used from the CLI commands. For `start` command this will be `true`, for all other commands this is `false`. 
- To minimize changes, default value of this argument was kept false. All occurrences of callers requiring this flag were modified and set to `true` in case of `start` or running the engine.
- For Enums, always replace the environment variable with its value for appropriate deserialization into an Enum value.

## How was this tested?

- [X] Integration Tests - Modified the End2EndTests, to test the commands `add`, `update` dont replace the environment variables, but `start` replaces it.
- [X] Unit Tests - Modified the `CheckConfigEnvParsingTest` to test when replaceEnvVar is `false` we dont replace them. The `true` scenario was already being tested. Also modified the `database-type` property to be an env variable for testing enums are always replaced with values.


abhishekkumams:
`replaceVar` should be false by default when running through CLI, but it should be true for when running directly through Engine.
I would say make it true by default, and send false from CLI in other cases. This will fix the failing tests.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -112,9 +124,11 @@ public static JsonSerializerOptions GetSerializationOptions()
         options.Converters.Add(new EnumMemberJsonEnumConverterFactory());
         options.Converters.Add(new RestRuntimeOptionsConverterFactory());
         options.Converters.Add(new GraphQLRuntimeOptionsConverterFactory());
-        options.Converters.Add(new EntitySourceConverterFactory());
+        options.Converters.Add(new EntitySourceConverterFactory(replaceEnvVar));
+        options.Converters.Add(new EntityGraphQLOptionsConverterFactory(replaceEnvVar));
+        options.Converters.Add(new EntityRestOptionsConverterFactory(replaceEnvVar));
         options.Converters.Add(new EntityActionConverterFactory());
-        options.Converters.Add(new StringJsonConverterFactory());
+        options.Converters.Add(new StringJsonConverterFactory(replaceEnvVar));
```
aaronpowell, line 131: Rather than having to pass down this variable, would it be easier to not include this converter if the environment variable is set?

I was thinking that you'd just put a check around this line of _if flag disables replacement, don't include_ rather than having to have logic within the converters.

abhishekkumams (approved):

src/Config/RuntimeConfigLoader.cs
Aniruddh25: You mean in those cases rely on default String converter? Are we losing any other functionality because of which we needed this custom converter? Or was the custom converter only needed for environment variable replacement?

aaronpowell: Yes, the only job of that `JsonConverter` is to apply the env-var replacement rules.

Aniruddh25: makes sense. fixed.

src/Config/Converters/EntityGraphQLOptionsConverterFactory.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// value or not while deserializing.</param>
+        public EntityGraphQLOptionsConverter(bool replaceEnvVar)
+        {
+            _replaceEnvVar = replaceEnvVar;
+        }
+
+        /// <inheritdoc/>
+        public override EntityGraphQLOptions? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            if (reader.TokenType == JsonTokenType.StartObject)
```
aaronburtle, line 51: nit: we favor use of `is` key word instead of `==`, though I think this has been missed in some classes. Many cases here if you want to change them.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -181,7 +209,7 @@ private static void SetEnvVariables()
         /// </summary>
         /// <param name="reps">Replacement strings.</param>
         /// <returns>Json string with replacements.</returns>
-        public static string GetModifiedJsonString(string[] reps)
+        public static string GetModifiedJsonString(string[] reps, string enumString)
```
seantleonard, line 212: nit: comment needed for `enumString`. From the test's code, it seems this value is used to provide an actual value for 'database-type' regardless of whether `replaceEnvVar` is true or false?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -60,15 +62,40 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+                {
+                    Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
+                        GetModifiedJsonString(repValues, @"""mssql"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
+                        "Should read the expected config");
+                }
+                else
+                {
+                    Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
+                        GetModifiedJsonString(repKeys, @"""@env('enumVarName')"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
+                        "Should read the expected config");
```
seantleonard, line 93: For Enum's i thought you were trying to always replace the env_var with an actual value otherwise deserializing the json would throw an exception?

src/Cli/ConfigGenerator.cs
```diff
@@ -969,7 +969,8 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
             loader.UpdateBaseConfigFileName(runtimeConfigFile);
 
             // Validates that config file has data and follows the correct json schema
-            if (!loader.TryLoadKnownConfig(out RuntimeConfig? deserializedRuntimeConfig))
+            // replace all the environment variables while deserializing.
```
seantleonard, line 972:
```suggestion
            // Replaces all the environment variables while deserializing when starting DAB.
```

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -52,7 +54,7 @@ public void TestCleanup()
     public Task TestInitForCosmosDBNoSql()
     {
         string[] args = { "init", "-c", TEST_RUNTIME_CONFIG_FILE, "--database-type", "cosmosdb_nosql",
-                          "--connection-string", "localhost:5000", "--cosmosdb_nosql-database",
+                          "--connection-string", TEST_ENV_CONN_STRING, "--cosmosdb_nosql-database",
```
seantleonard, line 57: This test uses `Verify()` , are there verify.txt files that need updating due to `TEST_CONNECTION_STRING` representing an env_var key?

src/Cli.Tests/TestHelper.cs
```diff
@@ -41,6 +44,7 @@ public static Process ExecuteDabCommand(string command, string flags)
                 StartInfo =
                 {
                     FileName = @"./Microsoft.DataApiBuilder",
+                    CreateNoWindow = true,
```
seantleonard, line 47: Does this mitigate the issues seen in the previously ignored test in EnvironmentTests caused by hangs?

seantleonard (commented):
Some questions

src/Config/RuntimeConfigLoader.cs
```diff
@@ -112,9 +124,16 @@ public static JsonSerializerOptions GetSerializationOptions()
         options.Converters.Add(new EnumMemberJsonEnumConverterFactory());
         options.Converters.Add(new RestRuntimeOptionsConverterFactory());
         options.Converters.Add(new GraphQLRuntimeOptionsConverterFactory());
-        options.Converters.Add(new EntitySourceConverterFactory());
+        options.Converters.Add(new EntitySourceConverterFactory(replaceEnvVar));
+        options.Converters.Add(new EntityGraphQLOptionsConverterFactory(replaceEnvVar));
+        options.Converters.Add(new EntityRestOptionsConverterFactory(replaceEnvVar));
```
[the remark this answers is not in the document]
aaronburtle: I think factory pattern just fits nicely here.

The CLI is going to be using the same `TryLoadKnownConfig` function that comes from the `FileSystemRuntimeConfigLoader`, but it has no reason to do the environmental variable parsing in most cases. The factory pattern here lets the CLI generate the right derived class that has the behavior it wants with as you noticed a super simple addition of simply passing in this flag. So with that flag we get the derived class with the deserialization behavior that won't replace all those env variables.

It could be done other ways, but this is pattern that can slot into the existing code nicely in my opinion.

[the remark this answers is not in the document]
aaronpowell: When I did the original implementation, the reason that some used factories and others didn't was because there were some instances where we would want to remove the converter dynamically. This would happen when the fully expanded .NET type mapped directly to the JSON structure (when not using a shorthand), and for those cases, rather than walking through the JSON Reader it'd just call into the deserialization engine again. The only problem would be that if the converter was still there, it would become a recursive call, eventually resulting in exceeding the call stack depth.

The only workaround for this is to "manually" add the converter with a factory rather than relying on the attribute-discovery, as if it's discovered by an attribute you can't remove it dynamically.

src/Cli.Tests/EndToEndTests.cs
aaronpowell: Connection strings are ignored in verify tests to avoid accidentally committing credentials to the repo.

src/Config/Converters/StringJsonConverterFactory.cs
```diff
@@ -44,8 +48,7 @@ class StringJsonConverter : JsonConverter<string>
             if (reader.TokenType == JsonTokenType.String)
             {
                 string? value = reader.GetString();
-
-                return Regex.Replace(reader.GetString()!, ENV_PATTERN, new MatchEvaluator(ReplaceMatchWithEnvVariable));
+                return Regex.Replace(value!, ENV_PATTERN, new MatchEvaluator(ReplaceMatchWithEnvVariable));
```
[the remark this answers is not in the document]
Aniruddh25: value is nothing but reader.GetString() which is already stored in the previous line. We don't need to get the string again from the reader - its gonna give us the same 'value'. reader doesn't advance unless we go out of this function. Also, its guaranteed to be non-null since we already checked that the TokenType is String.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -54,6 +54,7 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
     /// is known by the loader.
     /// </summary>
     /// <returns>The RuntimeConfig instance.</returns>
+    /// <remark>Dont use this method if environment variables need to be retained.</remark>
```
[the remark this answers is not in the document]
Aniruddh25: Yes.

src/Config/RuntimeConfigLoader.cs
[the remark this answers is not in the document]
Aniruddh25: And the reason I added the factory is yes, because I wanted to pass on `replaceEnvVar` to the constructor otherwise there was chance the default constructor would be picked up. Even if the right constructor is invoked, we wouldn't be able to pass in the argument of the `GetSerializationOptions` function downstream to the constructor that would be invoked.

src/Config/Converters/EnumMemberJsonEnumConverterFactory.cs
```diff
@@ -113,7 +113,8 @@ public JsonStringEnumConverterEx()
         /// <inheritdoc/>
         public override TEnum Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
-            string? stringValue = reader.DeserializeString();
+            // Always replace env variable in case of Enum otherwise string to enum conversion will fail.
```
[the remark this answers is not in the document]
Aniruddh25: yes, or even the databasetype enum

src/Config/Converters/EntityRestOptionsConverterFactory.cs
```diff
@@ -0,0 +1,162 @@
...
+            if (value.Methods is not null && value.Methods.Length > 0)
+            {
+
+                writer.WriteStartArray("methods");
+                foreach (SupportedHttpVerb method in value.Methods)
+                {
+                    writer.WriteStringValue(JsonSerializer.SerializeToElement(method, options).GetString());
+                }
+
+                writer.WriteEndArray();
```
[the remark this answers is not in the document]
Aniruddh25: Yes, makes sense!

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
Aniruddh25: Thats right, fixed. Yes its used to provide actual value when called for the expectedConfig. For actual config, this would be `@env('enumVarName')`

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
Aniruddh25: Thats right, fixed it.

aaronburtle (approved):
Looks good, thanks for adding this fix!

src/Cli.Tests/TestHelper.cs
Aniruddh25: Yes. seems like it, when I ran the test locally, I noticed VS debug window popping up to get user input for debugging the failed start up of engine in the child process(- the failure is what the test expects). I had to click cancel and seems like in the pipeline it may have been hanging waiting for a similar user input. After adding this change, looks like the hang is gone from the pipeline so that confirms this fixed the issue.

Aniruddh25:
> `replaceVar` should be false by default when running through CLI, but it should be true for when running directly through Engine. I would say make it true by default, and send false from CLI in other cases. This will fix the failing tests.

Prefer to explicitly call replaceEnvVar=true since we dont want to accidentally leak any connectionstring if its happens by default.

Issue #1598, opened 2023-07-20
[Bug]: Failed to convert parameter value from a String to a Guid.


yorek:
### What happened?

Using GraphQL, columns using GUIDs cannot be filtered, for example:

```graphql
query {
  todos(filter: { category_id: {eq: "0B216E5C-C302-4555-9F3D-F2E856A6D79B"}}) {
    items {
      id
      title
      owner_id
    }
  }
}
```

generates and error. The table is defined as:

```sql
CREATE TABLE [dbo].[todos](
	[id] [uniqueidentifier] NOT NULL,
	[category_id] [uniqueidentifier] NULL,
	[title] [nvarchar](1000) NOT NULL,
	[completed] [bit] NOT NULL,
	[custom] [nvarchar](max) NULL,
	[owner_id] [varchar](128) NOT NULL,
	[created_on] [datetime2](7) NOT NULL,
	[position] [int] NULL,
	[created_on_date]  AS (CONVERT([date],[created_on]))
)
```

and the entity configuration, in the dab-config, is:

```json
    "Todos": {
      "source": "dbo.todos",
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            "*"
          ]
        }
      ],
      "rest": {
        "path": "/todos"
      }
```

with REST everything works fine, and I can filter the rows:

```http
https://localhost:5001/api/todos?$filter=category_id%20eq%2000000000-0000-0000-0000-000000000000
```

### Version

Microsoft.DataApiBuilder 0.7.6+61de247acf65280d93783072226f695536591ffb

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

GraphQL

### Relevant log output

```Text
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'Hot Chocolate GraphQL Pipeline'
dbug: Azure.DataApiBuilder.Service.Resolvers.IQueryExecutor[0]
      e46c92e2-17f4-4ed7-998f-6849745d18cc: Executing query:
SELECT TOP 100 [table0].[id] AS [id], [table0].[title] AS [title], [table0].[owner_id] AS [owner_id] FROM [dbo].[todos] AS [table0] WHERE [table0].[category_id] = @param1 ORDER BY [table0].[id] ASC FOR JSON PATH, INCLUDE_NULL_VALUES
fail: Azure.DataApiBuilder.Service.Startup[0]
[...]

PR #1644, opened 2023-08-22
Fix for guid filter in GraphQL

ayush3797:
## Why make this change?
Fix https://github.com/Azure/data-api-builder/issues/1598 to ensure GQL filter works with guid types.

## What is this change?
Correctly resolved guid type to Hotchocolate type of `UUID` (per: https://chillicream.com/docs/hotchocolate/v12/defining-a-schema/scalars#uuid-type) which was earlier resolved to string. Accordingly parsed the guid values using `Guid.Parse()` method instead of parsing it as string.

## How was this tested?
- [x] Unit Tests - `SchemaConverterTests`
- [X] Integration Tests - Added filter tests with GUID to `GraphQLSupportedTypesTestsBase.QueryTypeColumnFilterAndOrderBy`


src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -46,6 +46,7 @@ public static bool IsBuiltInType(ITypeNode typeNode)
             {
                 "ID",
                 BYTE_TYPE,
+                GUID_TYPE,
```
Aniruddh25, line 49: How is "ID" different than GUID_TYPE?

src/Service.Tests/DatabaseSchema-PostgreSql.sql
```diff
@@ -328,6 +328,7 @@ INSERT INTO type_table(id, short_types, int_types, long_types, string_types, sin
     (3, -32768, -2147483648, -9223372036854775808, '', -3.4E38, -1.7E308, 2.929292E-19, true, '1753-01-01 00:00:00.000', '\x00000000'),
     (4, 32767, 2147483647, 9223372036854775807, 'null', 3.4E38, 1.7E308, 2.929292E-14, true, '9999-12-31 23:59:59.997', '\xFFFFFFFF'),
     (5, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
+INSERT INTO type_table(id, guid_types) values(10, 'D1D021A8-47B4-4AE4-B718-98E89C41A161');
```
Aniruddh25, line 331: Why not MySQL?

Aniruddh25 (changes requested):
Thanks for the quick fix! Few suggestions

src/Service.Tests/DatabaseSchema-PostgreSql.sql
ayush3797: There is not data type for GUID in MySql.

src/Service.GraphQLBuilder/GraphQLUtils.cs
ayush3797: Actually, this change I think is not required. GUID_TYPE is not a built-in type for hotchocolate. Guids correspond to ID type in Hotchocolate : https://chillicream.com/docs/hotchocolate/v13/defining-a-schema/scalars#id.

src/Core/Services/ResolverMiddleware.cs
```diff
@@ -265,6 +265,8 @@ protected static bool IsInnerObject(IMiddlewareContext context)
                 SINGLE_TYPE => ((FloatValueNode)value).ToSingle(),
                 FLOAT_TYPE => ((FloatValueNode)value).ToDouble(),
                 DECIMAL_TYPE => ((FloatValueNode)value).ToDecimal(),
```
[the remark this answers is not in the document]
ayush3797: Yeah the different scalars supported by Hotchocolate are listed here: https://chillicream.com/docs/hotchocolate/v13/defining-a-schema/scalars. We can create a constant for ID type as well.

Aniruddh25 (approved):
LGTM.

src/Core/Services/ResolverMiddleware.cs
```diff
@@ -265,6 +265,8 @@ protected static bool IsInnerObject(IMiddlewareContext context)
...
+                // If we reach here, we can be sure that the value will not be null.
+                ID_TYPE => Guid.Parse(value.Value!.ToString()!),
```
Aniruddh25, line 269: Do we have tests for CosmosDB that check the Guid type? If yes, are those affected in any way. Just checking to ensure there are no regressions.

abhishekkumams: Cosmos DB has an entity called Planet, which has ID of Guid type.

src/Core/Services/ResolverMiddleware.cs
```diff
@@ -265,6 +265,10 @@ protected static bool IsInnerObject(IMiddlewareContext context)
                 SINGLE_TYPE => ((FloatValueNode)value).ToSingle(),
                 FLOAT_TYPE => ((FloatValueNode)value).ToDouble(),
                 DECIMAL_TYPE => ((FloatValueNode)value).ToDecimal(),
+                // If we reach here, we can be sure that the value will not be null.
+                // ID type can store int, string or Guid. First we try to parse the value as Guid.
+                // If the value is not a Guid, we just return the value.
+                ID_TYPE => Guid.TryParse(value.Value!.ToString(), out Guid guidValue) ? guidValue : value.Value,
```
ayush3797, line 271: @Aniruddh25 cosmos can treat int values like 1,2,3.. etc for (fields like `id` shown in the image) and guids as well, as Hotchocolate ID:
image

whereas Sql only treats `System.Guid` type as Hotchocolate id.

So, when we find a field with Hotchocolate type as 'ID', we will first try to parse it as a GUID. If the parsing returned false, we will return the value.Value as it is.

This is kind of a fix for comos as well because for cosmos, if they try to have a GUID column today with hotchocolate type of 'ID', filters would not work, just like they don't work for Sql.

src/Core/Services/ResolverMiddleware.cs
ayush3797: Continuing conversation here: https://github.com/Azure/data-api-builder/pull/1644#discussion_r1304146910

src/Core/Services/TypeHelper.cs
```diff
@@ -43,7 +43,8 @@ public static class TypeHelper
             [typeof(byte[])] = DbType.Binary,
             [typeof(TimeOnly)] = DbType.Time,
             [typeof(TimeSpan)] = DbType.Time,
-            [typeof(object)] = DbType.Object
+            [typeof(object)] = DbType.Object,
+            [typeof(Guid)] = DbType.Guid
```
seantleonard, line 47:
```suggestion
```

This already exists on line 41

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -220,7 +220,7 @@ public static string GetGraphQLTypeFromSystemType(Type type)
             return type.Name switch
             {
                 "String" => STRING_TYPE,
-                "Guid" => STRING_TYPE,
+                "Guid" => ID_TYPE,
```
seantleonard, line 223: Why `ID_TYPE` and not `UUID_TYPE` https://chillicream.com/docs/hotchocolate/v12/defining-a-schema/scalars/#uuid-type

src/Core/Services/ResolverMiddleware.cs
seantleonard: I don't understand this assumption: "whereas Sql only treats System.Guid type as Hotchocolate id."

Your code comment captured an important point raised by GraphQL spec: 
> https://spec.graphql.org/October2021/#sel-FAHXZDCAOCEBpjG
> The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is serialized in the same way as a String; however, it is not intended to be human-readable. While it is often numeric, it should always serialize as a String.

We follow this portion of the spec(assisted by HotChocolate serializing the field as a string), and then it is up to us (the server) to coerce the value into an "expected format". 

> [Result Coercion](https://spec.graphql.org/October2021/#sec-ID.Result-Coercion)
GraphQL is agnostic to ID format, and serializes to string to ensure consistency across many formats ID could represent, from small auto-increment numbers, to large 128-bit random numbers, to base64 encoded values, or string values of a format like [GUID](https://en.wikipedia.org/wiki/Globally_unique_identifier).
> 
> GraphQL services should coerce as appropriate given the ID formats they expect. When coercion is not possible they must raise a field error.
>
> [Input Coercion](https://spec.graphql.org/October2021/#sec-ID.Input-Coercion)
When expected as an input type, any string (such as "4") or integer (such as 4 or -4) input value should be coerced to ID as appropriate for the ID formats a given GraphQL service expects. Any other input value, including float input values (such as 4.0), must raise a request error indicating an incorrect type

What I don't understand with this PR's changes is that we are saying that a GUID column in MsSql **is a GraphQL ID field type**. However, such a column may not actually be a unique identifier for the particular row in the database, even though GUID is inherently a unique value.

src/Core/Services/TypeHelper.cs
ayush3797: Thanks, removed.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
ayush3797: Seemed to be the appropriate choice. Thanks @Aniruddh25 for making the changes.

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
```diff
@@ -8,6 +8,7 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder.GraphQLTypes
     /// </summary>
     public static class SupportedTypes
     {
+        public const string ID_TYPE = "ID";
```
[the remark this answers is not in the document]
ayush3797: GUID_TYPE  = 'Guid' is not a hotchocolate type while ID_TYPE is.

src/Core/Services/ResolverMiddleware.cs
ayush3797: I see your point here. Apologies for the confusion but I meant that with the implementation I had, sql will treat guid as hotchocolate id type. You have a valid point here, a guid column can possibly not be the unique identifier aka ID of a record. Mapping of System.Guid to Hotchocolate.UUID makese sense, yes.

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
```diff
@@ -8,6 +8,7 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder.GraphQLTypes
     /// </summary>
     public static class SupportedTypes
     {
+        public const string UUID_TYPE = "UUID";
```
seantleonard, line 11: Do we need to remove GUID_TYPE from this list? i'm a little confused on the mixed usage of `UUID_TYPE` and `GUID_TYPE`

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -715,7 +754,7 @@ private static string TypeNameToGraphQLType(string typeName)
         {
             if (typeName is GUID_TYPE)
             {
-                return STRING_TYPE;
+                return UUID_TYPE;
```
seantleonard, line 757: instance of mixed GUID_TYPE and UUID_TYPE usage

seantleonard (commented):
just last two questions.

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
Aniruddh25: Removed. GUID_TYPE was only being used to identify the column name "guid_types" in the test type_table. Renamed the column to uuid_types to avoid confusion. But essentially, GUID_TYPE is equivalent to UUID_TYPE

src/Service.Tests/Unittests/ODataASTVisitorUnitTests.cs
```diff
@@ -144,23 +144,23 @@ public void VisitorLeftFieldGreaterThanRightNullFilterTest()
         // Constant on left side and OData EDM object on right side of binary operator. (L->R)
         [DataRow("'1' eq int_types", false, DisplayName = "L->R: Cast token claim of type string to integer")]
         [DataRow("12.24 eq float_types", false, DisplayName = "L->R: Cast token claim of type single to type double (CLR) which maps to (SQL) float")]
-        [DataRow("'13B4F4EC-C45B-46EC-99F2-77BC22A256A7' eq guid_types", false, DisplayName = "L->R: Cast token claim of type string to GUID")]
+        [DataRow("'13B4F4EC-C45B-46EC-99F2-77BC22A256A7' eq UUID_TYPEs", false, DisplayName = "L->R: Cast token claim of type string to GUID")]
```
Aniruddh25, line 147:
```suggestion
        [DataRow("'13B4F4EC-C45B-46EC-99F2-77BC22A256A7' eq uuid_types", false, DisplayName = "L->R: Cast token claim of type string to GUID")]
```

src/Service.Tests/Unittests/ODataASTVisitorUnitTests.cs
```diff
@@ -144,23 +144,23 @@ public void VisitorLeftFieldGreaterThanRightNullFilterTest()
...
         [DataRow("'true' eq boolean_types", false, DisplayName = "L->R: Cast token claim of type string to bool (true)")]
         [DataRow("'false' eq boolean_types", false, DisplayName = "L->R: Cast token claim of type string to bool (false)")]
         [DataRow("1 eq string_types", false, DisplayName = "L->R: Cast token claim of type int to string")]
         [DataRow("true eq string_types", false, DisplayName = "L->R: Cast token claim of type bool to string")]
         // Constant on right side and OData EDM object on left side of binary operator. (R->L)
         [DataRow("int_types eq '1'", false, DisplayName = "R->L: Cast token claim of type string to integer")]
         [DataRow("float_types eq 12.24", false, DisplayName = "R->L: Cast token claim of type single to type double (CLR) which maps to (SQL) float")]
-        [DataRow("guid_types eq '13B4F4EC-C45B-46EC-99F2-77BC22A256A7'", false, DisplayName = "R->L: Cast token claim of type string to GUID")]
+        [DataRow("UUID_TYPEs eq '13B4F4EC-C45B-46EC-99F2-77BC22A256A7'", false, DisplayName = "R->L: Cast token claim of type string to GUID")]
```
Aniruddh25, line 155:
```suggestion
        [DataRow("uuid_types eq '13B4F4EC-C45B-46EC-99F2-77BC22A256A7'", false, DisplayName = "R->L: Cast token claim of type string to GUID")]
```

src/Service.Tests/Unittests/ODataASTVisitorUnitTests.cs
```diff
@@ -144,23 +144,23 @@ public void VisitorLeftFieldGreaterThanRightNullFilterTest()
...
         [DataRow("boolean_types eq 'true'", false, DisplayName = "R->L: Cast token claim of type string to bool (true)")]
         [DataRow("boolean_types eq 'false'", false, DisplayName = "R->L: Cast token claim of type string to bool (false)")]
         [DataRow("string_types eq 1", false, DisplayName = "R->L: Cast token claim of type int to string")]
         [DataRow("string_types eq true", false, DisplayName = "R->L: Cast token claim of type bool to string")]
         // Comparisons expected to fail due to inability to cast
         [DataRow("boolean_types eq 2", true, DisplayName = "Fail to cast arbitrary int to bool")]
-        [DataRow("guid_types eq 1", true, DisplayName = "Fail to cast arbitrary int to GUID")]
-        [DataRow("guid_types eq 'stringConstant'", true, DisplayName = "Fail to cast arbitrary string to GUID")]
+        [DataRow("UUID_TYPEs eq 1", true, DisplayName = "Fail to cast arbitrary int to GUID")]
```
Aniruddh25, line 162:
```suggestion
        [DataRow("uuid_types eq 1", true, DisplayName = "Fail to cast arbitrary int to GUID")]
```

src/Service.Tests/Unittests/ODataASTVisitorUnitTests.cs
```diff
@@ -144,23 +144,23 @@ public void VisitorLeftFieldGreaterThanRightNullFilterTest()
...
+        [DataRow("UUID_TYPEs eq 'stringConstant'", true, DisplayName = "Fail to cast arbitrary string to GUID")]
```
Aniruddh25, line 163:
```suggestion
        [DataRow("uuid_types eq 'stringConstant'", true, DisplayName = "Fail to cast arbitrary string to GUID")]
```

seantleonard (approved):
thanks for addressing concerns!

PR #1647, opened 2023-08-23
Fixing bug with handling output columns - MySql

ayush3797:
## Why make this change?
Fix https://github.com/Azure/data-api-builder/issues/1648

Currently Output columns are incorrectly handled while generating SELECT query for MySql after update via UPSERT operation.
Relevant code snippet:
image
Link to file:
https://github.com/Azure/data-api-builder/blob/5e3874e6a77179523d270a0358efdcbbea9762fa/src/Core/Resolvers/MySqlQueryBuilder.cs#L293

**Where is the bug?**
As we can observe that in the attached code snippet, we are processing the output columns one by one. The IF block for line 293 is hit when we find a column which is present in the `structure.InsertColumns`. We add such a column to the final `selections` and increment the index:

`selections.Add($"{structure.Values[index]} AS {quotedColName}");`
`index++;`

**What have we assumed here?**
We have assumed that the order of columns in `structure.InsertColumns` is same as `structure.OutputColumns`. Well that is certainly not true as output columns can have columns other that insertColumns and even if the columns are same, the order can differ. **So, we cannot just reference `structure.Values[index]` on line 293 and increment the index on line 294.**. We need to be sure that this output column, which is also an insert column, corresponds to this parameter name at `structure.Values[index]` (structure.Values essentially stores parameter names for each of the insert columns).
## What is this change?
- Instead of incrementing index, we need a dictionary `insertColumnsToParamName` which certainly tells us as to what paramName does each of the insertColumns map to. **This also prevents an O(n) lookup** that we were doing to check if the output column is present in the list of `structure.InsertColumns`.

## How was this tested?
Will be adding tests in: https://github.com/Azure/data-api-builder/pull/1596, if required at all. The existing tests are still passing which confirms the change is good.


src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -280,26 +280,26 @@ private string MakeInsertSelections(SqlInsertStructure structure)
...
             {
                 string quotedColName = QuoteIdentifier(column.Label);
-
-                if (structure.InsertColumns.Contains(column.ColumnName))
+                ColumnDefinition columnDefinition = structure.GetColumnDefinition(column.ColumnName);
+                if (insertColumnsToParamName.TryGetValue(column.ColumnName, out string? paramName))
                 {
-                    selections.Add($"{structure.Values[index]} AS {quotedColName}");
-                    index++;
+                    selections.Add($"{paramName!} AS {quotedColName}");
```
Aniruddh25, line 294: Do we need the null forgiving operator here after paramName? Since its true, isnt it guaranteed to be not null?

Aniruddh25 (approved):
LGTM, Thanks for fixing this!

abhishekkumams (approved):

src/Core/Resolvers/MySqlQueryBuilder.cs
ayush3797: yeah correct thanks.

PR #1649, opened 2023-08-24
Enable Application Insights For Non Hosted Scenario

### LINKED ISSUE #1601: RFC: Enable Application Insights
https://learn.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core?tabs=netcorenew%2Cnetcore6
```[tasklist]
### Tasks
- [ ] https://github.com/Azure/data-api-builder/issues/1704
- [ ] https://github.com/Azure/data-api-builder/pull/1649
```

abhishekkumams:
## Why make this change?

- Closes #1601
  - This PR aims to Integrate Application Insights to Data API builder, which will allow developers to monitor and gain insights into the performance and usage of their applications.

## What is this change?

- Added a new optional property called `Telemetry` to runtime config that will contain information about ApplicationInsights connection string. users can enable/disable it as well.
- If Application Insights is enabled, All the Application Logs that is currently generated by Data API builder will now be sent to Application Insights.

## How was this tested?

- [X] Manual Tests: Logs are being populated in the Application Insights portal
- [X] Unit Tests

## Images from Azure Portal
1. Live Metrics

image

2. Logs

image

3. Requests

image

4. Exceptions

image


src/Config/ObjectModel/TelemetryOptions.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
+
+/// <summary>
+/// Represents the options for telemetry.
+/// </summary>
+public record TelemetryOptions(ApplicationInsightsOptions ApplicationInsights)
```
seantleonard, line 9: How is this level of abstraction going to be utilized when/if we ever support other telemetry mechanisms? Would this record be updated to something like
```csharp
TelemetryOptions(ApplicationInsightsOptions? ApplicationInsights = null, OtherTelemetryProviderOptions providerOptions? = null)

File: src/Core/Services/ResolverMiddleware.cs [hunk 2]
```diff
@@ -62,6 +66,12 @@ public async Task InvokeAsync(IMiddlewareContext context)
 
             if (context.Selection.Field.Coordinate.TypeName.Value == "Mutation")
             {
+                // Track event when request is received
+                if (_telemetryClient is not null)
+                {
+                    TrackEventForGraphQLRequests(context, _telemetryClient);
+                }
```
seantleonard, line 73: Should these events come in a separate PR? I'd imagine we'd want to allow users to customize which events are logged, if they want any additional events at all.

For a highly utilized app, this event (and others) would flood the logs and increase costs to use AppInsights.

src/Service/Controllers/RestController.cs
```diff
@@ -47,15 +49,17 @@ public class RestController : ControllerBase
         public const string REDIRECTED_ROUTE = "favicon.ico";
 
         private readonly ILogger<RestController> _logger;
+        private readonly TelemetryClient? _telemetryClient;
 
         /// <summary>
         /// Constructor.
         /// </summary>
-        public RestController(RestService restService, IOpenApiDocumentor openApiDocumentor, ILogger<RestController> logger)
+        public RestController(RestService restService, IOpenApiDocumentor openApiDocumentor, ILogger<RestController> logger, TelemetryClient? telemetryClient = null)
```
seantleonard, line 57: if log events emmitted using ILogger are pushed to app insights, why do we also need `TelemetryClient`? Couldn't we just add additional `ILogger.Info` events with the desired custom properties?

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.ApplicationInsights;
+using Microsoft.Extensions.Logging;
+
+/// <summary>
+/// Provides a logger implementation that sends telemetry to Azure Application Insights.
```
seantleonard, line 10: Is there a sample/Microsoft doc used to describe implementing a custom logger provider that you can link in this comment?

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return true;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+        {
+            // Whenever ILogger.Log is called for Error, we send an event to Application Insights.
+            if (logLevel == LogLevel.Error)
```
seantleonard, line 58: logLevel is a good candidate for allowing users to customize which, if any, logLevel events are sent to app insights.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                _telemetryClient.TrackEvent("ErrorCaught", new Dictionary<string, string>
+                {
+                    { "CategoryName", _categoryName },
+                    { "Message", formatter(state, exception) },
```
seantleonard, line 63: When we use structured logging (add properties into the log message), do we need to do anything here to ensure the properties are tracked/uploaded?: 
```csharp
ILogger.Error(message: "This is a message with {propertyName}", propertyNameVariable);
```

src/Service/Telemetry/MyTelemetryInitializer.cs
```diff
@@ -0,0 +1,28 @@
...
+        // Add more custom properties here
+    };
+
+    public void Initialize(ITelemetry telemetry)
+    {
+        telemetry.Context.Cloud.RoleName = "DataApiBuilder";
+        telemetry.Context.Session.Id = Guid.NewGuid().ToString();
+        telemetry.Context.Component.Version = ProductInfo.GetProductVersion();
+        telemetry.Context.Device.OperatingSystem = Environment.OSVersion.ToString();
+        telemetry.Context.User.Id = $"{Environment.MachineName}_{Environment.UserName}";
```
seantleonard, line 21: This assumes the user has these environment variables, right? This should be documented or exposed as options that the user can configure.

src/Service/Telemetry/MyTelemetryInitializer.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static readonly IReadOnlyDictionary<string, string> GlobalProperties = new Dictionary<string, string>
+    {
+        { "UserAgent", $"{ProductInfo.GetDataApiBuilderUserAgent()}" },
+        // Add more custom properties here
+    };
+
+    public void Initialize(ITelemetry telemetry)
+    {
+        telemetry.Context.Cloud.RoleName = "DataApiBuilder";
```
seantleonard, line 17: Since we hardcode the role name, do we need to recommend that customers have one appinsights instance per Data API builder deployment?

src/Service/Telemetry/MyTelemetryInitializer.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Collections.Generic;
+using Azure.DataApiBuilder.Product;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility;
+
+public class MyTelemetryInitializer : ITelemetryInitializer
```
seantleonard, line 7: Probably would need a more concrete name like `AppInsightsTelemetryInitializer` or `DabTelemetryInitializer`, since using `My` indicates that this is a demo or sample code.

src/Service/Startup.cs
```diff
@@ -536,6 +546,41 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+        {
+            if (runtimeConfigurationProvider.TryGetConfig(out RuntimeConfig? runtimeConfig) && runtimeConfig.Runtime.Telemetry is not null
+                && runtimeConfig.Runtime.Telemetry.ApplicationInsights.Enabled)
+            {
+                _applicationInsightsOptions = runtimeConfig.Runtime.Telemetry.ApplicationInsights;
+
+                if (string.IsNullOrWhiteSpace(_applicationInsightsOptions.ConnectionString))
+                {
+                    _logger.LogError("Logs won't be sent to Application Insights as connection string is not set in the runtime config.");
+                }
```
seantleonard, line 565: Should this method fail early / return here? if the connectionstring is missing, shouldn't we avoid instantiating TelemetryConfiguration?

src/Service/Program.cs
```diff
@@ -119,6 +119,18 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel)
...
+
+                    // For Sending all the ILogger logs to Application Insights
+                    if (Startup._applicationInsightsOptions.Enabled)
+                    {
+                        builder.AddApplicationInsights(configureTelemetryConfiguration: (config) =>
+                            config.ConnectionString = Startup._applicationInsightsOptions.ConnectionString,
+                            configureApplicationInsightsLoggerOptions: (options) => { }
+                        )
+                        .AddFilter<ApplicationInsightsLoggerProvider>("", logLevel)
+                        .AddProvider(new ApplicationInsightsLoggerProvider(Startup._telemetryClient!));
```
seantleonard, line 131: Would ILogger be a valid provider here for appinsights?

seantleonard (changes requested):
Some questions/suggestions prior to continuing further review

src/Core/Services/ResolverMiddleware.cs
abhishekkumams: makes sense, will move this to a seperate PR.

ayush3797: @seantleonard how do we, in future see allowing user to customize events that are logged? Via config file?

src/Core/Services/ResolverMiddleware.cs
```diff
@@ -371,5 +387,19 @@ private static void SetNewMetadata(IMiddlewareContext context, IMetadata? metada
         {
             context.ScopedContextData = context.ScopedContextData.SetItem(_contextMetadata, metadata);
         }
+
+        /// <summary>
+        /// Track event when GraphQL request is received and send EntityName, Operation and Description for Telemetry.
+        /// </summary>
+        private static void TrackEventForGraphQLRequests(IMiddlewareContext context, TelemetryClient _telemetryClient)
+        {
+            _telemetryClient.TrackEvent("GraphQLRequestReceived", new Dictionary<string, string>
```
ayush3797, line 396: Probably have constants for all the event names in that PR (I saw the comment from Sean to have these events in a separate PR)

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2409,6 +2409,17 @@ private static async Task<HttpStatusCode> GetGraphQLResponsePostConfigHydration(
             { entityName, entity }
         };
 
+            // Adding an entity with only Authorized Access
+            Entity anotherEntity = new(
```
ayush3797, line 2413: Why is this required?

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,257 @@
...
+    /// <summary>
+    /// Sets up the test environment by creating a runtime config with telemetry options.
+    /// </summary>
+    [ClassInitialize]
+    public static void SetUpTelemetryInconfig(TestContext testContext)
+    {
+        DataSource dataSource = new(DatabaseType.MSSQL,
+            GetConnectionStringFromEnvironmentConfig(environment: TestCategory.MSSQL), Options: null);
+
+        _configuration = InitMinimalRuntimeConfig(dataSource, new(), new());
```
ayush3797, line 46: nit: helpful to have named parameters.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,257 @@
...
+        Assert.IsTrue(telemetryItems.Any(item =>
+            item is EventTelemetry
+            && ((EventTelemetry)item).Name == "GraphQLRequestReceived"
+            && ((EventTelemetry)item).Properties["GraphQLOperation"] == "Query"
+            && ((EventTelemetry)item).Properties["GraphQLEntityOperationName"] == "book_by_pk"
+            && ((EventTelemetry)item).Properties["GraphQLRequestMethod"] == "POST"));
+
+        Assert.IsTrue(telemetryItems.Any(item =>
+            item is EventTelemetry
+            && ((EventTelemetry)item).Name == "RestRequestReceived"
```
ayush3797, line 116: Can use string.equals instead for comparing string literals.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,257 @@
...
+        using (HttpClient client = server.CreateClient())
+        {
+            JsonContent content = GetPostStartupConfigParams(TestCategory.MSSQL, _configuration, configurationEndpoint);
+
+            HttpResponseMessage postResult =
+            await client.PostAsync(configurationEndpoint, content);
+            Assert.AreEqual(HttpStatusCode.OK, postResult.StatusCode);
+
+            HttpStatusCode restResponseCode = await GetRestResponsePostConfigHydration(client);
+
```
ayush3797, line 153: nit: blank lines not really required here.

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: to generate a failure event while making an anonymous request to an Entity with authorized only access.

src/Service/Startup.cs
```diff
@@ -46,6 +48,9 @@ public class Startup
 
         public static bool IsLogLevelOverriddenByCli;
 
+        public static ApplicationInsightsOptions _applicationInsightsOptions = new();
+        public static TelemetryClient? _telemetryClient;
```
ayush3797, line 52: Naming convention is not right for members. Should start with Capital letter.

src/Config/ObjectModel/TelemetryOptions.cs
abhishekkumams: Correct, This will be updated based on future requirements. since, currently we don't know what other telemetry Options we would provide support for and how connection to them will be estabilished.

src/Core/Services/ResolverMiddleware.cs
abhishekkumams: Yeah, in cli we will add a command to add telemetry, there we can give them a switch to enable some of the predefined event tracking feature embedded in DAB, for more customization. They can create an issue or create their own custom events inside the code.

src/Service/Controllers/RestController.cs
abhishekkumams: telemetry client is required for tracking event, ILogger would just send logs. For creating an event in ApplicationInsights, we need to use `TrackEvent` method provided by `Microsoft.ApplicationInsights` package.

src/Core/Services/ResolverMiddleware.cs
ayush3797: I see, but providing ability to them to create custom event means adding logging to different places in the code, right? I know that's looking too far ahead, but won't that be a bit tricky? Are such abilities provided by any other APIs that you know of?

src/Service/Program.cs
abhishekkumams: ApplicationInsightsLoggerProvider is based on top of ILoggerProvider itself, we need to customize it for capturing errors and creating event for Error case.

src/Service/Startup.cs
abhishekkumams: right, thanks for the catch.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
abhishekkumams: Sure, I used this one: https://learn.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider

src/Core/Services/ResolverMiddleware.cs
abhishekkumams: Not that I'm aware of, trackEvent is mostly to capture different features within the code or to monitor performance of specific section of code, such as a graphQL request or database operation, etc. For this PR, i won't be keeping this custom trackEvents and will check if there are better ways to do it.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
seantleonard: I meant add to the code comment for future reference.

src/Core/Services/ResolverMiddleware.cs
seantleonard: I see Ayush's point. My perspective is that we should be careful with what logging level we assign to various logging events. If we add a lot of error events to a hot code path, then that greatly increases the number of logs customers have to store. 
In other cloud services, I have seen Microsoft recommend that customers do not have "Trace/verbose/informational" logs in production, but instead experiment with enabling critical errors.

seantleonard: Customers are sensitive (as they should be) to increased logging added unnecessarily.. leading to rising costson their part without notice/documentation/guidance from microsoft.

src/Service/Telemetry/MyTelemetryInitializer.cs
abhishekkumams: updated it to not use hardcodded value, rather use dab_oss_version/dab_hosted_version accordingly based on oss/hosted scenario resp.

src/Core/Services/ResolverMiddleware.cs
abhishekkumams: > I see Ayush's point. My perspective is that we should be careful with what logging level we assign to various logging events. If we add a lot of error events to a hot code path, then that greatly increases the number of logs customers have to store. In other cloud services, I have seen Microsoft recommend that customers do not have "Trace/verbose/informational" logs in production, but instead experiment with enabling critical errors.

Currently, DAB follows the same approach, in Production environment, we don't generate console logs below error level and same for application insights we don't send logs below error level.

Also, A cutom event for tracking errors will only be sent for errors.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
abhishekkumams: I updated it so that the event "ErrorCaught", will have all the properties from Structured Logging

src/Service/Program.cs
```diff
@@ -119,6 +119,18 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel)
...
                     builder.AddFilter(category: "Default", logLevel);
+
+                    // For Sending all the ILogger logs to Application Insights
+                    if (Startup.AppInsightsOptions.Enabled)
+                    {
+                        builder.AddApplicationInsights(configureTelemetryConfiguration: (config) =>
+                            config.ConnectionString = Startup.AppInsightsOptions.ConnectionString,
+                            configureApplicationInsightsLoggerOptions: (options) => { }
+                        )
+                        .AddFilter<ApplicationInsightsLoggerProvider>("", logLevel)
```
seantleonard, line 130:
```suggestion
                        .AddFilter<ApplicationInsightsLoggerProvider>(category: string.Empty, logLevel)
```

src/Service/Startup.cs
```diff
@@ -534,6 +544,42 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+                AppTelemetryClient = app.ApplicationServices.GetRequiredService<TelemetryClient>();
+
+                // Update the TelemetryConfiguration object
+                TelemetryConfiguration telemetryConfiguration = AppTelemetryClient.TelemetryConfiguration;
+                telemetryConfiguration.ConnectionString = AppInsightsOptions.ConnectionString;
+
+                if (AppTelemetryClient is null)
+                {
+                    _logger.LogError("Telemetry client is not initialized.");
+                }
```
seantleonard, line 575: Shouldn't this null check be two lines up? you have the following code preceding this check which uses `AppTelemetryClient` already.
```csharp
TelemetryConfiguration telemetryConfiguration = AppTelemetryClient.TelemetryConfiguration;
```

src/Service/Startup.cs
```diff
@@ -534,6 +544,42 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+
+                // Updating Startup Logger to Log from Startup Class.
+                ILoggerFactory? loggerFactory = Program.GetLoggerFactoryForLogLevel(MinimumLogLevel);
+                _logger = loggerFactory.CreateLogger<Startup>();
```
seantleonard, line 579: I don't understand why this is needed. You're already using `_logger` prior to this. `_logger` is populated in the Startup class constructor with an instance of `ILogger<Startup> logger`

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+        {
+            // Whenever ILogger.Log is called for Error, we send an event to Application Insights.
+            if (logLevel == LogLevel.Error)
+            {
+                // The state parameter is an object that contains additional information about the log event.
+                // The code checks if state is an IEnumerable<KeyValuePair<string, object>> and uses the ToDictionary method to convert it to a dictionary of key-value pairs.
+                // If state is not an IEnumerable<KeyValuePair<string, object>>, the code creates an empty dictionary.
+                IDictionary<string, string> properties = (state as IEnumerable<KeyValuePair<string, object>>)?.ToDictionary(kv => kv.Key, kv => kv.Value?.ToString() ?? string.Empty) ?? new Dictionary<string, string>();
```
seantleonard, line 67: can break this line up for formatting. Really long line.

src/Service/Startup.cs
```diff
@@ -534,6 +544,42 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+            {
+                AppInsightsOptions = runtimeConfig.Runtime.Telemetry.ApplicationInsights;
+
+                if (string.IsNullOrWhiteSpace(AppInsightsOptions.ConnectionString))
+                {
+                    _logger.LogError("Logs won't be sent to Application Insights as connection string is not set in the runtime config.");
+                    return;
+                }
+
+                AppTelemetryClient = app.ApplicationServices.GetRequiredService<TelemetryClient>();
```
seantleonard, line 566: This doesn't make sense. Having a static AppTelemetryClient AND registering it as a singleton?

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ApplicationInsights;
+using Microsoft.Extensions.Logging;
+
+/// <summary>
+/// Provides a custom logger implementation that sends telemetry events to Azure Application Insights.
+/// See <see href="https://learn.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider">Custom Logging Provider</see> for more information.
+/// </summary>
+public class ApplicationInsightsLoggerProvider : ILoggerProvider
```
seantleonard, line 14: Why are you creating your own?  ApplicationInsightsLoggerProvider and ApplicationInsightsLogger are provided by the nuget package. See source: https://github.com/microsoft/ApplicationInsights-dotnet/tree/main/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation

src/Service/Startup.cs
```diff
@@ -70,6 +75,11 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton(fileSystem);
             services.AddSingleton(configProvider);
             services.AddSingleton(configLoader);
+
+            // Register a TelemetryClient instance and custom ITelemetryInitializer implementation with the dependency injection
+            services.AddSingleton<TelemetryClient>();
```
seantleonard, line 80: Why not use `services.AddApplicationInsightsTelemetry()`? That automatically registers `telemetryclient` for you and no need for static instance.

src/Service/Startup.cs
abhishekkumams: the logger doesn't have telemetryClient info, so the logs from StartUp class won't go to the App Insights, that's why updating it.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
abhishekkumams: They're obsolete, also we wanted to customize it for our use case, therefore not using those.

src/Service/Startup.cs
abhishekkumams: removed the static variable

src/Service/Startup.cs
abhishekkumams: Thanks, updated it accordingly.

abhishekkumams: for static field, i created it to use telemetryClient in Program.cs when using constructing Provider `ApplicationInsightsLoggerProvider` for ILogger

src/Service/Startup.cs
```diff
@@ -46,6 +48,9 @@ public class Startup
 
         public static bool IsLogLevelOverriddenByCli;
 
+        public static ApplicationInsightsOptions AppInsightsOptions = new();
+        // public static TelemetryClient? AppTelemetryClient;
```
seantleonard, line 52:
```suggestion
```
nit commented code

src/Directory.Packages.props
```diff
@@ -13,16 +13,20 @@
     <PackageVersion Include="Humanizer" Version="2.14.1" />
     <PackageVersion Include="Humanizer.Core" Version="2.14.1" />
     <PackageVersion Include="DotNetEnv" Version="2.5.0" />
+    <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.21.0" />
+    <PackageVersion Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.15.0" />
```
seantleonard, line 17: 2.15.0 is a deprecated version of `Microsoft.ApplicationInsights.AspNetCore` and must be updated before merging.

src/Service/Program.cs
seantleonard: still need named param

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
seantleonard: I did more digging and they do not appear to be obsolete. In fact, `AddApplicationInsights()` uses the library provided `ApplicationInsightsLoggerProvider`. Source code: https://github.com/microsoft/ApplicationInsights-dotnet/blob/2.21.0/LOGGING/src/ILogger/ApplicationInsightsLoggerProvider.cs

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    public void Dispose()
+    {
+        // Nothing to dispose
+    }
+
+    /// <summary>
+    /// This logger implementation is used to track errors caught by the application.
+    /// </summary>
+    private class ApplicationInsightsLogger : ILogger
```
seantleonard, line 38: There is a lot more built in functionality that the library's included `ApplicationInsightsLogger` provides https://github.com/microsoft/ApplicationInsights-dotnet/blob/2.21.0/LOGGING/src/ILogger/ApplicationInsightsLogger.cs

Based on how you configure AppInsights (`applicationInsightsLoggerOptions.TrackExceptionsAsExceptionTelemetry` == true/false) app insights knows to `TrackTrace()` vs `TrackException()`

After looking through app insights docs, I don't think a custom logger to execute `TrackEvent()` is the right way forward at this time. Right now we just want to upload logs emitted by DAB. 

Per API summary, TrackEvent is used for "User actions and other events. Used to track user behavior or to monitor performance." I think we need to rethink what we consider as events and how we should use them. Then, have a follow up PR that adds additional features. Ideally, we want to reduce rolling our own as much as possible. 
https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#api-summary

seantleonard: https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#trackevent

src/Service/Program.cs
```diff
@@ -119,6 +121,18 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel)
...
+
+                    // For Sending all the ILogger logs to Application Insights
+                    if (Startup.AppInsightsOptions.Enabled && appTelemetryClient is not null)
+                    {
+                        builder.AddApplicationInsights(configureTelemetryConfiguration: (config) =>
+                            config.ConnectionString = Startup.AppInsightsOptions.ConnectionString,
+                            configureApplicationInsightsLoggerOptions: (options) => { }
+                        )
+                        .AddFilter<ApplicationInsightsLoggerProvider>("", logLevel)
+                        .AddProvider(new ApplicationInsightsLoggerProvider(appTelemetryClient));
```
seantleonard, line 133: Won't need this (see comments in app insights logger provider file), because `.AddApplicationInsights()` already does this.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
abhishekkumams: removed created `ApplicationInsightsLoggerProvider` Class.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
abhishekkumams: makes sense, removed the class.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+        server.Dispose();
+
+        Console.WriteLine(JsonSerializer.Serialize(telemetryItems));
+
+        // Assert that we are not sending any Traces/Requests/Exceptions to Telemetry
+        Assert.IsTrue(telemetryItems.IsNullOrEmpty());
+    }
+
+    /// <summary>
+    /// this method tests the ability of the server to handle GraphQL and REST requests,
```
Aniruddh25, line 231:
```suggestion
    /// This method tests the ability of the server to handle GraphQL and REST requests,
```

Aniruddh25 (changes requested):
Looks good overall, need some clarification on hosted scenarios.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            HttpResponseMessage graphQLResponse = await client.SendAsync(graphQLRequest);
+            Assert.AreEqual(HttpStatusCode.OK, graphQLResponse.StatusCode);
+
+            // POST request on non-accessible entity
+            HttpRequestMessage restRequest = new(HttpMethod.Post, "/api/Publisher/id/1?name=Test");
+            HttpResponseMessage restResponse = await client.SendAsync(restRequest);
+
+            Assert.AreEqual(HttpStatusCode.Forbidden, restResponse.StatusCode);
+
```
seantleonard, line 262:
```suggestion
```

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+        Assert.IsTrue(telemetryItems.Any(item =>
+        item is ExceptionTelemetry
+        && ((ExceptionTelemetry)item).Message.Equals("Authorization Failure: Access Not Allowed.")));
+    }
+
+    /// <summary>
+    /// Tests that No telemetry data is sent when disabled.
+    /// Both Application Insight Flag should be enabled and Telemetry Connection String should be provided
+    /// to send telemetry.
```
seantleonard, line 200:
```suggestion
    /// Validates that no telemetry data is sent to CustomTelemetryChannel when 
    /// Appsights is disabled OR when no valid connectionstring is provided.
```

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+    {
+        // Disable parallel execution for this test method
+        TestContext.Properties.Add("ParallelScope", "Individual");
+
+        SetUpTelemetryInconfig(CONFIG_WITH_TELEMETRY, isTelemetryEnabled: true, TEST_APP_INSIGHTS_CONN_STRING);
+
+        string[] args = new[]
+        {
+            $"--ConfigFileName={CONFIG_WITH_TELEMETRY}"
+        };
```
seantleonard, line 136: The `args` array doesn't seem to be needed and doesn't look to be used in this test.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+            HttpStatusCode restResponseCode = await GetRestResponsePostConfigHydration(client, "Publisher/id/1");
+
+            Assert.AreEqual(expected: HttpStatusCode.Forbidden, actual: restResponseCode);
+
+            HttpStatusCode graphqlResponseCode = await GetGraphQLResponsePostConfigHydration(client);
+
+            Assert.AreEqual(expected: HttpStatusCode.OK, actual: graphqlResponseCode);
+        }
+
+        server.Dispose();
```
seantleonard, line 160: nit: you can use a `using` statement with TestServer setup, this is done in other tests in ConfigurationTests.cs, that way you don't need to manually dispose the server object.
```csharp
        using (TestServer server = new(Program.CreateWebHostFromInMemoryUpdateableConfBuilder(Array.Empty<string>())))
        using (HttpClient client = server.CreateClient())
        {
            ...
        }
```

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        Assert.IsTrue(telemetryItems.Any(item =>
+        item is RequestTelemetry
+        && ((RequestTelemetry)item).Name.Equals("GET Rest/Find [route]")
+            && ((RequestTelemetry)item).ResponseCode.Equals("403")
+            && ((RequestTelemetry)item).Url.PathAndQuery.Equals("/api/Publisher/id/1")));
+
+        Assert.IsTrue(telemetryItems.Any(item =>
+        item is ExceptionTelemetry
+        && ((ExceptionTelemetry)item).Message.Equals("Authorization Failure: Access Not Allowed.")));
```
seantleonard, line 194: should be indented to remain consistent with first batch of Asserts.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+    /// <summary>
+    /// Testing the Hosted Scenario for both configuration endpoint.
+    /// Tests that different telemetry items such as Traces or logs, Exceptions and Requests
+    /// are correctly sent to application Insights when enabled.
+    /// Also asserting on their respective properties.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow(CONFIGURATION_ENDPOINT)]
+    [DataRow(CONFIGURATION_ENDPOINT_V2)]
+    public async Task TestTelemetryItemsAreSentCorrectly_HostedScenario(string configurationEndpoint)
```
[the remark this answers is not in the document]
seantleonard: " if we can do this for hosted case." if we can use AppInsights within SWA Database connection instance of DAB?

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+        };
+
+        List<ITelemetry> telemetryItems = new();
+        ITelemetryChannel telemetryChannel = new CustomTelemetryChannel(telemetryItems);
+        Startup.CustomTelemetryChannel = telemetryChannel;
+        TestServer server = new(Program.CreateWebHostBuilder(args));
+
+        await TestRestAndGraphQLRequestsOnServerInNonHostedScenario(server);
+
+        server.Dispose();
```
seantleonard, line 86: wrap in using statement to avoid needing to manually dispose the server object.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,295 @@
...
+namespace Azure.DataApiBuilder.Service.Tests.Configuration;
+
+/// <summary>
+/// Contains tests for telemetry functionality.
+/// </summary>
+[TestClass, TestCategory(TestCategory.MSSQL)]
+public class TelemetryTests
+{
+    public TestContext TestContext { get; set; }
+    private const string TEST_APP_INSIGHTS_CONN_STRING = "InstrumentationKey=testKey;IngestionEndpoint=https://unitTest.com/;LiveEndpoint=https://unittest2.com/";
```
seantleonard, line 30: nit: unittest.com resolves to a domain. Any way to just specify local host for these endpoints?
also for the InstrumentationKey, isn't usage of that deprecated? is it even needed here?

src/Service/Startup.cs
```diff
@@ -535,6 +556,49 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+        /// <summary>
+        /// Configure Application Insights Telemetry based on the loaded runtime configuration. If Application Insights
+        /// is enabled, we can track different events and metrics.
+        /// </summary>
+        /// <param name="runtimeConfigurationProvider">The provider used to load runtime configuration.</param>
+        /// <seealso cref="https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core#enable-application-insights-telemetry-collection"/>
+        private void ConfigureApplicationInsightsTelemetry(IApplicationBuilder app, RuntimeConfigProvider runtimeConfigurationProvider)
+        {
+            if (runtimeConfigurationProvider.TryGetConfig(out RuntimeConfig? runtimeConfig) && runtimeConfig.Runtime.Telemetry is not null
+                && runtimeConfig.Runtime.Telemetry.ApplicationInsights.Enabled)
```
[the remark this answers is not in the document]
abhishekkumams: yes, it won't be null. the default value initializes it.

src/Service.Tests/Configuration/TelemetryTests.cs
abhishekkumams: connection string is a combination of all those key/value. direct usage of instrumentation key is deprecated.

abhishekkumams: done,thanks

src/Service/Telemetry/AppInsightsTelemetryInitializer.cs
```diff
@@ -0,0 +1,31 @@
...
+        // Add more custom properties here
+    };
+
+    /// <summary>
+    /// Initializes the telemetry context.
+    /// </summary>
+    /// <param name="telemetry">The telemetry object to</param>
+    public void Initialize(ITelemetry telemetry)
+    {
+        telemetry.Context.Cloud.RoleName = ProductInfo.GetDataApiBuilderUserAgent();
```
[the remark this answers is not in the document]
abhishekkumams: image

I did a search on what people are using, and I couldn't find anything most suitable for role name as of now. open to suggestions. Could be something configurable by user later as well.

src/Service/Startup.cs
```diff
@@ -535,6 +556,49 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+            {
+                AppInsightsOptions = runtimeConfig.Runtime.Telemetry.ApplicationInsights;
+
+                if (string.IsNullOrWhiteSpace(AppInsightsOptions.ConnectionString))
+                {
+                    _logger.LogError("Logs won't be sent to Application Insights as connection string is not set in the runtime config.");
```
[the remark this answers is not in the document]
abhishekkumams: According to the RFC: https://github.com/Azure/data-api-builder/issues/1601, It was mentioned to raise an error because of configuration mismatch.

src/Service/Startup.cs
```diff
@@ -535,6 +556,49 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+                    return;
+                }
+
+                TelemetryClient? appTelemetryClient = app.ApplicationServices.GetService<TelemetryClient>();
+
+                if (appTelemetryClient is null)
+                {
+                    _logger.LogError("Telemetry client is not initialized.");
```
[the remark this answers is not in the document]
abhishekkumams: Acc to RFC, Since user enabled application insights, this should be properly initialized to send the telemetry, otherwise error shown.

src/Config/ObjectModel/TelemetryOptions.cs
```diff
@@ -0,0 +1,10 @@
...
+{ }
```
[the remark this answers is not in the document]
abhishekkumams: Yes If Telemetry option is provided, by default it is configured with default values of ApplicationInsightsOptions, which is disabled and empty connection string.

src/Service/Startup.cs
```diff
@@ -557,6 +621,10 @@ private async Task<bool> PerformOnConfigChangeAsync(IApplicationBuilder app)
             {
                 RuntimeConfigProvider runtimeConfigProvider = app.ApplicationServices.GetService<RuntimeConfigProvider>()!;
                 RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
+
+                // Configure Application Insights Telemetry
```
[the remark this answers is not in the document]
abhishekkumams: yes, otherwise we won't see telemetry in hosted case

src/Service.Tests/Configuration/TelemetryTests.cs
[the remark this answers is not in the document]
abhishekkumams: I didn't get. Do we not want to use application insights for hosted case?

src/Service/Telemetry/AppInsightsTelemetryInitializer.cs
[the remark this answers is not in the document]
abhishekkumams: sure

src/Config/ObjectModel/TelemetryOptions.cs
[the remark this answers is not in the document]
abhishekkumams: making it nullable

src/Service/Startup.cs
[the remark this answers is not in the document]
abhishekkumams: this should be error, if Application Insights is enabled and connection string is also provided but telemetry client in not initialized. this can lead to loss of telemetry, hence error.

src/Service.Tests/Configuration/TelemetryTests.cs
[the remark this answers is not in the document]
abhishekkumams: https://github.com/Azure/data-api-builder/issues/1735 For Adding support of AppInsights in Hosted Scenario.

src/Product/ProductInfo.cs
```diff
@@ -12,6 +12,8 @@ public static class ProductInfo
     public const string DAB_APP_NAME_ENV = "DAB_APP_NAME_ENV";
     public static readonly string DEFAULT_APP_NAME = $"dab_oss_{ProductInfo.GetProductVersion()}";
```
seantleonard, line 14: nit: need blank line?

src/Service/Startup.cs
```diff
@@ -535,6 +559,55 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+
+                if (!AppInsightsOptions.Enabled)
+                {
+                    _logger.LogInformation("Application Insights are disabled.");
+                    return;
+                }
+
+                if (string.IsNullOrWhiteSpace(AppInsightsOptions.ConnectionString))
+                {
+                    _logger.LogWarning("Logs won't be sent to Application Insights as connection string is not available in the runtime config.");
```
seantleonard, line 583:
```suggestion
                    _logger.LogWarning("Logs won't be sent to Application Insights because an Application Insights connection string is not available in the runtime config.");
```

src/Config/ObjectModel/TelemetryOptions.cs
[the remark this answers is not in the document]
Aniruddh25: For the record, this is a shorthand for defining the property itself/constructor.
Did you mean to set the property nullable? or want to change it only for the constructor? 

If the latter, you need to define a separate constructor. See my PR for initializing optional properties for `RuntimeOptions` for an e.g.

src/Service/Startup.cs
Aniruddh25: From offline discussion - Already emitted logs will not be sent to AppInsights after this update.

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
```diff
@@ -1,85 +0,0 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
-
-using System;
```
Aniruddh25: Why dont we need the ApplicationInsightsLoggerProvider anymore?

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        // Assert that we are not sending any Traces/Requests/Exceptions to Telemetry
+        Assert.IsTrue(telemetryItems.IsNullOrEmpty());
+    }
+
+    /// <summary>
+    /// This method tests the ability of the server to handle GraphQL and REST requests,
+    /// and verifies that the server returns the expected response status codes.
+    /// Makes a valid GraphQL Request and one Invalid REST Request.
+    /// </summary>
```
seantleonard, line 173: Can you elaborate on how this relates to the Telemetry tests? ie why successful execution validates telemetry config

src/Service/Telemetry/ApplicationInsightsLoggerProvider.cs
seantleonard: `AddApplicationInsights()` uses the AppInsights SDK provided `ApplicationInsightsLoggerProvider`. Source code: https://github.com/microsoft/ApplicationInsights-dotnet/blob/2.21.0/LOGGING/src/ILogger/ApplicationInsightsLoggerProvider.cs

No need to roll our own.

src/Service.Tests/Configuration/TelemetryTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            $"--ConfigFileName={CONFIG_WITH_TELEMETRY}"
+        };
+
+        ITelemetryChannel telemetryChannel = new CustomTelemetryChannel()
+        {
+            EndpointAddress = "https://localhost/"
+        };
+        Startup.CustomTelemetryChannel = telemetryChannel;
+        using (TestServer server = new(Program.CreateWebHostBuilder(args)))
```
Aniruddh25, line 90: Is the telemetry client being disposed off correctly - so that the Flush gets called?

src/Service.Tests/Configuration/TelemetryTests.cs
abhishekkumams: this method is just used as helper for other test methods to execute REST and GRaphQL requests

src/Config/ObjectModel/TelemetryOptions.cs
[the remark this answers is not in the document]
Aniruddh25: nvm, I realized if we dont keep the optional property itself nullable, I will modify my PR - it might be an issue with the merge configuration file feature as well: [Using two config files](https://learn.microsoft.com/en-us/azure/data-api-builder/data-api-builder-cli#using-data-api-builder-with-two-configuration-files)
If the overriding config has minimal values, it will now have the optional values set to their defaults as opposed to being null. These defaults may not be the same as the base config and this could lead to a situation where the merged configuration overrides the base values just because they were written to the file system by us in the overriding config even when the developer might not have provided those initially. 
 
I think the best way to handle this is to re-introduce optional properties to be nullable, and leave them as is – don’t write the expanded version on the file system - so we don’t accidentally merge and create an incorrect merged configuration.

Aniruddh25 (approved):
LGTM, Thanks for enabling AppInsights!

src/Service.Tests/Configuration/TelemetryTests.cs
abhishekkumams: The TelemetryClient class does not implement the IDisposable interface, which means that it does not need to be explicitly disposed of.

But if you mean the server, i think the using statement will make sure it's disposed off.

src/Service.Tests/Configuration/TelemetryTests.cs
seantleonard: In that case, the description and method name should be updated to reflect that. Right now the description and name read as if this were its test.

I don't think the asserts need to be kept either. Essentially the point of this method is to issue requests which trigger the logging system to emit logs.

abhishekkumams: Gotcha

seantleonard (approved):
thank you for making changes and building this feature!

PR #1653, opened 2023-08-24
OpenAPI document resolves custom configured REST path

### LINKED ISSUE #1633: [Bug]: Wrong REST path generated in OpenAPI spec
### What happened?

When adding the entity via DAB CLI, the generated path for REST contains an unescaped slash. This is the DAB CLI command I used:

```
dab add Todos --source dbo.todos --permissions "anonymous:*" --rest todos
```

and this is the configuration I got for the created entity:

```
"Todos": {
      ...
      "rest": {
        "enabled": true,
        "path": "/todos",
        "methods": [
          "get",
          "post",
          "put",
          "patch",
          "delete"
        ]
      },
    ...
    }
```

the path at which REST API are available is:

```
/api/odos
```

instead of the correct:

```
/api/Todos
```

### Version

Microsoft.DataApiBuilder 0.8.44-rc+ebe4ef414deaf7670b579fe160b872c6b682c114

Azure SQL

Local (including CLI)

REST

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
## Why make this change?

- Closes #1633
  - The wrong 'path' value is generated in the OpenAPI description document created. This was observed when looking at the `/swagger` endpoint to see the REST endpoint documentation.
  - When an entity's REST path property is explicitly set AND includes a starting slash `/`, the openapidocumentor does not properly create the PATH used to access the entity in the REST endpoint.

## What is this change?

- Correctly process an entity's REST path configuration. Before this change, the method at fault was improperly calling `Substring()` on the wrong variable. Although the variable was previously `restPath`, the value at the time of calling substring was the top-level entity name which did not have a starting slash `/`. 
- The variable `restPath` is modified depending on whether the entity has explicit rest path config. If no explicit rest path is configured, the top level entity name is returned as the resolved rest path. 
- Now, calling `entityRestSettings.Path.TrimStart('/')` will correctly remove a starting slash IF one exists. This accommodates the following formatting of the explicit rest path:

1. `/customRestPath`
2. `customRestPath`
3. `//customRestPath` -> while this option is not advisable to use in practice, we still support it and this PR is not meant to change that behavior. But for complete test coverage, this case is tested.

## How was this tested?

- [x] Integration Tests

### DIFF
--- a/src/Core/Services/OpenAPI/OpenApiDocumentor.cs
+++ b/src/Core/Services/OpenAPI/OpenApiDocumentor.cs
@@ -677,7 +677,7 @@ namespace Azure.DataApiBuilder.Core.Services
         /// If no override exists, this method returns the passed in entity name.
         /// </summary>
         /// <param name="entityName">Name of the entity.</param>
-        /// <returns>Returns the REST path name for the provided entity.</returns>
+        /// <returns>Returns the REST path name for the provided entity with no starting slash: {entityName} or {entityRestPath}.</returns>
         private string GetEntityRestPath(string entityName)
         {
             string entityRestPath = entityName;
--- a/src/Core/Services/OpenAPI/OpenApiDocumentor.cs
+++ b/src/Core/Services/OpenAPI/OpenApiDocumentor.cs
@@ -673,8 +673,8 @@ namespace Azure.DataApiBuilder.Core.Services
         }
 
         /// <summary>
-        /// Resolves any REST path overrides present for the provided entity in the runtime config.
-        /// If no overrides exist, returns the passed in entity name.
+        /// Attempts to resolve the REST path override set for an entity in the runtime config.
+        /// If no override exists, this method returns the passed in entity name.
         /// </summary>
         /// <param name="entityName">Name of the entity.</param>
         /// <returns>Returns the REST path name for the provided entity.</returns>
@@ -683,10 +683,10 @@ namespace Azure.DataApiBuilder.Core.Services
             string entityRestPath = entityName;
             EntityRestOptions entityRestSettings = _runtimeConfig.Entities[entityName].Rest;
 
-            if (!string.IsNullOrEmpty(entityRestSettings.Path) && entityRestSettings.Path.StartsWith('/'))
+            if (!string.IsNullOrEmpty(entityRestSettings.Path))
             {
-                // Remove slash from start of rest path.
-                entityRestPath = entityRestPath.Substring(1);
+                // Remove slash from start of REST path.
+                entityRestPath = entityRestSettings.Path.TrimStart('/');
             }
 
             return entityRestPath;
--- a/src/Service.Tests/OpenApiDocumentor/OpenApiTestBootstrap.cs
+++ b/src/Service.Tests/OpenApiDocumentor/OpenApiTestBootstrap.cs
@@ -1,0 +1,94 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Abstractions;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
+
+namespace Azure.DataApiBuilder.Service.Tests.OpenApiIntegration
+{
+    // Defines helpers used to help generate an OpenApiDocument object which
+    // can be validated in tests so that a constantly running DAB instance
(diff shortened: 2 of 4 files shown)


src/Service.Tests/OpenApiDocumentor/StoredProcedureGeneration.cs
seantleonard, line 1: Portions of this file were deleted and moved to a common helper class `OpenApiTestBootstrap`

src/Service.Tests/OpenApiDocumentor/OpenApiTestBootstrap.cs
seantleonard, line 1: This class is for the most part not new code. Code was moved from `StoredProcedureGeneration.cs` to this class as a common helper. 

The change i made was adding to the method signature for creating the test server `GenerateOpenApiDocument(...)` so it takes arguments from the various tests to create a dab server and generate an openapi doc.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -683,10 +683,10 @@ private string GetEntityRestPath(string entityName)
             string entityRestPath = entityName;
             EntityRestOptions entityRestSettings = _runtimeConfig.Entities[entityName].Rest;
 
-            if (!string.IsNullOrEmpty(entityRestSettings.Path) && entityRestSettings.Path.StartsWith('/'))
```
seantleonard, line 686: by removing `entityRestSettings.Path.StartsWith('/')` from the conditional, i avoid improperly setting the PATH due to the edge case of when a user does not define their custom rest path with a starting slash `/`

src/Service.Tests/OpenApiDocumentor/OpenApiTestBootstrap.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// where all actions are permitted.
+        /// </summary>
+        /// <returns>Array of EntityPermission objects.</returns>
+        internal static EntityPermission[] CreateBasicPermissions()
+        {
+            List<EntityPermission> permissions = new()
+            {
+                new EntityPermission("anonymous", new EntityAction[]
+                {
+                    new(EntityActionOperation.All, null, new())
```
Aniruddh25, line 83: nit: name arguments: null, new() for clarity

src/Service.Tests/OpenApiDocumentor/PathValidation.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// Validates that the OpenApiDocument object's Paths property for an entity is generated
+        /// with the entity's explicitly configured REST path, if set. Otherwise, the top level
+        /// entity name is used.
+        /// </summary>
+        /// <param name="entityName">Top level entity name defined in runtime config.</param>
+        /// <param name="restPath">Entity's configured REST path.</param>
+        /// <param name="expectedOpenApiPath">Expected path generated for OpenApiDocument.Paths</param>
+        [DataRow("entity", "/customEntityPath", "/customEntityPath", DisplayName = "Entity REST path has leading slash - REST path override used.")]
```
Aniruddh25, line 34: If the leading slash is trimmed, how is the expected path value still = `/customEntity`?

src/Service.Tests/OpenApiDocumentor/PathValidation.cs
```diff
@@ -0,0 +1,70 @@
...
+        [DataRow("entity", "//customEntityPath", "/customEntityPath", DisplayName = "Entity REST path has leading slashes - REST path override used.")]
```
Aniruddh25, line 35: If we trim only one slash, and OpenApiDocumentor somehow still adds a slash before the expected path as per all the other DataRows, shouldn't the expected path here be `//customEntityPath`

Aniruddh25 (approved):
Thanks for the quick investigation and fix! Few questions, but otherwise LGTM.

src/Service.Tests/OpenApiDocumentor/PathValidation.cs
seantleonard: Added clarifying comments. The OpenApi path value created takes the form `"/{entityRestPath}"` where `{entityRestPath}` is stripped of all leading slashes when openapidocumentor evaluates whether to use the entity name or configured rest path as the value of rest path to use.

src/Service.Tests/OpenApiDocumentor/PathValidation.cs
seantleonard: Adding the clarifying comment to this test:
```csharp
        /// When OpenApiDocumentor.BuildPaths() is called, the entityBasePathComponent is created using
        /// the formula "/{entityRestPath}" where {entityRestPath} has no starting slashes and is either
        /// the entity name or the explicitly configured entity REST path
```

aaronburtle (approved):
Looks good to me.

PR #1663, opened 2023-08-28
Fixing indentation in the generated configuration

Aniruddh25:
## Why make this change?

- With the overhaul of the config system using our own Write method while serializing objects, the config that is generated is misaligned with respect to indentation for all the keys under each entity name.
For e.g.:
In the following picture, the entity name `Todo` is indented with 4 leading spaces, subsequently its child key `source` should have been indented with 6 leading spaces, however its currently indented with only 2 spaces.

image

- This is a regression in 0.8.44-rc.

## What is this change?

- Use the same writer to serialize the subkeys under the entity name so that the depth of indentation is maintained.

## How does this change fix the issue?
- `WriteRawValue` writes the json string argument provided as is without doing any additional formatting on it. It doesn't utilize the `_currentDepth` private member of the writer to determine how to indent.

- Prior to this change, we were generating the inner json string first then writing the generated inner json "as is". While generating that json on line 29

https://github.com/Azure/data-api-builder/blob/31b754251832ffebf4c4ce36ffcbead9baff6355/src/Config/Converters/RuntimeEntitiesConverter.cs#L29

Line 29 in [31b7542](https://github.com/Azure/data-api-builder/commit/31b754251832ffebf4c4ce36ffcbead9baff6355)

a new writer is being used by the `JsonSerializer.Serialize()` function, effectively resetting the `_currentDepth`. The `_currentDepth` of the writer in context was not getting honored.

## How was this tested?

- Manual test, building solution and using `dab add` to simulate automatic writing of an entity to the config file. With the change, the indentation is fixed.

image

### DIFF
--- a/src/Config/Converters/RuntimeEntitiesConverter.cs
+++ b/src/Config/Converters/RuntimeEntitiesConverter.cs
@@ -26,9 +26,8 @@ class RuntimeEntitiesConverter : JsonConverter<RuntimeEntities>
         writer.WriteStartObject();
         foreach ((string key, Entity entity) in value.Entities)
         {
-            string json = JsonSerializer.Serialize(entity, options);
             writer.WritePropertyName(key);
-            writer.WriteRawValue(json);
+            JsonSerializer.Serialize(writer, entity, options);
         }
 
         writer.WriteEndObject();


src/Config/Converters/RuntimeEntitiesConverter.cs
```diff
@@ -26,9 +26,8 @@ public override void Write(Utf8JsonWriter writer, RuntimeEntities value, JsonSer
         writer.WriteStartObject();
         foreach ((string key, Entity entity) in value.Entities)
         {
-            string json = JsonSerializer.Serialize(entity, options);
             writer.WritePropertyName(key);
-            writer.WriteRawValue(json);
+            JsonSerializer.Serialize(writer, entity, options);
```
seantleonard, line 30: Can you clarify why this works? Is it because WriteRawValue (according to quick stack overflow search) doesn't know how to preserve any indentation formatting in context of where it is writing json?

Would other places where `WriteRawValue()` have this issue too?

Aniruddh25: Yes, WriteRawValue writes the json string argument provided as is without doing any additional formatting on it. It doesnt utilize the `_currentDepth` private member of the writer to determine how to indent. 

Prior to this change, we were generating the inner json string first then writing the generated inner json "as is". While generating that json on line 29 https://github.com/Azure/data-api-builder/blob/31b754251832ffebf4c4ce36ffcbead9baff6355/src/Config/Converters/RuntimeEntitiesConverter.cs#L29 previously, a new writer is being used by the JsonSerializer.Serialize() function, effectively resetting the `_currentDepth`. The `_currentDepth` of the writer in context was not getting honored.

aaronpowell: Nice find out how to fix this. I remember doing it and going "ugh, formatting is busted, I'll fix that in the future" and, well, clearly didn't 🤣

aaronpowell (approved):
Should we regenerate the config files in the test projects to make sure they are formatted properly (or are they already?)

severussundar (approved):
Thank you for spotting and fixing this!

Issue #1517, opened 2023-05-25
Structured Logging - Refactor logging statements to address CA2254


aaronpowell:
Currently we have a lot of [CA2254](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2254) warnings in the codebase due to how we generate log messages.

Here's an example:

```c#
_logger.LogInformation($"Using merged config file based on environment:{configToBeUsed}.");
```

The issue is that we're not using [log templates](https://learn.microsoft.com/dotnet/core/extensions/logging?tabs=command-line#log-message-template) which enable better structured logging.

Ideally, the code should be:

```c#
_logger.LogInformation("Using merged config file based on environment:{configToBeUsed}.", configToBeUsed);
```

NLog has a good overview on [structured logging](https://github.com/NLog/NLog/wiki/How-to-use-structured-logging).

This issue is to track the codebase-wide refactor.

```[tasklist]
### Tasks
- [ ] https://github.com/Azure/data-api-builder/pull/1594
- [x] remove dupe exception logging
- [x] update service/config project
- [x] update core project
```

aaronpowell:
Branching from https://github.com/Azure/data-api-builder/pull/1402#discussion_r1202762244

seantleonard:
Whenever this is picked up, I would suggest and highly appreciate breaking up this refactor into multiple PRs to make reviewing quick and straightforward.

aaronpowell:
Yep, we should create a feature branch and do PRs into that branch so it can be reviewed piece by piece.

seantleonard:
As a follow up to #1541, which has tests that validates the logger logs the expected messages:
We can have more explicit validation of the logged messages:
1. This requires us to turn our message templates into constants with variables used for structured logging present within curly braces: `string messageTemplateError = "{runtimeConfigFile} does not exist."
2. Then, validate that the expected log message template is logged  (this value is available to us without having to run String.Contains() or String.Split()
3. Optionally validate that the expected variable value is used. in this example, the runtimeConfigFile value is as expected.

You can do this with the following (copied from my comment on `dotnet/runtime/issues/ 67577 #issuecomment-1625920776`):

-----------------------------------------------
> I think unit testing should be done using the interface (`IReadOnlyList<KeyValuePair<string, object>>`). I now think the most valid reason for these methods is for re-implementing your own log methods, not unit testing.

[...]

PR #1664, opened 2023-08-28
Remove duplicate exception logged to console during startup failure.

seantleonard:
## Why make this change?

-Work done as part of structured logging overhaul. #1517 

## What is this change?

- Removes one instance where an exception is logged twice:
1. once automatically by dotnet
2. second when we manually include the exception message/stacktrace in exception handling. [Remove this instance]

-Adds named parameters in ilogger call while making message generation for the console more concise. ILogger logError() and other ilogger calls allow you to include a parameter for the exception. That is logged automatically in ilogger so we don't have to additionally log the exception message in the logger message manually.

## How was this tested?

manual testing:
before:
DupeExceptionLoggingStartup

after:
NoDupeExceptionLoggingStartup

## Sample Request(s)

- put a bad entity rest path (all else okay) to trigger startup failure. i.e. `"path": "     /pathName"`


aaronpowell (approved):

abhishekkumams:
Can you check if it works for cases where exception is thrown outside Program.cs,

image

To reproduce:
1. Start the engine with table1 configured in your db and exposed in config file.
2. now drop that table, and make a query on it.

Aniruddh25 (approved):

Issue #1614, opened 2023-07-28
[Bug]: "--config" option is not honoring the user provided config file


abhishekkumams:
### What happened?

the Option `--config` or `-c` is not honoring the user provided config file.
Currently if DAB_ENVIRONMENT or ASPNETCORE_ENVIRONMENT is setup, the environment file is given more prefrenece.

Idealy, It should directly use the config file if provided by the user

### Version

0.8.44-rc

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1669, opened 2023-08-29
Remove redundant logging of found config file

Aniruddh25:
## Why make this change?
- With the overhaul of config system PR #1402 , we started seeing redundant logs about which config file is being used.

## What is this change?

- Store the environment based config file name found at construction time in the property `ConfigFileName`. 
- Remove Console writeline statements while checking for existence of file since this leads to multiple logs - one from CLI before starting the engine and second from within the engine itself when calling TryPargeConfig
- Add a single log of the loaded file in CLI
- Also catch exception for incorrect connection string parsing

## How was this tested?

Before:

After:
image

### DIFF
--- a/src/Cli/ConfigGenerator.cs
+++ b/src/Cli/ConfigGenerator.cs
@@ -949,7 +949,7 @@ namespace Cli
         /// It will use the config provided by the user, else based on the environment value
         /// it will either merge the config if base config and environmentConfig is present
         /// else it will choose a single config based on precedence (left to right) of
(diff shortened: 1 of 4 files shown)


ayush3797 (approved):

abhishekkumams:
does this PR also address the [issue](https://github.com/Azure/data-api-builder/issues/1614) here ?

Aniruddh25:
> does this PR also address the [issue](https://github.com/Azure/data-api-builder/issues/1614) here ?

yes it does

abhishekkumams (approved):

PR #1676, opened 2023-09-01
Parsing parameters based on ApiType

ayush3797:
## Why make this change?

Per this comment: https://github.com/Azure/data-api-builder/pull/1636#issuecomment-1695050975, need to differently parse `System.DateTime` values for GQL/REST. Specifically:

1. For REST, parse `System.DateTime` values using `System.DateTime` class and populate the `DbType` while passing the parsed value to the database.
2. For GQL, parse `System.DateTime` values using `System.DateTimeOffset` class and don't populate the `DbType` while passing the parsed value to the database so as to avoid mismatch between the parsed value type (`System.DateTimeOffset`) and passed DbType (`DbType.DateTime`/`DbType.DateTime2`/`DbType.Date`/`DbType.SmallDateTime`). (Since `smalldatetime`, `date`, `datetime`,  `datetime2` types in MsSql map to .NET type of `System.DateTime`.)

## What is this change?
1. Based on the ApiType, differently parse System.DateTime values for REST/GQL in `BaseSqlQueryStructure.ParseParamAsSystemType` method.
2. Avoid populate DbType for System.DateTime for GQL in `BaseQueryStructure.MakeDbConnectionParam()` method.

## How was this tested?
Passing of existing test cases confirms the refactor.

### DIFF
--- a/src/Config/DatabasePrimitives/DatabaseObject.cs
+++ b/src/Config/DatabasePrimitives/DatabaseObject.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT License.
 
 using System.Data;
+using System.Net;
 using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Service.Exceptions;
 
 namespace Azure.DataApiBuilder.Config.DatabasePrimitives;
 
@@ -119,6 +121,21 @@ public class StoredProcedureDefinition : SourceDefinition
 
         return null;
     }
+
+    /// <inheritdoc/>
+    public override Type GetSystemTypeForParam(string paramName)
+    {
+        if (Parameters.TryGetValue(paramName, out ParameterDefinition? paramDefinition))
+        {
+            return paramDefinition.SystemType;
+        }
+
+        throw new DataApiBuilderException(
+            message: "Unexpected parameter name encountered",
+            statusCode: HttpStatusCode.InternalServerError,
+            subStatusCode: DataApiBuilderException.SubStatusCodes.UnexpectedError
+            );
+    }
 }
 
 public class ParameterDefinition
@@ -185,6 +202,28 @@ public class SourceDefinition
         }
 
         return null;
+    }
+
+    /// <summary>
+    /// Method to get the system type for:
+    /// 1. column for table/view,
+    /// 2. parameter for stored procedure.
+    /// </summary>
+    /// <param name="paramName">The parameter whose system type is to be determined.
+    /// For table/view paramName refers to the backingColumnName if aliases are used.</param>
+    /// <returns>System type for the parameter.</returns>
+    public virtual Type GetSystemTypeForParam(string paramName)
+    {
+        if (Columns.TryGetValue(paramName, out ColumnDefinition? columnDefinition))
+        {
+            return columnDefinition.SystemType;
+        }
+
+        throw new DataApiBuilderException(
+            message: "Unexpected column name encountered",
+            statusCode: HttpStatusCode.InternalServerError,
+            subStatusCode: DataApiBuilderException.SubStatusCodes.UnexpectedError
+            );
     }
 }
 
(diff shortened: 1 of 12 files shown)


src/Core/Resolvers/BaseQueryStructure.cs
```diff
@@ -111,13 +120,27 @@ public class BaseQueryStructure
...
+                Parameters.Add(encodedParamName, new(value));
             }
             else
             {
-                Parameters.Add(encodedParamName, new(value));
+                SourceDefinition sourceDefinition = GetUnderlyingSourceDefinition();
+                DbType? dbTypeForParam = sourceDefinition.GetDbTypeForParam(paramName);
+                Type systemTypeForParam = sourceDefinition.GetSystemTypeForParam(paramName);
+
+                if (dbTypeForParam is null || ApiType is ApiType.GraphQL && systemTypeForParam == typeof(DateTime))
```
aaronburtle, line 133: nit: should still be able to say `systemTypeForParam is typeof(DateTime)`

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -346,8 +346,20 @@ protected List<LabelledColumn> GenerateOutputColumns()
         /// <param name="systemType"></param>
         /// <returns></returns>
         /// <exception cref="NotSupportedException"></exception>
-        protected static object ParseParamAsSystemType(string param, Type systemType)
+        protected object ParseParamAsSystemType(string param, Type systemType)
         {
+            if (systemType.Name.Equals("DateTime"))
```
aaronburtle, line 351: Can we just make this 

`if (systemType.Name.Equals("DateTime") && ApitType is ApiType.GraphQL)`

and then include the `return DateTime.Parse(param)` in the below switch case?

aaronburtle (approved):
Looks good!

I believe that our existing test cases covers this, but please double check and make sure that we have an edge case that will cover this change.

src/Core/Resolvers/BaseQueryStructure.cs
```diff
@@ -13,6 +15,11 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 {
     public class BaseQueryStructure
     {
+        /// <summary>
+        /// ApiType is required to correctly parse the params of DateTime types.
+        /// </summary>
+        protected ApiType ApiType { get; set; }
```
severussundar, line 21: nit: can be re-named to `ApiRequestType`

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -186,6 +203,28 @@ public bool IsAnyColumnNullable(List<string> columnsToCheck)
 
         return null;
     }
+
+    /// <summary>
+    /// Method to get the system type for:
+    /// 1. column for table/view,
+    /// 2. parameter for stored procedure.
```
severussundar, line 210: nit: description related to stored procedure parameters can be removed if inapplicable here

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -346,8 +346,20 @@ protected List<LabelledColumn> GenerateOutputColumns()
...
+            {
+                if (ApiType is ApiType.GraphQL)
+                {
+                    // To continue supporting 'Z' notation for GraphQL, we parse datetime values as datetimeoffset.
```
severussundar, line 355: Would be helpful to add links to `DateTimeOffset` and `DateTime` docs in the comments.

src/Core/Resolvers/BaseQueryStructure.cs
severussundar, line 133: nit: add additional brackets around conditions (based on evaluation order). Though it is possible to understand based on the logic below, it would help with readability.

severussundar (approved):

ayush3797:
Not required. See https://github.com/Azure/data-api-builder/pull/1636#issuecomment-1754875128.

PR #1681, opened 2023-09-04
Fix Config File Path  Issue

### LINKED ISSUE #1674: [Bug]: ACA deployment with latest sample script and latest Docker image cannot find DAB config
### What happened?

When using the sample script found at https://github.com/Azure/data-api-builder/blob/main/samples/azure/azure-container-apps-deploy.sh with a valid DAB config using latest schema version, latest `mcr.microsoft.com/azure-databases/data-api-builder` Docker container image, and properly configured environment variables in the sample script to deploy to Azure Container Apps, DAB can't seem to find the config on startup and returns HTTP 503 on entity REST requests.

After some troubleshooting, it was found that an older Docker image with tag `0.7.6` does not trigger this issue and returns requests fine; this issue seems to occur only on images after `0.7.6` (`0.8.44-rc` and the latest of this post, `0.8.49`, which is also the same as `latest` tag) regardless if the schema version is different from the Docker image version.

### Version

0.8.49

Azure SQL

Container Apps

REST

### Relevant log output

```Text
2023-08-31T17:16:02.452130428Z warn: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[60]
2023-08-31T17:16:02.452159362Z       Storing keys in a directory '/root/.aspnet/DataProtection-Keys' that may not be persisted outside of the container. Protected data will be unavailable when container is destroyed.
2023-08-31T17:16:02.452399157Z info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[62]
2023-08-31T17:16:02.452407764Z       User profile is available. Using '/root/.aspnet/DataProtection-Keys' as key repository; keys will not be encrypted at rest.
2023-08-31T17:16:02.465330499Z info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[58]
2023-08-31T17:16:02.465346428Z       Creating key {bfc3e456-e8bc-4fc8-8788-7792304f6555} with creation date 2023-08-31 17:16:02Z, activation date 2023-08-31 17:16:02Z, and expiration date 2023-11-29 17:16:02Z.
2023-08-31T17:16:02.472850394Z warn: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[35]
2023-08-31T17:16:02.472867196Z       No XML encryptor configured. Key {bfc3e456-e8bc-4fc8-8788-7792304f6555} may be persisted to storage in unencrypted form.
2023-08-31T17:16:02.476072036Z info: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[39]
2023-08-31T17:16:02.476080652Z       Writing data to file '/root/.aspnet/DataProtection-Keys/key-bfc3e456-e8bc-4fc8-8788-7792304f6555.xml'.
2023-08-31T17:16:02.491547038Z Unable to find config file:  does not exist.
2023-08-31T17:16:02.672510416Z Unable to find config file:  does not exist.
2023-08-31T17:16:02.687951924Z info: Microsoft.Hosting.Lifetime[14]
2023-08-31T17:16:02.687990687Z       Now listening on: http://[::]:5000
2023-08-31T17:16:02.689067467Z info: Microsoft.Hosting.Lifetime[0]
2023-08-31T17:16:02.689083808Z       Application started. Press Ctrl+C to shut down.
2023-08-31T17:16:02.689670331Z info: Microsoft.Hosting.Lifetime[0]
2023-08-31T17:16:02.689681321Z       Hosting environment: Production
2023-08-31T17:16:02.689686020Z info: Microsoft.Hosting.Lifetime[0]
2023-08-31T17:16:02.689690148Z       Content root path: /App
2023-08-31T17:17:00.902103944Z info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
2023-08-31T17:17:00.902166371Z       Request starting HTTP/1.1 GET http://***.westus2.azurecontainerapps.io/api/Product - -
2023-08-31T17:17:00.905627200Z warn: Microsoft.AspNetCore.HttpsPolicy.HttpsRedirectionMiddleware[3]
2023-08-31T17:17:00.905651636Z       Failed to determine the https port for redirect.
2023-08-31T17:17:00.950390183Z info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
2023-08-31T17:17:00.950445215Z       Request finished HTTP/1.1 GET http://***.westus2.azurecontainerapps.io/api/Product - - - 503 0 - 48.8507ms
```

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #1674 ConfigFile was not found when running in the Azure Container Apps.

## What is this change?

- Because of the way we used to fetch the cofigFileName, the path was lost, due to which using config file present in a different directory was giving `file not found` error.
- So we updated the method to fetch the configFileNamewihoutExtension to keep the original path.
- To run Azure Container Apps `--ConfigFileName` is provided with the full path name to the mounted config file.

NOTE:
* This issue might be caught in Az Container apps but this is not specifically related to Containers, because the current file name generation process was itself incorrect in 0.8.* release.
* When we try to get the `fileNameWithoutExtension` in the method `GetFileNameForEnvironment`, we used to lose directory information, for example: if baseConfigFile is`dab/configs/dab-config.json`, then `configFileNameWithoutExtension` was coming as `dab-config`, which is incorrect.
* This issue started happening from 0.8.* because in the earlier release (0.7.6), if config file name is provided by user, we directly used it without searching for overrides and environments, unlike 0.8.*. And the precedence check used to happen for cases where config file is not provided by user.
* Now, since 0.8.*, we started using the method `GetFileNameWithoutExtension`, which is incorrect causing File not found issue whenever the config file is not in current directory.
* The method `DoesFileExistInCurrentDirectory()` checks if the file exists by combining the current directory with the fileName. Now if file is existing in some other directory inside current directory, then for proper searching the file should contain complete info about the directory. So, if current dicrectory is `C:/dab`, and file is in `C:/dab/configs/dab-config.json`, then the file name should either be full path or `/configs/dab-configs.json` for it to be searched.

## Additional Change
1. Fixing name of variables and function name to correctly reflect their behavior.
2. Fixing issue with not honoring the user provided config file, (caused after 0.8 release). (Putting in this PR, because this PR aims at fixing how we are using the config files and not just a bug fix.)

## How was this tested?

- [X] Running In Azure Container Apps
BEFORE:
image

AFTER:
image

- [X] Local Testing
BEFORE:
image

AFTER:
image

- [X] Unit Tests

### DIFF
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -58,7 +58,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
         _fileSystem = fileSystem;
         _baseConfigFileName = baseConfigFileName;
         ConfigFileName = GetFileNameForEnvironment(Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"), false);
-        // If fileName based on environment is empty, then use the default config file name.
+        // If file for environment is not found, then the baseConfigFile is used as the final configFile for runtime engine.
         if (string.IsNullOrWhiteSpace(ConfigFileName))
         {
             ConfigFileName = baseConfigFileName;
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -173,7 +173,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
         string configFileNameWithExtension = $"{configFileName}{fileExtension}";
         string overriddenConfigFileNameWithExtension = GetOverriddenName(configFileName);
 
-        Console.WriteLine($"FileSystemRuntimeConfigLoader2, fileNameWithoutExtension: {fileNameWithoutExtension}, configFileName: {configFileName}, overriddenConfigFileNameWithExtension: {overriddenConfigFileNameWithExtension}");
+        Console.WriteLine($"FileSystemRuntimeConfigLoader2, fileNameWithoutExtension: {fileNameWithoutExtension}, configFileName: {configFileName}, overriddenConfigFileNameWithExtension: {overriddenConfigFileNameWithExtension}, configFileNameWithExtension: {configFileNameWithExtension}");
 
         if (considerOverrides && DoesFileExistInCurrentDirectory(overriddenConfigFileNameWithExtension))
         {
@@ -208,7 +208,9 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     public bool DoesFileExistInCurrentDirectory(string fileName)
     {
         string currentDir = _fileSystem.Directory.GetCurrentDirectory();
-        Console.WriteLine($"FileSystemRuntimeConfigLoader, currentDir: {currentDir}");
+        Console.WriteLine($"FileSystemRuntimeConfigLoader2, currentDir: {currentDir}");
+        Console.WriteLine($"FileSystemRuntimeConfigLoader2, fileName: {fileName}");
+        Console.WriteLine($"FileSystemRuntimeConfigLoader2, _fileSystem.Path.Combine(currentDir, fileName): {_fileSystem.Path.Combine(currentDir, fileName)}");
         return _fileSystem.File.Exists(_fileSystem.Path.Combine(currentDir, fileName));
     }
 
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -58,6 +58,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
         _fileSystem = fileSystem;
         _baseConfigFileName = baseConfigFileName;
         ConfigFileName = GetFileNameForEnvironment(Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"), false);
+        // if fileName based on environment is empty, then use the default config file name.
         if (string.IsNullOrWhiteSpace(ConfigFileName))
         {
             ConfigFileName = baseConfigFileName;
@@ -158,6 +159,9 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// <returns></returns>
     public string GetFileName(string? environmentValue, bool considerOverrides)
     {
+        // if the baseConfigFileName contains a path, we need to insure that is not lost. for example: baseConfigFileName = "config/dab-config.json"
+        // in this case, we need to get the directory name and the file name without extension and then combine them back. Else, we will lose the path
+        // and the file will be searched in the current directory.
         string fileNameWithoutExtension = _fileSystem.Path.Combine(_fileSystem.Path.GetDirectoryName(_baseConfigFileName) ?? string.Empty, _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFileName));
         string fileExtension = _fileSystem.Path.GetExtension(_baseConfigFileName);
         string configFileName =
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -150,7 +150,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// </summary>
     private string GetFinalConfigFilePath()
     {
-        if(!string.Equals(_baseConfigFilePath, DEFAULT_CONFIG_FILE_NAME))
+        if (!string.Equals(_baseConfigFilePath, DEFAULT_CONFIG_FILE_NAME))
         {
             // user provided config file is honoured.
             return _baseConfigFilePath;
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -27,7 +27,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
 {
     // This stores either the default config name e.g. dab-config.json
     // or user provided config file.
-    private string _baseConfigFileName;
+    private string _baseConfigFilePath;
 
     private readonly IFileSystem _fileSystem;
 
@@ -45,24 +45,20 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     public const string DEFAULT_CONFIG_FILE_NAME = $"{CONFIGFILE_NAME}{CONFIG_EXTENSION}";
 
     /// <summary>
-    /// Stores the config file name actually loaded by the engine.
+    /// Stores the config file actually loaded by the engine.
     /// It could be the base config file (e.g. dab-config.json), any of its derivatives with
     /// environment specific suffixes (e.g. dab-config.Development.json) or the user provided
     /// config file name.
+    /// It could also be the config file provided by the user.
     /// </summary>
-    public string ConfigFileName { get; internal set; }
+    public string ConfigFilePath { get; internal set; }
 
-    public FileSystemRuntimeConfigLoader(IFileSystem fileSystem, string baseConfigFileName = DEFAULT_CONFIG_FILE_NAME, string? connectionString = null)
+    public FileSystemRuntimeConfigLoader(IFileSystem fileSystem, string baseConfigFilePath = DEFAULT_CONFIG_FILE_NAME, string? connectionString = null)
         : base(connectionString)
     {
         _fileSystem = fileSystem;
-        _baseConfigFileName = baseConfigFileName;
-        ConfigFileName = GetFileNameForEnvironment(Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"), false);
-        // If file for environment is not found, then the baseConfigFile is used as the final configFile for runtime engine.
-        if (string.IsNullOrWhiteSpace(ConfigFileName))
-        {
-            ConfigFileName = baseConfigFileName;
-        }
+        _baseConfigFilePath = baseConfigFilePath;
+        ConfigFilePath = GetFinalConfigFilePath();
     }
 
     /// <summary>
@@ -104,7 +100,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// <returns>True if the config was loaded, otherwise false.</returns>
     public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false)
     {
-        return TryLoadConfig(ConfigFileName, out config, replaceEnvVar);
+        return TryLoadConfig(ConfigFilePath, out config, replaceEnvVar);
     }
 
     /// <summary>
@@ -150,6 +146,29 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     }
 
     /// <summary>
+    /// This method returns the final config file name that will be used by the runtime engine.
+    /// </summary>
+    private string GetFinalConfigFilePath()
+    {
+        if(!string.Equals(_baseConfigFilePath, DEFAULT_CONFIG_FILE_NAME))
+        {
+            // user provided config file is honoured.
+            return _baseConfigFilePath;
+        }
+
+        // ConfigFile not explicitly provided by user, so we need to get the config file name based on environment.
+        string configFilePath = GetFileNameForEnvironment(Environment.GetEnvironmentVariable(ASP_NET_CORE_ENVIRONMENT_VAR_NAME), false);
+
+        // If file for environment is not found, then the baseConfigFile is used as the final configFile for runtime engine.
+        if (string.IsNullOrWhiteSpace(configFilePath))
+        {
+            return _baseConfigFilePath;
+        }
+
+        return configFilePath;
+    }
+
+    /// <summary>
     /// Generates the config file name and a corresponding overridden file name,
     /// With precedence given to overridden file name, returns that name
     /// if the file exists in the current directory, else an empty string.
@@ -160,24 +179,24 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// <returns></returns>
     public string GetFileName(string? environmentValue, bool considerOverrides)
     {
-        // If the baseConfigFileName contains a path, we need to ensure that it is not lost. for example: baseConfigFileName = "config/dab-config.json"
+        // If the baseConfigFilePath contains directory info, we need to ensure that it is not lost. for example: baseConfigFilePath = "config/dab-config.json"
         // in this case, we need to get the directory name and the file name without extension and then combine them back. Else, we will lose the path
         // and the file will be searched in the current directory.
-        string fileNameWithoutExtension = _fileSystem.Path.Combine(_fileSystem.Path.GetDirectoryName(_baseConfigFileName) ?? string.Empty, _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFileName));
-        string fileExtension = _fileSystem.Path.GetExtension(_baseConfigFileName);
-        string configFileName =
+        string filePathWithoutExtension = _fileSystem.Path.Combine(_fileSystem.Path.GetDirectoryName(_baseConfigFilePath) ?? string.Empty, _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFilePath));
+        string fileExtension = _fileSystem.Path.GetExtension(_baseConfigFilePath);
+        string configFilePath =
             !string.IsNullOrEmpty(environmentValue)
-            ? $"{fileNameWithoutExtension}.{environmentValue}"
-            : $"{fileNameWithoutExtension}";
-        string configFileNameWithExtension = $"{configFileName}{fileExtension}";
-        string overriddenConfigFileNameWithExtension = GetOverriddenName(configFileName);
+            ? $"{filePathWithoutExtension}.{environmentValue}"
+            : $"{filePathWithoutExtension}";
+        string configFileNameWithExtension = $"{configFilePath}{fileExtension}";
+        string overriddenConfigFileNameWithExtension = GetOverriddenName(configFilePath);
 
-        if (considerOverrides && DoesFileExistInCurrentDirectory(overriddenConfigFileNameWithExtension))
+        if (considerOverrides && DoesFileExistInDirectory(overriddenConfigFileNameWithExtension))
         {
             return overriddenConfigFileNameWithExtension;
         }
 
-        if (DoesFileExistInCurrentDirectory(configFileNameWithExtension))
+        if (DoesFileExistInDirectory(configFileNameWithExtension))
         {
             return configFileNameWithExtension;
         }
@@ -185,9 +204,9 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
         return string.Empty;
     }
 
-    private static string GetOverriddenName(string fileName)
+    private static string GetOverriddenName(string filePath)
     {
-        return $"{fileName}.overrides{CONFIG_EXTENSION}";
+        return $"{filePath}.overrides{CONFIG_EXTENSION}";
     }
 
     /// <summary>
@@ -199,10 +218,16 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
         return $"{fileName}.{environmentValue}{CONFIG_EXTENSION}";
     }
 
-    public bool DoesFileExistInCurrentDirectory(string fileName)
+    /// <summary>
+    /// Checks if the file exists in the directory.
+    /// Works for both relative and absolute paths.
+    /// </summary>
+    /// <param name="filePath"></param>
+    /// <returns>True if file is found, else false.</returns>
+    public bool DoesFileExistInDirectory(string filePath)
     {
         string currentDir = _fileSystem.Directory.GetCurrentDirectory();
-        return _fileSystem.File.Exists(_fileSystem.Path.Combine(currentDir, fileName));
+        return _fileSystem.File.Exists(_fileSystem.Path.Combine(currentDir, filePath));
     }
 
     /// <summary>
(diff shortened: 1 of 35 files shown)


src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -58,6 +58,11 @@ public FileSystemRuntimeConfigLoader(IFileSystem fileSystem, string baseConfigFi
         _fileSystem = fileSystem;
         _baseConfigFileName = baseConfigFileName;
         ConfigFileName = GetFileNameForEnvironment(Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"), false);
+        // if fileName based on environment is empty, then use the default config file name.
```
Aniruddh25, line 61: The function: `GetFileNameForEnvironment` should return the default config file name if the fileName based on environment is empty - the 3rd index is string.Empty in the environment variable loop. Did you observe this is not happening?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -154,7 +160,10 @@ public string GetFileNameForEnvironment(string? aspnetEnvironment, bool consider
     /// <returns></returns>
     public string GetFileName(string? environmentValue, bool considerOverrides)
     {
-        string fileNameWithoutExtension = _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFileName);
+        // if the baseConfigFileName contains a path, we need to insure that is not lost. for example: baseConfigFileName = "config/dab-config.json"
+        // in this case, we need to get the directory name and the file name without extension and then combine them back. Else, we will lose the path
+        // and the file will be searched in the current directory.
+        string fileNameWithoutExtension = _fileSystem.Path.Combine(_fileSystem.Path.GetDirectoryName(_baseConfigFileName) ?? string.Empty, _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFileName));
```
Aniruddh25, line 166: In the case of remote docker, how would the `_baseConfigFileName` contain a directory path?
The constructor of `FileSystemRuntimeConfigLoader` defaults this parameter to be DEFAULT_CONFIG_FILE_NAME which is dab_config.json

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,36 @@ string[] linkingTargetFields
...
+        /// This test checks that the final config used by runtime engine doesn't lose the directory information
+        /// provided in the base config file path when loading the config based on the environment.
+        /// </summary>
+        /// <param name="addDirectory"></param>
+        /// <param name="environmentValue"></param>
+        /// <param name="baseConfigFilePath"></param>
+        /// <param name="finalConfigFilePath"></param>
+        [DataTestMethod]
+        [DataRow(false, "", "current-dir-dab-config.json", "current-dir-dab-config.json", DisplayName = "Config file in the current directory")]
+        [DataRow(true, "", "test\\diff-dir-dab-config.json", "test\\diff-dir-dab-config.json", DisplayName = "Config file in a different directory")]
```
Aniruddh25, line 2030: What if its an absolute path to the config file? for example, `C:\\dab\\test\\diff-dir-dab-config.json` ?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,36 @@ string[] linkingTargetFields
...
+        [DataRow(false, "Test", "current-dir-dab-config.json", "current-dir-dab-config.Test.json", DisplayName = "Config file in the current directory")]
```
Aniruddh25, line 2031:
```suggestion
        [DataRow(false, "Test", "current-dir-dab-config.json", "current-dir-dab-config.Test.json", DisplayName = "Config file in the current directory with non-empty environment value")]
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,36 @@ string[] linkingTargetFields
...
+        [DataRow(true, "Test", "test\\diff-dir-dab-config.json", "test\\diff-dir-dab-config.Test.json", DisplayName = "Config file in a different directory")]
```
Aniruddh25, line 2032:
```suggestion
        [DataRow(true, "Test", "test\\diff-dir-dab-config.json", "test\\diff-dir-dab-config.Test.json", DisplayName = "Config file in a different directory with non-empty environment value")]
```

src/Config/FileSystemRuntimeConfigLoader.cs
abhishekkumams: `_baseConfigFileName` would contain the default when nothing is provided as part of `--ConfigFileName`. If the `_baseConfigFileName` doesn't contain drectory path, then it searches in the current directory.

src/Config/FileSystemRuntimeConfigLoader.cs
abhishekkumams: yes, it was returning empty when the config file is not found, as can be seen in the Logs. Due to which logs are like this `Unable to find config File: does not exist.`

src/Config/FileSystemRuntimeConfigLoader.cs
Aniruddh25: I understand when --ConfigFileName is not provided, default is picked up. 

My question is why would _baseConfigFileName contain a directory name in the remote docker scenario? Essentially, why does this fix work?

Aniruddh25: Discussed offline, PR description needs to be updated. To run Azure Container Apps, 
`--ConfigFileName` is provided with the full path name to the mounted config file, hence the remote docker scenario is fixed with this solution.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -58,6 +58,11 @@ public FileSystemRuntimeConfigLoader(IFileSystem fileSystem, string baseConfigFi
         _fileSystem = fileSystem;
         _baseConfigFileName = baseConfigFileName;
         ConfigFileName = GetFileNameForEnvironment(Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"), false);
+        // If fileName based on environment is empty, then use the default config file name.
```
Aniruddh25, line 61: nit: Comment needs to be updated.

Aniruddh25 (approved):
Thanks for investigating this, and fixing with high priority for 0.8! LGTM after resolving comments as discussed.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -154,7 +160,10 @@ public string GetFileNameForEnvironment(string? aspnetEnvironment, bool consider
     /// <returns></returns>
     public string GetFileName(string? environmentValue, bool considerOverrides)
     {
-        string fileNameWithoutExtension = _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFileName);
+        // If the baseConfigFileName contains a path, we need to insure that is not lost. for example: baseConfigFileName = "config/dab-config.json"
```
Aniruddh25, line 163:
```suggestion
        // If the baseConfigFileName contains a path, we need to ensure that it is not lost. for example: baseConfigFileName = "config/dab-config.json"
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,47 @@ string[] linkingTargetFields
...
         }
 
+        /// <summary>
+        /// This test checks that the final config used by runtime engine doesn't lose the directory information
+        /// if provided in the base config file path when loading the config based on the environment.
+        /// </summary>
+        /// <param name="addDirectory"></param>
+        /// <param name="environmentValue"></param>
+        /// <param name="baseConfigFilePath"></param>
+        /// <param name="finalConfigFilePath"></param>
```
seantleonard, line 2027: Outdated

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -154,7 +160,10 @@ public string GetFileNameForEnvironment(string? aspnetEnvironment, bool consider
     /// <returns></returns>
     public string GetFileName(string? environmentValue, bool considerOverrides)
     {
-        string fileNameWithoutExtension = _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFileName);
+        // If the baseConfigFileName contains a path, we need to ensure that it is not lost. for example: baseConfigFileName = "config/dab-config.json"
+        // in this case, we need to get the directory name and the file name without extension and then combine them back. Else, we will lose the path
+        // and the file will be searched in the current directory.
+        string fileNameWithoutExtension = _fileSystem.Path.Combine(_fileSystem.Path.GetDirectoryName(_baseConfigFileName) ?? string.Empty, _fileSystem.Path.GetFileNameWithoutExtension(_baseConfigFileName));
```
seantleonard, line 166: You have code here which preserves the path (per comment), however, the next few lines of code check `DoesFileExistInCurrentDirectory()` which would presumably fail and hit line 185 to return empty string.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,47 @@ string[] linkingTargetFields
...
+            bool IsDifferentDirectory,
+            string environmentValue,
+            bool isAbsolutePath,
+            string baseConfigFilePath,
+            string finalConfigFilePath)
+        {
+            MockFileSystem fileSystem = new();
+            if (IsDifferentDirectory)
+            {
+                fileSystem.AddDirectory("test");
```
seantleonard, line 2047: In `runtimeConfigLoader.GetFileName(...)` there is a call to `_fileSystem.Directory.GetCurrentDirectory();`. Is the current directory in the mock file system representative of the value returned in a real Az Container Apps deployment? 

The PR description is not clear about what change after `0.7.6` caused the bug this PR is fixing. The requirement to now supply a 'full' file path looks to make everything work, but it is not clear why it works because per the container deploy script, the config file is uploaded to the root of the config file share:
https://github.com/Azure/data-api-builder/blob/f8373b8e231ddae1f8bba99160f6de1932d9f4a6/samples/azure/azure-container-apps-deploy.sh#L50-L52

What would be an example of the full file path provided in Az container apps? do you need to know the mount name/location of the file share that is created?

seantleonard:
In the past, we've had issues with local Docker containers finding the desired config file. The root cause of those issues where an incorrect file/folder mapping to the Docker container. could that be the case with this issue too? 
In the PR description, can you describe what behavior you observed in Az Container apps?
- Was the desired config file being overwritten in the file share, when looking for a config file in the root of the file share?
- Is this issue only seen when the config file is placed in a folder? (not root of file share) e.g. `FileShare::\\config.json`vs. `FileShare::\\subFolder\config.json`

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
aaronburtle: > What would be an example of the full file path provided in Az container apps? do you need to know the mount name/location of the file share that is created?

As I understand how the `az storage file upload` command works when uploading to an azure file share, the full path would be made up of <mount-location>/<file-share-name>/<directory-path>/<filename>

so if the mount location was `/mnt/azfiles`, and the file share name was `fileshare`, and the directory path was `DAb/config`, and the file name was `config.json`, then the full path would be

`/mnt/azfiles/fileshare/DAb/config/config.json`

I suppose this is something we can document, but I guess that the person using the az file share should know what a full path is within this context.

aaronburtle (commented):
Looks good, just waiting on some clarification for what we mean by full path, and the comment that sean raised about returning empty.

aaronburtle (approved):

src/Config/FileSystemRuntimeConfigLoader.cs
abhishekkumams: It won't fail because the file name would contain the path details as well. for example if fileName is `/dab/configs/dab-config.json` and we are in root directory(`/`), then it would search the file using the directory information contained in the file Name. which is why it was failing before because earlier it used to search for `dab-config.json` in the current directory as the directory info was getting lost.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
abhishekkumams: @Sean, when deploying in Azure Container Apps, we need to mount the volume as well from Azure Container Environment, so if the mounted volume contained the config in `/mnt/configs.dab-config.json`, then we need to provide the correct path in the `--ConfigFileName`. Due to changes in 0.8 release, the directory info was getting lost and we were left with `dab-config.json`, and hence getting config file not found error.

src/Config/FileSystemRuntimeConfigLoader.cs
abhishekkumams: updated PR description as well

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
abhishekkumams: updated PR description

seantleonard: Okay, I was confused because of the usage of the term "file name" which to me would be `dab-config.json`. If there are path components included like `mnt/folder/dab-config.json`, i would  use the term 'full path' or 'relative path' depending on the path..

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -190,10 +218,16 @@ public static string GetEnvironmentFileName(string fileName, string environmentV
...
+    /// Checks if the file exists in the directory.
+    /// Works for both relative and absolute paths.
+    /// </summary>
+    /// <param name="filePath"></param>
+    /// <returns>True if file is found, else false.</returns>
+    public bool DoesFileExistInDirectory(string filePath)
     {
         string currentDir = _fileSystem.Directory.GetCurrentDirectory();
-        return _fileSystem.File.Exists(_fileSystem.Path.Combine(currentDir, fileName));
+        return _fileSystem.File.Exists(_fileSystem.Path.Combine(currentDir, filePath));
```
seantleonard, line 230: Originally, the plan was to stop using `Path.Combine`. However, after further inspection of [documentation]https://learn.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=net-7.0#system-io-path-combine(system-string-system-string)()
> This method is intended to concatenate individual strings into a single string that represents a file path. However, if an argument other than the first contains a rooted path, any previous path components are ignored, and the returned string begins with that rooted path component.


`Path.Combine(currentDir, filePath)` resolves to `"C:\dab\configs\dab-config.json"` for
- currentDir == `"C:\dab\configs\"`
- filePath == `"C:\dab\configs\dab-config.json"` or `"configs\dab-config.json"` or `"dab-config.json"`

if `filePath` is a rooted path different that the first param of `currentDir`, the rooted path in `filePath` takes precedence (Per docs):

So, `Path.Combine(currentDir, filePath)` resolves to `"D:\dab2\configfiles\dab-config.json"` for
- currentDir == `"C:\dab\configs\"`
- filePath == `"D:\dab2\configfiles\dab-config.json"`

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -190,10 +218,16 @@ public static string GetEnvironmentFileName(string fileName, string environmentV
         return $"{fileName}.{environmentValue}{CONFIG_EXTENSION}";
     }
 
-    public bool DoesFileExistInCurrentDirectory(string fileName)
+    /// <summary>
+    /// Checks if the file exists in the directory.
+    /// Works for both relative and absolute paths.
+    /// </summary>
+    /// <param name="filePath"></param>
```
seantleonard, line 225: would want to note that absolute and relative path are accepted here.

src/Config/FileSystemRuntimeConfigLoader.cs
abhishekkumams: Thanks for the details and the link.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,72 @@ string[] linkingTargetFields
...
+        [DataRow("test-configs/my-config.json", "Test", true, true, "test-configs/dab-config.Test.json", "test-configs/my-config.json", DisplayName = "Config file in the different directory provided by user and environment variable is set and environment file is present and absolute path is provided")]
+        [DataRow(null, "", false, false, null, "dab-config.json", DisplayName = "Config file not provided by user and environment variable is not set")]
+        [DataRow(null, "Test", false, false, "dab-config.Test.json", "dab-config.json", DisplayName = "Config file not provided by user and environment variable is set and environment file is present")]
+        [DataRow(null, "Test", false, false, null, "dab-config.json", DisplayName = "Config file not provided by user and environment variable is set and environment file is not present")]
+        public void TestCorrectConfigFileIsSelectedForRuntimeEngine(
+            string userProvidedConfigFilePath,
+            string environmentValue,
+            bool IsDifferentDirectory,
```
seantleonard, line 2051:
```suggestion
            bool isDifferentDirectory,
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,72 @@ string[] linkingTargetFields
...
+            string environmentFile,
+            string finalConfigFilePath)
+        {
+            MockFileSystem fileSystem = new();
+            if (IsDifferentDirectory)
+            {
+                fileSystem.AddDirectory("test-configs");
+            }
+
+            if (isAbsolutePath)
```
seantleonard, line 2062: Would make test easier to read if `isAbsolutePath` was not an additional test parameter. (Just lots of booleans to read and parse).
You could instead use [https://learn.microsoft.com/en-us/dotnet/api/system.io.path.ispathrooted?view=net-6.0#system-io-path-ispathrooted(system-string)](`Path.IsPathRooted`)

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,72 @@ string[] linkingTargetFields
...
+            if (userProvidedConfigFilePath is not null)
+            {
+                runtimeConfigLoader = new(fileSystem, userProvidedConfigFilePath);
+            }
+            else
+            {
+                runtimeConfigLoader = new(fileSystem);
+            }
+
+            Assert.AreEqual(finalConfigFilePath, runtimeConfigLoader.ConfigFilePath);
```
seantleonard, line 2083: Will `runtimeConfigLoader.ConfigFilePath` always be a full path? this test seems to indicate that it could be either an absolute path, a relative path, or just a file name.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,72 @@ string[] linkingTargetFields
             }
         }
 
+        /// <summary>
+        /// This test checks that the final config used by runtime engine doesn't lose the directory information
+        /// if provided by the user.
+        /// It also validates that if config file is provided by the user, it will be used directly irrespective of
+        /// environment variable being set or not. 
+        /// When user doesn't provide a config file, we check if environment variable is set and if it is, we use
+        /// the config file specified by the environment variable, else we use the default config file.
```
seantleonard, line 2026: nit for future consideration. each of this bullet points would have been good to be separate tests for readability purposes. While there may be some duplicate code, the tests would be easier to read and have single purpose.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2017,6 +2017,72 @@ string[] linkingTargetFields
...
+        public void TestCorrectConfigFileIsSelectedForRuntimeEngine(
+            string userProvidedConfigFilePath,
+            string environmentValue,
+            bool IsDifferentDirectory,
+            bool isAbsolutePath,
+            string environmentFile,
+            string finalConfigFilePath)
+        {
+            MockFileSystem fileSystem = new();
+            if (IsDifferentDirectory)
```
seantleonard, line 2057: instead of a test method input param, perhaps use https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getdirectoryname?view=net-6.0#system-io-path-getdirectoryname(system-string) `Path.GetDirectoryName()` to determine if a path value is included.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
abhishekkumams: I'm using this variable not to know if the provided path is Absolute path or not, but to make it absolute, since it's a MockFileSystem, we need to make the path absolute programmatically.

abhishekkumams: updated the variable name to `useAbsolutePath` to indicate it's correct usage

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -27,7 +27,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
 {
     // This stores either the default config name e.g. dab-config.json
     // or user provided config file.
```
Aniruddh25, line 29:
```suggestion
    // or user provided config file which could a relative file path, absolute file path or simply the file name assumed to be in current directory.
```

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -26,7 +26,7 @@ namespace Azure.DataApiBuilder.Config;
 public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
 {
     // This stores either the default config name e.g. dab-config.json
-    // or user provided config file.
+    // or user provided config file which could a relative file path, absolute file path or simply the file name assumed to be in current directory.
```
Aniruddh25, line 29:
```suggestion
    // or user provided config file which could be a relative file path, absolute file path or simply the file name assumed to be in current directory.
```

Aniruddh25 (approved):

Issue #1679, opened 2023-09-01
[Enhancement] Boolean values in CLI


JerryNixon:
# Summary
Adjust CLI boolean switches to clearly allow both `true` and `false`.

> Note: this should be implemented soon to prevent future code churn. 

# Motivation
The example `-rest.disabled` is unclear as it seems like it should require the value `true` or `false` like `dab init -rest.disabled true`; however, the behavior is actually like this: its presence alone is the value `true` & the value `false` is only possible when the switch is not present. This is quite unintuitive and prone to error for developers who are not meticulously referencing our documentation. This original approach was loosely based on alignment with the AZ CLI but proves far too unintuitive to leave like it is. See below. 

# Old behavior
`dab init -rest.disabled` As stated above, the presence of the switch is a `true` value for the "negative" switch. The `false` value of this "negative switch, is to omit it from the CLI op. 

# New behavior
`dab init -rest.enabled false` As stated above, the positive switch receives either a `true` or `false` value to indicate the operation. It has defaults, but its omission is not necessary to set a specific value. It can set either `true` or `false` explicitly or fall back to the default in omission. In the case of some CLI switches, there may not be a default value, making its presence required.

## Important notes:

1. The original "negative" switch remains unchanged but is _deprecated_. The reason for this is **only** for backwards compatibility with the old behavior as we move forward. All "negative" switches will also be removed from any documentation & samples to discourage use going forward. We can discuss breaking changes before GA for "negative" switches. 
2. The new "positive" switch is not boolean, but a `true/false` enum. The reason for this is to overcome any limitations in the args library we are using. It will appear to be a boolean to the developer. As a result, the enum will simulate boolean behavior of passing `true` and `false`. 
3. Going forward, all boolean-like switches in the CLI will use this approach - both avoiding any "negative" switches and simulating boolean ops with a `true/false` enum.

ayush3797:
Makes sense to me! A question: Since negative switches are deprecated but in a weird scenario (which may still occur) when a user uses both the flags in the init command, there can be a conflict: 
`dab init  ..... --rest.disabled --rest.enabled true`

I think we should:

1. Log a warning whenever the negative flag is included in the init command - that this is a deprecated feature and we recommend to use `rest.enabled` instead.
2. Log a warning that precedence will be given to the rest.enabled flag if included in the init command.

Same goes for other boolean flags.

JerryNixon:
> 2. When there is a conflict between the two flags, throw an exception.

[...]

PR #1686, opened 2023-09-05
Adding support for positive boolean options in CLI

ayush3797:
## Why make this change?
To adjust CLI boolean options to properly allow true/false values as per: https://github.com/Azure/data-api-builder/issues/1679

## What is this change?
1. Added new options - `--rest.enabled` and `--graphql.enabled` to `dab init` options.
2. Introduce an enum `CliBool` with values - `None`,`True`,`False`, to represent boolean options in CLI. When the option is omitted from the init command, the value of `CliBool.None` would be assigned to the option, or else appropriate `True`/`False` value would be parset out and assigned to the option.
2. When both positive and negative flags are included in the init command then:
    - Config generation fails when the semantics differ.
    - If the semantics match, then appropriate value is assigned to `Runtime.Rest.Enabled`/`Runtime.GraphQL.Enabled` properties.
3. When only one of `--rest.enabled`/`--rest.disabled ` (or corresponding flags for graphql) option is included in the init command, the value assigned to that option is honored.
## How was this tested?
- [x] Unit Tests - are added via `EndToEndTests.TestEnabledDisabledFlagsForApis` test confirming the behavior for different cases.


src/Config/ObjectModel/CliBool.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel
+{
+    public enum CliBool
+    {
+        // The enum value None is required to determine whether a value was provided for a CLI option. In case no value is provided for the option,
```
abhishekkumams, line 8: why we need `None`, if the user doesn't want to provide a value, they can just not use the option.

ayush3797: This is because of opposite nature of `rest.disabled` and `rest.enabled` options. 
Assume a dab init command where `rest.disabled` is included in the init command and `rest.enabled` is omitted.
Since `rest.disabled` is included, it means we want to disable the REST endpoint. 
Now for `rest.enabled` we have an enum. And when we do not provide a value for an enum, it assumes the default value of first enum value. If don't include `None`, it would assume the value of True/False depending on whichever is written first in the enum declaration. This is not what we want. We want to explicitly know that the `rest.enabled` option was not included in the init command and no consideration should be given to it in the case when it is coupled with `rest.disabled` flag. Exclusion of `rest.enabled` coupled with inclusion of `rest.disabled` means the user is not bothered about the behavior of `rest.enabled` and he would be assuming that the rest endpoint will be disabled.

src/Cli/ConfigGenerator.cs
```diff
@@ -206,8 +226,8 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
                 Schema: dabSchemaLink,
                 DataSource: dataSource,
                 Runtime: new(
-                    Rest: new(!restDisabled, restPath ?? RestRuntimeOptions.DEFAULT_PATH),
-                    GraphQL: new(!options.GraphQLDisabled, graphQLPath),
```
severussundar, line 210: If only `--graphql.disabled` was used without `--graphql.enabled false`, would graphQL still get disabled?

src/Cli/ConfigGenerator.cs
```diff
@@ -221,6 +241,46 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
...
+        /// <summary>
+        /// Helper method to determine if the api is enabled or not based on the enabled/disabled options in the dab init command.
+        /// The method also validates that there is no mismatch in semantics of enabling/disabling the REST/GraphQL API(s)
+        /// based on the values supplied in the enabled/disabled options for the API in the init command.
+        /// </summary>
+        /// <param name="apiDisabledOptionValue">Value of disabled option as in the init command. If the option is omitted in the command, default value is assigned.</param>
+        /// <param name="apiEnabledOptionValue">Value of enabled option as in the init command. If the option is omitted in the command, default value is assigned.</param>
+        /// <param name="apiType">ApiType - REST/GraphQL.</param>
+        /// <param name="isApiEnabled">Boolean value indicating whether the API endpoint is enabled or not.</param>
+        /// <exception cref="Exception">Thrown when the semantics of enabled/disabled options differ.</exception>
```
severussundar, line 253: nit: An exception is not thrown.

src/Cli/ConfigGenerator.cs
```diff
@@ -92,13 +92,33 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
             Dictionary<string, JsonElement> dbOptions = new();
 
             HyphenatedNamingPolicy namingPolicy = new();
-            bool restDisabled = options.RestDisabled;
+
+            // If --rest.disabled flag is included in the init command, we log a warning to not use this flag as it is a deprecated feature.
+            if (options.RestDisabled is true)
+            {
+                _logger.LogWarning("The option --rest.disabled is deprecated and its support will be removed in future versions." +
```
seantleonard, line 99:
```suggestion
                _logger.LogWarning("The option --rest.disabled will be deprecated and its support will be removed in future versions." +
```

src/Cli/ConfigGenerator.cs
```diff
@@ -92,13 +92,33 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
...
+                    " We recommend to use the --rest.enabled option instead.");
+            }
+
+            // If --graphql.disabled flag is included in the init command, we log a warning to not use this flag as it is a deprecated feature.
+            if (options.GraphQLDisabled is true)
+            {
+                _logger.LogWarning("The option --graphql.disabled is deprecated and its support will be removed in future versions." +
```
seantleonard, line 106:
```suggestion
                _logger.LogWarning("The option --graphql.disabled will be deprecated and support for the option will be removed in future versions of Data API builder." +
```

src/Cli/ConfigGenerator.cs
```diff
@@ -92,13 +92,33 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
...
+                    " We recommend to use the --graphql.enabled option instead.");
```
seantleonard, line 107:
```suggestion
                    " We recommend that you use the --graphql.enabled option instead.");
```

src/Cli/ConfigGenerator.cs
```diff
@@ -92,13 +92,33 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
             Dictionary<string, JsonElement> dbOptions = new();
 
             HyphenatedNamingPolicy namingPolicy = new();
-            bool restDisabled = options.RestDisabled;
+
+            // If --rest.disabled flag is included in the init command, we log a warning to not use this flag as it is a deprecated feature.
+            if (options.RestDisabled is true)
+            {
+                _logger.LogWarning("The option --rest.disabled is deprecated and its support will be removed in future versions." +
+                    " We recommend to use the --rest.enabled option instead.");
```
seantleonard, line 100:
```suggestion
                    " We recommend that you use the --rest.enabled option instead.");
```

src/Cli/ConfigGenerator.cs
```diff
@@ -221,6 +241,46 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
...
+            if (apiEnabledOptionValue is CliBool.None)
+            {
+                // This means that the --api.enabled option was not included in the init command.
+                isApiEnabled = !apiDisabledOptionValue;
+                return true;
+            }
+
+            // We hit this code only when both --api.enabled and --api.disabled flags are included in the init command.
+            isApiEnabled = bool.Parse(apiEnabledOptionValue.ToString());
+            if (!apiDisabledOptionValue != isApiEnabled)
```
abhishekkumams, line 273: The !apiDisabledOptionValue != isApiEnabled expression can be simplified to apiDisabledOptionValue == isApiEnabled. This will make the code easier to read and understand.

src/Cli/ConfigGenerator.cs
ayush3797: thanks for noticing, forgot to remove.

src/Cli/ConfigGenerator.cs
ayush3797: right thanks!

seantleonard (approved):
Looks good! one last question

abhishekkumams (approved):

PR #1691, opened 2023-09-07
Avoid escaping HTML sensitive characters when writing configuration to file

### LINKED ISSUE #1687: [Bug]: Escape characters in Config from CLI
### What happened?

When you run this CLI command:

`dab init --host-mode development --database-type mssql --connection-string "@env('my-connection-string')"`

The `connection-string` line in the configuration file looks like this:

````JSON
{
  "$schema": "https://github.com/Azure/data-api-builder/releases/download/v0.8.49/dab.draft.schema.json",
  "data-source": {
    "database-type": "mssql",
    "connection-string": "@env(\u0027my-connection-string\u0027)",
    "options": {
      "set-session-context": false
    }
  },
````

This value `\u0027` is the unicode escape sequence for single quote. Does it work? Yes. However, this _cryptic_ value is not obvious to developers (especially new or emerging developers) and using the actual single quote character `'` instead of this lets the CLI create expected values that do not require mental parsing to read. I do not suggest we stop supporting escape sequences, but I do suggest we use the character supplied by the developer and don't translate it under them. 

### Version

Microsoft.DataApiBuilder 0.8.49+0aebfe2179ee27d1e1723776dd4662459674e987

Azure SQL

Local (including CLI)

REST, GraphQL

### Relevant log output

_No response_

### Code of Conduct

### PR DESCRIPTION
**Pull request by Aniruddh25**
## Why make this change?

- Closes #1687 

## What is this change?

- Set the Encoder for JsonSerializationOptions to be [`JavaScriptEncoder.UnsafeRelaxedJsonEscaping`](https://learn.microsoft.com/en-us/dotnet/api/system.text.encodings.web.javascriptencoder.unsaferelaxedjsonescaping?view=net-7.0) instead of the default encoder. 
- Although [this article](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/character-encoding#serialize-all-characters) mentions use this encoder with caution, since we are sure the file contents will be deserialized as UTF-8 JSON, we should be safe to use this encoder.
- We were using this encoder previously from the fix in #1386, but with #1402, that original fix was essentially not being used since #1402 introduced usage of the new `GetJsonSerializerOptions()` function in `RuntimeConfigLoader`.

## How was this tested?

- [X] Unit Tests. `TestSpecialCharactersInConnectionString` is moved to the `ConfigGeneratorTests` since keeping it as part of `InitTests` with the Verify framework wasn't really testing the scenario since we ignore connection strings in the snapshots. 
- This regression from #1402 was not caught because this test scenario was effectively being skipped.
- Note, this test is modified to explicitly compare the expected and actual json string from file since any usage of JsonSerializer/JObject.Parse would require using the same encoder to deserialize making the test itself run similar code which it is testing. Parsing into JSON objects would treat `\u0027` and `'` as equal whereas string comparison wont. Hence, comparison using string is necessary to catch regressions here.

## Sample Request(s)

BEFORE:
`dab init --host-mode development --database-type mssql --connection-string "@env('my-connection-string')"`

creates the configuration file as follows:

```
{
  "$schema": "https://github.com/Azure/data-api-builder/releases/download/v0.8.49/dab.draft.schema.json",
  "data-source": {
    "database-type": "mssql",
    "connection-string": "@env(\u0027my-connection-string\u0027)",
    "options": {
      "set-session-context": false
    }
  }
```
Note, the value of connection string.

AFTER: Same command creates the following file:
```
{
  "$schema": "https://github.com/Azure/data-api-builder/releases/download/v0.8.49/dab.draft.schema.json",
  "data-source": {
    "database-type": "mssql",
    "connection-string": "@env('my-connection-string')",
    "options": {
      "set-session-context": false
    }
  }
```

### DIFF
--- a/src/Cli.Tests/ConfigGeneratorTests.cs
+++ b/src/Cli.Tests/ConfigGeneratorTests.cs
@@ -176,15 +176,19 @@ public class ConfigGeneratorTests
               ""entities"": {}
             }");
 
-        expectedRuntimeConfigJson = expectedRuntimeConfigJson.Replace("\n", string.Empty);
         expectedRuntimeConfigJson = expectedRuntimeConfigJson.Replace(" ", string.Empty);
+        expectedRuntimeConfigJson = expectedRuntimeConfigJson.Replace("\n", string.Empty);
+        expectedRuntimeConfigJson = expectedRuntimeConfigJson.Replace("\r\n", string.Empty);
 
         Assert.IsTrue(TryGenerateConfig(options, _runtimeConfigLoader!, _fileSystem!));
 
         StringBuilder actualRuntimeConfigJson = new(_fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE, Encoding.Default));
         actualRuntimeConfigJson = actualRuntimeConfigJson.Replace(" ", string.Empty);
         actualRuntimeConfigJson = actualRuntimeConfigJson.Replace("\r\n", string.Empty);
+        actualRuntimeConfigJson = actualRuntimeConfigJson.Replace("\n", string.Empty);
 
+        // Comparing explicit strings here since parsing these into JSON would lose
+        // the test scenario of verifying escaped chars are not written to the file system.
         Assert.AreEqual(expectedRuntimeConfigJson.ToString(), actualRuntimeConfigJson.ToString());
     }
 
--- a/src/Cli.Tests/ConfigGeneratorTests.cs
+++ b/src/Cli.Tests/ConfigGeneratorTests.cs
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Text;
+using System.Text.RegularExpressions;
+
 namespace Cli.Tests;
 
 /// <summary>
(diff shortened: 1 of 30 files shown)

### CODE REVIEW

File: src/Cli.Tests/ConfigGeneratorTests.cs [hunk 1]
```diff
@@ -119,6 +121,73 @@ public void TestCleanup()
...
+            }");
+
+        expectedRuntimeConfigJson = expectedRuntimeConfigJson.Replace("\n", string.Empty);
+        expectedRuntimeConfigJson = expectedRuntimeConfigJson.Replace(" ", string.Empty);
+
+        Assert.IsTrue(TryGenerateConfig(options, _runtimeConfigLoader!, _fileSystem!));
+
+        StringBuilder actualRuntimeConfigJson = new(_fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE, Encoding.Default));
+        actualRuntimeConfigJson = actualRuntimeConfigJson.Replace(" ", string.Empty);
+        actualRuntimeConfigJson = actualRuntimeConfigJson.Replace("\r\n", string.Empty);
```
seantleonard, line 186: Should expected runtime config also replace `"\r\n"`?

seantleonard (approved):
One question otherwise, lgtm!

src/Cli.Tests/ConfigGeneratorTests.cs
abhishekkumams: Comparing JSON as a string can be problematic because it does not take into account differences in formatting or ordering of the JSON properties. This can lead to false negatives in tests. I have updated it to use JToken.DeepEquals,  it compares the JSON objects in a more robust way that takes into account differences in formatting and ordering.

abhishekkumams (approved):

src/Cli.Tests/ConfigGeneratorTests.cs
Aniruddh25: Parsing it as JSON objects loses the test scenario - as I mentioned in the PR description as well:

> Note, this test is modified to explicitly compare the expected and actual json string from file since any usage of JsonSerializer/JObject.Parse would require using the same encoder to deserialize making the test itself run similar code which it is testing.


If the test failing, replacing `\r\n` might also be needed to consider platform differences.

PR #1697, opened 2023-09-11
Fix $select clause behavior for Find Requests

### LINKED ISSUE #1505: REST - Using $select always returns the primary key fields and $orderby fields in the response
For the below GET request ,

```http
GET https://localhost:5001/api/Book?$select=title
```
the response contains the `id` field as well.

Response:

```json
{
    "value": [
        {
            "title": "New title #1",
            "id": 1
        },
        {
            "title": "Also Awesome book",
            "id": 2
        },
        {
            "title": "Great wall of china explained",
            "id": 3
        },
        ...
    ]
}
```

For the below GET request ,

```http
GET https://localhost:5001/api/Book?$select=id,publisher_id&$orderby=name
```
the response contains the `name` field as well.

Response:

```json

{
    "value": [
        {
            "id": 2,
            "publisher_id": 1234,
            "name": "Also Awesome book"
        },
       ...
       ]
}    
```

severussundar:
## Why make this change?

- Closes #1505 
  - Responses for Find requests with `$select` query string fields in addition to those requested in `$select`

For the below GET request,
```http
GET https://localhost:5001/api/Book?$select=title
```
the response contains the `id` field as well. If a table contains composite PK, all the Pks are returned in the response.

Response:

```json
{
    "value": [
        {
            "title": "New title #1",
            "id": 1
        },
        {
            "title": "Also Awesome book",
            "id": 2
        },
        {
            "title": "Great wall of china explained",
            "id": 3
        },
        ...
    ]
}

```

When both $select and $orderby are used together, the response contains $orderby fields in addition to the ones in $select

```http
GET https://localhost:5001/api/Book?$select=id,publisher_id&$orderby=name
```
the response contains the `name` field as well.

Response:

```json

{
    "value": [
        {
            "id": 2,
            "publisher_id": 1234,
            "name": "Also Awesome book"
        },
       ...
       ]
}    
```
Similar behavior was observed with views, where the configured key-fields were returned in the response. If multiple key-fields are configured, all of them were returned.

image

## What is this change?

- To support `pagination` and `$first` with Find requests, it is necessary to provide the `nextLink` field in the response. For the calculation of `nextLink`, the fields such as primary keys, fields in `$orderby` clause are retrieved from the database in addition to the fields requested in the $select clause. The `nextLink` calculation is done at DAB layer after executing the database query because it is impossible to know beforehand whether a table/view would contain more than 100 records (or more entries than requested in the `$first` clause if that is applicable).

- Seeing that these additional fields needs to be retrieved from the database, the database queries cannot be modified to select only the fields present in `$select` clause. So, the additional fields are removed at the DAB layer before returning the response.

## Why can't the database queries by modifed to select only the fields present in `$select` query string?
When the table contains more than 100 records, the nextLink URL field is returned in the response. The nextLink URL contains $after which is a base 64 encoded value. $after is calculated using all the primary key and $orderby column values of the 100th record (or nth when $first=n is used). 

The row values from the 100th (or nth when $first=n is used) record is necessary because of two reasons

a) The last record's values are necessary to correctly fetch the subsequent set of records. Let's say the first GET request fetches 100 records. To be able to fetch the next set of rows: 101-200 in the right ordering, the database query will include the condition ~ `WHERE ( ( [orderby_column] > [orderby_value_of_row100] ) OR ( [orderby_column] == [orderby_value_of_row100] && [pk_column] > [pk_value_of_row100])`. Here `orderby_value_of_row100` and `pk_value_of_row100` refers to the values of orderby column of the 100th row and pk column of the 100th row respectively.
b) $orderby can be performed in ASC or DESC order. The 100th record will be different in both the ordering and we rely on the value of the 100th row to be able to select the next set of rows (101-200, etc.) 

Logic for $after field calculation: [SqlPaginationUtil.MakeCursorFromJsonElement()](https://github.com/Azure/data-api-builder/blob/7b4c8b5606012d6c9e8093a1ed9b63e5116faf75/src/Core/Resolvers/SqlPaginationUtil.cs#L123C2-L123C2) 

Granted, all of this is applicable only when nextLink field is necessary in the response, it is not possible to know beforehand whether a table contains more than 100 records ( or n records when first n records is selected using $first clause). We select 101 records (or n+1 records for $first=n) and then determine at DAB layer if nextLink field is necessary. If the database query returns 101 (or n+1 records), then we conclude that a nextLink field is necessary. Essentailly, not being able to know beforehand the number of records in the table forces to select the additional fields.

## How was this tested?

- [x] Integration Tests
- [x] Manual Tests

## Sample Request(s)

- Primary Key fields are not returned in the response when absent in $select query string
image

- Orderby fields are not returned in the response
image


src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -186,20 +187,34 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
         {
             JsonElement jsonElement = jsonDoc.RootElement.Clone();
 
+            // When there are no rows returned from the database, the jsonElement will be an empty array.
+            // In that case, the response is returned as is.
+            if (jsonElement.ValueKind is JsonValueKind.Array && jsonElement.EnumerateArray().Count() == 0)
```
seantleonard, line 192: instead of `jsonElement.EnumerateArray().Count() == 0` could you do something like `!jsonElement.Any()`? if the count >0 then you're enumerating the whole array unnecessarily.

seantleonard:
In your pr description, you state: 
> Seeing that these additional fields needs to be retrieved from the database, the database queries cannot be modified to select only the fields present in $select clause. 

Why can't the db queries be modified? for example
```http
GET localhost/api/Author?$select=name$orderby=id desc
```

can translate into:
```tsql
SELECT [name]
  FROM [dbo].[authors]
  ORDER BY [id] desc
```

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
seantleonard, line 1: for your net new tests, it would be easier to read if you added a comment beside each test describing what you are validating

src/Service.Tests/SqlTests/RestApiTests/Find/MsSqlFindApiTests.cs
```diff
@@ -462,6 +462,80 @@ public class MsSqlFindApiTests : FindApiTestBase
                 $"SELECT [id], [category], [revenue], [accessible_role] FROM { _tableWithSecurityPolicy } " +
                 $"WHERE [id] = 3 " +
                 $"FOR JSON PATH, INCLUDE_NULL_VALUES, WITHOUT_ARRAY_WRAPPER"
+            },
+            {
+                "FindByIdTestWithSelectFieldsOnViewWithoutKeyFields",
+                $"SELECT [title] FROM { _simple_all_books } " +
+                $"WHERE [id] = 1 FOR JSON PATH, INCLUDE_NULL_VALUES, WITHOUT_ARRAY_WRAPPER"
```
seantleonard, line 469: It would be easier to read if you bump this segment to a new line: (this comment applies to multiple tests below this comment, and makes test entries consisten with the tests already present.)
`FOR JSON PATH, INCLUDE_NULL_VALUES, WITHOUT_ARRAY_WRAPPER`

src/Service.Tests/SqlTests/RestApiTests/Find/MySqlFindApiTests.cs
```diff
@@ -414,7 +414,7 @@ LIMIT 100
             {
                 "FindTestWithPrimaryKeyContainingForeignKey",
                 @"
-                    SELECT JSON_OBJECT('id', id, 'book_id', book_id, 'content', content) AS data
+                    SELECT JSON_OBJECT('id', id, 'content', content) AS data
```
seantleonard, line 417: two lines below this there is the statement:
```sql
SELECT id, book_id, content
```
Would it be valid to also remove book_id from the select statement here as you did for this outer select?

src/Service.Tests/SqlTests/RestApiTests/Find/MySqlFindApiTests.cs
```diff
@@ -737,7 +737,7 @@ LIMIT 1
             {
                 "FindTestWithSingleMappedFieldsToBeReturned",
                 @"
-                  SELECT JSON_ARRAYAGG(JSON_OBJECT('treeId', treeId, 'Scientific Name', species)) AS data
+                  SELECT JSON_ARRAYAGG(JSON_OBJECT('Scientific Name', species)) AS data
```
seantleonard, line 740: Below, can `SELECT treeId, species` be modified to exclude the field `treeId`?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -233,8 +248,87 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+            }
+            else
+            {
+                foreach (JsonProperty property in response.EnumerateArray().Last().EnumerateObject())
+                {
+                    fieldsPresentInResponse.Add(property.Name);
+                }
+            }
+
+            return fieldsPresentInResponse.Except(fieldsToBeReturned);
```
seantleonard, line 290:
```suggestion
            return fieldsPresentInResponse.Except(context.FieldsToBeReturned);
``` 

would directly using `context.FieldsToBeReturned` allow you to avoid the extra hashset allocation?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -233,8 +248,87 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+        /// <param name="jsonElement"> Json Element with extra fields</param>
+        /// <param name="extraFields">Additional fields that needs to be removed from the Json element</param>
+        /// <returns>Json Element after removing the additional fields</returns>
+        private static JsonElement RemoveExtraFieldsInResponseWithSingleItem(JsonElement jsonElement, IEnumerable<string> extraFields)
+        {
+            JsonObject? jsonObject = JsonObject.Create(jsonElement);
+
+            if (jsonObject is null)
+            {
+                throw new DataApiBuilderException(message: "While processing your request the server ran into an unexpected error",
```
seantleonard, line 321: nit: message param to new line

seantleonard (changes requested):
I'm not following why we have to do all the introduced post processing when the SQL query could be modified to return the correct results (which exclude the pk fields/select/orderby clause fields.

That would be helpful to have in the pr description. In your tests, you make adjustments to the validation queries to address this ask, so why not in the actual engine code?

severussundar:
> I'm not following why we have to do all the introduced post processing when the SQL query could be modified to return the correct results (which exclude the pk fields/select/orderby clause fields.
> 
> That would be helpful to have in the pr description. In your tests, you make adjustments to the validation queries to address this ask, so why not in the actual engine code?

When the table contains more than 100 records, the nextLink URL field is returned in the response. The nextLink URL contains $after which is a base 64 encoded value. $after is calculated using all the primary key and $orderby column values of the 100th record (or nth when $first=n is used). 

The row values from the 100th (or nth when $first=n is used) record is necessary because of two reasons

a) The last record's values are necessary to correctly fetch the subsequent set of records. Let's say the first GET request fetches 100 records. To be able to fetch the next set of rows: 101-200 in the right ordering, the database query will include the condition ~ `WHERE ( ( [orderby_column] > [orderby_value_of_row100] ) OR ( [orderby_column] == [orderby_value_of_row100] && [pk_column] > [pk_value_of_row100])`. Here `orderby_value_of_row100` and `pk_value_of_row100` refers to the values of orderby column of the 100th row and pk column of the 100th row respectively.
b) $orderby can be performed in ASC or DESC order. The 100th record will be different in both the ordering and we rely on the value of the 100th row to be able to select the next set of rows (101-200, etc.) 

Logic for $after field calculation: [SqlPaginationUtil.MakeCursorFromJsonElement()](https://github.com/Azure/data-api-builder/blob/7b4c8b5606012d6c9e8093a1ed9b63e5116faf75/src/Core/Resolvers/SqlPaginationUtil.cs#L123C2-L123C2) 

Granted, all of this is applicable only when nextLink field is necessary in the response, it is not possible to know beforehand whether a table contains more than 100 records ( or n records when first n records is selected using $first clause). We select 101 records (or n+1 records for $first=n) and then determine at DAB layer if nextLink field is necessary. If the database query returns 101 (or n+1 records), then we conclude that a nextLink field is necessary. Essentailly, not being able to know beforehand the number of records in the table forces to select the additional fields.


>In your tests, you make adjustments to the validation queries to address this ask, so why not in the actual engine code?

The tests primarily validate that no extra columns in addition to those requested through `$select` are returned in the response. To construct the expected response, I've used database queries to pick the necessary fields. These are not the exact database queries that DAB generates for the corresponding Find API requests. And, importantly, they cannot be the same queries as DAB selects additional fields to ensure correct ordering [The earlier section explains this in greater detail]. Since, there is no change in the `nextLink` generation logic, no new tests are added to validate that. The existing pagination and `$first` clause tests validate the following.
a) `nextLink` field is returned in the response when necessary
b) The correctness of the `nextLink` field.

src/Core/Resolvers/SqlQueryEngine.cs
severussundar: Good catch. I've updated to use `jsonElement.GetArrayLength() == 0`

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -186,20 +187,34 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
             if (jsonElement.ValueKind is not JsonValueKind.Array || !SqlPaginationUtil.HasNext(jsonElement, context.First))
             {
-                // Clones the root element to a new JsonElement that can be
-                // safely stored beyond the lifetime of the original JsonDocument.
-                return OkResponse(jsonElement);
+                return jsonElement.ValueKind is JsonValueKind.Array ? OkResponse(JsonSerializer.SerializeToElement(RemoveExtraFieldsInResponseWithMultipleItems(jsonElement.EnumerateArray().ToList(), extraFieldsInResponse)))
+                                                                    : OkResponse(RemoveExtraFieldsInResponseWithSingleItem(jsonElement, extraFieldsInResponse));
             }
 
+            IEnumerable<JsonElement> rootEnumerated = jsonElement.EnumerateArray().ToList();
```
seantleonard, line 209: Why do you need ToList() if you are using IEnumerable<T>? The original code did not use ToList()

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -233,8 +248,88 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+        /// To support pagination and $first clause with Find requests, it is necessary to provide the nextLink
+        /// field in the response. For the calculation of nextLink, the fields such as primary keys, fields in $orderby clause
+        /// are retrieved from the database in addition to the fields requested in the $select clause.
+        /// However, these fields are not required in the response.
+        /// This function helps to determine those additional fields that are present in the response.
+        /// </summary>
+        /// <param name="response">Response json retrieved from the database</param>
+        /// <param name="context">FindRequestContext for the GET request.</param>
+        /// <returns>Additional fields that are present in the response</returns>
+        private static IEnumerable<string> DetermineExtraFieldsInResponse(JsonElement response, FindRequestContext context)
```
seantleonard, line 266: will there always be extra fields in a response when the count of returned items is >= 1? Is there no way to short circuit some of these loops if no $first, $after, $select query string params exist?

src/Core/Resolvers/SqlQueryEngine.cs
severussundar: Removed it

seantleonard:
I have updated the nuget feed so that it can pull the correct version of packages.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -186,20 +187,45 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+                else
+                {
+                    return jsonElement.ValueKind is JsonValueKind.Array ? OkResponse(JsonSerializer.SerializeToElement(RemoveExtraFieldsInResponseWithMultipleItems(jsonElement.EnumerateArray().ToList(), extraFieldsInResponse)))
+                                                                        : OkResponse(RemoveExtraFieldsInResponseWithSingleItem(jsonElement, extraFieldsInResponse));
+                }
             }
 
-            // More records exist than requested, we know this by requesting 1 extra record,
-            // that extra record is removed here.
             IEnumerable<JsonElement> rootEnumerated = jsonElement.EnumerateArray();
```
seantleonard, line 220: Would be better to just have this as a list since .Count() and .Last() and both traverse the enumerable.

severussundar: I've updated to use List and am performing index based access which is O(1). [List docs](https://learn.microsoft.com/dotnet/api/system.collections.generic.list-1.item?#remarks)

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -186,22 +187,47 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+                                                                        : OkResponse(RemoveExtraFieldsInResponseWithSingleItem(jsonElement, extraFieldsInResponse));
+                }
             }
 
+            List<JsonElement> rootEnumerated = jsonElement.EnumerateArray().ToList();
+
             // More records exist than requested, we know this by requesting 1 extra record,
             // that extra record is removed here.
-            IEnumerable<JsonElement> rootEnumerated = jsonElement.EnumerateArray();
+            rootEnumerated = rootEnumerated.Take(rootEnumerated.Count - 1).ToList();
```
seantleonard, line 224: why not just remove the last element in the list with `Remove(rootEnumerated.Count - 1)`? 

Currently, you're enumerating `rootEnumerated.Count - 1` elements, and then creating a brand new list to then assign back to `rootEnumerated`.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -186,22 +187,47 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
+            if (jsonElement.ValueKind is JsonValueKind.Array && jsonElement.GetArrayLength() == 0)
+            {
+                return OkResponse(jsonElement);
+            }
+
+            HashSet<string> extraFieldsInResponse = (jsonElement.ValueKind is not JsonValueKind.Array)
+                                                  ? DetermineExtraFieldsInResponse(jsonElement, context)
+                                                  : DetermineExtraFieldsInResponse(jsonElement.EnumerateArray().First(), context);
+
+            int countOfExtraFieldsInResponse = extraFieldsInResponse.Count;
```
seantleonard, line 201: Do you need to have a separate variable for Count?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -233,8 +257,83 @@ private OkObjectResult FormatFindResult(JsonDocument jsonDoc, FindRequestContext
...
-            rootEnumerated = rootEnumerated.Append(nextLink);
-            return OkResponse(JsonSerializer.SerializeToElement(rootEnumerated));
+
+            // When there are extra fields present, they are removed before returning the response.
+            if (extraFieldsInResponse.Count > 0)
+            {
+                rootEnumerated = RemoveExtraFieldsInResponseWithMultipleItems(rootEnumerated, extraFieldsInResponse);
+            }
+
+            return OkResponse(JsonSerializer.SerializeToElement(rootEnumerated.Append(nextLink)));
```
seantleonard, line 267: Now that `rootEnumerated` is a list, you can use `Add()`. Using `Append()` creates a new list: https://stackoverflow.com/a/52355756/18174950

seantleonard (approved):
One last nit on List usage. Otherwise, thanks for making these changes!

abhishekkumams (approved):

PR #1726, opened 2023-09-18
Hot Reload of Runtime Section

aaronburtle:
## Why make this change?

Related to, covering the runtime section of the config only https://github.com/Azure/data-api-builder/issues/67

Because we are not able to make easy use of the `IOptionsMonitor` which would have allowed for the updating of individual sections of the config automatically (ie: just refreshing the `runtime options`), we monitor and reload the entire file. However, we currently do not have support for reloading the sections outside of the `Runtime` section of the config. Because of this, we will not document and announce this feature until the other sections can be safely reloaded.

Additionally, certain options are used to configure services at startup, and despite refreshing the `RuntimeConfig`, those services will still be configured with the values from initialization on startup. This means adding a mechanism for refreshing those services to make use of the updated options after a Hot Reload, and that work will come in a future PR.

## What is this change?

Adds a config file watcher during startup which will monitor for any file changes to the config and then trigger an update event in the case that the file is changed.

We instantiate the config file watcher when the runtime provider is first asked to get a config. We then register a method owned by the runtime config provider that will be called back to in the case that we wish to hot reload. However, we only do this for when we start the service in local development. So that means when we are `not` late configured, and we are in a Host Mode of Development on startup. In other cases we do not watch the config file for changes, and do not make this callback in the case that the config file does change.

To get the location of the file that we are going to watch we use the config loader's file system and stored data regarding the config file. 

## How was this tested?

Unit testing was added to verify correct hot reloading of the runtime options that we intend to have changing.

## Sample Request(s)

Any valid request after the change ought to work.


abhishekkumams:
Some pointers:
Add a Logger to Log Information saying `Found changes in the config: Executing Hot Reload`, you can add screenshot for the same from manual testing.

src/Service/Startup.cs
```diff
@@ -153,8 +160,11 @@ public void ConfigureServices(IServiceCollection services)
 
             //Enable accessing HttpContext in RestService to get ClaimsPrincipal.
             services.AddHttpContextAccessor();
-
-            ConfigureAuthentication(services, configProvider);
+            services.AddTransient(implementationFactory: (serviceProvider) =>
```
abhishekkumams, line 163: interesting, what does `AddTransient` do?

src/Service/Startup.cs
```diff
@@ -153,8 +160,11 @@ public void ConfigureServices(IServiceCollection services)
...
+            {
+                ConfigureAuthentication(services, serviceProvider.GetRequiredService<RuntimeConfigProvider>());
+                return string.Empty;
```
abhishekkumams, line 166: is it giving any error, if we don't return anything?

abhishekkumams:
Also if possible can you add logs to show what happens when file is modified.

src/Service/Startup.cs
```diff
@@ -153,8 +160,11 @@ public void ConfigureServices(IServiceCollection services)
 
             //Enable accessing HttpContext in RestService to get ClaimsPrincipal.
             services.AddHttpContextAccessor();
-
-            ConfigureAuthentication(services, configProvider);
+            services.AddTransient(implementationFactory: (serviceProvider) =>
+            {
+                ConfigureAuthentication(services, serviceProvider.GetRequiredService<RuntimeConfigProvider>());
```
seantleonard, line 165: I don't understand why this is wrapped in a `services.Add..()` block

src/Service/Startup.cs
seantleonard: I think this section needs to be approached differently. I think the intent is to react to a new config file after startup and configure services (and consequently the DI container) is already created. 

This PR aims to address hot reloading the runtime section of the runtime config. So that means any new config which changes up the authentication mechanism, will need to have a different authentication scheme configured. 

Due to how authentication is configured only at startup, we only have one chance to configure it. I think we'll need to refactor the authN setup section to register all authentication schema unconditionally and allow the mechanism which "authenticate" a request, to dynamically determine which authN scheme to choose based on runtime configuration settings. 

```csharp
services.AddAuthentication("<no default or default SWA>")
    .AddEasyAuthAuthentication(easyAuthAuthenticationProvider: EasyAuthType.AppService)
    .AddEasyAuthAuthentication(easyAuthAuthenticationProvider: EasyAuthType.StaticWebApps)
    .AddSimulatorAuthentication()
    .AddJwtBearer("<use ioptionsmonitor here because jwt settings may change in new config files>");
```

To then register changes to jwtbeareroptions, this asp.net repo thread provides suggestions on how to monitor the jwt options changes and signal that to the underlying authentication config: https://github.com/dotnet/aspnetcore/issues/49586#issuecomment-1671838595

seantleonard (changes requested):

src/Service/Startup.cs
seantleonard: Then in `ClientRoleHeaderAuthenticationMiddleware.InvokeAsync(...)` all calls to `httpContext.AuthenticateAsync(..)` and `httpContext.ChallengeAsync(...)` need to use the method signature that requires you to pass in the authetnication scheme. The authentication scheme can be dynamically resolved by helper methods on the runtime config/provider class.

src/Service/Startup.cs
aaronburtle: this is old code, a new approach was taken, but to answer your question, it was to get access to the service provider.

src/Service/Startup.cs
aaronburtle: i see what you mean with how we do the authentication configuring only at startup. Let's sync up on this one, and thanks for pointing this out!

aaronburtle: Just to update the decisions here. After syncing with Sean, we've determined a plan for updating the authentication options, and this new design will be added to the related design document, and then come in the future iterations on hot reload.

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+            EnableRaisingEvents = true
+        };
+
+        _configProvider = configProvider;
+        _fileWatcher.Changed += OnConfigFileChange;
+    }
+
+    public ConfigFileWatcher() { }
+
+    private void OnConfigFileChange(object sender, FileSystemEventArgs e)
```
abhishekkumams, line 36: please add method summary

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            fileSystem.File.WriteAllText(configName, initialConfig);
+            FileSystemRuntimeConfigLoader configLoader = new(fileSystem, configName, string.Empty);
+            RuntimeConfigProvider configProvider = new(configLoader);
+            // Must GetConfig() to start file watching
+            RuntimeConfig runtimeConfig = configProvider.GetConfig();
+            Assert.IsNotNull(runtimeConfig);
+            // Simulate change to the config file
+            fileSystem.File.WriteAllText(configName, updatedConfig);
+            // Give file watcher enough time to see the change
+            System.Threading.Thread.Sleep(1000);
```
abhishekkumams, line 128: is it the recomended sleep time for file watcher? what if this is not enough?

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            // Hot Reloaded config
+            runtimeConfig = configProvider.GetConfig();
+            Assert.AreEqual(updatedRestEnabled, runtimeConfig.Runtime.Rest.Enabled);
+            Assert.AreEqual(updatedRestPath, runtimeConfig.Runtime.Rest.Path);
+            Assert.AreEqual(updatedGQLEnabled, runtimeConfig.Runtime.GraphQL.Enabled);
+            Assert.AreEqual(updatedGQLPath, runtimeConfig.Runtime.GraphQL.Path);
+            Assert.AreEqual(updatedGQLIntrospection, runtimeConfig.Runtime.GraphQL.AllowIntrospection);
+            Assert.AreEqual(updatedMode, runtimeConfig.Runtime.Host.Mode);
```
abhishekkumams, line 136: this asserts tells us that config file is changed.
Can we do something like hitting a rest endpoint before hot reload and then change the endpoint and then try to hit the updated endpoint to show it's updated.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
aaronburtle: there isn't really a recommended sleep time for this case, but 1000ms is well more than enough time to pick up the change and it doesnt significantly affect the time to complete the test.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
aaronburtle: that's a good idea for integration testing, ill figure out a way to add some tests like that

aaronpowell:
I was pondering on how to tackle the idea of hot reload when I reworked the config system.

I thought to tackle it slightly differently, that we would create an additional `ConfigLoader` type, most likely a new class that inherits `FileSystemRuntimeConfigLoader` and then runs the watcher in itself. This would mean that you don't need to expose internal stuff, such as the `IFileSystem` out (and have casts from `ConfigLoader`), and removes the need for any part of the application to be aware that there is a watcher as it just sees a `ConfigLoader` being passed around.

Then within the `Startup.cs` when we create the instance of the loader there is conditional logic to whether we want to have the "watcher version" or not created (and if, in the future there are other loader types we continue the branching logic).

aaronburtle:
> I was pondering on how to tackle the idea of hot reload when I reworked the config system.
> 
> I thought to tackle it slightly differently, that we would create an additional `ConfigLoader` type, most likely a new class that inherits `FileSystemRuntimeConfigLoader` and then runs the watcher in itself. This would mean that you don't need to expose internal stuff, such as the `IFileSystem` out (and have casts from `ConfigLoader`), and removes the need for any part of the application to be aware that there is a watcher as it just sees a `ConfigLoader` being passed around.
> 
> Then within the `Startup.cs` when we create the instance of the loader there is conditional logic to whether we want to have the "watcher version" or not created (and if, in the future there are other loader types we continue the branching logic).

I sent you a message on team to follow up on this, but just to summarize briefly here for the comment chain. This was my initial thought too, but I ran into a circular reference trying to put the file watcher in the config loader, because we need to referene the RuntimeConfigProvider from the file watcher, but the runtime config provider already references the config loader, causing the circular reference problem.

aaronpowell:
Ah yes, a circular reference problem as the loader is stateless, so how would the provider know about a change to the file it's monitoring.

Thinking about it a bit more (and refreshing my memory with the code) I wonder if an event system might not be the best solution and there's some more design in the pattern that needs to be thought through. Here's an interesting observation - when using DAB in the hosted model like SWA does, it creates the `FileSystemRuntimeConfigLoader` type as the loader, even though there's no file on disk, it's provided via the HTTP endpoint. This does mean that we have to have some defensive coding for when the file doesn't exist, which it won't in that case.

So in this case as well we basically ignore the loader, but it could be where an event model makes sense.

Config received from external source (FS watcher notice, Controller Action) -> Loader event of "new config" raised -> Loader event of "refresh config" raised -> Provider updates its internal reference

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
aaronburtle: integration testing introduces some new challenges with race conditions. I have opened an issue to track adding these tests, they will come in a future PR.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -159,7 +159,7 @@ public RuntimeConfig(string? Schema, DataSource DataSource, RuntimeEntities Enti
         this.DataSource = DataSource;
         this.Runtime = Runtime;
         this.Entities = Entities;
-        _defaultDataSourceName = Guid.NewGuid().ToString();
+        _defaultDataSourceName = "DEFAULT_DATASOURCE_NAME";
```
aaronburtle, line 162: Changed this to a constant value so that we don't have to save the unique ID for reloads. We have correlation ID to track individual requests already, not sure having unique UIDs here grants us any advantage.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            HostMode initialMode = HostMode.Development;
+            HostMode updatedMode = HostMode.Production;
+
+            string initialConfig = @"
+{
+  ""$schema"": ""https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json"",
+  ""data-source"": {
+    ""database-type"": ""mssql"",
+    ""connection-string"": ""Server=test;Database=test;User ID=test;Password=test;"",
```
Aniruddh25, line 47: Provide `placeholder` as the password value if it absolutely needs to be set

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                File.Delete(configName);
+            }
+
+            FileSystem fileSystem = new();
+            fileSystem.File.WriteAllText(configName, initialConfig);
+            FileSystemRuntimeConfigLoader configLoader = new(fileSystem, configName, string.Empty);
+            RuntimeConfigProvider configProvider = new(configLoader);
+            // Must GetConfig() to start file watching
+            RuntimeConfig runtimeConfig = configProvider.GetConfig();
```
Aniruddh25, line 123: Perhpas, good to assert on the initial values as well before changing

Aniruddh25 (changes requested):
Good starter, need some clarifications on the ConfigFilePath being watched

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -29,7 +29,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     // or user provided config file which could be a relative file path, absolute file path or simply the file name assumed to be in current directory.
     private string _baseConfigFilePath;
 
-    private readonly IFileSystem _fileSystem;
+    public readonly IFileSystem _fileSystem;
```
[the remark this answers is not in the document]
aaronburtle: I use the file system inside of the config file watcher to make sure that we are monitoring the correct directory and file.

src/Service/Startup.cs
```diff
@@ -80,9 +80,8 @@ public void ConfigureServices(IServiceCollection services)
             IFileSystem fileSystem = new FileSystem();
             FileSystemRuntimeConfigLoader configLoader = new(fileSystem, configFileName, connectionString);
             RuntimeConfigProvider configProvider = new(configLoader);
-
-            services.AddSingleton(fileSystem);
             services.AddSingleton(configProvider);
+            services.AddSingleton(fileSystem);
```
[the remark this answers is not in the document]
aaronburtle: nope, this just reordered during a refactor, ill change it back to avoid unnecessary changes

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -44,6 +44,8 @@ public class RuntimeConfigProvider
 
     public RuntimeConfigLoader ConfigLoader { get; private set; }
 
+    private ConfigFileWatcher? ConfigFileWatcher { get; set; }
```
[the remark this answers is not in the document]
aaronburtle: a field here is fine

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
aaronburtle: ill simply remove

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+
+public class ConfigFileWatcher
+{
+    private FileSystemWatcher? _fileWatcher;
+    RuntimeConfigProvider? _configProvider;
+
+    public ConfigFileWatcher(RuntimeConfigProvider configProvider)
+    {
+        FileSystemRuntimeConfigLoader loader = (FileSystemRuntimeConfigLoader)configProvider.ConfigLoader;
+        string configFileName = loader.ConfigFilePath;
```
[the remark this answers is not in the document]
aaronburtle: it is, I was just used to only seeing the name when running locally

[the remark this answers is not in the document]
aaronburtle: it can either be a simple file name in the case where the config is already in the current directory you can just supply the name, or it can be a complete path. I will update the code to handle both cases correctly.

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+        IFileSystem fileSystem = (IFileSystem)loader._fileSystem;
+        string? currentDirectoryPath = fileSystem.Directory.GetCurrentDirectory();
+        string configFilePath = Path.Combine(currentDirectoryPath!, configFileName);
```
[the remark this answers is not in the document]
aaronburtle: right, I need to be more careful with how this is combined, will fix.

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+        string path = Path.GetDirectoryName(configFilePath)!;
```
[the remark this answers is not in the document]
aaronburtle: old code Ive removed, wasnt being used for anything.

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+        _fileWatcher = new FileSystemWatcher(Path.GetDirectoryName(configFilePath)!)
```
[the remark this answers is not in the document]
aaronburtle: right, we should only get the current directory in the case that the complete path isnt provided and the config is simply in the current directory.

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            Filter = Path.GetFileName(configFilePath),
```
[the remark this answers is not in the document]
aaronburtle: it can be either, so I will see if we can get a directory from the path provided, in which case we want to use that directory and not do any additional combinations. if no directly is provided than we should get the current directory and combine.

seantleonard (changes requested):
some design change suggestions for improved readability and code intent.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -236,7 +236,7 @@ public bool DoesFileExistInDirectory(string filePath)
     /// </summary>
     public override string GetPublishedDraftSchemaLink()
     {
-        string? assemblyDirectory = _fileSystem.Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+        string? assemblyDirectory = FileSystem.Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
```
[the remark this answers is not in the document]
aaronburtle: we don't mock the file system for the tests with hot reload. this change was just a rename from the public/private naming conventions.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -79,9 +81,31 @@ public RuntimeConfig GetConfig()
                 subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
         }
 
+        CheckForAndSetupConfigFileWatcher();
```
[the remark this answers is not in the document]
aaronburtle: I thought about adding `Try` to the method name, but this implies that it will return a bool and I wanted to keep it void. I suppose we can return a bool that isn't being used here for clarity to the method name.

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    public ConfigFileWatcher(RuntimeConfigProvider configProvider)
+    {
+        FileSystemRuntimeConfigLoader loader = (FileSystemRuntimeConfigLoader)configProvider.ConfigLoader;
+        string configFilePath = loader.ConfigFilePath;
+        IFileSystem fileSystem = (IFileSystem)loader.FileSystem;
+        string? directoryName = Path.GetDirectoryName(configFilePath);
+        directoryName = string.IsNullOrWhiteSpace(directoryName) ?
+            fileSystem.Directory.GetCurrentDirectory() :
+            directoryName;
```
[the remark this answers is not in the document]
aaronburtle: We want that `path` to come from the FileSystem here. And when we invoke the file watcher's constructor, we only have access to the loader since we are inside of the `RuntimeConfigProvider`. How are you proposing we get the `path` that we need here if not from the `FileSystem` contained in the loader?

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            Assert.AreEqual(initialRestEnabled, runtimeConfig.Runtime.Rest.Enabled);
+            Assert.AreEqual(initialRestPath, runtimeConfig.Runtime.Rest.Path);
+            Assert.AreEqual(initialGQLEnabled, runtimeConfig.Runtime.GraphQL.Enabled);
+            Assert.AreEqual(initialGQLPath, runtimeConfig.Runtime.GraphQL.Path);
+            Assert.AreEqual(initialGQLIntrospection, runtimeConfig.Runtime.GraphQL.AllowIntrospection);
+            Assert.AreEqual(initialMode, runtimeConfig.Runtime.Host.Mode);
+            // Simulate change to the config file
+            fileSystem.File.WriteAllText(configName, updatedConfig);
+            // Give file watcher enough time to see the change
+            System.Threading.Thread.Sleep(1000);
```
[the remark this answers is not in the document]
aaronburtle: I am not sure what you mean here. We have a FileWatcher that will raise an event when the config file changes, but that is within the config file watcher class. Do you mean to add another file watcher inside of this test that then can raise another event for us to wait on?

I am not sure this is worth the effort because we still need to thread.sleep in order to give the hot reload enough time to take place. We could in theory have an event that gets signaled when the hot reload is finished, but that would be an event within our config file watcher code that we signal only for testing.

It seems like for a simple unit test, simply sleeping for 1 second is a more natural approach to me.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            // Hot Reloaded config
+            runtimeConfig = configProvider.GetConfig();
```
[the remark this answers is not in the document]
aaronburtle: I think it makes more sense to have this coverage as part of the integration testing. This unit test is really just for testing the unit of code that is doing the actual hot reloading itself. In this case we'd just be testing the same unit of code twice but it would be triggered from two different locations. With integration testing we can cover all of these unique code paths and make sure that they all work together. The integration tests depend on covering some additional scenarios that will be in with future work (multiple config files in particular).

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (!_configProvider!.IsLateConfigured && _configProvider!.GetConfig().IsDevelopmentMode())
+            {
+                _configProvider!.HotReloadConfig();
+            }
+        }
+        catch (Exception ex)
+        {
+            // Need to remove the dependency configuring authentication has on the RuntimeConfigProvider
+            // before we can have an ILogger here.
+            Console.WriteLine("Unable to Hot Reload configuration file due to " + ex.Message);
```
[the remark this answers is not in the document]
aaronburtle: In this case you will need to get the config back into a healthy state. So as long as you are trying to hot reload or otherwise load in an unhealthy configuration, you will run into errors with loading the runtime config and get messages stating as such. However, you can recover from this simply by providing a health configuration file, and every way that I have been able to break the hot reloading by getting an exception to trigger has had pretty clear error messaging.

But it is true that if you hot reload with a bad config file, the service will not work until you provide something to get the states back to be consistent and healthy. This is something that was considered early on, and we decided to just let people hot reload as they desire, and then they can reset the state, rather than trying to see if we can reload and if not return it to the last healthy state for them. So this was a choice that we made intentionally.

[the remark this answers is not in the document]
aaronburtle: Just to clarify what you'd see with an example:

Let's say I have a healthy config but then I do a bad hot reload. Suppose I put in some invalid characters so that we throw an exception when we try to parse the config during the hot reload call.

This will fail and I will have error messages that relate to the way that the loading of the config during hot reload went bad.  Then if I send a request, since the runtime config isn't setup correctly, it will try to set it up again, but since we havent fixed the broken config, it will fail and you will get an error in the response to your request.

{
    "error": {
        "code": "ErrorInInitialization",
        "message": "Runtime config isn't setup.",
        "status": 503
    }
}

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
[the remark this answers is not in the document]
seantleonard: My intent was to mitigate the need to have an explicit thread.sleep. While it is 1 second and 1 test now, that wouldn't really be scalable if more tests had to do that. 

The current comment you have above the sleep code indicates that the wait is required because it may take time for the file changed event to be fired and received by the file watcher. 

If that's not the case and the wait is because it takes a certain amount of time to complete `hotReload()` related activities like recreating the runtime config object, then the comment should be updated.

My suggestion was to have another event listener in this test that listens for the file changed event to be fired by the filesystem, assuming the thread.sleep was used because of latency between filechanged notification being fired and filewatcher processing that notification. In other words, "wait for the filechanged notification and right when we detect that, begin our asserts" Ultimately, that doesn't address your latest point of the time required to finish `hotreload()`

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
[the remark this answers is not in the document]
seantleonard: sounds good.

src/Core/Configurations/ConfigFileWatcher.cs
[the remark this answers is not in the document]
aaronburtle: do you mean adding something like a `GetPath()` function to the Loader that we can call to have to internally get the path for us without making the `FileSystem` public? We could call this from the `RuntimeConfigProvider` before we instantiate the file watcher.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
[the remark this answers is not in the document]
aaronburtle: I see what you mean, I'll update the comment so it is more clear that we really need to wait on the event plus the actual hot reloading.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -29,7 +29,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     // or user provided config file which could be a relative file path, absolute file path or simply the file name assumed to be in current directory.
     private string _baseConfigFilePath;
 
-    private readonly IFileSystem _fileSystem;
+    public readonly IFileSystem FileSystem;
```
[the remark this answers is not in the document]
aaronburtle: Added a helper function to the loader and then have this private again, good suggestion!

src/Core/Configurations/RuntimeConfigProvider.cs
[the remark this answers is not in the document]
seantleonard: To align with your wanting to keep it void, consider whether number 1 would make more sense. The downside I mention may not necessarily be a downside, and the method name `SetupConfigFileWatcher()` clearly communicates what side effects occur as a result of calling it and is functionally cohesive (Code Complete Ch7.2) 

I only called this out because when I read `CheckForAndSetupConfigFileWatcher()`, I couldn't confidently answer these questions:
1. What happens when `checkFor` either finds or doesn't find a configFileWatcher? Is it finding or not finding that cause positive/negative impacts or neither?
2. Is the method always guaranteed to succeed?  there's no try/catch wrapping this method call so the side effect the method name conveys is successfully applied?

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -1,30 +1,18 @@
...
-
 namespace Azure.DataApiBuilder.Core.Configurations;
 
 public class ConfigFileWatcher
 {
     private FileSystemWatcher? _fileWatcher;
     private readonly RuntimeConfigProvider? _configProvider;
 
-    public ConfigFileWatcher(RuntimeConfigProvider configProvider)
+    public ConfigFileWatcher(RuntimeConfigProvider configProvider, string directoryName, string configFileName)
```
seantleonard, line 11: Is there any way to combine these into a single parameter? e.g. FullConfigFilePath? then the process of breaking that  down into a path + filename.extension can be handled as an implementation detail.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -81,29 +82,39 @@ public RuntimeConfig GetConfig()
...
-                _configFileWatcher = new(this);
+                FileSystemRuntimeConfigLoader loader = (FileSystemRuntimeConfigLoader)ConfigLoader;
+                string configFilePath = loader.ConfigFilePath;
+                string configFileName = Path.GetFileName(configFilePath);
+                string? directoryName = Path.GetDirectoryName(configFilePath);
+                directoryName = string.IsNullOrWhiteSpace(directoryName) ?
+                    loader.GetCurrentDirectoryFromFileSystem() :
+                    directoryName;
+                _configFileWatcher = new(this, directoryName, configFileName);
+                return true;
```
seantleonard, line 109: Return true would be valid for both conditions where `_configFileWatcher = new(...)` is called, wouldn't it?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -56,11 +56,20 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
...
         _baseConfigFilePath = baseConfigFilePath;
         ConfigFilePath = GetFinalConfigFilePath();
     }
 
+    /// <summary>
+    /// Gets the current directory of the file system and returns
+    /// it as a string.
+    /// </summary>
+    /// <returns></returns>
+    public string GetCurrentDirectoryFromFileSystem()
```
seantleonard, line 69: `FromFileSystem` is an implementation detail. The caller of this method just wants to get a directory value that represents.. ? 

However, simplifying to `GetCurrentDirectory()` is vague and seems like we're exposing filesystem internals. 

Can you think of a name that would represents:
-  What the method is returning (directory)
- The method'significance/relevance to the caller? 
- What's the relation of 'current directory' to `FileSystemRuntimeConfigLoader`? Current directory of what?

src/Core/Configurations/RuntimeConfigProvider.cs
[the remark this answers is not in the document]
aaronburtle: I think returning a bool makes sense because it can immediately let the caller know whether or not they have just instantiated a file watcher with the call or not. There is no need to handle the returned bool if the caller does not have a need to do so and the function name is easiest to understand.

src/Core/Configurations/ConfigFileWatcher.cs
aaronburtle: unfortunately no, the directory name depends on the loader, so we'd have to pass the loader into the config file watcher which is less desirable than just combining these strings I reckon.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -79,9 +82,41 @@ public RuntimeConfig GetConfig()
...
+            }
+            else
+            {
+                _configFileWatcher = new();
```
seantleonard, line 113: Why create this and return false?

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -0,0 +1,151 @@
...
+{
+    [TestClass]
+    public class ConfigFileWatcherUnitTests
+    {
+        /// <summary>
+        /// Use the file system to add a file that matches the format of our config file.
+        /// Use that to create the needed config loader and provider, and then get the
+        /// actual config object from the provider, instantiating our config file watcher
+        /// in the process. Modify that file we created originally and verify we reload
+        /// the config correctly for the desired RuntimeOptions.
```
ayush3797, line 21: nit: The summary reads more like all the steps performed in the steps. Not really needed here - can include within the test as comments. Summary should be around what we are testing here - the hot reload successfully updates the runtime section.

ayush3797:
@aaronburtle, just trying to understand here. Are we able to hot reload runtime section only because throughout the code, the properties in the runtime section are directly referenced from the runtimeconfig object as opposed to other sections in the runtimeconfig like entities where the metadata has already been used in creating objects like SqlMetaDataProvider etc? And updating that metadata is not that straightforward and would require re-creating those objects using something like IConfigurableService?

seantleonard (approved):
thank you for making this change and addressing comments!

aaronburtle:
> @aaronburtle, just trying to understand here. Are we able to hot reload runtime section only because throughout the code, the properties in the runtime section are directly referenced from the runtimeconfig object as opposed to other sections in the runtimeconfig like entities where the metadata has already been used in creating objects like SqlMetaDataProvider etc? And updating that metadata is not that straightforward and would require re-creating those objects using something like IConfigurableService?

That's right, this is the initital step, which simply updates the objects we are storing in memory as the runtimeconfig, but without doing the additional work to re-initialize the settings that are based on that config object during startup. This will come in subsequent PRs using an event driven design.

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: followup up comment for clarity, after our sync on teams decided to keep GetCurrentDiretoryFromFileSystem

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            Filter = configFileName,
+            EnableRaisingEvents = true
+        };
+
+        _configProvider = configProvider;
+        _fileWatcher.Changed += OnConfigFileChange;
+    }
+
+    public ConfigFileWatcher() { }
```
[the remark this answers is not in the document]
aaronburtle: we simply use this to avoid instantiating the actual FileWatching mechanic but being able to check if we have already setup this wrapper object for cases like lateconfig, let's us avoid tracking this information by just making a null check..

src/Core/Configurations/RuntimeConfigProvider.cs
aaronburtle: fixed

src/Core/Configurations/ConfigFileWatcher.cs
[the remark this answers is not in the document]
aaronburtle: that would be another way to do this, but then you're running that logic every single time you try to get config, this let's us just see if the object is null or not and then skip all of that logic.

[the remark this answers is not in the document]
aaronburtle: updated to use bool for beauty.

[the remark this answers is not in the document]
seantleonard: talked offline and updated to not create empty watcher object.

seantleonard (approved):
Thank you for bearing with me on feedback and discussing the design in depth in order to ensure proper encapsulation of class behavior and not leaking implementation details.

seantleonard (approved):

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -79,9 +86,34 @@ public RuntimeConfig GetConfig()
...
+    /// that we have attempted to do so to avoid repeat checks in future calls.
+    /// Returns true if we instantiate a file watcher.
+    /// </summary>
+    private bool TrySetupConfigFileWatcher()
+    {
+        if (!_hasAttemptedFileWatcherConfiguration)
+        {
+            if (!IsLateConfigured && _runtimeConfig is not null && _runtimeConfig.IsDevelopmentMode())
+            {
+                FileSystemRuntimeConfigLoader loader = (FileSystemRuntimeConfigLoader)ConfigLoader;
```
Aniruddh25, line 106: This is assuming the ConfigLoader HAS to be a FileSystemRuntimeConfigLoader - today it is - but goes against the notion of using a base RuntimeConifgloader. Suggest to put this code in a try catch block to prevent throwing exceptions if the cast fails - adhering to the naming convention of TryXXX

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (!_configProvider!.IsLateConfigured && _configProvider!.GetConfig().IsDevelopmentMode())
+            {
+                _configProvider!.HotReloadConfig();
+            }
+        }
+        catch (Exception ex)
+        {
+            // Need to remove the dependency configuring authentication has on the RuntimeConfigProvider
+            // before we can have an ILogger here.
+            Console.WriteLine("Unable to Hot Reload configuration file due to " + ex.Message);
```
Aniruddh25, line 50:
```suggestion
            Console.WriteLine("Unable to hot reload configuration file due to " + ex.Message);
```

Aniruddh25 (changes requested):
Much clear separation of responsibilities, but a few more suggestions to avoid assumptions

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+    private void OnConfigFileChange(object sender, FileSystemEventArgs e)
+    {
+        try
+        {
+            if (_configProvider is null)
+            {
+                throw new ArgumentNullException("_configProvider can not be null.");
+            }
+
+            if (!_configProvider!.IsLateConfigured && _configProvider!.GetConfig().IsDevelopmentMode())
```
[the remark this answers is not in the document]
aaronburtle: No, it can be removed, thanks.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -79,9 +86,34 @@ public RuntimeConfig GetConfig()
...
+    /// <summary>
+    /// Checks if we have already attempted to configure the file watcher, if not
+    /// instantiate the file watcher if we are in the correct scenario. If we
+    /// are not in the correct scenario, do not setup a file watcher but remember
+    /// that we have attempted to do so to avoid repeat checks in future calls.
+    /// Returns true if we instantiate a file watcher.
+    /// </summary>
+    private bool TrySetupConfigFileWatcher()
+    {
+        if (!_hasAttemptedFileWatcherConfiguration)
```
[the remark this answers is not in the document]
aaronburtle: we either need a boolean here or we have to do something like I had before where we check if the config file watcher is not null. this let's us avoid instantiating an empty config file watcher in the case where we are not actually doing any file watching (in the case that we are not in a hot reload scenario)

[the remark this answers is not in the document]
aaronburtle: or we can repeat the check of which scenario we are in every time this gets called. the bool prevents us from doing that more complicated check over and over.

src/Core/Configurations/ConfigFileWatcher.cs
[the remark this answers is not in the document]
seantleonard: that may have been missed. @aaronburtle, you'll need to remove this now.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -103,12 +103,18 @@ private bool TrySetupConfigFileWatcher()
...
+                try
+                {
+                    FileSystemRuntimeConfigLoader loader = (FileSystemRuntimeConfigLoader)ConfigLoader;
+                    _configFileWatcher = new(this, loader.GetConfigDirectoryName(), loader.GetConfigFileName());
+                    _hasAttemptedFileWatcherConfiguration = true;
+                    return true;
+                }
+                catch (Exception)
+                {
+                    return false;
```
Aniruddh25: We should at least log a warning here that an attempt to configure hot reloading failed

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -103,12 +103,18 @@ private bool TrySetupConfigFileWatcher()
...
+                }
             }
-
-            _hasAttemptedFileWatcherConfiguration = true;
-            return true;
         }
 
         return false;
```
Aniruddh25, line 114: this line of code is unreachable now, instead of returning from both the try and catch blocks, you can simply return if `_configFileWatcher is not null` at the end

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -38,16 +38,16 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
                 throw new ArgumentNullException("_configProvider can not be null.");
             }
 
-            if (!_configProvider!.IsLateConfigured && _configProvider!.GetConfig().IsDevelopmentMode())
+            if (!_configProvider!.IsLateConfigured && _configProvider.GetConfig().IsDevelopmentMode())
```
Aniruddh25:
```suggestion
            if (!_configProvider.IsLateConfigured && _configProvider.GetConfig().IsDevelopmentMode())
```

Aniruddh25 (changes requested):

src/Core/Configurations/RuntimeConfigProvider.cs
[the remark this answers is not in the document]
Aniruddh25: In the case where we are do NOT actually WANT to do any file watching, do we even need to call this function? Are we calling it in the late configured scenario? Looking at the name of the function, feels natural to use it only if you KNOW you loaded from a file. 

Would we ever end up in a case where you either need to check for the boolean OR create an empty object?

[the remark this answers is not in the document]
aaronburtle: followup from our sync on teams, because of how we check for the runtimeconfig being null and additionally the usage of tryloadconfig, we really dont need this bool. removed for conciseness.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -99,22 +94,21 @@ public RuntimeConfig GetConfig()
...
-                    return false;
-                }
+                FileSystemRuntimeConfigLoader loader = (FileSystemRuntimeConfigLoader)ConfigLoader;
+                _configFileWatcher = new(this, loader.GetConfigDirectoryName(), loader.GetConfigFileName());
             }
+            catch (Exception ex)
+            {
+                // Need to remove the dependencies in startup on the RuntimeConfigProvider
+                // before we can have an ILogger here.
+                Console.WriteLine($"Attempt to configure config file watcher for hot reload failed due to: {ex.Message}.");
```
Aniruddh25, line 108: Dont we have access to logger?

Aniruddh25: nvm see the comment now

Aniruddh25 (approved):
Looks good, thanks for addressing all comments and making sure all cases are covered!

src/Core/Configurations/RuntimeConfigProvider.cs
aaronburtle: before this line of code is executed, we will :)

aaronburtle: i need to refactor and resolve some dependencies on the runtimeconfig provider in startup to get ILogger here.

verschaevesiebe:
@aaronburtle i've had a quick look at this through the use of docker and a fileshare / volume mount. (Both in Azure as well as locally) and i don't seem to be able to perform the hotreload scenario.

I see in the code that there's a "IsLateConfigured" boolean that's checked. What's the definition of this since the only way i know how to run dab in docker is through the use of the --ConfigFileName argument. (does this fall under being late configured ?)

I'm just trying to reason what i'm doing wrong in this scenario or if it might be a bug. This would be usefull in the future if dab-configs are stored on a storage account in the cloud and we might need to update it withour restarting the full container.

Thanks.

Aniruddh25:
hi @verschaevesiebe, thank you for your patience. The hot reload of runtime config feature is not completely done yet. `IsLateConfigured` means the configuration wasn't present at start up of DataApiBuilder but is configured post startup. 

If you run dab in docker through the use of --ConfigFileName, it doesnt fall under late configured. It may not be working as expected at the moment since the feature isn't complete. We are still tracking it as part of the issue:: https://github.com/Azure/data-api-builder/issues/67

PR #1739, opened 2023-09-21
Fix setting cosmos configuration at runtime by checking whether the graphql schema is set

Mathos1432:
## Why make this change?
Failing to set the configuration at runtime if there is not "schema" file set in the config. This is not required when setting the config at runtime since the graphql schema is set already.

## How was this tested?

- [x] Validated locally
- [x] Updated the tests so the schema property is removed from the config. Verified that the tests were failing before the fix and passing after.

### DIFF
--- a/src/Core/Configurations/RuntimeConfigValidator.cs
+++ b/src/Core/Configurations/RuntimeConfigValidator.cs
@@ -128,6 +128,8 @@ namespace Azure.DataApiBuilder.Core.Configurations
                         HttpStatusCode.ServiceUnavailable,
                         DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
 
+                // The schema is provided through GraphQLSchema and not the Schema file when the configuration
+                // is received after startup.
                 if (string.IsNullOrEmpty(cosmosDbNoSql.GraphQLSchema))
                 {
                     if (string.IsNullOrEmpty(cosmosDbNoSql.Schema))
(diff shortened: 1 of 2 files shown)


src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -128,17 +128,20 @@ public void ValidateConfig()
                         HttpStatusCode.ServiceUnavailable,
                         DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
 
-                if (string.IsNullOrEmpty(cosmosDbNoSql.Schema))
+                if (string.IsNullOrEmpty(cosmosDbNoSql.GraphQLSchema))
```
Aniruddh25, line 131: Good find! in order to ensure there's no regression in future, can we add a test that does this?

`Modified the test config file to remove the schema, verified that the late config tests were failing`

Mathos1432: Sure, but it would required either a second configuration file for cosmos which will then have to be maintained, or to parse the config file in the test to change that variable. Any preference on the approach?

Aniruddh25: The latter is preferable - parse it, and change that variable - then use it to configure

seantleonard: similar to how in ConfigurationTests.cs `ConfigurationPostParameters configParams = GetCosmosConfigurationParameters();` sets the values for the CosmosOptions

seantleonard: also worth a code comment here calling out that `GraphQLSchema` (and not `Schema`) will be populated when configuration is received post-startup

seantleonard (approved):

Aniruddh25 (approved):
LGTM, thanks for the quick fix and the accompanying test!

Issue #1736, opened 2023-09-21
[Bug]: dab start command throws unhandled exception if connection string is malformed for MsSql


Aniruddh25:
### What happened?

1. After building the solution, from the root of the repo: `cd src\out\cli\net6.0`
1. Microsoft.DataApiBuilder.exe init --database-type mssql --connection-string "abcd"
2. Microsoft.DataApiBuilder.exe add Book  --source dbo.books --permissions "anonymous:*"
3. Microsoft.DataApiBuilder.exe start

CLI will throw unhandled exception. 

![image](https://github.com/Azure/data-api-builder/assets/3513779/fdc2bdc2-ad2d-49bd-b543-b9d4dfccc7ea)

### Version

main

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

REST, GraphQL

### Relevant log output

```Text
Information: Microsoft.DataApiBuilder 1.0.0+38557d3f45316631c12a6f6f2b82da6f08f7e98d
Information: User provided config file: issue.json
Loading config file from issue.json.
Unhandled exception. Azure.DataApiBuilder.Service.Exceptions.DataApiBuilderException: A valid Connection String should be provided.
 ---> System.ArgumentException: Format of the initialization string does not conform to specification starting at index 0.
   at System.Data.Common.DbConnectionOptions.GetKeyValuePair(String connectionString, Int32 currentPosition, StringBuilder buffer, Boolean useOdbcRules, String& keyname, String& keyvalue)
   at System.Data.Common.DbConnectionOptions.ParseInternal(Dictionary`2 parsetable, String connectionString, Boolean buildChain, Dictionary`2 synonyms, Boolean firstKey)
   at System.Data.Common.DbConnectionOptions..ctor(String connectionString, Dictionary`2 synonyms, Boolean useOdbcRules)
   at System.Data.Common.DbConnectionStringBuilder.set_ConnectionString(String value)
   at Microsoft.Data.SqlClient.SqlConnectionStringBuilder..ctor(String connectionString)
   at Azure.DataApiBuilder.Config.RuntimeConfigLoader.GetConnectionStringWithApplicationName(String connectionString) in D:\hawaii-gql\src\Config\RuntimeConfigLoader.cs:line 224
   --- End of inner exception stack trace ---
[...]

PR #1741, opened 2023-09-22
Unhandled Exception for malformed connection string bug fix

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/1736

When a connection string was not properly formed, in the case of `MSSQL` database types it was possible to have an exception when modifying the connection string as part of our normal code. However, that exception was not being handled and would cause an ungraceful crash of the application. We now handle this exception.

## What is this change?

Include `DataApiBuilderException` in the exception handling for where we modify the connection string in this particular case. And include a regression test case.

New error messaging can be seen here:

image

It should also be noted that some tests had expected behavior that did not account for our new exception handling. Some test cases have been changed to `postgresql` data base type to avoid conflicts with our new post processing error handling, when that new error handling is not relevant to what those tests were covering. Since the post processing only happens in the following conditional, this is an easy way to keep our test coverage without major changes.

image

Our tests for error checking bad connection strings for mssql was also relying on having the actual exception available, but we now have exception handling that will catch that and return a different exception than the test is expecting. To keep this test working we redirect standard error and then checking against that messaging for those cases where we are both mssql db type and have a malformed connection string . The remaining difference for those cases that do not have valid connection strings and are mssql db type is that we have an `InternalServerError` as the status code now, so we modify the test to look for that instead in this case.

## How was this tested?

A regression test is added to the unit tests for the runtime config loader. This is where the change happened, and the new case covers a malformed connection string, verifying that we handle the exception correctly by validating that we return false from the function that was throwing the unhandled exception.

## Sample Request(s)

You can manually repro with the instructions from the linked issue this closes which are the following steps:

1. After building the solution, from the root of the repo: cd src\out\cli\net6.0
2. Microsoft.DataApiBuilder.exe init --database-type mssql --connection-string "abcd"
3. Microsoft.DataApiBuilder.exe add Book --source dbo.books --permissions "anonymous:*"
4. Microsoft.DataApiBuilder.exe start

### DIFF
--- a/src/Cli.Tests/EnvironmentTests.cs
+++ b/src/Cli.Tests/EnvironmentTests.cs
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using Azure.DataApiBuilder.Config.Converters;
+using Microsoft.Extensions.Hosting;
 
 namespace Cli.Tests;
 
@@ -157,7 +159,10 @@ public class EnvironmentTests
         );
 
         string? output = await process.StandardError.ReadLineAsync();
-        Assert.AreEqual(output, "Deserialization of the configuration file failed during a post-processing step.");
+        Assert.AreEqual("Deserialization of the configuration file failed during a post-processing step.", output);
+        output = await process.StandardError.ReadToEndAsync();
+        StringAssert.Contains(output, "Environmental Variable, "
+            + expectedEnvVarName + ", not found.", StringComparison.Ordinal);
         process.Kill();
     }
 
--- a/src/Cli.Tests/EnvironmentTests.cs
+++ b/src/Cli.Tests/EnvironmentTests.cs
@@ -1,9 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System;
 using Azure.DataApiBuilder.Config.Converters;
-using Microsoft.Extensions.Hosting;
 
 namespace Cli.Tests;
 
--- a/src/Cli.Tests/EnvironmentTests.cs
+++ b/src/Cli.Tests/EnvironmentTests.cs
@@ -143,13 +143,6 @@ public class EnvironmentTests
     /// variable CONN_STRING. This will cause an exception during the post
     /// processing of the deserialization of the config. We verify the expected
     /// error message returns.
-    /// This test has been disabled as it is causing the build server to hang indefinitely.
-    /// There is something problematic with reading from stderr and stdout in this test
-    /// that is causing the issue. It's possible that the stream is not being flushed
-    /// by the process so when the test tries to read it, it hangs waiting for the stream
-    /// to be readable, but it will require more investigation to determine the root cause.
-    /// I feel confident that the overarching scenario is covered through other testing
-    /// so disabling temporarily while we investigate should be acceptable.
     /// </summary>
     [TestMethod]
     public async Task FailureToStartEngineWhenEnvVarNamedWrong()
@@ -164,7 +157,7 @@ public class EnvironmentTests
         );
 
         string? output = await process.StandardError.ReadLineAsync();
-        Assert.AreEqual(output, "Deserialization of the configuration file failed.");
+        Assert.AreEqual(output, "Deserialization of the configuration file failed during a post-processing step.");
         process.Kill();
     }
 
--- a/src/Cli.Tests/EnvironmentTests.cs
+++ b/src/Cli.Tests/EnvironmentTests.cs
@@ -140,7 +140,9 @@ public class EnvironmentTests
     /// Validates that engine startup fails when the CONN_STRING environment
     /// variable is not found. This test simulates this by defining the
     /// environment variable COMM_STRINGX and not setting the expected
-    /// variable CONN_STRING.
+    /// variable CONN_STRING. This will cause an exception during the post
+    /// processing of the deserialization of the config. We verify the expected
+    /// error message returns.
     /// This test has been disabled as it is causing the build server to hang indefinitely.
     /// There is something problematic with reading from stderr and stdout in this test
     /// that is causing the issue. It's possible that the stream is not being flushed
@@ -162,8 +164,7 @@ public class EnvironmentTests
         );
 
         string? output = await process.StandardError.ReadLineAsync();
-        StringAssert.Contains(output, "Environmental Variable, "
-            + expectedEnvVarName + ", not found.", StringComparison.Ordinal);
+        Assert.AreEqual(output, "Deserialization of the configuration file failed.");
         process.Kill();
     }
 
--- a/src/Config/RuntimeConfigLoader.cs
+++ b/src/Config/RuntimeConfigLoader.cs
@@ -145,7 +145,9 @@ public abstract class RuntimeConfigLoader
             }
 
         }
-        catch (JsonException ex)
+        catch (Exception ex) when (
+            ex is JsonException ||
+            ex is DataApiBuilderException)
         {
             string errorMessage = "Deserialization of the configuration file failed.";
 
--- a/src/Config/RuntimeConfigLoader.cs
+++ b/src/Config/RuntimeConfigLoader.cs
@@ -149,7 +149,8 @@ public abstract class RuntimeConfigLoader
             ex is JsonException ||
             ex is DataApiBuilderException)
         {
-            string errorMessage = "Deserialization of the configuration file failed.";
+            string errorMessage = ex is JsonException ? "Deserialization of the configuration file failed." :
+                "Deserialization of the configuration file failed during a post-processing step.";
 
             // logger can be null when called from CLI
             if (logger is null)
--- a/src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
+++ b/src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
@@ -159,7 +159,7 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
         [DataRow("\"notsupporteddb\"", "")]
         [DataRow("\"mssql\"", "\"notsupportedconnectionstring\"")]
         [TestMethod("Validates that JSON deserialization failures are gracefully caught.")]
-        public void TestDeserializationFailures(string dbType, string connectionString)
+        public void TestDataSourceDeserializationFailures(string dbType, string connectionString)
         {
             string configJson = @"
 {
--- a/src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
+++ b/src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
@@ -156,8 +156,10 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
             Assert.ThrowsException<DataApiBuilderException>(() => JsonSerializer.Deserialize<StubJsonType>(json, options));
         }
 
-        [DataRow("\"notsupporteddb\"", "")]
-        [DataRow("\"mssql\"", "\"notsupportedconnectionstring\"")]
+        [DataRow("\"notsupporteddb\"", "",
+            DisplayName="Tests that a database type which will not deserialize correctly fails.")]
+        [DataRow("\"mssql\"", "\"notsupportedconnectionstring\"",
+            DisplayName="Tests that a malformed connection string fails during post-processing.")]
         [TestMethod("Validates that JSON deserialization failures are gracefully caught.")]
         public void TestDataSourceDeserializationFailures(string dbType, string connectionString)
         {
@@ -169,6 +171,7 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
     },
     ""entities"":{ }
 }";
+            // replaceEnvVar: true is needed to make sure we do post-processing for the connection string case
             Assert.IsFalse(RuntimeConfigLoader.TryParseConfig(configJson, out RuntimeConfig deserializedConfig, replaceEnvVar: true));
             Assert.IsNull(deserializedConfig);
         }
--- a/src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
+++ b/src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
@@ -83,13 +83,13 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
                 if (replaceEnvVar)
                 {
                     Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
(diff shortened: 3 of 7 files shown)


src/Config/RuntimeConfigLoader.cs
```diff
@@ -145,7 +145,9 @@ public RuntimeConfigLoader(string? connectionString = null)
             }
 
         }
-        catch (JsonException ex)
+        catch (Exception ex) when (
+            ex is JsonException ||
+            ex is DataApiBuilderException)
```
Aniruddh25, line 150: if the exception is DataApiBuilderException, is the error message here still a valid message? How is the deserialization failure message valid for a connection string failure?

Aniruddh25 (changes requested):
Thanks for the quick fix and test but needs better messaging to convey exactly what is wrong with the configuration so the users can self help themselves.

src/Config/RuntimeConfigLoader.cs
aaronburtle: I thought it was still descriptive for what is happening because this is a post processing step of the deserialization. We deserialize directly from the json string to create the objects, then we have a conditional update as part of the post processing. So in my eyes that was still part of the deserialization process and giving it another name in the error messaging might be confusing.

I considered calling this "parsing" instead but thought it was clearer like this. this is what it looks like now

image

src/Config/RuntimeConfigLoader.cs
```diff
@@ -145,7 +145,9 @@ public RuntimeConfigLoader(string? connectionString = null)
             }
 
         }
-        catch (JsonException ex)
+        catch (Exception ex) when (
+            ex is JsonException ||
```
seantleonard, line 149: I see what Ani is saying. JsonException directly correlates to a deserialization issue. However, the DAB exception could be something else, like in this case with a failure to build a connection string.
Perhaps a separate catch for the DAB exception and log the associated error message.

aaronburtle: I am just not sure that differentiating between the post processing done during deserialization and the core deserialization itself is going to be a key distinction we want to make in the error messaging.

aaronburtle: As an example of what I mean, suppose that the Json Deserializer added a post processing option where you could pass in a function and it would execute that function as a post processing step for you.

from the users perspective, the same thing is happening, would we differentiate that exception? 

this seems like a detail that is really only visible at the implementation level, unless we don't consider the post processing steps part of the deserialization process generally, which I do.

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -162,8 +164,7 @@ public async Task FailureToStartEngineWhenEnvVarNamedWrong()
         );
 
         string? output = await process.StandardError.ReadLineAsync();
-        StringAssert.Contains(output, "Environmental Variable, "
-            + expectedEnvVarName + ", not found.", StringComparison.Ordinal);
+        Assert.AreEqual(output, "Deserialization of the configuration file failed.");
```
seantleonard, line 167: The previous check explicitly checked for the failure that the test was trying to trigger. Can you include both your new addition to check the high level deserialization failure notice as well as checking `StringAssert.Contains(output, "Environmental Variable, "`?

Also, the test name indicates that this is testing an invalid env_var name provided in config, as opposed to a invalid connection string failing post-processing. Can you clarify test intent here?

src/Config/RuntimeConfigLoader.cs
seantleonard: include that screenshot in the pr description to demonstrate this PR's affect.

src/Config/RuntimeConfigLoader.cs
seantleonard: I see. I think you point then begs the question of whether the error message we currently set is too specific?

> Deserialization of the configuration file failed.

Since that message specifically distinguishes between deserialization and post-processing, a still relevant but less granular message that wouldn't affect how a user reacts could be:

> Configuration file processing failed.

src/Cli.Tests/EnvironmentTests.cs
aaronburtle: The error message contains whatever we decide to put in the messaging, which right now is, "Deserialization of the configuration file failed.", to add more details than that we would have to add it to the original message.

seantleonard: Understood that we have error handling now and set the message "Deserialization of the configuration file failed." My point is that the test can also check the specific error causing the deserialization issue. Those details are printed to the console as shown by your screenshot: 
image

The environment variable error messaging may still be available to check here, perhaps if you continue reading errors printed to the console `process.StandardError.ReadLineAsync();`.

That way, the test can be specific so you know that no other possible issues are causing the test to pass/fail:
- if you're testing env_var issues, check for both the "Deserialization of the configuration file failed." and exception specific messaging printed to the console
- if you're testing connectionstring builder errors, check for both the "Deserialization of the configuration file failed." and exception specific messaging printed to the console

aaronburtle: sure we can read out the entire error message and verify the underlying exception includes the environment variable error.

src/Config/RuntimeConfigLoader.cs
aaronburtle: I added some language about this being during a post processing step, but I am still not sure this really helps the end understand better what is wrong.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -83,13 +83,13 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
                     Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
-                        GetModifiedJsonString(repValues, @"""mssql"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
+                        GetModifiedJsonString(repValues, @"""postgresql"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
                         "Should read the expected config");
                 }
                 else
                 {
                     Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
-                        GetModifiedJsonString(repKeys, @"""mssql"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
+                        GetModifiedJsonString(repKeys, @"""postgresql"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
```
seantleonard, line 92: why these changes?

aaronburtle: this test doesnt use a valid connection string format, so if we use mssql db type we end up inside the code where we will get an exception now during post processing. we want to avoid that for this test, so using postgressql is an easy way to do that.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -156,16 +156,20 @@ public void CheckConfigEnvParsingThrowExceptions(string invalidEnvVarName)
...
     ""data-source"": {
-        ""database-type"": ""notsupporteddb""
-     }
+        ""database-type"": " + dbType + @",
+        ""connection-string"": " + connectionString + @"
+    },
+    ""entities"":{ }
 }";
-            Assert.IsFalse(RuntimeConfigLoader.TryParseConfig(configJson, out RuntimeConfig deserializedConfig));
+            Assert.IsFalse(RuntimeConfigLoader.TryParseConfig(configJson, out RuntimeConfig deserializedConfig, replaceEnvVar: true));
```
seantleonard, line 172: what is replaceEnvVar doing for the test? could yo add a test description/displaynames for the data rows?

src/Config/RuntimeConfigLoader.cs
seantleonard: That message was fine. When exceptions are raised their errors are automatically output to the console, so that gives more detail.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: Sure I added a comment. This is just used to get into the portion of the code that does the post-processing, which is what we are testing for the connection string.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
seantleonard: followed up on Teams: Our configloader looks for "MSSQL" dbtype to decide whether to scrutinize the connection string and subject it to the connectionstringbuilder. 

This test is evaluating env_Var replacement which may end up creating an intentionally bad connection string, however, the test is only checking for whether env_vars were processed properly by checking whether configloading succeeded. Now that we have connecitonstringbuilder validation in the runtimeconfigloader, the test broke because of a different untargeted failure.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
seantleonard: if we don't replace environment variables, the connectionstringBuilder check / exception won't be hit? wouldn't an unreplace env_var be an invalid connection string?

aaronburtle: the connection string builder is called within this call that happens, and this call is only hit for MSSQL with replace environment variables being true
image

seantleonard:
How does this affect the late-bound config scenario?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
seantleonard: gotcha, add that to the PR description too and/or note that in the test description comment for the test you changed from mssql->pgsql

aaronburtle:
> How does this affect the late-bound config scenario?

the late bound scenario is just calling this function, the TryParse function, with a string that was sent to the configuration controller, rather than one that was read from a file. this will not cause any special behavior, but it will catch unhandled exceptions that otherwise would have happened if this case was triggered in the hosted scenario. so it does provide coverage there, but nothing special.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -73,16 +74,25 @@ public async Task CheckNoExceptionForNoForeignKey()
...
-        [DataRow("DO NOT EDIT, look at CONTRIBUTING.md on how to run tests")]
-        [DataRow("")]
-        public async Task CheckExceptionForBadConnectionStringForMsSql(string connectionString)
+        [DataRow(";;;;;fooBarBAZ", true)]
+        [DataRow("!&^%*&$$%#$%@$%#@()", true)]
+        [DataRow("Server=<>;Databases=<>;Persist Security Info=False;Integrated Security=True;MultipleActiveResultSets=False;Connection Timeout=5;", true)]
+        [DataRow("Servers=<>;Database=<>;Persist Security Info=False;Integrated Security=True;MultipleActiveResultSets=False;Connection Timeout=5;", true)]
+        [DataRow("DO NOT EDIT, look at CONTRIBUTING.md on how to run tests", true)]
+        [DataRow("", false)]
+        public async Task CheckExceptionForBadConnectionStringForMsSql(string connectionString, bool invalidConnectionBuilderString)
```
Aniruddh25, line 83:
```suggestion
        public async Task CheckExceptionForBadConnectionStringForMsSql(string connectionString, bool isInvalidConnectionBuilderString)
```

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -178,9 +188,19 @@ private static async Task CheckExceptionForBadConnectionStringHelperAsync(string
             catch (DataApiBuilderException ex)
             {
                 // use contains to correctly cover db/user unique error messaging
-                Assert.IsTrue(ex.Message.Contains(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE));
-                Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
+                // if sw is not null it holds the error messaging
+                string error = sw is null ? ex.Message : sw.ToString();
+                Assert.IsTrue(error.Contains(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE));
                 Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ErrorInInitialization, ex.SubStatusCode);
+                if (sw is not null)
```
Aniruddh25, line 195: Check for the newly added post processing error message here in the exception to stdout?

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -178,9 +188,19 @@ private static async Task CheckExceptionForBadConnectionStringHelperAsync(string
...
+                {
+                    Assert.AreEqual(HttpStatusCode.InternalServerError, ex.StatusCode);
```
Aniruddh25, line 197: Why is this an internal server error? Why not serviceunavailable?
Did the server not come up due to invalid user input or something wrong with the server?

Aniruddh25 (changes requested):
Good capture of all errors, few more suggestions.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
aaronburtle: This is the exception that comes back trying to invoke `GetConfig()` through the `RuntimeConfigProvider` in the case where trying to load the known config results in that config not loading but instead being `null`. In this case we had decided on an `InternalServerError`. I think this makes sense because it is indicating back to the user that the problem is with our internal processing, in this case of the config file.

image

aaronburtle: after discussion we decided to use `ServiceUnavailable` for all of the config related errors. this is the status code used in the config validator already, so changing this matches the rest of the messaging we use.

Aniruddh25: Since we decided to go with ServiceUnavailable, shouldn't the assert here be changed?

Aniruddh25 (approved):
LGTM, after fixing the remaining test failures. Thanks for the quick fix

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
aaronburtle: yes, I had this locally only, pushed the changes now.

seantleonard (approved):
thank you for getting this error handling improvement in and addressing feedback

seantleonard (commented):
feedback

PR #1768, opened 2023-10-03
Validations to ensure that the reference config files stay upto date

### LINKED ISSUE #1766: Add tests to check staleness of checked-in reference config files
There are a few reference config files checked-in in the `Service.Tests` directory. Whenever there is a change in the way CLI generates a config file, these reference config file needs to be updated to reflect the latest changes.

At the moment, there is no validation to ensure that these reference files are as per the latest format. So, these config files tends to go out of date. 

Hence, add tests to ensure that these reference config files stay up to date.

severussundar:
## Why make this change?

- Closes #1766 
  - There are 4 reference config files in the repo - `dab-config.MsSql.json`, `dab-config.MySql.json`, `dab-config.PostgreSql.json` and `dab-config.CosmsDb_NoSql.json`
  - When there are changes to the config file - addition of new elements, roles, entities, etc. these reference config files also need to be updated so that they stays up to date.
  - At the moment, there are no validations to ensure this and as a result, the config files tend to miss out on some of the changes. 

## What is this change?

- Validations for each config file are added to ensure that the reference config file is up to date with the latest way the CLI generates the config files. For each database type, the checked in reference file and the one generated by CLI for the integration tests are compared.
- All the reference config files are renamed from `dab-config.*.json` to `dab-config.*.reference.json` - The config generator scripts generate config files for integration tests with the name `dab-config.<database_type>.json`. In the repo, the reference file are named the same (`dab-config.<database_type>.json`). The config generator script, first deletes the file with the same name before executing the `dab` commands (in-order to generate the necessary config file for integration test).  To be able to compare the contents of the both the files - checked in reference file and the one generated by config generator script; both needs to exist during integration tests. Hence, the renaming of the reference config files. 
- Validations are added to compare the reference config file and the CLI generated config file.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests


src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -102,6 +102,7 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
         /// <code>Check: </code> Making sure table prefix matches expected prefix.
         /// </summary>
         [DataTestMethod]
+        [TestCategory(TestCategory.MSSQL)]
```
severussundar, line 105: This test looks for `dab-config.MsSql.json` file but it is unavailable in the build pipeline. The logic for fetching the file cannot be modified to read from `dab-config.MsSql.reference.json` as other tests also share that logic. 
Hence, running this test as part of the MsSql integration test pipeline (The other tests that use the same helper method to look for `dab-config.MsSql.json` already run as part of the MsSql integration test pipeline)

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -943,6 +944,95 @@ private Task ConfigFileDeserializationValidationHelper(string jsonString)
...
 
+        /// <summary>
+        /// Helper method to parse the config json string and set the connection string field to a given string.
+        /// </summary>
+        /// <param name="configJsonString">Config Json string</param>
+        /// <param name="newConnectionString">Value to update the connection string with</param>
+        /// <returns>Config Json string with connection string set to the given value</returns>
+        private static string UpdateConnectionStringWithCustomValue(string configJsonString, string newConnectionString)
+        {
+            JObject configJson = JObject.Parse(configJsonString);
```
seantleonard, line 955: Did you rule out usage of `JsonNode` from system.text.json?  https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-dom#use-jsonnode

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -943,6 +944,95 @@ private Task ConfigFileDeserializationValidationHelper(string jsonString)
             return Verify(runtimeConfig);
         }
 
+        /// <summary>
+        /// Helper method to parse the config json string and set the connection string field to a given string.
+        /// </summary>
+        /// <param name="configJsonString">Config Json string</param>
+        /// <param name="newConnectionString">Value to update the connection string with</param>
+        /// <returns>Config Json string with connection string set to the given value</returns>
+        private static string UpdateConnectionStringWithCustomValue(string configJsonString, string newConnectionString)
```
seantleonard, line 953:
```suggestion
        private static string UpdateConnectionString(string configJsonString, string newConnectionString)
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -943,6 +944,95 @@ private Task ConfigFileDeserializationValidationHelper(string jsonString)
...
+        {
+            JObject configJson = JObject.Parse(configJsonString);
+            configJson["data-source"]["connection-string"] = newConnectionString;
+            return configJson.ToString();
+        }
+
+        /// <summary>
+        /// Validates that the checked-in reference config file for MsSql (dab-config.MsSql.reference.json) reflects the latest format with which
+        /// the CLI generates the config files.
+        /// The pipeline that runs the integration tests for MsSql, generates a config file (dab-config.MsSql.json) and uses this for all the tests.
```
seantleonard, line 963: Do these integration tests run locally and pass?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -943,6 +944,95 @@ private Task ConfigFileDeserializationValidationHelper(string jsonString)
...
+        /// This test compares these two files to make sure that the reference config file is up to date.
+        /// </summary>
+        [TestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        public void TestReferenceConfigFileIsInLatestFormatForMsSql()
+        {
+            string referenceConfigJson = File.ReadAllText($"{CONFIGFILE_NAME}.{MSSQL_ENVIRONMENT}.{REFERENCE_CONFIG_FILE_SUFFIX}{CONFIG_EXTENSION}");
```
seantleonard, line 970: Is the mssql config file only present in the file system when the `[TestCategory(TestCategory.MSSQL)]` attribute is added to this test method?

seantleonard (changes requested):

src/Service.Tests/Configuration/ConfigurationTests.cs
Aniruddh25: Instead of different functions for all the dbs, can we use DataRow for each db type and run these tests at the end of Build similar to the Unit Tests?

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
Aniruddh25: can we move the generation of `dab-config.MsSql.json` from the integration test pipeline to the build pipeline so that these unit tests stay as unit tests?

Aniruddh25 (approved):
LGTM, but wonder if the validation can be done as part of unit test instead of integration

seantleonard:
Latest update: new approach to be taken, so @severussundar will be closing this pr.

seantleonard:
Need to call out specifically in either tracking issue/new pr: ensure that this doesn't extend the runtime of the pipelines. Each integration test pipeline takes 2.5-3 minutes to run the configscript.txt files that use the cli to create a config. Doing this in unit test pipeline seems excessive.

Aniruddh25:
If moving the generation of files to unit test pipeline increases build time, we need to instead make one of the tests in SqlMetadataProviderUnitTests as integration tests since that needs `[TestCategory(TestCategory.MSSQL)]`.

severussundar:
Will be performing the validation through a new approach - so closing this PR for now.

PR #1769, opened 2023-10-03
Adding support for query parameter in Swagger UI

### LINKED ISSUE #1572: OpenAPI - Document query string
In the `OpenApiDocumentor` service:
- Document the appropriate query string for Get and Get(All) requests for entities that have their rest endpoint enabled
  -  `"rest": { "enabled": true }` , no rest property defined (we default to rest enabled true), essentially as long as rest enabled != false
  - Separate logic for Tables/Views
  - Separate logic for Stored procedures

Helpful resources:
- https://spec.openapis.org/oas/v3.0.1#parameter-object -> parameter location will be `query`
> query - Parameters that are appended to the URL. For example, in /items?id=###, the query parameter is id.
- Example swagger openapiv3  https://petstore3.swagger.io/#/store/getOrderById
image

- Example openapiv3 json with query parameters: [https://petstore3.swagger.io/api/v3/openapi.json](https://petstore3.swagger.io/api/v3/openapi.json)
- Example json
```json
"/pet/findByStatus": {
            "get": {
                "tags": [
                    "pet"
                ],
                "summary": "Finds Pets by status",
                "description": "Multiple status values can be provided with comma separated strings",
                "parameters": [
                    {
                        "name": "status",
                        "in": "query",
                        "description": "Status values that need to be considered for filter",
                        "required": false,
                        "explode": true,
                        "schema": {
                            "type": "string",
                            "default": "available",
                            "enum": [
                                "available",
                                "pending",
                                "sold"
                            ]
                        }
                    }
                ]
            }
        },
```

abhishekkumams:
## Why make this change?

- Closes #1572 
  - This is necessary to allow REST calls for GET and GET ALL support query parameters. 

## What is this change?

- Updating the OpenApi document to include the supported query parameters
- Separate Logic for StoredProcedures and Table/Views.
- `$filter`,`$orderBy`, `$first`, and `$after` will be available in swagger UI for GET/GET_ALL REST requests in Table/Views.
- StoredProcedures will also support query parameters.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

### DIFF
--- a/src/Core/Services/OpenAPI/OpenApiDocumentor.cs
+++ b/src/Core/Services/OpenAPI/OpenApiDocumentor.cs
@@ -285,6 +285,7 @@ namespace Azure.DataApiBuilder.Core.Services
                 // which is returned when using Enum.ToString("D").
                 // The "D" format specified "displays the enumeration entry as an integer value in the shortest representation possible."
                 OpenApiOperation getOperation = CreateBaseOperation(description: GETONE_DESCRIPTION, tags: tags);
+                getOperation.Parameters = CreateTableAndViewQueryParameters();
                 getOperation.Responses.Add(HttpStatusCode.OK.ToString("D"), CreateOpenApiResponse(description: nameof(HttpStatusCode.OK), responseObjectSchemaName: entityName));
                 openApiPathItemOperations.Add(OperationType.Get, getOperation);
 
@@ -317,6 +318,7 @@ namespace Azure.DataApiBuilder.Core.Services
             {
                 // Primary key(s) are not included in the URI paths of the GET (all) and POST operations.
                 OpenApiOperation getAllOperation = CreateBaseOperation(description: GETALL_DESCRIPTION, tags: tags);
+                getAllOperation.Parameters = CreateTableAndViewQueryParameters();
                 getAllOperation.Responses.Add(
                     HttpStatusCode.OK.ToString("D"),
                     CreateOpenApiResponse(description: nameof(HttpStatusCode.OK), responseObjectSchemaName: entityName, includeNextLink: true));
@@ -360,6 +362,7 @@ namespace Azure.DataApiBuilder.Core.Services
             if (configuredRestOperations[OperationType.Get])
             {
                 OpenApiOperation getOperation = CreateBaseOperation(description: SP_EXECUTE_DESCRIPTION, tags: tags);
+                getOperation.Parameters = CreateStoredProcedureParameters((StoredProcedureDefinition)sourceDefinition);
                 getOperation.Responses.Add(
                     HttpStatusCode.OK.ToString("D"),
                     CreateOpenApiResponse(
@@ -429,6 +432,103 @@ namespace Azure.DataApiBuilder.Core.Services
             };
 
             return operation;
+        }   
+
+        /// <summary>
+        /// This method creates a list of OpenApiParameter objects for the parameters of a stored procedure.
+        /// </summary>
+        private static List<OpenApiParameter> CreateStoredProcedureParameters(StoredProcedureDefinition spDefinition)
+        {
+            List<OpenApiParameter> parameters = new();
+
+            foreach ((string paramKey, ParameterDefinition parameterDefinition) in spDefinition.Parameters)
+            {
+                parameters.Add(
+                    GetOpenApiParameter(
+                        name: paramKey,
+                        description: "Query parameter for stored procedure parameter",
+                        required: !parameterDefinition.HasConfigDefault,
+                        type: TypeHelper.GetJsonDataTypeFromSystemType(parameterDefinition.SystemType).ToString().ToLower()
+                    )
+                );
+            }
+
+            return parameters;
+        }
+
+        /// <summary>
+        /// 
+        /// <summary>
+        private static List<OpenApiParameter> CreateTableAndViewQueryParameters()
+        {
+            List<OpenApiParameter> parameters = new ();
+
+            // Add $select query parameter
+            parameters.Add(
+                GetOpenApiParameter(
+                    name: "$select",
+                    description: "The query parameter $select allow to specify which fields must be returned.",
+                    required: false,
+                    type: "string"
+                )
+            );
+
+            // Add $filter query parameter
+            parameters.Add(
+                GetOpenApiParameter(
+                    name: "$filter",
+                    description: "The value of the $filter option is predicate expression (an expression that returns a boolean value) using entity's fields.",
+                    required: false,
+                    type: "string"
+                )
+            );
+
+            // Add $orderby query parameter
+            parameters.Add(
+                GetOpenApiParameter(
+                    name: "$orderby",
+                    description: "The value of the orderby parameter is a comma-separated list of expressions used to sort the items.",
+                    required: false,
+                    type: "string"
+                )
+            );
+
+            // Add $first query parameter
+            parameters.Add(
+                GetOpenApiParameter(
+                    name: "$first",
+                    description: "The value of the $first option is an integer value that specifies the number of items to return.",
+                    required: false,
+                    type: "integer"
+                )
+            );
+
+            // Add $after query parameter
+            parameters.Add(
+                GetOpenApiParameter(
+                    name: "$after",
+                    description: "The value of the $after option is a base64 encoded string value that specifies the cursor position after which items should be returned.",
+                    required: false,
+                    type: "string"
+                )
+            );
+
+            return parameters;
+        }
+
+        private static OpenApiParameter GetOpenApiParameter(string name, string description, bool required, string type)
+        {
+            return new OpenApiParameter
+            {
+                Name = name,
+                In = ParameterLocation.Query,
+                Description = description,
+                Required = required,
+                Schema = new OpenApiSchema
+                {
+                    Type = type
+                }
+            };
         }
 
         /// <summary>


src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            RuntimeEntities runtimeEntities = new(entities);
+            OpenApiDocument openApiDocument = await OpenApiTestBootstrap.GenerateOpenApiDocument(
+                runtimeEntities: runtimeEntities,
+                configFileName: CUSTOM_CONFIG,
+                databaseEnvironment: MSSQL_ENVIRONMENT);
+
+            OpenApiPathItem pathItem = openApiDocument.Paths.First().Value;
+            foreach (OpenApiOperation operation in pathItem.Operations.Values)
+            {
+                if (entityName is "UpdateBookTitle")
```
ayush3797, line 173: For string data types, should use the string specific comparison function - `.Equals()`.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// </summary>
+        /// <param name="entityName"></param>
+        /// <param name="objectName"></param>
+        /// <returns></returns>
+        [DataTestMethod]
+        [DataRow("CountBooks", "count_books", DisplayName = "StoredProcedure with no parameters")]
+        [DataRow("UpdateBookTitle", "update_book_title", DisplayName = "StoredProcedure with parameters")]
+        public async Task ValidateQueryParametersForStoredProcedures(string entityName, string objectName)
+        {
+            Dictionary<string, object> parameterDefaults = null;
```
ayush3797, line 144: Is null not by default?

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                        // No query parameters for other OperationTypes. 
+                        Assert.IsFalse(operation.Parameters.Any(param => param.In is ParameterLocation.Query));
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Validates that Query parameters are correctly generated for Stored Procedures.
+        /// Query Parameters for stored procedure will be generated for irrespective of the Operation Type.
```
ayush3797, line 133:
```suggestion
        /// Query Parameters for stored procedure will be generated for each operation irrespective of its Operation Type.
```

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                    Assert.IsTrue(pathItem.Parameters.Count is 0);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Validates that Query parameters are correctly generated for GET methods in Table/Views.
+        /// $select, $filter, $orderby, $first, $after are the query parameters.
+        /// </summary>
+        /// <param name="entityName"></param>
```
ayush3797, line 79: Remove or populate params.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// <param name="objectName"></param>
+        /// <param name="entitySourceType"></param>
+        /// <returns></returns>
+        [DataTestMethod]
+        [DataRow("BooksTable", "books", EntitySourceType.Table, DisplayName = "Table")]
```
ayush3797, line 84: Need better display names.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Validates that Query parameters are correctly generated for Stored Procedures.
+        /// Query Parameters for stored procedure will be generated for irrespective of the Operation Type.
+        /// If a parameter default value is not provided in the config, it will marked as required.
+        /// </summary>
+        /// <param name="entityName"></param>
```
ayush3797, line 136: nit: remove/populate.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                Mappings: null,
+                Relationships: null);
+
+            Dictionary<string, Entity> entities = new()
+            {
+                { entityName, entity }
+            };
+
+            RuntimeEntities runtimeEntities = new(entities);
+            OpenApiDocument openApiDocument = await OpenApiTestBootstrap.GenerateOpenApiDocument(
```
ayush3797, line 51: The code to create the openApiDocument in all the tests is same. Can be moved to a common helper method.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
abhishekkumams: i don't think so, because i was getting error to assign it before using.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+            Mappings: null,
+            Relationships: null);
+
+        Dictionary<string, Entity> entities = new()
+        {
+            { entityName, entity }
+        };
+
+        RuntimeEntities runtimeEntities = new(entities);
+        return await OpenApiTestBootstrap.GenerateOpenApiDocument(
```
seantleonard, line 171: probably need to fix this method name too: `GenerateOpenApiDocumentAsync(...)`

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+                    && param.Name is "title"
+                    && param.Schema.Type is "string"
+                    && param.Required is false));
+            }
+            else
+            {
+                // CountBookTitle doesn't require any query parameters.
+                Assert.IsTrue(operation.Parameters.IsNullOrEmpty());
+            }
+
```
seantleonard, line 146:
```suggestion
```

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// Validates that Query parameters are correctly generated for Stored Procedures.
+    /// Query parameters for stored procedures will be generated for each operation irrespective of its Operation Type.
+    /// If a parameter default value is not provided in the config, it will marked as required.
+    /// </summary>
+    /// <param name="entityName">The name of the entity.</param>
+    /// <param name="objectName">The name of the database object.</param>
+    /// <returns></returns>
+    [DataTestMethod]
+    [DataRow("CountBooks", "count_books", DisplayName = "StoredProcedure with no parameters")]
+    [DataRow("UpdateBookTitle", "update_book_title", DisplayName = "StoredProcedure with parameters")]
```
seantleonard, line 109: In the display name can you include what the expectation is?

StoredProcudure with no parameters... results in 0 created query params.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+                {
+                    // No query parameters for other OperationTypes. 
+                    Assert.IsFalse(operation.Parameters.Any(param => param.In is ParameterLocation.Query));
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Validates that Query parameters are correctly generated for Stored Procedures.
```
seantleonard, line 100:
```suggestion
    /// Validates that Query parameters are generated for Stored Procedures.
```

Are you validating more than the presence of the stored procedure query parameters? or are you performing additional validation on parameter metadata? if you are performing more than existence checks, can you elaborate here?

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// Query parameters for stored procedures will be generated for each operation irrespective of its Operation Type.
+    /// If a parameter default value is not provided in the config, it will marked as required.
+    /// </summary>
+    /// <param name="entityName">The name of the entity.</param>
+    /// <param name="objectName">The name of the database object.</param>
+    /// <returns></returns>
```
seantleonard, line 106:
```suggestion
```

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    private const string CUSTOM_CONFIG = "parameter-config.MsSql.json";
+    private const string MSSQL_ENVIRONMENT = TestCategory.MSSQL;
+
+    /// <summary>
+    /// Validates the path parameters are correctly generated for GET methods.
+    /// </summary>
+    /// <param name="entityName">The name of the entity.</param>
+    /// <param name="objectName">The name of the database object.</param>
+    /// <param name="entitySourceType">The source type of the entity.</param>
+    /// <returns></returns>
```
seantleonard, line 31:
```suggestion
```

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+                Assert.IsTrue(pathItem.Parameters.Count is 1);
+                OpenApiParameter pathParameter = pathItem.Parameters.First();
+                Assert.AreEqual("id", pathParameter.Name);
+                Assert.AreEqual(ParameterLocation.Path, pathParameter.In);
+                Assert.IsTrue(pathParameter.Required);
+            }
+            else
+            {
+                // Get All method with path /entityName, will have no path parameters.
+                Assert.IsTrue(pathItem.Parameters.Count is 0);
```
seantleonard, line 54: woudldn't `POST` hit this condition too? POST requests hit `/entityName` paths.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+/// </summary>
+[TestCategory(TestCategory.MSSQL)]
+[TestClass]
+public class ParameterValidationTests
+{
+    private const string CUSTOM_CONFIG = "parameter-config.MsSql.json";
+    private const string MSSQL_ENVIRONMENT = TestCategory.MSSQL;
+
+    /// <summary>
+    /// Validates the path parameters are correctly generated for GET methods.
```
seantleonard, line 26: What would constitute 'incorrectly' generated path parameters?

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.OpenApi.Models;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Azure.DataApiBuilder.Service.Tests.OpenApiIntegration;
+
+/// <summary>
+/// Integration tests validating correct path and query parameters are created
```
seantleonard, line 15: You may want to describe why we use the terminology "path parameter" -> OpenApi description document considers path components after `/entity` in  `/entity/pk_key/pk_value` as path parameters?

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /// <summary>
+    /// Validates that Query parameters are correctly generated for GET methods in Table/Views.
+    /// $select, $filter, $orderby, $first, $after are the query parameters.
+    /// </summary>
+    /// <param name="entityName">The name of the entity.</param>
+    /// <param name="objectName">The name of the database object.</param>
+    /// <param name="entitySourceType">The source type of the entity.</param>
+    /// <returns></returns>
```
seantleonard, line 66:
```suggestion
```

seantleonard (changes requested):
Suggestions on improved comments and updated async method names

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
abhishekkumams: right, POST will also hit this path

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
abhishekkumams: ahaha, right. I'll update the description.

abhishekkumams: But by incorrectly it could be like no pathParameter is there

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -431,6 +438,115 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
...
+
+        /// <summary>
+        /// Creates a list of OpenAPI parameters for querying tables and views.
+        /// The query parameters include $select, $filter, $orderby, $first, and $after, which allow the user to specify which fields to return,
+        /// filter the results based on a predicate expression, sort the results, and paginate the results.
+        /// </summary>
+        /// <returns>A list of OpenAPI parameters.</returns>
+        private static List<OpenApiParameter> CreateTableAndViewQueryParameters()
+        {
+            List<OpenApiParameter> parameters = new();
```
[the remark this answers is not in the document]
ayush3797: What was the resolution for this?

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        foreach (OpenApiPathItem pathItem in openApiDocument.Paths.Values)
+        {
+            foreach ((OperationType operationType, OpenApiOperation operation) in pathItem.Operations)
+            {
+                if (operationType is OperationType.Get)
+                {
+                    Assert.IsTrue(operation.Parameters.Count is 5);
+
+                    // Assert that it contains all the query parameters with appropriate type
+                    Assert.IsTrue(operation.Parameters.Any(param => param.In is ParameterLocation.Query && param.Name is "$select" && param.Schema.Type is "string"));
```
ayush3797, line 87: Can you use `.equals()` for string comparison?

ayush3797 (approved):
LGTM after resolving one nit. Also, unresolved a conversation seeking resolution.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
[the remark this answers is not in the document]
abhishekkumams: created a static list _tableAndViewQueryParameters, which will be used again and again.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+    /// <param name="entitySourceType">The source type of the entity.</param>
+    [DataTestMethod]
+    [DataRow("BooksTable", "books", EntitySourceType.Table, DisplayName = "Table with path parameter id")]
+    [DataRow("BooksView", "books_view_all", EntitySourceType.View, DisplayName = "View with path parameter id")]
+    public async Task ValidatePathParametersForTablesAndViews(string entityName, string objectName, EntitySourceType entitySourceType)
+    {
+        EntitySource entitySource = new(Object: objectName, entitySourceType, null, null);
+        OpenApiDocument openApiDocument = await GenerateOpenApiDocumentForGivenEntityAsync(entityName, entitySource);
+
+        foreach ((string pathName, OpenApiPathItem pathItem) in openApiDocument.Paths)
```
seantleonard, line 44: If the expected path is not created or `openApiDocument.Paths` is empty, this test would still pass because no assertions could be violated. Is that correct? You'll need to add another assert to ensure that your expected path is actually present (and only present the expected number of occurrences) in the list. by 'getting' the path object from the list of paths that matches your expected created path, you can then avoid having loop code in your test here. Are you expecting the loop to encounter >1 matching path to evaluate assertions?

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            else
+            {
+                // Get All and POST method with path /entityName, will have no path parameters.
+                Assert.IsTrue(pathItem.Parameters.Count is 0);
+            }
+        }
+    }
+
+    /// <summary>
+    /// Validates that Query parameters are generated for GET methods in Table/Views.
```
seantleonard, line 64:
```suggestion
    /// Validates that the default set of query parameters are generated for GET methods in Table/Views.
```
nit.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -431,6 +437,115 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
...
+
+            // Add $select query parameter
+            parameters.Add(
+                GetOpenApiQueryParameter(
+                    name: "$select",
+                    description: "A comma separated list of fields to return in the response.",
```
seantleonard, line 477: if you want to compare against another deployed microsoft api (Microsoft Graph), see this doc for the descriptions they use for query params: https://learn.microsoft.com/en-us/graph/query-parameters?tabs=http#odata-system-query-options

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+    [DataRow("BooksTable", "books", EntitySourceType.Table, DisplayName = "Table with query parameters")]
+    [DataRow("BooksView", "books_view_all", EntitySourceType.View, DisplayName = "View with query parameters")]
+    public async Task ValidateQueryParametersForTablesAndViews(string entityName, string objectName, EntitySourceType entitySourceType)
+    {
+        EntitySource entitySource = new(Object: objectName, entitySourceType, null, null);
+        OpenApiDocument openApiDocument = await GenerateOpenApiDocumentForGivenEntityAsync(entityName, entitySource);
+
+        foreach (OpenApiPathItem pathItem in openApiDocument.Paths.Values)
+        {
+            foreach ((OperationType operationType, OpenApiOperation operation) in pathItem.Operations)
```
seantleonard, line 80: instead of having loops, especially a double for loop, could you try to 'get' the expected path item from `openApiDocument.Paths.Values` and then 
validate you retrieve an object AND the expected number of objects.

Then, you can 'get' the expected tuple of operationType and operation to
- validate an expected number of tuples exists in `pathItem.Operations`
- perform your pre-existing assertions.
- assert that GET by ID does not contain query parameters irrelevant to single object query. (e.g. filter, orderby, first, after aren't applicable to get by id). 

For easier reading, I would argue that you have a separate test which validates zero query parameters exist for non-GET operation types.
- Test 1 -> `QueryParametersAddedForGEToperationOnTablesAndViews`
- Test 2 -> `QueryParametersExcludedFromNonReadOperationsOnTablesAndViews` -> this test would then explicitly validate lack of query params for POST, PUT, PATCH, DELETE, and GET (by ID)

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+                    // No query parameters for other OperationTypes. 
+                    Assert.IsFalse(operation.Parameters.Any(param => param.In is ParameterLocation.Query));
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Validates that Input parameters are generated for Stored Procedures with GET operation.
+    /// Input parameters for stored procedures will be generated for GET operation.
```
seantleonard, line 104:
```suggestion
```
redundant

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+    /// If a parameter default value is not provided in the config, it will marked as required.
```
seantleonard, line 105:
```suggestion
    /// If a parameter default value is not provided in the config, it will be marked as required.
```
nit

seantleonard: this may also change with Rohan's latest PR affecting input parameters validation and expected presence for stored procedures.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+    /// It also validates parameter metadatas like type, name, location, required, etc.
```
seantleonard, line 106:
```suggestion
    /// It also validates parameter metadata like type, name, location, required, etc.
```

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            // Adding a parameter default value.
+            parameterDefaults = new Dictionary<string, object> { { "title", "Test Title" } };
+        }
+
+        EntitySource entitySource = new(Object: objectName, EntitySourceType.StoredProcedure, parameterDefaults, null);
+        OpenApiDocument openApiDocument = await GenerateOpenApiDocumentForGivenEntityAsync(entityName, entitySource);
+
+        OpenApiPathItem pathItem = openApiDocument.Paths.First().Value;
+        foreach ((OperationType operationType, OpenApiOperation operation) in pathItem.Operations)
```
seantleonard, line 126: You'll probably want to validate that `pathItem.Operations` contains the expected element and expected number of elements. (This is assuming that test will pass when zero assertions are hit.) This would indicate a 'false-negative'. In other words the test would pass (negative for any issue detected) and we would have false confidence that the expected behavior occurred during open api doc generation. 

I realize this comment applies to other pre-existing tests outside the scope of this pr, and we'll have to go fix those.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            if (entityName.Equals("UpdateBookTitle"))
+            {
+                if (operationType is OperationType.Get)
```
seantleonard, line 130: you can avoid the 'if' statements by performing a 'get' on the collection that contains the object that you want. Then you don't risk a chance these conditions are never hit.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+                {
+                    Assert.IsTrue(operation.Parameters.Any(param =>
+                    param.In is ParameterLocation.Query
+                    && param.Name.Equals("id")
+                    && param.Schema.Type.Equals("number")
+                    && param.Required is true
+                    ));
+
+                    // Parameter with default value will be an optional query parameter.
+                    Assert.IsTrue(operation.Parameters.Any(param =>
```
seantleonard, line 140: `IEnumerable.Single()` will ensure that the expected param is present only once and throw if that expectation fails.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,187 @@
...
+                {
+                    // Input Parameters are supported only for GET requests
+                    // For Other requests requestBody is used.
+                    Assert.IsTrue(operation.Parameters.IsNullOrEmpty());
+                }
+            }
+            else
+            {
+                // CountBookTitle doesn't require any query parameters.
+                Assert.IsTrue(operation.Parameters.IsNullOrEmpty());
```
seantleonard, line 156: could this be its own test? that way you don't have a bunch of conditional logic in this test. Distinct tests which validate zero query parameters exist, when that scenario is expected. (e.g. non- GET requests and stored procs which have zero input params)

seantleonard (changes requested):
suggestions on making tests more readable and less prone to flakiness.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -431,6 +437,115 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
...
+                    required: false,
+                    type: "integer"
+                )
+            );
+
+            // Add $after query parameter
+            parameters.Add(
+                GetOpenApiQueryParameter(
+                    name: "$after",
+                    description: "A base64 encoded string that specifies the cursor position after which items should be returned.",
```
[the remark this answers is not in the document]
abhishekkumams: why opaque?

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
abhishekkumams: yes, with and  without /id/ in the path

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
abhishekkumams:
```
Test 2 -> QueryParametersExcludedFromNonReadOperationsOnTablesAndViews -> this test would then explicitly validate lack of query params for POST, PUT, PATCH, DELETE, and GET (by ID)
```
why GET (by ID) should not have query params? I agree some parameters won't makes sense with it, but parameters like $select does. I can see if there is a way to add a condition.

abhishekkumams: When you're retrieving a single resource by its ID, these parameters generally don't make sense because there's only one resource to retrieve, so there's no need to filter or sort the results, and it's usually desirable to return all of the resource's fields.

However, there might be some cases where $select could be useful in a getById request.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
abhishekkumams: I feel like little more descriptive is good, as it's creates less confusion on the usage.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
seantleonard: Yeah, agreed. I should have been more explicit, but what i meant with "GET by ID does not contain query parameters irrelevant to single object query" was that only query params that make sense like `$select` that you mention, should be applicable.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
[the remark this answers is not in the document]
seantleonard: Encoding type is an implementation detail and by specifying the type of encoding, we may be leading the developer to thing they can create/modify their `$after` string. The string should be used as provided by DAB's paging mechanisms and not built from scratch by the user.

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
```diff
@@ -0,0 +1,240 @@
...
+        Assert.IsTrue(openApiDocument.Paths.ContainsKey($"/{entityName}"));
+        Assert.IsTrue(openApiDocument.Paths.ContainsKey($"/{entityName}/id/{{id}}"));
+
+        // Assert that path without id has all the query parameters.
+        Assert.IsTrue(openApiDocument.Paths[$"/{entityName}"].Operations.ContainsKey(OperationType.Get));
+        Assert.AreEqual(5, openApiDocument.Paths[$"/{entityName}"].Operations[OperationType.Get].Parameters.Count);
+
+        // Asserting on all the parameters
+        List<OpenApiParameter> openApiParameters = openApiDocument.Paths[$"/{entityName}"].Operations[OperationType.Get].Parameters.ToList();
+        Assert.IsTrue(openApiParameters.Any(param => param.In is ParameterLocation.Query && param.Name.Equals("$select") && param.Schema.Type.Equals("string")));
```
Aniruddh25, line 91: do we have defined constants for all the query parameters $select, $filter etc?

Aniruddh25 (changes requested):
Some nits, and question on providing only $select for FindById..

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -284,7 +287,9 @@ private OpenApiPaths BuildPaths()
                 // The OpenApiResponses dictionary key represents the integer value of the HttpStatusCode,
                 // which is returned when using Enum.ToString("D").
                 // The "D" format specified "displays the enumeration entry as an integer value in the shortest representation possible."
+                // It will only contain $select query parameter to allow the user to specify which fields to return.
                 OpenApiOperation getOperation = CreateBaseOperation(description: GETONE_DESCRIPTION, tags: tags);
+                getOperation.Parameters = new List<OpenApiParameter>() { _tableAndViewSelectQueryParameter };
```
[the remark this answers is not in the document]
abhishekkumams: As discussed here: https://github.com/Azure/data-api-builder/pull/1769#discussion_r1380877727

we only provide `$select` for get by id as When we're retrieving a single resource by its ID, other parameters like `$filter`,`$items`, `$after` generally don't make sense as there's only one resource to retrieve, so there's no need to filter or sort the results, and it's usually desirable to return all of the resource's fields.

However, there might be some cases where $select could be useful in a getById request if the resource has a large number of fields and we only need a few of them

src/Service.Tests/OpenApiDocumentor/ParameterValidationTests.cs
abhishekkumams: added. Thanks for pointing out

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -431,6 +438,124 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
...
+        {
+            List<OpenApiParameter> parameters = new();
+
+            foreach ((string paramKey, ParameterDefinition parameterDefinition) in spDefinition.Parameters)
+            {
+                parameters.Add(
+                    GetOpenApiQueryParameter(
+                        name: paramKey,
+                        description: "Input parameter for stored procedure parameter",
+                        required: !parameterDefinition.HasConfigDefault,
```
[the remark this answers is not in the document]
abhishekkumams: As pointed out by Rohann, it's difficult to get the default values of arguments from metadata, so if default values are provided in the config, `HasConfigDefault` would be made true, hence it wouldn't require values to be explicitly provided when making a request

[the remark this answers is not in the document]
Aniruddh25: With this, we are still unable to catch the case if the config doesnt specify the value, whereas the stored proc definition has it set as optional.. we would end up saying this parameter is required. 

But thats not the case, I think the `required` argument should always be set to false, we cant really say if its really required or not. It may not be correctly indicative.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -42,6 +42,13 @@ public class OpenApiDocumentor : IOpenApiDocumentor
         public const string SCHEMA_OBJECT_TYPE = "object";
         public const string RESPONSE_ARRAY_PROPERTY = "array";
 
+        // Query parameters name
```
Aniruddh25, line 45: Aren't these constants already defined somewhere in the project?

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -493,8 +500,8 @@ private static List<OpenApiParameter> CreateTableAndViewQueryParameters()
             // Add $filter query parameter
             parameters.Add(
                 GetOpenApiQueryParameter(
-                    name: "$filter",
-                    description: "A predicate expression (an expression that returns a boolean value) using the entity's fields to retrieve a subset of the results.",
+                    name: FILTER_QUERY_PARAMETER_NAME,
+                    description: "A OData expression (an expression that returns a boolean value) using the entity's fields to retrieve a subset of the results.",
```
Aniruddh25, line 504:
```suggestion
                    description: "An OData expression (an expression that returns a boolean value) using the entity's fields to retrieve a subset of the results.",
```

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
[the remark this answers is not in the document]
Aniruddh25: Even if they are not applicable, do we allow them by definition or error out in validation? If we allow them, I think the OpenAPI doc should mention those. 
To me, $filter is also applicable, I want to getbyId but filter out if that item doesnt satisfy some condition.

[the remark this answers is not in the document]
abhishekkumams: The purpose of a getByID call is to retrieve a single, specific resource based on its unique identifier. what would we filter on a single resource? this would just not result in any output if the condition fails

also in case we would want to do it, we could achive this in getALL call as well by addding a filter on id with the condition. So, this wouldn't add any value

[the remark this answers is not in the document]
Aniruddh25: GetbyId only if that id belongs to a certain country for e.g.

If we are thinking of usefulness, GetALLfilterOnId = GetById anyway, so how is GetById useful? 
We cant determine what might be useful to developers, in some cases GetById with filter could be useful in others, GetAllFilteronId+Othercondition. 

Irrespective of usefulness, the OpenAPI doc is to provide information about what APIs are available. If we support $filter on getById today, we should mention it in OpenAPI. But if the engine itself blocks support for it already, then we dont need to implement it now.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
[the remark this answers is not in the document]
abhishekkumams: make sense, will update it.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -431,6 +438,115 @@ private OpenApiOperation CreateBaseOperation(string description, List<OpenApiTag
...
+        /// </summary>
+        /// <returns>A list of OpenAPI parameters.</returns>
+        private static List<OpenApiParameter> CreateTableAndViewQueryParameters()
+        {
+            List<OpenApiParameter> parameters = new();
+
+            // Add $select query parameter
+            parameters.Add(
+                GetOpenApiQueryParameter(
+                    name: RequestParser.FIELDS_URL,
```
aaronburtle, line 477: Do we want to couple this class directly to the RequestParser class's implementation of how it stores the names of query parameters? This will now make the RequestParser class responsible for providing those names to other classes. Do we want to have it doing that, and if so, should we just be exposing them as public fields?

Aniruddh25 (approved):
LGTM, thanks for addressing all the feedback!

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
abhishekkumams: i think request parser is the right place for the query names, because they're specifically used there, openAPI params needs the same queries, it's okay to source them from there.

PR #1782, opened 2023-10-06
Update dab.draft.schema.json

abhishekkumams:
## Why make this change?

- Currently our config and schema file had some differences, which is why even the config we support was showing to contain errors.

## What is this change?
- updating schema to reflect the correct supported config file.

## How was this tested?

- Online Tool: https://www.jsonschemavalidator.net/

### Before
image

### After
image


seantleonard:
on the topic of fixing json schema validation, should this PR address #1793? Can our `schema.json` file validate that no extraneous fields are provided?

seantleonard (changes requested):
as a pre-req to merging this, please open a PR updating the public documentation's description of GraphQL and REST settings for an entity. There is no reference to "enabled" property and there is also still reference to allowing the config:
```json
rest: "true",
graphql: "true"
```
this section also seems to be inaccurate: 
https://learn.microsoft.com/en-us/azure/data-api-builder/configuration-file#graphql-operation
```json
{
    "graphql": "true",
    "operation": "query"
  }
```

abhishekkumams:
> as a pre-req to merging this, please open a PR updating the public documentation's description of GraphQL and REST settings for an entity. There is no reference to "enabled" property and there is also still reference to allowing the config:
> 
> ```json
> rest: "true",
> graphql: "true"
> ```
> 
> this section also seems to be inaccurate: https://learn.microsoft.com/en-us/azure/data-api-builder/configuration-file#graphql-operation
> 
> ```json
> {
>     "graphql": "true",
>     "operation": "query"
>   }
> ```

Created a PR for the above in docs

aaronburtle (approved):
LGTM!

PR #1787, opened 2023-10-09
To file scoped namespace core

gmottajr:
## Why make this change?
GitHub open-source project collaboration requested.

## What is this change?

Move to file scoped namespace Core part 3


JerryNixon:
If we do this, should we do this in every `cs` file?

seantleonard:
> If we do this, should we do this in every `cs` file?

Our suggestion was to break up this change into multiple PR’s to avoid extra large single PR. This is one of a few.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -16,562 +16,561 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.OData.UriParser;
 
-namespace Azure.DataApiBuilder.Core.Resolvers
+namespace Azure.DataApiBuilder.Core.Resolvers;
+
```
seantleonard, line 20: nit: don't need extra line here.

seantleonard (changes requested):
one nit, then should be good to go!

src/Core/Resolvers/Factories/QueryManagerFactory.cs
```diff
@@ -8,111 +8,110 @@
...
+        {
+            throw new DataApiBuilderException(
+                $"{nameof(databaseType)}:{databaseType} could not be found within the config",
+                HttpStatusCode.BadRequest,
+                DataApiBuilderException.SubStatusCodes.DataSourceNotFound);
         }
 
+        return exceptionParser;
     }
+
```
aaronburtle, line 116: nit: I know you didnt add this empty line, but for consistency with the other files we should remove it.

aaronburtle (approved):
Looks good, and thanks for cleaning this up. Just an empty line here or there to remove.

Aniruddh25 (approved):
Looks good to merge, thanks for your patience while we review and improve our CI process. Has some conflicts that need to be resolved.

src/Core/Resolvers/Factories/QueryManagerFactory.cs
seantleonard: can resolve in follow up.

seantleonard:
@gmottajr can you resolve merge conflicts (take all latest from main) and then we can get this merged

PR #1801, opened 2023-10-10
REST and GraphQL Query Caching Design Doc

seantleonard:
## Why make this change?

- Proposed design of implementing query caching for DAB. This explicitly covers an in-memory cache, not a distributed cache.


docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
+
+## Why cache?
+
+Reduce the number of round-trip requests between Data API builder (DAB) and the configured back-end database. Reducing the number of round-trip requests results in the end-user experiencing quicker response times.
+
+### Objectives
+
+- Cache responses to read operations
+    - REST: GET requests REST
```
Aniruddh25, line 9:
```suggestion
    - REST: GET requests
```

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+### Cache entry
+
+#### Key
+
+`dataSourceName:queryText:queryParameters`
+
+The following example illustrates a real value created using the key schema listed above (new lines added for readability):
+
+```text
+7a07f92a-1aa2-4e2a-81d6-b9af0a25bbb6:
```
Aniruddh25, line 40: in case of multiple data sources, this may result in potentially heterogenous query texts - belonging to potentially different types of backends within the same cache, e.g. sql query for mssql v/s pgsql.

How does this affect caching for Cosmos db sources?

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+| View                | Yes                                                          |
+| Stored-Procedure    | Yes, only for stored procedures which query data. See below. |
+
+Stored procure entities must have the following configuration to allow for response caching.
+
+```json
+{
+  "entityName": {
+    "rest": {
+      "methods": [
```
Aniruddh25, line 215: So, by default data returned by stored procs is NOT cache compatible unless they change the methods/operation properties. 
Also, what happens if REST method is GET but graphql operation is not modified to query and vice versa..

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+
+__DAB will return unexpired cache entries when a database goes offline.__ Depending on the developer configured ttl for an entity, the cache entries will persist in an offline database scenario for the ttl defined duration. A cache entry is evicted after the ttl is surpassed.
+
+### Data in transit
+
+__Data returned by the database is received over the same TLS connection established to execute queries.__ The caching service described in this document is an in-memory cache and not a distributed cache. Consequently, the cache does not further transmit data returned by the database, by design. 
+
+### Access token metadata usage
+
+__Cache entries will not be created for database query text generated from database policies.__ This is because database policies directly affect values set within SQL Server's `-set-session-context` variables and is an implementation detail of database specific query executors. The caching service will be implemented independent of and before execution of query executor code.
```
Aniruddh25, line 98: If the query executor code decides whether to add database policies metadata in the query corresponding to a given request, how would we identify the request to be a cache miss if a similar request with NO database policies was already cached? 

Would that lead to incorrect cache hits preventing addition of additional predicates based on policies which would have otherwise happened if the cache were to be populated AFTER evaluation of policies?

Basically at which point do we check the cache for a hit or miss? - after complete generation of the query text OR right before executing the query?

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+      }
+    }
+  }
+}
+```
+
+| Property | JSON Data Type | Default Value | Description Value | Required |
+|----------|------------|-------------|---------------|----------|
+| enabled  | Boolean | false | Whether caching is enabled for read requests on this entity. |  Yes |
+| ttl      | Number | 30 | Number of seconds a cache entry is valid before cache eviction. Min: 1s Max: 24hrs/1440s |  Yes |
```
Aniruddh25, line 122: how can we make the fact that ttl is in seconds easily interpreted?

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+    publishers {
+      id,
+      name
+    }
+  }
+}
+```
+
+Because DAB resolves the request into a single database query, DAB is unable to query the cache for one half the request and query the database for the other half. 
+**DAB will only honor the cache configuration for the top-level referenced entity.** The developer will be responsible for ensuring their cache settings are consistently set across related entities to experience caching behavior consistent across entities.
```
Aniruddh25, line 152: What would be the guidance if the top level entity is mostly READ only but the related nested entities can be modified frequently?

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+For additional context see this document by Mark Nottingham titled "Caching Tutorial for Web Authors and Webmasters"
+> https://www.mnot.net/cache_docs/. 
+This document aims to explain web caching and usage of (HTTP) control caches. 
+
+Additionally, HTTP cache-control is not applicable to GraphQL due to (described later) the fact that most GraphQL requests would come via HTTP POST (not GET). More context can be found in article titled "GraphQL & Caching: The Elephant in the Room" by Marc-André Giroux.
+> https://www.apollographql.com/blog/backend/caching/graphql-caching-the-elephant-in-the-room/ 
+
+##### Overview
+
+The `cache-control`` header is defined as:
```
Aniruddh25, line 183:
```suggestion
The `cache-control` header is defined as: 
```

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+
+Due to how cache keys and values are created in the scope of this document, caching Data API builder responses in CDNs and browsers is not in scope. This means that `cache-control` **response** directives are not in scope. While the spec does state that the server can return `cache-control` if desired, DAB performs authentication and authorization on requests prior to accessing the in-memory cache. That same level of validation can't occur in the browser or CDN headers for security purposes. We can validate that ASP.NET core returns `private` or `no-store` as the `cache-control` header value.
+
+- Honor HTTP Header `cache-control` per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control with the following two **request** directives:
```
Aniruddh25, line 193: So, are we saying we honor `cache-control` headers specified ONLY in ANONYMOUS request directives?

Will the following directives override whatever is configured for the respective entity in question?

Is there a directive that enforces cache for a given request even if the config has it disabled for that entity (and vice-versa)?

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+
+For additional context see this document by Mark Nottingham titled "Caching Tutorial for Web Authors and Webmasters"
+> https://www.mnot.net/cache_docs/. 
+This document aims to explain web caching and usage of (HTTP) control caches. 
+
+Additionally, HTTP cache-control is not applicable to GraphQL due to (described later) the fact that most GraphQL requests would come via HTTP POST (not GET). More context can be found in article titled "GraphQL & Caching: The Elephant in the Room" by Marc-André Giroux.
```
Aniruddh25, line 178:
```suggestion
Additionally, HTTP cache-control (described later) is not applicable to GraphQL due to the fact that most GraphQL requests would come via HTTP POST (not GET). More context can be found in article titled "GraphQL & Caching: The Elephant in the Room" by Marc-André Giroux.
```

Aniruddh25 (approved):
Looks pretty good coverage of all considerations related to caching.. Left some questions and suggestions.

docs/design/query-caching.md
seantleonard: I should probably just move this entire section to a different doc. I didn't remove it to avoid losing the progress made. However, this section now begins with a disclaimer that `cache-control` will come with any work related to etags when we decide to take that on. This design will not implement caching mechanisms for upstream caches like cdns/browsers, even for anonymous get requests.

docs/design/query-caching.md
seantleonard: I would suggest, if necessary, we do this by updating the property name to `ttl-seconds`

docs/design/query-caching.md
seantleonard: at the moment, Cosmos is not included since Cosmos uses a distinct `CosmosQueryEngine`. I know this adds drift in functionality between SQL dbs and Cosmos, but would need more Cosmos involvement to help implement with their query engine.

docs/design/query-caching.md
seantleonard: Good point, I'll need to add logic which knows to skip cache for db policy affected requests. 

Your point also leads me to observe that the database policy/set session context query text generation should not be in the query executor. Ideally, that logic should be in the query /mutation engines or the sql query builders. The Query Executor should have a single purpose of executing the request metadata provided to it and not attempt to further modify the request metadata sent to it.

docs/design/query-caching.md
seantleonard: Then we'd need to suggest that when using GraphQL, you need to determine whether caching is right in that circumstance. 

Or we can look at skipping the cache altogether when nested entities. are present.  Currently, I'm proposing caching the entire SQL query because otherwise we'd need to modify our GraphQL resolver code to know when or when not to generate sql text and instead fetch from the cache, and then proceed to know how to piece cache results with non-cached results. DAB hasn't really been built to accommodate that which may require a much larger refactor / different design.

docs/design/query-caching.md
seantleonard: right.

so if there are conflicting REST: GET and GraphQL: Mutation for a stored proc,then the cache service may need to know the context of the request and check config accordingly: graphql mutation -> skip cache.

docs/design/query-caching.md
Aniruddh25: Would you please create a separate issue for CosmosDb sources, make Sajee and Neeraj aware of the same?

docs/design/query-caching.md
Aniruddh25: I vaguely remember @ayush3797 ran into some issues with adding session context query text to the query builders, @ayush3797 could you please provide reasoning behind why db policy query text generation was added to the executor instead of the builder?

docs/design/query-caching.md
Aniruddh25: makes sense, we should incorporate the guidance on which entities to cache when we provide documentation updates for the caching feature.

docs/design/query-caching.md
Aniruddh25: sounds good to me: `ttl-seconds`

docs/design/query-caching.md
Aniruddh25: Its good to have it here, what confused me was this section: 

`Honor HTTP Header `cache-control` per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control with the following two **request** directives:` -> which implies to me that we would honor the header for ANONYMOUS requests, but as per the disclaimer, are you saying that DAB will NOT honor this header in any case?

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+`dataSourceName:queryText:queryParameters`
+
+The following example illustrates a real value created using the key schema listed above (new lines added for readability):
+
+```text
+7a07f92a-1aa2-4e2a-81d6-b9af0a25bbb6:
+
+SELECT TOP 1 [dbo_books].[id] AS [id], [dbo_books].[title] AS [title], [dbo_books].[publisher_id] AS [publisher_id] FROM [dbo].[books] AS [dbo_books] WHERE [dbo_books].[id] = @param0 ORDER BY [dbo_books].[id] ASC FOR JSON PATH, INCLUDE_NULL_VALUES,WITHOUT_ARRAY_WRAPPER:
+
+{"@param0":{"Value":1,"DbType":11},"@param1":{"Value":"id","DbType":null},"@param2":{"Value":"title","DbType":null},"@param3":{"Value":"publisher_id","DbType":null}}
```
aaronburtle, line 44: Order here will matter, will this meta data always be ordered with Param0 first, etc? If not, it may need to be sorted to ensure you get consistent keys.

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### Updates
+
+Detect if the operation is a follow up read to a mutation. We can then map that follow up read generated query to a cache key entry and attempt to invalidate the cache entry. This would not allow us to expire all variations of read queries generated for the specific entity(s) because querystrings, role based access, and column based access may be cached with different keys. This behavior backs up our guidance to limit caching to rarely changing database objects/records because cache entries may be stale and fed back to the user.
+
+## Security and privacy
+
+### Data residency
+
+__Cache data is stored in-memory__ and remains in memory until evicted due to ttl rules, manual eviction, or DAB process shutdown. Cache data is not persisted to disk.
```
aaronburtle, line 85: Are there security concerns with holding information from the database locally in memory? Was just thinking about the possibility of potentially sensitive information living in memory for too long, potentially creating a vulnerability.

docs/design/query-caching.md
seantleonard: context is: https://github.com/Azure/data-api-builder/pull/1084#discussion_r1066278125

> Creating the query earlier won't simplify things - the sqltext is specific to the request intent (PUT/PATCH/INSERT) etc but the session context query is constant for all of them, so instead of separately adding it to the query at distributed places, we can do it at one single place. Also, I'll keep the name of the dictionary to something more generic like sessionParameters because its a good possibility we will have a similar functionality for postgres as well, not sure of MySql.

My interpretation is that there isn't a technical limitation to moving. Result would be more places in code which add the session parameter sql text. To me, that is a valid tradeoff for separation of concerns. During a refactor, we'd be able to better determine how much code de-duplication can occur.

docs/design/query-caching.md
seantleonard: > which implies to me that we would honor the header for ANONYMOUS requests, but as per the disclaimer, are you saying that DAB will NOT honor this header in any case?

Yes. With this in-memory cache, that is the case.

docs/design/query-caching.md
seantleonard: This is why cache config is per-entity. Entities with more sensitive fields can choose whether to opt in to using the cache. 

What vulnerabilities came to mind? If there is a compromise of the server hosting dab, then that's already a bigger problem. 

One case that is worth considering is either documenting or identifying a mitigating action to handle:
- database user represented by the connection string has a change of permissions, possibly has permissions removed, this should result in a purge of the cache which could then be rebuilt using most recent permissions. This would require event notification to dab, if one exists. 
- if database connection goes down, the cache is meant to kick in. Users who only want the db results in memory for a certain amount of time after the db goes down can configure their cache entry ttl accordingly.

docs/design/query-caching.md
seantleonard: Even if params are in order, there may be possibility that their values are not 1:1 across generated queries when users change up the order of fields in the request: 
- REST: change in order of composite PK, or order of query string params
- GraphQL: change in order of fields in selection set.

I observed that other caching services caveat cache performance is dependent on ordering of such values I mention above. We can document that too. We'd have to do perf tests to determine whether sorting keys and values adds overhead that defeats the purpose of a cache.

docs/design/query-caching.md
seantleonard: Also, session_context is all or nothing. all queries will get session context containing token claims when runtimeconfig's mssqloptions.SetSessionContext == true. so this is another boolean that will gate usage of the cache.

docs/design/query-caching.md
aaronburtle: that's a good point about other orderings being important too. I would say one thing though, I would have less sympathy for someone suffering from cache misses if theyre swapping around the ordering in their request on their own versus if the ordering is coming out differently just because of how some data structure is being iterated under the hood. not sure how easy it is to separate those, but we could certainly condition the cache performance on the user not messing with the order without me losing any sleep :)

docs/design/query-caching.md
aaronburtle: Ya, similar to the cases you describe my immediate concerns would be easy to describe scenario like some information that should not have been exposed on the database is exposed, or a policy for how the data is exposed changes, but dab has cached that information, so now the behavior between sending the query to the backend database yourself and what you see from dab is different, and in a way where dab is exposing information that the database does not want to. Maybe some way to flush the cache from the user would alleviate this?

seantleonard: >scenario like some information that should not have been exposed on the database is exposed

I'm not sure i follow. Is the scenario you mean one where the connection string is granted too much access, then the cache fills with that unintentionally exposed data, and then connection string access is fixed, and now the cache has data that should not be accessible?
In that case, e
- dab should have mechanism to purge cache when connection string changes. 
- data in the cache was allowed to be accessed and read at the time of access. If the protection level changes at the database level, users will not be able to get new fresh data. and are subject to the stale data. That's where  the developer will need to carefully manage their infrastructure.
 
> a policy for how the data is exposed changes

A policy in the dab runtime config permissions? in this case, all authN and authZ occur before cache access occurs. If a user meets all permissions requirements, their GraphQL or REST request will be translated into a SQL query and subsequently have the query executed against the database. cache access should be allowed then too. 

> some way to flush the cache from the user

We shouldn't allow a request to dab to clear the cache. the protection mechanisms that would need to be built to accomplish this would be covered by the developer's ability to restart dab.

aaronburtle: > I'm not sure i follow. Is the scenario you mean one where the connection string is granted too much access, then the cache fills with that unintentionally exposed data, and then connection string access is fixed, and now the cache has data that should not be accessible?

I was thinking something more simple. Consider a database that has mistakenly exposed users social security numbers into a column that can be read, but should not be so. Someone uses dab to connect to and read this information. Then the database is corrected to not expose this information. But dab will still have the information in its cache until the cache is updated to reflect the new state of the database.

Granted, a bad actor could have saved that information on their own, so perhaps this isn't a scenario we should worry about too much, but wanted to make sure it is something that is considered.

yorek:
Fantastic work! I would recommend to add support for enabling caching at global level, in the runtime section. I would also love to be able to have, always in the runtime section, a way to specify second-level cache. AFAIK we'll be using FusionCache which can also use Redis, so it will be great to be able to specify something like:

```
"runtime": {
    "cache": {
      "enabled": true,
      "ttl": "30",
      "redis" {
			"address": ""
			"key":""
		}	  
    }
```

This is not needed for 0.10 for sure, but would be nice to add it in 0.11 or 0.12 :)

docs/design/query-caching.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+The goals of the created key:
+- Uniquely store query text per data source to avoid collisions.
+- Agnostic towards GraphQL or REST request metadata.
+- Determinate size based on the size of the string generated.
+- Equivalent query metadata will map to the same cache value.
+  - This protects against reference type comparison where two distinct C# objects with same property values (request metadata) are treated as two separate cache entries instead of the same entry.
+
+- Rely on the underlying cache implementation to store the key/value pair in a concurrency compatible collection. The implementation defined collection will abstract away the work done to handle cache key hash collisions and appropriate bucketing.
```
severussundar, line 54: It would be helpful to note that the implementation would make use of `FusionCache`

docs/design/query-caching.md
severussundar: Let's say a user is hosting DAB in their own hosted setup and they are able to take a memory dump in that server. In this case, would they be able to in some way see the contents of the cached responses (that would essentially enable them to read contents of a database that they might not necessarily have access to).

[I acknowledge I'm not too familiar with dumps and could be something that we may not have to worry about. But, just wanted to bring it up so that we've thought through this possibility]

docs/design/query-caching.md
seantleonard: [#1884](https://github.com/Azure/data-api-builder/issues/1884)

docs/design/query-caching.md
seantleonard: Your having a level of access to the host machine such that you can acquire a memory dump means that you can see all DAB metadata such as the connection string, which you could use to access the db directly instead of by looking at cache data. The host machine should be hardened such that only permitted users can access for production related maintenance/troubleshooting.

docs/design/query-caching.md
seantleonard: updated entire section to say "NOT APPLICABLE TO IN MEMORY CACHE"

seantleonard:
@microsoft-github-policy-service rerun

PR #1811, opened 2023-10-12
Add design doc and diagram for Hot Reload feature

aaronburtle:
## Why make this change?

This is related to the following Hot Reload PR https://github.com/Azure/data-api-builder/pull/1726

and this issue https://github.com/Azure/data-api-builder/issues/67

It adds a design doc to get feedback and iterate on how to implement the Hot Reloading feature. Hot Reloading will be on going work as all of the needed features to support it entirely are added.

## What is this change?

Adds the design doc and a basic diagram for the Hot Reload feature.


docs/design/HotReloadDoc.md
```diff
@@ -0,0 +1,49 @@
...
+### Dependency Updating:
+
+Implement a mechanism to update services that are already registered during startup and depend on configuration settings that have changed. Authorization in particular is only configured on startup. In order to reset those authorization settings when we change the config file we need a way to have those particular settings reset and update. This is a detail that remains in design.
+
+# Potential Classes still in consideration
+
+## IConfigurableService Interface:
+
+### Responsibilities:
+Interface to be implemented by services that depend on configuration settings.
```
seantleonard, line 37: I don't agree with this approach.. instead, i think we could consider:
-Ioptionsmonitor to detect onchange event to then modify class internals
-OR ensure that when RuntimeConfigProvider is injected in class constructors, it's internal runtimeconfig is not set in the class. Instead set the runtimeconfig provider so that any usage of the runtime config can be fed from runtimeconfig provider (runtimeconfigprovider abstracts away the need to know about any hot loading config). This points us in the right direction but still requires more thought because our runtimeconfigprovider/runtimeconfig relationship is a service locator antipattern.

seantleonard (commented):
some thoughts

docs/design/HotReloadDoc.md
```diff
@@ -0,0 +1,49 @@
...
...de is set to `development`, in other words, in a local development scenario. To start we will implement this feature only working for the "runtime" section of the config file, but care must be made to implement this in a way that we do not block or have to change our implementation significantly for the other sections. This will allow users to update the configuration without the need to restart the service, providing more flexibility and efficiency within our application. After investigation it became apparent that hot reloading only 1 section of the config would be additional work without being able to make use of `IOptionsMonitor` to automatically update an individual section of the config that followed the `option` pattern. Because of the way we instantiate our object model of the configuration file, using `IOptionsMonitor`` in such a way was not possible at this time. We therefore will hot reload the entire config, but will lack the validation and implementation details for this to function inititally outside of Runtime section. Support for the other sections will be added incrementally, and when the entire hot reload feature is working, it will be documented for customer use.
```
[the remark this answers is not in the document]
aaronburtle: > > After investigation it became apparent that hot reloading only 1 section of the config would be additional work without being able to make use of `IOptionsMonitor` to automatically update an individual section of the config that followed the `option` pattern. Because of the way we instantiate our object model of the configuration file, using `IOptionsMonitor`` in such a way was not possible at this time.
> 
> I don't really follow. Are you saying you couldn't identify how to use `IOptionsMonitor` such that it only updates a single section of the config?

From my investigation, the `IOptionsMonitor` can not automatically reload the options from our config because of how we are creating our object model separately from a simply registering of the config file with the underlying .net framework. I was not able to get `IOptionsMonitor` to recognize or update when we change our config file in any of the ways that the documentation of tutorial describe.



> > We therefore will hot reload the entire config, but will lack the validation and implementation details for this to function inititally outside of Runtime section.
> 
> So the behavior of the first PR's worth of implementation will load a brand new file, but only use metadata from a single section (runtime) to rehydrate the configuration object maintained in the DAB engine at runtime?

So we will follow the same deserialization process as we do on startup, but without the ability to for example modify the schema hot reloading is not fully supported outside of the runtime section. Which is why we will not document this change until it is entirely supported.

docs/design/HotReloadDoc.md
```diff
@@ -0,0 +1,49 @@
...
...s is because the `RuntimeConfigProvider` has a `RuntimeConfigLoader`, and therefore the namespace that the `RuntimeConfigProvider` is within, `namespace Azure.DataApiBuilder.Core.Configurations` needs to reference the namespace that the `RuntimeConfigLoader` is within, `namespace Azure.DataApiBuilder.Config`. Since the `ConfigFileWatcher` will need a reference to the `RuntimeConfigProvider`, it will not be possible to have it live within the class or the project of the `RuntimeConfigLoader`, because this will create a circular dependancy. We therefore have decided to place the `ConfigFileWatcher` in the same namespace as the `RuntimeConfigProvider` and strongly couple the `ConfigFileWatcher` to the `RuntimeConfigProvider`, since it is a callback within the file watcher which will trigger a call into the provider, these classes are programmatically coupled already. This will then enable the file watcher to access the data that is needed from the file loader through the reference it has to the provider, since the provider has a file loader. Our file watcher then registers a trigger function which will itself call a function within the provider to handle the actual hot reload process.
```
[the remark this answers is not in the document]
aaronburtle: > > so that hot reloading is possible prior to the application being usable.
> 
> what do you mean by 'usable'?

Finished with initialization and ready for requests

docs/design/HotReloadDoc.md
aaronburtle: > I don't agree with this approach.. instead, i think we could consider: -Ioptionsmonitor to detect onchange event to then modify class internals
> 
> -OR ensure that when RuntimeConfigProvider is injected in class constructors, it's internal runtimeconfig is not set in the class.
> 
> * Instead set the runtimeconfig provider so that any usage of the runtime config can be fed from runtimeconfig provider (runtimeconfigprovider abstracts away the need to know about any hot loading config).
> * This points us in the right direction but still requires more thought because our runtimeconfigprovider/runtimeconfig relationship is a service locator antipattern.


After our discussion yesterday it seems that we might need multiple approaches to make sure that this work as we want with all of our services.

for example updating the jwt options seems to lend itself to something like `IOptionsMonitor` approach, but then the `OpenApiDocumentor` could be modified to use the `RuntimeConfigProvider` to get a `RuntimeConfig` rather than storing one internally to make sure it correctly reflects the hot reloads.

Aniruddh25 (approved):
Thanks for the detailed design and diagram! Left a few nits.

docs/design/HotReloadDoc.md
ayush3797, line 3:
```suggestion
The objective of this task is to enable "hot reloading" of the config file when we are not in a hosted scenario and we the Host Mode is set to `development`, in other words, in a local development scenario. To start, we will implement this feature only for the "runtime" section of the config file, but care must be taken to implement this in a way that we do not block or have to change our implementation significantly for the other sections. This will allow users to update the configuration without the need to restart the service, providing more flexibility and efficiency within our application. After investigation it became apparent that hot reloading only 1 section of the config would be additional work without being able to make use of `IOptionsMonitor` to automatically update an individual section of the config that followed the `option` pattern. Because of the way we instantiate our object model of the configuration file, using `IOptionsMonitor`` in such a way was not possible at this time. We therefore will hot reload the entire config, but will lack the validation and implementation details for this to function initially outside of Runtime section. Support for the other sections will be added incrementally, and when the entire hot reload feature is working, it will be documented for customer use.
```

docs/design/HotReloadDoc.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Current Setup:
+Currently, the application uses a RuntimeConfig class, which is a record. It holds a string as the `Schema`, and record types implemented as custom classes `DataSource`, `RuntimeOptions`, and `RuntimEntities`. The `RuntimeConfig` is populated through the `RuntimeConfigProvider` which makes use of the `RuntimeConfigLoader` to handle the reading of the configuration and deserializing into the needed objects. As mentioned, the `RuntimeConfig` consists of three major sections: Connection, Runtime, and Entities. While we consider all sections, to start this task, we will focus on the Runtime section.
+
+## Proposed Changes:
+We will modify the application to support hot reloading of the confiuration file when we are in a local development scenario. This means when the config file is changed while the application is running, and we are not in a hosted scenarior, and the config file is set to development mode, the objects populated to form the RuntimeConfig should be automatically updated to reflect the changes to the config file without requiring a service restart. Because the `RuntimeConfig` is a record, a new one will need to be created entirely.
```
ayush3797, line 9:
```suggestion
We will modify the application to support hot reloading of the confiuration file when we are in a local development scenario. This means that when the config file is changed and the application is running in development mode for a non-hosted scenario, the objects populated to form the RuntimeConfig should be automatically updated to reflect the changes to the config file without requiring a service restart. Because the `RuntimeConfig` is a record, a new one will need to be created entirely.
```

rohkhann (approved):

PR #1822, opened 2023-10-17
Add design doc for dab validate

abhishekkumams:
## Why make this change?

 design doc of implementing `dab validate`.


docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+This design specification outlines the process for validating the Data API builder config file used by a runtime engine. The validation process ensures that the config file is correctly formatted and contains all the required information for the runtime engine to function correctly.
+
+## Requirements
+
+The following requirements must be met by the validation process:
+
+- The validation process must check that the configuration file is correctly formatted and is in sync with the schema file.
+- If the configuration file is not valid, the validation process must 
+    - throw an exception in case of Metadata Initialization failure with a descriptive error message.
+    - Provide a list of errors that occur when the configuration file does not conform to the schema, including the line numbers where each error occurs.
```
seantleonard, line 14: How doable is this? I'd imagine that the first error encountered results in an exception and log event. are you just going to continue processing so that all errors can be emitted in one pass through?

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+- If the configuration file is valid, the validation process must return without throwing an exception.
+
+## Design
+
+The validation process will be implemented as a method that takes a file path as input and returns a boolean value indicating whether the file is valid or not. The method will perform the following steps:
+
+1. The validation can be initiated using the CLI command `dab validate`.
+2. This command will have an optional flag `-c` or `--config` to provide the config file. If not used it will pick the default config file `dab-config.{DAB_ENVIRONMENT}.json`, or `dab-config.json` if `DAB_ENVIRONMENT` is not set.
```
seantleonard, line 22: If deferring to the 'default' or 'environment var' resolved config, be sure to explicitly log the full path of the file being validated. or as much of the path as allowed by permissions granted to the process running dab validate.

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+> [!NOTE]
+> If the provided file is multi-data-source file, then all the files mentioned in `data-source-files` will be validated sequentially.
+3. Check that the file exists and is readable.
```
seantleonard, line 25: Check that which file exists? the multi data source linked files or the json.schema file?

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+4. Read the contents of the file into memory.
+5. Validate it against the schema file. We will use `NschemaJson.Net` nuget package.
+6. Deserialize the contents of the file into a configuration object.
```
seantleonard, line 28: Will this result in multiple `RuntimeConfig` objects? I believe the intent of multi data source is that the extra linked files just result in additional entities added to the primary runtimeconfig object's entities section. maybe @rohkhann can give some pointers here

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+7. Check that the configuration object contains all the required information.
+8. SetUp Metadata and check for any errors.
+9. If the configuration object is not valid, throw an exception with a descriptive error message.
+10. If the configuration object is valid, return true.
+
+The config file will be validated against the schema file mentioned in the config file property `$Schema`. If the config file is created by DAB CLI, it will contain the schema file link based on the DAB version.
+
+
+
+## Testing
```
seantleonard, line 38: I propose breaking this up into different PRs. At the very least:
1. Single config file validate
2. multi data source / multi config file validate.

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+3. Check that the file exists and is readable.
+4. Read the contents of the file into memory.
+5. Validate it against the schema file. We will use `NschemaJson.Net` nuget package.
+6. Deserialize the contents of the file into a configuration object.
+7. Check that the configuration object contains all the required information.
+8. SetUp Metadata and check for any errors.
+9. If the configuration object is not valid, throw an exception with a descriptive error message.
+10. If the configuration object is valid, return true.
+
+The config file will be validated against the schema file mentioned in the config file property `$Schema`. If the config file is created by DAB CLI, it will contain the schema file link based on the DAB version.
```
Aniruddh25, line 34: `$schema` is an optional property. If the config doesnt contain that property, what would be the behavior of the validate command?

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+
+The validation process will be tested using unit/integration tests that cover the following scenarios:
+
+Unit tests:
+- Valid configuration file
+- Invalid configuration file (missing required information)
+- Invalid configuration file (incorrect format)
+
+Integration tests:
+- Invalid Config containing invalid entities (to catch errors occuring while setting up metadata).
```
Aniruddh25, line 48:
```suggestion
- Invalid Config containing invalid entities (to catch errors occurring while setting up metadata).
```

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
...
+- If the configuration file is valid, the validation process must return without throwing an exception.
+
+## Design
+
+The validation process will be implemented as a method that takes a file path as input and returns a boolean value indicating whether the file is valid or not. The method will perform the following steps:
+
+1. The validation can be initiated using the CLI command `dab validate`.
+2. This command will have an optional flag `-c` or `--config` to provide the config file. If not used it will pick the default config file `dab-config.{DAB_ENVIRONMENT}.json`, or `dab-config.json` if `DAB_ENVIRONMENT` is not set.
+> [!NOTE]
+> If the provided file is multi-data-source file, then all the files mentioned in `data-source-files` will be validated sequentially.
```
Aniruddh25, line 24: What happens if the multiple data source files have different $schema properties used in them?

docs/design/dab-validate.md
```diff
@@ -0,0 +1,53 @@
+# Design Specification: Validating a Config File
+
+## Overview
+
+This design specification outlines the process for validating the Data API builder config file used by a runtime engine. The validation process ensures that the config file is correctly formatted and contains all the required information for the runtime engine to function correctly.
```
Aniruddh25, line 5: what do you mean by "correctly formatted"?
- compliant with the schema OR/(and)
- correct usage of spaces/tabs/indentation?

docs/design/dab-validate.md
abhishekkumams: make sense, will do that

docs/design/dab-validate.md
abhishekkumams: Thanks, will look into it, and lay it down in a separate PR.

docs/design/dab-validate.md
abhishekkumams: compliant with schema. will update to reflect it correctly.

docs/design/dab-validate.md
abhishekkumams: yes, instead of throwing the error we will collect it, but some places we would still throw the error like if the connection string is not valid, we won't continue to fetch entity metadata from the database.

Aniruddh25 (approved):

ayush3797 (approved):

Issue #1737, opened 2023-09-21
[Bug]: Merged configuration file always has the optional properties with their default values if they are not specified in the overridden file


Aniruddh25:
### What happened?

Example:
dab-config.json  => base config has graphql enabled = false.
```
{
“graphql”:  {
               “enabled”: false
               } 
}
```
 
dab-config.development.json => no graphql options provided explicitly
```
{
“graphql”: { }
}
```
 
We deserialize and initialize the “graphql” options with enabled = true for dab-config.development.json
 
Merging the base dab-config.json + the override- dab-config.development.json would lead to this:
```
{
“graphql”:  {
               “enabled”: true
               } 
}
```
Even though, correct merge should have retained the base config, i.e. `enabled =false. 
`

### Version

0.8.51

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

REST, GraphQL

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1823, opened 2023-10-17
[Configuration] Make optional object sub properties nullable

Aniruddh25:
## Why make this change?

- As per the JSON config schema, not all properties in the config file are mandatory.
- With #1402, in versions 0.8.49-0.8.52 those properties that are not explicitly set, were initialized with defaults before writing to the file system.
- However, config files generated using version 0.7.6 or lesser, may not have these properties initialized to their defaults when writing to the file system. Using config files generated with <= 0.7.6, to start engine with 0.8.49 would therefore lead to null reference exceptions when the optional properties are dereferenced.
- With PR #1684 the optional properties were initialized with their default values to avoid the null reference, any config file that is created or modified using the CLI results in a fully expanded file. This is not a problem when a single config is used but leads to undesired behavior when using the merge configuration file feature. A fully expanded higher precedence config file when merged with a base config file will override the values set for the optional properties in the base file with their default values if those properties were not explicitly defined in the higher precedence config file to begin with. 
In order to retain the base config file values for the optional properties, the higher precedence file would have to define every single optional value exactly the same as in the base file if they desire those to not be overridden. That defeats the purpose of providing the higher precedence file which should ideally only have those properties that need to be overriden. For more details with examples, please refer to #1737.

- This PR fixes this problem by allowing optional properties to be null and ensures there are no null reference exceptions as well  when starting engine with configs that don't have optional properties. 

## What is this change?
- Modify object model constructors to accept nullable arguments for optional properties - represented as fields in the record types. The scalar optional properties have not been made nullable yet, will have a follow up PR for those.
- All references to optional properties check for nulls
- During merge of config files, null + non-null values will pick up non-null values. After the merge, if the value is still null, they are considered as default values. To easily get default values, added some properties to `RuntimeConfig`.
- Default value of Host.Mode is `Production`. Keep it that way. Default of GraphQL.AllowIntrospection is true currently - will have another PR to determine it based on host mode.

## How was this tested?
- Removed optional properties from config and ensured engine could be started without them.
- [X] Unit Tests for deserialization and serialization of nullable optional properties see `RuntimeConfigLoaderJsonDeserializerTests`
- [X] In the merge configuration tests, in `TestHelper`, modify base config to contain a non-default value, env based config to not specify that value and confirm the merged config file has the base-config value. 

## Sample Request(s)

- Before fix, remove `runtime` section from config, run dab start with 0.8.49 leads to NullReferenceException.
- After the fix, starting engine is successful even when `runtime` section is absent in the config.

// IN THE BASE config file
MISSING = use default
{ empty } = use default

// IN THE OVERRIDDEN config file 
MISSING = use base
{ empty } = use base

Examples:
// all subproperties are missing
base: "user": { "first": "jerry", "last": "nixon" }
override: "user": { empty } 
merged: "user": { "first": "jerry", "last": "nixon" }

// some sub properties are missing 
base: "user": { "first": "jerry", "last": "nixon" }
override: "user": { "first": "jerry" } 
merged: "user": { "first": "jerry", "last": "nixon" }

// parent object property is missing altogether
base: "user": { "first": "jerry", "last": "nixon" }
override: "user": MISSING (not specified in override)
merged: "user": { "first": "jerry", "last": "nixon" }

## TO DO for 0.10-rc
- A followup PR to make scalar optional properties nullable too.
- A followup PR that makes specifying an explicit NULL imply Default values


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1574,7 +1574,7 @@ public async Task ValidateLocationHeaderFieldForPostRequests(EntitySourceType en
             HostOptions staticWebAppsHostOptions = new(null, AuthenticationOptions);
 
             RuntimeOptions runtimeOptions = configuration.Runtime;
-            RuntimeOptions baseRouteEnabledRuntimeOptions = new(runtimeOptions.Rest, runtimeOptions.GraphQL, staticWebAppsHostOptions, "/data-api");
+            RuntimeOptions baseRouteEnabledRuntimeOptions = new(runtimeOptions?.Rest, runtimeOptions?.GraphQL, staticWebAppsHostOptions, "/data-api");
```
seantleonard, line 1577: nit: the above line `configuration.Runtime` returns a nullable `RuntimeOptions`. so `RuntimeOptions runtimeOptions` should be `RuntimeOptions? runtimeOptions`

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -128,6 +128,51 @@ public void CheckCommentParsingInConfigFile()
...
+                                    ""database-type"": ""mssql"",
+                                    ""connection-string"": ""@env('test-connection-string')""
+                                    },
+                                ""entities"": { }");
+
+            TryParseAndAssertOnDefaults("{" + minJson + "}", out _);
+
+            // Test with an empty runtime property
+            minJson.Append(@", ""runtime"": ");
+            string emptyRuntime = minJson + "{ }}";
```
seantleonard, line 152:
```suggestion
            minJson.Append(@", ""runtime"": { } ");
            string emptyRuntime = minJson + "}";
```
Would adding the empty runtime object to the first line work? It's more readable that way.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -128,6 +128,51 @@ public void CheckCommentParsingInConfigFile()
...
+            minJsonWithHostSubProps.Append(@"{ ""cors"": { }, ""authentication"": { } } }");
+            string emptyHostSubProps = minJsonWithHostSubProps + "}";
+            TryParseAndAssertOnDefaults("{" + emptyHostSubProps, out _);
+
+            // Test with empty telemetry sub-properties
+            minJsonWithTelemetrySubProps.Append(@"{ ""application-insights"": { } } }");
+
+            string emptyTelemetrySubProps = minJsonWithTelemetrySubProps + "}";
+            TryParseAndAssertOnDefaults("{" + emptyTelemetrySubProps, out _);
+        }
```
seantleonard, line 174: I'm not following what the test is doing with empty object's `{ }` for the properties `telemetry` and `application-insights`. There is no reference in the `TryParseAndAssertOnDefaults(..)` to those properties.

Are you just ensuring that the properties set to empty object `{}` result in successful config parsing due to default values and nullable optional properties... as opposed to null ref exceptions?

seantleonard (commented):
Some questions to help me understand your testing approach and implications of making config properties nullable.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -122,10 +126,10 @@ public void TestInitializingRestAndGraphQLGlobalSettings()
...
         Assert.AreEqual(DatabaseType.MSSQL, runtimeConfig.DataSource.DatabaseType);
         Assert.IsNotNull(runtimeConfig.Runtime);
-        Assert.AreEqual("/rest-api", runtimeConfig.Runtime.Rest.Path);
-        Assert.IsFalse(runtimeConfig.Runtime.Rest.Enabled);
-        Assert.AreEqual("/graphql-api", runtimeConfig.Runtime.GraphQL.Path);
-        Assert.IsTrue(runtimeConfig.Runtime.GraphQL.Enabled);
+        Assert.AreEqual("/rest-api", runtimeConfig.Runtime.Rest?.Path);
+        Assert.IsFalse(runtimeConfig.Runtime.Rest?.Enabled);
+        Assert.AreEqual("/graphql-api", runtimeConfig.Runtime.GraphQL?.Path);
+        Assert.IsTrue(runtimeConfig.Runtime.GraphQL?.Enabled);
```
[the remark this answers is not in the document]
Aniruddh25: right, since the CLI command includes the option in the command line, the optional properties should not be null. And eventually give us the right sub property. If they are null, the test will fail

src/Cli.Tests/TestHelper.cs
```diff
@@ -990,7 +985,7 @@ public static Process ExecuteDabCommand(string command, string flags)
           ""graphql"": {
             ""path"": ""/graphql"",
             ""enabled"": true,
-            ""allow-introspection"": true
+            ""allow-introspection"": false
```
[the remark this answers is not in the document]
Aniruddh25: yes, as mentioned in the description:

>  In the merge configuration tests, in TestHelper, modify base config to contain a non-default value, env based config to not specify that value and confirm the merged config file has the base-config value.

src/Cli/Utils.cs
```diff
@@ -177,50 +177,6 @@ public static bool IsURIComponentValid(string? uriComponent)
...
-        //     "cors": {
-        //         "origins": [],
-        //         "allow-credentials": true
-        //     },
-        //     "authentication": {
-        //         "provider": "StaticWebApps"
-        //     }
-        // }
-        /// </summary>
-        public static HostOptions GetDefaultHostOptions(
```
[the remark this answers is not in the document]
Aniruddh25: right, function not referenced anywhere

src/Service.Tests/Configuration/ConfigurationTests.cs
Aniruddh25: the test project doesnt have "nullable" context enabled. So, I couldnt do the suggested. Enabling nullable context for test project will be a refactoring issue in itself.

It errors:

> Severity	Code	Description	Project	File	Line	Suppression State
> Error	CS8632	The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.		1576	Active

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
seantleonard: meant to delete this comment, this is not possible after looking at logic in the lines that follow.

Aniruddh25: I agree thats more readable - but Im building up the minJson by incrementally appending to it, so on line 156 - I add the sub properties to the `runtime` object. If I append the { } to `minJson`, would need to remove those first before adding the subproperties - wanted to avoid that.

seantleonard: Understood

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
Aniruddh25: yes, Im ensuring there are no null ref exceptions. Also added a test for checking disabled `AppInsights`

seantleonard (approved):

src/Config/Converters/GraphQLRuntimeOptionsConverterFactory.cs
```diff
@@ -25,12 +25,12 @@ private class GraphQLRuntimeOptionsConverter : JsonConverter<GraphQLRuntimeOptio
     {
         public override GraphQLRuntimeOptions? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
-            if (reader.TokenType == JsonTokenType.True)
+            if (reader.TokenType == JsonTokenType.True || reader.TokenType == JsonTokenType.Null)
```
aaronburtle, line 28:
```suggestion
            if (reader.TokenType is JsonTokenType.True || reader.TokenType is JsonTokenType.Null)
```

src/Config/Converters/GraphQLRuntimeOptionsConverterFactory.cs
```diff
@@ -25,12 +25,12 @@ private class GraphQLRuntimeOptionsConverter : JsonConverter<GraphQLRuntimeOptio
...
             {
                 return new GraphQLRuntimeOptions();
             }
 
-            if (reader.TokenType == JsonTokenType.Null || reader.TokenType == JsonTokenType.False)
+            if (reader.TokenType == JsonTokenType.False)
```
aaronburtle, line 33:
```suggestion
            if (reader.TokenType is JsonTokenType.False)
```

aaronburtle (approved):
Looks good! The improvements to the `RuntimeConfig` properties are very nice!

PR #1825, opened 2023-10-18
Handling missing environment variables cleaner for SWA

### LINKED ISSUE #1820: [Bug]: DAB shouldn't throw exceptions for scenarios that are ONLY applicable to file based configs
### What happened?

Exceptions thrown while parsing configuration for example:

https://github.com/Azure/data-api-builder/blob/038c17d57c6d0ac4406b090e5ca84a56732a573c/src/Config/Converters/StringJsonConverterFactory.cs#L79

should only be thrown when the configuration is sourced from a file. 

For configurations provided to the configuration endpoint - these exceptions should NOT be thrown. 

This issue to track down such exceptions that are ONLY applicable to file based configuration and disable throwing them for non-file based configs.

E.g. Exception
image

### Version

0.8.52

Azure SQL

Static Web Apps (SWA)

GraphQL

### Relevant log output

_No response_

### Code of Conduct

aaronpowell:
## Why make this change?

- Fixes #1820 

## What is this change?

- New test coverage for #1820
- Added a way to control error handling when env vars aren't set - default is left to throw, override to ignore

## How was this tested?

- [x] Integration Tests
- [ ] Unit Tests


src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        RuntimeConfigProvider configProvider = server.Services.GetService(typeof(RuntimeConfigProvider)) as RuntimeConfigProvider;
+        RuntimeConfig loadedConfig = configProvider.GetConfig();
+
+        Assert.AreEqual(config.Schema, loadedConfig.Schema);
+    }
+
+    [TestMethod("Testing that environment variables can be replaced at runtime not only when config is loaded."), TestCategory(TestCategory.COSMOSDBNOSQL)]
+    [DataRow(CONFIGURATION_ENDPOINT_V2)]
+    public async Task CanLoadConfigWithEnvironmentVariables(string configurationEndpoint)
```
Aniruddh25, line 57: Now that I think of it: do we really want to allow replacement of the environment variables at all if they exist when the config is sent to the endpoint? Before 0.8.49, we would attempt to replace the environment variables only when the source of config is the local file system. 

A side effect would be any default environment variable already set on the container will now be available for consumption - e.g. `@env('ASPNETCORE_ENVIRONMENT')`

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
```diff
@@ -0,0 +1,95 @@
...
+using static Azure.DataApiBuilder.Service.Tests.Configuration.ConfigurationEndpoints;
+using static Azure.DataApiBuilder.Service.Tests.Configuration.TestConfigFileReader;
+
+namespace Azure.DataApiBuilder.Service.Tests.Configuration;
+
+[TestClass]
+public class LoadConfigViaEndpointTests
+{
+    [TestMethod("Testing that missing environment variables won't cause runtime failure."), TestCategory(TestCategory.COSMOSDBNOSQL)]
+    [DataRow(CONFIGURATION_ENDPOINT_V2)]
```
Aniruddh25, line 23: Why no test for `CONFIG_ENDPOINT`? We still have that in use...

Aniruddh25 (changes requested):
Thanks for quick fix! Left some questions..

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
aaronpowell: If we don't want to do env replacement then it's a much simpler fix, as we can disable that code path entirely for the HTTP config setting. I assumed that SWA would have some things they send in, such as the connection info.

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
aaronpowell: My understanding was that that endpoint is deprecated so I didn't include it, but can add it pretty easily.

Aniruddh25: It will be deprecated only after this gets deployed successfully :) When that happens, we will remove the code for the old endpoint altogether.

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
```diff
@@ -53,7 +54,9 @@ public async Task CanLoadConfigWithMissingEnvironmentVariables(string configurat
     }
 
     [TestMethod("Testing that environment variables can be replaced at runtime not only when config is loaded."), TestCategory(TestCategory.COSMOSDBNOSQL)]
+    [DataRow(CONFIGURATION_ENDPOINT)]
     [DataRow(CONFIGURATION_ENDPOINT_V2)]
+    [Ignore("We don't want to environment variable substitution in Phoenix, but test is left in for if this changes.")]
```
Aniruddh25, line 59:
```suggestion
    [Ignore("We don't want to environment variable substitution in late configuration, but test is left in for if this changes.")]
```

Aniruddh25 (approved):
LGTM! Thanks for addressing this.

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
```diff
@@ -0,0 +1,98 @@
...
+        RuntimeConfig loadedConfig = configProvider.GetConfig();
+
+        Assert.AreNotEqual(config.Schema, loadedConfig.Schema);
+        Assert.AreEqual(Environment.GetEnvironmentVariable("schema"), loadedConfig.Schema);
+    }
+
+    [TestCleanup]
+    public void Cleanup()
+    {
+        Environment.SetEnvironmentVariable("schema", null);
```
seantleonard, line 96: Why clean if the env var is not there in the first place?

seantleonard: I see it is in the ignored test. So i guess this doesn't hurt.

seantleonard (changes requested):

src/Config/Converters/StringJsonConverterFactory.cs
```diff
@@ -76,10 +89,17 @@ private static string ReplaceMatchWithEnvVariable(Match match)
...
+            if (_replacementFailureMode == EnvironmentVariableReplacementFailureMode.Throw)
+            {
+                return envValue is not null ? envValue :
+                    throw new DataApiBuilderException(message: $"Environmental Variable, {envName}, not found.",
+                                                   statusCode: System.Net.HttpStatusCode.ServiceUnavailable,
+                                                   subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
+            }
+            else
+            {
+                return envValue ?? match.Value;
```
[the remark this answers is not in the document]
aaronpowell: Yep, that's the value it'll return

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -136,7 +137,8 @@ public bool TryGetLoadedConfig([NotNullWhen(true)] out RuntimeConfig? runtimeCon
         if (RuntimeConfigLoader.TryParseConfig(
                 configuration,
                 out RuntimeConfig? runtimeConfig,
-                replaceEnvVar: true))
+                replaceEnvVar: false,
+                replacementFailureMode: EnvironmentVariableReplacementFailureMode.Ignore))
```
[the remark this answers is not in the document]
aaronpowell: Oh yeah. That's probably going to be a failing test one the build issue is fixed 🤣

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
```diff
@@ -0,0 +1,98 @@
...
+    [DataRow(CONFIGURATION_ENDPOINT)]
+    [DataRow(CONFIGURATION_ENDPOINT_V2)]
+    public async Task CanLoadConfigWithMissingEnvironmentVariables(string configurationEndpoint)
+    {
+        TestServer server = new(Program.CreateWebHostFromInMemoryUpdateableConfBuilder(Array.Empty<string>()));
+        HttpClient httpClient = server.CreateClient();
+
+        RuntimeConfig config = ReadCosmosConfigurationFromFile() with { Schema = "@env('schema')" };
+
+        ConfigurationPostParametersV2 @params = new(
```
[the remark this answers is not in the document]
aaronpowell: Good point, I forgot there were different structures between v1 and v2 endpoints and that simply adding the `DataRow` wasn't enough.

I used `@params` as I couldn't think of a better variable name than `params`, but since that's a reserved word you can use an `@` prefix to define it. You can [read it in more detail on the docs](https://learn.microsoft.com/dotnet/csharp/language-reference/language-specification/lexical-structure#643-identifiers)

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
aaronpowell: True, it doesn't _need_ to be done since the test is ignored, but if we didn't have it, chances are we'd add the test back in but not add that in and it'd result in unexpected side effects.

src/Service.Tests/Configuration/LoadConfigViaEndpoint.cs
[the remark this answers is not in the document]
seantleonard: understood. I didn't doubt that it worked but a name like `configParams` would have been fine too. 

No need to change. resolving.

PR #1847, opened 2023-10-26
Ensure customers do not have to replicate stored proc parameters in the config.

rohkhann:
Stored procedure today, is trying to validate that number of params in config == number of params in sp definition. this makes it difficult for customer as they have to respecify all the parameters in the  sp in the config again. Our validation should only ever fail in two scenarios:
1. The customer provides inputs in the config that dont match whats there in the stored proc. 
2. The customer does not provide input in the query and there is no backing default in the config. (ideally if there was a default on the db stored proc we should have detected it but it seems very hard for sql). As a workaround, customer can just define that default in the config if they want. 

Difficulties in determining whether a stored procedure parameter has default values:
1. Documented by microsoft - 
image
https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-sql-modules-transact-sql?view=sql-server-ver16
2. There are some alternatives but they are flaky -https://copyprogramming.com/howto/get-default-value-of-stored-procedure-parameter 

We can try the SQL DOM approach but the focus on this pr is to ensure that irrespective of whether there is or is not a default parameter customers can query their endpoint. If there is no query parameter provided and nothing specified in config we error out (even if for now they have a default defined in the db.)

### DIFF
--- a/src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
+++ b/src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
@@ -4,6 +4,7 @@
 using System.Text.Json;
 using Azure.DataApiBuilder.Config.DatabasePrimitives;
 using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Service.GraphQLBuilder.Sql;
 using HotChocolate.Language;
 using HotChocolate.Types;
 using static Azure.DataApiBuilder.Service.GraphQLBuilder.GraphQLNaming;
@@ -37,24 +38,30 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder
             // which are needed because parameter and column names can differ.
             StoredProcedureDefinition spdef = (StoredProcedureDefinition)dbObject.SourceDefinition;
 
-            // Create input value definitions from parameters defined in runtime config.
-            if (entity.Source.Parameters is not null)
+            // Create input value definitions from parameters defined in stored proc.
+            if (spdef is not null)
             {
-                foreach (string param in entity.Source.Parameters.Keys)
+                foreach ((string param, ParameterDefinition definition) in spdef.Parameters)
                 {
                     // Input parameters defined in the runtime config may denote values that may not cast
                     // to the exact value type defined in the database schema.
                     // e.g. Runtime config parameter value set as 1, while database schema denotes value type decimal.
                     // Without database metadata, there is no way to know to cast 1 to a decimal versus an integer.
-                    string defaultValueFromConfig = entity.Source.Parameters[param].ToString()!;
-                    Tuple<string, IValueNode> defaultGraphQLValue = ConvertValueToGraphQLType(defaultValueFromConfig, parameterDefinition: spdef.Parameters[param]);
+
+                    IValueNode? valueNode = null;
+                    if (entity.Source.Parameters is not null && entity.Source.Parameters.TryGetValue(param, out object? value))
+                    {
+                        Tuple<string, IValueNode> defaultGraphQLValue = ConvertValueToGraphQLType(value.ToString()!, parameterDefinition: spdef.Parameters[param]);
+                        valueNode = defaultGraphQLValue.Item2;
+                    }
+
                     inputValues.Add(
                         new(
                             location: null,
                             name: new(param),
                             description: new StringValueNode($"parameters for {name.Value} stored-procedure"),
-                            type: new NamedTypeNode(defaultGraphQLValue.Item1),
-                            defaultValue: defaultGraphQLValue.Item2,
+                            type: new NamedTypeNode(SchemaConverter.GetGraphQLTypeFromSystemType(type: definition.SystemType)),
+                            defaultValue: valueNode,
                             directives: new List<DirectiveNode>())
                         );
                 }
--- a/src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
+++ b/src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
@@ -48,11 +48,11 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder
                     // e.g. Runtime config parameter value set as 1, while database schema denotes value type decimal.
                     // Without database metadata, there is no way to know to cast 1 to a decimal versus an integer.
 
-                    IValueNode? valueNode = null;
+                    IValueNode? defaultValueNode = null;
                     if (entity.Source.Parameters is not null && entity.Source.Parameters.TryGetValue(param, out object? value))
                     {
                         Tuple<string, IValueNode> defaultGraphQLValue = ConvertValueToGraphQLType(value.ToString()!, parameterDefinition: spdef.Parameters[param]);
-                        valueNode = defaultGraphQLValue.Item2;
+                        defaultValueNode = defaultGraphQLValue.Item2;
                     }
 
                     inputValues.Add(
@@ -61,7 +61,7 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder
                             name: new(param),
                             description: new StringValueNode($"parameters for {name.Value} stored-procedure"),
                             type: new NamedTypeNode(SchemaConverter.GetGraphQLTypeFromSystemType(type: definition.SystemType)),
-                            defaultValue: valueNode,
+                            defaultValue: defaultValueNode,
                             directives: new List<DirectiveNode>())
                         );
                 }
--- a/src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMsSql.verified.txt
+++ b/src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMsSql.verified.txt
@@ -2095,16 +2095,13 @@
       GetPublisher: {
         Source: {
           Object: get_publisher_by_id,
-          Type: stored-procedure,
-          Parameters: {
-            id: 1
-          }
+          Type: stored-procedure
         },
         GraphQL: {
           Singular: GetPublisher,
           Plural: GetPublishers,
           Enabled: true,
-          Operation: Mutation
+          Operation: Query
         },
         Rest: {
           Methods: [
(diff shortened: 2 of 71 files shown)


aaronburtle:
Changes look good, but I did not see a negative case to provide test coverage for the cases you mention in the description about validation failing in the config validation tests or in the sp builder tests. Can you double check if we have a test of this, and if not add an issue to track adding if not adding to this PR itself?

aaronburtle (approved):

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -37,24 +38,30 @@ public static class GraphQLStoredProcedureBuilder
             // which are needed because parameter and column names can differ.
             StoredProcedureDefinition spdef = (StoredProcedureDefinition)dbObject.SourceDefinition;
 
-            // Create input value definitions from parameters defined in runtime config.
-            if (entity.Source.Parameters is not null)
+            // Create input value definitions from parameters defined in stored proc.
+            if (spdef is not null)
             {
-                foreach (string param in entity.Source.Parameters.Keys)
+                foreach ((string param, ParameterDefinition definition) in spdef.Parameters)
```
seantleonard, line 44: So by iterating over the resolved Stored procedure definition retreived from the db, you can handle the case when the customer provides default values for the input param in the runtime config that dont match whats there in the stored proc?

seantleonard (changes requested):
Need tests to validate scenario that you are trying to accommodate. 
- If input parameter in db schema has a default value and 1) none is supplied in request 2) no default value defined in runtime config -> what is the expected outcome? 

What kind of test do you think could validate your failure case proposal:
> The customer does not provide input in the query and there is no backing default in the config. (ideally if there was a default on the db stored proc we should have detected it but it seems very hard for sql). As a workaround, customer can just define that default in the config if they want.

Aniruddh25 (changes requested):
Need validation of at least the params that are specified

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -37,24 +38,30 @@ public static class GraphQLStoredProcedureBuilder
...
                 {
                     // Input parameters defined in the runtime config may denote values that may not cast
                     // to the exact value type defined in the database schema.
                     // e.g. Runtime config parameter value set as 1, while database schema denotes value type decimal.
                     // Without database metadata, there is no way to know to cast 1 to a decimal versus an integer.
-                    string defaultValueFromConfig = entity.Source.Parameters[param].ToString()!;
-                    Tuple<string, IValueNode> defaultGraphQLValue = ConvertValueToGraphQLType(defaultValueFromConfig, parameterDefinition: spdef.Parameters[param]);
+
+                    IValueNode? valueNode = null;
+                    if (entity.Source.Parameters is not null && entity.Source.Parameters.TryGetValue(param, out object? value))
```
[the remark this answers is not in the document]
ayush3797: Instead of retrieving the default value, is it not sufficient to just have the metadata which tells whether there is a default value available for a stored proc param? In that case, we can just invoke the SP without passing in a value for that param in case no value was specified by the user.

ayush3797:
Ideally, I think these cases make sense to me:

1. The parameter does not have a default value specified in the stored proc definition and in the config:
-> User needs to provide value for that param
2. The parameter does not have a default value specified in the SP definition but has a default value specified in the config:
-> User does not need to provide value for the param and we will use the value from the config.
3. The parameter has a default value specified in the SP definition and in the config (do we want to allow this scenario?): 
-> User does not need to provide value for the param and we will use the value from the config.
4. The parameter has a default value specified in the SP definition but not in the config:
-> User does not need to provide value for the param and we will use the value from the SP definition. 

For doing this, we would need the metadata about whether a param has default value in the SP definition or not.

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
rohkhann: This is already validated while creating the stored procedure definition here: 
image

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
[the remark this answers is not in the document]
rohkhann: Ive added description in the pr header on why this is very hard to do. The existing solutions are flaky and it requires proper review. However, this pr main aim is to unblock you from having to readd every parameter from your sql db into your config even without lack of default values. Now atleast customer can respecify the default value in the config or have it in their frontend code and atleast make queries without having to repeat every parameter in the config.

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -37,24 +38,30 @@ public static class GraphQLStoredProcedureBuilder
...
+
                     inputValues.Add(
                         new(
                             location: null,
                             name: new(param),
                             description: new StringValueNode($"parameters for {name.Value} stored-procedure"),
-                            type: new NamedTypeNode(defaultGraphQLValue.Item1),
-                            defaultValue: defaultGraphQLValue.Item2,
+                            type: new NamedTypeNode(SchemaConverter.GetGraphQLTypeFromSystemType(type: definition.SystemType)),
+                            defaultValue: valueNode,
```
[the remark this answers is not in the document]
rohkhann: config vs sp validation is already done at time of creation of the stored proc. When we reach here every parameter in the config is validated to exist in the stored proc: 
image

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -37,24 +38,30 @@ public static class GraphQLStoredProcedureBuilder
...
+                foreach ((string param, ParameterDefinition definition) in spdef.Parameters)
                 {
                     // Input parameters defined in the runtime config may denote values that may not cast
                     // to the exact value type defined in the database schema.
                     // e.g. Runtime config parameter value set as 1, while database schema denotes value type decimal.
                     // Without database metadata, there is no way to know to cast 1 to a decimal versus an integer.
-                    string defaultValueFromConfig = entity.Source.Parameters[param].ToString()!;
-                    Tuple<string, IValueNode> defaultGraphQLValue = ConvertValueToGraphQLType(defaultValueFromConfig, parameterDefinition: spdef.Parameters[param]);
+
+                    IValueNode? valueNode = null;
```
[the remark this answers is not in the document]
rohkhann: fixed.

Aniruddh25:
> Ideally, I think these cases make sense to me:

1. The parameter does not have a default value specified in the stored proc definition and in the config:
-> User needs to provide value for that param
2. The parameter does not have a default value specified in the SP definition but has a default value specified in the config:
-> User does not need to provide value for the param and we will use the value from the config.
3. The parameter has a default value specified in the SP definition and in the config (do we want to allow this scenario?): 
-> User does not need to provide value for the param and we will use the value from the config.
4. The parameter has a default value specified in the SP definition but not in the config:
-> User does not need to provide value for the param and we will use the value from the SP definition. 

For doing this, we would need the metadata about whether a param has default value in the SP definition or not. 

Looks like the problem is SP definition cannot reliably provide us information whether the parameter has a default or not + we dont know what the value would be.

Aniruddh25:
LGTM. Thanks for fixing this! Would be nice if we can have a follow up for checking SP metadata using sys.sql_modules or OBJECT_DEFINITION as the docs suggest for finding if the parameter has default or not.

Aniruddh25 (approved):

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -374,6 +375,22 @@ public async Task QueryAgainstSPWithOnlyTypenameInSelectionSet()
             await QueryAgainstSPWithOnlyTypenameInSelectionSet(dbQuery);
         }
 
+        /// <summary>
+        /// Checks failure on providing arguments with no default in runtimeconfig.
+        /// </summary>
```
seantleonard, line 380: Can you elaborate on the state of the system under test in the comment:
1. Which fields are expected to be present in the request for the request to pass? 
2. What runtime config properties are included/excluded to ensure this test is in the proper state? e.g. no default values specified in runtime config 
3. What are the relevant db schema traits of the stored procedure under test? e.g. defines X number of default value params which are named a, b, and c.

I think one more test is warranted where the GraphQL request includes the a value for the field, where the field:
- has default value defined in db
- no default value defined in config

Result: successful response.

rohkhann: This test seems to be already confirmed by this test case right: 
image

seantleonard: The test in the screenshot may have unintentionally covered the use case I mentioned. 

If that pre-existing test has its underlying entity swapped out, then covering this test case may no longer apply, even though the test in the screenshot still accomplishes what it claims to be testing :

> The GraphQL request includes a value for the field, where the field:
> 
> - has default value defined in db
> - no default value defined in config

-> Action: Update the comment on that test `TestStoredProcedureQueryForGettingSingleRow(string dbQuery)` to elaborate on the state of the system under test like in my first comment of this thread.

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMsSql.verified.txt
```diff
@@ -2095,16 +2095,13 @@
...
-            id: 1
-          }
+          Type: stored-procedure
         },
         GraphQL: {
           Singular: GetPublisher,
           Plural: GetPublishers,
           Enabled: true,
-          Operation: Mutation
+          Operation: Query
```
abhishekkumams, line 2104: why this change?

seantleonard (approved):
Pending response to Abhishek's comment.

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMsSql.verified.txt
rohkhann: GetPublisher is a getter method returning results and hence should be marked as query as typically mutations are used to create/update/delete from db.

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
rohkhann: Makes sense. added comment.

PR #1851, opened 2023-10-30
Change the form of our NextLink used for pagination cursors

aaronburtle:
## Why make this change?

Updates the form of our NextLink cursor used for pagination. This only updates the part used for the response, which is then used to continue the paginated requests. Matches the desired form.

## What is this change?

Rather than using part of the `SqlQueryStructures` to form the cursor, we create an object that is only used specifically for the NextLink with respect to the response/request. This means we don't need all of the fields that are included in the pagination columns, and that the information contained in this object is really unrelated to the query structure.

We therefore introduce a new class in our pagination utilities which can be used to facilitate the serialization and deserialization of the NextLink cursor in the response/request. Using an anonymous object was considered, but we re-use this object in enough places, and it holds complex enough data, that it will be easier to understand and be more reusable if we define a new class here inside of our pagination utilities.

The Json structure that is contained within the response/request that we (de)serialize with this new class is therefore modified.

Previous

```
[{"Value":0,"Direction":0,"TableSchema":"foo","TableName":"bar","ColumnName":"baz"}]")
```

Updated

```
[{"EntityName":"foobar","FieldName":"qux","Value":0,"Direction":0}]
```

Complete NextLink without being decoded

```
"nextLink": "https://localhost:5001/rest/Book?$first=1&$after=W3siVmFsdWUiOjEsIkRpcmVjdGlvbiI6MCwiVGFibGVTY2hlbWEiOiJkYm8iLCJUYWJsZU5hbWUiOiJib29rcyIsIkNvbHVtbk5hbWUiOiJpZCJ9XQ%3D%3D"
```
## How was this tested?

Modified our current test suite to match the new format.

## Sample Request(s)

and paginated request should work as long as it uses or generated a cursor for pagination.

one such example for when REST has a path of /rest against the Book entity.

https://localhost:5001/rest/Book?$first=1

### DIFF
--- a/src/Core/Resolvers/SqlPaginationUtil.cs
+++ b/src/Core/Resolvers/SqlPaginationUtil.cs
@@ -310,8 +310,6 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                     paginationCursorColumnsForQuery.Add(pageColumn);
                     // holds exposed name mapped to exposed pagination column
                     exposedFieldNameToBackingColumn.Add(column.FieldName, pageColumn);
-                    // overwrite with backing column's name for query generation
-                    column.FieldName = backingColumnName;
                 }
 
                 // verify that primary keys is a sub set of after's column names
@@ -324,11 +322,11 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                     // which will get the exposed name for safe messaging in the response.
                     // Since we are looking for primary keys we expect these columns to
                     // exist.
-                    string safePK = GetExposedColumnName(entityName, pk, sqlMetadataProvider);
-                    if (!exposedFieldNameToBackingColumn.ContainsKey(safePK))
+                    string exposedFieldName = GetExposedColumnName(entityName, pk, sqlMetadataProvider);
+                    if (!exposedFieldNameToBackingColumn.ContainsKey(exposedFieldName))
                     {
                         throw new DataApiBuilderException(
-                            message: $"Pagination token is not well formed because it is missing an expected field: {safePK}",
+                            message: $"Pagination token is not well formed because it is missing an expected field: {exposedFieldName}",
                             statusCode: HttpStatusCode.BadRequest,
                             subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
                     }
@@ -340,18 +338,18 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                 SqlQueryStructure structure = paginationMetadata.Structure!;
                 foreach (OrderByColumn column in structure.OrderByColumns)
                 {
-                    string columnName = GetExposedColumnName(entityName, column.ColumnName, sqlMetadataProvider);
+                    string exposedFieldName = GetExposedColumnName(entityName, column.ColumnName, sqlMetadataProvider);
 
-                    if (!exposedFieldNameToBackingColumn.ContainsKey(columnName) ||
-                        exposedFieldNameToBackingColumn[columnName].Direction != column.Direction)
+                    if (!exposedFieldNameToBackingColumn.ContainsKey(exposedFieldName) ||
+                        exposedFieldNameToBackingColumn[exposedFieldName].Direction != column.Direction)
                     {
                         // REST calls this function with a non null sqlMetadataProvider
                         // which will get the exposed name for safe messaging in the response.
                         // Since we are looking for valid orderby columns we expect
                         // these columns to exist.
-                        string safeColumnName = GetExposedColumnName(entityName, columnName, sqlMetadataProvider);
+                        string exposedOrderByFieldName = GetExposedColumnName(entityName, column.ColumnName, sqlMetadataProvider);
                         throw new DataApiBuilderException(
-                            message: $"Could not match order by column {safeColumnName} with a column in the pagination token with the same name and direction.",
+                            message: $"Could not match order by column {exposedOrderByFieldName} with a column in the pagination token with the same name and direction.",
                             statusCode: HttpStatusCode.BadRequest,
                             subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
                     }
--- a/src/Core/Resolvers/SqlPaginationUtil.cs
+++ b/src/Core/Resolvers/SqlPaginationUtil.cs
@@ -295,7 +295,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                     if (backingColumnName is null)
                     {
                         throw new DataApiBuilderException(
-                            message: $"Cursor for Pagination Predicates is not well formed, {column.FieldName} is not valid.",
+                            message: $"Pagination token is not well formed because {column.FieldName} is not valid.",
                             statusCode: HttpStatusCode.BadRequest,
                             subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
                     }
@@ -328,7 +328,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                     if (!exposedFieldNameToBackingColumn.ContainsKey(safePK))
                     {
                         throw new DataApiBuilderException(
-                            message: $"Cursor for Pagination Predicates is not well formed, missing primary key column: {safePK}",
+                            message: $"Pagination token is not well formed because it is missing an expected field: {safePK}",
                             statusCode: HttpStatusCode.BadRequest,
                             subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
                     }
--- a/src/Core/Resolvers/SqlPaginationUtil.cs
+++ b/src/Core/Resolvers/SqlPaginationUtil.cs
@@ -121,7 +121,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         /// continue to the next page. These can then be used to form the pagination
         /// columns that will be needed for the actual query.
         /// </summary>
-        protected class NextLinkColumn
+        protected class NextLinkField
         {
             public string EntityName { get; set; }
             public string FieldName { get; set; }
@@ -129,7 +129,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             public string? ParamName { get; set; }
             public OrderBy Direction { get; set; }
 
-            public NextLinkColumn(
+            public NextLinkField(
                 string entityName,
                 string fieldName,
                 object? fieldValue,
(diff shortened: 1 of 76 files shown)


abhishekkumams (approved):

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -234,45 +260,56 @@ public static JsonDocument CreatePaginationConnectionFromJsonDocument(JsonDocume
...
-                                                       statusCode: HttpStatusCode.BadRequest,
-                                                       subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                        throw new DataApiBuilderException(
+                            message: $"Cursor for Pagination Predicates is not well formed, {column.FieldName} is not valid.",
+                            statusCode: HttpStatusCode.BadRequest,
+                            subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
                     }
 
+                    PaginationColumn pageColumn = new(
+                        tableName: "",
```
severussundar, line 302: When parsing the $after string, is it okay to not populate the actual table and schema name for further processing?

aaronburtle: it turns out that we don't actually generate the table name in the query from this information. we get the table name from the lookup we do from the request information itself.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -234,45 +260,56 @@ public static JsonDocument CreatePaginationConnectionFromJsonDocument(JsonDocume
...
+            string afterJsonString,
+            PaginationMetadata paginationMetadata,
+            ISqlMetadataProvider sqlMetadataProvider,
+            string entityName,
+            RuntimeConfigProvider runtimeConfigProvider
+            )
         {
-            IEnumerable<PaginationColumn>? after;
+            List<PaginationColumn>? after = new();
+            IEnumerable<NextLinkColumn>? requestAfter;
```
seantleonard, line 272: can you give this variables more descriptive names?

src/Core/Resolvers/SqlPaginationUtil.cs
seantleonard: If that is the case, then why do we need `NextLinkColumn.EntityName` ? since the entity name will be present in the request URL anyway?

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -234,45 +260,56 @@ public static JsonDocument CreatePaginationConnectionFromJsonDocument(JsonDocume
...
+                        tableName: "",
+                        tableSchema: "",
+                        columnName: backingColumnName,
+                        value: column.Value,
+                        paramName: column.ParamName,
+                        direction: column.Direction);
+                    after.Add(pageColumn);
                     // holds exposed name mapped to exposed pagination column
-                    afterDict.Add(column.ColumnName, column);
+                    afterDict.Add(column.FieldName, pageColumn);
```
seantleonard, line 310: can you give a more descriptive name to `afterDict` ? i get that the name already existed before this pr, but along with your other collections with the `after` name, then these all get confusing about what their purpose is.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -115,6 +115,34 @@ public static JsonDocument CreatePaginationConnectionFromJsonDocument(JsonDocume
...
+            public NextLinkColumn(
+                string entityName,
+                string fieldName,
+                object? value,
+                string? paramName = null,
+                OrderBy direction = OrderBy.ASC)
+            {
+                EntityName = entityName;
+                FieldName = fieldName;
+                Value = value;
```
seantleonard, line 140: is there a better name for this? either more descriptive or a comment describing what 'value' represents.

seantleonard (changes requested):
I appreciate your efforts to update the nextLink. I like the new format. To be in a good state to merge, please add more comments for new classes, more descriptive variable names.

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: that's there for clarification, it was what davide suggested, and it makes sense to include since the information is safe to expose and could be helpful if youre tracking down an issue related to pagination.

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: i changed it to FieldValue, but this is just the value of the column, a database row or record being a set of columns.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -288,9 +325,10 @@ RuntimeConfigProvider runtimeConfigProvider
                     string safePK = GetExposedColumnName(entityName, pk, sqlMetadataProvider);
                     if (!afterDict.ContainsKey(safePK))
                     {
-                        throw new DataApiBuilderException(message: $"Cursor for Pagination Predicates is not well formed, missing primary key column: {safePK}",
-                                                       statusCode: HttpStatusCode.BadRequest,
-                                                       subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                        throw new DataApiBuilderException(
+                            message: $"Cursor for Pagination Predicates is not well formed, missing primary key column: {safePK}",
```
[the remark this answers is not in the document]
aaronburtle: why do we not want to give the exposed name of the missing primary key?

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: cleaned up the names

abhishekkumams:
can you paste screenshot of what value it is returning. Is it returning Base64 encoded cursor or next url link that we can directly paste?

src/Core/Resolvers/SqlPaginationUtil.cs
Aniruddh25: is the tableName not required even when `PaginationColumn` is used for the GraphQL case?

Aniruddh25 (approved):
Looks good, thanks for fixing this so quickly, need a check for GraphQL..

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: That's right, it isn't needed for either REST or GraphQL.

src/Core/Resolvers/SqlPaginationUtil.cs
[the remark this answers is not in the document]
Aniruddh25: perhaps mention it as missing an expected field with the exposed name of the field. Not necessarily describe its the primary key/column.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -292,7 +295,7 @@ RuntimeConfigProvider runtimeConfigProvider
                     if (backingColumnName is null)
                     {
                         throw new DataApiBuilderException(
-                            message: $"Cursor for Pagination Predicates is not well formed, {column.FieldName} is not valid.",
+                            message: $"Pagination token is not well formed because {column.FieldName} is not valid.",
```
seantleonard, line 298: This should also be the exposed/mapped field name and not the backingColumnName in this error message.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -322,10 +325,10 @@ RuntimeConfigProvider runtimeConfigProvider
                     // Since we are looking for primary keys we expect these columns to
                     // exist.
                     string safePK = GetExposedColumnName(entityName, pk, sqlMetadataProvider);
-                    if (!afterDict.ContainsKey(safePK))
+                    if (!exposedFieldNameToBackingColumn.ContainsKey(safePK))
```
seantleonard, line 328: after further reflection, the name `safePK` could be `exposedFieldName` to be consistent with other parts of our codebase.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -339,8 +342,8 @@ RuntimeConfigProvider runtimeConfigProvider
                 {
                     string columnName = GetExposedColumnName(entityName, column.ColumnName, sqlMetadataProvider);
 
-                    if (!afterDict.ContainsKey(columnName) ||
-                        afterDict[columnName].Direction != column.Direction)
+                    if (!exposedFieldNameToBackingColumn.ContainsKey(columnName) ||
```
seantleonard, line 345: usage of var name `columnName` could be revised to be `exposedFieldName` to directly align with the dictionary `exposedFieldNameToBackingColumn`. To be consistent with other code in this file, you'd want to stick with 'fieldName' usage for user facing field names, and 'columnName' for any metadata directly mapping to the database/not exposed to the user.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -267,20 +270,20 @@ protected class NextLinkColumn
             RuntimeConfigProvider runtimeConfigProvider
             )
         {
-            List<PaginationColumn>? after = new();
-            IEnumerable<NextLinkColumn>? requestAfter;
+            List<PaginationColumn>? paginationCursorColumnsForQuery = new();
+            IEnumerable<NextLinkColumn>? paginationCursorColumnsFromRequest;
```
seantleonard, line 274:
```suggestion
            IEnumerable<NextLinkColumn>? paginationCursorFieldsFromRequest;
```

seantleonard: see comment about naming: 'fields' refer to values received from user/request and 'columns' is a DAB internal reference to actual database metadata. This will also help with readability: knowing what collection holds data directly provided by request versus values internally resolved/translated by DAB into database metdata.

seantleonard (changes requested):
a few suggestions about naming consistency and what var names we should attribute with data received from request versus DAB internal datastructures mapping to database metadata.

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: that's why we use the column.FieldName instead of the backingColumnName, this is safe.

seantleonard: i see. the use of column / field together confused me

aaronburtle: Ya, the translation between exposed and backing columns/fields is a bit confuding, I thought about calling these NextLinkFields instead of columns, not sure if that's better or worse for readability.

aaronburtle: I think for consistency, now the we are movin everything to the expose = field, backing = column dichotomy, I will rename the class to NextLinkFields, the we can iterate through Fields and have Field.FieldName which should be easier to read

severussundar (approved):
LGTM. Thank you for fixing this and incorporating the suggestions!

seantleonard (approved):
Thank you for addressing feedback! lgtm!

PR #1854, opened 2023-11-01
Upgrade HotChocolate

Aniruddh25:
## Why make this change?

- Pick up latest patch version for HotChocolate 12 for bug fixes.

## What is this change?

- Upgrade the patch version number in the global properties for all HotChocolate.* packages

## How was this tested?
- Verified BCP is launched for local dev and can query GraphQL.
- Existing integration tests -fix some to accommodate how upgraded hotchocolate behaves

### DIFF
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -325,6 +325,11 @@
                       "required": [
                         "methods"
                       ]
+                    },
+                    {
+                      "required": [
+                        "enabled"
+                      ]
                     }
                   ]
                 }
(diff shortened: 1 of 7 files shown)


seantleonard (changes requested):
are the failing tests a result of this change or transient that would resolve on pipeline retry?

Aniruddh25:
> are the failing tests a result of this change or transient that would resolve on pipeline retry?

some of them were due to this change since hotchocolate changed the response codes, I tested with previous version - 0.8.49... 
I have modified the expected response accordingly since this is out of our control. Moving to different response codes shouldnt be a breaking change. Waiting on some Pg/MySql tests as well.

seantleonard (approved):
thanks for keeping dependencies up to date!

abhishekkumams (approved):

aaronburtle (approved):
Looks good!

PR #1856, opened 2023-11-02
Add Validate command

abhishekkumams:
## Why make this change?

- This is a follow up PR for #1845.
- This aims at adding necessary changes to the runtime engine to handle validation when triggered by validate vs start command.

## What is this change?

- Created a new class to validate config file against our dab schema file using `NJsonSchema` nuget package.
- Added a new Method for Validation which will call all the different validation methods for different checks in config including schema validation.
- This method will display all the error at once. for schema errors it will also display the line number containing the error.
- Updated Exception handling in `RuntimeConfigProvider.cs` and `SqlMetadaProvider.cs` class to make sure when validation is triggered by `validate` command, the engine should not exit and instead capture and display all the errors at once.
- It also contains changes related to CLI

## Data Flow:
1. `validate` command called by CLI (if user provided the config, it will be used or it will use the same way to get the config as `start` command).
2. CLI calls the method `IsConfigValid` which returns true if config is valid, else returns false.
3. `IsConfigValid` method calls another method in the engine `TryValidateConfig()` which contains all the different methods we have for validating different parts of the config file.
4. we have a `JsonConfigSchemaValidator` class containing containing methods to do schema related validation of the config.
5. it then does metada validation as well where it basically test the entities against the database, here instead of throwing the exception, we pass it to a method called `HandleOrRecordException()`. it stores the exception when called from `validate` command (i.e `isValidateOnly=true`).
6. These handling has been placed in such a way that when some exceptions such as connection-string issue or invalid entity issue are thrown, other checks are not performed.
7. for other exceptions, it continues the validation through every entity and stores it in a list. And displays all the exception message at once after validation is complete.

## Order of Validation
1. **Schema validation:** fetches the schema file from the `$schema` property, if not available uses the schema present in the package.
2. **Config properties validation:** validates datasource, runtime-section, rest and graphQL endpoints
3. **Config permission validation:** validates semantic correctness of the permission defined for each entity.
4. **Database connection validation:** validates connection to database.
5. **Entities Metadata validation:** validates entites configuration against database. this check won't run if database connection fails.

## How was this tested?

- [X] Integration Tests
- [X] Unit Tests

## NOTE:
`NJsonSchema` currently has an open issue for not supporting schema validation inside if-else condition:
https://github.com/RicoSuter/NJsonSchema/issues/1240


seantleonard (changes requested):
see comments

src/Core/Configurations/JsonConfigSchemaVaidator.cs
```diff
@@ -0,0 +1,147 @@
...
+    /// </summary> 
+    /// <param name="jsonConfig">The JSON config file to validate.</param> 
+    public static async Task ValidateJsonConfig(string jsonConfig)
+    {
+        // deserialize jsonData to runtimeConfig
+        if (!RuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig? config))
+        {
+            Console.WriteLine("The config is invalid.");
+            return;
+        }
```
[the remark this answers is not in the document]
abhishekkumams: to get the $schema property

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -645,34 +744,34 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
                         message: $"Cannot define relationship for entity: {entity}",
                         statusCode: HttpStatusCode.ServiceUnavailable,
-                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
             }
 
             string databaseName = runtimeConfig.GetDataSourceNameFromEntityName(entityName);
             ISqlMetadataProvider sqlMetadataProvider = sqlMetadataProviderFactory.GetMetadataProvider(databaseName);
 
-            foreach ((string relationshipName, EntityRelationship relationship) in entity.Relationships!)
```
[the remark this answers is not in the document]
abhishekkumams: yes

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -102,16 +117,59 @@ public void ValidateConfig()
...
+
+    public async Task ValidateEntitiesMetadata()
+    {
+        LoggerFactory loggerFactory = new();
+        QueryManagerFactory queryManagerFactory = new(
+            runtimeConfigProvider: _runtimeConfigProvider,
+            logger: loggerFactory.CreateLogger<IQueryExecutor>(),
+            contextAccessor: null!);
+        
+        MetadataProviderFactory metadataProviderFactory = new(
```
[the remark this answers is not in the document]
abhishekkumams: validation when triggered by cli, won't have these setup in StartUp, because it's role is simply to validate the config file. It won't setup the application

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -102,16 +117,59 @@ public void ValidateConfig()
...
         ValidateDatabaseType(runtimeConfig, fileSystem, logger);
     }
 
+    public async void ValidateConfig(string configFilePath)
+    {
+        _isValidateOnly = true;
+        await ValidateConfigSchema(configFilePath);
+        ValidateConfigProperties();
+        ValidatePermissionsInConfig(_runtimeConfigProvider.GetConfig());
+        await ValidateEntitiesMetadata();
```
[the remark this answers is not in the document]
abhishekkumams: the idea was to give as many errors as possible in one go. but we can do this as well, specifying users that `schema validation failed. Correct schema and run validation again for running other validations`.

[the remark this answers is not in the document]
seantleonard: You can still do that. Just logically separate code paths but both can be executed within Dabcli.validate

src/Core/Configurations/JsonConfigSchemaVaidator.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// <summary> 
+    /// Retrieves the JSON schema from the assembly package. 
+    /// </summary>
+    /// <returns>The contents of the JSON schema file.</returns> 
+    private static string GetSchemaFromAssemblyPackage()
+    {
+        string assemblyPath = Assembly.GetExecutingAssembly().Location;
+        string directoryPath = _fileSystem.Path.GetDirectoryName(assemblyPath)!;
+        string jsonPath = _fileSystem.Path.Combine(directoryPath, "dab.draft.schema.json");
```
Aniruddh25, line 142: Use a const string for the schema name

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -102,16 +117,59 @@ public void ValidateConfig()
...
+                );
             }
         }
 
         ValidateDatabaseType(runtimeConfig, fileSystem, logger);
     }
 
+    public async void ValidateConfig(string configFilePath)
+    {
+        _isValidateOnly = true;
```
Aniruddh25, line 134: Keeping a static member `_isValidateOnly` and modifying it via Test, here doesnt seem intuitive. You can create separate object for RuntimeConfigValidator in CLI with this property set to true when its the validate command. For all other cases, you can keep the default value of this property to be false. 

Static members should be defined to share state between different objects - as a class wide property not for convenience of use.

src/Core/Configurations/JsonConfigSchemaVaidator.cs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /// <summary> 
+    /// Validates the given json config. 
+    /// </summary> 
+    /// <param name="jsonConfig">The JSON config file to validate.</param> 
+    public static async Task ValidateJsonConfig(string jsonConfig)
+    {
+        // deserialize jsonData to runtimeConfig
+        if (!RuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig? config))
```
Aniruddh25, line 69: Cant we get the config from the RuntimeConfigProvider which could have it already parsedf? 
If CLI wont have it, can we do the Parsing only for the CLI portion when you implement `validate` but when its being called from Service, reuse the already parsed config object?

src/Core/Configurations/JsonConfigSchemaVaidator.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        string? jsonSchema = await GetJsonSchema(config);
+
+        if (string.IsNullOrWhiteSpace(jsonSchema))
+        {
+            _logger!.LogWarning("The schema file is invalid. Unable to verify provided config against schema.");
+            return;
+        }
+
+        (bool isValid, ICollection<ValidationError>? validationErrors) = await ValidateJsonConfigWithSchemaAsync(jsonSchema, jsonConfig);
```
Aniruddh25, line 84: Is passing a null jsonSchema to this function appropriate?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -977,21 +1041,182 @@ public void TestRuntimeEnvironmentVariable()
...
             RuntimeConfigProvider configProvider = TestHelper.GetRuntimeConfigProvider(configPath);
 
             Mock<ILogger<RuntimeConfigValidator>> configValidatorLogger = new();
-            IConfigValidator configValidator =
-                new RuntimeConfigValidator(
+            RuntimeConfigValidator configValidator =
+                new (
                     configProvider,
                     new MockFileSystem(),
                     configValidatorLogger.Object);
```
Aniruddh25, line 1059: you could provide _isValidateOnly property as you construct this object RuntimeConfigValidator here.

src/Core/Configurations/RuntimeConfigValidator.cs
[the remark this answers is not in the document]
Aniruddh25: In that case, we should create these objects only in the CLI validate implementation. Not create duplicates for this in the Service. Do we perform the ValidateRelationshipsInConfig in the service too?

If these are new objects we are adding as dependencies to the RuntimeConfigValidator, we should do that. And in the cli, create a RuntimeConfigValidator and supply all these objects explicitly.

Aniruddh25 (changes requested):
Still need to review ConfigurationTests, but posting comments so far since the `static` variables implementation needs to be modified, correct objects created separately in CLI/Service.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -102,16 +117,59 @@ public void ValidateConfig()
...
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization)
+                );
             }
         }
 
         ValidateDatabaseType(runtimeConfig, fileSystem, logger);
     }
 
+    public async void ValidateConfig(string configFilePath)
```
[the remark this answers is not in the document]
ayush3797: Can you add comments when this function will be called?

src/Core/Configurations/JsonConfigSchemaVaidator.cs
```diff
@@ -0,0 +1,147 @@
...
+        
+        if (!isValid && validationErrors != null)
```
ayush3797, line 86: Curious why only a check on `isValid` is not sufficient here

ayush3797:
@abhishekkumams , is the functionality which `dab validate` will provide not a superset of what runtime config validator does? If it is a superset, can we have a class which contains :
1. An object of `RuntimeConfigValidator` type and will do all the existing validations? 
2. Contains references to all other dependencies needed injected via DI?

src/Core/Configurations/RuntimeConfigValidator.cs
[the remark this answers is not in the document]
ayush3797: I think we should keep the existing `RuntimeConfigValidator` untouched and all the additional work that dab validate needs to perform goes into another class which will have these extra added dependencies.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -30,6 +32,9 @@ public class RuntimeConfigValidator : IConfigValidator
     // Only characters from a-z,A-Z,0-9,.,_ are allowed to be present within the claimType.
     private static readonly string _invalidClaimChars = @"[^a-zA-Z0-9_\.]+";
 
+    private static bool _isValidateOnly;
+    public static HashSet<Exception> SqlMetadataProviderExceptions {get; } = new();
```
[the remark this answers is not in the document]
abhishekkumams: we will just log the exception message and not the name of the class where it happened. Users should not know the internal implementation. Aim is to provide logging message which will let users know what they need to fix in the config.

src/Service.Tests/TestHelper.cs
```diff
@@ -26,6 +26,8 @@ public static void UnsetAllDABEnvironmentVariables()
             Environment.SetEnvironmentVariable(FileSystemRuntimeConfigLoader.RUNTIME_ENVIRONMENT_VAR_NAME, null);
             Environment.SetEnvironmentVariable(FileSystemRuntimeConfigLoader.ASP_NET_CORE_ENVIRONMENT_VAR_NAME, null);
             Environment.SetEnvironmentVariable(FileSystemRuntimeConfigLoader.RUNTIME_ENV_CONNECTION_STRING, null);
+            RuntimeConfigValidator.TurnOffValidationOnlyMode();
+            RuntimeConfigValidator.SqlMetadataProviderExceptions.Clear();
```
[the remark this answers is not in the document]
abhishekkumams: not required anymore, after moving out of static methods

src/Core/Configurations/JsonConfigSchemaVaidator.cs
abhishekkumams: updated

src/Core/Configurations/JsonConfigSchemaVaidator.cs
abhishekkumams: for cases where it failed to validate the schema due to exception caused while validating or parsing the schema

src/Core/Configurations/RuntimeConfigValidator.cs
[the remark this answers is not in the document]
abhishekkumams: I'm trying to have a single method in RuntimeConfigValidator class which is called by CLI for validate

seantleonard:
@abhishekkumams Does `dotnet format` in local work for you to identify and mitigate formatting issues? that way you don't have to wait for the pipeline to alert you about formatting issues.

abhishekkumams:
> @abhishekkumams Does `dotnet format` in local work for you to identify and mitigate formatting issues? that way you don't have to wait for the pipeline to alert you about formatting issues.

it does give some errors sometimes when running locally, it's on my radar to fix it.

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    [TestCleanup]
+    public void TestCleanup()
+    {
+        _fileSystem = null;
+        _runtimeConfigLoader = null;
+    }
+
+    /// <summary>
+    /// this method  validates that the IsConfigValid method returns false when the config is invalid.
```
seantleonard, line 36:
```suggestion
    /// This method validates that the IsConfigValid method returns false when the config is invalid.
```

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -0,0 +1,62 @@
...
+    [TestMethod]
+    public void TestConfigWithInvalidConfigProperties()
+    {
+        ((MockFileSystem)_fileSystem!).AddFile(TEST_RUNTIME_CONFIG_FILE, CONFIG_WITH_DISABLED_GLOBAL_REST_GRAPHQL);
+
+        ValidateOptions validateOptions = new(TEST_RUNTIME_CONFIG_FILE);
+
+        bool isConfigValid = ConfigGenerator.IsConfigValid(validateOptions, _runtimeConfigLoader!, _fileSystem!);
+        Assert.IsFalse(isConfigValid);
+    }
```
seantleonard, line 61: Comment doesn't align with test.

src/Cli/Utils.cs
```diff
@@ -840,5 +840,15 @@ public static bool IsEntityProvided(string? entity, ILogger cliLogger, string co
 
             return true;
         }
+
+        /// <summary>
+        /// Returns ILoggerFactory with CLI custom logger provider.
+        /// </summary>
+        public static ILoggerFactory GetLoggerFactoryForCli()
```
seantleonard, line 847: should callers be asking for a logger factory or a logger? ILoggerFactory has a method called `CreateLogger(string categoryName)` which returns a logger.

src/Cli/ConfigGenerator.cs
```diff
@@ -1091,6 +1082,64 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
+            }
+            else
+            {
+                _logger.LogInformation("Loaded config file: {runtimeConfigFile}", runtimeConfigFile);
+            }
+
+            RuntimeConfigProvider runtimeConfigProvider = new(loader);
+
+            ILoggerFactory loggerFactory = Utils.GetLoggerFactoryForCli();
+            ILogger<RuntimeConfigValidator> runtimeConfigValidatorLogger = loggerFactory.CreateLogger<RuntimeConfigValidator>();
```
seantleonard, line 1110: You alrady create a logger factory in `Program.cs` in the CLI project. Why another factory? now loggers are being created from different(multiple) factories?

src/Cli/ConfigGenerator.cs
```diff
@@ -1091,6 +1082,64 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
+
+            return runtimeConfigValidator.TryValidateConfig(runtimeConfigFile, deserializedRuntimeConfig, loggerFactory).Result;
+        }
+
+        /// <summary>
+        /// Tries to fetch the config file based on the precedence.
+        /// if config provided by the user, it will be the final config used, else will check based on the environment variable.
+        /// Returns true if the config file is found, else false.
+        /// </summary>
+        public static bool TryGetFinalConfigForRuntimeEngine(
```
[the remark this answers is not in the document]
abhishekkumams: final config is basically the final config selected based on precedence or merging, etc that will be supplied to engine for starting up.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -77,18 +76,43 @@ public abstract class SqlMetadataProvider<ConnectionT, DataAdapterT, CommandT> :
...
+            {
+                SqlMetadataExceptions.Add(e);
+            }
+            else
+            {
+                throw e;
+            }
+        }
+
+        public void LogSqlMetadataExceptions()
```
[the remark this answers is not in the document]
abhishekkumams: removed, not required

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1049,10 +1052,11 @@ public void TestConfigPropertiesAreValid()
 
             Mock<ILogger<RuntimeConfigValidator>> configValidatorLogger = new();
             RuntimeConfigValidator configValidator =
-                new (
+                new(
                     configProvider,
                     new MockFileSystem(),
-                    configValidatorLogger.Object);
+                    configValidatorLogger.Object,
+                    isValidateOnly: true);
```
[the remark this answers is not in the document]
abhishekkumams: correct, thanks for pointing out. updated.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -944,13 +1096,18 @@ public static bool IsValidPermissionAction(EntityActionOperation action, Entity
...
                     statusCode: HttpStatusCode.ServiceUnavailable,
-                    subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
             }
 
             return action is EntityActionOperation.All || EntityAction.ValidPermissionOperations.Contains(action);
         }
     }
+
+    public void ClearValidationExceptionList()
```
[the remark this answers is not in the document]
abhishekkumams: we actually don't require this method. Since, we are not reusing it, no need for extra operation.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -926,16 +1078,16 @@ private static DataApiBuilderException GetInvalidActionException(string entityNa
     /// <param name="entity">Used to identify entity's representative object type.</param>
     /// <param name="entityName">Used to supplement error messages.</param>
     /// <returns>Boolean value indicating whether the action is valid or not.</returns>
-    public static bool IsValidPermissionAction(EntityActionOperation action, Entity entity, string entityName)
+    public bool IsValidPermissionAction(EntityActionOperation action, Entity entity, string entityName)
```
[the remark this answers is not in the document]
abhishekkumams: right, made it private

src/Service.Tests/TestHelper.cs
```diff
@@ -155,6 +162,57 @@ public static RuntimeConfig AddMissingEntitiesToConfig(RuntimeConfig config, str
             ""entities"": {}" +
           "}";
 
+        /// <summary>
+        /// <summary>
+        /// A minimal valid config json with invalid schema due to custom property `description`.
+        /// This config string is used in unit tests for schema validation tests.
+        /// </summary>
+        public const string CONFIG_WITH_INVALID_SCHEMA =
```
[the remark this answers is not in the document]
abhishekkumams: right, it was redundant.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -973,24 +1040,253 @@ public void TestRuntimeEnvironmentVariable()
...
 
-            configValidator.ValidateConfig();
+            await configValidator.ValidateEntitiesMetadata(configProvider.GetConfig(), mockLoggerFactory);
+            Assert.IsTrue(configValidator.ConfigValidationExceptions.IsNullOrEmpty());
             TestHelper.UnsetAllDABEnvironmentVariables();
```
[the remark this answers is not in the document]
abhishekkumams: Unsetting explicitly is not required it already has a TestCleanUp, which does unsetDAB environment variables after each tests. updated the TestClass

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -973,24 +1040,253 @@ public void TestRuntimeEnvironmentVariable()
...
+            Assert.IsTrue(result.ErrorMessage.Contains("NoAdditionalPropertiesAllowed: #/description"));
+
+            File.WriteAllText(CUSTOM_CONFIG, combinedJson);
+            string[] args = new[]
+            {
+                $"--ConfigFileName={CUSTOM_CONFIG}"
+            };
+
+            // Non-Hosted Scenario
+            using (TestServer server = new(Program.CreateWebHostBuilder(args)))
```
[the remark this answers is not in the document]
abhishekkumams: No, program startup will not fail due to extra properties. This test is specifically for that, it verifies that the engine startup is not affected due to some extra properties.

src/Core/Services/MetadataProviders/IMetadataProviderFactory.cs
```diff
@@ -19,6 +19,8 @@ public interface IMetadataProviderFactory
         /// </summary>
         public IEnumerable<ISqlMetadataProvider> ListMetadataProviders();
 
+        public List<Exception> GetAllMetadataExceptions();
```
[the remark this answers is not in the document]
abhishekkumams: Could be, but i need a method that will give me all the exceptions occured in different metadataProviders in a single place for logging.

src/Cli/ConfigGenerator.cs
abhishekkumams: created a static variable to not create multiple new instances of loggerFactory

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -262,6 +254,22 @@ public async Task InitializeAsync()
         {
             System.Diagnostics.Stopwatch timer = System.Diagnostics.Stopwatch.StartNew();
             GenerateDatabaseObjectForEntities();
+            if (_isValidateOnly)
+            {
+                // Currently Validate mode only support single datasource,
```
seantleonard, line 259: this needs to be documented in public facing docs for Dab validate.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -321,17 +329,32 @@ private void LogPrimaryKeys()
...
-            DataTable procedureMetadata = await conn.GetSchemaAsync(collectionName: "Procedures", restrictionValues: procedureRestrictions);
+                procedureMetadata = await conn.GetSchemaAsync(collectionName: "Procedures", restrictionValues: procedureRestrictions);
+            }
+            catch (Exception ex)
+            {
+                string message = $"Cannot obtain Schema for entity {entityName} " +
+                            $"with underlying database object source: {schemaName}.{storedProcedureSourceName} " +
+                            $"due to: {ex.Message}";
+
+                throw new DataApiBuilderException(
```
seantleonard, line 353: include `ex` as innerException param for `DataApiBuilderException`

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -925,19 +994,26 @@ private void GenerateExposedToBackingColumnMapsForEntities()
...
-                // InCase of StoredProcedures, result set definitions becomes the column definition.
-                Dictionary<string, string>? mapping = GetMappingForEntity(entityName);
-                EntityBackingColumnsToExposedNames[entityName] = mapping is not null ? mapping : new();
-                EntityExposedNamesToBackingColumnNames[entityName] = EntityBackingColumnsToExposedNames[entityName].ToDictionary(x => x.Value, x => x.Key);
-                SourceDefinition sourceDefinition = GetSourceDefinition(entityName);
-                foreach (string columnName in sourceDefinition.Columns.Keys)
+                try
                 {
-                    if (!EntityExposedNamesToBackingColumnNames[entityName].ContainsKey(columnName) && !EntityBackingColumnsToExposedNames[entityName].ContainsKey(columnName))
+                    // InCase of StoredProcedures, result set definitions becomes the column definition.
```
seantleonard, line 999:
```suggestion
                    // For StoredProcedures, result set definitions become the column definition.
```

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -100,21 +109,122 @@ public void ValidateConfig()
...
+        string configFilePath,
+        RuntimeConfig runtimeConfig,
+        ILoggerFactory loggerFactory,
+        bool isValidateOnly = false)
+    {
+        JsonSchemaValidationResult validationResult = await ValidateConfigSchema(runtimeConfig, configFilePath, loggerFactory);
+        ValidateConfigProperties();
+        ValidatePermissionsInConfig(runtimeConfig);
+
+        // if the ConfigValidationExceptions list doesn't contain a DataApiBuilderException with message "I am Khan",
```
seantleonard, line 142: what is "I am khan" ?

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Cli.Tests;
+/// <summary>
+/// Test for config file initialization.
+/// </summary>
+[TestClass]
+public class ValidateConfigTests
+    : VerifyBase
+{
+    private IFileSystem? _fileSystem;
```
aaronburtle, line 12: how come we need this `IFileSystem` as its own field? can't this just live in the loader?

src/Cli/ConfigGenerator.cs
```diff
@@ -1091,6 +1082,63 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
             return Azure.DataApiBuilder.Service.Program.StartEngine(args.ToArray());
         }
 
+        /// <summary>
+        /// Runs all the validations on the config file and returns true if the config is valid.
+        /// </summary>
+        public static bool IsConfigValid(ValidateOptions options, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
```
aaronburtle, line 1088: similar comment as above, the `FileSystemRuntimConfigLoader` has a `IFileSystem` is there a reason we aren't just grabbing that filesystem? convenience?

src/Cli.Tests/ValidateConfigTests.cs
abhishekkumams: we are using that fileSystem to create the loader. following the same convention as other tests.

src/Cli/ConfigGenerator.cs
abhishekkumams: yes, i had this question as well, after checking when the fileSystem was introduced, it was like this and it was passing both loader and fileSystem separately. I believe the fileSystem is intended to be used only within the class and not modified externally, so it was better to keep it private and not create a getter or something.

seantleonard (approved):

aaronburtle (approved):
Looks good!

PR #1863, opened 2023-11-06
Removed hash from application name (fix #1759)

### LINKED ISSUE #1759: Reduce DAB Application Name Length
As per documentation (https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection.connectionstring?view=sqlclient-dotnet-standard-5.1), application name is limited to 128 char, and right now DAB is injecting in the connection string a very long string, like

```text
dab_oss_0.9.5-rc+73a02dcd3b85a73ac36a5675432a45b7a3cc4a7d
```

can it be reduced to just the version number without the commit hash, to avoid to use too much space? for example:

```text
dab_oss_0.9.5-rc
```

yorek:
## Why make this change?

Proposed fix for #1759

## What is this change?

Remove the hashed value after the "+" sign when injecting DAB version in the MSSQL connection string 

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests
- [x] Manual Test

## Sample Request(s)

N/A


src/Product/ProductInfo.cs
```diff
@@ -8,15 +8,15 @@ namespace Azure.DataApiBuilder.Product;
 
 public static class ProductInfo
 {
-    public const string DEFAULT_VERSION = "1.0.0";
+    public const string DEFAULT_VERSION = "0.0.0";
```
abhishekkumams, line 11: I believe default was given 1.0.0 for tests, because it is what gets generated automatically by assembly when nuget is not installed, or installed without specifying a version.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -195,7 +195,9 @@ internal static string GetConnectionStringWithApplicationName(string connectionS
         }
 
         // Get the application name using ProductInfo.GetDataApiBuilderUserAgent().
-        string applicationName = ProductInfo.GetDataApiBuilderUserAgent();
+        string dabVersion = ProductInfo.GetDataApiBuilderUserAgent();
+        int hashStartPosition = dabVersion.LastIndexOf('+');
```
abhishekkumams, line 199: we can add a method in ProductInfo class to do the same. this might be needed somewhere else.

src/Product/ProductInfo.cs
yorek: I see. The point of using something different than 1.0.0 is to make it possible to understand when 1.0.0 really means 1.0.0 vs when it means there was an issue getting the version number

src/Config/RuntimeConfigLoader.cs
yorek: Make sense, if keeping the hash (is the build hash or the signature hash? not sure what that hashed value represents) is not useful for other purposes, then it make sense to move this into the `GetDataApiBuilderUserAgent` method.

yorek: I wonder if we want to keep it to show it in the `dab --version` command. Is the build hash useful to you for any reasons?

src/Product/ProductInfo.cs
yorek: I updated the unit tests and, at least locally, it seems everything is fine with "0.0.0"

src/Config/RuntimeConfigLoader.cs
Aniruddh25: The commit hash is useful to easily identify the last change that went into that version - although we can derive it from the version branch commit history - but thats a 2 step process. 

Currently, the hash is appended as a side-effect of maintaining a healthy nuget package status with regards to source link, deterministic dlls. i,e, in the nuget explorer we see:

image
 
This health status will be red if we attempt to remove the hash from `dab --version`

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -502,11 +502,7 @@ public void CleanupAfterEachTest()
 
             string actualUpdatedConnectionString = updatedRuntimeConfig.DataSource.ConnectionString;
 
-            // SourceLink appends the commit ID to the assembly metadata. The application name is extracted from the
-            // assembly metadata. So, the connection string after appending the application name will be of the form
-            // dab_oss_1.0.0+<commit-id> or dab_hosted_1.0.0+<commit-id> depending on oss or hosted scenario respectively.
-            // So, the updated connection string is validated to check if it starts with dab_oss_1.0.0 or dab_hosted_1.0.0.
-            Assert.IsTrue(actualUpdatedConnectionString.StartsWith(expectedUpdatedConnectionString));
+            Assert.IsTrue(actualUpdatedConnectionString == expectedUpdatedConnectionString);
```
Aniruddh25, line 505:
```suggestion
            Assert.AreEqual(actualUpdatedConnectionString, expectedUpdatedConnectionString);
```

Using Equal is better for troubleshooting because in case of failures, it will print out both the string in the logs so we would know immediately why the test could have failed.

Aniruddh25 (changes requested):
Looks good but seems like one other Unit Test is still failing

src/Config/RuntimeConfigLoader.cs
yorek: Make sense, so let's keep it, and strip the hash only from the application name injected in the connection string

yorek:
Yep, I'll fix the test ASAP

src/Config/RuntimeConfigLoader.cs
seantleonard: To prevent a regression in meeting all the Nuget health check properties, can you add a test checking that `dab --version` includes the commit hash. Since the commit hash changes and may not be straightforward to identify locally, just checking that a `+` character and additional characters follow the version number.

seantleonard: Relevant test to update `TestVersionInfoAndConfigIsCorrectlyDisplayedWithDifferentCommand` in Cli\Tests\EndToEndTests.cs

seantleonard (changes requested):
One test to fix in End2End tests to get pipeline passing. Also a test in that file can be updated to ensure commit hash is still included when expected to ensure nuget health check succeeds and doesn't regress.

src/Config/RuntimeConfigLoader.cs
yorek: added the test

yorek:
@Aniruddh25 @seantleonard build is failing on "Check formatting", complaining that .NET 2 is not supported....

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -668,7 +668,12 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
         Assert.IsNotNull(output);
 
         // Version Info logged by dab irrespective of commands being parsed correctly.
-        StringAssert.Contains(output, $"{Program.PRODUCT_NAME} {ProductInfo.GetProductVersion()}", StringComparison.Ordinal);
+        StringAssert.Contains(ProductInfo.GetProductVersion(), $"{Program.PRODUCT_NAME} {ProductInfo.GetProductVersion()}", StringComparison.Ordinal);
```
Aniruddh25, line 671:
```suggestion
        StringAssert.Contains(output, $"{Program.PRODUCT_NAME} {ProductInfo.GetProductVersion()}", StringComparison.Ordinal);
```

Keep the test string to check the contents of to be the output from the dab command itself, otherwise this assert is not testing anything - infact it would fail coz the first string will then just be a subset of the second..

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -668,7 +668,12 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
+
+        // Check that build hash is returned as part of  version number
+        string[] versionParts = output.Split('+');
+        Assert.AreEqual(2, versionParts.Length, "Build hash not returned as part of version number.");
+        Assert.AreEqual(41, versionParts[1].Length, "Build hash is not of expected length.");
```
Aniruddh25, line 676: The `output` here could be different based on the command used in the test case. If you look at the various DataRows specified as a function header, only the first entry about version will have the second part to be exactly equal to the hash length.

The same wouldn't hold true for other cases. For example, for the test case: `edit --new-option` the output seen would be: 

> Microsoft.DataApiBuilder 0.9.6-rc+529471ba3b50e617439c57548891fdff7ec1c3fa
> c Microsoft Corporation. All rights reserved.
> 
> ERROR(S):
>   Verb 'edit' is not recognized.
> 
>   --help       Display this help screen.
> 
>   --version    Display version information.

So, we should not assert on the hash length - it will always be a subset of the whole output. Same for the versionParts.Length, there could be more than one `+` in the entire output

Aniruddh25: if you just want to verify ProductVersion contains the hash, we can simply do that with unit test asserts on the `ProductInfo.GetProductVersion()` - that it contains the +, and hash is of right length. This doesnt need to be added in this test: `TestVersionInfoAndConfigIsCorrectlyDisplayedWithDifferentCommand` - you can define a separate unit test method altogether as well

Aniruddh25:
> @Aniruddh25 @seantleonard build is failing on "Check formatting", complaining that .NET 2 is not supported....

Did you get a chance to setup this git pre-commit hook? 
https://github.com/Azure/data-api-builder/blob/main/CONTRIBUTING.md#enforcing-code-style-with-git-hooks

Aniruddh25:
> @Aniruddh25 @seantleonard build is failing on "Check formatting", complaining that .NET 2 is not supported....

The actual error from here https://msdata.visualstudio.com/CosmosDB/_build/results?buildId=110183176&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=2b20c7d0-7587-5b0f-aadc-b456e6f0b807 is : 

> /home/vsts/work/1/s/src/Cli.Tests/EndToEndTests.cs(6,1): warning IDE0005: Using directive is unnecessary. [/home/vsts/work/1/s/src/Cli.Tests/Cli.Tests.csproj]

So, you need to remove the `using` directive from that EndToEndTests file on line 6

src/Cli.Tests/EndToEndTests.cs
yorek: Good point :)

src/Cli.Tests/EndToEndTests.cs
yorek: Make sense. I created a new `TestVersionHasBuildHash` test

yorek:
> One test to fix in End2End tests to get pipeline passing. Also a test in that file can be updated to ensure commit hash is still included when expected to ensure nuget health check succeeds and doesn't regress.

Fixed

yorek:
All green! :)

src/Config/RuntimeConfigLoader.cs
```diff
@@ -194,8 +194,8 @@ internal static string GetConnectionStringWithApplicationName(string connectionS
             return connectionString;
         }
 
-        // Get the application name using ProductInfo.GetDataApiBuilderUserAgent().
-        string applicationName = ProductInfo.GetDataApiBuilderUserAgent();
+        // Get the application name using ProductInfo.GetDataApiBuilderApplicationName().
```
seantleonard, line 197:
```suggestion
```

src/Product/ProductInfo.cs
```diff
@@ -36,5 +36,12 @@ public static string GetDataApiBuilderUserAgent()
     {
         return Environment.GetEnvironmentVariable(DAB_APP_NAME_ENV) ?? DEFAULT_APP_NAME;
     }
+
+    public static string GetDataApiBuilderApplicationName()
```
seantleonard, line 40: nit: this does more than just return the app name. i.e. returns: `dab_oss_0.10.5` 

a comment to not how this function differs from `GetProductVersion()` would be helpful.

seantleonard (approved):
last nit about function name != function behavior.

src/Config/RuntimeConfigLoader.cs
yorek: Good catch! :)

src/Product/ProductInfo.cs
yorek: Make sense, added

seantleonard (approved):

Aniruddh25 (approved):
Thank you for this change! LGTM

PR #1864, opened 2023-11-06
Addition of new database type DWSQL.

rohkhann:
As part of ongoing effort to add support for Data warehouse to data-api-builder, this small pr adds the database type DWSQL to database type enum.


aaronburtle (approved):
LGTM! Can you add a comment somewhere or to our docs in some place that says DWSQL is DataWarehouse SQL, just because the other types are self explanatory, but this one is not.

seantleonard (approved):

PR #1865, opened 2023-11-07
Add Cache properties to runtime config file with JSON Serialize/Deserialize

seantleonard:
## Why make this change?

- first work attributed to #1618 #1617 

### Global Cache Config property `cache`
- Adds global config property `cache` with value type of boolean in the runtime section:
```json
{
  "runtime": {
    "cache": false
  }
}
```
Behavior:
  - Nullable and defaults to false.
  - When null, will not be written to the CLI created runtime config JSON

### Entity Cache Config property/object `cache`
- Adds entity specific config property `cache` in entity section:
```json
{
  "entityName": {
    "cache": {
      "enabled": true,
      "ttl-seconds": 5
    }
  }
}
```
Behavior:
- Nullable and defaults to `"enabled": false, "ttl-seconds": 5`
- When an empty object -> caching disabled
- When null -> caching disabled
- ttl-seconds defaults to 5 regardless but due to private field state, tracks whether the value was user provided or set with the default value due to not being explicitly set in runtime config. This enables CLI runtime config creation to exclude the `ttl-seconds` property and value if not explicitly set by the user.

## How was this tested?
- [X] manually tested CLI handling of writing `ttl-seconds` to file. but CLI specific work will come in separate pr.
- [X] Unit Tests -> System under test is `RuntimeConfigLoader.TryParseConfig(string json, ...)` such that I test the behavior of any json parsing that goes on. JSON parsing mechanisms are implementation details and aren't tested directly because the implementation can change. I explicitly test for the desired behavior. I test Behavior such as providing null and empty json objects in order to validate how CLI merge config will work. and to avoid regressions that have come up in the past due to improper/lackof null handling.

## Sample Request(s)

Caching is not yet enabled. Behavior can be observed in unit test class `CachingConfigDeserializationTests` or you can create a config using the CLI:

1. Create a folder called `dabconfigs` at root of C:\ drive to work with these sample commands.
2. run DAB cli to create a config file and one entity called Book.
```sh
Microsoft.DataApiBuilder init --database-type mssql -c "C:\dabconfigs\dab-config.json"              
   
Microsoft.DataApiBuilder add Book -c "C:\dabconfigs\dab-config.json" --source books --permissions "anonymous:create,read,update,delete" --graphql "book:books"
```

3. Modify config file directly to add cache config to entity.
```json
    "Book": {
      "source": {
        "object": "books",
        "type": "table"
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "book",
          "plural": "books"
        }
      },
      "rest": {
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            {
              "action": "*"
            }
          ]
        }
      ],
      "cache": {
        "enabled": true
      }
    }
```

4. add another entity, and validate that `ttl-seconds` is not added to the `Book` entity config because a non-explicitly set ttl should not result in a property/value being written to the config from CLI. Under the hood, the default 5 second value is used, which is validated in unit tests.
```sh
Microsoft.DataApiBuilder add Dog -c "C:\dabconfigs\dab-config.json" --source books --permissions "anonymous:create,read,update,delete" --graphql "dog:dogs"
```


src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
```diff
@@ -0,0 +1,242 @@
...
+    [DataRow(@",""cache"": { ""enabled"": true, ""ttl-seconds"": 2147483648 }", DisplayName = "EntityCacheOptions.TtlSeconds set to Int.MaxValue+1 is invalid for parser.")]
+    [DataRow(@",""cache"": { ""enabled"": true, ""ttl-seconds"": -2147483649 }", DisplayName = "EntityCacheOptions.TtlSeconds set to Int.MinValue-1 is invalid for parser.")]
+    [DataTestMethod]
+    public void EntityCacheOptionsDeserialization_InvalidJson(string entityCacheConfig)
+    {
+        // Arrange
+        string fullConfig = GetRawConfigJson(globalCacheConfig: string.Empty, entityCacheConfig: entityCacheConfig);
+
+        // Act
+        bool parsingSuccessful = RuntimeConfigLoader.TryParseConfig(
```
Aniruddh25, line 91:
```suggestion
        bool isParsingSuccessful = RuntimeConfigLoader.TryParseConfig(
```

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
```diff
@@ -0,0 +1,242 @@
...
+            out _,
+            logger: null,
+            connectionString: null,
+            replaceEnvVar: false,
+            dataSourceName: string.Empty,
+            datasourceNameToConnectionString: null,
+            replacementFailureMode: EnvironmentVariableReplacementFailureMode.Throw);
+
+        // Assert
+        Assert.IsFalse(parsingSuccessful, message: "Expected JSON parsing to fail.");
```
Aniruddh25, line 102:
```suggestion
        Assert.IsFalse(isParsingSuccessful, message: "Expected JSON parsing to fail.");
```

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
```diff
@@ -0,0 +1,242 @@
...
+    string entityCacheConfig,
+    bool expectedEnabled,
+    int expectedTTL,
+        bool expectedUserDefinedTtl)
+    {
+        // Arrange
+        string fullConfig = GetRawConfigJson(globalCacheConfig: string.Empty, entityCacheConfig: entityCacheConfig);
+
+        // Act
+        RuntimeConfigLoader.TryParseConfig(
```
Aniruddh25, line 51: In addition to deserialization, can we have verification that after serialization, we get back the same `fullConfig`?

src/Config/Converters/EntityCacheOptionsConverterFactory.cs
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.ObjectModel;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+
+internal class EntityCacheOptionsConverterFactory : JsonConverterFactory
```
abhishekkumams, line 10: maybe a comment on what is this class used for, would be useful.

src/Config/Converters/EntityCacheOptionsConverterFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+    public override JsonConverter? CreateConverter(Type typeToConvert, JsonSerializerOptions options)
+    {
+        return new EntityCacheOptionsConverter();
+    }
+
+    private class EntityCacheOptionsConverter : JsonConverter<EntityCacheOptions>
+    {
+        /// <summary>
+        /// Defines how DAB reads an entity's cache options and defines which values are
+        /// written to the runtime configuration.
```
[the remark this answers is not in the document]
seantleonard: -> Defines how DAB reads an entity's cache options and defines which values are used to instantiate EntityCacheOptions.

src/Config/ObjectModel/EntityCacheOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+/// <summary>
+/// Entity specific in-memory cache configuration.
+/// Properties are nullable to support DAB CLI merge config
+/// expected behavior.
+/// </summary>
+public record EntityCacheOptions
+{
+    public const int DEFAULT_TTL_SECONDS = 60;
+    /// <summary>
```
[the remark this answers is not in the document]
seantleonard: that's equal. i'll lower to 45 seconds.

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
```diff
@@ -0,0 +1,242 @@
...
+            Assert.AreEqual(expectedUserDefinedTtl, resolvedEntityCacheOptions.UserProvidedTtlOptions, message: "UserProvidedTtlOptions expected to be: " + expectedUserDefinedTtl);
+        }
+    }
+
+    /// <summary>
+    /// Validates that unexpected values provided for EntityCacheOptions in the JSON config result in
+    /// a failure to deserialize the runtime config.
+    /// </summary>
+    /// <param name="entityCacheConfig">Escaped JSON string defining entity cache configuration.</param>
+    [DataRow(@",""cache"": { ""enabled"": true, ""ttl-seconds"": 2147483648 }", DisplayName = "EntityCacheOptions.TtlSeconds set to Int.MaxValue+1 is invalid for parser.")]
```
aaronburtle, line 82: Can you add a case for `""ttl-seconds"": 0` and also some other small values, perhaps -1, and a small real number like .0001?

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
```diff
@@ -0,0 +1,242 @@
...
+            Assert.IsNotNull(config.Runtime?.CacheEnabled, message: "Expected global cache property to be non-null.");
+        }
+    }
+
+    /// <summary>
+    /// Validates that unexpected values provided for global runtime.cache property in the JSON config result in
+    /// a failure to deserialize the runtime config.
+    /// </summary>
+    /// <param name="globalCacheConfig">Escaped JSON string defining entity cache configuration.</param>
+    [DataRow(@",""cache"": {}", DisplayName = "Global cache property is empty: deserialization fails.")]
```
aaronburtle, line 149: what happens if someone puts a 1 or a 0 instead of true or false? might be a case to test here since that should be invalid.

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
```diff
@@ -0,0 +1,242 @@
...
+    /// <summary>
+    /// Returns a JSON string of the runtime config with the test-provided
+    /// cache configuration.
+    /// </summary>
+    /// <param name="globalCacheConfig">Escaped JSON string defining global cache configuration.</param>
+    /// <param name="entityCacheConfig">Escaped JSON string defining entity cache configuration.</param>
+    /// <returns></returns>
+    private static string GetRawConfigJson(string globalCacheConfig, string entityCacheConfig)
+    {
+        StringBuilder expectedRuntimeConfigJson = new(
```
aaronburtle, line 180: is the StringBuilder needed? I would expected you could just return this directly as a string. Also, is it possible to remove the connection string or will that break the parsing in the test?

aaronburtle (approved):
Couple questions about testing but looks good to me!

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
seantleonard: thanks for the suggestion, adding. all three of those cases which would be invalid.

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
seantleonard: added these as test cases in `GlobalCacheOptionsDeserialization_InvalidJson`. 1 and 0 are not boolean values so will fail deserialization.

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
seantleonard: With the multiple side effects that occur due to stringbuilder.Replace calls, I'd say it's better to have stringbuilder since you're not constantly copying/allocating strings for the test.

seantleonard: also, yes i removed the connection string and all still works.

src/Config/ObjectModel/RuntimeOptions.cs
```diff
@@ -1,11 +1,14 @@
...
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
 public record RuntimeOptions(
     RestRuntimeOptions? Rest,
     GraphQLRuntimeOptions? GraphQL,
     HostOptions? Host,
     string? BaseRoute = null,
-    TelemetryOptions? Telemetry = null);
+    TelemetryOptions? Telemetry = null,
+    [property: JsonPropertyName("cache")] bool? CacheEnabled = null);
```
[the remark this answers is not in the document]
seantleonard: updated to use same `EntityCacheOptions` object both at the entity level and globally. Will use a separate PR to update naming in code. That would change almost all aspects of this pr hiding any additional tests/logic i added within name change diffs.

aaronburtle (approved):

src/Service.Tests/Caching/CachingConfigDeserializationTests.cs
seantleonard: I addressed this with tests to ensure the `cache` property and its values are serialized when expected, and are not serialized when not expected (not explicitly defining ttl)

PR #1868, opened 2023-11-08
Add telemetry client only if it is enabled

Aniruddh25:
## Why make this change?

- Saw this exception while running locally via VS:

image

- it is triggered asynchronously when the grapqhql schema is being generated in the background
- Also, noticed telemetry client is being added irrespective of it is enabled or not.

## What is this change?
- Add AppInsights Telemetry only when it is enabled.
- Use `TryAdd` instead of simply doing an `Add` to the dictionary of global properties in the Initialize function that is used to initialize any telemetry object.
- Depending on if the telemetry object had already been initialized with the 2 global properties we intend to add, they may or may not be already present. Examples of such telemetry objects include `MetricsTelemetry`, `DependencyTelemetry` etc. 

## How was this tested?

- Reran the engine to verify the exception was gone upon starting.
- Added asserts in TelemetryTests to verify the client is created only when telemetry is enabled.

### DIFF
--- a/src/Service.Tests/Configuration/TelemetryTests.cs
+++ b/src/Service.Tests/Configuration/TelemetryTests.cs
@@ -8,12 +8,16 @@ using System.Net.Http;
 using System.Net.Http.Json;
 using System.Threading.Tasks;
 using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Core.Configurations;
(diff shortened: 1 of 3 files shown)


abhishekkumams (approved):
Thanks for addressing this.

aaronburtle (approved):
LGTM! Thanks for adding quickly

Issue #1869, opened 2023-11-08
Hot Reload of the OpenApi Document


aaronburtle:
The Open Api Document is generated on startup, and so to properly update the doc when a hot reload occurs we need to trigger that create document call as part of the hot reload process. This issues covers implementation of that plus the other changes needed to make sure that the open api docs properly update during a hot reload.

seantleonard:
Moving this + hot reload related tasks to post GA -> tracking via 1.2rc milestone.

seantleonard:
Current OpenApiDocumentor implementation has function `CreateDocument()` which has this preliminary check
```csharp
            if (_openApiDocument is not null)
            {
                throw new DataApiBuilderException(
                    message: DOCUMENT_ALREADY_GENERATED_ERROR,
                    statusCode: HttpStatusCode.Conflict,
                    subStatusCode: DataApiBuilderException.SubStatusCodes.OpenApiDocumentAlreadyExists);
            }
```

Enabling hot reload involes removing this check and rethinking what situations necessitate calling `OpenApiDocumentor::CreateDocument()` again. 

At first glance, adding a hot reload function directly to IOpenApiDocumentor or OpenApiDocumentor is not appropriate. Another service should call `openapidocumentor.CreateDocument()` when new config comes in.

Consider:
- Should we lock or wait for new createdocument operation to complete before allowing new ` GET https://dab/api/openapi` requests to be fulfilled?
- Do we need a feature switch for anything hot reload related here? (probably not, but please consider as part of design).

PR #1870, opened 2023-11-08
Hot Reload the OpenApi Document

aaronburtle:
## Why make this change?

Part of the ongoing work for https://github.com/Azure/data-api-builder/issues/67

this closes https://github.com/Azure/data-api-builder/issues/1869

This change will reload the open api docs during a hot reload scenario.

## What is this change?

Update the documentor class to get a new runtimeconfig rather than storing one internally. Then add a documentor to the runtimconfig provider, and pass it into the config file watcher. We then can call the create document function when we are hot reloading.

## How was this tested?

Testing is still needed. Will keep as draft until tested.

## Sample Request(s)


ayush3797:
This will also hot reload only the runtimeconfig section in the openapi doc? If rest.enabled changes from true to false, will the openapi doc no longer be available?

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft Corporation.
```
abhishekkumams, line 1: why is this class being shown as newly added, this was already a part of your last PR that got merged.

src/Service/Startup.cs
```diff
@@ -79,10 +79,10 @@ public void ConfigureServices(IServiceCollection services)
                 null);
             IFileSystem fileSystem = new FileSystem();
             FileSystemRuntimeConfigLoader configLoader = new(fileSystem, configFileName, connectionString);
+            services.AddSingleton<IOpenApiDocumentor, OpenApiDocumentor>();
```
abhishekkumams, line 82: Add a check to make sure we do this only when Rest is enabled. ref:  https://github.com/Azure/data-api-builder/pull/1872/files

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1337,6 +1337,67 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+
+                // we write to the config which is being monitored and sleep to allow hot reload to occur
+                fileSystem.File.WriteAllText(configFilePath, hotReloadConfig.ToJson());
+                Thread.Sleep(1000);
+                requestPath += "/MappedBookmarks";
+                request = new(httpMethod, requestPath);
+                response = await client.SendAsync(request);
+                Assert.IsTrue(response.StatusCode is HttpStatusCode.OK);
+                // overwrites the changes we made when hot reloading
+                fileSystem.File.WriteAllText(configFilePath, config.ToJson());
```
abhishekkumams, line 1397: what is the use of this last line in testing?

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -166,8 +167,9 @@ public void CreateDocument()
         private OpenApiPaths BuildPaths()
         {
             OpenApiPaths pathsCollection = new();
+            RuntimeConfig config = _configProvider.GetConfig();
```
abhishekkumams, line 170: should we pass the config here from `CreateDocument()` ? is there a possibility the config that is sourced in `CreateDocument()` is different than the config sourced in `BuildPaths()`?

seantleonard:
closing since this is really old. Please reopen a new PR with conflicts resolved and latest codebase changes included.

PR #1873, opened 2023-11-09
DataWarehouse query support for data-api-builder

rohkhann:
## Why make this change?
Adds support for Data warehouse to data-api-builder project.

## What is this change?
Created a new DWSQLQueryBuilder. DWSQL reuses metadataprovider, dbexceptionParser and queryExecutor from mssql as those files would be same.
Also since we are jsonifying and returning a string, we return the empty string for null values. For string objects, this means we will return empty whenever there is null, for all other objects we ensure to continue to return null.
This change DOES NOT support following:
1. Mutations (Update, delete, execute)
2. Stored procedures

These will come in later prs.

## How was this tested?
1. Manual integration test done with local datawarehouse. 
2. Ran the schema file on a DW to ensure test schema is generated correctly. 
3. Added automated integration tests.
4. Note: With string aggregate a string is returned as the json. The graphql code however, formats the data according to the return type expected by the query, hence deep equal to json is not working. For now the tests are being run against the mssql db and hence the mssql results can be compared to the return of the dw query executed against the mssql db.


src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+            string stringAgg = string.Empty;
+            int i = 0;
+            // Iterate through all the columns and build the string_agg
+            foreach (LabelledColumn column in structure.Columns)
+            {
+                // Generate the col value.
+                bool subQueryColumn = structure.IsSubqueryColumn(column);
+                string col_value = column.Label;
+
+                // If the column is not a subquery column and is not a string, cast it to string
```
aaronburtle, line 38: why do we want to cast this to a string when it isn't one already and not a subquery column?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+            return string.Join(", ",
+                structure.Columns.Select(
+                    c => structure.IsSubqueryColumn(c) ?
+                        WrapSubqueryColumn(c, structure.JoinQueries[c.TableAlias!]) + $" AS {QuoteIdentifier(c.Label)}" :
+                        Build(c)
+            ));
+        }
+
+        /// <summary>
+        /// Builds the query to fetch result set details of stored-procedure.
```
aaronburtle, line 163: is this summary still valid for data warehouse?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+        /// <summary>
+        /// Builds the query to get all the read-only columns in an DWSql table.
+        /// For DWSql, the columns:
+        /// 1. That have data_type of 'timestamp', or
+        /// 2. are computed based on other columns,
+        /// are considered as read only columns. The query combines both the types of read-only columns and returns the list.
+        /// </summary>
+        /// <param name="schemaOrDatabaseParamName">Param name of the schema/database.</param>
+        /// <param name="tableParamName">Param name of the table.</param>
+        /// <returns></returns>
```
aaronburtle, line 184:
```suggestion
        /// <returns>String representing the query needed to get a combined list of read only columns.</returns>
```

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -73,7 +73,7 @@ public class MsSqlQueryExecutor : QueryExecutor<SqlConnection>
                   httpContextAccessor)
         {
             RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
-            IEnumerable<KeyValuePair<string, DataSource>> mssqldbs = runtimeConfig.GetDataSourceNamesToDataSourcesIterator().Where(x => x.Value.DatabaseType == DatabaseType.MSSQL);
+            IEnumerable<KeyValuePair<string, DataSource>> mssqldbs = runtimeConfig.GetDataSourceNamesToDataSourcesIterator().Where(x => x.Value.DatabaseType == DatabaseType.MSSQL || x.Value.DatabaseType == DatabaseType.DWSQL);
```
aaronburtle, line 76:
```suggestion
            IEnumerable<KeyValuePair<string, DataSource>> mssqldbs = runtimeConfig.GetDataSourceNamesToDataSourcesIterator().Where(x => x.Value.DatabaseType is DatabaseType.MSSQL || x.Value.DatabaseType is DatabaseType.DWSQL);
```

nit: prefer is to ==

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Core.Models;
+using Microsoft.Data.SqlClient;
+
+namespace Azure.DataApiBuilder.Core.Resolvers
+{
+    /// <summary>
+    /// Class for building DwSql queries.
+    /// </summary>
+    public class DWSqlQueryBuilder : BaseSqlQueryBuilder, IQueryBuilder
```
vadeveka, line 14: NIT: might be useful to stick to camelcase convention i.e. DwSql along the lines of MsSql

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+                i++;
+
+                // Add comma if not last column. example: {"id":"1234","name":"Big Company"}
+                // the below ensures there is a comman after id but not after name.
+                if (i != structure.Columns.Count)
+                {
+                    stringAgg += ",";
+                }
+            }
+
```
vadeveka, line 55: NIT: can we add the sample content of stringAgg as code comment here for better understanding?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+                                    Build(structure.Predicates),
+                                    Build(structure.PaginationMetadata.PaginationPredicate));
+            string columns = WrappedColumns(structure);
+            string orderBy = $" ORDER BY {Build(structure.OrderByColumns)}";
+
+            string query = $"SELECT TOP {structure.Limit()} {columns}"
+                + $" FROM {fromSql}"
+                + $" WHERE {predicates}"
+                + orderBy;
+            return query;
```
vadeveka, line 102: Can we add the sample content of query as code comment for better understanding?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        }
+
+        /// <summary>
+        /// Build internal query for DW.
+        /// This will generate the query that will return results in sql.
+        /// Results of this query will be jsonified and returned to the user.
+        /// </summary>
+        public string BuildSqlQuery(SqlQueryStructure structure)
```
vadeveka, line 80: Should this be private method?

vadeveka:
I presume relationships are not supported either - did you want to mention in the description?

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: This is because STRING_AGG which we are using to create the json requires strings. Hence all other data types need to be cast into a string.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: Stored procedures are supported in DW. This will be addressed in later pr's

src/Core/Services/ResolverMiddleware.cs
```diff
@@ -218,9 +219,20 @@ private static object PreParseLeaf(IMiddlewareContext context, string leafJson)
...
+            {
+                // if the field asked for is not a string type but element return kind is a string
+                // check rawtext, if rawtext is empty then it is a null value.
+                if (element.ValueKind == JsonValueKind.String && string.IsNullOrEmpty(element.ToString()))
+                {
+                    dwNullCheck = true;
+                }
+            }
+
+            return (string.IsNullOrEmpty(element.ToString()) && element.GetRawText() == "null") || dwNullCheck;
```
seantleonard, line 235: Can this be simplified to the following where new param is not needed?:
```suggestion
return element.ValueKind is not (JsonValueKind.Undefined or JsonValueKind.Null)
```

seantleonard (commented):
still have more files to review.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: Fixed.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: Added.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: fixed.

src/Core/Resolvers/MsSqlQueryExecutor.cs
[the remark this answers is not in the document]
rohkhann: as per my understanding session context is supported https://learn.microsoft.com/en-us/sql/t-sql/functions/session-context-transact-sql?view=azure-sqldw-latest. I did my integrated test with a dw using an identity token so that is supported for sure.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+
+using System.Data.Common;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Core.Models;
+using Microsoft.Data.SqlClient;
+
+namespace Azure.DataApiBuilder.Core.Resolvers
+{
+    /// <summary>
+    /// Class for building MsSql queries.
```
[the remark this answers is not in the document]
rohkhann: Somehow, this comment only showed up now. Have fixed this in earlier commit.

src/Core/Services/ResolverMiddleware.cs
rohkhann: For DW, since we return a string the jsonValueKind will be string. We need to check if the return type being asked is not a string. if return type is int and we are returning empty from db then we can infer that the ISNULL check replaced the null with empty string while constructing the json and hence returned null. For strings however, we wont be able to differentiate between empty and null. Could not think of a way around this for now.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: fixed.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+    /// </summary>
+    public class DWSqlQueryBuilder : BaseSqlQueryBuilder, IQueryBuilder
+    {
+
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        /// <summary>
+        /// Builds the query to fetch result set details of stored-procedure.
+        /// result_field_name is the name of the result column.
+        /// result_type contains the sql type, i.e char,int,varchar. Using TYPE_NAME method
+        /// allows us to get the type without size constraints. example: TYPE_NAME for both
+        /// varchar(100) and varchar(max) would be varchar.
+        /// is_nullable is a boolean value to know if the result column is nullable or not.
+        /// </summary>
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Services/ResolverMiddleware.cs
seantleonard: Can you give an example of the JSON contents of the JsonElement in this case?

Is this what you mean?
```json
{
"propertyName": "null"
}
```

Is this not possible?
```json
{
"propertyName": null
}
```

rohkhann: it will return "propertyName": "", I chose to keep it string.empty but can be changed to null. Will not impact the string case. for all other datatypes we would have to make a check and determine if the db value is null.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// SELECT COALESCE('['+STRING_AGG('{"title":"' + STRING_ESCAPE(ISNULL(title,''),'json') + '"}',', ')+']','[]')
+        /// FROM (
+        ///     SELECT TOP 100 [table0].[title] AS [title]
+        ///     FROM [dbo].[books] AS [table0]
+        ///     WHERE 1 = 1 ORDER BY [table0].[id] ASC
+        ///     ) AS [table0]
+        /// </summary>
+        public string Build(SqlQueryStructure structure)
+        {
+            string columns;
```
seantleonard, line 36: Can you note the significance of one set to string.Empty and the other using the default of null?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -73,7 +73,7 @@ public class MsSqlQueryExecutor : QueryExecutor<SqlConnection>
                   httpContextAccessor)
         {
             RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
-            IEnumerable<KeyValuePair<string, DataSource>> mssqldbs = runtimeConfig.GetDataSourceNamesToDataSourcesIterator().Where(x => x.Value.DatabaseType == DatabaseType.MSSQL);
+            IEnumerable<KeyValuePair<string, DataSource>> mssqldbs = runtimeConfig.GetDataSourceNamesToDataSourcesIterator().Where(x => x.Value.DatabaseType is DatabaseType.MSSQL || x.Value.DatabaseType is DatabaseType.DWSQL);
```
Aniruddh25, line 76: Does the RuntimeConfig property about `SqlSourceUsed` need to be modified for multiple config scenario where one of the configs is Dw?

Aniruddh25 (commented):
Posting comments so far, need to go through QueryBuilder

Aniruddh25 (commented):
Posting comments so far, need to go through QueryBuilder

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+            string stringAgg = string.Empty;
```
Aniruddh25, line 37: Use StringBuilder to do concatenation

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+            int i = 0;
+            // Iterate through all the columns and build the string_agg
+            foreach (LabelledColumn column in structure.Columns)
+            {
+                // Generate the col value.
+                bool subQueryColumn = structure.IsSubqueryColumn(column);
+                string col_value = column.Label;
+
+                // If the column is not a subquery column and is not a string, cast it to string
+                if (!structure.IsSubqueryColumn(column) && structure.GetColumnSystemType(column.ColumnName) != typeof(string))
```
Aniruddh25, line 47: What happens to subquery columns?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// SELECT TOP 100 [table0].[title] AS[title],
+        /// JSON_QUERY([table1_subq].[data]) AS[publishers],
+        /// JSON_QUERY(COALESCE('['+[table7_subq].[data]+']', '[]')) AS[authors]
+        /// FROM dbo_books AS[table0]
+        /// OUTER APPLY(SubQuery generated by recursive call to build function, will create the _subq tables)
+        /// </summary>
+        private string BuildSqlQuery(SqlQueryStructure structure)
+        {
+            string dataIdent = QuoteIdentifier(SqlQueryStructure.DATA_IDENT);
+            string fromSql = $"{QuoteIdentifier(structure.DatabaseObject.SchemaName)}.{QuoteIdentifier(structure.DatabaseObject.Name)} " +
```
Aniruddh25, line 97: Use `stringbuilder`

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+            string predicates = JoinPredicateStrings(
+                                    structure.GetDbPolicyForOperation(EntityActionOperation.Read),
+                                    structure.FilterPredicates,
+                                    Build(structure.Predicates),
+                                    Build(structure.PaginationMetadata.PaginationPredicate));
+            string columns = WrappedColumns(structure);
+            string orderBy = $" ORDER BY {Build(structure.OrderByColumns)}";
+
+            string query = $"SELECT TOP {structure.Limit()} {columns}"
+                + $" FROM {fromSql}"
```
Aniruddh25, line 113: Is TOP the most performant for DW ?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        /// <summary>
+        /// Build columns and wrap columns which represent join subqueries
+        /// </summary>
+        private string WrappedColumns(SqlQueryStructure structure)
+        {
+            return string.Join(", ",
+                structure.Columns.Select(
+                    c => structure.IsSubqueryColumn(c) ?
+                        WrapSubqueryColumn(c, structure.JoinQueries[c.TableAlias!]) + $" AS {QuoteIdentifier(c.Label)}" :
```
Aniruddh25, line 171: how can we be sure TableAlias is non null here?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// 2. are computed based on other columns,
+        /// are considered as read only columns. The query combines both the types of read-only columns and returns the list.
+        /// </summary>
+        /// <param name="schemaOrDatabaseParamName">Param name of the schema/database.</param>
+        /// <param name="tableParamName">Param name of the table.</param>
+        /// <returns>String representing the query needed to get a combined list of read only columns.</returns>
+        public string BuildQueryToGetReadOnlyColumns(string schemaParamName, string tableParamName)
+        {
+            // For 'timestamp' columns sc.is_computed = 0.
+            string query = "SELECT ifsc.column_name from sys.columns as sc INNER JOIN INFORMATION_SCHEMA.COLUMNS as ifsc " +
```
Aniruddh25, line 195: Can this metadata query correctly identify read only columns? Do we have tests for DW?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -0,0 +1,212 @@
...
+                $"AND sc.object_id = object_id({schemaParamName}+'.'+{tableParamName}) and ifsc.table_name = {tableParamName} " +
+                $"AND ifsc.table_schema = {schemaParamName} and ifsc.column_name = sc.name;";
+
+            return query;
+        }
+
+        /// <inheritdoc/>
+        public string BuildFetchEnabledTriggersQuery()
+        {
+            string query = "SELECT STE.type_desc FROM sys.triggers ST inner join sys.trigger_events STE " +
```
Aniruddh25, line 206: Same, is this query to identify triggers are enabled and if yes, does it help us update table with triggers as well - Perhaps, this would come with the support for update..

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
```diff
@@ -0,0 +1,304 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Azure.DataApiBuilder.Service.Tests.SqlTests.GraphQLQueryTests
+{
+    /// <summary>
+    /// Test GraphQL Queries validating proper resolver/engine operation.
```
Aniruddh25, line 10: Need to have similar tests for REST API GET query support for DW.

Aniruddh25 (changes requested):
Looks good, but left some suggestions.

src/Core/Resolvers/MsSqlQueryExecutor.cs
rohkhann: I have modified the runtimeconfig sqlSourceUsed in this pr:
image

src/Service.Tests/DatabaseSchema-DwSql.sql
```diff
@@ -0,0 +1,135 @@
+-- Copyright (c) Microsoft Corporation.
+-- Licensed under the MIT License.
+
+BEGIN TRANSACTION
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: There is no system type or db type for subquery columns like say authors. They are whole entities being represented as columns. The recursion and the logic for the subquery portion handles returning these as string and hence there is no explicit cast needed here.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: fixed.

src/Core/Services/ResolverMiddleware.cs
rohkhann: This is now fixed. using alternatives discussed in dab sync up meeting.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: Fixed after review in dab sync up

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: fixed.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: Based on review with UC, this seemed to be ok. We are waiting on contact from DW team. any perf related improvements will likely come in following pr's as we keep enhancing the query.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -33,6 +33,13 @@ public static JsonDocument CreatePaginationConnectionFromJsonElement(JsonElement
             // Maintains the connection JSON object *Connection
             Dictionary<string, object> connectionJson = new();
 
+            // in dw we wrap array with "" and hence jsonValueKing is string instead of array.
+            if (root.ValueKind == JsonValueKind.String)
```
aaronburtle, line 37:
```suggestion
            if (root.ValueKind is JsonValueKind.String)
```

nit: prefer `is` to `==`

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
aaronburtle: Can a case with pagination and a case with filter predicate (where foo=bar) be added to? And same for REST.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -33,6 +33,13 @@ public static JsonDocument CreatePaginationConnectionFromJsonElement(JsonElement
             // Maintains the connection JSON object *Connection
             Dictionary<string, object> connectionJson = new();
 
+            // in dw we wrap array with "" and hence jsonValueKing is string instead of array.
```
Aniruddh25, line 36:
```suggestion
            // in dw we wrap array with "" and hence jsonValueKind is string instead of array.
```

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: For subquery columns, we use table alias to get the results in a subquery by doing outer apply. Same logic is used for mssql case as well.

src/Service.Tests/dab-config.DwSql.json
```diff
@@ -0,0 +1,931 @@
+{
+  "$schema": "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json",
+  "data-source": {
+    "database-type": "dwsql",
+    "connection-string": "Server=tcp:127.0.0.1,1433;Persist Security Info=False;User ID=sa;Password=REPLACEME;MultipleActiveResultSets=False;Connection Timeout=5;",
```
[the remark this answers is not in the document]
rohkhann: You would ideally want it to be Data Source= and used access token to access this string. However, given that we are testing this with mssql db in the pipeline I have kept the same format.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: This is the same query used in mssql that works for DW. This query is tested as part of the integration tests for DW

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: This would be handled in update. same query as mssql with no changes.

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
rohkhann: Added file for DW tests. All tests run successfully that are part of the RESTFINDbase class.

aaronburtle (approved):
Looks good, thanks for addressing issues quickly!

Aniruddh25 (approved):
Looks good overall, left few questions and suggestions.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -314,10 +314,13 @@ public async Task<bool> ReadAsync(DbDataReader reader)
             // Parse Results into Json and return
             if (dbDataReader.HasRows)
             {
-                // Make sure to get the complete json string in case of large document.
-                jsonResult =
-                    JsonSerializer.Deserialize<TResult>(
-                        await GetJsonStringFromDbReader(dbDataReader));
+                string jsonString = await GetJsonStringFromDbReader(dbDataReader);
+
+                if (!string.IsNullOrEmpty(jsonString))
```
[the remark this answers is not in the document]
rohkhann: Json string can be null or empty when the returned result from the db is NULL. This is one of the test cases we have. For other databases this case would return 0 rows, because we are not constructing any json in the sql query. in dw case this would return a single row with an empty string owing to the isNull check.

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1635,6 +1635,12 @@ public virtual async Task TestInvalidFilterParamQuery()
             DatabaseType dbType,
             string testEnvironment)
         {
+            if (dbType is DatabaseType.DWSQL)
```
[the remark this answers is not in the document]
rohkhann: fixed!

src/Service.Tests/dab-config.DwSql.json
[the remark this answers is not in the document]
rohkhann: Makes sense. fixed!

src/Core/Services/ResolverMiddleware.cs
```diff
@@ -220,13 +221,13 @@ private static object PreParseLeaf(IMiddlewareContext context, string leafJson)
...
         {
-            return string.IsNullOrEmpty(element.ToString()) && element.GetRawText() == "null";
+            return (string.IsNullOrEmpty(element.ToString()) && element.GetRawText() == "null");
         }
 
         protected static bool TryGetPropertyFromParent(IMiddlewareContext context, out JsonElement jsonElement)
         {
             JsonDocument result = JsonDocument.Parse(JsonSerializer.Serialize(context.Parent<JsonElement>()));
-            if (result is null)
+            if (result is null || (result.RootElement.ValueKind == JsonValueKind.Array && result.RootElement.EnumerateArray().Count() == 0))
```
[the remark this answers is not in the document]
rohkhann: fixed this, was initially used for testing. Dont think its needed now. I re ran all tests with removing this condition and all looks good.

src/Core/Resolvers/QueryExecutor.cs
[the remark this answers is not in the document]
Aniruddh25: So for other dbs, HasRows check returns false right? For those situations, we would log the subsequent message and return a default value for the `jsonResult`. in case of DW, we will miss logging this information and yet return a default value. 

Could you please:
1) Add this reasoning in the code,
2) Log the information even in case of DW..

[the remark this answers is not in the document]
rohkhann: For sure, let me add this.

src/Service.Tests/dab-config.DwSql.json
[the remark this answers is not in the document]
Aniruddh25: The sample [here](https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-connection-strings#sample-adonet-connection-string) says the connection string should be of this format: 

`Server=tcp:{your_server}.database.windows.net,1433;Database={your_database};User ID={your_user_name};Password={your_password_here};Encrypt=True;Connection Timeout=30;`

Would just specifying the DataSource be sufficient for an application that is connecting to a  standalone dedicated sql pool endpoint?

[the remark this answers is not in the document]
rohkhann: We have typically followed the datasource= format and then used a token to auth into the db. Can switch to above as it seems more clear mor scenario where we want to use username and password

PR #1874, opened 2023-11-09
Adds helpful error message for mutations executed without read permissions - SQL database types

### LINKED ISSUE #1464: GraphQL - Roles defined without read action returns error
When a role is defined without `read` action, mutations performed in the context of that role, returns an error response. However, at the database level, the mutation(update,insert, etc. action) succeeds. 

Repro steps:
1) In the config file available in the main branch, remove the `read` action for the `Anonymous` role
2) Execute the following mutation request in the context of `Anonymous` role

```graphql
mutation {
     updatebook(id: 1, item: {title: "Harry Potter and the Goblet of Fire"}){
          id
          title
     }
}
```

Mutation - Read role not defined bug

This scenario needs to be better handled.
a) For roles defined without read action, only graphQL special field `__typename` could be allowed. Other entity fields can be disallowed in the selection set.
b) Behavior needs to be better documented.

severussundar:
## Why make this change?

- Closes #1464 
  - At the moment, when a mutation operation is executed using a role which does not have read permissions (and Anonymous does not have read permissions), the mutation operation succeeds at the database layer but a generic AuthZ error message is returned. This could lead to the user concluding that the database operation did not go through because of the error response.  
  - This PR adds changes to return a helpful error message to the users - `The mutation operation {operation name} was successful but the current user is unauthorized to view the response due to lack of read permissions`
## What is this change?

 -  The objective of this change is to return a helpful error message that lets users know a) what has happened as a result of executing this mutation request b) the reason as why this error message is returned 
 - Presence of read permission is evaluated for the current role (with which the request was executed) as well as Anonymous role (because, for graphQL, read permissions is inherited from Anonymous role by other roles). If read permission is inapplicable, then a custom error message is returned.  
 

## How was this tested?
- [x] Integration Testing
- [x] Manual Testing

The request is executed in the context of Anonymous role. Anonymous role does not have read permissions configured but has the other three - create, update and delete configured.

**Response:**
image

**Create operation executed successfully**
image

## Error Messages

**Current Format (for mutation without read permission)**
image

**New Format introduced by this change (for mutation without read permission)**
image

The operation name such as `createbook`/`createPublisher` is included in the error message to help identify the operation which resulted in error when there are multiple mutations in the same request
image


ayush3797:
Do we need any tests to verify this change?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -170,6 +184,13 @@ public class SqlMutationEngine : IMutationEngine
                 throw _dabExceptionWithTransactionErrorMessage;
             }
 
+            if (!isReadPermissionConfigured)
+            {
+                throw new DataApiBuilderException(message: "The database operation was successful but the current user is unable to view the response due to lack of read permissions",
```
Aniruddh25, line 189:
```suggestion
                throw new DataApiBuilderException(message: "The mutation operation was successful but the current user is unable to view the response due to lack of read permissions",
```

Aniruddh25 (changes requested):
Looks good, although a test would be useful for catching any regressions.

src/Core/Resolvers/SqlMutationEngine.cs
seantleonard: beat me to this suggestion!

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -170,6 +184,13 @@ public class SqlMutationEngine : IMutationEngine
...
+                                                  statusCode: HttpStatusCode.Unauthorized,
```
seantleonard, line 190:
```suggestion
                                                  statusCode: HttpStatusCode.Forbidden,
```

Forbidden maps to HTTP 403 -> Authorization error
Unauthorized maps to HTTP 401 -> Unauthenticated/Authentication error.

src/Core/Resolvers/SqlMutationEngine.cs
seantleonard: Perhaps the end of the message should read:
> but the current user is unauthorized to view the response.

seantleonard (changes requested):
Need tests. Once those are added, I/you will be able to better determine if additional changes are needed to the engine code.

src/Core/Resolvers/SqlMutationEngine.cs
severussundar: I think it is worth explicitly calling out the lack of `read` permissions - as it is easy to think that just create permission should be sufficient for create mutation while actually we need create + read permissions.

So, I've updated the error message to - `The mutation operation was successful but the current user is unauthorized to view the response due to lack of read permissions`

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1424,6 +1424,243 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
             }
         }
 
+        /// <summary>
```
severussundar, line 1427: Alternative approaches considered for writing tests:

1. Removing the `read` action from `Anonymous` role from any of the existing entities leads to failure of tests. So, alternative options were to a) Have different config file setup for each test while using existing entities (or) b) Introduce new entities without read action in Anonymous role. 
I've gone ahead with having a different config file setup for each test.

2. I tried writing these two tests as a single DataTestMethod as there is an overlap in the mutation and cleanup steps. But that led to lots of conditionals and made the test logic difficult to follow. So, I've written these two as separate tests for better readability.

Aniruddh25 (approved):
LGTM, thank you for adding the integration tests!!

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -101,19 +101,31 @@ public class SqlMutationEngine : IMutationEngine
             // If authorization fails, an exception will be thrown and request execution halts.
             AuthorizeMutationFields(context, parameters, entityName, mutationOperation);
 
+            string roleName = GetHttpContext().Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER];
```
seantleonard, line 104: see `AuthorizeMutationFields()` later in this file. I suggest breaking out the code that resolves the role into a separate private helper which can be reused here. Essentially, resolving the role should occur by pulling the client role header out of the IMiddlewareContext and not the HTTP context.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -170,6 +184,13 @@ public class SqlMutationEngine : IMutationEngine
                 throw _dabExceptionWithTransactionErrorMessage;
             }
 
+            if (!isReadPermissionConfigured)
+            {
+                throw new DataApiBuilderException(message: "The mutation operation was successful but the current user is unauthorized to view the response due to lack of read permissions",
+                                                  statusCode: HttpStatusCode.Forbidden,
+                                                  subStatusCode: DataApiBuilderException.SubStatusCodes.AuthorizationCheckFailed);
+            }
```
seantleonard, line 192: in your pr description can you add a screenshot of the full GraphQL response before this change? by throwing an exception here, are we losing any details from the original error (before this change) and the more informative error message contents. e.g. are we missing the hotchocolate inserted error metadata that states which fields are error'd out?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1424,6 +1424,243 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+                      stock_by_pk(categoryid: 5001, pieceid: 5001) {
+                        categoryid
+                        pieceid
+                        categoryName
+                      }
+                    }";
+
+                string authToken = AuthTestHelper.CreateStaticWebAppsEasyAuthToken();
+
+                JsonElement queryResponse = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
```
seantleonard, line 1527: this next section is now a new test, I would break this into a separate test method. New role execution context, running a GraphQL query to validate db data which results in different assertion expectations.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1424,6 +1424,243 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+
+                _ = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
+                    client,
+                    server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                    query: deleteMutation,
+                    queryName: "deleteStock",
+                    variables: null,
+                    authToken: authToken,
+                    clientRoleHeader: AuthorizationResolver.ROLE_AUTHENTICATED);
+            }
```
seantleonard, line 1556: If this mutation fails then the database is not in an expected state if this test were to run again.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1424,6 +1424,243 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
...
+        /// Authenticated role has read permission configured. A pk query is executed in the context of Authenticated role to validate that a new
+        /// record was created in the database.
+        /// </summary>
+        [TestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        public async Task ValidateErrorMessageForMutationWithoutReadPermission()
+        {
+            GraphQLRuntimeOptions graphqlOptions = new(Enabled: true);
+            RestRuntimeOptions restRuntimeOptions = new(Enabled: false);
+
```
seantleonard, line 1444: Does this test really belong in ConfigurationTests.cs? Why not in authorization tests, or perhaps more consisely, its own test class. Then you can also add tests for additional edge cases explicitly to what you're testing.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -101,19 +101,31 @@ public class SqlMutationEngine : IMutationEngine
             // If authorization fails, an exception will be thrown and request execution halts.
             AuthorizeMutationFields(context, parameters, entityName, mutationOperation);
 
+            string roleName = GetHttpContext().Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER];
+
+            // The presence of READ permission is checked in the current role (with which the request is executed) as well as Anonymous role. This is because, for GraphQL requests,
+            // READ permission is inherited by other roles from Anonymous role when present.
+            bool isReadPermissionConfigured = _authorizationResolver.AreRoleAndOperationDefinedForEntity(entityName, roleName, EntityActionOperation.Read)
+                                              || _authorizationResolver.AreRoleAndOperationDefinedForEntity(entityName, AuthorizationType.Anonymous.ToString(), EntityActionOperation.Read);
+
```
seantleonard, line 110: read operation permissions checks are performed by hot chocolate by observing the `@authorize` directives present on the object type definition and/or its fields. 

so for my understanding, you just have this check here to determine whether to make follow-up roundtrip to the db to get the inserted data?

You'll also need to consider the case when an entity has the "read" permission defined, but excludes all fields. probably another opportunity to break this into a private helper.

seantleonard (changes requested):

src/Core/Resolvers/SqlMutationEngine.cs
severussundar: I've added a helper method and updated the code here and within AuthorizeMutationFields() to make use of the helper method for role determination

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: I've added the cleanup delete mutation within a finally block to ensure that it gets executed always. 

AFAICT, the delete GraphQL mutation should succeed always as the role has the required permissions.

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: I've introduced a helper method to perform this validaton

src/Service.Tests/Configuration/ConfigurationTests.cs
severussundar: I started with a new test class but then realized that a lot of the methods present in the ConfigurationTest class are necessary. This sounds like an opportunity to refactor the helper methods to another common class and update existing tests.  

I'm leaning towards having this test in ConfigurationTests class for now so that it is not blocker for 0.10.* release. There are some more tests that don't necessarily belong in ConfigurationTests class. I can work on a follow-up refactor PR once this gets merged.

src/Core/Resolvers/SqlMutationEngine.cs
severussundar: >read operation permissions checks are performed by hot chocolate by observing the @authorize directives present on the object type definition and/or its fields. so for my understanding, you just have this check here to determine whether to make follow-up roundtrip to the db to get the inserted data?

True, but HotChocolate performs this check when constructing the response (after all the database operations have taken place). But, the `select` query doesn't have to be executed when it can be determined that an error response is going to be returned anyways.  Hence, added this check before the select query execution.

> You'll also need to consider the case when an entity has the "read" permission defined, but excludes all fields. probably another opportunity to break this into a private helper.


It looks the execution of `select` query can be skipped for more scenarios. For example, in this mutation, I've excluded only `publisher_id` field. But, the selection set contains `publisher_id`. Even, in cases such as this, we don't return any data (for other valid fields). 
image

So, even there is one field in the selection set that is not queryable, the select query execution can be skipped as no data for any of the fields is returned. I'm looking into adding this logic change

src/Core/Resolvers/SqlMutationEngine.cs
seantleonard: The new error message is now incorrect then compared to the original error message returned to the user for the following scenario:

Execute a read (or a mutation follow-up read) within the anonymous role context when the read role is only assigned to users in the authenticated role. The error the user should receive is `UNAUTHENTICATED` as referenced in the original error message. The authentication checks occur first prior to the authorization checks. With HotChocolate/GraphQL, the `@authorize` directive ensures that:

1. The user is authenticated
2. The user's role membership and `x-ms-api-role` header map to a role defined in the directive.

This PR's changes would only result in a valid error message for a follow-up read when both of these criteria hold true:

1. The user is authenticated
2. Both system roles `anonymous` and `authenticated` lack assigned read permissions.

src/Core/Resolvers/SqlMutationEngine.cs
seantleonard: >  looks the execution of select query can be skipped for more scenarios. For example, in this mutation, I've excluded only publisher_id field. But, the selection set contains publisher_id. Even, in cases such as this, we don't return any data (for other valid fields). [This was performed using version 0.9.7 ]

Can you include the config you used here? Also, can you try this where no anonymous role is defined? only authenticated and other custom roles. Is the behavior the same?

src/Core/Resolvers/SqlMutationEngine.cs
severussundar: No, we get the same `UNAUTHENTICATED` error as before. 

Scenario: Anonymous has no read, Authenticated has read permission configured.

When a query is executed in the context of Anonymous role, the error message is as follows (with the changes in this PR):

image

For the same query, the error message in version 0.9.7 is as follows:

image

src/Core/Resolvers/SqlMutationEngine.cs
severussundar: Yes, the behavior is same when anonymous is not defined. 

```json
"Book": {
      "source": {
        "object": "books",
        "type": "table"
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "book",
          "plural": "books"
        }
      },
      "rest": {
        "enabled": true
      },
      "permissions": [
        {
          "role": "authenticated",
          "actions": [
            {
              "action": "create"
            },
            {
              "action": "update"
            },
            {
              "action": "delete"
            },
            {
               "action": "read",
               "fields": {
                   "exclude": [ "publisher_id" ] 
                }  
            }
          ]
        }
   ]
   ....
 }
```

Aniruddh25 (approved):

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1480,79 +1480,75 @@ public async Task ValidateErrorMessageForMutationWithoutReadPermission()
                 $"--ConfigFileName={CUSTOM_CONFIG}"
             };
 
+            string authToken = AuthTestHelper.CreateStaticWebAppsEasyAuthToken();
```
seantleonard, line 1483: This `authToken` looks to be used only for the `finally` block where the cleanup occurs. this line should be located in the cleanup code since it isn't used in your actual test.

severussundar: It is also used for validating that the database operation corresponding to the mutation request ran successfully. This check is done at L1529

seantleonard (approved):

PR #1876, opened 2023-11-09
Configured CosmosDB Local Emulator for running integration tests on the Windows agent

neeraj-sharma2592:
## Why make this change?

Currently, we are running integration tests against a production CosmosDB account. Recently, we've encountered 429 errors during these tests because the same production account is used for each OS agent. To address this issue, we have started using the CosmosDB Local Emulator when running integration tests on the Windows agent.

## What is this change?
Changed the Windows agent to windows-latest. This agent has an emulator preinstalled.
Removed the FileTransform task for the Windows agent. Temporarily removed the Mac-os job to prevent 429 errors. 
Todo: Use an emulator in the Mac-os agent and Linux agent.

## How was this tested?

- [x] Integration Tests

### DIFF
--- a/.pipelines/cosmos-pipelines.yml
+++ b/.pipelines/cosmos-pipelines.yml
@@ -18,11 +18,8 @@ strategy:
     linux:
       imageName: "ubuntu-latest"
       ADDITIONAL_TEST_ARGS: '--collect "XPlat Code coverage"'
-    mac:
-      imageName: "macOS-latest"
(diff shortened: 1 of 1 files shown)


Aniruddh25 (approved):
LGTM! Thanks for this mitigation.

seantleonard (approved):
Thanks for adding this change! looking forward to follow up on adjusting macos/linux runners.

PR #1881, opened 2023-11-14
In-Memory Cache 

seantleonard:
## Why make this change?

- Second set of work attributed to #1618 #1617 

## What is this change?
- Implements the design in #1801. The one behavior which isn't in this PR is caching GraphQL "list" queries. That will come in a separate commit once i resolve the errors that arise with generics and the DataReaderHandler delegates.
- Creates a `DabCacheService` which is used in database queries (REST/GraphQL) such that requests and responses are cached in memory. 
- This PR is not dependent on #1865. However, once that PR and this PR are merged, a third follow-up PR will make a small adjustment in `SqlQueryEngine.ExecuteAsync(...)` such that the `cacheEntryTtl` and `cacheEnabled` values are resolved from runtime config and used determine whether to use the cache and how to use the cache.

## How was this tested?

- [x] Integration Tests -> integration tests included which mock the QueryExecutor and test how the DabCacheService handles request metadata and caching method execution. The integration tests aim to test the **behavior** of the DabCacheService and intentionally do not make private methods public or break out key generation and entry size calculation into separate classes. Additionally, the integration tests are not meant to test the functionality of FusionCache. That library has its own tests which validate functionality. 

## Sample Request(s)

Until all related PRs are merged, this feature is gated by a feature flag. To add the feature flag, add the following snippet to the `AppConfig.json`:
```json
// Define feature flags in config file
    "FeatureManagement": {
        "CachingPreview": true, // Turn-On feature
    }
``` 

Point requests will be cached with a default 10 second ttl. You can try out whether response times are quicker for your use case.


src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -205,15 +213,29 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
...
+
+            JsonDocument? jsonDocument = await queryExecutor.ExecuteQueryAsync(
+                sqltext: queryString,
+                parameters: structure.Parameters,
+                dataReaderHandler: queryExecutor.GetJsonResultAsync<JsonDocument>,
+                httpContext: _httpContextAccessor.HttpContext!,
+                args: null,
+                dataSourceName: dataSourceName);
             return jsonDocument;
+
```
seantleonard, line 238: will remove

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -205,15 +213,29 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
...
-                    sqltext: queryString,
-                    parameters: structure.Parameters,
-                    dataReaderHandler: queryExecutor.GetJsonResultAsync<JsonDocument>,
-                    httpContext: _httpContextAccessor.HttpContext!,
-                    args: null,
-                    dataSourceName: dataSourceName);
+
+            if (await _featureManager.IsEnabledAsync("CachingPreview"))
+            {
+                if (_runtimeConfigProvider.GetConfig().SqlDataSourceUsed
```
Aniruddh25, line 219: Have we created a separate issue to track caching for CosmosDB GraphQL?

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -0,0 +1,345 @@
...
+            Assert.AreEqual(expected: true, actual: mockQueryExecutor.Invocations.Count is 1, message: ERROR_UNEXPECTED_INVOCATIONS);
+
+            IReadOnlyList<object> actualExecuteQueryAsyncArguments = mockQueryExecutor.Invocations[0].Arguments;
+            Assert.AreEqual(expected: queryMetadata.QueryText, actual: actualExecuteQueryAsyncArguments[0], message: "QueryText " + ERROR_FAILED_ARG_PASSTHROUGH);
+            Assert.AreEqual(expected: queryMetadata.QueryParameters, actual: actualExecuteQueryAsyncArguments[1], message: "Query parameters " + ERROR_FAILED_ARG_PASSTHROUGH);
+            Assert.AreEqual(expected: queryMetadata.DataSource, actual: actualExecuteQueryAsyncArguments[5], message: "Data source " + ERROR_FAILED_ARG_PASSTHROUGH);
+            Assert.AreEqual(expected: null, actual: result, message: "Expected factory to return a null result.");
+        }
+
+        [TestMethod]
```
seantleonard, line 209: need comment

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -0,0 +1,345 @@
...
+            JsonElement? result = await dabCache.GetOrSetAsync<JsonElement?>(queryExecutor: mockQueryExecutor.Object, queryMetadata: queryMetadata, cacheEntryTtl: cacheEntryTtl);
+
+            // Assert
+            Assert.IsFalse(mockQueryExecutor.Invocations.Count is 1, message: "Unexpected cache hit when cache entry size exceeded cache capacity.");
+            Assert.IsTrue(mockQueryExecutor.Invocations.Count is 2, message: ERROR_UNEXPECTED_INVOCATIONS);
+            Assert.AreEqual(expected: expectedDatabaseResponseJson, actual: result.ToString(), message: ERROR_UNEXPECTED_RESULT);
+        }
+
+        [TestMethod]
+        public async Task CacheServiceFactoryInvocationReturnsNull()
```
seantleonard, line 180: need comment

src/Core/Resolvers/SqlQueryEngine.cs
seantleonard: #1884

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -205,15 +213,29 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
...
+                && (!_runtimeConfigProvider.GetConfig().DataSource.GetTypedOptions<MsSqlOptions>()?.SetSessionContext ?? true))
```
Aniruddh25, line 220: If session context is set, are we NOT caching?

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
...
+        string cacheKey = CreateCacheKey(queryMetadata);
+        JsonElement? result = await _cache.GetOrSetAsync(
+               key: cacheKey,
+               async (FusionCacheFactoryExecutionContext<JsonElement> ctx, CancellationToken ct) =>
+               {
+                   // Need to handle undesirable results like db errors or null.
+                   JsonElement? result = await queryExecutor.ExecuteQueryAsync(
+                       sqltext: queryMetadata.QueryText,
+                       parameters: queryMetadata.QueryParameters,
+                       dataReaderHandler: queryExecutor.GetJsonResultAsync<JsonElement>,
```
Aniruddh25, line 62: The call `GetJsonResultAsync<JsonDocument>` takes `JsonDocument` if caching is disabled. Here we are giving it a `JsonElement` Why the difference and do we need the discrepancy?

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+        StringBuilder cacheKeyBuilder = new();
+        cacheKeyBuilder.Append(queryMetadata.DataSource);
+        cacheKeyBuilder.Append(KEY_DELIMITER);
+        cacheKeyBuilder.Append(queryMetadata.QueryText);
+        cacheKeyBuilder.Append(KEY_DELIMITER);
+        cacheKeyBuilder.Append(JsonSerializer.Serialize(queryMetadata.QueryParameters));
+        string cacheKey = cacheKeyBuilder.ToString();
+
+        if (_logger?.IsEnabled(LogLevel.Trace) ?? false)
```
Aniruddh25, line 92: should we disable this logging in late configured scenario?

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
...
+                   // Need to handle undesirable results like db errors or null.
+                   JsonElement? result = await queryExecutor.ExecuteQueryAsync(
+                       sqltext: queryMetadata.QueryText,
+                       parameters: queryMetadata.QueryParameters,
+                       dataReaderHandler: queryExecutor.GetJsonResultAsync<JsonElement>,
+                       httpContext: _httpContextAccessor.HttpContext!,
+                       args: null,
+                       dataSourceName: queryMetadata.DataSource);
+
+                   ctx.Options.SetSize(EstimateCacheEntrySize(cacheKey: cacheKey, cacheValue: result?.ToString()));
```
Aniruddh25, line 67: Is the cache entry size variable and can be modified per request?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -205,15 +213,29 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
...
-                await queryExecutor.ExecuteQueryAsync(
-                    sqltext: queryString,
-                    parameters: structure.Parameters,
-                    dataReaderHandler: queryExecutor.GetJsonResultAsync<JsonDocument>,
-                    httpContext: _httpContextAccessor.HttpContext!,
-                    args: null,
-                    dataSourceName: dataSourceName);
+
+            if (await _featureManager.IsEnabledAsync("CachingPreview"))
+            {
```
Aniruddh25, line 218: Could the SQL connection string specify the application intent as Read only or read write?
Should we base the decision of whether to cache or not on this insight? Or perhaps use it in addition to the feature flag in config

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -0,0 +1,345 @@
...
+
+            // Act
+            int cacheEntryTtl = 1;
+            JsonElement? result = await dabCache.GetOrSetAsync<JsonElement?>(queryExecutor: mockQueryExecutor.Object, queryMetadata: queryMetadata, cacheEntryTtl: cacheEntryTtl);
+
+            // Assert
+            Assert.AreEqual(expected: true, actual: mockQueryExecutor.Invocations.Count is 1, message: ERROR_UNEXPECTED_INVOCATIONS);
+
+            IReadOnlyList<object> actualExecuteQueryAsyncArguments = mockQueryExecutor.Invocations[0].Arguments;
+            Assert.AreEqual(expected: queryMetadata.QueryText, actual: actualExecuteQueryAsyncArguments[0], message: "QueryText " + ERROR_FAILED_ARG_PASSTHROUGH);
```
Aniruddh25, line 69: Would these indices change if their order is modified in the executor? If yes, Can we somehow enforce this modification to indices if the order of arguments changes?

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -0,0 +1,345 @@
...
+            {
+                {"param1", new DbConnectionParam(value: "param1Value") }
+            };
+
+            DatabaseQueryMetadata queryMetadata = new(queryText: "select * from MyTable", dataSource: "dataSource1", queryParameters: parameters);
+            DabCacheService dabCache = CreateDabCacheService(cache);
+
+            // Act
+            int cacheEntryTtl = 1;
+            JsonElement? result = await dabCache.GetOrSetAsync<JsonElement?>(queryExecutor: mockQueryExecutor.Object, queryMetadata: queryMetadata, cacheEntryTtl: cacheEntryTtl);
```
Aniruddh25, line 63: Can we also validate the result is as expectedDatabaseResponseJson?

Aniruddh25 (commented):
Looks good, posting comments so far, tests remaining..

Aniruddh25 (changes requested):
Very Comprehensive tests, left some questions.

src/Core/Resolvers/SqlQueryEngine.cs
ayush3797: Can you elaborate a bit more here, Ani? Are you suggesting we don't need a feature flag to turn on/off caching?

The application intent in the connection string is related to the level of permission which the user has, right? Even if this user has read-only connection, there might be some other user who is performing mutation on the data the first user is fetching. So if we based our decision on the application intent for the first user, he would get back stale data right?

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
```
severussundar, line 3: Something to consider:  We could have an interface like `ICacheService` and have `DabCacheService` implement that. This could be useful if we want to extend to support distributed caching, etc.

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// <typeparam name="JsonElement">Response payload</typeparam>
+    /// <param name="queryExecutor">Factory method. Only executed after a cache miss.</param>
+    /// <param name="queryMetadata">Metadata used to create a cache key or fetch a response from the database.</param>
+    /// <param name="cacheEntryTtl">Number of seconds the cache entry should be valid before eviction.</param>
+    /// <returns>JSON Response</returns>
+    /// <exception cref="Exception">Throws when the cache-miss factory method execution fails.</exception>
+    public async ValueTask<JsonElement?> GetOrSetAsync<JsonElement>(IQueryExecutor queryExecutor, DatabaseQueryMetadata queryMetadata, int cacheEntryTtl)
+    {
+        string cacheKey = CreateCacheKey(queryMetadata);
+        JsonElement? result = await _cache.GetOrSetAsync(
```
severussundar, line 54: For my understanding: Does the `.GetOrSetAsync()` method takes care of creating a cache entry incase of a cache miss?

src/Core/Services/Cache/DabCacheService.cs
seantleonard: Yes, cache entry size is based on contents of the key (request metadata) and response (db response).

The cache entry size is calculated once the response is received from the db and then inserted into the cache.

src/Service/Startup.cs
```diff
@@ -167,12 +171,24 @@ public void ConfigureServices(IServiceCollection services)
...
             AddGraphQLService(services);
+            services.AddFusionCache()
+                .WithOptions(options =>
+                {
+                    options.FactoryErrorsLogLevel = LogLevel.Debug;
+                    options.EventHandlingErrorsLogLevel = LogLevel.Debug;
+                })
+                .WithDefaultEntryOptions(new FusionCacheEntryOptions
+                {
+                    Duration = TimeSpan.FromMinutes(1)
```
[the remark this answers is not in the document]
seantleonard: default cache entry ttl when user doesn't specify in the runtime config for an entity.

src/Core/Services/Cache/DabCacheService.cs
seantleonard: We could, in that case we would need to evaluate logging across the dab solution to determine whether to log in late configured scenario wouldn't we? 

Are you highlighting a scenario specific to caching that should preclude logging in late bound config scenario?

src/Core/Resolvers/SqlQueryEngine.cs
seantleonard: I may be failing to see the implications of the connection string declaring read vs read/write. As long as data can be read from the db, the data can be cached based on
- the runtime configuration settings defined
- other checks performed like ensuring MsSql SetSessionContext feature is not enabled.

src/Core/Services/Cache/DabCacheService.cs
seantleonard: I did consider that. However, I have not determined whether another implementation of such an `ICacheService` is necessary to accommodate distributed caching versus extending the configuration of `FusionCache` within the original `DabCacheService`. Creating an interface is premature in this case when it is unclear what that interface should enforce in its contract. 
While an interface with only a single implementation could be helpful for testing, I did not want to mock `DabCacheService` in tests because mocking the query executor achieved what i wanted to test: behavior of the cache service as a result of results returned by the query executor.

src/Core/Services/Cache/DabCacheService.cs
seantleonard: 2/3. Working on response, tbd.

src/Service/Startup.cs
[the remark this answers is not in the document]
severussundar: I think you had mentioned after taking into consideration the lifetime of a container in the hosted scenario, the new choice of default is 45 seconds. If so, this needs to be updated

src/Core/Services/Cache/DabCacheService.cs
severussundar: Thanks for the explanation

src/Service/Startup.cs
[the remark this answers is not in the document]
seantleonard: updated to 5 seconds

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
seantleonard: They would change if the method signature of the `queryExecutor` changed the order. In that case the test would break because Moq does not support setting up a mock with named parameters (which may help mitigate the issue you highlight). 

Moq setup looks like the following with no named params supported (compiler error):
```csharp
                    mockQueryExecutor.Setup(x => x.ExecuteQueryAsync(
                        It.IsAny<string>(),
                        It.IsAny<IDictionary<string, DbConnectionParam>>(),
                        It.IsAny<Func<DbDataReader?, List<string>?, Task<JsonElement?>>>(),
                        httpContext,
                        args,
                        It.IsAny<string>()).Result)
                        .Returns((JsonElement?)null);
```

Additionally, the MOQ object's property `mockQueryExecutor.Invocations[0].Arguments` is a readonly list of objects. There is no key representing the parameter/argument name. I would only be able to check what object type each list entry is, but that falls short when you have multiple params of the same type (e.g. string).

src/Directory.Packages.props
```diff
@@ -28,6 +28,8 @@
     <PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="6.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.ApplicationInsights" Version="2.21.0" />
+    <PackageVersion Include="Microsoft.FeatureManagement" Version="3.0.0" />
```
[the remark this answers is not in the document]
seantleonard: yes

src/Core/Services/Cache/DabCacheService.cs
seantleonard: I intended to have the caching code avoid dealing with managing JsonDocument (IDisposable). Callers shouldn't need to know to dispose any created jsondocuments and so i pass back and utilize json element.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -199,20 +207,34 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
...
-            JsonDocument? jsonDocument =
-                await queryExecutor.ExecuteQueryAsync(
-                    sqltext: queryString,
-                    parameters: structure.Parameters,
-                    dataReaderHandler: queryExecutor.GetJsonResultAsync<JsonDocument>,
-                    httpContext: _httpContextAccessor.HttpContext!,
-                    args: null,
-                    dataSourceName: dataSourceName);
+
+            if (await _featureManager.IsEnabledAsync("CachingPreview") && runtimeConfig.IsCachingEnabled)
```
severussundar, line 218: nit: "CachingPreview" can be declared as a const string

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
...
+               key: cacheKey,
+               async (FusionCacheFactoryExecutionContext<JsonElement> ctx, CancellationToken ct) =>
+               {
+                   // Need to handle undesirable results like db errors or null.
+                   JsonElement? result = await queryExecutor.ExecuteQueryAsync(
```
severussundar, line 59: Something to consider: Would it be helpful to add log messages (maybe at debug level) that indicates whether the query was served from the cache or database?

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
...
+                size += cacheKey.Length * sizeof(char);
+                size += cacheValueSize * sizeof(char);
+                return size;
+            }
+        }
+        catch (OverflowException)
+        {
+            if (_logger?.IsEnabled(LogLevel.Trace) ?? false)
+            {
+                _logger.LogTrace(message: "Cache entry is too big.");
```
severussundar, line 131: The cache key for which the entry is big can be logged. Could be helpful for investigating/debugging

severussundar (approved):
Neat. Good usage of mocks for writing the integration tests. 

Few nits, otherwise, LGTM!

src/Core/Services/Cache/DabCacheService.cs
seantleonard: Removing for now, since cachekey will contain lots of request metadata that we may/may not want to log. Can revisit.

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -0,0 +1,137 @@
...
+    private long EstimateCacheEntrySize(string cacheKey, string? cacheValue)
+    {
+        if (string.IsNullOrWhiteSpace(cacheKey))
+        {
+            throw new ArgumentException(message: "Cache key should not be empty.");
+        }
+
+        try
+        {
+            checked
```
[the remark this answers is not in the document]
seantleonard: add this to method comment:
   ```csharp
 /// <seealso cref="https://learn.microsoft.com/dotnet/csharp/language-reference/statements/checked-and-unchecked"/>
```

File: src/Core/Resolvers/SqlQueryEngine.cs [hunk 20]
```diff
@@ -199,20 +207,34 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
...
+                {
+                    DatabaseQueryMetadata queryMetadata = new(queryText: queryString, dataSource: dataSourceName, queryParameters: structure.Parameters);
+                    JsonElement result = await _cache.GetOrSetAsync<JsonElement>(queryExecutor, queryMetadata, cacheEntryTtl: 5);
+                    byte[] jsonBytes = JsonSerializer.SerializeToUtf8Bytes(result);
+                    JsonDocument doc = JsonDocument.Parse(jsonBytes);
+                    return doc;
+                }
+            }
+
+            JsonDocument? jsonDocument = await queryExecutor.ExecuteQueryAsync(
```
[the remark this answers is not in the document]
seantleonard: renamed

src/Core/Services/Cache/DabCacheService.cs
seantleonard: Cache key logging can be enabled with appsettings.json configuration to change logging level for fusion cache.

src/Core/Services/Cache/DabCacheService.cs
seantleonard: Cache key logging can be enabled with appsettings.json configuration to change logging level for fusion cache.

abhishekkumams (approved):

src/Core/Services/Cache/DabCacheService.cs
seantleonard: can revisit if needed.

PR #1883, opened 2023-11-14
Fix MSSQL default value with built_in function support 

### LINKED ISSUE #1857: [Bug]: CLI create API incorrectly requires field value that has been excluded in the config
### What happened?

The DAB CLI hosted API for creating a new record incorrectly requires field value that has been specifically excluded in the config.

### Version

0.8.52

Azure SQL

GraphQL

### Relevant log output

_No response_

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #1857
- When DAB resolves default column values from SQL, we don't account for the fact that the default value can be a system function. Due to this while setting this methods as graphQL fields we get conversion errors.

## What is this change?

- before setting default value for graphQL input fields, we check if it is a built_in function. If yes, we do not set a default value from graphQL and let database handle it.

## How was this tested?

- [X] Integration Tests


src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -40,6 +41,64 @@ public async Task InsertMutation(string dbQuery)
...
+                        user_value
+                        current_date
+                        current_utc_date
+                        random_number
+                        start_of_day
+                        end_of_day
+                        default_string_with_paranthesis
+                        default_function_string_with_paranthesis
+                        default_integer
+                    }
```
seantleonard, line 65: can you give examples of raw values returned in these columns/fields?

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -40,6 +41,64 @@ public async Task InsertMutation(string dbQuery)
...
+                }
+            ";
+
+            JsonElement actual = await ExecuteGraphQLRequestAsync(graphQLMutation, graphQLMutationName, isAuthenticated: true);
+            string expected = await GetDatabaseResultAsync(dbQuery);
+            SqlTestHelper.PerformTestEqualJsonStrings(expected, actual.ToString());
+
+            // Assert the values
+            Assert.AreEqual(1234, actual.GetProperty("user_value").GetInt32());
```
seantleonard, line 74: How is this specific check relevant to what you're testing?

src/Service.Tests/SqlTests/GraphQLMutationTests/MsSqlGraphQLMutationTests.cs
```diff
@@ -58,6 +58,27 @@ public async Task InsertMutation()
             await InsertMutation(msSqlQuery);
         }
 
+        /// <summary>
+        /// <code>Do: </code> Inserts new row with random_number = 1234
+        /// <code>Check: </code> Mutation fails because the database policy (@item.name ne 'New publisher') prohibits insertion of records with name = 'New publisher'.
```
Aniruddh25, line 63: seems like this comment needs to be updated

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -40,6 +41,64 @@ public async Task InsertMutation(string dbQuery)
             SqlTestHelper.PerformTestEqualJsonStrings(expected, actual.ToString());
         }
 
+        /// <summary>
+        /// <code>Do: </code> Inserts new row and return all its columns
+        /// <code>Check: </code> A row is inserted in the table that has rows with default values as built_in methods.
+        /// it should insert it correctly with default values correctly handled by database.
+        /// </summary>
+        public async Task InsertMutationWithDefaultBuiltInFunctions(string dbQuery)
```
seantleonard, line 49: It's not clear which columns have default values and whether those values are `('string')``, `((1))` (int), or `((myFunc()))` which are functions

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -285,6 +286,20 @@ create table books_sold
     last_sold_on_date as last_sold_on,
 )
 
+CREATE TABLE default_with_function_table
+(
+    id INT PRIMARY KEY IDENTITY(1,1),
+    user_value INT,
+    [current_date] DATETIME DEFAULT GETDATE(),
+    current_utc_date DATETIME DEFAULT GETUTCDATE(),
```
Aniruddh25, line 294: we should add a test for NOT NULL column, but from GraphQL createItem try omitting the input.

Aniruddh25 (changes requested):
Needs to handle all cases..

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -2836,6 +2836,32 @@
...
       }
     },
+    "DefaultBuiltInFunction": {
+      "source": {
+        "object": "default_with_function_table",
+        "type": "table"
+      },
+      "graphql": {
+        "enabled": true
+      },
```
seantleonard, line 2846: Would your added test for this change still pass if default values were defined here in the runtime config? If so, I think it is necessary to have an additional test within `MutationBuilderTests` to validate the schema generated for this entity: validate the expected presence (or absence) of a default value on the mutation defintion

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -40,6 +41,64 @@ public async Task InsertMutation(string dbQuery)
...
+                }
+            ";
+
+            JsonElement actual = await ExecuteGraphQLRequestAsync(graphQLMutation, graphQLMutationName, isAuthenticated: true);
+            string expected = await GetDatabaseResultAsync(dbQuery);
+            SqlTestHelper.PerformTestEqualJsonStrings(expected, actual.ToString());
+
+            // Assert the values
+            Assert.AreEqual(1234, actual.GetProperty("user_value").GetInt32());
+            Assert.IsNotNull(actual.GetProperty("current_date").GetString());
```
seantleonard, line 75: perhaps checking that these values are not null **and** not empty

seantleonard (changes requested):

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1313,12 +1313,39 @@ internal string GetTablePrefix(string databaseName, string schemaName)
 
                     if (hasDefault)
                     {
-                        columnDefinition.DefaultValue = columnInfo["COLUMN_DEFAULT"];
+                        // If default value is a built_in function let database handle it.
+                        if (!IsBuiltInFunction(columnInfo["COLUMN_DEFAULT"].ToString()!))
```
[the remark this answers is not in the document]
abhishekkumams: we already check two lines above that the value is Not Null for getting `hasDefault`

src/Service.Tests/dab-config.MsSql.json
abhishekkumams: we don't support default values for tables or views in the runtime config yet.

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -40,6 +41,64 @@ public async Task InsertMutation(string dbQuery)
...
+            DateTime today = DateTime.Now;
+            DateTime tomorrow = today.AddDays(1);
+
+            // Assert the dates
+            // Note: Due to the time it takes to execute the code, the dates might not be exactly equal.
+            // So, comparing only the date part, or allow for a small difference.
+            Assert.AreEqual(today.Date, currentDate.Date);
+            Assert.AreEqual(today.Date, currentUtcDate.Date);
+            Assert.AreEqual(today.Date, startOfDay.Date);
+            Assert.AreEqual(tomorrow.Date, endOfDay.Date);
```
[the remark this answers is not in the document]
abhishekkumams: make sense. many of these asserts are redundant.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -130,7 +130,7 @@ private static InputValueDefinitionNode GenerateSimpleInputType(NameNode name, F
                 location: null,
                 f.Name,
                 new StringValueNode($"Input for field {f.Name} on type {GenerateInputTypeName(name.Value)}"),
-                defaultValue is not null ? f.Type.NullableType() : f.Type,
+                defaultValue is not null ? f.Type.NullableType() : (f.Type.IsNonNullType() ? f.Type.NullableType() : f.Type),
```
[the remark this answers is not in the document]
abhishekkumams: this is done so that we can create an input field with null for graphQL request, so that Database can handle it by itself.

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
seantleonard: how was this resolved?

abhishekkumams: I have added in the test summary highlighting the default columns value

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
abhishekkumams: makes sense, updated the test.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1313,12 +1313,39 @@ internal string GetTablePrefix(string databaseName, string schemaName)
...
+        /// </summary>
+        /// <param name="value">given value for database column</param>
+        /// <returns>true if given value is a built-in function</returns>
+        private static bool IsBuiltInFunction(string value)
+        {
+            if (value.StartsWith("(") && value.EndsWith(")"))
+            {
+                // remove starting and ending paranthesis
+                value = value.Substring(1, value.Length - 2);
+                if (!(value.StartsWith("'") && value.EndsWith("'")) && !(value.StartsWith("(") && value.EndsWith(")")))
```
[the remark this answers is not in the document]
abhishekkumams: nope, it won't be. That's why i have put negation in place.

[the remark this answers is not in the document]
abhishekkumams: i have simplified the logic

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
abhishekkumams: added in the test summary

config-generators/mssql-commands.txt
```diff
@@ -43,6 +43,7 @@ add MappedBookmarks --config "dab-config.MsSql.json" --source "mappedbookmarks"
 add FteData --config "dab-config.MsSql.json" --source fte_data --permissions "anonymous:*" --graphql "FteData:FteData"
 add InternData --config "dab-config.MsSql.json" --source intern_data --permissions "anonymous:*" --graphql "InternData:InternData"
 add BooksSold --config "dab-config.MsSql.json" --source "books_sold" --rest true --graphql "books_sold:books_sold" --permissions "anonymous:*"
+add DefaultBuiltInFunction --config "dab-config.MsSql.json" --source "default_with_function_table" --rest true --graphql true --permissions "anonymous:*"
```
Aniruddh25, line 46: Do we really need a new entity? Can this default function be a new column of an existing entity/table pair?

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -102,7 +102,11 @@ public override async Task PopulateTriggerMetadataForTable(string entityName, st
 
                     if (hasDefault)
                     {
-                        columnDefinition.DefaultValue = columnInfo["COLUMN_DEFAULT"];
+                        // If default value is a built_in function let database handle it.
+                        if (!IsBuiltInFunction(columnInfo["COLUMN_DEFAULT"].ToString()!))
+                        {
+                            columnDefinition.DefaultValue = columnInfo["COLUMN_DEFAULT"];
```
Aniruddh25, line 108: Do we also ensure that if the column has as builtin function as a default, we make it nullable so it is not required in the input?

abhishekkumams: yes, we are making sure of that in the `CreateMutationBuilder.cs` line:129

config-generators/mssql-commands.txt
abhishekkumams: That is a very good point. I went through the existing entities, all of them have specific use cases, adding some columns to fit in our use case seemed odd for them, like for a column with default function as string, or default string with paranthesis.

But keeping this point in mind we can be more cognizant from here.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -236,5 +240,24 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
                 return false;
             }
         }
+
+        /// <summary>
+        /// This method returns whether the given value is a built_in function or not.
+        /// built_in funtions are surrounded by single parenthesis,i.e. (getdate())
```
Aniruddh25, line 246: nit: Please clarify why you are considering these three categories. 1 reason I can think of is "given value could be in one of these 3 formats: 1) when it is a built-in function 2) when it is a string 3) when it is of type int/float". 
As a followup, what happens if the data type is DateTime with a specific default value?

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -126,11 +126,29 @@ private static InputValueDefinitionNode GenerateSimpleInputType(NameNode name, F
                 defaultValue = value.Fields[0].Value;
             }
 
+            ITypeNode type;
+            if (defaultValue is not null)
+            {
+                type = f.Type.NullableType();
```
Aniruddh25, line 132: Can ID field be nullable if it has a default value as a built in function in the database?

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -236,5 +240,24 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+        /// string will be with parenthesis and single quotes, i.e. ('value')
+        /// int/float will be with double parenthesis, i.e. ((20))
+        /// </summary>
+        /// <param name="columnDefinitionDefaultValue">given value for database column</param>
+        /// <returns>true if given value is a built-in function</returns>
+        private static bool IsBuiltInFunction(string columnDefinitionDefaultValue)
+        {
+            if ((columnDefinitionDefaultValue.StartsWith("('") && columnDefinitionDefaultValue.EndsWith("')"))
+                || (columnDefinitionDefaultValue.StartsWith("((") && columnDefinitionDefaultValue.EndsWith("))")))
```
Aniruddh25, line 255: Is there a separate issue created for mysql/postgresql assuming this logic wont work for those db types?

Aniruddh25: Actually looks like we have added tests for other db types as well. In thast case, shouldn't this function be in the base claas. And each derived db type metadataprovider inherit from that class?

Aniruddh25 (changes requested):
Overall looks good, but few followups needed

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
abhishekkumams: other Dbs are handling it differently than MSSQL. While MsSQL drivier is not parsing the inbuilt function, mysql and postgres are able to do it, due to which this issue is specific to MsSQL, which is verfied with the tests as well.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
abhishekkumams: Nope, primary key fields cannot be nullable.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
abhishekkumams: if DateTime field has a specific default value, it won't cause any problem since it would be parsed into DateTime.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -236,5 +240,25 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
                 return false;
             }
         }
+
+        /// <summary>
+        /// This method returns whether the given value is a built_in function or not.
+        /// ColmnDefinitiondefault value can be in one of the below formats.
```
Aniruddh25, line 246:
```suggestion
        /// colmnDefinitionDefaultValue can be in one of the below formats.
```

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
Aniruddh25: In that case, wouldn't MySQL/Postgres insert wrong values for default if in_built function is get_current_date() for example. If their drivers parse it and provide the current date value when GraphQL schema is generated, it would insert incorrect value at request time. How do we handle such incorrect mutations?

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
Aniruddh25: It may not be causing any parsing problems, but what is the format in which the default value is generated? Is it similar to int/float e.g.((01-04-2024)) ? or is it like a string ('01-04-2024') ? How about other scalar types than string? 
Basically my question, are we sure the default value format regardless of any other underlying type will be 1 of these 3 formats?

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
abhishekkumams: i may not have understood your point, but the problem with mssql is it is considering built_in function as string, so we let database handle it, but for other dbs, it's handling the built_in functions correctly so we insert exactly aligned with the column type.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
abhishekkumams: it will be like string. added a test for the same. Thanks for bringing up this case.

abhishekkumams: it will be like string ('01-04-2024'), added a test as well

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -40,6 +40,56 @@ public async Task InsertMutation(string dbQuery)
...
+        ///   "current_timestamp": "2023-12-15T16:24:48.267Z",
+        ///   "random_number": 0,
+        ///   "next_day": "2023-12-16T00:00:00.000Z",
+        ///   "default_string_with_paranthesis": "()",
+        ///   "default_function_string_with_paranthesis": "NOW()",
+        ///   "default_integer": 100,
+        ///   "default_date_string": "1999-01-08T10:23:54.000Z"
+        /// }
+        /// </summary>
+        public virtual async Task InsertMutationWithDefaultBuiltInFunctions()
```
seantleonard, line 62: You should add one more (distinct) test which executes this mutation twice and saves both results. In this assertion section, validate that the `execution1.current_date != execution2.current_date` and all other function fields that should return different values for each request. 

This will validate that the error scenario Ani highlighted does not occur.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -126,11 +126,29 @@ private static InputValueDefinitionNode GenerateSimpleInputType(NameNode name, F
...
+            }
+            else
+            {
+                // ID field cannot be nullable.
+                if (f.Type.NamedType().Name.Value != "ID")
+                {
+                    type = f.Type.NullableType();
+                }
```
seantleonard, line 140: This is new logic, not clear if/how it relates to the function support changes. This also needs a test to validate

seantleonard (changes requested):
two tests can be added to ensure coverage of changes made in this PR:
- Execute test mutation twice and compare the values to ensure they are different because that will prove that the functions are executing per request. (this will be important for proving valid behavior with mysql/pgsql too).
- Add test to validate the behavior you expect with ensuring GraphQL "ID" fields are non-nullable.

seantleonard:
I also reopened two previously resolved threads, those need to be revisited in order to close out.

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
Aniruddh25: Good suggestion - specifically for Pg/MySql

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
Aniruddh25: please add it to the comment summary section as well!

abhishekkumams:
Not Required

abhishekkumams:
Closed because of #1986

PR #1893, opened 2023-11-20
 Verifying the read access of anonymous users prior to transmitting the Mutation result.

### LINKED ISSUE #1464: GraphQL - Roles defined without read action returns error

neeraj-sharma2592:
## Why make this change?
Resolves #1464 for CosmosDB.

## What is this change?
Verifying the read access of anonymous users prior to transmitting the Mutation result

## How was this tested?

- [X] Integration Tests


severussundar (commented):
Need to add tests. After feedback, the format of the error message is slightly changed to include the mutation operation name as well. Please refer to [this PR](https://github.com/Azure/data-api-builder/pull/1874) for the latest format

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -233,6 +235,14 @@ private static async Task<ItemResponse<JObject>> HandleUpdateAsync(IDictionary<s
             }
         }
 
+        private void CheckReadAccess(string entityName)
+        {
+            if (!_authorizationResolver.AreRoleAndOperationDefinedForEntity(entityName, AuthorizationType.Anonymous.ToString(), EntityActionOperation.Read))
+            {
+                throw new Exception("The database operation was successful but the current user is unable to view the response due to lack of read permissions");
+            }
+        }
```
seantleonard, line 244: Why hardcoding to `AuthorizationType.Anonymous`? 

You should be acquiring the current role from the GraphQL execution context. See `AuthorizeMutationFields(...)` which has a snippet that you can use.

seantleonard (changes requested):
need tests too. @neeraj-sharma2592

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -233,6 +235,14 @@ private static async Task<ItemResponse<JObject>> HandleUpdateAsync(IDictionary<s
             }
         }
 
+        private void CheckReadAccess(string entityName)
```
abhishekkumams, line 238: Add Summary for the function

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -390,6 +402,246 @@ public async Task UpdateMutationWithOnlyTypenameInSelectionSet()
...
+
+        /// <summary>
+        /// For mutation operations, the respective mutation operation type(create/update/delete) + read permissions are needed to receive a valid response.
+        /// For graphQL requests, if read permission is configured for Anonymous role, then it is inherited by other roles.
+        /// In this test, Anonymous role has read permission configured. Authenticated role has only create permission configured.
+        /// A create mutation operation is executed in the context of Authenticated role and the response is expected to have no errors because
+        /// the read permission is inherited from Anonymous role.
+        /// </summary>
+        [TestMethod]
+        [TestCategory(TestCategory.MSSQL)]
```
severussundar, line 546: nit: TestCategory is incorrect. 

Can be removed as the test category is already defined at the class level

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -2,13 +2,25 @@
...
+using Azure.DataApiBuilder.Core.Configurations;
 using Azure.DataApiBuilder.Core.Resolvers;
 using Azure.DataApiBuilder.Service.Exceptions;
+using Azure.DataApiBuilder.Service.Tests.Configuration;
+using Microsoft.AspNetCore.TestHost;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.UriParser;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Sprache;
```
severussundar, line 23: It could be worth double checking if any of the functions from this library is being used?

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -258,6 +273,29 @@ private static async Task<ItemResponse<JObject>> HandleUpdateAsync(IDictionary<s
...
         }
 
+        /// <summary>
+        /// Helper method to get the role with which the GraphQL API request was executed.
+        /// </summary>
+        /// <param name="context">HotChocolate context for the GraphQL request</param>
+        private static string GetRoleOfGraphQLRequest(IMiddlewareContext context)
+        {
+            string role = string.Empty;
+            if (context.ContextData.TryGetValue(key: AuthorizationResolver.CLIENT_ROLE_HEADER, out object? value) && value is StringValues stringVals)
```
severussundar, line 283: `AuthorizeMutationFields()` method already contains logic to extract the role. For SQL, I refactored the logic for extracting the role from graphQL context to a separate helper function. It could be worth checking if that can be re-used here.

severussundar (changes requested):
Looks good for the most part. Some suggestions around re-using helper functions and few nits in tests

src/Core/Resolvers/CosmosMutationEngine.cs
neeraj-sharma2592: Sure. I'll check that

neeraj-sharma2592: I examined the AuthorizeMutationFields, and it indeed retrieves the role. While I could store that information in a global variable and access it within my scenario, I believe that approach might not be clean and may not effectively communicate the intention behind the change. I am more inclined to obtain the role from this helper function. Thoughts?

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -74,6 +74,21 @@ private async Task<JObject> ExecuteAsync(IMiddlewareContext context, IDictionary
                 _ => throw new NotSupportedException($"unsupported operation type: {resolver.OperationType}")
             };
 
+            string roleName = GetRoleOfGraphQLRequest(context);
+
+            // The presence of READ permission is checked in the current role (with which the request is executed) as well as Anonymous role. This is because, for GraphQL requests,
+            // READ permission is inherited by other roles from Anonymous role when present.
+            bool isReadPermissionConfigured = _authorizationResolver.AreRoleAndOperationDefinedForEntity(entityName, roleName, EntityActionOperation.Read)
+                                              || _authorizationResolver.AreRoleAndOperationDefinedForEntity(entityName, AuthorizationType.Anonymous.ToString(), EntityActionOperation.Read);
```
seantleonard, line 82:
```suggestion
                                              || _authorizationResolver.AreRoleAndOperationDefinedForEntity(entityName, AuthorizationResolver.ROLE_ANONYMOUS, EntityActionOperation.Read);
```

src/Core/Resolvers/CosmosMutationEngine.cs
seantleonard: `AuthorizeMutationFields()` should be updated to now call `GetRoleOfGraphQLRequest()` to eliminate code duplication.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -390,6 +399,245 @@ public async Task UpdateMutationWithOnlyTypenameInSelectionSet()
...
+                    JsonElement queryResponse = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
+                                                client,
+                                                server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                                                query: graphQLQuery,
+                                                queryName: queryName,
+                                                variables: null,
+                                                authToken: authToken,
+                                                clientRoleHeader: AuthorizationResolver.ROLE_AUTHENTICATED);
+
+                    Assert.IsFalse(queryResponse.TryGetProperty("errors", out _));
```
seantleonard, line 517: can you add param `message` with a value something like:  `message: "The response was not expected to have errors."`

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -390,6 +399,245 @@ public async Task UpdateMutationWithOnlyTypenameInSelectionSet()
...
+                        client,
+                        server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                        query: _createPlanetMutation,
+                        queryName: "createPlanet",
+                        variables: new() { { "item", input } },
+                        authToken: AuthTestHelper.CreateStaticWebAppsEasyAuthToken(),
+                        clientRoleHeader: AuthorizationResolver.ROLE_AUTHENTICATED
+                        );
+                        
+                    Assert.IsFalse(mutationResponse.TryGetProperty("errors", out _));
```
seantleonard, line 624: can you add param `message` with a value something like:  `message: "The response was not expected to have errors."`

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -390,6 +399,245 @@ public async Task UpdateMutationWithOnlyTypenameInSelectionSet()
...
+                    JsonElement mutationResponse = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
+                        client,
+                        server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                        query: _createPlanetMutation,
+                        queryName: "createPlanet",
+                        variables: new() { { "item", input } },
+                        clientRoleHeader: null
+                        );
+
+                    Assert.IsTrue(mutationResponse.ToString().Contains("The mutation operation createPlanet was successful but the current user is unauthorized to view the response due to lack of read permissions"));
```
seantleonard, line 495: can you add param `message` with a value something like:  `message: "The response was not expected to have errors."`

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -390,6 +399,245 @@ public async Task UpdateMutationWithOnlyTypenameInSelectionSet()
...
+        ///
+        /// A create mutation operation in the context of Anonymous role is executed and the expected error message is validated.
+        /// Authenticated role has read permission configured. A pk query is executed in the context of Authenticated role to validate that a new
+        /// record was created in the database.
+        /// </summary>
+        [TestMethod]
+        public async Task ValidateErrorMessageForMutationWithoutReadPermission()
+        {
+            const string SCHEMA = @"
+type Planet @model {
```
seantleonard, line 417: please add value to `@model` directive per https://learn.microsoft.com/azure/data-api-builder/database-specific-features#user-provided-graphql-schema

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -390,6 +399,245 @@ public async Task UpdateMutationWithOnlyTypenameInSelectionSet()
...
+        /// For graphQL requests, if read permission is configured for Anonymous role, then it is inherited by other roles.
+        /// In this test, Anonymous role has read permission configured. Authenticated role has only create permission configured.
+        /// A create mutation operation is executed in the context of Authenticated role and the response is expected to have no errors because
+        /// the read permission is inherited from Anonymous role.
+        /// </summary>
+        [TestMethod]
+        public async Task ValidateInheritanceOfReadPermissionFromAnonymous()
+        {
+            const string SCHEMA = @"
+type Planet @model {
```
seantleonard, line 546: please add value to `@model` directive per https://learn.microsoft.com/azure/data-api-builder/database-specific-features#user-provided-graphql-schema

severussundar (approved):
LGTM

seantleonard (approved):
thanks for updating with this behavior and addressing feedback!

Issue #1799, opened 2023-10-10
Investigate and fix use of encoding for after/nextlink


aaronburtle:
We are currently encoding the nextlink used for pagination using a base64 encode. We need to investigate how client applications are processing this afterlink and confirm if we really need to be encoding this ourselves, or allowing the client application to handle such encodings.

Ultimately, once this information is found, we need to fix the encodings so that the after cursor is opaque with valid base 64 encode.

seantleonard:
The part of the `nextLink` that is encoded is intended to be opaque to the client. Clients should not do anything with the contents of the encoded piece of the nextLink besides provide it to DAB. This allows us to make non-breaking changes to the structure of the encoded bits without the user being affected.

seantleonard:
- Confirm Base64 encoding is URL compatible.

Aniruddh25:
We need to investigate this because the url encoded `nextLink` or `after` clause no longer provides a valid base-64 encoded value. So, the link turns out to not be useful if it contains special characters. This was evident in the OpenAPI response as well.

PR #1895, opened 2023-11-21
Fix base64 encoding of $after query parameter value in NextLink

aaronburtle:
## Why make this change?

We were not providing a valid base64 encoding of the after cursor in the `NextLink` for pagination.

Closes https://github.com/Azure/data-api-builder/issues/1799

## What is this change?

Re-factor the logic of creating a `NextLink`. The desired behavior is for the `after` portion of the `NextLink` to be a valid base 64 encode, we change the logic to ensure this is the case.

## How was this tested?

Run against our current test suite.

## Sample Request(s)

Any valid pagination request which will have characters that would be uri encoded to contain invalid base 64 encode characters. Such as, `https://localhost:5001/rest/Book?$first=1&$orderby=title` which will include spaces in the `after` portion.


abhishekkumams (approved):
Looks good. Thanks for the fix.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -526,18 +527,13 @@ public static string Base64Decode(string base64EncodedData)
...
-                nvc = new();
-            }
-
+            string queryString = FormatQueryString(queryParameters: nvc);
             if (!string.IsNullOrWhiteSpace(after))
             {
-                nvc["$after"] = after;
+                string afterPrefix = string.IsNullOrWhiteSpace(queryString) ? "?" : "&";
+                queryString += $"{afterPrefix}{RequestParser.AFTER_URL}={after}";
             }
```
seantleonard, line 535: You could take advantage of the .NET helper inside our `FormatQueryString` function: `QueryHelpers.AddQueryString()`, to avoid creating your own conditional logic determining whether to use the `?` or `&` prefix.

Also, i see this logic ensures we no longer add a hard coded query param key name: `nvc["$after"]`. Was there a bug you identified that required updating the code/could a test be added to address that?

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -823,7 +822,7 @@ public async Task FindTestWithPaginationVerifMultiplePrimaryKeysInAfter()
                 queryString: $"?$first=1",
                 entityNameOrPath: _entityWithCompositePrimaryKey,
                 sqlQuery: GetQuery(nameof(FindTestWithPaginationVerifMultiplePrimaryKeysInAfter)),
-                expectedAfterQueryString: $"&$after={Uri.EscapeDataString(SqlPaginationUtil.Base64Encode(after))}",
+                expectedAfterQueryString: $"&$after={SqlPaginationUtil.Base64Encode(after)}",
```
seantleonard, line 825: I noticed some tests have `expectedAfterQueryString` base64 encode the after query param value directly, and some tests don't do that. Do you think it is worth making the tests consistent?

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -1015,7 +1014,7 @@ public async Task FindTestWithFirstTwoVerifyAfterFormedCorrectlyWithOrderBy()
             string after = $"[{{\"EntityName\":\"Author\",\"FieldName\":\"birthdate\",\"FieldValue\":\"2001-01-01\",\"Direction\":0}}," +
                            $"{{\"EntityName\":\"Author\",\"FieldName\":\"name\",\"FieldValue\":\"Aniruddh\",\"Direction\":0}}," +
                            $"{{\"EntityName\":\"Author\",\"FieldName\":\"id\",\"FieldValue\":125,\"Direction\":1}}]";
-            after = $"&$after={Uri.EscapeDataString(SqlPaginationUtil.Base64Encode(after))}";
+            after = $"&$after={SqlPaginationUtil.Base64Encode(after)}";
```
seantleonard, line 1017: For consistency, should other tests which have long `after` values get base64 encoded like this prior to being passed in `expectedAfterQueryString`

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: > You could take advantage of the .NET helper inside our `FormatQueryString` function: `QueryHelpers.AddQueryString()`, to avoid creating your own conditional logic determining whether to use the `?` or `&` prefix.



That helper function is what causes the incorrect URI encoding within the after, which is why it was separated out to how we have it now.


> Also, i see this logic ensures we no longer add a hard coded query param key name: `nvc["$after"]`. Was there a bug you identified that required updating the code/could a test be added to address that?

It is because we had these constants with the query param names as private inside of the request parser, but now we are using them in other parts of the code, so they are public. So now I think we should just be re-using them whenever possible so that everything has 1 source of truth for those names. Ultimately it might make sense to not have just a request parser class, since it is a bit weird to have the "parser" holding names like this, but rather have a request class that has these things and then also parsing methods or sub classes etc

for now this makes the most sense though I think.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
aaronburtle: if it doesn't take me too much time to get these consistent I will add it here, otherwise I will note it down to tackle later when there is more time.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
aaronburtle: ill look through the code and try to get it consistent, this is kind of an on-going theme where we have some inconsistencies in style.

src/Core/Resolvers/SqlPaginationUtil.cs
seantleonard: thanks for the clarification!

seantleonard: Optional, but would be helpful for future devs:
1. A comment above the following line to warn of the side effects that you learned about when making this fix:
```csharp
queryString = QueryHelpers.AddQueryString(queryString, key, queryParamValues);
```
2. Comment for `FormatQueryString(...)` warning of the side effects

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
seantleonard: since the test functionality isnt affected by this, I'm closing this thread. thanks for the background.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
seantleonard: closing since this doesn't affect the test functionality and can come later.

seantleonard (approved):
last nits. looks good!

ARDHA3:
Thanks @seantleonard, is there any documentation for creating issue?

seantleonard:
> Thanks @seantleonard, is there any documentation for creating issue?

Go to https://github.com/Azure/data-api-builder/issues and select `New Issue`, there is a template for create a Bug Report.

PR #1901, opened 2023-11-27
Fixing dw query builder to handle Unicode characters

rohkhann:
## Why make this change?

Dw pipeline test case was failing due to incorrect parsing for chinese characters. 
In addition, dw pipeline was not enabled for linux. 

## What is this change?

Use the N' prefix that handles unicode characters correctly when building the json in dw.
Added the db type in the config-generator.sh

## Tests
Unique characters test case passes along with other test cases.

### DIFF
--- a/config-generators/config-generator.sh
+++ b/config-generators/config-generator.sh
@@ -9,15 +9,19 @@ databaseTypes=();
 # The argument represents the database type. Valid arguments are MsSql, MySql, PostgreSql and Cosmos
 # When invoked with a database type, config file for that database type will be generated.
 # When invoked without any arguments, config files for all the database types will be generated.
+
+allowedDbTypes=("mssql" "mysql" "postgresql" "cosmosdb_nosql" "dwsql")
+databaseTypes=()
+
 if [[ $# -eq 0 ]]; then
-    databaseTypes=("mssql" "mysql" "postgresql" "cosmosdb_nosql")
+    databaseTypes=("${allowedDbTypes[@]}")
 elif [[ $# -eq 1 ]]; then
-    databaseType=$1;
-    if ! { [ $databaseType == "mssql" ] || [ $databaseType == "mysql" ] || [ $databaseType == "postgresql" ] || [ $databaseType == "cosmosdb_nosql" ]; }; then
-        echo "Valid arguments are mssql, mysql, postgresql or cosmosdb_nosql";
+    databaseType=$1
+    if [[ ! ${allowedDbTypes[@]} =~ ${databaseType} ]]; then
+        echo "Valid arguments are mssql, mysql, postgresql, cosmosdb_nosql, or dwsql";
         exit 1;
     fi
-    databaseTypes+=$databaseType;    
+    databaseTypes+=$databaseType;
 else
     echo "Please run with 0 or 1 arguments";
     exit 1;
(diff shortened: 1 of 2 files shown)


src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -109,12 +109,12 @@ private static string GenerateColumnsAsJson(SqlQueryStructure structure, bool su
                 {
                     col_value = $"CAST([{col_value}] AS NVARCHAR(MAX))";
                     // Create json. Example: "book.id": 1 would be a sample output.
-                    stringAgg.Append($"\"{escapedLabel}\":\' + ISNULL(STRING_ESCAPE({col_value},'json'),'null') + \'");
+                    stringAgg.Append($"N\'\"{escapedLabel}\":\' + ISNULL(STRING_ESCAPE({col_value},'json'),'null')");
```
seantleonard, line 112: Is this perf issue applicable to this modification? https://github.com/sequelize/sequelize/issues/3752#issuecomment-334866215

Can you add more details to how you identified `N\'` as a solution?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -123,11 +123,11 @@ private static string GenerateColumnsAsJson(SqlQueryStructure structure, bool su
                 // the below ensures there is a comma after id but not after name.
                 if (i != structure.Columns.Count)
                 {
-                    stringAgg.Append(",");
+                    stringAgg.Append("+\',\'+");
```
seantleonard, line 126: so this is used to escape the single quotes used when appending the char `,` ? can you add more details about the escape sequence usage/ requirement?

seantleonard (changes requested):
need more background on changes

seantleonard:
and a test to validate

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: Im not sure if the N' will cause a whole table scan because its just on the key and not the value right? From my understanding the concern is when you would do it on the result returned by the sql query.

rohkhann:
> and a test to validate

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: This just adds the escape characters for the , that we are adding to the string. We are adding a string character , and hence need to literals.

src/Core/Resolvers/DWSqlQueryBuilder.cs
rohkhann: I confirmed with UC to ensure this would not lead to any perf impacts

seantleonard (approved):

sajeetharan (approved):

config-generators/config-generator.sh
```diff
@@ -9,15 +9,19 @@ databaseTypes=();
 # The argument represents the database type. Valid arguments are MsSql, MySql, PostgreSql and Cosmos
 # When invoked with a database type, config file for that database type will be generated.
 # When invoked without any arguments, config files for all the database types will be generated.
+
+allowedDbTypes=("mssql" "mysql" "postgresql" "cosmosdb_nosql" "dwsql")
```
severussundar, line 13: Could you please make similar changes to `config-generator.ps1` script too?

seantleonard (changes requested):
some changes to get things near passing.

PR #1907, opened 2023-12-01
Fixing dw-sql pipeline and enabling DW Query Builder to handle Unicode characters

severussundar:
- Fixes errors in `dw-sql` pipeline for Linux OS type
- Original PR with the changes: [#1901 ](https://github.com/Azure/data-api-builder/pull/1901) 

# Description from the mentioned PR

## Why make this change?

Dw pipeline test case was failing due to incorrect parsing for chinese characters. 
In addition, dw pipeline was not enabled for linux. 

## What is this change?

Use the N' prefix that handles unicode characters correctly when building the json in dw.
Added the db type in the config-generator.sh

## Tests
Unique characters test case passes along with other test cases.


.pipelines/dwsql-pipelines.yml
```diff
@@ -50,27 +52,43 @@ jobs:
       dockerVersion: 17.09.0-ce
       releaseType: stable
 
+  - task: Bash@3
+    displayName: 'Generate password'
+    inputs:
+      targetType: 'inline'
+      script: |
+        password=$(cat /dev/urandom | tr -dc 'A-Za-z0-9' | head -c24)
```
severussundar, line 60: Using special characters leads to the following error in the `Get and Start Ubuntu SQL Server Image Docker with SSL enabled`

Error: `Detected characters in arguments that may not be executed correctly by the shell. Please escape special characters using backslash (\)`

abhishekkumams (approved):

.pipelines/dwsql-pipelines.yml
seantleonard: Is this all special characters? I only saw documentation that complained about `@` and `$`  (this is because semi colon `;` and spaces ` ` and quotes `'` `"` were not in the regex here.)
https://learn.microsoft.com/en-us/sql/relational-databases/security/strong-passwords?view=sql-server-ver16#:~:text=Starts%20with%20the%20%24%20or%20%40%20character.

.pipelines/dwsql-pipelines.yml
```diff
@@ -50,27 +52,43 @@ jobs:
...
+      arguments: '-p:generateConfigFileForDbType=dwsql --configuration $(buildConfiguration)' # Update this to match your need
 
   - task: FileTransform@1.206.0
     displayName: 'Generate dab-config.DwSql.json'
     inputs:
       folderPath: '$(System.DefaultWorkingDirectory)'
       fileType: 'json'
       targetFiles: 'src/out/tests/net6.0/dab-config.DwSql.json'
-
+
```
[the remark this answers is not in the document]
seantleonard: this suggestion meant to display removing the few spaces in the new line.

seantleonard (approved):

PR #1928, opened 2023-12-14
update mssql pipeline for migration

abhishekkumams:
## Why make this change?

- #1923 
- As part of pipeline migration.

## How was this tested?

- pipeline running fine when triggered by comment


.pipelines/mssql-pipelines.yml
```diff
@@ -49,13 +51,33 @@ jobs:
...
       releaseType: stable
+  
+  - task: PowerShell@2
+    displayName: 'Generate DB password'
+    inputs:
+      targetType: 'inline'
+      script: |
+        $sourceString = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+        $password = -join ($sourceString.ToCharArray() | Get-Random -Count 20)
+        Write-Host "##vso[task.setvariable variable=dbPassword]$password"
```
seantleonard, line 62: can we keep consistent password generation between linux steps in various pipelines? DW introduced this: https://github.com/Azure/data-api-builder/blob/40dab94d85b68f538e336e0b732f8cf73fca026e/.pipelines/dwsql-pipelines.yml#L55-L61

abhishekkumams: updated

seantleonard (approved):
Thanks for making these changes!

severussundar (approved):

PR #1929, opened 2023-12-15
Adding validation to disallow multiple relationships between same source and target entities

ayush3797:
## Why make this change?

Currently, we allow defining multiple relationships between the same source and target entities. A user is allowed to define multiple relationships each with different source fields/target fields and even different cardinalities. This is not only logically incorrect and difficult to make sense out of, but will also pose a problem in the functioning of nested mutations.

## What is this change?
Added the above explained validation in `development` mode.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

### DIFF
--- a/src/Core/Configurations/RuntimeConfigValidator.cs
+++ b/src/Core/Configurations/RuntimeConfigValidator.cs
@@ -651,9 +651,20 @@ public class RuntimeConfigValidator : IConfigValidator
 
             string databaseName = runtimeConfig.GetDataSourceNameFromEntityName(entityName);
             ISqlMetadataProvider sqlMetadataProvider = sqlMetadataProviderFactory.GetMetadataProvider(databaseName);
-
+            Dictionary<string, string> targetEntityNameToRelationshipName = new();
             foreach ((string relationshipName, EntityRelationship relationship) in entity.Relationships!)
             {
+                string targetEntityName = relationship.TargetEntity;
+                if (targetEntityNameToRelationshipName.TryGetValue(targetEntityName, out string? duplicateRelationshipName))
+                {
+                    throw new DataApiBuilderException(
+                        message: $"Found multiple relationships: {duplicateRelationshipName}, {relationshipName} between source entity: {entityName} and target entity: {targetEntityName}",
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
+                }
+
+                // Add entry for this relationship to the dictionary tracking all the relationships for this entity.
+                targetEntityNameToRelationshipName[targetEntityName] = relationshipName;
                 // Validate if entity referenced in relationship is defined in the config.
                 if (!runtimeConfig.Entities.ContainsKey(relationship.TargetEntity))
                 {


src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -333,6 +333,110 @@ public void TestAddingRelationshipWithDisabledGraphQL()
...
+
+            _sqlMetadataProvider.Setup(x => x.EntityToDatabaseObject).Returns(mockDictionaryForEntityDatabaseObject);
+            Mock<IMetadataProviderFactory> _metadataProviderFactory = new();
+            _metadataProviderFactory.Setup(x => x.GetMetadataProvider(It.IsAny<string>())).Returns(_sqlMetadataProvider.Object);
+
+            if (isGQLEnabledForSource)
+            {
+                // Assert for expected exception.
+                DataApiBuilderException ex = Assert.ThrowsException<DataApiBuilderException>(() =>
+                configValidator.ValidateRelationshipsInConfig(runtimeConfig, _metadataProviderFactory.Object));
```
seantleonard, line 430: nit: need more indententation

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -333,6 +333,110 @@ public void TestAddingRelationshipWithDisabledGraphQL()
             Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
         }
 
+        /// <summary>
+        /// Test to validate that we throw a validation error when GraphQL is enabled on the source entity and
+        /// the user defines multiple relationships between the same source and target entities.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(true, DisplayName = "Validate that exception is thrown when GQL is enabled and user defines multiple relationship between source and target entities.")]
+        [DataRow(false, DisplayName = "Validate that no exception is thrown when GQL is disabled and user defines multiple relationship between source and target entities.")]
```
severussundar, line 342: Something to consider: Should we log a warning (instead of throwing an exception which halts engine start-up) when graphQL is disabled for the entity?

severussundar (commented):
I think this particular validation can be added directly to the `main` branch

abhishekkumams:
> I think this particular validation can be added directly to the `main` branch.
> 
> Also, would this validation run as part of `dab validate` command?

It will run both for start and validate command

abhishekkumams (approved):

abhishekkumams:
what is causing multiple relationships between same source and target entities? 
Did we figure out why it was added in the first place?
If it is not required, why not update the method which adds multiple relationship b/w same source-target, instead of throwing an exception if it happens?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: This was agreed upon in the scrum (including Jerry), that we should throw an exception.

ayush3797:
> what is causing multiple relationships between same source and target entities? Did we figure out why it was added in the first place? If it is not required, why not update the method which adds multiple relationship b/w same source-target, instead of throwing an exception if it happens?

This is happening because DAB currently allows users to define multiple relationships between the same source, target entities in the config file.

abhishekkumams:
This issue raised here: #1940 seems connected to your changes. Can you please check if this addresses this issue?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
severussundar: Right. No change in that.

My suggestion was that we could consider throwing a warning (instead of ignoring it) when the graphQL is disabled. Throwing a warning lets users know that something is wrong in the way they have defined the config, but, at the same time, engine does not crash since graphQL is disabled.

ayush3797: But if the user does not want graphql, why should they see an unnecessary warning which should not bother them ideally?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -652,8 +652,24 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+            // are getting more than 1 entry for a target entity, we throw a validation error as it indicates the user has
+            // defined multiple relationships between the same source and target entities.
+            Dictionary<string, string> targetEntityNameToRelationshipName = new();
             foreach ((string relationshipName, EntityRelationship relationship) in entity.Relationships!)
             {
+                string targetEntityName = relationship.TargetEntity;
+                if (targetEntityNameToRelationshipName.TryGetValue(targetEntityName, out string? duplicateRelationshipName))
+                {
+                    throw new DataApiBuilderException(
+                        message: $"Found multiple relationships: {duplicateRelationshipName}, {relationshipName} between source entity: {entityName} and target entity: {targetEntityName}",
```
Aniruddh25, line 665: Perhaps, also mention that this is not supported and is an invalid config? The substatus code is really for the triaging if an issue is opened.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -333,6 +333,110 @@ public void TestAddingRelationshipWithDisabledGraphQL()
...
+                { targetEntityName, targetEntity }
+            };
+
+            RuntimeConfig runtimeConfig = new(
+                Schema: "UnitTestSchema",
+                DataSource: new DataSource(DatabaseType: DatabaseType.MSSQL, "", Options: null),
+                Runtime: new(
+                    Rest: new(),
+                    GraphQL: new(),
+                    Host: new(null, null)
```
Aniruddh25, line 402: how do you ensure `development` mode is picked up? Is that the default today?

Aniruddh25 (approved):
Looks good, 1 question and a nit

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
severussundar: My initial thoughts were this ---> Because defining multiple relationships is an incorrect thing to have in the config file even if graphQL is disabled. Throwing a warning will let them know that the config is an invalid state.  It will be helpful if they have intentions of enabling graphQL in the future.


I guess, when they enable graphQL, they would hit this error and can modify the relationships. So, I don't have a strong preference as to throw/not throw a warning. 
 

Marking this resolved.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
ayush3797: The default is production mode. However, the check for the host mode is done in startup:

image

So, we are unit testing the method for which it does not really matter what is the mode- since the method does not have a dependency on the mode. We can make mode check a part of this method (which I think is something we should do- would make the method more complete and encapsulate all the logic together). Let me know what your thoughts are.

seantleonard (approved):

abhishekkumams (approved):

PR #1931, opened 2023-12-15
fixing datetime and bit issue in DW.

rohkhann:
## Why make this change?

Datawarehouse was having issues with datetime. This is because we are using string agg. if we dont add the "" to datetime, it is being deserialized as a number and we get a bug saying there is a - in the number. FOR JSON converts bit to true or false and hence we need to do same.

## What is this change?

Hence for datetime we need to add the wrapping "" to ensure correct deserialization. 
if datatype is bit set to true or false.

## How was this tested?

unit tests added. 
integration test done with dab running and dw datetime supported. image:

### DIFF
--- a/src/Core/Resolvers/DWSqlQueryBuilder.cs
+++ b/src/Core/Resolvers/DWSqlQueryBuilder.cs
@@ -116,7 +116,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                         // need to wrap datetime in quotes to ensure correct deserialization.
                         stringAgg.Append($"N\'\"{escapedLabel}\":\"\' + ISNULL(STRING_ESCAPE({col_value},'json'),'null') + \'\"\'+");
                     }
-                    else if(col_type == typeof(Boolean))
+                    else if (col_type == typeof(Boolean))
                     {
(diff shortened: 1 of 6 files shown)


src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -108,8 +108,23 @@ private static string GenerateColumnsAsJson(SqlQueryStructure structure, bool su
...
                 {
                     col_value = $"CAST([{col_value}] AS NVARCHAR(MAX))";
-                    // Create json. Example: "book.id": 1 would be a sample output.
-                    stringAgg.Append($"N\'\"{escapedLabel}\":\' + ISNULL(STRING_ESCAPE({col_value},'json'),'null')");
+
+                    Type col_type = structure.GetColumnSystemType(column.ColumnName);
+
+                    if (col_type == typeof(DateTime))
+                    {
+                        // need to wrap datetime in quotes to ensure correct deserialization.
```
aaronburtle, line 116:
```suggestion
                        // Need to wrap datetime in quotes to ensure correct deserialization.
```

aaronburtle (approved):
LGTM!

abhishekkumams (approved):

PR #1943, opened 2023-12-22
Multiple-create: Authorization 

ayush3797:
## Why make this change?
Since GraphQL insertions now support nested insertions, we need to authorize entity and fields not only for the top-level entity in the insertion, but also the nested entities and fields. This PR aims to address that logic of collecting all the unique entities and fields belonging to those entities in a data structure, and then sequentially iterate over all entities and fields to check whether the given role is authorized to perform the action (here nested insertion).

## What is this change?

1. The presence of the model directive on the output type of the mutation object will be used to determine whether we are dealing with a point mutation or a nested insertion. Presence of model directive indicates we are doing a point insertion while its absence indicates the opposite. This logic is added to `SqlMutationEngine.ExecuteAsync()` method. This logic determines whether the input argument name is `item` (for point mutation) or `items` (for insert many).
2. A new method `SqlMutationEngine.AuthorizeEntityAndFieldsForMutation()` is added. The name is kept generic (instead of using 'Insertion') because the same method can be used later for nested updates as well. As the name indicates, this method iterates over all the entities and fields and does the required authorization checks.
3. The method mentioned in the point above depends on another newly added method `SqlMutationEngine.PopulateMutationFieldsToAuthorize()` whose job is to populate all the unique entities referred in the mutation and their corresponding fields into a data structure of the format:
`Dictionary<string, HashSet<string>> entityAndFieldsToAuthorize` - where for each entry in the dictionary:
     -> Key represents the entity name
     -> Value represents the unique set of fields referenced from the entity
4. The method `SqlMutationEngine.PopulateMutationFieldsToAuthorize()` recursively calls itself for nested entities based on different criteria explained in code comments.
5. When the field in a nested mutation is a list of ObjectFieldNode (fieldName: fieldValue) or the field has an object value, the fields are added to the data structure mentioned in (3) using a newly added method: `SqlMutationEngine.ProcessObjectFieldNodesForAuthZ()` which sequentially goes over all the fields and add it to the list of fields to be authorized. Since a field might represent a relationship- and hence a nested entity, this method again calls its parent caller i.e. `SqlMutationEngine.PopulateMutationFieldsToAuthorize()`.
6. The method `SqlMutationEngine.ProcessObjectFieldNodesForAuthZ()` contains the logic to ensure that the fields belonging to linking tables are not added to the list of fields to be authorized.
7. Moved the method `GetRoleOfGraphQLRequest()` from `Cosmos`/`SqlMutationEngine` to `AuthorizationResolver`.
## How was this tested?
To be added.

## Sample Request(s)
1. Config:
image
image

1. Request/Response - AuthZ failure because `piecesAvailable` field is not accessible to `test_role_with_excluded_fields_on_create` role. 
image

2. Request/Response: Removing `piecesAvailable` field from request body leads to successful authz checks (request fails during query generation).

image


src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -441,6 +441,7 @@ internal static List<string> GetSubArgumentNamesFromGQLMutArguments
                     mutationInput = new Dictionary<string, object?>();
                     foreach (ObjectFieldNode node in mutationInputRaw)
                     {
+                        //node.
```
severussundar, line 444: super nit: remove

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -86,20 +87,32 @@ public class SqlMutationEngine : IMutationEngine
...
             ISqlMetadataProvider sqlMetadataProvider = _sqlMetadataProviderFactory.GetMetadataProvider(dataSourceName);
             IQueryEngine queryEngine = _queryEngineFactory.GetQueryEngine(sqlMetadataProvider.GetDatabaseType());
 
             Tuple<JsonDocument?, IMetadata?>? result = null;
             EntityActionOperation mutationOperation = MutationBuilder.DetermineMutationOperationTypeBasedOnInputType(graphqlMutationName);
 
+            string inputArgumentName = isPointMutation ? MutationBuilder.ITEM_INPUT_ARGUMENT_NAME : MutationBuilder.ARRAY_INPUT_ARGUMENT_NAME;
             // If authorization fails, an exception will be thrown and request execution halts.
-            AuthorizeMutationFields(context, parameters, entityName, mutationOperation);
+            AuthorizeMutationFields(inputArgumentName, context, IMutationEngine.GetClientRoleFromMiddlewareContext(context), parameters, entityName, mutationOperation);
```
severussundar, line 115: There is a function in the same class : `GetRoleOfGraphQLRequest` which extracts the role from the hotchocolate context. It can be re-used here

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1073,14 +1085,116 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                        message: $"The client has insufficient permissions on one or more fields in the entity: {entityNameInMutation} referenced in this mutation.",
+                        statusCode: HttpStatusCode.Forbidden,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.AuthorizationCheckFailed
+                    );
+                }
+            }
+            
+            return fieldsToAuthorize;
+        }
+
```
severussundar, line 1129: nit: add method summary and param descriptions

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1073,14 +1085,116 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+        private void ProcessObjectFieldNodesForAuthZ(Dictionary<string, HashSet<string>> fieldsToAuthorize, string entityName, IMiddlewareContext context, RuntimeConfig runtimeConfig, InputObjectType schemaObject, IReadOnlyList<ObjectFieldNode> fieldNodes)
+        {
+            foreach (ObjectFieldNode field in fieldNodes)
+            {
+                Tuple<IValueNode?, SyntaxKind> fieldDetails = GetUnderlyingKindForField(field.Value, context.Variables);
+                SyntaxKind underlyingFieldKind = fieldDetails.Item2;
+                if (underlyingFieldKind != SyntaxKind.ObjectValue && underlyingFieldKind != SyntaxKind.ListValue)
+                {
+                    fieldsToAuthorize[entityName].Add(field.Name.Value);
```
severussundar, line 1166: In case of M:N relationships, the fields from the linking table would also show up as a field in the related entity. Do we require a condition to check whether the field translates to a column in the related entity?

ayush3797: Yep, that logic is yet to be added.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1073,14 +1085,116 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+            object? parameters,
+            RuntimeConfig runtimeConfig)
+        {
+            InputObjectType schemaObject = ResolverMiddleware.InputObjectTypeFromIInputField(schema);
+            if (!fieldsToAuthorize.ContainsKey(entityName))
+            {
+                fieldsToAuthorize.Add(entityName, new HashSet<string>());
+            }
+
+            if (parameters is List<ObjectFieldNode> fields)
```
seantleonard, line 1144: can you elaborate on each of these conditions you are testing? e.g. when do you expect to processing `List<ObjectFieldNode>` versus an `ObjectFieldNode` or `List<IValueNode>`

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1073,14 +1085,116 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+            }
+        }
+
+        private void ProcessObjectFieldNodesForAuthZ(Dictionary<string, HashSet<string>> fieldsToAuthorize, string entityName, IMiddlewareContext context, RuntimeConfig runtimeConfig, InputObjectType schemaObject, IReadOnlyList<ObjectFieldNode> fieldNodes)
+        {
+            foreach (ObjectFieldNode field in fieldNodes)
+            {
+                Tuple<IValueNode?, SyntaxKind> fieldDetails = GetUnderlyingKindForField(field.Value, context.Variables);
+                SyntaxKind underlyingFieldKind = fieldDetails.Item2;
+                if (underlyingFieldKind != SyntaxKind.ObjectValue && underlyingFieldKind != SyntaxKind.ListValue)
```
seantleonard, line 1164: can you elaborate on what it means to meet this condition?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1073,14 +1085,116 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                        schemaObject.Fields[relationshipName],
+                        targetEntityName,
+                        context,
+                        fieldDetails.Item1,
+                        runtimeConfig);
+                }
+            }
+        }
+
+        private Tuple<IValueNode?, SyntaxKind> GetUnderlyingKindForField(IValueNode? value, IVariableValueCollection variables)
```
seantleonard, line 1183: Can you elaborate on what is "kind" and the significance ?

seantleonard: a name more descriptive than `value` would be helpful here.

src/Core/Resolvers/SqlMutationEngine.cs
seantleonard, line 1: Your PR description/design doc/ specification needs to show example requests and how your authorization logic processes such requests. That will help explain the code you've added here.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Moved that method to authz resolver.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Added

src/Core/Resolvers/IMutationEngine.cs
```diff
@@ -41,15 +45,36 @@ public interface IMutationEngine
...
         /// <exception cref="DataApiBuilderException"></exception>
         public void AuthorizeMutationFields(
+            string inputArgumentName,
             IMiddlewareContext context,
+            string clientRole,
             IDictionary<string, object?> parameters,
             string entityName,
             EntityActionOperation mutationOperation);
+
+        protected static string GetClientRoleFromMiddlewareContext(IMiddlewareContext context)
```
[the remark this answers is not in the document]
ayush3797: I agree. Moved that method to authz resolver.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Added example and explaination

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: added

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: added method summary

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1073,14 +1085,116 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+        private void PopulateMutationFieldsToAuthorize(
+            Dictionary<string, HashSet<string>> fieldsToAuthorize,
+            IInputField schema,
+            string entityName,
+            IMiddlewareContext context,
+            object? parameters,
+            RuntimeConfig runtimeConfig)
+        {
+            InputObjectType schemaObject = ResolverMiddleware.InputObjectTypeFromIInputField(schema);
+            if (!fieldsToAuthorize.ContainsKey(entityName))
```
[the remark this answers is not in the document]
ayush3797: Nope, can use TryAdd.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1073,14 +1085,116 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
                         );
             }
 
-            if (!isAuthorized)
+            return isAuthorized;
+        }
+
+        private Dictionary<string, HashSet<string>> AuthorizeEntityAndFieldsForMutation(EntityActionOperation operation, string clientRole, string inputArgumentName, IMiddlewareContext context, IDictionary<string, object?> parametersDictionary, string entityName)
```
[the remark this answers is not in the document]
ayush3797: Added

src/Service.Tests/Authorization/GraphQL/GraphQLMutationAuthorizationTests.cs
```diff
@@ -84,7 +84,9 @@ public void MutationFields_AuthorizationEvaluation(bool isAuthorized, string[] c
             try
             {
                 engine.AuthorizeMutationFields(
+                    MutationBuilder.ITEM_INPUT_ARGUMENT_NAME,
                     graphQLMiddlewareContext.Object,
+                    MIDDLEWARE_CONTEXT_ROLEHEADER_VALUE,
```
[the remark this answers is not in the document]
ayush3797: Can you elaborate more on why this is considered as a breaking change? The additional changes are to support more functionalities while keeping the support of existing functionalities as is.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1048,41 +1069,52 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
         public void AuthorizeMutationFields(
+            string inputArgumentName,
             IMiddlewareContext context,
+            string clientRole,
             IDictionary<string, object?> parameters,
             string entityName,
             EntityActionOperation mutationOperation)
         {
-            string role = GetRoleOfGraphQLRequest(context);
+            if (mutationOperation is EntityActionOperation.Create)
```
Aniruddh25, line 1081: Looks like for create, we call AuthorizeEntityAndFieldsForMutation irrespective of point or multiple mutations, right?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1132,227 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+        }
+
+        /// <summary>
+        /// Performs authorization checks on entity level permissions and field level permissions for every entity and field
+        /// referenced in a GraphQL mutation for the given client role.
+        /// </summary>
+        /// <param name="context">Middleware context.</param>
+        /// <param name="clientRole">Client role header value extracted from the middleware context of the mutation</param>
+        /// <param name="entityName">entity name</param>
+        /// <param name="entityName">Top level entity name.</param>
```
Aniruddh25, line 1145: nit: entityName is specified twice in the params

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1132,227 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+
+            // Perform authorization check at entity level.
+            foreach(string entityNameInMutation in entityNames)
             {
-                throw new DataApiBuilderException(
-                    message: "Unauthorized due to one or more fields in this mutation.",
-                    statusCode: HttpStatusCode.Forbidden,
-                    subStatusCode: DataApiBuilderException.SubStatusCodes.AuthorizationCheckFailed
-                );
+                if (!_authorizationResolver.AreRoleAndOperationDefinedForEntity(entityNameInMutation, clientRole, operation))
```
Aniruddh25, line 1176: Why do we need to separately perform authZ at entity level if we are anyway doing the same (albeit in more detail) at the field level? If the role and operation is not defined for entity, the field level authZ we perform below should also return that the field is unauthorized. In essence, arent the checks we do in AreFieldAuthorizedForEntity a superset of `AreRoleAndOperationDefinedForEntity`?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1132,227 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                }
+            }
+
+            // Perform authorization checks at field level.
+            foreach ((string entityNameInMutation, HashSet<string> fieldsInEntity) in entityAndFieldsToAuthorize)
+            {
+                if (!AreFieldsAuthorizedForEntity(clientRole, entityNameInMutation, EntityActionOperation.Create, fieldsInEntity))
+                {
+                    throw new DataApiBuilderException(
+                        message: $"Access is forbidden to one or more fields in the entity: {entityNameInMutation} referenced in this mutation.",
```
Aniruddh25, line 1192: Why cant we maintain the same exception message as we had previously for point Create mutation? And still have it for update, delete => "Unauthorized due to one or more fields in this mutation."?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1132,227 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+        ///                         title
+        ///                     }
+        ///                 }</example>
+        private void PopulateMutationFieldsToAuthorize(
+            Dictionary<string, HashSet<string>> entityAndFieldsToAuthorize,
+            IInputField schema,
+            string entityName,
+            IMiddlewareContext context,
+            object? parameters,
+            RuntimeConfig runtimeConfig)
```
Aniruddh25, line 1246: why do we need the runtimeConfig argument if we can reference the _runtimeConfigProvider.GetConfig directly inside this method?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1132,227 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+        private void ProcessObjectFieldNodesForAuthZ(
+            IMiddlewareContext context,
+            string entityName,
+            InputObjectType schemaObject,
+            IReadOnlyList<ObjectFieldNode> fieldNodes,
+            Dictionary<string, HashSet<string>> fieldsToAuthorize,
+            RuntimeConfig runtimeConfig)
+        {
+            fieldsToAuthorize.TryAdd(entityName, new HashSet<string>());
+            string dataSourceName = GraphQLUtils.GetDataSourceNameFromGraphQLContext(context, runtimeConfig);
```
Aniruddh25, line 1294: will nested insertion work for multiple config files as well?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1132,227 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+            }
+        }
+
+        /// <summary>
+        /// Helper method to get the field details i.e. the field value and the field kind, from the GraphQL mutation body.
+        /// If the field value is being provided as a variable in the mutation, a recursive call is made to the method
+        /// to get the actual value of the variable.
+        /// </summary>
+        /// <param name="value">Value of the field.</param>
+        /// <param name="variables">Colelction of variables declared in the GraphQL mutation request.</param>
```
Aniruddh25, line 1339:
```suggestion
        /// <param name="variables">Collection of variables declared in the GraphQL mutation request.</param>
```

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1132,227 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                if (underlyingFieldKind is not SyntaxKind.ObjectValue && underlyingFieldKind is not SyntaxKind.ListValue)
+                {
+                    // It might be the case that we are currently processing the fields for a linking input object.
+                    // Linking input objects enable users to provide input for fields belonging to the target entity and the linking entity.
+                    // Hence the backing column for fields belonging to the linking entity will not be present in the source definition of this target entity.
+                    // We need to skip such fields belonging to linking table as we do not perform authorization checks on them.
+                    if (metadataProvider.TryGetBackingColumn(entityName, field.Name.Value, out string? _))
+                    {
+                        // Only add those fields to this entity's set of fields which belong to this entity and not the linking entity.
+                        fieldsToAuthorize[entityName].Add(field.Name.Value);
```
Aniruddh25, line 1313: do we expect the fields we want to authorize to be mapped names or underlying table's column names?

Aniruddh25 (changes requested):
Overall looks good, some ques and nits

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: yes

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Right, I just looked at the implementation of the AuthorizationResolver.AreColumnsAllowedForOperation(). It has checks for whether the role and action is defined for an entity. So, yes we don't need the `AreRoleAndOperationDefinedForEntity` check.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: right, we should be able to do it.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: I think so, yes. However, nested insertion would be possible for entities specified in the same config since in the `relationships` section of the config file, a relationship can only be defined with an entity present in the same config.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: We expect them to be mapped names. The `AuthorizationResolver.AreColumnsAllowedForOperation()` takes care of getting the backing column names:

image

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -83,18 +86,36 @@ public class SqlMutationEngine : IMutationEngine
 
             dataSourceName = GetValidatedDataSourceName(dataSourceName);
             string graphqlMutationName = context.Selection.Field.Name.Value;
-            string entityName = GraphQLUtils.GetEntityNameFromContext(context);
+            IOutputType outputType = context.Selection.Field.Type;
+            string entityName = outputType.TypeName();
```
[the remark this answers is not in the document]
ayush3797: added comments and moved the entire logic to a separate method

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: we can. I'll incorporate this.

src/Core/Resolvers/SqlMutationEngine.cs
Aniruddh25: We should add a test for that - attempt to do nested insert between entities across different configs leads to failure

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1149,199 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
-            if (string.IsNullOrEmpty(role))
+            else if (parameters is List<IValueNode> listOfIValueNode)
             {
-                throw new DataApiBuilderException(
-                    message: "No ClientRoleHeader available to perform authorization.",
-                    statusCode: HttpStatusCode.Unauthorized,
-                    subStatusCode: DataApiBuilderException.SubStatusCodes.AuthorizationCheckFailed);
+                // For the example createbooks mutation written above, the list value for `items` is interpreted as a List<IValueNode>
+                // i.e. items is a list of ObjectValueNode(s).
+                listOfIValueNode.ForEach(iValueNode => PopulateMutationEntityAndFieldsToAuthorize(
```
Aniruddh25, line 1262: What is the schemaObject for this `if` condition evaluate to? where parameters is List<IValueNode>... Do we really need the schemaObject for this if and the last else if?

Aniruddh25 (commented):
Some more questions on latest

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: > What is the schemaObject for this if condition evaluate to? 

This evaluates to the input type of `items` argument without the array modifier, i.e. for createBooks, it will evaluate to `createBookInput` type.

> Do we really need the schemaObject for this if and the last else if?

No, I'll move this evaluation to where it is required.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -218,6 +218,44 @@ public class SqlMutationEngine : IMutationEngine
...
+            bool isPointMutation;
+            if (GraphQLUtils.TryExtractGraphQLFieldModelName(_underlyingFieldType.Directives, out string? modelName))
+            {
+                isPointMutation = true;
+                entityName = modelName;
+            }
+            else
+            {
+                // Absence of model directive on the mutation indicates that we are dealing with a `mutate many`
+                // mutation like createBooks.
```
severussundar, line 242: Model directive is not added on `*Connection` object types (which is the input type for multiple create operation type). Hence, the conclusion.

It would be helpful to add more details (wherever applicable) on why the failure to meet a certain condition leads to a certain conclusion.

Aniruddh25 (changes requested):
Waiting for a unit test on a successful authZ check (even though it might fail further in query execution)

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                SyntaxKind underlyingFieldKind = fieldDetails.Item2;
+
+                // For a column field, we do not have to recurse to process fields in the value - which is required for relationship fields.
+                if (GraphQLUtils.IsScalarField(underlyingFieldKind))
+                {
+                    // It might be the case that we are currently processing the fields for a linking input object.
+                    // Linking input objects enable users to provide input for fields belonging to the target entity and the linking entity.
+                    // Hence the backing column for fields belonging to the linking entity will not be present in the source definition of this target entity.
+                    // We need to skip such fields belonging to linking table as we do not perform authorization checks on them.
+                    if (metadataProvider.TryGetBackingColumn(entityName, field.Name.Value, out string? _))
```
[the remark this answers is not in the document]
ayush3797: Yes, it contains the names of the exposed fields. I'll add a comment in the param summary for this variable. I'd rename the variable name as well.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                    string relationshipName = field.Name.Value;
+                    string targetEntityName = runtimeConfig.Entities![entityName].Relationships![relationshipName].TargetEntity;
+
+                    // Recurse to process fields in the value of this relationship field.
+                    PopulateMutationEntityAndFieldsToAuthorize(
+                        entityTofields,
+                        schemaObject.Fields[relationshipName],
+                        targetEntityName,
+                        context,
+                        fieldDetails.Item1);
```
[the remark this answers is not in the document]
ayush3797: The fieldDetails.Item1 represent the value of a column field. And its perfectly ok to assign a null value to a column which is nullable (not to mention, the column should not be a relationship field - which is captured during request validation). Hence, yes, it needs to be nullable.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+        ///                         id
+        ///                         title
+        ///                     }
+        ///                 }</example>
+        private void PopulateMutationEntityAndFieldsToAuthorize(
+            Dictionary<string, HashSet<string>> entityToFields,
+            IInputField schema,
+            string entityName,
+            IMiddlewareContext context,
+            object? parameters)
```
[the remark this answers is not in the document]
ayush3797: I think, it should be fine to keep it non-nullable. I'd need to add another conditional for null values which I think would be more clear rather than making this param nullable.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Makes sense. I'll add the comment.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: I think this validation, and the corresponding test should be added to `RuntimeConfigValidator` stage itself, where it prohibits the user to add relationship between entities across different config files. With that in place, the generated schema won't even contain such an invalid relationship. I had a chat with @rohkhann some weeks back, and he confirmed that this is a validation we need to have in place.

Aniruddh25: Please create a separate issue for adding this cross config file validation + test.

rohkhann: yes this makes sense, since we dont support cross database relationships.

seantleonard:
Can you elaborate on the expected behavior of what happens when authorization requirements are met for some, but not all of the input? e.g. A single component of the input failing authorization fails the entire request?

in your example in PR description of inserting multiple review objects: `reviews [ {title: "mytitle"}, {body: "reviewBody" ] ` if "body" column is permitted, but "title" is not, then the whole request fails?

src/Core/Resolvers/IMutationEngine.cs
```diff
@@ -41,13 +42,15 @@ public interface IMutationEngine
...
         /// Authorization check on mutation fields provided in a GraphQL Mutation request.
         /// </summary>
-        /// <param name="context">Middleware context of the mutation</param>
+        /// <param name="clientRole">Client role header value extracted from the middleware context of the mutation</param>
         /// <param name="parameters">parameters in the mutation query.</param>
         /// <param name="entityName">entity name</param>
         /// <param name="mutationOperation">mutation operation</param>
         /// <exception cref="DataApiBuilderException"></exception>
         public void AuthorizeMutationFields(
+            string inputArgumentName,
```
seantleonard, line 51: need comment for param and for param comments to be in correct order.

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -94,17 +93,17 @@ private async Task<JObject> ExecuteAsync(IMiddlewareContext context, IDictionary
...
         /// <inheritdoc/>
         public void AuthorizeMutationFields(
+            string inputArgumentName,
             IMiddlewareContext context,
+            string clientRole,
             IDictionary<string, object?> parameters,
             string entityName,
             EntityActionOperation mutationOperation)
         {
-            string role = GetRoleOfGraphQLRequest(context);
```
seantleonard, line 102: why remove this and then require it to be sent as an argument?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -218,6 +218,44 @@ public class SqlMutationEngine : IMutationEngine
...
+            if (GraphQLUtils.TryExtractGraphQLFieldModelName(_underlyingFieldType.Directives, out string? modelName))
+            {
+                isPointMutation = true;
+                entityName = modelName;
+            }
+            else
+            {
+                // Absence of model directive on the mutation indicates that we are dealing with a `mutate many`
+                // mutation like createBooks.
+                isPointMutation = false;
```
seantleonard, line 243: `isPointMutation` is false by default, so this assignment is redundant.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -218,6 +218,44 @@ public class SqlMutationEngine : IMutationEngine
...
+
+                // For a mutation like createBooks which inserts multiple records into the Book entity, the underlying field type is a paginated response
+                // type like 'BookConnection'. To determine the underlying entity name, we have to look at the type of the `items` field which stores a list of items of
+                // the underlying entity's type - here, Book type.
+                IOutputType entityOutputType = _underlyingFieldType.Fields[MutationBuilder.ARRAY_INPUT_ARGUMENT_NAME].Type;
+                ObjectType _underlyingEntityFieldType = GraphQLUtils.UnderlyingGraphQLEntityType(entityOutputType);
```
seantleonard, line 249: why prefix with an underscore?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -218,6 +218,44 @@ public class SqlMutationEngine : IMutationEngine
...
+                if (GraphQLUtils.TryExtractGraphQLFieldModelName(_underlyingEntityFieldType.Directives, out modelName))
```
seantleonard, line 250: Would we benefit from an else statement here to throw an exception in case any scenarios exist where this condition doesn't hold true?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1048,41 +1086,53 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
             EntityActionOperation mutationOperation)
         {
-            string role = GetRoleOfGraphQLRequest(context);
-
-            List<string> inputArgumentKeys;
-            if (mutationOperation != EntityActionOperation.Delete)
+            if (mutationOperation is EntityActionOperation.Create)
             {
-                inputArgumentKeys = BaseSqlQueryStructure.GetSubArgumentNamesFromGQLMutArguments(MutationBuilder.ITEM_INPUT_ARGUMENT_NAME, parameters);
+                AuthorizeEntityAndFieldsForMutation(context, clientRole, entityName, mutationOperation, inputArgumentName, parameters);
```
seantleonard, line 1100: This method's name: "AuthorizeMutationFields" implies that Entity authorization already occurred. Seems odd that now you are expanding the scope of this method to now call `AuthorizeEntityAndFieldsForMutation(...)`

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+            Dictionary<string, HashSet<string>> entityToFields = new();
+            object? parameters;
+            if (parametersDictionary.TryGetValue(inputArgumentName, out parameters))
+            {
+                // Get all the entity names and field names referenced in the mutation.
+                PopulateMutationEntityAndFieldsToAuthorize(entityToFields, schemaForArgument, entityName, context, parameters);
+            }
+
+            // Perform authorization checks at field level.
+            foreach ((string entityNameInMutation, HashSet<string> fieldsInEntity) in entityToFields)
```
seantleonard, line 1188: Seems like there's a case where `entityToFields` can be empty? e.g. when `if (parametersDictionary.TryGetValue(inputArgumentName, out parameters))` returns false?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+            IMiddlewareContext context,
+            object? parameters)
         {
-            string role = string.Empty;
-            if (context.ContextData.TryGetValue(key: AuthorizationResolver.CLIENT_ROLE_HEADER, out object? value) && value is StringValues stringVals)
+            if (parameters is List<ObjectFieldNode> listOfObjectFieldNode)
             {
-                role = stringVals.ToString();
+                // For the example createbook mutation written above, the object value for `item` is interpreted as a List<ObjectFieldNode> i.e.
+                // all the fields present for item namely- title, reviews, publisher, authors are interpreted as ObjectFieldNode.
```
seantleonard, line 1250: `title` is resolved as an `ObjectFieldNode` and not an `IValueNode`? title isn't an object, it's a scalar field.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                // For the example createbook mutation written above, the node for publisher field is interpreted as an ObjectValueNode.
+                // Similarly the individual node (elements in the list) for the reviews, authors ListValueNode(s) are also interpreted as ObjectValueNode(s).
+                ProcessObjectFieldNodesForAuthZ(
+                    context: context,
+                    entityName: entityName,
+                    schemaObject: ResolverMiddleware.InputObjectTypeFromIInputField(schema),
+                    fieldNodes: objectValueNode.Fields,
+                    entityTofields: entityToFields);
+            }
+            else if (parameters is ListValueNode listValueNode)
```
seantleonard, line 1280: since this last condition is `else if` and not `else` do you know of other cases that would be hit here resulting in `entityToFields` remaining empty?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                {
+                    // It might be the case that we are currently processing the fields for a linking input object.
+                    // Linking input objects enable users to provide input for fields belonging to the target entity and the linking entity.
+                    // Hence the backing column for fields belonging to the linking entity will not be present in the source definition of this target entity.
+                    // We need to skip such fields belonging to linking table as we do not perform authorization checks on them.
+                    if (metadataProvider.TryGetBackingColumn(entityName, field.Name.Value, out string? _))
+                    {
+                        // Only add those fields to this entity's set of fields which belong to this entity and not the linking entity.
+                        entityTofields[entityName].Add(field.Name.Value);
+                    }
```
seantleonard, line 1329: nit: could this be a private helper named such that it self-describes the condition? 
`IsNonLinkingTableField(string entityName, string fieldName)`

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                }
+                else
+                {
```
seantleonard, line 1332: this else condition indicates field is a relationship field?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1100,37 +1150,198 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                    schema: schema,
+                    entityName: entityName,
+                    context: context,
+                    parameters: iValueNode));
+            }
+            else if (parameters is ObjectValueNode objectValueNode)
+            {
+                // For the example createbook mutation written above, the node for publisher field is interpreted as an ObjectValueNode.
+                // Similarly the individual node (elements in the list) for the reviews, authors ListValueNode(s) are also interpreted as ObjectValueNode(s).
+                ProcessObjectFieldNodesForAuthZ(
```
seantleonard, line 1273: super nit: any way that this method `ProcessObjectFieldNodesForAuthZ(...)` can have the same param order as `PopulateMutationEntityAndFieldsToAuthorize(...)`

```
context: 
entityName:
schemaObject:
fieldNodes:
entityTofields: 
```

vs 

```
entityToFields: ,
schema: ,
entityName: ,
context: 
parameters: )
```

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -375,6 +376,43 @@ private static string GenerateDataSourceNameKeyFromPath(IMiddlewareContext conte
...
+        /// <summary>
+        /// Helper method to determine whether a field is a column or complex (relationship) field based on its syntax kind.
+        /// If the SyntaxKind for the field is not ObjectValue and ListValue, it implies we are dealing with a column/scalar field which
+        /// has an IntValue, FloatValue, StringValue, BooleanValue, NullValue or an EnumValue.
+        /// </summary>
+        /// <param name="fieldSyntaxKind">SyntaxKind of the field.</param>
+        /// <returns>true if the field is a scalar field, else false.</returns>
+        public static bool IsScalarField(SyntaxKind fieldSyntaxKind)
+        {
+            return fieldSyntaxKind is not SyntaxKind.ObjectValue && fieldSyntaxKind is not SyntaxKind.ListValue;
```
seantleonard, line 388: there's no syntaxKind or object property to condition on the affirmative? no property on `IValueNode` ?

seantleonard (commented):
more questions, still need to look at tests file again

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Created issue here: https://github.com/Azure/data-api-builder/issues/2078.

src/Core/Resolvers/CosmosMutationEngine.cs
ayush3797: The caller methods already had role evaluated, so I thought why evaluate again inside this method. But I think this is just creating an unnecessary dependency, since the role can be evaluated from context which is being passed anyways. I'll revert it back to previous state.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: yeah should not be.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: yes

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: I think it's a bit of overkill. It's very specific to this particular code block and covers two different cases (please see the comment).

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Yes, title is resolved as an `ObjectFieldNode` which represents a key-value pair where the value is a scalar value. IValueNode is the interface for `ObjectValueNode` and `ListValueNode`. For `ObjectValueNode`, the value in KV pair is an object and not a scalar value. Similarly for `ListValueNode` , the value is a list value/array value. 

Please don't confuse it with the scalar/complex field we have been referring to in the schema generation portion.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: No, the schema is generated such that parameters would never be null. I'll anyways add an else block to throw an exception.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Renaming the method `AuthorizeMutationFields` -> `AuthorizeMutation`.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: I don't think this condition would ever not be true.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: I don't think there are any other cases. I'll change it to else condition.

ayush3797:
> Can you elaborate on the expected behavior of what happens when authorization requirements are met for some, but not all of the input? e.g. A single component of the input failing authorization fails the entire request?
> 
> in your example in PR description of inserting multiple review objects: `reviews [ {title: "mytitle"}, {body: "reviewBody" ] ` if "body" column is permitted, but "title" is not, then the whole request fails?

Whether the role lacks the permission for `create` operation on any entity or any field belonging to any entity referenced in the mutation, AuthZ check will fail. 

So, to your question, the answer is yes, the whole request will fail.

src/Service.GraphQLBuilder/GraphQLUtils.cs
ayush3797: Whatever is not a list or an object is a scalar. I have changed the function implementation to rather use affirmatives.

On a side note, there are 8 possible SyntaxKind enum values which represent a GQL node value:

1. IntValue,
2. FloatValue,
3. StringValue,
4. BooleanValue,
5. EnumValue,
6. NullValue,
7. ListValue,
8. ObjectValue,

The last 2 represent a complex input field (relationship field).

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -229,25 +229,27 @@ public class SqlMutationEngine : IMutationEngine
...
-                IOutputType entityOutputType = _underlyingFieldType.Fields[MutationBuilder.ARRAY_INPUT_ARGUMENT_NAME].Type;
-                ObjectType _underlyingEntityFieldType = GraphQLUtils.UnderlyingGraphQLEntityType(entityOutputType);
-                if (GraphQLUtils.TryExtractGraphQLFieldModelName(_underlyingEntityFieldType.Directives, out modelName))
+                // For a mutation like createBooks which inserts multiple records into the Book entity,
+                // the underlying field type is a paginated response type like 'BookConnection'.
+                // To determine the underlying entity name, we have to look at the type of the `items` field
+                // which stores a list of items of the underlying entity's type - here, Book type.
+                IOutputType entityOutputType = underlyingFieldType.Fields[MutationBuilder.ARRAY_INPUT_ARGUMENT_NAME].Type;
+                ObjectType underlyingEntityFieldType = GraphQLUtils.UnderlyingGraphQLEntityType(entityOutputType);
+                if (GraphQLUtils.TryExtractGraphQLFieldModelName(underlyingEntityFieldType.Directives, out modelName))
```
seantleonard, line 252: A few lines up you say
```csharp
// Model directive is not added to the output type of 'mutate many' mutations.
// Thus, absence of model directive here indicates that we are dealing with a 'mutate many'
```
and here you try to extract model name. Is that only possible here because you've retrieved the underlying type from the paginated reponse type?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1316,16 +1333,22 @@ private bool AreFieldsAuthorizedForEntity(string clientRole, string entityName,
...
                     // We need to skip such fields belonging to linking table as we do not perform authorization checks on them.
+                    //
+                    // Case 3. When a relationship field is assigned a null value. Such a field also needs to be ignored.
                     if (metadataProvider.TryGetBackingColumn(entityName, field.Name.Value, out string? _))
                     {
-                        // Only add those fields to this entity's set of fields which belong to this entity and not the linking entity.
-                        entityTofields[entityName].Add(field.Name.Value);
+                        // Only add those fields to this entity's set of fields which belong to this entity and not the linking entity,
+                        // i.e. for Case 1.
+                        entityToExposedColumns[entityName].Add(field.Name.Value);
```
seantleonard, line 1351: tryadd?

src/Service.Tests/Authorization/GraphQL/NestedMutationAuthorizationUnitTests.cs
seantleonard, line 1: are you adding more tests to this file? 
such as field level authorization

seantleonard (commented):
looks good so far. Anticipating review of additional test cases.

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -387,6 +388,45 @@ private static string GenerateDataSourceNameKeyFromPath(IMiddlewareContext conte
...
+        /// Helper method to determine whether a field is a column or complex (relationship) field based on its syntax kind.
+        /// If the SyntaxKind for the field is not ObjectValue and ListValue, it implies we are dealing with a column/scalar field which
+        /// has an IntValue, FloatValue, StringValue, BooleanValue or an EnumValue.
+        /// </summary>
+        /// <param name="fieldSyntaxKind">SyntaxKind of the field.</param>
+        /// <returns>true if the field is a scalar field, else false.</returns>
+        public static bool IsScalarField(SyntaxKind fieldSyntaxKind)
+        {
+            return fieldSyntaxKind is SyntaxKind.IntValue || fieldSyntaxKind is SyntaxKind.FloatValue ||
+                fieldSyntaxKind is SyntaxKind.StringValue || fieldSyntaxKind is SyntaxKind.BooleanValue ||
```
Aniruddh25, line 401: Should we check for `SyntaxKind.NullValue` here to call it a scalar field?

Aniruddh25 (approved):
Looks close to finish line as well, except for +ve authorization successful test case

src/Service.GraphQLBuilder/GraphQLUtils.cs
ayush3797: Nope, NullValue can also be for a relationship field since relationship fields are optional.

src/Service.Tests/Authorization/GraphQL/NestedMutationAuthorizationUnitTests.cs
ayush3797: yes, adding them.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: This is a hashset, so we don't have a TryAdd, and we don't need one. Can safely use Add.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Yes.

src/Service.Tests/Authorization/GraphQL/NestedMutationAuthorizationUnitTests.cs
```diff
@@ -0,0 +1,71 @@
...
+            string nestedCreateOneBook = @"mutation {
+                    createbook(item: { title: ""My New Book"", publishers: { name: ""New publisher""}}) {
+                        id
+                        title
+                    }
+                }";
+
+            // The anonymous role has create permissions on the Book entity but not on the Publisher entity.
+            // Hence the request will fail during authorization check.
+            await ValidateRequestIsUnauthorized(createBookMutationName, nestedCreateOneBook);
```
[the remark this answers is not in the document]
ayush3797: Added tests, will let Shyam know.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Will add examples to the doc and PR.

src/Core/Resolvers/IMutationEngine.cs
```diff
@@ -41,15 +45,36 @@ public interface IMutationEngine
...
             IMiddlewareContext context,
+            string clientRole,
```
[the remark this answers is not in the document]
ayush3797: stale comment

seantleonard:
The PR description example doesn't touch on authorization of "multiple mutations." Yes, there is a relationship field in the mutation, but your example doesn't demonstrate authorization of fields/entities defined by that relationship field.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -441,6 +441,7 @@ internal static List<string> GetSubArgumentNamesFromGQLMutArguments
                     Dictionary<string, object?> mutationInput = new();
                     foreach (ObjectFieldNode node in mutationInputRaw)
                     {
+                        //node.
```
seantleonard, line 444:
```suggestion
```

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -1101,37 +1145,219 @@ private void PopulateParamsFromRestRequest(Dictionary<string, object?> parameter
...
+                        );
+                }
+
+                // Perform authorization checks at field level.
+                foreach ((string entityNameInMutation, HashSet<string> exposedColumnsInEntity) in entityToExposedColumns)
+                {
+                    if (!AreFieldsAuthorizedForEntity(clientRole, entityNameInMutation, operation, exposedColumnsInEntity))
+                    {
+                        throw new DataApiBuilderException(
+                            message: $"Unauthorized due to one or more fields in this mutation.",
```
seantleonard, line 1195: nit: would it make sense to include `entityNameInMutation` here to help user triangulate which part of the mutation is at fault?
Current response looks like:
```json
{
  "errors": [
    {
      "message": "Unauthorized due to one or more fields in this mutation.",
      "extensions": {
        "message": "Unauthorized due to one or more fields in this mutation.",
        "stackTrace": "<stack_trace>",
        "code": "AuthorizationCheckFailed"
      }
    }
  ],
  "data": {
    "createbooks": null
  }
}
```

for the mutation executing as `Anonymous`
```graphql
mutation authz2Test {
  createbooks(items:[
    {title:"magicTreehouse3", publishers:{name: "RandomApartmentPubHouse"}}
    ])
  {
    items{
      id,
      title,
      publisher_id
    }
  }
}
```

src/Service.Tests/Authorization/GraphQL/CreateMutationAuthorizationTests.cs
```diff
@@ -0,0 +1,390 @@
...
+                isAuthenticated: false,
+                clientRoleHeader: "anonymous"
+                );
+        }
+
+        /// <summary>
+        /// Test to validate that a 'create one' point mutation will fail the AuthZ checks if the user does not have create permission
+        /// on one more columns belonging to the entity in the mutation.
+        /// </summary>
+        /// <returns></returns>
```
seantleonard, line 58:
```suggestion
```

src/Service.Tests/Authorization/GraphQL/CreateMutationAuthorizationTests.cs
```diff
@@ -0,0 +1,390 @@
...
+            await ValidateRequestIsUnauthorized(
+                graphQLMutationName: createBookMutationName,
+                graphQLMutation: createOneBookMutation,
+                expectedExceptionMessage: "Unauthorized due to one or more fields in this mutation.",
+                expectedExceptionStatusCode: DataApiBuilderException.SubStatusCodes.AuthorizationCheckFailed.ToString(),
+                isAuthenticated: false,
+                clientRoleHeader: "anonymous"
+                );
+
+            // The authenticates role has create permissions on both the Book and Publisher entities.
```
seantleonard, line 119:
```suggestion
            // The authenticated role has create permissions on both the Book and Publisher entities.
```

src/Service.Tests/Authorization/GraphQL/CreateMutationAuthorizationTests.cs
```diff
@@ -0,0 +1,390 @@
...
+            string createMultipleBookMutation = @"mutation {
+                    createbooks(items: [{ title: ""Book #1"", publishers: { name: ""Publisher #1""}},
+                                        { title: ""Book #2"", publishers: { name: ""Publisher #2""}}]) {
+                        items{
+                            id
+                            title
+                        }
+                    }
+                }";
+
```
seantleonard, line 146: another test to add would be to ensure that the whole request fails if even one of the inserts is forbidden:
```graphql
mutation {
                    createbooks(items: [{ title: ""Book #1"", publisher_id: 1234},
                                        { title: ""Book #2"", publishers: { name: ""Publisher #2""}}]) {
                        items{
                            id
                            title
                        }
                    }
                }";
```

src/Service.Tests/Authorization/GraphQL/CreateMutationAuthorizationTests.cs
```diff
@@ -0,0 +1,390 @@
...
+                                                items
+                                                {
+                                                   categoryid
+                                                   pieceid
+                                                }
+                                             }
+                                          }";
+
+            // The 'test_role_with_excluded_fields_on_create' role does not have create permissions on
+            // stocks.piecesAvailable field and hence the authorization check should fail.
```
seantleonard, line 278: Seems that all the tests in this file check for permissions on either:
- top level entity
- top level entity fields
- sub-entity

I don't see tests that cover permissions check on 
- sub-entity fields. (which would validate behavior of recursive authZ calls when processing the relationship `stocks_price` , or `publishers` in other tests)

seantleonard (approved):
Looks good so far. I'd say only thing that this PR would benefit from is testing authz of sub-entity fields. Unless i missed it, i only saw tests for Top-level entity and fields and sub-entity (no fields)

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -84,17 +85,18 @@ public class SqlMutationEngine : IMutationEngine
...
             Tuple<JsonDocument?, IMetadata?>? result = null;
             EntityActionOperation mutationOperation = MutationBuilder.DetermineMutationOperationTypeBasedOnInputType(graphqlMutationName);
+            string roleName = AuthorizationResolver.GetRoleOfGraphQLRequest(context);
+            string inputArgumentName = isPointMutation ? MutationBuilder.ITEM_INPUT_ARGUMENT_NAME : MutationBuilder.ARRAY_INPUT_ARGUMENT_NAME;
 
             // If authorization fails, an exception will be thrown and request execution halts.
-            AuthorizeMutationFields(context, parameters, entityName, mutationOperation);
-
-            string roleName = GetRoleOfGraphQLRequest(context);
+            AuthorizeMutation(inputArgumentName, context, parameters, entityName, mutationOperation);
```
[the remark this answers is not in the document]
seantleonard: If you think this optimization makes sense, open an issue to track and then we can revisit. Don't need to make more changes to how this PR works.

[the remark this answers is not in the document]
ayush3797: Most of the implementation is being continued to what it was and is common with cosmos. Minor optimization for evaluating inputArgumentName makes sense and should be quick. I'll incorporate it.

src/Service.Tests/Authorization/GraphQL/CreateMutationAuthorizationTests.cs
ayush3797: added

src/Service.Tests/Authorization/GraphQL/CreateMutationAuthorizationTests.cs
ayush3797: The test that you're suggesting would cover this test case as well. Added that.

src/Core/Resolvers/SqlMutationEngine.cs
ayush3797: Will take this up as another issue: https://github.com/Azure/data-api-builder/issues/2123. Discussed over chat - fine to mark the comment as resolved for now.

PR #1952, opened 2024-01-04
Add support for app insights in CLI

### LINKED ISSUE #1745: [RFC]: Adding Cli option to enable Application Insights telemtry
Dab engine supports config with the below telemetry options.
```
{
  "runtime": {
    "telemetry": {
      "application-insights": {
        "enabled": true,
        "connection-string": "{YOUR_APP_INSIGHTS_CONN_STRING}"
      }
  }
}
```

Below is the proposed cli command to add application insights telemetry to the config
```
dab add-telemetry --app-insights-enabled true --app-insights-conn-string "connection string"
```

The `add-telemetry` command is used to add telemetry support to a Data API Builder project. This command will modify the runtime config file to add application Insights telemetry and configure the connection string.

Flags/Options
`--app-insights-enabled`: A boolean flag that specifies whether Application Insights telemetry should be enabled. This flag is required and must be set to either true or false.
`--app-insights-conn-string`: A string that specifies the connection string for the Application Insights resource to which telemetry data should be sent. This flag is required whenever `--app-insights-enabled` is set to true and must be set to a valid connection string.

Here's an example usage of the dab add-telemetry command:
```
dab add-telemetry --app-insights-enabled true --app-insights-conn-string "InstrumentationKey=12345678-1234-1234-1234-123456789012,IngestionEndpoint=<<ingestion_url>>"
```

This command will enable Application Insights telemetry and configure the connection string for the specified resource. Note that the connection string should be replaced with the actual connection string for your Application Insights resource.

abhishekkumams:
## Why make this change?

- Closes #1745 
  - This change will allow users to add application insights to config through CLI

## What is this change?

- Added a new cli command `add-telemetry` to add telemetry resource where the data will be sent. Currently DAB supports sending telemetry to Application Insights.
- this command has two flags/options:
`--app-insights-enabled`: A boolean flag that specifies whether Application Insights telemetry should be enabled. This flag is optional and default value is true.
`--app-insights-conn-string`: A string that specifies the connection string for the Application Insights resource to which telemetry data should be sent. This flag is required must be set to a valid connection string.

- This command will add application insights telemetry resource details in the config, when dab engine starts with this config it will send telemetry details to applications insights.

## How was this tested?
- [X] Unit Tests


src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+using CommandLine;
+using Microsoft.Extensions.Logging;
+using static Cli.Utils;
+
+namespace Cli.Commands
+{
+    /// <summary>
+    /// Telemetry command options
+    /// </summary>
+    [Verb("add-telemetry", isDefault: false, HelpText = "Add telemtry for Data Api Builder Application", Hidden = false)]
```
Aniruddh25, line 16:
```suggestion
    [Verb("add-telemetry", isDefault: false, HelpText = "Add telemtry for Data Api builder Application", Hidden = false)]
```

src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+        public AddTelemetryOptions(bool appInsightsEnabled, string appInsightsConnString, string? config) : base(config)
+        {
+            AppInsightsEnabled = appInsightsEnabled;
+            AppInsightsConnString = appInsightsConnString;
+        }
+
+        [Option("app-insights-enabled", Default = false, HelpText = "Enable/Disable Application Insights")]
+        public bool AppInsightsEnabled { get; }
+
+        [Option("app-insights-conn-string", HelpText = "Connection string for the Application Insights resource for telemetry data", Required = true)]
```
Aniruddh25, line 28: If app-insights-conn-string is provided, wouldn't it mean the user wants to enable appinsights - so in that case why would they need to additionally specify enabled = true, isnt simply specifying the conn-string sufficient to infer?

src/Cli/ConfigGenerator.cs
```diff
@@ -1317,5 +1317,51 @@ private static EntityGraphQLOptions ConstructUpdatedGraphQLDetails(Entity entity
...
+            }
+
+            ApplicationInsightsOptions applicationInsightsOptions = new(
+                Enabled: options.AppInsightsEnabled,
+                ConnectionString: options.AppInsightsConnString
+            );
+
+            runtimeConfig = runtimeConfig with
+            {
+                Runtime = runtimeConfig.Runtime with
```
Aniruddh25, line 1358: if runtimeConfig.Runtime already has telemetry section, will this override it?
What if both telemetry/appinsights sections are present? 
What if only telemetry section is present?
What if telemetry + other future option than application insights is present in the read config file?

src/Cli.Tests/AddTelemetryTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        /// <summary>
+        /// Testing to check telemetry options are correctly added to the config.
+        /// Verifying scenarios such as enabling/disabling telemetry and providing a valid/empty connection string.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(true, "", false, DisplayName = "Add Telemetry with empty connection string should fail.")]
+        [DataRow(true, "InstrumentationKey=00000000-0000-0000-0000-000000000000", true, DisplayName = "Add Telemetry with valid connection string")]
+        [DataRow(false, "InstrumentationKey=00000000-0000-0000-0000-000000000000", true, DisplayName = "Add Telemetry but disabled")]
```
Aniruddh25, line 29: Please also test with non-empty connection string but empty telemetryenabled flag

src/Cli.Tests/AddTelemetryTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            MockFileSystem fileSystem = FileSystemUtils.ProvisionMockFileSystem();
+            string configPath = "test-app-insights-config.json";
+            fileSystem.AddFile(configPath, new MockFileData(INITIAL_CONFIG));
+
+            // Initial State
+            Assert.IsTrue(fileSystem.FileExists(configPath));
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(fileSystem.File.ReadAllText(configPath), out RuntimeConfig? config));
+            Assert.IsNotNull(config);
+            Assert.IsNull(config.Runtime!.Telemetry);
```
Aniruddh25, line 40: Can we add a test where telemetry section already exists, and then we try to add app-insights section

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -167,6 +167,35 @@ public void TestAddEntity()
...
+        Program.Execute(initArgs, _cliLogger!, _fileSystem!, _runtimeConfigLoader!);
+
+        Assert.IsTrue(_runtimeConfigLoader!.TryLoadConfig(TEST_RUNTIME_CONFIG_FILE, out RuntimeConfig? runtimeConfig));
+
+        // Perform assertions on various properties.
+        Assert.IsNotNull(runtimeConfig);
+        Assert.IsNotNull(runtimeConfig.Runtime);
+        Assert.IsNull(runtimeConfig.Runtime.Telemetry);
+
+        string[] addTelemetryArgs = { "add-telemetry", "-c", TEST_RUNTIME_CONFIG_FILE, "--app-insights-enabled", "true", "--app-insights-conn-string", "InstrumentationKey=00000000" };
```
Aniruddh25, line 187: These arguments may not be necessary: "--app-insights-enabled", "true" since conn-string is provided...

Aniruddh25 (changes requested):
Good starters, can simplify cli options, add more test scenarios.

src/Cli/Commands/AddTelemetryOptions.cs
abhishekkumams: i think both of them should be independent, since there could be cases where conn-string is set but enabled is false to maybe use it in hot reload case, or in case where the user would want to keep the connection string info in the config file but not enable it.

src/Cli/ConfigGenerator.cs
abhishekkumams: no matter it is present or not, A config file will only have 1 appication insights section under telemetry, which would be replaced if present.

src/Cli.Tests/EndToEndTests.cs
abhishekkumams: i feel this should be considering scenarios where the developer may want to store the app-insights conn string but not use it for certain cases where it's impacting performance or due to temp issues with AppInsights server.

src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            bool isSuccess = ConfigGenerator.TryAddTelemetry(this, loader, fileSystem);
+
+            if (isSuccess)
+            {
+                logger.LogInformation("Successfully added telemetry to the configuration file.");
+            }
+            else
+            {
+                logger.LogError("Failed to add telemetry to the configuration file.");
```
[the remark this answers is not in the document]
abhishekkumams: All the failure causes will already be logged and are present inside the `TryAddTelemetry` method.

Aniruddh25 (commented):
> **if runtimeConfig.Runtime already has telemetry section, will this override it?**
-> yes
**What if both telemetry/appinsights sections are present?**
-> it will update the appinsights section.
**What if only telemetry section is present?**
-> it will add application insights section.
**What if telemetry + other future option than application insights is present in the read config file?**
-> it would only add application insights section without any change in other telemetry options. 

Overriding an already present telemetry section seems like an incorrect behavior to me. We should retain the original telemetry section and replace the appinsights section if present inside it,

src/Cli.Tests/AddTelemetryTests.cs
```diff
@@ -0,0 +1,165 @@
...
+                    }},
+                    {telemetrySection}
+                }},
+                ""entities"": {{}}";
+        }
+
+        /// <summary>
+        /// Represents a JSON string for the telemetry section of the config, with Application Insights enabled and a specified connection string.
+        /// </summary>
+        public const string TELEMETRY_SECTION_WITH_APP_INSIGHTS = @"
```
Aniruddh25, line 150: why make these public const? are these referenced outside this class?

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -311,7 +311,7 @@ private async Task<(string? idValue, string? partitionKeyValue)> GetIdAndPartiti
         /// <param name="parameter"></param>
         /// <returns></returns>
 #nullable enable
-        private string? GetPartitionKeyValue(string? partitionKeyPath, object? parameter)
+        private static string? GetPartitionKeyValue(string? partitionKeyPath, object? parameter)
```
Aniruddh25, line 314: why is this change needed in this PR?

src/Cli/ConfigGenerator.cs
```diff
@@ -1317,5 +1317,53 @@ private static EntityGraphQLOptions ConstructUpdatedGraphQLDetails(Entity entity
...
+                return false;
+            }
+
+            if (runtimeConfig.Runtime is null)
+            {
+                _logger.LogError("Invalid or missing 'runtime' section in config file: {runtimeConfigFile}.", runtimeConfigFile);
+                return false;
+            }
+
+            if (string.IsNullOrWhiteSpace(options.AppInsightsConnString))
```
Aniruddh25, line 1345: If we are checking for the validity of the connstring, shouldn't this ensure the connection actually succeeds to the appinsights telemetry resource? Also, we should add this validation to the `dab validate` command.

src/Core/Resolvers/CosmosQueryEngine.cs
abhishekkumams: because this method was not using any instance data, so compiler was giving warnings.

src/Cli/ConfigGenerator.cs
abhishekkumams: As of now i could only found one way to check whether connection succeds or not is by sending telemetry to App Insights and checking in the portal itself.

abhishekkumams:
> > **if runtimeConfig.Runtime already has telemetry section, will this override it?**
> > -> yes
> > **What if both telemetry/appinsights sections are present?**
> > -> it will update the appinsights section.
> > **What if only telemetry section is present?**
> > -> it will add application insights section.
> > **What if telemetry + other future option than application insights is present in the read config file?**
> > -> it would only add application insights section without any change in other telemetry options.
> 
> Overriding an already present telemetry section seems like an incorrect behavior to me. We should retain the original telemetry section and replace the appinsights section if present inside it,

Yes, by overrriding i meant only application Insights section of the telemetry not the complete telemetry section.

src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+    public class AddTelemetryOptions : Options
+    {
+        public AddTelemetryOptions(string appInsightsConnString, CliBool appInsightsEnabled, string? config) : base(config)
+        {
+            AppInsightsConnString = appInsightsConnString;
+            AppInsightsEnabled = appInsightsEnabled;
+        }
+
+        // Connection string for the Application Insights resource to which telemetry data should be sent.
+        // This optional is required and must be provided a valid connection string.
```
severussundar, line 27:
```suggestion
        // This optional is required and must be provided with a valid connection string.
```

src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+        [Option("app-insights-conn-string", Required = true, HelpText = "Connection string for the Application Insights resource for telemetry data")]
+        public string AppInsightsConnString { get; }
+
+        // To specifies whether Application Insights telemetry should be enabled. This flag is optional and default value is true.
```
severussundar, line 31:
```suggestion
        // To specify whether Application Insights telemetry should be enabled. This flag is optional and default value is true.
```

src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public void Handler(ILogger logger, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
+        {
+            logger.LogInformation("{productName} {version}", PRODUCT_NAME, ProductInfo.GetProductVersion());
+
+            bool isSuccess = ConfigGenerator.TryAddTelemetry(this, loader, fileSystem);
+
+            if (isSuccess)
+            {
+                logger.LogInformation("Successfully added telemetry to the configuration file.");
```
severussundar, line 43:
```suggestion
                logger.LogInformation("Successfully added App Insights options to the configuration file.");
```

src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+            else
+            {
+                logger.LogError("Failed to add telemetry to the configuration file.");
```
severussundar, line 47:
```suggestion
                logger.LogError("Failed to add App Insights options to the configuration file.");
```

src/Cli.Tests/AddTelemetryTests.cs
```diff
@@ -0,0 +1,165 @@
...
+            MockFileSystem fileSystem = FileSystemUtils.ProvisionMockFileSystem();
+            string configPath = "test-app-insights-config.json";
+            string runtimeSection = isEmptyTelemetry ? RUNTIME_SECTION_WITH_EMPTY_TELEMETRY_SECTION : RUNTIME_SECTION_WITH_APP_INSIGHTS_TELEMETRY_SECTION;
+            string configData = $"{{{SAMPLE_SCHEMA_DATA_SOURCE},{runtimeSection}}}";
+            fileSystem.AddFile(configPath, new MockFileData(configData));
+
+            // Initial State
+            Assert.IsTrue(fileSystem.FileExists(configPath));
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(fileSystem.File.ReadAllText(configPath), out RuntimeConfig? config));
+            Assert.IsNotNull(config);
```
severussundar, line 85: Maybe, adding one more Assert statement to check config.Runtime is not null could potentially help get away with the usage of `!` operator in the next few lines?

severussundar:
If the connection-string for the Application Insights resource is accessed through a .env file, does it work successfully?

abhishekkumams:
> If the connection-string for the Application Insights resource is accessed through a .env file, does it work successfully?

Yes, it does. users can use env variable in the config to source the value from the environment file.

Aniruddh25 (approved):
LGTM, thanks for completing the App-Insights feature!

severussundar:
> > If the connection-string for the Application Insights resource is accessed through a .env file, does it work successfully?
> 
> Yes, it does. users can use env variable in the config to source the value from the environment file.

Thank you for verifying this!

src/Cli.Tests/AddTelemetryTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            Assert.IsNotNull(config.Runtime);
+            Assert.IsNotNull(config.Runtime.Telemetry);
+
+            if (isEmptyTelemetry)
+            {
+                Assert.IsNull(config.Runtime!.Telemetry!.ApplicationInsights);
```
severussundar, line 91: nit: Since, `config.Runtime` and `config.Runtime.Telemetry` has been validated to be not null, check if it is possible to write this Assert without `!` operator.

src/Cli.Tests/AddTelemetryTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+            else
+            {
+                Assert.IsNotNull(config.Runtime!.Telemetry!.ApplicationInsights);
```
severussundar, line 95: nit: Same suggestion as above

src/Cli.Tests/AddTelemetryTests.cs
```diff
@@ -0,0 +1,166 @@
...
+                new AddTelemetryOptions("InstrumentationKey=11111-1111-111-11-1", CliBool.False, configPath),
+                new FileSystemRuntimeConfigLoader(fileSystem),
+                fileSystem);
+
+            // Assert after adding telemetry
+            Assert.IsTrue(isSuccess);
+            Assert.IsTrue(fileSystem.FileExists(configPath));
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(fileSystem.File.ReadAllText(configPath), out config));
+            Assert.IsNotNull(config);
+            Assert.IsNotNull(config.Runtime!.Telemetry);
```
severussundar, line 111: nit: Same suggestion ~ asserting `config.Runtime` to be not null could help get away with `!` operator

src/Cli/ConfigGenerator.cs
```diff
@@ -1317,5 +1317,53 @@ private static EntityGraphQLOptions ConstructUpdatedGraphQLDetails(Entity entity
 
             return graphQLType with { Operation = graphQLOperation };
         }
+
+        /// <summary>
+        /// This method will add the telemetry options to the config file. if the config file already has telemetry options,
```
severussundar, line 1322: super nit:
```suggestion
        /// This method will add the telemetry options to the config file. If the config file already has telemetry options,
```

PR #1958, opened 2024-01-10
Make relationship fields in referenced entities nullable

### LINKED ISSUE #1747: [Bug]: Cannot return null for non-nullable field.
### What happened?

If, in a One-to-Many or Many-to-One relationship, there is no value for the "one" side of the relationship, an error is returned, instead than returning a null value. As a result, this make impossible to represent correctly One-to-One relationship. Take this example, using the "School Database" available here: https://learn.microsoft.com/en-us/ef/ef6/resources/school-database

There is a one-to-one relationship between `Course` and `OnlineCourse` as `OnlineCourse` is a "specialization" of `Course`. Same goes between `Course` and `OnsiteCourse` and This is the following DAB config:

```json
{
  "$schema": "https://github.com/Azure/data-api-builder/releases/download/v0.8.52/dab.draft.schema.json",
  "data-source": {
    "database-type": "mssql",
    "connection-string": "@env('MSSQL')",
    "options": {
      "set-session-context": false
    }
  },
  "runtime": {
    "rest": {
      "enabled": true,
      "path": "/api"
    },
    "graphql": {
      "enabled": true,
      "path": "/graphql",
      "allow-introspection": true
    },
    "host": {
      "cors": {
        "origins": [],
        "allow-credentials": false
      },
      "authentication": {
        "provider": "StaticWebApps"
      },
      "mode": "development"
    }
  },
  "entities": {
    "Course": {
      "source": {
        "object": "dbo.Course",
        "type": "table"
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "Course",
          "plural": "Courses"
        }
      },
      "rest": {
        "path": "/courses",
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            {
              "action": "*"
            }
          ]
        }
      ],
      "relationships": {
        "OnlineCourse": {
          "cardinality": "one",
          "target.entity": "OnlineCourse"
        },
        "OnsiteCourse": {
          "cardinality": "one",
          "target.entity": "OnsiteCourse"
        }
      }
    },
    "OnlineCourse": {
      "source": {
        "object": "dbo.OnlineCourse",
        "type": "table"
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "OnlineCourse",
          "plural": "OnlineCourses"
        }
      },
      "rest": {
        "path": "/online-courses",
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            {
              "action": "*"
            }
          ]
        }
      ],
      "relationships": {
        "Course": {
          "cardinality": "one",
          "target.entity": "Course"
        }
      }
    },
    "OnsiteCourse": {
      "source": {
        "object": "dbo.OnsiteCourse",
        "type": "table"
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "OnsiteCourse",
          "plural": "OnsiteCourses"
        }
      },
      "rest": {
        "path": "/onsite-courses",
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            {
              "action": "*"
            }
          ]
        }
      ],
      "relationships": {
        "Course": {
          "cardinality": "one",
          "target.entity": "Course"
        }
      }
    }
  }
}
```

if I run the following GraphQL query:

```
query {
  courses {
    items {
      Title
      OnlineCourse {   
          URL
      }
      OnsiteCourse {
          Location
          Days
      }
    }
  }
}
```

I get the following error:

```
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field.",
      "locations": [
        {
          "line": 5,
          "column": 7
        }
      ],
      "path": [
        "courses",
        "items",
        0,
        "OnlineCourse"
      ],
      "extensions": {
        "code": "HC0018"
      }
    }
  ]
}
```

As for not all courses the field returned by the referenced entities will be available.

### Version

Microsoft.DataApiBuilder 0.8.52+c69925060e1942d28515b9c4b89ea24832da0c7c

Azure SQL

Local (including CLI)

GraphQL

### Relevant log output

```Text
Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'Hot Chocolate GraphQL Pipeline'
dbug: Azure.DataApiBuilder.Core.AuthenticationHelpers.ClientRoleHeaderAuthenticationMiddleware[0]
      1c429351-7b86-4eb2-9298-2e56e41dafe9: Request authentication state: Anonymous.
dbug: Azure.DataApiBuilder.Core.AuthenticationHelpers.ClientRoleHeaderAuthenticationMiddleware[0]
      1c429351-7b86-4eb2-9298-2e56e41dafe9: The request will be executed in the context of Anonymous role
      1c429351-7b86-4eb2-9298-2e56e41dafe9: Executing query:
SELECT TOP 100 [table0].[Title] AS [Title], JSON_QUERY ([table1_subq].[data]) AS [OnlineCourse], JSON_QUERY ([table3_subq].[data]) AS [OnsiteCourse] FROM [dbo].[Course] AS [table0] OUTER APPLY (SELECT TOP 1 [table1].[URL] AS [URL] FROM [dbo].[OnlineCourse] AS [table1] WHERE [table1].[CourseID] = [table0].[CourseID] ORDER BY [table1].[CourseID] ASC FOR JSON PATH, INCLUDE_NULL_VALUES,WITHOUT_ARRAY_WRAPPER) AS [table1_subq]([data]) OUTER APPLY (SELECT TOP 1 [table3].[Location] AS [Location], [table3].[Days] AS [Days] FROM [dbo].[OnsiteCourse] AS [table3] WHERE [table3].[CourseID] = [table0].[CourseID] ORDER BY [table3].[CourseID] ASC FOR JSON PATH, INCLUDE_NULL_VALUES,WITHOUT_ARRAY_WRAPPER) AS [table3_subq]([data]) WHERE 1 = 1 ORDER BY [table0].[CourseID] ASC FOR JSON PATH, INCLUDE_NULL_VALUES
      Request finished HTTP/2 POST https://localhost:5001/graphql application/json 212 - 500 - application/json;+charset=utf-8 5.6927ms
```

### Code of Conduct

Aniruddh25:
## Why make this change?

- Closes #1747 
- Records in referenced entities may not always be related to the referencing entity. E.g. in a `book-websiteplacement`  relationship, not all `books` (referenced entity) may have a `websiteplacement` (referencing) yet. This change is to ensure we get all records of the referenced entity including the ones that DON'T have any relationships. 

## What is this change?

- Previously, we used to rely on the nullability of the referenced fields to determine whether the relationship field in the referenced entity should be nullable or not. While this is applicable when we are considering the relationship fields of a referencing entity, it is actually restricting when used for a referenced entity. 
- For example, the referencing entity (BookWebsitePlacement) should have a nullable `books` (relationship field) based on whether the foreign key `book_id` in `book_website_placements` is nullable or not -> indicating whether a website placement MUST have a book or not. 
- On the other hand, the referenced entity `books` should always have a NULLABLE `websiteplacement` relationship field in order to include those books that don't have any `websiteplacement` published yet. Relying on the nullability of the `id` - the referenced field in the book->book_website_placement foreign key would make the relationship NON-NULLABLE but this restricts inclusion of those books that don't have any website placements hence the `error: "Cannot return null for non-nullable field"`. We need to make the relationship field nullable in such cases. 

- The source entity could be both the referencing and referenced entity
  in case of missing foreign keys in the db or self referencing relationships.
- Use the nullability of referencing columns to determine the nullability of the relationship field only if
1. there is exactly one relationship where source is the referencing entity.
DAB doesn't support multiple relationships at the moment.
and
2. when the source is not a referenced entity in any of the relationships.
- Identifies the scenario where the relationship field is from the referenced entity and always sets its nullability to `true`.

## How was this tested?

- [X] Integration Tests - existing data already had this bug, needed to modify the query to expose it.
- Ran the query in #1747 to verify this fix solved that issue.

## Sample Request(s)

```GraphQL
{
  books {
    items {
      id
      title
      websiteplacement {
        price
      }
    }
  }
}
```

BEFORE:
image

AFTER FIX:
There are no more errors and note that we now actually return `null` as the relationship field value(here, `websiteplacement`) when querying the referenced entity (here, `book`) which doesn't have any relationship with the referencing entity.
image

## NOTE
The issue is only exposed in a 1:1 relationship. This is because the only other scenario for querying a referenced entity is a 1:many relationship. And for records in the referenced entity(e.g. publisher) which are NOT related to the referencing entity(e.g. book), we explicitly check for nulls and return an empty array. See here: https://github.com/Azure/data-api-builder/blob/dacf3bccbe0a36da51776bf4afd1b4a0d4348ff4/src/Core/Resolvers/SqlQueryEngine.cs#L123

E.g.
```GraphQL Query for 1:many relationship
{ publishers {
    items {
      id
      name
      books {
        items {
          title
        }
      }
      }
    }
}
```

Expected Response
image

NOTE:
When DAB supports multiple relationships, nullability of each relationship field should be determined based on foreign keys associated with each relationship.


src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -130,7 +130,9 @@ public static class SchemaConverter
                         relationshipInfo.TargetEntityToFkDefinitionMap.TryGetValue(targetEntityName,
                             out List<ForeignKeyDefinition>? listOfForeignKeys))
                     {
-                        ForeignKeyDefinition? foreignKeyInfo = listOfForeignKeys.FirstOrDefault();
+                        ForeignKeyDefinition? foreignKeyInfo =
+                            listOfForeignKeys.Where(fk => fk.ReferencingColumns.Count > 0
```
severussundar, line 134: For 1:1 and n:1 relationships, the foreign key info is added in both the direction when iterating through the relationships defined in the config file - with the current entity as the referenced (and the related entity as referencing) as well as with current entity as the referencing entity (and the related entity as the referenced entity); as it is not possible to determine this without looking at the database metadata. Later when the database metadata is inferred, it is possible to know exactly the referenced/referencing entity, at which point, the referencing column and referenced column for exactly one pair (either current entity is referenced or current entity is referencing) would get updated. So, the entry with referencing/referenced column count = 0 is essentially invalid entry. 

Is this why, the condition (referencing & referenced columns count > 0) is introduced?

If so, adding the reasoning (in a better concise manner) would be helpful.

severussundar (approved):
LGTM

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
Aniruddh25: yes, that's exactly the reason, will add it in comments

Aniruddh25 (commented):
/azp run

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -146,7 +154,11 @@ public static class SchemaConverter
                             }
                             else
                             {
-                                isNullableRelationship = sourceDefinition.IsAnyColumnNullable(foreignKeyInfo.ReferencedColumns);
+                                // if the entity is "referenced", it may or may not have a relationship with the referencing
+                                // entity irrespective of nullability of the referenced columns.
+                                // Setting the relationship field to nullable ensures even those records
+                                // that are not related are considered while querying.
+                                isNullableRelationship = true;
```
abhishekkumams, line 161: good catch

abhishekkumams (approved):
LGTM

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -146,7 +154,11 @@ public static class SchemaConverter
                             }
                             else
                             {
-                                isNullableRelationship = sourceDefinition.IsAnyColumnNullable(foreignKeyInfo.ReferencedColumns);
+                                // if the entity is "referenced", it may or may not have a relationship with the referencing
```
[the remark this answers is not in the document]
Aniruddh25: no, I meant in the database - if a record of referenced table is related to any of the records in the referencing table.
At this point in code, runtime config has a relationship section defined for sure. Modified the explanation.

seantleonard (approved):

julianadormon:
@Aniruddh25  Hoping you can help :)
I have a website table in my database.
I also have a PaymentProvidersSettings table in my database which is linked to the Website table by WebsiteId.
While a record in PaymentProviderSetting will always be linked to the Website, there may not be any related records (yet).

I updated my dab.config using the following:
dab update Website --relationship "PaymentProviderSettings" --target.entity "PaymentProviderSetting" --cardinality one --relationship.fields "WebsiteId:WebsiteId"

And my schema.graphql shows this as  PaymentProviderSettings: PaymentProviderSetting! (non nullable).

When I query the following:

website_by_pk(WebsiteId: $input)
  {
    TimeZone
    CurrencyTypeId
    Culture
    PaymentProviderSettings {
      WebsiteId
      PaypalBusinessEmail
}

This returns, "Cannot return null for non-nullable field".
I upgraded to the latest dab, hoping the above may fix the issue, but it has not. I know if it is not quite the same point, but I think my database set-up is pretty standard. Any suggestions would be greatly appreciated.

julianadormon:
I just ended up making it a list of one, but I'm curios if there is a better way. Thanks!

PR #1962, opened 2024-01-11
Improve token claims processing by not erroring out on duplicate claim types.

### LINKED ISSUE #1957: [Bug]: Array in claims taken as duplicate claims with "Duplicate claims are not allowed within a request."
### What happened?

The claim type is "scope", stored as "array" value
Data api builder simply crash because we have array in our access token

image
image
image

### Version

Azure SQL

Local (including CLI)

REST

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
Closes #1957

## Why make this change?

- The referenced issue highlights the fact that duplicate claim types in an access token are not gracefully handled by DAB and result in a request error -> HTTP 403 Forbidden.
- This error occurs when a request utilizes a database policy for authorization and/or has `set-session-context` true in DAB's runtime config for mssql datasources (set-session-context is true by default when creating a DAB config using DABCLI.).

## What is this change?

JWT tokens may have claims whose values are structured data like JSON objects or JSON arrays. Token processing depends on the library used in code and in DAB's case: DotNet authentication helpers. Token processing in DAB's case results in the flattening of the JWT tokens key value pairs:
- For JSON arrays ` { "roles": ["role1", "role2"]}` the following claims are created -> `roles: role1` `roles:role2`
- For space delimted strings `{ "scp": "openid profile scope1" }` will be split into -> `scp: openid` `scp: profile` `scp: scope1`
- JSON objects will have the json object serialized as the value of the top level claim:
```json
{
  "_claim_sources": {
    "src1": {
      "endpoint": "https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects"
    }
  }
}
```
-> `_claim_sources: "<jsonobject>"`

With this change, DAB will ignore claims that are "duplicates" e.g. claims that have the same claim type after JWT flattening occurs. The `scp` example above illustrates this. If multiple scopes are present in an access token, those scope claims are not passed to SQL's session-context nor are the claims made available in database policy processing. While before this change, an exception and error was returned to the requestor, now the request will succeed if a db policy does not reference a claim that was ignored. 
- Using the `scp` example, if only one scp claim exists, that claim will be made available to db policy process and session context  creationg.
- when multiple scopes are present, those claims are ignored and dab will not fail the request as mentioned in the linked issue.

The `roles` claim is a special case. DAB will always make available to session context and database policy processing exactly one roles: the role that matches the client role header value (if in fact the user is a member of this role). If the client role header value doesn't match the user's role membership, earlier authentication processing will reject the request.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

### DIFF
--- a/src/Core/Authorization/AuthorizationResolver.cs
+++ b/src/Core/Authorization/AuthorizationResolver.cs
@@ -438,8 +438,23 @@ public class AuthorizationResolver : IAuthorizationResolver
     }
 
     /// <summary>
-    /// Helper method to extract all claims available in the HttpContext's user object and add the claims
-    /// to the claimsInRequestContext dictionary to be used for claimType -> claim lookups.
+    /// Returns a dictionary (claim name -> claim) populated from claims within a ClaimsPrincipal's ClaimsIdentity object.
+    /// Collect all unique token claims.
+    /// Claims with a value whose data type is one of the following may result in duplicate claim types which are not supported:
+    /// 1. JSON object
+    /// 2. JSON array: e.g. roles and groups claim, which are arrays of strings and are flattened to: roles: role1, roles: role2, groups: group1, groups: group2
+    /// 3. String delimited with spaces -> e.g. scp claim, which is a space delimited string of scopes and is flattened to: scp: scope1, scp: scope2. scp: scope3
+    /// claims
+    /// Background:
+    /// Claims are name valued pairs and nothing more. https://github.com/dotnet/aspnetcore/issues/13647#issuecomment-527523224
+    /// The library that parses the jwt token into claims is the one that decides *how* to resolve the claims from the token's JSON payload.
+    /// DotNet flattens the claims into a list: https://github.com/dotnet/aspnetcore/blob/282bfc1b486ae235a3395150a8d53073a57b7f43/src/Security/Authentication/OAuth/src/JsonKeyClaimAction.cs#L39-L53
+    /// Constraints
+    /// Only ignore duplicates because to ingore specific claim types (besides the role claim) is a breaking change.
+    /// e.g. When only one scope exists in a JWT token, one scp claim will be resolved: scp: scope1
+    /// However, a delimited scp claim will result in multiple claims -> scp: scope1 , scp: scope2.
+    /// We can't know which scp the user intends to use in a database policy or session context variable, so we remove all scp claims when duplicates exist.
+    /// Because claims are resolved into a dictionary by DAB, dupes have been and continue to be unsupported.
     /// </summary>
     /// <param name="context">HttpContext object used to extract the authenticated user's claims.</param>
     /// <returns>Dictionary with claimType -> claim mappings.</returns>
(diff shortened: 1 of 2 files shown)


src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,8 +438,23 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+    /// Background:
+    /// Claims are name valued pairs and nothing more. https://github.com/dotnet/aspnetcore/issues/13647#issuecomment-527523224
+    /// The library that parses the jwt token into claims is the one that decides *how* to resolve the claims from the token's JSON payload.
+    /// DotNet flattens the claims into a list: https://github.com/dotnet/aspnetcore/blob/282bfc1b486ae235a3395150a8d53073a57b7f43/src/Security/Authentication/OAuth/src/JsonKeyClaimAction.cs#L39-L53
+    /// Constraints
+    /// Only ignore duplicates because to ingore specific claim types (besides the role claim) is a breaking change.
+    /// e.g. When only one scope exists in a JWT token, one scp claim will be resolved: scp: scope1
+    /// However, a delimited scp claim will result in multiple claims -> scp: scope1 , scp: scope2.
+    /// We can't know which scp the user intends to use in a database policy or session context variable, so we remove all scp claims when duplicates exist.
+    /// Because claims are resolved into a dictionary by DAB, dupes have been and continue to be unsupported.
```
Aniruddh25, line 457: Now that we understand that inherently there could be multiple claims in the request, why do we restrict ourselves to resolve them into a only simple dictionary of <string, value> in DAB? Can't we resolve them into <string, list of values>? Accordingly, for DB policy, IIRC we add WHERE clause, the equality condition can simply be expanded to an `IN` condition, no?

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -479,19 +494,23 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
                 }
             }
         }
 
+        // Remove any claimType detected to have more than one instance.
+        // For example: scp claim is a json array of strings. If the user has multiple scopes,
+        // the claimType will be present multiple times.
+        foreach (string claimType in duplicateClaimNames)
+        {
+            claimsInRequestContext.Remove(claimType);
```
Aniruddh25, line 511: Doesn't simply ignoring the claims array misinform the constructed session context? Today, we have been erroring out, going forward if we dont want to error, I think better would be to still include these somehow in the session context.

Aniruddh25 (changes requested):
Need to rethink what we really want to do here.

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: Database policies would benefit from using your suggestion on how to translate a policy like `@claims.scp eq 'myHardCodedScope'` which would translate to the following tsql IF there are >1 scopes / >1 `scp` claims:
```tsql
WHERE 'myHardCodedScope' IN ('scp1', 'openid', 'profile', 'myHardCodedScope')
```
However, as you mention, we then would need to:
- Translate OData `eq` operator in all db policies referencing claims to the tsql `IN` clause ONLY when multiple instances of a claim exist. This isn't a simple change, unfortunately.
  - As opposed to implementing OData `IN` operator so developers can write db policies to match exactly how DAB will process the policy instead of needing to remember/look up in docs that `eq` translates to `in`. 
- I haven't finalized how to proceed with how `sp_set_session_context` should handle multiple instances of a claim type because that stored procedure only allows setting unique keys. e.g. `scp` claim could only appear once. and having the value be a space delimited list defeats the purpose and require even more custom tsql code in the database to parse.
  - @abratv 's suggestion could work here: _"why not simply using "scope_0", "scope_1",... "_ with the caveat that order of scopes is not guaranteed. `scope_0` may equal `openid` in one request and may be resolved as the value `scope_1` in another request.

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: I see your point. That would be better, yes.
Another thought: the issue should be split into two separate work items: 
1. Enable compatibility with dupe claim types in DB Policies 
2. Enable compatibility with dupe claim types in session context. (Quicker with fewer cascading requirements/changes to make this work.)

PR #1963, opened 2024-01-11
Cleanup of redundant class

abhishekkumams:
## Why make this change?

- Cleanup CliBoolean class as it is not adding any value and creating confusion with CliBool.cs

## How was this tested?
- Existing tests


seantleonard (approved):
thanks for making this change. I'm assuming `CliBoolean` was added by mistake at some point?

src/Config/ObjectModel/CliBoolean.cs
```diff
@@ -1,18 +0,0 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
-
-namespace Azure.DataApiBuilder.Config.ObjectModel;
-
-/// <summary>
-/// Class to represent boolean values in CLI. This is required over primitive boolean types because of the limitation of the CommandLineParser library
```
Aniruddh25, line 7: Please move this summary to the `CliBool.cs` file since looks like they serve the same purpose, but a summary is missing in CliBool.

Aniruddh25 (approved):

PR #1965, opened 2024-01-11
Exposing the field mappings helpers through interface.

rohkhann:
## Why make this change?
We use the ISqlMetadataProvider to interact with the metadataprovider of a class. This method helps get access to the field mappings for an entity.
With the current code, it is not possible to access field mappings of an entity.

## What is this change?
Add interface definitions in the ISqlMetadataProvider class.

## How was this tested?

Current tests should cover this


src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -123,6 +123,16 @@ public interface ISqlMetadataProvider
         /// <returns></returns>
         public IReadOnlyDictionary<string, DatabaseObject> GetEntityNamesAndDbObjects();
 
+        /// <summary>
+        /// Given entity name, gets the entity to column mappings if present.
+        /// </summary>
+        public bool TryGetEntityToFieldMappings(string entityName, [NotNullWhen(true)] out IReadOnlyDictionary<string, string>? mappings);
+
+        /// <summary>
+        /// Given entity name, gets the column to entity mappings if present.
```
Aniruddh25, line 132: nit: the description/function name should be fixed. This is getting the mapping from backing to exposed field names for the given entity. Neither the function name nor the description is suggesting this. Not your doing, it was there previously - but since you are exposing this in the interface now, I think we should fix the naming of both the functions and their descriptions. 
The name FieldToEntityMappings implies - we are retrieving a dictionary that maps a field name to its entity - but thats not what this function is doing

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -300,10 +298,8 @@ public bool TryGetEntityToColumnMappings(string entityName, [NotNullWhen(true)]
...
         }
 
-        /// <summary>
-        /// Given entity name, gets the column to entity mappings if present.
-        /// </summary>
-        public bool TryGetColumnToEntityMappings(string entityName, [NotNullWhen(true)] out IReadOnlyDictionary<string, string>? mappings)
+        /// <inheritdoc/>
+        public bool TryGetFieldToEntityMappings(string entityName, [NotNullWhen(true)] out IReadOnlyDictionary<string, string>? mappings)
         {
             Dictionary<string, string>? columntoEntityMappings;
```
Aniruddh25, line 304: please fix the name of this variable to `columnToExposedNameMap` for e.g.

Aniruddh25 (changes requested):
Overall looks fine, just needs to fix existing names/description. Also fix typo in PR title - should be `field` instead of `filed`

vadeveka (approved):

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: `TryGetBackingFieldToExposedFieldMap` to align with usage of `field`. Just so there is no confusion since we don't need to use the term `column` for an interface applying to all DBs such as Cosmos.

rohkhann: Fixed this to new name.

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
rohkhann: Fixed this to new name.

seantleonard (approved):
Thanks for making this change!

Aniruddh25 (approved):
Thanks for improving the codebase with clear naming conventions!

PR #1970, opened 2024-01-12
Update Dependencies and Suppress NU1603

seantleonard:
## Why make this change?
- Updates dependencies `SqlClient` and `Microsoft.AspNetCore.Authentication.JwtBearer`
- Explicitly set version of and include transitive dependencies: `System.IdentityModel.Tokens.Jwt` and `Microsoft.IdentityModel.JsonWebTokens`
- Suppress NU1603 when resolving dependencies from NuGet feed to prevent pipeline failures.
- Allows pipeline to pull compatible dependency version from private feed even when dependency is not the exact lower bounded version. DAB project treats warnings as errors and in this case, NU1603 is emitted and should be ignored.

- Background on issue: https://github.com/NuGet/Home/issues/5764

Even though our feed may have version `6.10.1` of the relevant package which meets the expectations of `>=6.10.0`, NuGet is expects and considers the "ideal" match to be `6.10.0` because that is what the package was built and tested against. Newer versions may or may not work and allows build systems to be consistent (per linked NuGet issue thread).

> ##[error]Core.csproj(0,0): Error NU1603: Microsoft.IdentityModel.Protocols 6.10.0 depends on Microsoft.IdentityModel.Tokens (>= 6.10.0) 
> but Microsoft.IdentityModel.Tokens 6.10.0 was not found. 
> An approximate best match of Microsoft.IdentityModel.Tokens 6.10.1 was resolved.

## What is this change?

- Suppress `NU1603` warning in all csproj files.
- Updates SqlClient dependency to version `5.1.4`.
- Updates Microsoft.AspNetCore.Authentication.JwtBearer to version `6.0.26`
- Updates System.IdentityModel.Tokens.Jwt and Microsoft.IdentityModel.JsonWebTokens to `6.35.0` explicitly even though transitive dependencies because latest SqlClient 5.1.4 references vulnerable 6.24.0 versions. Known issue in sqlclient https://github.com/dotnet/SqlClient/issues/2294 and workaround is to explicitly define the dependencies/needed version.

## How was this tested?

- [x] integration test: privately running the pipeline resolves expected dependencies without error.

### DIFF
--- a/src/Service.Tests/Azure.DataApiBuilder.Service.Tests.csproj
+++ b/src/Service.Tests/Azure.DataApiBuilder.Service.Tests.csproj
@@ -5,6 +5,7 @@
     <IsPackable>false</IsPackable>
     <Nullable>disable</Nullable>
     <OutputPath>$(BaseOutputPath)\tests</OutputPath>
+    <NoWarn>NU1603</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Service/Azure.DataApiBuilder.Service.csproj
+++ b/src/Service/Azure.DataApiBuilder.Service.csproj
@@ -6,6 +6,7 @@
         <OutputPath>$(BaseOutputPath)\engine</OutputPath>
         <RuntimeIdentifiers>win-x64;linux-x64;osx-x64</RuntimeIdentifiers>
         <EmbedUntrackedSources>true</EmbedUntrackedSources>
+        <NoWarn>NU1603</NoWarn>
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
@@ -28,7 +29,7 @@
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Docker|AnyCPU'">
-      <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+        <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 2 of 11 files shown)


src/Service/Azure.DataApiBuilder.Service.csproj
```diff
@@ -28,7 +29,7 @@
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Docker|AnyCPU'">
-      <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+        <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
```
Aniruddh25, line 32: nit: remove trailing space

aaronburtle (approved):
LGTM!

src/Service.Tests/Azure.DataApiBuilder.Service.Tests.csproj
```diff
@@ -5,6 +5,7 @@
     <IsPackable>false</IsPackable>
     <Nullable>disable</Nullable>
     <OutputPath>$(BaseOutputPath)\tests</OutputPath>
+    <NoWarn>NU1603</NoWarn>
   </PropertyGroup>
```
Aniruddh25, line 9: Why does PR title say `suppress NU1603 for Private feed`? this is for our public feed right?

Aniruddh25 (approved):
Thanks for chasing this down!

src/Service/Azure.DataApiBuilder.Service.csproj
seantleonard: The spacing was incorrect before. This is correct spacing (4 spaces)

Aniruddh25 (approved):

Issue #1976, opened 2024-01-16
[Feature] Mutation support for datawarehouse's


rohkhann:
Currently the query builders for mutations for data warehouse are not implemented. This feature is to track adding mutation support for graphql endpoint.

rohkhann:
Some explanation around the design: 
Datawarehouse create, update and delete are unable to generate result sets. This is because they do not have support for OUTPUT clause.

Alternate is to:
Create a DbOperationResult entity type and attach it to the mutations of data warehouses.  

DpOperationResult entity is added to the graphql schema. This entity has a single field called status that can take up success or failure. (Stored proc has gone with an empty result field – we could do similar ) 
Schema would look as follows:  

A screen shot of a computer code

Description automatically generated 

 

With the mutation return type looking as:  

A screenshot of a computer

Description automatically generated 

 

The call to the mutation would look as follows: 

A screenshot of a computer code

Description automatically generated 

 

Code changes in dab:  

In the GenerateSqlGraphQLObjects, we add an entity to the root node call DbOperationResult.  

A screen shot of a computer program

Description automatically generated 

During addition of mutation name the return type of the mutation to dbOperationResult: 

 

For multi-db scenario we were using the return type to determine which entity a mutation belongs to. However, since we are using a generic DbOperationResult as the return type, we would need to determine the entity type using the mutation name. This would  involved adding logic that can parse the entityname from the mutation name: 

 

A screen shot of a computer program

Description automatically generated 

Finally, if the mutation succeeds we need to send back the default result json document so code addition for that: 

A screen shot of a computer code

Description automatically generated

PR #1978, opened 2024-01-18
GraphQL Mutation support for datawarehouse.

rohkhann:
## Why make this change?

Issue: https://github.com/Azure/data-api-builder/issues/1976. 

GraphQL Mutations need to be supported for data warehouse. Reasons for listed changes listed in issue document.
This pr does not cover REST addition for insert/patch etc. Will be done in following pr's . issue tracking: https://github.com/Azure/data-api-builder/issues/1982

## What is this change?

1. Adding a DbOperationResult type that will be the return type for all dw mutation operations. 
2. Generating DWSQL queries for create, update, delete and upsert queries similar to that of other data sources.
3. Adding model directive to DW mutation nodes so that we can determine which db they belong to. 

## How was this tested?
Performed integration tests:
image

- [x] Unit Tests
Unit tests added for all dw scenarios.


src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -287,18 +288,51 @@ public static string GetDataSourceNameFromGraphQLContext(IMiddlewareContext cont
...
+        /// Get entity name from context object.
+        /// </summary>
+        public static string GetEntityNameFromContext(IMiddlewareContext context)
+        {
+            string entityName = context.Selection.Field.Type.TypeName();
+
+            if (entityName == "DbOperationResult")
+            {
+                // CUD for a mutation whose result set we do not have. Get Entity name mutation field directive.
+                if (GraphQLUtils.TryExtractGraphQLFieldModelName(context.Selection.Field.Directives, out string? modelName))
```
seantleonard, line 319: all dw schema fields will now have `@model` directive?

src/Service.GraphQLBuilder/Mutations/DeleteMutationBuilder.cs
```diff
@@ -17,13 +18,18 @@ public static class DeleteMutationBuilder
         /// <param name="name">Name of the GraphQL object to generate the delete field for.</param>
         /// <param name="objectTypeDefinitionNode">The GraphQL object type to generate for.</param>
         /// <param name="configEntity">Entity definition</param>
+        /// <param name="dbEntityName">Entity name in runtime config.</param>
+        /// <param name="databaseType">Database type the operation is targetting.</param>
+        /// <param name="returnEntityName">Name of return type entity for mutation.</param>
```
seantleonard, line 23: I think more context is necessary in the comments to explain why dbEntityName and returnEntityName could be different.

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -283,6 +298,8 @@ private DocumentNode GenerateCosmosGraphQLObjects(HashSet<string> dataSourceName
                     case DatabaseType.CosmosDB_NoSQL:
                         cosmosDataSourceNames.Add(_runtimeConfigProvider.GetConfig().GetDataSourceNameFromEntityName(entityName));
                         break;
+                    case DatabaseType.DWSQL:
```
seantleonard, line 301: is this needed? next case condition has DWSQL

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -232,6 +233,20 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
...
+            fields.TryAdd("result", field);
+
+            objectTypes.Add(GraphQLUtils.DB_OPERATION_RESULT_TYPE, new ObjectTypeDefinitionNode(
+                location: null,
+                name: nameNode,
+                description: null,
+                new List<DirectiveNode>(),
+                new List<NamedTypeNode>(),
+                fields.Values.ToImmutableList()));
+
```
seantleonard, line 249: this object is added even when DWSQL is not used as a datasource? Would this also be created if DWSQL only defined queries in its GraphQL schema? 

I'm also on the fence about this, but `GetDefaultResultFieldForMutation()` may be better placed as a private function in this class rather than the mutation builder class. While it is relevant to mutations, it is really just building a GraphQL Object type definition, that happens to be used for mutations, and not building a mutation explicitly.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -829,7 +838,8 @@ private FindRequestContext ConstructFindRequestContext(RestRequestContext contex
 
                 dbResultSetRow = dbResultSet is not null ?
                     (dbResultSet.Rows.FirstOrDefault() ?? new DbResultSetRow()) : null;
-                if (dbResultSetRow is not null && dbResultSetRow.Columns.Count == 0)
+
+                if (dbResultSetRow is not null && dbResultSetRow.Columns.Count == 0 && dbResultSet!.ResultProperties.TryGetValue("RecordsAffected", out object? recordsAffected) && (int)recordsAffected <= 0)
```
seantleonard, line 842: Looks like you're not guaranteed to have `recordsAffected` be non-null. So wouldn't the following && condition possibly fail in casting when recordsAffected is null?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -134,16 +127,28 @@ public class SqlMutationEngine : IMutationEngine
...
+                            // the result was not null previously, it indicates this DELETE lost
+                            // a concurrent request race. Hence, empty the non-null result.
+                            if (result is not null && result.Item1 is not null)
+                            {
+
+                                result = new Tuple<JsonDocument?, IMetadata?>(
+                                    default(JsonDocument),
+                                    PaginationMetadata.MakeEmptyPaginationMetadata());
+                            }
+                            else if (context.Selection.Type.TypeName() == GraphQLUtils.DB_OPERATION_RESULT_TYPE)
```
seantleonard, line 143: do we want the detection of `GraphQLUtils.DB_OPERATION_RESULT_TYPE` as a typename to be used outside of DW?

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -302,7 +302,7 @@ private List<OrderByColumn> PrimaryKeyAsOrderByColumns()
                 PaginationMetadata.Subqueries.Add(QueryBuilder.PAGINATION_FIELD_NAME, PaginationMetadata.MakeEmptyPaginationMetadata());
             }
 
-            EntityName = _underlyingFieldType.Name;
+            EntityName = GraphQLUtils.GetEntityNameFromContext(ctx);
```
seantleonard, line 305: Is there any way to only have this change apply if querying DW database?

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
```diff
@@ -0,0 +1,604 @@
...
+                    createbook(item: { id: 1, title: ""Awesome Book"", publisher_id: 1234 }) {
+                        __typename
+                    }
+                }
+            ";
+
+            JsonElement actual = await ExecuteGraphQLRequestAsync(graphQLMutation, graphQLMutationName, isAuthenticated: true);
+            string expected = @"
+              {
+                ""__typename"": ""DbOperationResult""
```
seantleonard, line 101: this needs to be documented because the typename of the result does not match the object type that the mutation affects. I think you've mentioned this in design discussion as point of confusion for developers because every mutation would return the same type.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
```diff
@@ -0,0 +1,604 @@
...
+              }
+            ";
+
+            SqlTestHelper.PerformTestEqualJsonStrings(expected, actual.ToString());
+        }
+
+        /// <summary>
+        /// <code>Do: </code>Update book in database and return its updated fields
+        /// <code>Check: </code>if the book with the id of the edited book and the new values exists in the database
+        /// and if the mutation query has returned the values correctly
```
seantleonard, line 111: all the test comments in this file should be updated to reflect that "return its updated fields" will not occur for DWSQL mutations. 

The "check" section should only really mention that a result value of "success" is verified. These tests aren't actually verifying the state of the db.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
```diff
@@ -0,0 +1,604 @@
...
+                ""__typename"": ""DbOperationResult""
+              }
+            ";
+
+            SqlTestHelper.PerformTestEqualJsonStrings(expected, actual.ToString());
+        }
+
+        /// <summary>
+        /// Demonstrates that using mapped column names for fields within the GraphQL mutation results in successful engine processing
+        /// of removal of the record where column1 = $id and the returned object representing the deleting record utilizes the mapped column values.
```
seantleonard, line 345: comment needs updating to reflect behavior of dwsql

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
```diff
@@ -0,0 +1,604 @@
...
+            string graphQLMutation = @"
+                mutation {
+                    updatebook(id: 1) {
+                        result
+                    }
+                }
+            ";
+
+            JsonElement result = await ExecuteGraphQLRequestAsync(graphQLMutation, graphQLMutationName, isAuthenticated: true);
+            SqlTestHelper.TestForErrorInGraphQLResponse(result.ToString(), message: $"item");
```
seantleonard, line 460: can you elaborate on what this is actually validating and what error you expect to see?

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
```diff
@@ -0,0 +1,604 @@
...
+            Assert.AreEqual("{\"result\":\"success\"}", response2.ToString());
+        }
+
+        [TestMethod]
+        [Ignore]
+        /// <inheritdoc/>
+        public override Task TestDbPolicyForCreateOperationReferencingFieldAbsentInRequest()
+        {
+            throw new NotImplementedException();
+        }
```
seantleonard, line 571: add comment for why this is ignored since your changes seem to include DB policies in your query building.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -134,16 +127,28 @@ public class SqlMutationEngine : IMutationEngine
                                 sqlMetadataProvider);
 
                         // If the number of records affected by DELETE were zero,
-                        // and yet the result was not null previously, it indicates this DELETE lost
-                        // a concurrent request race. Hence, empty the non-null result.
```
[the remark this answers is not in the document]
rohkhann: The comment is moved into the if statement at lines 134 and 135

src/Core/Resolvers/SqlMutationEngine.cs
rohkhann: In the design discussion, we had discussed this as a possibility that people could opt into. Sample scenarios: for Bulk inserts, bulk deletes etc user may not want to see the records returned but instead just want a summary represented by the db_operation_result type. We can expose this in the runtimeconfig. Similar thing can be used by users of other datasources if they just want summary results as opposed to modified records. Hence there is scope for this to be used for other data sources in the future.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: I can make it private within the same class. The reason that I did not add DW specific logic here to create this only when its DW was for two reasons:
1. This could be used by other datasources in the future. 
2. as per my understanding, adding this to the schema (even if its not used) should not have any perf impact or any impact on the schema. Hence, did not want to add any db specific logic here. 

Let me know if you think otherwise.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: fixed.

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -287,18 +288,51 @@ public static string GetDataSourceNameFromGraphQLContext(IMiddlewareContext cont
...
+
+                // Store dataSourceName on context for later use.
+                context.ContextData.TryAdd(GenerateDataSourceNameKeyFromPath(context), dataSourceName);
+            }
+            else
+            {
+                // Derive node from path - e.g. /books/{id} - node would be books.
+                // for this queryNode path we have stored the datasourceName needed to retrieve query and mutation engine of inner objects
+                object? obj = context.ContextData[GenerateDataSourceNameKeyFromPath(context)];
+                dataSourceName = obj?.ToString()!;
```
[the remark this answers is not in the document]
rohkhann: dataSourceName cannot be null over here as we have added it onto the path object. Also there is no logic change over here just code is refactored into this function.

src/Service.GraphQLBuilder/Mutations/DeleteMutationBuilder.cs
rohkhann: Added more context in the explanation.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -117,7 +110,7 @@ public class SqlMutationEngine : IMutationEngine
                     {
                         // When read permission is not configured, an error response is returned. So, the mutation result needs to
                         // be computed only when the read permission is configured.
-                        if (isReadPermissionConfigured)
+                        if (isReadPermissionConfigured && context.Selection.Type.TypeName() != GraphQLUtils.DB_OPERATION_RESULT_TYPE)
```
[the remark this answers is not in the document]
rohkhann: Makes sense. fixed this.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
rohkhann: Fixed.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
rohkhann: fixed.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
rohkhann: fixed.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
rohkhann: fixed.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
rohkhann: Do you mean just on the test or there is some specific place wherin we need to add this documentation?

src/Core/Resolvers/SqlMutationEngine.cs
rohkhann: Based on my understanding of the records affected field: https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbdatareader.recordsaffected?view=net-8.0. It says that this field must be implemented on dbdatareader closed. Even for select statements it returns a -1 but is a valid value. In existing DAB code today, I see similar logic used in the delete operation: 
image

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
rohkhann: Fixed message check.

seantleonard (approved):
thanks for making these changes. Looks good to me.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
rohkhann: Ok, where do i get access to add that?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -119,12 +112,16 @@ public class SqlMutationEngine : IMutationEngine
                         // be computed only when the read permission is configured.
                         if (isReadPermissionConfigured)
                         {
-                            // compute the mutation result before removing the element,
-                            // since typical GraphQL delete mutations return the metadata of the deleted item.
-                            result = await queryEngine.ExecuteAsync(
-                                        context,
-                                        GetBackingColumnsFromCollection(entityName: entityName, parameters: parameters, sqlMetadataProvider: sqlMetadataProvider),
-                                        dataSourceName);
+                            // For cases we only require summary, we can skip getting the impacted records.
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
```diff
@@ -0,0 +1,604 @@
...
+            string graphQLMutation = @"
+                mutation {
+                    updatebook(id: 1) {
+                        result
+                    }
+                }
+            ";
+
+            JsonElement result = await ExecuteGraphQLRequestAsync(graphQLMutation, graphQLMutationName, isAuthenticated: true);
+            SqlTestHelper.TestForErrorInGraphQLResponse(result.ToString(), message: $"The argument `item` is required");
```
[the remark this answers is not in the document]
rohkhann: This error message is only specific to the DW scenario. There is no impact to the MsSql, MySql, PG versions.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -219,6 +219,8 @@ private static NameNode GenerateInputTypeName(string typeName)
 
         /// <summary>
         /// Generate the `create` mutation field for the GraphQL mutations for a given Object Definition
+        /// ReturnEntityName can be different from dbEntityName in cases where user wants summary results returned (through the DBOperationResult entity)
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: Makses sense. created GetDbOperationResultField.

src/Service.GraphQLBuilder/GraphQLUtils.cs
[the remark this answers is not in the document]
rohkhann: added null check here.

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -287,18 +288,51 @@ public static string GetDataSourceNameFromGraphQLContext(IMiddlewareContext cont
...
+        }
 
+        /// <summary>
+        /// Get entity name from context object.
+        /// </summary>
+        public static string GetEntityNameFromContext(IMiddlewareContext context)
+        {
+            string entityName = context.Selection.Field.Type.TypeName();
+
+            if (entityName == "DbOperationResult")
```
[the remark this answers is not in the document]
rohkhann: Created an issue for this as I believe this is part of having a reserved name check for entity names in graphql on dab side: https://github.com/Azure/data-api-builder/issues/1988

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
rohkhann: Will create issue in docs repo.

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -287,18 +288,60 @@ public static string GetDataSourceNameFromGraphQLContext(IMiddlewareContext cont
             {
```
vadeveka, line 288: NIT: GetDataSourceNameFromGraphQLContext  method's description  seems to be misleading

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -249,7 +254,7 @@ private static NameNode GenerateInputTypeName(string typeName)
                 entities);
 
             // Create authorize directive denoting allowed roles
-            List<DirectiveNode> fieldDefinitionNodeDirectives = new();
+            List<DirectiveNode> fieldDefinitionNodeDirectives = new() { new(ModelDirectiveType.DirectiveName, new ArgumentNode("name", dbEntityName)) };
```
vadeveka, line 257: Do we have unit tests over this change or covered by integration tests?

vadeveka (approved):

src/Service.GraphQLBuilder/GraphQLUtils.cs
rohkhann: fixed.

src/Service.GraphQLBuilder/Mutations/UpdateMutationBuilder.cs
```diff
@@ -236,7 +239,7 @@ private static NameNode GenerateInputTypeName(string typeName)
                     new List<DirectiveNode>()));
 
             // Create authorize directive denoting allowed roles
-            List<DirectiveNode> fieldDefinitionNodeDirectives = new();
+            List<DirectiveNode> fieldDefinitionNodeDirectives = new() { new(ModelDirectiveType.DirectiveName, new ArgumentNode("name", dbEntityName)) };
```
Aniruddh25, line 242: "name" can be replaced with `ModelDirectiveType.ModelNameArgument` - in all 3 `MutationBuilders`

Aniruddh25 (commented):
Even though its merged, submitting some pending comments.

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -94,7 +94,7 @@ private static Entity GenerateEmptyEntity()
             FieldDefinitionNode createField =
                 query.Fields.Where(f => f.Name.Value == $"createFoo").First();
             Assert.AreEqual(expected: isAuthorizeDirectiveExpected ? 1 : 0,
-                actual: createField.Directives.Count);
+                actual: createField.Directives.Where(x => x.Name.Value is "authorize").Count());
```
Aniruddh25, line 97: Could have used the constant: `GraphQLUtils.AUTHORIZE_DIRECTIVE`

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -249,7 +254,7 @@ private static NameNode GenerateInputTypeName(string typeName)
                 entities);
 
             // Create authorize directive denoting allowed roles
```
Aniruddh25, line 256: nit: this comment applies to CreateAuthorizationDirectiveIfNecessary() call below

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -232,6 +233,20 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
...
+            Dictionary<string, FieldDefinitionNode> fields = new();
+            NameNode nameNode = new(value: GraphQLUtils.DB_OPERATION_RESULT_TYPE);
+            FieldDefinitionNode field = GetDbOperationResultField();
+
+            fields.TryAdd("result", field);
+
+            objectTypes.Add(GraphQLUtils.DB_OPERATION_RESULT_TYPE, new ObjectTypeDefinitionNode(
+                location: null,
+                name: nameNode,
+                description: null,
```
Aniruddh25, line 245: Better to provide some description here.

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -267,6 +282,21 @@ private DocumentNode GenerateCosmosGraphQLObjects(HashSet<string> dataSourceName
...
 
+        /// <summary>
+        /// Create and return a default GraphQL result field for a mutation which doesn't
+        /// define a result set and doesn't return any rows.
+        /// </summary>
+        private static FieldDefinitionNode GetDbOperationResultField()
+        {
+            return new(
+                location: null,
+                name: new("result"),
```
Aniruddh25, line 293: define a constant string for this "result" field name, in GraphQLUtils perhaps?

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
Aniruddh25: Actually, I dont see the need to special case it for only DW.. This is a query field right? So, shouldn't it be same irrespective of db type? Since, I thought the type change we are doing is for Mutation only.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -83,14 +83,7 @@ public class SqlMutationEngine : IMutationEngine
 
             dataSourceName = GetValidatedDataSourceName(dataSourceName);
             string graphqlMutationName = context.Selection.Field.Name.Value;
-            IOutputType outputType = context.Selection.Field.Type;
-            string entityName = outputType.TypeName();
-            ObjectType _underlyingFieldType = GraphQLUtils.UnderlyingGraphQLEntityType(outputType);
-
-            if (GraphQLUtils.TryExtractGraphQLFieldModelName(_underlyingFieldType.Directives, out string? modelName))
```
Aniruddh25, line 90: See this is exactly what we can do in SqlQueryStructure too

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -185,7 +215,65 @@ public string Build(SqlExecuteStructure structure)
         /// <inheritdoc />
         public string Build(SqlUpsertQueryStructure structure)
```
Aniruddh25, line 216: fyi: `Upsert` will only be used in REST API PATCH/PUT

src/Service.Tests/SqlTests/GraphQLMutationTests/DwSqlGraphQLMutationTests.cs
Aniruddh25: given that you honor policies while generating the corresponding query, why do you think this is not supported for DWsql?

Aniruddh25:
Opened 2 closed conversations to resolve: 
1. https://github.com/Azure/data-api-builder/pull/1978#discussion_r1465685806
2. https://github.com/Azure/data-api-builder/pull/1978#discussion_r1465696478

src/Service.GraphQLBuilder/Mutations/UpdateMutationBuilder.cs
rohkhann: fixed.

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
rohkhann: fixed.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: fixed.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: fixed.

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -287,18 +288,60 @@ public static string GetDataSourceNameFromGraphQLContext(IMiddlewareContext cont
...
+
+            return dataSourceName;
+        }
+
+        /// <summary>
+        /// Get entity name from context object.
+        /// </summary>
+        public static string GetEntityNameFromContext(IMiddlewareContext context)
+        {
+            string entityName = context.Selection.Field.Type.TypeName();
```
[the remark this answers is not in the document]
rohkhann: fixed.

PR #1980, opened 2024-01-18
Support environment variable serialization for data source options

### LINKED ISSUE #1956: [Bug]: env() transformation does not work after upgrading the project from 0.7.6 to 0.9.7
### What happened?

Hi,

After upgrading my project from `0.7.6` to `0.9.7`, the @env() transformation stopped working. The environment variables are not replaced at all. I have the same problem both locally (CLI update) and in Azure ContainerApps (image update). 

In the past, for `0.7.6`, when I was developing locally, I set the environment variables in the terminal in the following way and everything worked fine without any problems. I get the same problem when I try to use the `.env` file.

Is there a new undocumented option to enable environment variable replacement?

```powershell
$env:DATABASE_NAME = 'abc'
$env:DATABASE_CONNECTION_STRING = 'abc'
$env:GRAPHQL_SCHEMA_PATH = 'dab-config.schema.gql'
$env:JWT_AUDIENCE = 'abc'
$env:JWT_ISSUER = 'abc'

dab start
```

dab-config.json:
```json
{
  "$schema": "https://github.com/Azure/data-api-builder/releases/download/v0.7.6/dab.draft.schema.json",
  "data-source": {
    "database-type": "cosmosdb_nosql",
    "options": {
      "database": "@env('DATABASE_NAME')",
      "schema": "@env('GRAPHQL_SCHEMA_PATH')"
    },
    "connection-string": "@env('DATABASE_CONNECTION_STRING')"
  },
  "runtime": {
    "rest": {
      "enabled": true,
      "path": "/api"
    },
    "graphql": {
      "allow-introspection": true,
      "enabled": true,
      "path": "/graphql"
    },
    "host": {
      "mode": "development",
      "cors": {
        "origins": [],
        "allow-credentials": false
      },
      "authentication": {
        "provider": "AzureAD",
        "jwt": {
          "audience": "@env('JWT_AUDIENCE')",
          "issuer": "@env('JWT_ISSUER')"
        }
      }
    }
  },
  "entities": {
    "NoteCard": {
      "source": "NoteCards",
      "permissions": [
        {
          "role": "authenticated",
          "actions": [
            "*"
          ]
        }
      ]
    }
  }
}
```

### Version

0.9.7

CosmosDB NoSQL

Local (including CLI), Container Apps

REST, GraphQL

### Relevant log output

```Text
PS D:\r\DigestFusion\scripts\data-api-builder\cosmos> dab start
Information: Microsoft.DataApiBuilder 0.9.7+e560142426d1c080b9fd7b7fabff51a276f6bf61
Information: Environment variable DAB_ENVIRONMENT is development
Loading config file from dab-config.json.
Information: Setting default minimum LogLevel: Debug for Development mode.
Starting the runtime engine...
Loading config file from dab-config.json.
info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[63]
      User profile is available. Using 'C:\Users\nieli\AppData\Local\ASP.NET\DataProtection-Keys' as key repository and Windows DPAPI to encrypt keys at rest.
fail: Azure.DataApiBuilder.Service.Startup[0]
      Unable to complete runtime initialization. Refer to exception for error details.
      System.IO.FileNotFoundException: The GraphQL schema file at '@env('GRAPHQL_SCHEMA_PATH')' could not be found. Ensure that it is a path relative to the runtime.
         at Azure.DataApiBuilder.Core.Configurations.RuntimeConfigValidator.ValidateDatabaseType(RuntimeConfig runtimeConfig, IFileSystem fileSystem, ILogger logger)
         at Azure.DataApiBuilder.Core.Configurations.RuntimeConfigValidator.ValidateConfig()
         at Azure.DataApiBuilder.Service.Startup.PerformOnConfigChangeAsync(IApplicationBuilder app)
fail: Azure.DataApiBuilder.Service.Startup[0]
      Could not initialize the engine with the runtime config file: dab-config.json
info: Microsoft.Hosting.Lifetime[0]
      Application is shutting down...
Unable to launch the Data API builder engine.
Error: Failed to start the engine.
```

### Code of Conduct

abhishekkumams:
## Why make this change?

-  Closes #1956 
  - Data source section was not supporting env transformation.

## What is this change?

- Added a DataSourceConverterFactory which will properly deserialize the data-source section considering the _replaceEnvVariable to decide weather to transform or not.

## How was this tested?
- [X] Unit Tests


src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +106,51 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+        /// Tests the parsing of the runtime configuration with environment variables based on replaceEnvVar parameter.
+        /// </summary>
+        /// <param name="replaceEnvVar">A boolean indicating whether to replace environment variables in the configuration.</param>
+        [DataTestMethod]
+        [DataRow(false, DisplayName = "Do not replace environment variables.")]
+        [DataRow(true, DisplayName = "Replace environment variables.")]
+        public void TestConfigParsingWithEnvVarReplacement(bool replaceEnvVar)
+        {
+            SetEnvironmentVariablesFromText(_environmentFileContent);
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
```
seantleonard, line 119: Nit: it would be clearer to explicitly distinguish the "Act" apart from the "Assert" section.

e.g. save result of `RuntimeConfigLoader.TryParseConfig(...)` in the act section because you are validating the behavior of the runtimeconfig parsing. Then, perform asserts.

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+                            case "database-type":
+                                dataSource = dataSource with { DatabaseType = EnumExtensions.Deserialize<DatabaseType>(reader.DeserializeString(_replaceEnvVar)!) };
+                                break;
+                            case "connection-string":
+                                dataSource = dataSource with { ConnectionString = reader.DeserializeString(replaceEnvVar: _replaceEnvVar)! };
+                                break;
+                            case "options":
+                                Dictionary<string, JsonElement> optionsDict = new();
+                                while (reader.Read() && reader.TokenType is not JsonTokenType.EndObject)
+                                {
```
Aniruddh25, line 76: Can't the outer while have a similar condition `reader.Read() && reader.TokenType is not JsonTokenType.EndObject` instead of adding a `break` as the 1st statement inside the loop?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +106,51 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+        /// <summary>
+        /// Retrieves the value of an environment variable if replacement is enabled, otherwise returns a placeholder string.
+        /// </summary>
+        /// <param name="envVarName">The name of the environment variable.</param>
+        /// <param name="replaceEnvVar">A boolean indicating whether to replace the environment variable with its value.</param>
+        /// <returns>
+        /// If replacement is enabled, the value of the environment variable. 
+        /// Otherwise, a placeholder string in the format "@env('variableName')".
+        /// </returns>
+        private static string GetEnvironmentVariableIfReplaced(string envVarName, bool replaceEnvVar)
```
seantleonard, line 142: Name would better describe what function does by stating it's returning an expected value.
`GetExpectedPropertyValue(envVarName, replaceEnvVar)`

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -301,8 +346,7 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
   },
   ""data-source"": {
     ""database-type"": " + enumString + @",
-    ""connection-string"": ""server=dataapibuilder;database=" + reps[++index % reps.Length] + @";uid=" + reps[++index % reps.Length] + @";Password=" + reps[++index % reps.Length] + @";"",
-    ""resolver-config-file"": """ + reps[++index % reps.Length] + @"""
```
seantleonard, line 305: was `""resolver-config-file"": """ + reps[++index % reps.Length] + @"""` legacy code? I suppose since `resolver-config-file` is no longer used.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +423,102 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
 ";
         }
 
+        /// <summary>
+        /// Config containing environment variables.
+        /// </summary>
+        private static string _configWithEnvVar = @"
```
seantleonard, line 429: need a test to confirm behavior of MSSQL options.

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+                                        string stringValue = reader.DeserializeString(replaceEnvVar: _replaceEnvVar)!;
+
+                                        using (JsonDocument doc = JsonDocument.Parse($"\"{stringValue}\""))
+                                        {
+                                            element = doc.RootElement.Clone();
+                                        }
+                                    }
+                                    else
+                                    {
+                                        bool boolValue = reader.GetBoolean();
```
Aniruddh25, line 91: why are we assuming the element could only be a string or a boolean?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +106,51 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
         }
 
+        /// <summary>
+        /// Tests the parsing of the runtime configuration with environment variables based on replaceEnvVar parameter.
+        /// </summary>
+        /// <param name="replaceEnvVar">A boolean indicating whether to replace environment variables in the configuration.</param>
+        [DataTestMethod]
+        [DataRow(false, DisplayName = "Do not replace environment variables.")]
+        [DataRow(true, DisplayName = "Replace environment variables.")]
+        public void TestConfigParsingWithEnvVarReplacement(bool replaceEnvVar)
```
seantleonard, line 116: as mentioned in another comment, this test should have an additional param representing the config value because you'll need to test JSON converter behavior with MSSQL options as well as cosmos options.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +423,102 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+        private static string _environmentFileContent =
+            "DATABASE_TYPE=cosmosdb_nosql" + "\n" +
+            "DATABASE_NAME=planet" + "\n" +
+            "GRAPHQL_SCHEMA_PATH=gql-schema.gql" + "\n" +
+            "DATABASE_CONNECTION_STRING=TEST_CONNECTION_STRING" + "\n" +
+            "JWT_AUDIENCE=TEST_AUDIENCE" + "\n" +
+            "JWT_ISSUER=TEST_ISSUER" + "\n" +
+            "ENTITY_SOURCE=books" + "\n" +
+            "ENTITY_ROLE=anonymous" + "\n" +
+            "ENTITY_REST_PATH=/books";
```
seantleonard, line 506: why not create a dictionary so you can avoid all the helpers which parse through this text?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +423,102 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+        /// Config containing environment variables.
+        /// </summary>
+        private static string _configWithEnvVar = @"
+          {
+            ""$schema"": ""/dab-schema.json"",
+            ""data-source"": {
+              ""database-type"": ""cosmosdb_nosql"",
+              ""options"": {
+                ""database"": ""@env('DATABASE_NAME')"",
+                ""schema"": ""@env('GRAPHQL_SCHEMA_PATH')""
```
seantleonard, line 436: because you have a new DataSource converter, you'll need tests to validate behavior of:
- `options.database` and `options.schema` are **null** because this has caused regressions in the past.
- `options` is null or empty.

seantleonard (changes requested):
more tests to exercise MSSQL options and ensure options properties properly handle being set to null.  and some test enhancement suggestions.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
Aniruddh25, line 119: Don't we have any existing tests in ConfigurationTests that would check environment variables are correctly replaced? Why do we need to add a new test?

Aniruddh25 (changes requested):
Some questions on the approach..

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
abhishekkumams: yes, correct.

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <param name="replaceEnvVar">Whether to replace environment variable with its
+        /// value or not while deserializing.</param>
+        public DataSourceConverter(bool replaceEnvVar)
+        {
+            _replaceEnvVar = replaceEnvVar;
+        }
+
+        public override DataSource? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            DataSource dataSource = new(DatabaseType.MSSQL, string.Empty, null);
```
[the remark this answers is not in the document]
abhishekkumams: databaseType takes only enum value, so i initialized it with one of the possible values. Which is updated below.

src/Config/Converters/DataSourceConverterFactory.cs
abhishekkumams: yep, thanks for pointing

src/Config/Converters/DataSourceConverterFactory.cs
abhishekkumams: anything else would be an error. updated the code to throw exception.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
abhishekkumams: the existing ones doesn't check the DataSource section.

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+                                break;
+                            case "options":
+                                if (reader.TokenType == JsonTokenType.Null)
+                                {
+                                    dataSource = dataSource with { Options = null };
+                                    break;
+                                }
+                                else
+                                {
+                                    Dictionary<string, JsonElement> optionsDict = new();
```
aaronpowell, line 76: Given there's now a parser that's going to go over the properties of the `options` object, does it still make sense to keep it as a `Dictionary<string, JsonElement>` or would it be better to use `Dictionary<string, object>`?

By the looks of the logic, it is parsing each property value, but then it turns it back into a `JsonElement`, which we'll then have to parse out later anyway.

If the dictionary type was changed it would skip having to deserialise then serialise and later deserialise again.

Aniruddh25:
Looks good after @aaronpowell's comment is addressed.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +108,112 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+        [DataRow(true, "mssql", DisplayName = "Replace environment variables containing boolean values.")]
+        [DataRow(false, "cosmosdb_nosql", DisplayName = "Do not replace environment variables containing string values.")]
+        [DataRow(true, "cosmosdb_nosql", DisplayName = "Replace environment variables containing string values.")]
+        [DataRow(false, "mysql", DisplayName = "Do not replace environment variables when datasource option is empty.")]
+        [DataRow(true, "mysql", DisplayName = "Replace environment variables when datasource option is empty.")]
+        [DataRow(false, "postgresql", DisplayName = "Do not replace environment variables when datasource option is null.")]
+        [DataRow(true, "postgresql", DisplayName = "Replace environment variables when datasource option is null.")]
+        [DataRow(false, "dwsql", DisplayName = "Do not replace environment variables when datasource option is not given.")]
+        [DataRow(true, "dwsql", DisplayName = "Replace environment variables when datasource option is not given.")]
+        public void TestConfigParsingWithEnvVarReplacement(bool replaceEnvVar, string databaseType)
```
seantleonard, line 132: nit: can you be specific with test function name such that it says you're testing runtimeconfig.datasource.options property env var replacement?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +108,112 @@ public class RuntimeConfigLoaderJsonDeserializerTests
             }
         }
 
+        /// <summary>
+        /// Tests the parsing of the runtime configuration with environment variables based on replaceEnvVar parameter.
+        /// Verifies the below five different cases:
```
seantleonard, line 113: verifies that in the below 5 cases... what happens?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +108,112 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+        [DataRow(true, "mysql", DisplayName = "Replace environment variables when datasource option is empty.")]
+        [DataRow(false, "postgresql", DisplayName = "Do not replace environment variables when datasource option is null.")]
+        [DataRow(true, "postgresql", DisplayName = "Replace environment variables when datasource option is null.")]
+        [DataRow(false, "dwsql", DisplayName = "Do not replace environment variables when datasource option is not given.")]
+        [DataRow(true, "dwsql", DisplayName = "Replace environment variables when datasource option is not given.")]
+        public void TestConfigParsingWithEnvVarReplacement(bool replaceEnvVar, string databaseType)
+        {
+            // Act
+            SetEnvironmentVariablesFromDictionary(_environmentFileContentDict);
+            string configWithEnvVar = _configWithVariableDataSource.Replace("{0}", GetDataSourceConfigForGivenDatabase(databaseType));
```
seantleonard, line 136: nit: should be `// Arrange` because you're setting up your test in anticipation of what you're testing.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +108,112 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
```
seantleonard, line 138: nit: have single operation performing the `TryParseConfig()` and save the result.

Then you can have a different line perform the assert in the `// Assert` section.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +108,112 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+                    Assert.AreEqual(runtimeConfig.DataSource.Options.Count, 0);
+                    break;
+                case "postgresql":
+                    Assert.AreEqual(runtimeConfig.DataSource.DatabaseType, DatabaseType.PostgreSQL);
+                    Assert.AreEqual(runtimeConfig.DataSource.Options, null);
+                    break;
+                case "dwsql":
+                    Assert.AreEqual(runtimeConfig.DataSource.DatabaseType, DatabaseType.DWSQL);
+                    Assert.AreEqual(runtimeConfig.DataSource.Options, null);
+                    break;
```
seantleonard, line 165: need to describe in your test summary that these three db types are unaffected by `replaceEnvVar` because those db's don't support `options`.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +108,112 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+        /// <summary>
+        /// Test the parsing of DataSource section in runtime config for cosmosdb_nosql
+        /// where under cosmosDb options has database as null, container is not provided, and schema is an empty string.
+        /// </summary>
+        [TestMethod]
+        public void TestConfigParsingWhenDataSourceOptionsForCosmosDBContainsInvalidValues()
+        {
+            // Act
+            SetEnvironmentVariablesFromDictionary(_environmentFileContentDict);
+            string configWithEnvVar = _configWithVariableDataSource.Replace("{0}", GetDataSourceOptionsForCosmosDBWithInvalidValues());
```
seantleonard, line 181: nit: should be `// Arrange` for when setting up test. then `// Act` for the line that is exercising the system under test,

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -106,6 +108,112 @@ public class RuntimeConfigLoaderJsonDeserializerTests
...
+                    break;
+            }
+
+            // Cleanup
+            ClearEnvironmentVariablesFromDictionary(_environmentFileContentDict);
+        }
+
+        /// <summary>
+        /// Test the parsing of DataSource section in runtime config for cosmosdb_nosql
+        /// where under cosmosDb options has database as null, container is not provided, and schema is an empty string.
```
seantleonard, line 174: What do you expect the behavior to be when such a config is provided?The function name states `DataSourceOptionsForCosmosDBContainsInvalidValues` can you elaborate on:
- why no error is expected
- how is dab handling 'invalid values' for options?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +486,143 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
 ";
         }
 
+        /// <summary>
+        /// Config containing variable data-source.
```
seantleonard, line 490:
```suggestion
        /// Config containing data-source property.
```

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +486,143 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+                  ""path"": ""/graphql""
+                },
+                ""host"": {
+                  ""mode"": ""development"",
+                  ""cors"": {
+                    ""origins"": [],
+                    ""allow-credentials"": false
+                  },
+                  ""authentication"": {
+                    ""provider"": ""AzureAD""
```
seantleonard, line 513: nit: EntraID

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +486,143 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+                          }";
+                    break;
+                case "mysql":
+                    databaseTypeEnvVariable = $"@env('MYSQL_DB_TYPE')";
+                    options = @",""options"": {}";
+                    break;
+                case "postgresql":
+                    databaseTypeEnvVariable = $"@env('POSTGRESQL_DB_TYPE')";
+                    options = @",""options"": null";
+                    break;
```
seantleonard, line 554: by including values for `options` when the dbtype doesn't support more options, are you testing that DAB ignores those fields?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +486,143 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+            return $@"
+            {{
+                ""database-type"": ""{databaseTypeEnvVariable}"",
+                ""connection-string"": ""@env('DATABASE_CONNECTION_STRING')""
+                {options}
+            }}";
+        }
+
+        /// <summary>
+        /// Gets the data source options for CosmosDB with no container, null database, and empty schema.
```
seantleonard, line 570: need to explain purpose of including "invalid" values here.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +486,143 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+                ""connection-string"": ""@env('DATABASE_CONNECTION_STRING')"",
+                ""options"": {{
+                    ""database"": null,
+                    ""schema"": """"
+                }}
+            }}";
+        }
+
+        /// <summary>
+        /// Clears the environment variables defined in the provided text.
```
seantleonard, line 586: `provided text.` -> dictionary? test?

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +486,143 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+        private static void ClearEnvironmentVariablesFromDictionary(Dictionary<string, string> environmentFileContentDict)
+        {
+            foreach (KeyValuePair<string, string> entry in environmentFileContentDict)
+            {
+                Environment.SetEnvironmentVariable(entry.Key, null);
+            }
+        }
+
+        /// <summary>
+        /// A string representing environment variables for testing environment variable replacement in the config.
```
seantleonard, line 598:
```suggestion
        /// A dictionary representing environment variables for testing environment variable replacement in the config.
```

seantleonard: is that what you mean by 'string'?

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+                                                }
+                                            }
+                                            else
+                                            {
+                                                // CosmosDbNoSQLDataSourceOptions will contain string values.
+                                                using (JsonDocument doc = JsonDocument.Parse($"\"{stringValue}\""))
+                                                {
+                                                    element = doc.RootElement.Clone();
+                                                }
+                                            }
```
seantleonard, line 101: Wouldn't you already know the db type at this point of json parsing? I suppose only if the user places 'database-type` before the `options` property.

So if user tries to provide a bool for `CosmosDbNoSQLDataSourceOptions`, we could error out with appropriate message?

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+                                {
+                                    dataSource = dataSource with { Options = null };
+                                    break;
+                                }
+                                else
+                                {
+                                    Dictionary<string, JsonElement> optionsDict = new();
+                                    while (reader.Read() && reader.TokenType is not JsonTokenType.EndObject)
+                                    {
+                                        string key = reader.GetString()!;
```
seantleonard, line 79: Does `key` represent a property within the `options` object? if so, would be more descriptive to have var name be `optionsSubproperty` or something like that

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
abhishekkumams: the role of serializer is just to correctly translate the values from the config. This test is checking that we support different ways in which options can be provided.

abhishekkumams: if options are provided that is not required for a db, it will be ignored, but serializer role is just to translate the values correctly.

src/Config/Converters/DataSourceConverterFactory.cs
abhishekkumams: The role of the serializer is simply to correctly deserialize the config containing env values. All the checks on the reuqired options for a specific db is already done separately. we don't have to have it at multiple places.

src/Config/Converters/DataSourceConverterFactory.cs
abhishekkumams: The role of the serializer is simply to correctly deserialize the config containing env values. All the checks on the reuqired options for a specific db is already done separately. we don't require to have it at multiple places.

src/Config/Converters/DataSourceConverterFactory.cs
abhishekkumams: yep, updated the name to be more descriptive.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -550,12 +550,12 @@ public async Task ValidateInheritanceOfReadPermissionFromAnonymous()
 }";
             GraphQLRuntimeOptions graphqlOptions = new(Enabled: true);
             RestRuntimeOptions restRuntimeOptions = new(Enabled: false);
-            Dictionary<string, JsonElement> dbOptions = new();
+            Dictionary<string, object> dbOptions = new();
```
Aniruddh25, line 553:
```suggestion
            Dictionary<string, object?> dbOptions = new();
```

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -379,6 +490,147 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
...
+              ""entities"": {}
+            }
+            ";
+
+        /// <summary>
+        /// Returns a datasource option for a given database type.
+        /// MSSQL options has a boolean value, CosmosDB options has string value,
+        /// MySQL datasource option is empty, PostgreSQL datasource option is null,
+        /// and DWSQL has no options.
+        /// The role of serializer is just to correctly translate the values from the config.
```
Aniruddh25, line 530:
```suggestion
        /// The role of deserializer is just to correctly translate the values from the config.
```

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -301,8 +413,7 @@ public static string GetModifiedJsonString(string[] reps, string enumString)
   },
   ""data-source"": {
     ""database-type"": " + enumString + @",
-    ""connection-string"": ""server=dataapibuilder;database=" + reps[++index % reps.Length] + @";uid=" + reps[++index % reps.Length] + @";Password=" + reps[++index % reps.Length] + @";"",
-    ""resolver-config-file"": """ + reps[++index % reps.Length] + @"""
+    ""connection-string"": ""server=dataapibuilder;database=" + reps[++index % reps.Length] + @";uid=" + reps[++index % reps.Length] + @";Password=" + reps[++index % reps.Length] + @";""
```
Aniruddh25, line 416: if we would have modified this string to include `options` property with an environment variable, would it have been sufficient for testing?

Aniruddh25 (commented):
> Data source section was not supporting env transformation.

Only the options section within the data-source section was unsupported, right? wasnt env transformation for other properties like connection-string still supported? 

If that's the case, please fix the PR description.

Aniruddh25 (approved):
Left few questions, but LGTM otherwise.

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -0,0 +1,136 @@
...
+                                {
+                                    Dictionary<string, object?> optionsDict = new();
+                                    while (reader.Read() && reader.TokenType is not JsonTokenType.EndObject)
+                                    {
+                                        string optionsSubproperty = reader.GetString()!;
+                                        reader.Read();
+                                        object? optionsSubpropertyValue;
+                                        if (reader.TokenType is JsonTokenType.String)
+                                        {
+                                            string stringValue = reader.DeserializeString(replaceEnvVar: _replaceEnvVar)!;
```
seantleonard, line 84: comment right here would be helpful pointing out that this is where you determine whether to resolve the environment variable or keep as-is

seantleonard (approved):
one or two comment related items, othewise, looks good, thanks for addressing feedback and clarifying other comments and code :)

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
abhishekkumams: it would have been, but since there are so many different cases with data-source, i didn't wanted to merge it with it.

src/Service.Tests/CosmosTests/MutationTests.cs
abhishekkumams: In this test method, we are not storing null values, so not required. Also it gives warning if i do so.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
abhishekkumams: i created new tests since there are different cases that needs to be tested for the data-source section. So, separating tests made more sense to debug if any failures occur because of future changes.

Issue #1967, opened 2024-01-12
Change default data source name for Hot Reload


aaronburtle:
We are currently assigning a unique default data source name to each `RuntimeConfig` that we create. However, for a Hot Reload scenario this is problematic because we do not want to be updating this value after the Hot Reload, and so, it can not be uniquely generated each time.

Instead this value should be created in a deterministic way, ie: taking a hash of the connection string.

seantleonard:
> because we do not want to be updating this value after the Hot Reload

need to explain why in the issue so motivation behind this change is clear.

seantleonard:
Not moving this task at the moment, this fix seems necessary even without remaining HR features in this milestone.

Issue #1966, opened 2024-01-12
Refactor RuntimeConfigProvider to be more extensible for Hot Reloading


aaronburtle:
The current design is to have a single RuntimeConfigProvider class, which stores the RuntimeConfig internally. However, with the addition of hot-reload we need to take into account that a RuntimeConfig is constructed through a different process in Hosted and Local scenarios. To better facilitate hot-reloading, we will move the RuntimeConfig object over to the RuntimeConfigLoader. The RuntimeConfigLoader will therefore be responsible for maintaining the RuntimeConfig object which is consistent with the configuration provided by the user. This means that we need to move the hot-reload functions over to this class, as they are central to initiating the process for deservialization of the new RuntimeConfig object. This will help us to be certain that we do not mix the implementation details of the two scenarios and for example try to monitor a file in the local file system when we are in a hosted scenario.

seantleonard:
Moving this + hot reload related tasks to post GA -> tracking via 1.2rc milestone.

The current PR can be moved to draft or closed for the time being until we revisit this feature post-GA.

seantleonard:
Moving some hot reload tasks to backlog so we can focus on tasks that add hotreload support various services/classes (openapi, authentication config). Once the pre-reqs are done, it will be easier to start updating the runtime config classes directly.

PR #1981, opened 2024-01-18
Refactor RuntimeConfig into Interface with Local and Hosted Providers

aaronburtle:
## Why make this change?

-  Related to https://github.com/Azure/data-api-builder/issues/67
- Closes to https://github.com/Azure/data-api-builder/issues/1966
- Closes to https://github.com/Azure/data-api-builder/issues/1967

In order to better facilitate Hot Reload, we need to split the responsibility for loading and accessing the `RuntimeConfig` between the two different scenarios that our service can run within, local and hosted. We already have a `FileSystemRuntimeConfigLoader` which handles the loading of the config from a local file system. This refactor naturally extends that approach so that we can Hot Reload both in the local and eventually, the hosted scenario.

We also make some necessary changes for hot reloading to work within the service.

## What is this change?

As a part of the changes, we split the RuntimeConfigProvider into a `LocalRuntimeConfigProvider` and a `HostedRuntimeConfigProvider`. These classes then implement the `IRuntimeConfigProvider` interface in the way that makes sense for their scenario.

Rather than having these providers storing the internal copy of the `RuntimeConfig`, the natural approach is for the loaders to be responsible for maintaining this internal representation, and for the loaders to handle all of the loading, including any hot reloading that takes place for that `RuntimeConfig`.

This was previously not possible, because we could not store a copy of the `RuntimeConfig` with the loader due to circular dependency. This refactor removes that issue as part of its changes. We then move the internally stored `RuntimeConfig` to these loaders, as well as all of the logic for Hot Reloading. Since we are only handling Hot Reloading in the local scenario to being, we only have the loader for the local scenario currently setup to do this. When we extend Hot Reloading to the hosted scenario, that will mean adding a loader that can handle that specific logic in a hosted scenario.

We also change the way that the default data source names are generated so that this is deterministic. This is a requirement to pass validation during time of handling requests. In this case we are taking a hash of the connection string, though this means to hot reload the connection string we will need a new approach.

Within the tests, there was an attempt to use the `LocalRuntimeConfigProvider` when possible, since testing is naturally happening in a local scenario. However, some tests utilize functions that are only relevant for the `HostedRuntimeConfigProvider` and so in those cases, we used that class. This created some issues, and workarounds were found for these tests.

## How was this tested?

Current test suite.

## Sample Request(s)

Any valid request.

### DIFF
--- a/src/Config/ConfigFileWatcher.cs
+++ b/src/Config/ConfigFileWatcher.cs
@@ -1,8 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Azure.DataApiBuilder.Config.ObjectModel;
+
 namespace Azure.DataApiBuilder.Config;
 
+/// <summary>
+/// This class is responsible for monitoring the config file from the
+/// local file system, and if changes are detected, triggering the
+/// required logic to begin a hot reload scenario.
+/// </summary>
 public class ConfigFileWatcher
 {
     private FileSystemWatcher? _fileWatcher;
@@ -33,7 +40,7 @@ public class ConfigFileWatcher
         {
             if (_configLoader is null)
             {
-                throw new ArgumentNullException("_configProvider can not be null.");
+                throw new ArgumentNullException("ConfigLoader can not be null.");
             }
 
             if (_configLoader.RuntimeConfig is null)
@@ -43,7 +50,7 @@ public class ConfigFileWatcher
 
             if (_configLoader.RuntimeConfig.IsDevelopmentMode())
             {
-                _configLoader.HotReloadConfig();
+                _configLoader.HotReloadConfig(_configLoader.RuntimeConfig.DefaultDataSourceName);
             }
         }
         catch (Exception ex)
--- a/src/Config/ConfigFileWatcher.cs
+++ b/src/Config/ConfigFileWatcher.cs
@@ -1,14 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Azure.DataApiBuilder.Core.Configurations;
+namespace Azure.DataApiBuilder.Config;
 
 public class ConfigFileWatcher
 {
     private FileSystemWatcher? _fileWatcher;
-    private readonly LocalRuntimeConfigProvider? _configProvider;
+    private FileSystemRuntimeConfigLoader _configLoader;
 
-    public ConfigFileWatcher(LocalRuntimeConfigProvider configProvider, string directoryName, string configFileName)
+    public ConfigFileWatcher(FileSystemRuntimeConfigLoader configLoader, string directoryName, string configFileName)
     {
         _fileWatcher = new FileSystemWatcher(directoryName)
         {
@@ -16,7 +16,7 @@ public class ConfigFileWatcher
             EnableRaisingEvents = true
         };
 
-        _configProvider = configProvider;
+        _configLoader = configLoader;
         _fileWatcher.Changed += OnConfigFileChange;
     }
 
@@ -31,14 +31,19 @@ public class ConfigFileWatcher
     {
         try
         {
-            if (_configProvider is null)
+            if (_configLoader is null)
             {
                 throw new ArgumentNullException("_configProvider can not be null.");
             }
 
-            if (!_configProvider.IsLateConfigured && _configProvider.GetConfig().IsDevelopmentMode())
+            if (_configLoader.RuntimeConfig is null)
             {
-                _configProvider.HotReloadConfig();
+                throw new ArgumentNullException("RuntimeConfig can not be null.");
+            }
+
+            if (_configLoader.RuntimeConfig.IsDevelopmentMode())
+            {
+                _configLoader.HotReloadConfig();
             }
         }
         catch (Exception ex)
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -8,6 +8,7 @@ using System.Reflection;
 using System.Text.Json;
 using Azure.DataApiBuilder.Config.ObjectModel;
 using Azure.DataApiBuilder.Service.Exceptions;
+using Microsoft.Extensions.Logging;
 
 namespace Azure.DataApiBuilder.Config;
 
@@ -91,11 +92,16 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
 
     /// <summary>
     /// Checks if we have already attempted to configure the file watcher, if not
-    /// instantiate the file watcher if we are in the correct mode.
-    /// Returns true if we instantiate a file watcher.
+    /// instantiate the file watcher if we are in the development mode.
+    /// Returns true if we instantiate a new file watcher.
     /// </summary>
     private bool TrySetupConfigFileWatcher()
     {
+        if (_configFileWatcher is not null)
+        {
+            return false;
+        }
+
         if (RuntimeConfig is not null && RuntimeConfig.IsDevelopmentMode())
         {
             try
@@ -122,28 +128,40 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// <param name="config">The loaded <c>RuntimeConfig</c>, or null if none was loaded.</param>
     /// <param name="replaceEnvVar">Whether to replace environment variable with its
     /// value or not while deserializing.</param>
+    /// <param name="logger">ILogger for logging errors.</param>
     /// <returns>True if the config was loaded, otherwise false.</returns>
     public bool TryLoadConfig(
         string path,
         [NotNullWhen(true)] out RuntimeConfig? config,
-        bool replaceEnvVar = false)
+        bool replaceEnvVar = false,
+        ILogger? logger = null,
+        string? defaultDataSourceName = null)
     {
         if (_fileSystem.File.Exists(path))
         {
             Console.WriteLine($"Loading config file from {path}.");
             string json = _fileSystem.File.ReadAllText(path);
-            bool res = TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
+            bool configParsed = TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
             TrySetupConfigFileWatcher();
             config = RuntimeConfig;
-            return res;
+            if (RuntimeConfig is not null && defaultDataSourceName is not null)
+            {
+                RuntimeConfig.DefaultDataSourceName = defaultDataSourceName;
+            }
+
+            return configParsed;
+        }
+
+        string errorMessage = $"Unable to find config file: {path} does not exist.";
+        if (logger is null)
+        {
+            Console.Error.WriteLine(errorMessage);
         }
         else
         {
-            // Unable to use ILogger because this code is invoked before LoggerFactory
-            // is instantiated.
-            Console.WriteLine($"Unable to find config file: {path} does not exist.");
+            logger.LogError(message: errorMessage);
         }
-
+        
         config = null;
         return false;
     }
@@ -155,7 +173,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// <param name="replaceEnvVar">Whether to replace environment variable with its
     /// value or not while deserializing.</param>
     /// <returns>True if the config was loaded, otherwise false.</returns>
-    public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false)
+    public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false, string? defaultDataSourceName = null)
     {
         return TryLoadConfig(ConfigFilePath, out config, replaceEnvVar);
     }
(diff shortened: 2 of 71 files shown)


src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -104,7 +132,10 @@ public string GetConfigFileName()
         {
             Console.WriteLine($"Loading config file from {path}.");
             string json = _fileSystem.File.ReadAllText(path);
-            return TryParseConfig(json, out config, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
+            bool res = TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
```
seantleonard, line 135: nit: more descriptive var name.

seantleonard (changes requested):

src/Core/Configurations/LocalRuntimeConfigProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// Attempt to acquire runtime configuration metadata.
+    /// </summary>
+    /// <param name="runtimeConfig">Populated runtime configuration, if present.</param>
+    /// <returns>True when runtime config is provided, otherwise false.</returns>
+    public bool TryGetConfig([NotNullWhen(true)] out RuntimeConfig? runtimeConfig)
+    {
+        if (ConfigLoader.RuntimeConfig is null)
+        {
+            ConfigLoader.TryLoadKnownConfig(out _, replaceEnvVar: true);
+
```
ayush3797, line 78: nit: remove extra line

src/Config/ConfigFileWatcher.cs
```diff
@@ -1,22 +1,22 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Azure.DataApiBuilder.Core.Configurations;
+namespace Azure.DataApiBuilder.Config;
 
 public class ConfigFileWatcher
```
ayush3797, line 6: Since we are touching this code now, can you add a class summary to this class?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,6 +89,32 @@ public string GetConfigFileName()
         return configFileName;
     }
 
+    /// <summary>
+    /// Checks if we have already attempted to configure the file watcher, if not
```
ayush3797, line 93: Does it actually check if the config file watcher was instantiated previously? Not following from the code.

src/Config/FileSystemRuntimeConfigLoader.cs
ayush3797, line 135: Not something changed in this PR: In the TryParseConfig() method, the order of parameters is messed up. Could you please fix that?

Aniruddh25 (changes requested):
Posting comments so far

Aniruddh25 (changes requested):
Good starter, but needs clarity on when would `HostedRuntimeConfigProvider` will be created so that previous scenarios are not broken.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -104,7 +132,10 @@ public string GetConfigFileName()
...
             Console.WriteLine($"Loading config file from {path}.");
             string json = _fileSystem.File.ReadAllText(path);
-            return TryParseConfig(json, out config, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
+            bool res = TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
+            TrySetupConfigFileWatcher();
+            config = RuntimeConfig;
+            return res;
         }
         else
         {
```
ayush3797, line 141: The code inside else block can be moved out of it and the else block can be removed.

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: The function below? Seems correct to me.

image

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -169,7 +169,7 @@ public RuntimeConfig(string? Schema, DataSource DataSource, RuntimeEntities Enti
         this.DataSource = DataSource;
         this.Runtime = Runtime;
         this.Entities = Entities;
-        _defaultDataSourceName = Guid.NewGuid().ToString();
+        _defaultDataSourceName = DataSource.ConnectionString.GetHashCode().ToString();
```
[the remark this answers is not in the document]
aaronburtle: we can pass the Guid along through the hot reload function and then overwrite the RuntimeConfig's default data source name with that passed along Guid if it exists. I updated this, so we can avoid using the connection string's hash.

src/Core/Configurations/HostedRuntimeConfigProvider.cs
```diff
@@ -10,45 +10,34 @@
...
-    public Dictionary<string, string?> ManagedIdentityAccessToken { get; private set; } = new Dictionary<string, string?>();
-
-    public RuntimeConfigLoader ConfigLoader { get; private set; }
+    public Dictionary<string, string?> ManagedIdentityAccessToken { get; set; } = new Dictionary<string, string?>();
 
-    private ConfigFileWatcher? _configFileWatcher;
+    public RuntimeConfigLoader ConfigLoader { get; set; }
 
-    private RuntimeConfig? _runtimeConfig;
+    public RuntimeConfig? _runtimeConfig;
```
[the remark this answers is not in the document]
aaronburtle: There were tests that in order to make work with the new refactor need these to be public for the mocking that was required. I could not find a straight forward way to get those tests to work without making this public, but perhaps we could sync on this.

src/Core/Configurations/HostedRuntimeConfigProvider.cs
```diff
@@ -10,45 +10,34 @@
...
 
-    public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
+    public HostedRuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
     {
         ConfigLoader = runtimeConfigLoader;
```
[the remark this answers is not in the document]
aaronburtle: It is used for testing because our tests are using the initialize function, which is for the hosted scenario. I am looking into how to refactor the tests to eliminate this, but the answer is not obvious.

src/Config/FileSystemRuntimeConfigLoader.cs
seantleonard: perhaps Ayush meant that most other methods in our project have the `out` parameter as the last parameter in the method signature.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -169,18 +169,18 @@ public RuntimeConfig(string? Schema, DataSource DataSource, RuntimeEntities Enti
         this.DataSource = DataSource;
         this.Runtime = Runtime;
         this.Entities = Entities;
-        _defaultDataSourceName = Guid.NewGuid().ToString();
+        DefaultDataSourceName = Guid.NewGuid().ToString(); //DataSource.ConnectionString.GetHashCode().ToString();
```
seantleonard, line 172: commented code

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -130,7 +130,7 @@ public bool AllowIntrospection
         }
     }
 
-    private string _defaultDataSourceName;
+    public string DefaultDataSourceName;
```
seantleonard, line 133: This seems to be only public member not following getter setter syntax. `{ get; set; }` any reason not to?

src/Config/ObjectModel/RuntimeConfig.cs
seantleonard: also seems inconsistent where all other members are set using `this.Member` = Parameter

src/Config/ConfigFileWatcher.cs
```diff
@@ -31,14 +38,19 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
...
             {
-                throw new ArgumentNullException("_configProvider can not be null.");
+                throw new ArgumentNullException("RuntimeConfig can not be null.");
             }
 
-            if (!_configProvider.IsLateConfigured && _configProvider.GetConfig().IsDevelopmentMode())
+            if (_configLoader.RuntimeConfig.IsDevelopmentMode())
             {
-                _configProvider.HotReloadConfig();
+                _configLoader.HotReloadConfig(_configLoader.RuntimeConfig.DefaultDataSourceName);
```
seantleonard, line 53: may be helpful to print/log that hot reloading was skipped  because config definded production mode.

src/Config/ConfigFileWatcher.cs
```diff
@@ -31,14 +38,19 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
     {
         try
         {
-            if (_configProvider is null)
+            if (_configLoader is null)
```
seantleonard, line 41: summary comment needs completion for params.

also nit in the summary: `this trigger function` -> `this event handler`

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,32 +90,78 @@ public string GetConfigFileName()
...
+        string? defaultDataSourceName = null)
     {
         if (_fileSystem.File.Exists(path))
         {
             Console.WriteLine($"Loading config file from {path}.");
             string json = _fileSystem.File.ReadAllText(path);
-            return TryParseConfig(json, out config, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
+            bool configParsed = TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
+            TrySetupConfigFileWatcher();
+            config = RuntimeConfig;
```
seantleonard, line 146: Can you clarify these few lines here?

Why would we still attempt to execute `TrySetupConfigFileWatcher()` if `configParsed` is false indicating some failure?

Also, the line `config=Runtimeconfig` wouldn't work or shouldn't occur if `configParsed` is false.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -124,11 +173,20 @@ public string GetConfigFileName()
...
+    public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false, string? defaultDataSourceName = null)
     {
         return TryLoadConfig(ConfigFilePath, out config, replaceEnvVar);
     }
 
+    /// <summary>
+    /// Hot Reloads the runtime config when the file watcher
+    /// is active and detects a change to the underlying config file.
+    /// </summary>
+    public void HotReloadConfig(string defaultDataSourceName)
```
seantleonard, line 185: why is the param named `defaultDataSourceName`? what happens when the value is not from the default data source and, instead, from a different data source?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -22,6 +22,8 @@ public abstract class RuntimeConfigLoader
 {
     protected readonly string? _connectionString;
 
+    public RuntimeConfig? RuntimeConfig { get; set; }
```
seantleonard, line 25: should this class be renamed to `RuntimeConfigLoaderBase`?

src/Core/Configurations/HostedRuntimeConfigProvider.cs
```diff
@@ -66,93 +56,25 @@ public RuntimeConfig GetConfig()
...
-
-        return false;
-    }
-
-    /// <summary>
-    /// Attempt to acquire runtime configuration metadata.
-    /// </summary>
-    /// <param name="runtimeConfig">Populated runtime configuration, if present.</param>
-    /// <returns>True when runtime config is provided, otherwise false.</returns>
+    // Not currently used in hosted scenario, but needed for interface
```
seantleonard, line 68: comments need to explain why:
 `TryGetConfig([NotNullWhen(true)] out RuntimeConfig? runtimeConfig)`
 `public bool TryGetLoadedConfig([NotNullWhen(true)] out RuntimeConfig? runtimeConfig)`

 are not used in hosted scenario. It doesn't make sense to me that these two methods are promoted to the interface if they aren't used in implementating classes.

src/Core/Configurations/HostedRuntimeConfigProvider.cs
```diff
@@ -278,22 +200,22 @@ public void HotReloadConfig()
...
+    //        {
+    //            configLoadedTasks.Add(configLoadedHandler(this, _runtimeConfig));
+    //        }
+    //    }
+
+    //    bool[] results = await Task.WhenAll(configLoadedTasks);
+
+    //    // Verify that all tasks succeeded.
+    //    return results.All(x => x);
+    //}
```
seantleonard, line 218: commented code.

seantleonard (changes requested):

seantleonard:
The code base has changed significantly since this PR was opened. Closing so that you can resolve conflicts and refresh the design/approach before reopening.

src/Core/Configurations/HostedRuntimeConfigProvider.cs
```diff
@@ -278,7 +197,7 @@ public void HotReloadConfig()
         return false;
     }
 
-    private async Task<bool> InvokeConfigLoadedHandlersAsync()
+    public async Task<bool> InvokeConfigLoadedHandlersAsync()
```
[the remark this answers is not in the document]
aaronburtle: this function is invoked when we initialize in a hosted scenario. From the initialize function we have

122     bool configLoadSucceeded = await InvokeConfigLoadedHandlersAsync();

PR #1990, opened 2024-01-24
Table name collision bug fix

### LINKED ISSUE #1908: [Bug]: DAB seems to have issues with tables in different schemas that have the same name (SQL Server)
### What happened?

I have two tables with the same name, in two different schemas, in the same database:
schema1.tablename and 
schema2.tablename
I add them both to the "entities" section in dab-config.json, like this:

  "schema1_tablename": {
    "source": "schema1.tablename",
    "permissions": [
      {
        "actions": ["*"],
        "role": "anonymous"
      }
    ]
  },
  "schema2_tablename": {
    "source": "schema2.tablename",
    "permissions": [
      {
        "actions": ["*"],
        "role": "anonymous"
      }
    ]
  },

When I try to query the first using REST via /api/schema1_tablename, I get the right answer: a list of the rows.
When I try to query the second using REST via /api/schema2_tablename I get an error 500 that says "Invalid Column Name".  The column it names is a valid column in the schema1 table, but not in the schema2 table.  

If I reverse the order of the entities in the dab-config.json, then I can query the schema2 table, but not the schema1 table.  So, even though in the configuration I am providing the full name of the table using schema.tablename format, it doesn't seem capable of distinguishing between tables that have the sane tablename, even if the schema is different.

### Version

Azure SQL

REST

### Relevant log output

_No response_

### Code of Conduct

aaronburtle:
## Why make this change?

Closes #1908

## What is this change?

We were previously filling the schema information for the entity data set but only using the table name. We add in the schema name to deconflict between two tables with the same name in different schemas. This means that we retrieve information from the entity data set by using schema and tablename when schema is available, and just the table name otherwise.

The reason that we do not need to worry about collisions at the schema and table name level (eg. where we have the same schema and table names within different databases) is because this entity data set is a part of the SqlMetadataProvider, and we have a SqlMetadataProvider for each database (not database type but actual DBs).

## How was this tested?

A regression test with a collision between table names in different schemas was added. We also update the unit test for SqlMetadatProvider.

## Sample Request(s)

see #1908


src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1240,7 +1240,7 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
             string schemaName,
             string tableName)
         {
-            DataTable? dataTable = EntitiesDataSet.Tables[tableName];
+            DataTable? dataTable = EntitiesDataSet.Tables[$"{schemaName}.{tableName}"];
```
Aniruddh25, line 1243: For MySQL, schemaName will be empty so this will turn into ".tablename", do we need a separate implementation of this function for MySQL similar to how

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -127,6 +127,22 @@ public virtual async Task FindOnTableWithUniqueCharacters()
                 sqlQuery: GetQuery("FindOnTableWithUniqueCharacters"));
         }
 
+        /// <summary>
+        /// Tests the Rest Api to validate that queries work
+        /// when there are is the same table name in two different
```
Aniruddh25, line 132:
```suggestion
        /// when there is the same table name in two different
```

Aniruddh25 (changes requested):
Thanks for the quick fix! Need to make sure MySQL doesn't regress

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1240,12 +1240,14 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
             string schemaName,
             string tableName)
         {
-            DataTable? dataTable = EntitiesDataSet.Tables[tableName];
+            // MySQL does not use schema
+            string schemaAndTableName = _databaseType is DatabaseType.MySQL ? tableName : $"{schemaName}.{tableName}";
```
Aniruddh25, line 1244: The string `{queryPrefix}{SqlQueryBuilder.QuoteIdentifier(tableName)}` obtained in FillSchemaForTableAsync seems to be the right way to generate the complete schema name/table name, why dont we move that into a helper and reuse it here and in `FillSchemaForTableAsync`?
That should handle MySQL automatically, since it checks for string.IsNullOrEmpty...

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1366,7 +1369,7 @@ private async Task ValidateDatabaseConnection()
                 = $"SELECT * FROM {queryPrefix}{SqlQueryBuilder.QuoteIdentifier(tableName)}";
             adapterForTable.SelectCommand = selectCommand;
 
-            DataTable[] dataTable = adapterForTable.FillSchema(EntitiesDataSet, SchemaType.Source, tableName);
+            DataTable[] dataTable = adapterForTable.FillSchema(EntitiesDataSet, SchemaType.Source, schemaAndTableName);
```
Aniruddh25, line 1372: if we set `schemaAndTableName = {queryPrefix}{SqlQueryBuilder.QuoteIdentifier(tableName)}`, we need not add a new argument to this function.

Aniruddh25 (changes requested):
Need to reuse existing methods

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
aaronburtle: modified the helper function `GetTablePrefix` into `GetTableNameWithPrefix` and included the database name as part of the string we return, so that if we ever have a single meta data provider per data source name we will de-conflict collisions between schema and table names that are the same but in different data bases.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1240,7 +1245,8 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
             string schemaName,
             string tableName)
         {
-            DataTable? dataTable = EntitiesDataSet.Tables[tableName];
+            string tableNameWithPrefix = GetTableNameWithPrefix(schemaName, tableName);
```
severussundar, line 1248: Would this also take care of other object types like views and stored procedures too? It would good to test out similar scenario - views, stored procedures with same names in different schemas.

severussundar (commented):
Few things to consider:

1.  `SourceDefinition` class has a method called `FullName` which returns `schemaName.tableName`. A new method can be introduced that returns `DBname.SchemaName.objectName`.  

Advantages with introducing this method 
              a) This would cover all database object types.
              b) Logic of determining the full name would not get repeated at different places. 
`sourceDefinition.FullName()` and `sourceDefinition.FullNameIncludingDbName()` (just example method name) can be used instead of determining the complete name multiple times.
3. This issue was reported for REST Get APIs. But, would be good to test this for other types like `POST`, `PUT`, etc.
4. This can be an opportunity to validate similar scenarios with GraphQL (This can be a separate issue/PR)

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
aaronburtle: This is a good point. I will test this out, and create an issue to track. For this PR I want to avoid increasing the scope too much since adding in additional coverage may complicate the testing scenarios in ways I can't anticipate right now and don't want to create a blocker. I will add a link to the issue to this PR's description when I have it created.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -119,6 +122,8 @@ private void HandleOrRecordException(Exception e)
             }
 
             ConnectionString = runtimeConfig.GetDataSourceFromDataSourceName(dataSourceName).ConnectionString;
+            SqlConnectionStringBuilder connectionStringBuilder = new(ConnectionString);
```
Aniruddh25, line 125: What happens if the ConnectionString is invalid, and this throws an exception?

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -120,9 +120,16 @@ public void CheckTablePrefix(string databaseName, string schemaName, string expe
...
-            SqlMetadataProvider<SqlConnection, SqlDataAdapter, SqlCommand> provider = new MsSqlMetadataProvider(runtimeConfigProvider, queryManagerFactory.Object, sqlMetadataLogger, runtimeConfigProvider.GetConfig().GetDefaultDataSourceName());
-            string tableprefix = provider.GetTablePrefix(databaseName, schemaName);
-            Assert.AreEqual(tableprefix, expectedPrefix);
+            SqlMetadataProvider<SqlConnection, SqlDataAdapter, SqlCommand> provider = new MsSqlMetadataProvider(
+                runtimeConfigProvider,
+                queryManagerFactory.Object,
+                sqlMetadataLogger,
+                runtimeConfigProvider.GetConfig().GetDefaultDataSourceName())
+            {
+                DatabaseName = databaseName
```
Aniruddh25, line 129: This is not really testing the new code that you added in the constructor - to deduce the database name using ConnectionStringBuilder. Can you set this databaseName in the connection string property of the `runtimeConfigProvider` in some other ways like string replace on `InitialCatalog` for e.g.?

Aniruddh25 (changes requested):
Need to fix the test for databasename

seantleonard (commented):
need updated pr description based on review feedback:

>We were previously filling the schema information for the entity data set but only using the table name. 

- Where? and for only certain functions? Shyam mentioned that there are certain functions that provide full name vs just table name.

> We add in the schema name to deconflict between two tables with the same name in different schemas.

- where? in single location or in many places that need to use the full name that includes schema?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1360,26 +1366,25 @@ private async Task ValidateDatabaseConnection()
...
             adapterForTable.SelectCommand = selectCommand;
 
-            DataTable[] dataTable = adapterForTable.FillSchema(EntitiesDataSet, SchemaType.Source, tableName);
+            DataTable[] dataTable = adapterForTable.FillSchema(EntitiesDataSet, SchemaType.Source, tableNameWithPrefix);
             return dataTable[0];
         }
 
-        internal string GetTablePrefix(string databaseName, string schemaName)
-        {
+        internal string GetTableNameWithPrefix(string schemaName, string tableName)
```
seantleonard, line 1378: need comment summarizing function + params. also explain what is meant by "prefix"

seantleonard: and return value: what is an example of the final string?
- `[schema].[tableName]`
- `'schema'.'tableName'`
- `schema.tableName`

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1396,7 +1401,8 @@ internal string GetTablePrefix(string databaseName, string schemaName)
                 tablePrefix.Append(schemaName);
             }
 
-            return tablePrefix.ToString();
+            string queryPrefix = string.IsNullOrEmpty(tablePrefix.ToString()) ? string.Empty : $"{tablePrefix}.";
+            return $"{queryPrefix}{SqlQueryBuilder.QuoteIdentifier(tableName)}";
```
seantleonard, line 1405: comments will help me understand this better. what was the output of this function before your change, and what is the output of the function after your change?

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -127,6 +127,22 @@ public virtual async Task FindOnTableWithUniqueCharacters()
                 sqlQuery: GetQuery("FindOnTableWithUniqueCharacters"));
         }
 
+        /// <summary>
+        /// Tests the Rest Api to validate that queries work
+        /// when there is the same table name in two different
+        /// schemas.
+        /// </summary>
+        /// <returns></returns>
```
seantleonard, line 135:
```suggestion
```

src/Service.Tests/SqlTests/RestApiTests/Find/MySqlFindApiTests.cs
```diff
@@ -1023,6 +1023,13 @@ public override string GetQuery(string key)
             return _queryMap[key];
         }
 
+        [TestMethod]
+        [Ignore]
+        public override Task FindOnTableWithNamingCollision()
+        {
+            throw new NotImplementedException();
+        }
```
seantleonard, line 1031: need comment explaining why this isn't an issue for MySql

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -127,6 +127,22 @@ public virtual async Task FindOnTableWithUniqueCharacters()
                 sqlQuery: GetQuery("FindOnTableWithUniqueCharacters"));
         }
 
+        /// <summary>
+        /// Tests the Rest Api to validate that queries work
+        /// when there is the same table name in two different
```
seantleonard, line 132: need to explain which two tables share a name and what are the two different schema values?

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -127,6 +127,22 @@ public virtual async Task FindOnTableWithUniqueCharacters()
...
+        /// schemas.
+        /// </summary>
+        /// <returns></returns>
+        [TestMethod]
+        public virtual async Task FindOnTableWithNamingCollision()
```
seantleonard, line 137: what is the expected behavior of "FindOnTableWithNamingCollision" 
-> (Actionable Exception/error communicated to user) error is caught and properly messaged to user? 
-> (Gracefully handled) no error because you now use the schema name when making comparisons?

src/Service.Tests/SqlTests/SqlTestBase.cs
```diff
@@ -101,6 +101,13 @@ public abstract class SqlTestBase
                 _ => TestHelper.AddMissingEntitiesToConfig(runtimeConfig, "magazine", "foo.magazines"),
             };
 
+            // Add table name collision testing entity to the config
```
seantleonard, line 104: explain why not for mysql

src/Service.Tests/SqlTests/SqlTestBase.cs
```diff
@@ -101,6 +101,13 @@ public abstract class SqlTestBase
...
+            runtimeConfig = DatabaseEngine switch
+            {
+                TestCategory.MYSQL => runtimeConfig,
+                _ => TestHelper.AddMissingEntitiesToConfig(runtimeConfig, "bar_magazine", "bar.magazines", new string[] { "upc" })
```
seantleonard, line 108: need named parameters , assuming `new string[] { "upc" }` represents a column name?

Any reason not to add this to the test config files directly and, if necessary, make it a duplicate of an entity in the test config file(s)? I understand magazine is already added as a "missing entity", but we may not need to continue that pattern as it maintains entities in two separate places.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -98,15 +98,15 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
 
         /// <summary>
         /// <code>Do: </code> Tests with different combinations of connection string to ensure
-        /// tableprefix generated is correctly.
-        /// <code>Check: </code> Making sure table prefix matches expected prefix.
+        /// tableprefix generated is correct and forms the correct table name with said prefix.
```
seantleonard, line 101: provide an example of what "correct" means and what would be "incorrect"

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
seantleonard: A good test would be an integration tests where you test the scenario described by #1908 and ensure that error/messaging does not occur and results from both tables are returned as expected.

src/Service.Tests/DatabaseSchema-DwSql.sql
```diff
@@ -235,6 +245,7 @@ VALUES (1, 'Star Trek', 'SciFi', NULL), (2, 'Cinderella', 'Tales', 3001),(3,'Ún
 (5, 'Snow White', 'AnotherTales', 3001);
 
 INSERT INTO [foo].[magazines](id, title, issue_number) VALUES (1, 'Vogue', 1234), (11, 'Sports Illustrated', NULL), (3, 'Fitness', NULL);
+INSERT INTO [bar].[magazines](upc, comic_name, issue) VALUES (0, '0', 0);
```
seantleonard, line 248: super nit to not just use an int as a sample name.
```suggestion
INSERT INTO [bar].[magazines](upc, comic_name, issue) VALUES (0, 'MyComic', 0);
```

seantleonard (changes requested):

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -119,6 +122,8 @@ private void HandleOrRecordException(Exception e)
             }
 
             ConnectionString = runtimeConfig.GetDataSourceFromDataSourceName(dataSourceName).ConnectionString;
+            SqlConnectionStringBuilder connectionStringBuilder = new(ConnectionString);
+            DatabaseName = connectionStringBuilder.InitialCatalog;
```
severussundar, line 126: For MsSQL, it looks like `InitialCatalog` would always get populated. For other SQL database types, does it get populated always (it can be the case, but need to go over the docs and confirm that)

_A slightly different approach to consider for fetching database name:_

`DatabaseName` property is used inside the function `GetTableWithSchemaFromDataSetAsync`. In that function, there is a call to a method `FillSchemaForTableAsync`. Inside there, a connection to the database is opened, post that, the  
database name is fetched from the connection object using `conn.Database` L:1364. When database name is extracted this way (after successfully connecting to the actual database), we can be sure that a) Connection string is valid b) Database specific behavior (if any) w.r.t to how connection string is interpreted is not going to have any effect.

aaronburtle: I initially wanted to use this approach, but it means opening up a database connection only to parse out the database name, which is not very efficient. Correctly and carefully parsing the connection string to pull out the database name should be a lot faster and also entirely synchronous.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
aaronburtle: > A good test would be an integration tests where you test the scenario described by #1908 and ensure that error/messaging does not occur and results from both tables are returned as expected.

i added an entity and table with collisions that match the scenario in #1908, they are used in a new case in the Find Api Tests, `FindOnTableWithNamingCollision`, this is just fixing the unit test that we already had to make sure it matches the new format.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
aaronburtle: Previously, this function was returning the prefix for a given table. So that would either be the [schema], or an empty string. Now it returns the prefix along with the table name, so the entire correctly formatted table name with prefix. So either


[databaseName].[schemaName].[tableName]

or

[schemaName].[tableName]

or

[tableName]

depending on what information is available.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
aaronburtle: This test validates that the query works despite there being a naming collision, so gracefully handled now because we are using the schema.

src/Service.Tests/SqlTests/SqlTestBase.cs
aaronburtle: This is because MySql handles schemas differently than the rest of the the DB types that we have. We can not have the kinds of different schemas in our config file that is included for MySql. Instead we include a config file that represents the intersection between what is valid for all DB types, and then modify it here. This avoids having to maintain multiple config files for testing.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
Aniruddh25: Why are you still setting the DatabaseName here 
`{
                DatabaseName = databaseName`
}` ? Cant you let the constructor do the job given that you are replacing the connectionstring accurately on line 119 above?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
severussundar: Since, there is going to be a separate SqlMetadataProvider class for every database, it should be fine to not include the database name. Distinguishing the backing objects based on schema name + table/view/stored-procedure name should be sufficient. 

Since, this entire thread is focused on diferent approaches for fetching the database names, this is not applicable anymore. So, resolving this thread.

severussundar (approved):
LGTM, suggestions to validate few more scenarios as part of a different issue/PR.
1. Validating and adding tests for the same bug but with views and stored procedures
2. Validating and adding tests for the same bug with GraphQL

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
aaronburtle: As we discussed on teams, it turns out that we will never have more than one database per sqlmetadataprovider so we don't need to include database name in this way.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1360,26 +1364,38 @@ private async Task ValidateDatabaseConnection()
...
-        {
+        /// <summary>
+        /// Gets the correctly formatted table name with prefix, given the provided schema name
+        /// and table name. A prefix is everything that comes before the table name itself.
+        /// For example, the correct format with database, schema, and table name is of the form
+        /// [DatabaseName].[schemaName].[tableName]
+        /// In the case that the DatabaseName is empty, the correct form is
+        /// [schemaName].[tableName]
+        /// In the case that both the DatabaseName and schemaName are empty the correct form is
+        /// [tableName]
```
seantleonard, line 1384: Doesn't this case negate the purpose of the function? "GetTableNameWithPrefix" ? this case results in no prefix being returned?
>         /// In the case that both the DatabaseName and schemaName are empty the correct form is
>        /// [tableName]

Also, is this a usecase somewhere in our codebase? What would be the purpose of only passing in a table name just to get that value back? is it because this function is adding the "quoteIdentifiers"?
e.g. input: tablename: "myTable"
output: `[myTable]`

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1240,7 +1240,11 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
             string schemaName,
             string tableName)
         {
-            DataTable? dataTable = EntitiesDataSet.Tables[tableName];
+            // Because we have an instance of SqlMetadataProvider for each individual database
+            // (note: this means each actual database not each database type), we do not
+            // need to worry about collisions beyond that schema, hence no database name is needed.
+            string tableNameWithPrefix = GetTableNameWithPrefix(string.Empty, schemaName, tableName);
```
seantleonard, line 1246: named params.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -97,20 +97,24 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
         }
 
         /// <summary>
-        /// <code>Do: </code> Tests with different combinations of connection string to ensure
-        /// tableprefix generated is correctly.
-        /// <code>Check: </code> Making sure table prefix matches expected prefix.
+        /// <code>Do: </code> Tests with different combinations of data base, schema, and table names
```
seantleonard, line 100:
```suggestion
        /// <code>Do: </code> Tests with different combinations of database, schema, and table names
```

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -127,6 +127,23 @@ public virtual async Task FindOnTableWithUniqueCharacters()
                 sqlQuery: GetQuery("FindOnTableWithUniqueCharacters"));
         }
 
+        /// <summary>
+        /// Tests the Rest Api to validate that queries work
+        /// when there is the same table name in two different
+        /// schemas. In this test we have two table both
```
seantleonard, line 133:
```suggestion
        /// schemas. In this test we have two tables both
```

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -97,20 +97,24 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
...
+        /// mean [TestDB].[model].[TrainedModel], and any other form would be incorrect.
+        /// <code>Check: </code> Making sure table name with prefix matches expected name with prefix.
         /// </summary>
         [DataTestMethod]
-        [DataRow("", "", "")]
-        [DataRow("", "model", "[model]")]
-        [DataRow("TestDB", "", "[TestDB]")]
-        [DataRow("TestDB", "model", "[TestDB].[model]")]
-        public void CheckTablePrefix(string databaseName, string schemaName, string expectedPrefix)
+        [DataRow("", "", "", "[]")]
```
seantleonard, line 107: is a table name allowed to be an empty string? in the `SqlMetadataProvider.GetTableNameWithPrefix(...)` function, there is a "IsNullOrEmpty" check for dbname and schema, but not for table name. Also, should we not expect GetTableNameWithPrefix to return a prefix, even with an empty table name?

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -127,6 +127,23 @@ public virtual async Task FindOnTableWithUniqueCharacters()
...
+        /// named magazines but with one in the schema "foo" and
+        /// the other in the schema "bar".
+        /// </summary>
+        [TestMethod]
+        public virtual async Task FindOnTableWithNamingCollision()
+        {
+            await SetupAndRunRestApiTest(
```
seantleonard, line 140: Do you  need to verify a request against the second entity? and ensure results are as expected from expected database type?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
aaronburtle: It is because of how MySql deals with schema, we can have an empty schema for that DB Type. You can see that we even assign the default schema name as the empty string for MySql in the MySqlMetadatProvider.

image


I guess we could change the name if you think returning an empty prefix is confusing, but I thought with the summary it was pretty clear that the prefix can just be empty in certain cases.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
aaronburtle: Table names are not allowed to be empty in sql as far as I know, but this is just a unit test that is checking if the formatting is done correctly. I did not create these cases originally, but modified them to fit the bug fix. I am apprehensive to remove these cases, even though I don't think theyre valid for actual use in SQL.


I can change the name if you don't like the idea of a prefix being empty.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
aaronburtle: I think just what we are doing should be fine since this demonstrates that we correctly navigate the collision and get the expected results from the request on a tablename with a collision.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: Gotcha.
> And yes in the case that everything except table name is empty you get back the table name formatted with []

Sounds like empty table name is a special circumstance that we handle by returning `[]` and not `[].[].[]` indicating that all passed in params were empty.

seantleonard (approved):

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1361,26 +1368,38 @@ private async Task ValidateDatabaseConnection()
                 Connection = conn
             };
 
-            string tablePrefix = GetTablePrefix(conn.Database, schemaName);
-            string queryPrefix = string.IsNullOrEmpty(tablePrefix) ? string.Empty : $"{tablePrefix}.";
+            string tableNameWithPrefix = GetTableNameWithPrefix(conn.Database, schemaName, tableName);
```
Aniruddh25, line 1371: For a connection string that has a databasename, conn.Database will not be empty, so tableNameWithPrefix will be for books e.g. `[testdb].[dbo].[books]` for mssql. The EntitiesDataSet dictionary will store the DataTable with this as the key. Subsequently, when you are trying to retrieve in the function above, you are looking up by `[dbo].[books]` How will the lookup succeed?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1241,7 +1241,14 @@ public static bool IsGraphQLReservedName(Entity entity, string databaseColumnNam
...
         {
-            DataTable? dataTable = EntitiesDataSet.Tables[tableName];
+            // Because we have an instance of SqlMetadataProvider for each individual database
+            // (note: this means each actual database not each database type), we do not
+            // need to worry about collisions beyond that schema, hence no database name is needed.
+            string tableNameWithPrefix = GetTableNameWithPrefix(
+                databaseName: string.Empty,
+                schemaName: schemaName,
+                tableName: tableName);
+            DataTable? dataTable = EntitiesDataSet.Tables[tableNameWithPrefix];
```
Aniruddh25, line 1251: if the key is not found, I think this will throw exception. Instead of directly indexing the Dictionary in this manner, I think we should use `TryGet`

Aniruddh25 (changes requested):
Needs to handle exceptions, and correct lookups

Aniruddh25:
Looks good to merge once unit tests succeed in pipeline and merged with main.

Aniruddh25 (approved):
Looks good to merge, tests that include database name in the connection string can be added optionally.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
aaronburtle: DataTable doesn't have a `TryGet`, and if the key is not found it simply returns an empty data table.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
aaronburtle: I have totally removed the need for database name now. This was succeeding in our tests before due to lack of a database name being explicitly used in the test pipeline.

PR #1995, opened 2024-01-30
add current version number

yorek:
## Why make this change?

Add a default version that is aligned with what being built to be able to more easily identify the correct version of a dab application connected to MSSQL (or another supported database)

## What is this change?

The change adds current version information to the project files.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Notes

With this change, querying MSSQL to find the connected DAB instances will return also correct version information:

```sql
select session_id, host_name, program_name, client_interface_name 
from sys.dm_exec_sessions where is_user_process = 1 and program_name like '%dab%'
```

result:

image

### DIFF
--- a/src/Cli/Cli.csproj
+++ b/src/Cli/Cli.csproj
@@ -29,6 +29,8 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <NoWarn>NU1603</NoWarn>
+    <Version>0.11</Version>
+    <FileVersion>0.11</FileVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
--- a/src/Config/Azure.DataApiBuilder.Config.csproj
+++ b/src/Config/Azure.DataApiBuilder.Config.csproj
@@ -7,6 +7,8 @@
     <OutputPath>$(BaseOutputPath)\engine</OutputPath>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <NoWarn>NU1603</NoWarn>
+    <Version>0.11</Version>
+    <FileVersion>0.11</FileVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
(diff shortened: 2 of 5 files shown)


abhishekkumams:
adding the version in project file will override the version that is generated automatically in build. Therefore, it makes things more manual. we need to find some better way.

Aniruddh25:
> adding the version in project file will override the version that is generated automatically in build. Therefore, it makes things more manual. we need to find some better way.

build automatically generates only the patch number. Major/Minor versions are hard coded in the build pipeline, and since `FileTransform` will only be useful for automated builds, I guess the only way to have these versions in local cloned builds is adding them in the csproj files as well.

Aniruddh25 (approved):

Aniruddh25 (commented):
There seems to be some unit tests in the CLI project that are failing which still are expecting 1.0.0 version.

abhishekkumams:
> > adding the version in project file will override the version that is generated automatically in build. Therefore, it makes things more manual. we need to find some better way.
> 
> build automatically generates only the patch number. Major/Minor versions are hard coded in the build pipeline, and since `FileTransform` will only be useful for automated builds, I guess the only way to have these versions in local cloned builds is adding them in the csproj files as well.

We can't merge this PR, before updating the `FileTransform` task, because the version hardcoded in csproj file will take precedence over the nuget pack command that we have in build pipeline. this will create an issue as it won't use the generated version.

seantleonard:
> > > adding the version in project file will override the version that is generated automatically in build. Therefore, it makes things more manual. we need to find some better way.
> > 
> > 
> > build automatically generates only the patch number. Major/Minor versions are hard coded in the build pipeline, and since `FileTransform` will only be useful for automated builds, I guess the only way to have these versions in local cloned builds is adding them in the csproj files as well.
> 
> We can't merge this PR, before testing the `FileTransform` task, because the version hardcoded in csproj file will take precedence over the nuget pack command that we have in build pipeline. this will create an issue as it won't use the generated version.

What specific steps are needed beyond ensuring existing pipelines pass? What logic does ` nuget pack command that we have in build pipeline.` have and can we update that to accommodate this change?

seantleonard:
closing because this functionality was fixed in #2086

PR #2003, opened 2024-02-01
Enhanced claims processing for MS SQL Set session context

seantleonard:
# Improved Claims Handling for DB Policies and MS SQL Session Context

## Why make this change?

Initially reported in #1957, DAB does not gracefully handle instances where the provided JWT token has claims with value type JSON Array.

Two negative impacts:
1. Developers who don't override the default MS SQL `set-session-context` value of `true` to `false` will observe that requests fail for tokens that fit the above criteria. 
2. Developers who write database policies with `@claims.claimType` references will see requests fail when the claimType referenced fits the above criteria. 

**Note:** The issue raised in #1957 is unique because the user may be using a historical version of Duende's IdentityServer which emits token scopes in a `scope` claim here the value is a JSON array. That format differs from Entra ID which emits tokens scopes in the 'scp' claim whose value is a string of values delimited by spaces. Reference: [Entra ID Access Token Claims Reference](https://learn.microsoft.com/entra/identity-platform/access-token-claims-reference#payload-claims).

### 1. Prevents `@claims.claimType` references in a DB policy from failing a request when the claimType has a value type of JSON array.

**Example:** 
Given the database policy: `@claims.groups eq @item.groupid`, DAB would previously (before this PR's changes) fail a request if the provided access token had >1 group because the `groups` claim is a JSON string array in the JWT token which dotnet resolves into multiple `Claim` objects. (one claim per group where claimType is `groups` and value is `groupGUID`. The request failure no longer occurs.

**How is this implemented?** 
DAB now recognizes when multiple `Claim` objects exist for a single claim type. Given the policy `@claims.groups eq @item.groupid` DAB will replace `@claims.groups` with the first instance of the claimType `groups` that DAB finds. If DAB has resolved two `groups` claim objects such as `["groupGUID1", "groupGUID2"]`, DAB would only resolve the first it finds: `groupGUID1`. 

When/if implemented, issue #2004 addresses this behavior by adding the OData operator `in` so that the query predicate is generated to be `([dbo].[groupid] in ('tokenGroupGUID1', 'tokenGroupGUID2'))`.

### 2. Prevents multiple instances of a claimType from failing a request utilizing MSSQL's `set-session-context` feature

**How?**
Aggregates multiple `Claim` objects of the same `claimType` (claim name) into a JSON array serialized into a string.  The original value type in the JWT JSON array (bool, int, string) is preserved. DAB uses the serialized JSON as the value for a session context variable whose key is `claimType`.

When dotnet processes a JWT token, claims whose values are JSON arrays will be split into distinct `Claim` objects where `claimType` is the claim name and `value` is one of the values in the array. `Claim` objects are created for each object in the array. DAB uses the session context feature to pass token claims and claim values to the database as session context variables.

#### This is not a breaking change.
This is not a breaking change because access tokens that had claims which didn't result in DAB failing the request will still have the claim values passed as is -> scalar values. This is because usable tokens didn't contain claims with JSON arrays as values. 

A breaking change would be modifying this behavior to pass the scalar by its original type as present in the JWT token by setting `DbConnectionParam.DbType` explicitly when creating `DbConnectionParam`.

#### How can developers write security policies in SQL? 
When the value of a session context variable is a serialized JSON string containing JSON array, the value can be processed using the following tsql functionality:
- `JSON_QUERY`(SQL Server 2016(13.X) and later, SQL MI, Azure SQL Database, Azure Synapse Analytics)
- `JSON_VALUE` (SQL Server 2016(13.X) and later, SQL MI, Azure SQL Database, Azure Synapse Analytics)
- `JSON_ARRAY` (SQL Server 2022 (16.X), Azure SQL Database)

#### Example JWT token that is now handled without error 
```json
{
  "aud": "00000003-0000-0000-c000-000000000000",
  "iss": "https://sts.windows.net/25fd4421-0fff-4ed6-ad2f-c2ca00ed7207/",
  "iat": 1706642510,
  "acr": "1",
  "int_array": [1,2,3],
  "bool_array": [true, false, true],
  "groups":"src1",
  "_claim_sources":{
     "src1" : { "endpoint" : "https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects" }},
  "wids": ["d74b8d81-39eb-4201-bd9f-9f1c4011e3c9","18d14519-c4da-4ad4-936d-9a2de69d33cf","9e513fc0-e8af-43b1-a6c7-949edb1967a3"],
  "roles": ["anonymous", "authenticated", "myCustomRole"],
  "scp": "email openid profile User.Read",
  "scope": ["idServerScope1", "idServerScope2"]
}
```

#### Example tsql created by dab for session context

The following SQL shows how an Entra ID `scp` claim is still passed as a space delimited string and
how a `roles` claim, a JSON array in the JWT token, is properly serialized into a JSON array string and passed to MS SQL session context without the request failing:

Assumption: `x-ms-api-role` header value is `authenticated`
```sql
EXEC sp_set_session_context 'roles', @session_param0, @read_only = 1;
EXEC sp_set_session_context 'scp', @session_param1, @read_only = 1;
EXEC sp_set_session_context 'scopes', @session_param2, @read_only = 1;

SELECT TOP 1 [dbo_books].[id] AS [id], [dbo_books].[title] AS [title], [dbo_books].[publisher_id] AS [publisher_id] FROM [dbo].[books] AS [dbo_books] WHERE [dbo_books].[id] = @param0 ORDER BY [dbo_books].[id] ASC FOR JSON PATH, INCLUDE_NULL_VALUES,WITHOUT_ARRAY_WRAPPER

@param0 int,
@param1 nvarchar(2),
@param2 nvarchar(5),
@param3 nvarchar(12),
@session_param0 nvarchar(13),
@session_param1 nvarchar(37)
@session_param2 nvarchar(44)

@param0=1,
@param1=N'id',
@param2=N'title',
@param3=N'publisher_id',
@session_param0=N'["Authenticated"]',
@session_param1=N'GraphQL.ReadWrite REST.EndpointAccess'
@session_param2=N'["GraphQL.ReadWrite", "REST.EndpointAccess"]'
```

#### Matrix of how jwt claims are process by DAB and dotnet for session context

`Dictionary<string, string> GetProcessedUserClaims(HttpContext? context)` is populated based on the following rules:

| Jwt ClaimName      | JwtClaimValue                                                                             |
|--------------------|-------------------------------------------------------------------------------------------|
| scp                | openid profile customScope                                                                |
| scopes (idserver4) | ["openid", "profile","customScope"]                                                       |
| intArrayClaim      | [1,2,3]                                                                                   |
| boolArrayClaim     | [true, false, true]                                                                       |
| nullValueClaim     | null                                                                                      |
| _claim_sources     | {"src1": {"endpoint":"https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects"}} |

| DotNetProcessed Name                         | DotNet Processed Value                                                                    |
|----------------------------------------------|-------------------------------------------------------------------------------------------|
| scp                                          | openid profile customScope                                                                |
| scopes </br> scopes </br> scopes                         | openid </br> profile </br> customScope                                                                |
| intArrayClaim </br> intArrayClaim </br> intArrayClaim    | 1 </br> 2 </br> 3                                                                                     |
| boolArrayClaim </br> boolArrayClaim </br> boolArrayClaim | true </br> false </br> true                                                                           |
| nullValueClaim                               | //empty string                                                                            |
| _claim_sources                               | {"src1": {"endpoint":"https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects"}} |

| ProcessedValueForSessionCtx                                                               | Remarks                                                                                                                                                              |
|-------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| openid profile customScope                                                                | scp claim from Entra ID access token                                                                                                                                 |
| ["openid", "profile", "customScope"                                                       | Non-Entra ID identity providers may emit a "scopes" claim as a JSON string array. Dotnet processes the array into individual scope claims and doesn't retain the space delimited string. |
| [1,2,3]                                                                                   |                                                                                                                                                                      |
| [true, false, true]                                                                       |                                                                                                                                                                      |
|                                                                                           |                                                                                                                                                                      |
| {"src1": {"endpoint":"https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects"}} |                                                                                                                                                                      |

## Tests

-[x] Unit tests
-[x] Integration tests


src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -965,10 +970,10 @@ public void ParseValidDbPolicy(string policy, string expectedParsedPolicy)
         [DataRow(ClaimValueTypes.UInteger64,    "18446744073709551615",               true, DisplayName = "ulong")]
         [DataRow(ClaimValueTypes.Double,        "12.34",                              true, DisplayName = "decimal")]
         [DataRow(ClaimValueTypes.Double,        "12.345",                             true, DisplayName = "double")]
-        [DataRow(JsonClaimValueTypes.JsonNull,  "null",                               true, DisplayName = "Json null literal")]
+        [DataRow(Microsoft.IdentityModel.JsonWebTokens.JsonClaimValueTypes.JsonNull,  "null",                               true, DisplayName = "Json null literal")]
```
abhishekkumams, line 973: is this required since we already imported it above?

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1210,6 +1194,202 @@ public void ValidateClientRoleHeaderClaimIsAddedToClaimsInRequestContext()
             Assert.IsTrue(claimsInRequestContext.Count == 1);
             Assert.IsTrue(claimsInRequestContext.ContainsKey(AuthenticationOptions.ROLE_CLAIM_TYPE));
             Assert.IsTrue(TEST_ROLE.Equals(claimsInRequestContext[AuthenticationOptions.ROLE_CLAIM_TYPE].Value));
+        }*/
+
+        /// <summary>
+        /// JWT token JSON payloads may not be flat and may contain nested JSON objects or arrays.
```
abhishekkumams, line 1200: is this test method not required? why is it commented out?

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
seantleonard: thanks for the feedback. updated to not have namespace. confirmed via intellisense.

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
seantleonard: Updated to remove duplicate. but uncommented other relevant test and updated comments.

Aniruddh25:
> Entra ID emits the scope claim as 'scp' whose value is a string of values separated by spaces as described in the[Entra ID Access Token Claims Reference](https://learn.microsoft.com/entra/identity-platform/access-token-claims-reference#payload-claims).
> Access token scopes should be emitted as a single string where values are space-delimited per [RFC 6749 - The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3)

Was this mentioned as a Note in the PR description to mean that EntraID follows the RFC correctly but Duende's Identity Server doesn't?

Aniruddh25:
> How are claims processed in the DAB engine now?

`now` is relative here. Do you now with your change or now in 0.10.23?

Also, in the PR description Im unclear at which point have you started explaining what the changes are. The `What` heading is missing

Aniruddh25 (commented):
In the example table, did you mean the JwtToken Claim value `["openid", "profile","customScope"]` is processed into `["openid", "profile","customScope"]` as the set session context value?

Aniruddh25 (commented):
Posting comments so far, test files remaining

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1204,12 +1190,167 @@ public void ValidateClientRoleHeaderClaimIsAddedToClaimsInRequestContext()
...
+            Assert.AreEqual(expected: "[1,2,3]", actual: claimsInRequestContext["int_array"]);
+            Assert.AreEqual(expected: "[true,false,true]", actual: claimsInRequestContext["bool_array"]);
+            Assert.AreEqual(expected: @"{""src1"":{""endpoint"":""https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects""}}", actual: claimsInRequestContext["_claim_sources"]);
+            Assert.AreEqual(expected: "1706816426", actual: claimsInRequestContext["iat"]);
+            Assert.AreEqual(expected: "", actual: claimsInRequestContext["nullValuedClaim"]);
+        }
+
+        /// <summary>
+        /// JWT token JSON payloads may not be flat and may contain nested JSON objects or arrays.
+        /// This test validates that when dotnet (authentication jwt processing code) flattens the JWT token payload,
```
Aniruddh25, line 1260: nit: this sentence `This test validates that when dotnet (authentication jwt processing code) flattens the JWT token payload,` is incomplete

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1204,12 +1190,167 @@ public void ValidateClientRoleHeaderClaimIsAddedToClaimsInRequestContext()
...
+        /// This test also protects against regression in current functionality because
+        /// this scenario works today. If we ignored any claim that *may* result in duplicate claims using the same name,
+        /// we would be making a breaking change.
+        /// DAB's Database Policies and Session context features utilize access token claims and those features
+        /// can utilize the claims that occur only once.
```
Aniruddh25, line 1265: These two comments seem opposing:
/// If we ignored any claim that *may* result in duplicate claims using the same name,
/// we would be making a breaking change.
=> implies we should Not be ignoring any claim that results in duplicate claims

/// DAB's Database Policies and Session context features utilize access token claims and those features
/// can utilize the claims that occur only once.
=> says that we are unable to process claims that occur multiple times -> essentially, we will be ignoring those that appear multiple times...

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1204,12 +1190,167 @@ public void ValidateClientRoleHeaderClaimIsAddedToClaimsInRequestContext()
...
+            context.Setup(x => x.User).Returns(principal);
+            context.Setup(x => x.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER]).Returns(TEST_ROLE);
+
+            // Act
+            Dictionary<string, string> claimsInRequestContext = AuthorizationResolver.GetProcessedUserClaims(context.Object);
+
+            // Assert
+            Assert.AreEqual(expected: authenticatedUserclaims.Count, actual: claimsInRequestContext.Count, message: "Only two claims should be present.");
+            Assert.AreEqual(expected: "openid", actual: claimsInRequestContext["scp"], message: "Unexpected scp claim returned.");
+            Assert.AreEqual(expected: TEST_ROLE, actual: claimsInRequestContext[AuthenticationOptions.ROLE_CLAIM_TYPE], message: "Unexpected roles claim returned.");
```
Aniruddh25, line 1352: `roles` claim seems to be very special -> because it will always have a single role even if the there were 2 authenticated identities. So, this test doesnt seem to point out the distinction between authenticated v/s unauthenticated. Can we check a regular claim `foo`, if it gets added twice like a JSON array?

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1282,6 +1423,56 @@ private static RuntimeConfig BuildTestRuntimeConfig(EntityPermission[] permissio
...
+            {
+                Dictionary<string, object> distributedClaimSources = new()
+                {
+                    ["src1"] = new { endpoint = "https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects" }
+                };
+
+                userClaims = new()
+                {
+                    { "scope", "scope1 scope2 scope3" },
+                    { "groups", "src1" },
```
Aniruddh25, line 1450: `groups` isnt an array here as the function comment expects.

Aniruddh25 (changes requested):
Some questions on what we really expect for `roles` claim

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -456,38 +553,43 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+            // which represents the role resolved in the x-ms-api-role http header.
+            // DAB authentication middleware validates the x-ms-api-role value.
             if (!claimsInRequestContext.ContainsKey(AuthenticationOptions.ROLE_CLAIM_TYPE) &&
                 identity.HasClaim(type: AuthenticationOptions.ROLE_CLAIM_TYPE, value: clientRoleHeader))
             {
-                claimsInRequestContext.Add(AuthenticationOptions.ROLE_CLAIM_TYPE, new Claim(AuthenticationOptions.ROLE_CLAIM_TYPE, clientRoleHeader, ClaimValueTypes.String));
-            }
+                List<Claim> roleClaim = new()
+                {
+                    new Claim(AuthenticationOptions.ROLE_CLAIM_TYPE, clientRoleHeader, ClaimValueTypes.String)
```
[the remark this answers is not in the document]
seantleonard: yes

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -456,38 +553,43 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+                };
 
-            // If identity is not authenticated, we don't honor any other claims present in this identity.
-            if (!identity.IsAuthenticated)
-            {
-                continue;
+                claimsInRequestContext.Add(AuthenticationOptions.ROLE_CLAIM_TYPE, roleClaim);
```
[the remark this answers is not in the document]
seantleonard: updated comment to reflect that this is the expected behavior. To not be a breaking change for set-session-context/db policies, when AuthorizationResolver resolves claims in this method, it only resolves the role claim whose value matches the client role header value.

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,14 +440,109 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+                        processedClaims.Add(claimName, value: JsonSerializer.Serialize(claimValues.Select(claim => int.Parse(claim.Value))));
+                        break;
+                    case ClaimValueTypes.Double:
+                        processedClaims.Add(claimName, value: JsonSerializer.Serialize(claimValues.Select(claim => double.Parse(claim.Value))));
+                        break;
+                    case ClaimValueTypes.String:
+                    case JsonClaimValueTypes.JsonNull:
+                    case JsonClaimValueTypes.Json:
+                    default:
+                        string json = JsonSerializer.Serialize(claimValues.Select(claim => claim.Value));
```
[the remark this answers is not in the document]
seantleonard: Code now does not have any code specific to these two claim types.

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,14 +440,109 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+        foreach ((string claimName, List<Claim> claimValues) in userClaims)
+        {
+            // Special case to accommodate historical versions of Duende IdentityServer which emitted
+            // the scope claim as a JSON array in the access token.
+            // Entra ID emits the scope claim as 'scp' whose value is a string of values separated by spaces
+            // as described in the Entra ID Access Token Claims Reference
+            // When >1 scope/scp claim exists, DAB resolves the values into a single string where scopes are delimited by spaces.
+            // https://learn.microsoft.com/entra/identity-platform/access-token-claims-reference#payload-claims
+            // https://docs.duendesoftware.com/identityserver/v5/apis/aspnetcore/authorization/#scope-claim-format
+            if ((claimName == "scp" || claimName == "scope") && claimValues.Count > 1)
```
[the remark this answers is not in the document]
seantleonard: I no longer have specific logic for these claim types.

Updated the behavior to be (some slight variance to what i sent you in email):
- `scp` claim from Entra ID is a string w/ values delimited by spaces, and dotnet DOES NOT split this up into different `Claim` objects. When populating session context variables, `scp` will be of type "string" and DAB will preserve that. Developers must write appropriate tsql code in security policy to parse out the scopes and that expectation is the same as before this PR. (If i change this, it would be a breaking change).
- `scopes` claim from other identity providers may come as a JSON array. While this never worked prior to this PR, DAB now handles this appropriately and preserves the JSON array when setting session context variable value. While this deviates from my email where we'd want similar behavior between all identity providers, I have to avoid creating a breaking change and will document that users must consult their identity provider docs to know how to process scopes in a tsql security policy.

Comments and PR description updated to reflect this.

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,14 +440,109 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+            {
+                StringBuilder spaceDelimitedClaimPayload = new();
+                for (int i = 0; i < claimValues.Count - 1; i++)
```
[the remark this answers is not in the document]
seantleonard: No limit and `Count` is type `int`.

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1176,8 +1160,10 @@ public void ParsePolicyWithDuplicateUserClaims(bool exceptionExpected, params st
         }
 
         /// <summary>
-        /// Test to validate the AuthorizationResolver.GetAllUserClaims() successfully adds role claim to the claimsInRequestContext dictionary.
-        /// Only the role claim corresponding to the X-MS-API-ROLE header is added to the claimsInRequestContext.
+        /// Test to validate the AuthorizationResolver.GetAllAuthenticatedUserClaims() successfully adds role claim to the claimsInRequestContext dictionary.
+        /// Only one "roles" claim is added to the dictionary claimsInRequestContext and only one Claim is added to the list maintained
+        /// for the key "roles" where the value corresponds to the X-MS-API-ROLE header.
+        /// e.g [key: "roles"] -> [value: List (Claim) { Claim(type: "roles", value: "{x-ms-api-role value") }]
```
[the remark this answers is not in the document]
seantleonard: Updated PR description.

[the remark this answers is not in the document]
seantleonard: only single role will be present, one that matches x-ms-api-role header.

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
seantleonard: updated. 

```csharp
        /// JWT token JSON payloads may not be flat and may contain nested JSON objects or arrays.
        /// This test validates that when dotnet's JWT processing code flattens the JWT token payload
        /// into Claim objects, the AuthorizationResolver.GetProcessedUserClaims(...) method correctly
        /// returns a dictionary where key=claimType and value= scalar value or serialized JSON array.
        /// This test enforces that DAB only processes one 'roles' claim whose value matches the x-ms-api-role
        /// header, which protects against regression.
```

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
seantleonard: Clarified by removing conflicting statement that were effectively outdated. The only claims that are ignored are `roles` claims that do not match the x-ms-api-role header. All other claims are retained and processed as a scalar string or a JSON array serialized to a string.

```
        /// JWT token JSON payloads may not be flat and may contain nested JSON objects or arrays.
        /// This test validates that when dotnet's JWT processing code flattens the JWT token payload
        /// into Claim objects, the AuthorizationResolver.GetProcessedUserClaims(...) method correctly
        /// returns a dictionary where key=claimType and value= scalar value or serialized JSON array.
        /// This test enforces that DAB only processes one 'roles' claim whose value matches the x-ms-api-role
        /// header, which protects against regression.
```

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
seantleonard: Updated. added an `oid` claim to both identities and asserted that the oid claim value used for the unauthenticated identity was NOT resolved. Also updated the test for 'roles' claim (even if special) to validate that the unauthenticated roles claim was not used instead of checking that the authenticated role claim was used. While essentially the same test, the updated version aligns with the test name explicitly.

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
seantleonard: updated comment to no longer state 'groups' is an array. it is covered by being described as a "distributed groups claim" and "json object"

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1058,64 +1063,43 @@ public void ParseInvalidDbPolicyWithUserNotPossessingAllClaims(string policy)
...
             AuthorizationResolver authZResolver = AuthorizationHelpers.InitAuthorizationResolver(runtimeConfig);
             Mock<HttpContext> context = new();
 
             // Add identity to the readAction, updateAction.
             ClaimsIdentity identity = new(TEST_AUTHENTICATION_TYPE, TEST_CLAIMTYPE_NAME, AuthenticationOptions.ROLE_CLAIM_TYPE);
-            foreach (string claimType in claimTypes)
-            {
-                identity.AddClaim(new Claim(type: claimType, value: defaultClaimValue, ClaimValueTypes.String));
-            }
+            identity.AddClaim(new Claim(type: "scope", value: "profile", ClaimValueTypes.String));
```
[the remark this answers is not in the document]
seantleonard: Yes, good call. Added test with 'groups' claim.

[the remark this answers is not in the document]
seantleonard: because a 'groups' claim is a json array, processed by dotnet as multiple `Claim` objects.

seantleonard:
> In the example table, did you mean the JwtToken Claim value ["openid", "profile","customScope"] is processed into ["openid", "profile","customScope"] as the set session context value?

yes. Also updated the table to be more readable by splitting up into three tables. 

> "How are claims processed in the DAB engine now?"
> `now` is relative here. Do you now with your change or now in 0.10.23?
> Also, in the PR description Im unclear at which point have you started explaining what the changes are. The `What` heading is missing

updated to mean that "now" meant "with this change, here's the new behavior"

> "Entra ID emits the scope claim as 'scp' whose value is a string of values separated by spaces as described in the[Entra ID Access Token Claims Reference](https://learn.microsoft.com/entra/identity-platform/access-token-claims-reference#payload-claims).
> Access token scopes should be emitted as a single string where values are space-delimited per [RFC 6749 - The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3)"
> 
> Was this mentioned as a Note in the PR description to mean that EntraID follows the RFC correctly but Duende's Identity Server doesn't?

I removed references to Duende Identity Server because I found additional identity providers (Okta) that may emit scopes as a "scopes" claim and uses a JSON array. DAB doesn't care which way scopes are presented, however they are presented, they will be forwarded in their original format to set-session-context. No need to say whether one way is right or wrong now.

src/Core/Authorization/AuthorizationResolver.cs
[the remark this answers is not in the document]
seantleonard: also removed this special handling of scp/scope so that DAB is no longer creating this space separated string. That space separated string is only preserved when using EntraID . so no need to recreated.

your question still applies to max count though. JWT tokens issue by identity providers are capped by max payload of HTTP header. If I were to put a cap on claim count, I'm uncertain what would be a "best" value without unintentionally truncating data a developer expects to be present.  For example, the `groups` claim is capped at Entra ID at 250 groups, before the groups claim values are replaced with a "Distributed claim" which means the value is a URL to microsoft graph where the app can query for a list of groups because the number was too large to include in a token.

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,61 +440,150 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
-                /*
-                 * An example claim would be of format:
-                 * claim.Type: "user_email"
-                 * claim.Value: "authz@microsoft.com"
-                 * claim.ValueType: "string"
-                 */
-                // At this point, only add non-role claims to the collection and only throw an exception for duplicate non-role claims.
-                if (!claim.Type.Equals(AuthenticationOptions.ROLE_CLAIM_TYPE) && !claimsInRequestContext.TryAdd(claim.Type, claim))
+                // 'roles' claim has already been processed.
+                if (claim.Type == AuthenticationOptions.ROLE_CLAIM_TYPE)
```
severussundar, line 574: super nit: `claim.Type.Equals(AuthenticationOptions.ROLE_CLAIM_TYPE)`

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,61 +440,150 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
             {
-                /*
-                 * An example claim would be of format:
-                 * claim.Type: "user_email"
-                 * claim.Value: "authz@microsoft.com"
-                 * claim.ValueType: "string"
-                 */
-                // At this point, only add non-role claims to the collection and only throw an exception for duplicate non-role claims.
-                if (!claim.Type.Equals(AuthenticationOptions.ROLE_CLAIM_TYPE) && !claimsInRequestContext.TryAdd(claim.Type, claim))
+                // 'roles' claim has already been processed.
```
severussundar, line 573: nit: Having a more detailed explanation (like you have in your PR description) would make the code easier to understand.

For my understanding: Since you have already created a list with a value for `roles` claims type in the previous code block, you are ignoring other values for that same claim type(Only the first encountered value is honored), right?

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -525,13 +616,19 @@ private static string GetPolicyWithClaimValues(string policy, Dictionary<string,
...
+        if (claimsInRequestContext.TryGetValue(claimType, out List<Claim>? claims))
         {
-            return GetClaimValue(claim);
+            // Database policies do not support operators like "in" or "contains".
+            // Return the first value in the list of claims.
+            // This is not a breaking change since historically,
+            // if the user had >1 role AND wrote a db policy to include the token '@claims.role',
+            // DAB would fail the request. Now, the request won't fail, but the value
+            // resolved is the first claim encountered (when there are multiple claim instances).
+            return GetClaimValue(claims.First());
```
severussundar, line 631: nit: Add a `null` check for `claims` before using `.First()` method

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,61 +440,150 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+            {
+                // Remaining claims will be collected as string scalar values.
+                // While Claim.ValueType may indicate the token value was not a string (int, bool),
+                // resolving the actual type here would be a breaking change because 
+                // DAB has historically sent single instance claims as value type string.
+                // This block also accommodates Entra ID access tokens to avoid a breaking change because
+                // the 'scp' claim is a space delimited string which is not broken up into separate claim objects
+                // by dotnet. The Entra ID 'scp' claim should be passed to MSSQL's session context as-is.
+                // https://learn.microsoft.com/entra/identity-platform/access-token-claims-reference#payload-claims
+                processedClaims.Add(claimName, value: claimValues[0].Value);
```
severussundar, line 514: nit: Is there a chance that the `claimValues` array is empty when this else block executes?

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,61 +440,150 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+        Dictionary<string, List<Claim>> userClaims = GetAllAuthenticatedUserClaims(context);
+
+        foreach ((string claimName, List<Claim> claimValues) in userClaims)
+        {
+            // Some identity providers (other than Entra ID) may emit a 'scope' claim as a JSON string array. dotnet will
+            // create claim objects for each value in the array. DAB will honor that format
+            // and processes the 'scope' claim objects as a JSON array serialized to a string.
+            if (claimValues.Count > 1)
+            {
+                switch (claimValues.First().ValueType)
```
severussundar, line 480: nit: Is it possible for the `claimValues` to be an empty list here? In that, `.First()` would throw an exception

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,61 +440,150 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+                {
+                    case ClaimValueTypes.Boolean:
+                        processedClaims.Add(claimName, value: JsonSerializer.Serialize(claimValues.Select(claim => bool.Parse(claim.Value))));
+                        break;
+                    case ClaimValueTypes.Integer:
+                    case ClaimValueTypes.Integer32:
+                    case ClaimValueTypes.Integer64:
+                    case ClaimValueTypes.UInteger32:
+                    case ClaimValueTypes.UInteger64:
+                        processedClaims.Add(claimName, value: JsonSerializer.Serialize(claimValues.Select(claim => int.Parse(claim.Value))));
```
severussundar, line 490: Here, would performing `int.Parse()` take care of all the cases as opposed to `Int32.Parse()`, `Int64.Parse()` etc.

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
```diff
@@ -98,5 +106,106 @@ public static class WebHostBuilderHelper
...
+                                        // The signing key must match
+                                        ValidateIssuerSigningKey = true,
+                                        IssuerSigningKey = key,
+                                        // Lifetime
+                                        ValidateLifetime = true,
+                                        // Instructs the asp.net core middleware to use the data in the "roles" claim for User.IsInRole()
+                                        // See https://learn.microsoft.com/en-us/dotnet/api/system.security.claims.claimsprincipal.isinrole?view=net-6.0#remarks
+                                        RoleClaimType = AuthenticationOptions.ROLE_CLAIM_TYPE
+                                    };
+                                    //options.MapInboundClaims = false;
```
severussundar, line 145: super nit: remove

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: correct. Making comment clearer:
```
         // Process all remaining claims adding all `Claim` objects with the same claimType (claim name)
            // into a list and storing that in resolvedClaims using the claimType as the key.
```

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: thanks for calling this out. While the list of claims won't be null/empty due because the code populating the list won't add the list/claim type to the dictionary if one doesn't exist, still good to have defensive code. Adding the condition:
```csharp
&& claims is not null && claims.Count > 0
```
so that if that condition is not met an exception is thrown.

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: No, For context we're talking about this:

```csharp
Dictionary<string, List<Claim>> userClaims = GetAllAuthenticatedUserClaims(context);

foreach ((string claimName, List<Claim> claimValues) in userClaims)
```

`userClaims` will not have an entry for a claim that doesn't exist. If there is an entry in the dictionary, there is at least one entry in the list of claims.

seantleonard: also, the following `First()` is only used when `Count()` is > 1.
```csharp
if (claimValues.Count > 1)
{
    switch (claimValues.First().ValueType)
```

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
seantleonard: thanks for catching that

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: added separate cases for those. thanks for calling out

src/Core/Authorization/AuthorizationResolver.cs
severussundar: My bad, I meant to add this in the `else` block. There `0`th element is accessed directly, hence I had the question. But, if there will be atleast one element in the `claimValues`, then it should be good.

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: thanks for the suggestion, updated

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
```diff
@@ -98,5 +106,105 @@ public static class WebHostBuilderHelper
                 })
                 .StartAsync();
         }
+
+        public static async Task<IHost> CreateWebHostCustomIssuer(SecurityKey key)
```
abhishekkumams, line 110: summary?

ayush3797:
> Example:
Given the database policy: @claims.groups eq @item.groupid, DAB would fail a request if the provided access token had >1 group because the groups claim is a JSON string array in the JWT token which dotnet resolves into multiple Claim objects. (one claim per group where claimType is groups and value is groupGUID.

> How is this implemented?
DAB now recognizes when multiple Claim objects exist for a single claim type. Given the policy @claims.groups eq @item.groupid DAB will replace @claims.groups with the first instance of the claimType groups that DAB finds. If DAB has resolved two groups claim objects such as ["groupGUID1", "groupGUID2"], DAB would only resolve the first it finds: groupGUID1.

This PR description doesn't feel up-to-date. While first para says we fail the request, the second says we use the first value in the array. Maybe the first one needs an update?

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -438,61 +440,157 @@ public IEnumerable<string> GetAllowedExposedColumns(string entityName, string ro
...
+                    case ClaimValueTypes.Integer:
+                    case ClaimValueTypes.Integer32:
+                        processedClaims.Add(claimName, value: JsonSerializer.Serialize(claimValues.Select(claim => int.Parse(claim.Value))));
+                        break;
+                    // Per Microsoft Docs: UInt32's CLS compliant alternative is Integer64
+                    case ClaimValueTypes.UInteger32:
+                    case ClaimValueTypes.Integer64:
+                        processedClaims.Add(claimName, value: JsonSerializer.Serialize(claimValues.Select(claim => long.Parse(claim.Value))));
+                        break;
+                    // Per Microsoft Docs: UInt64's CLS compliant alternative is decimal
```
ayush3797, line 494: Can link the mentioned ms doc in the method summary if not done already

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -525,13 +623,20 @@ private static string GetPolicyWithClaimValues(string policy, Dictionary<string,
     /// <param name="claimsInRequestContext">Dictionary populated with all the user claims.</param>
     /// <returns>The claim value for the given claimTypeMatch.</returns>
```
ayush3797, line 624: Maybe this comment needs an update. It rather returns the first element/value from the list of values in the claim value.

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -32,6 +33,7 @@ public class AuthorizationResolver : IAuthorizationResolver
     public const string AUTHORIZATION_HEADER = "Authorization";
     public const string ROLE_ANONYMOUS = "anonymous";
     public const string ROLE_AUTHENTICATED = "authenticated";
+    public const string ENTRA_ID_SCP_CLAIM = "scp";
```
ayush3797, line 36: Looks like an unused constant.

ayush3797 (approved):
Some nits. Looks good!

seantleonard:
> > Example:
> > Given the database policy: @claims.groups eq @item.groupid, DAB would fail a request if the provided access token had >1 group because the groups claim is a JSON string array in the JWT token which dotnet resolves into multiple Claim objects. (one claim per group where claimType is groups and value is groupGUID.
> 
> > How is this implemented?
> > DAB now recognizes when multiple Claim objects exist for a single claim type. Given the policy @claims.groups eq @item.groupid DAB will replace @claims.groups with the first instance of the claimType groups that DAB finds. If DAB has resolved two groups claim objects such as ["groupGUID1", "groupGUID2"], DAB would only resolve the first it finds: groupGUID1.
> 
> This PR description doesn't feel up-to-date. While first para says we fail the request, the second says we use the first value in the array. Maybe the first one needs an update?

Clarified to mean that request failure occurred before this change and this change fixes that:

_Given the database policy: `@claims.groups eq @item.groupid`, DAB would previously (before this PR's changes) fail a request if the provided access token had >1 group because the `groups` claim is a JSON string array in the JWT token which dotnet resolves into multiple `Claim` objects. (one claim per group where claimType is `groups` and value is `groupGUID`. The request failure no longer occurs._

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
seantleonard: added

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
```diff
@@ -1204,12 +1200,172 @@ public void ValidateClientRoleHeaderClaimIsAddedToClaimsInRequestContext()
...
+            // Assert
+            // The only role that should be present is the role set in the client role header. The claims collection
+            // configured in the 'arrange' section of this test must include that role in the list of role claims.
+            // DAB historically only sent a single role into MS SQL's session context, so changing this behavior
+            // is a breaking change.
+            Assert.AreEqual(expected: true, claimsInRequestContext.ContainsKey(AuthenticationOptions.ROLE_CLAIM_TYPE), message: "The 'roles' claim must be present.");
+            Assert.AreEqual(expected: TEST_ROLE, claimsInRequestContext[AuthenticationOptions.ROLE_CLAIM_TYPE], message: "The roles claim should have the value:" + TEST_ROLE);
+
+            // Ensure JSON arrays are reconstructed correctly aligning to the JWT token claim's value type
+            Assert.AreEqual(expected: @"[""d74b8d81-39eb-4201-bd9f-9f1c4011e3c9"",""18d14519-c4da-4ad4-936d-9a2de69d33cf"",""9e513fc0-e8af-43b1-a6c7-949edb1967a3""]", claimsInRequestContext["wids"]);
```
Aniruddh25, line 1260: Missing Assert on groups, scp claims

Aniruddh25 (approved):
Approving a priori, looks good overall, need to go through PR description once again.

src/Service.Tests/Authorization/AuthorizationResolverUnitTests.cs
seantleonard: fixed

PR #2011, opened 2024-02-05
Updated Json Schema which includes `cache` property

seantleonard:
## Why this change?

This change follows up on https://github.com/Azure/data-api-builder/pull/1865, which adds caching support. 
VSCode suggesting the new `cache` properties were errors when included in the config file. That is because the properties needed to be added to the json schema.

Global `cache` property in the `runtime` section which adjacent to `rest`, `graphql`, and `host`.

```json
{
  "runtime": {
    "cache": {
      "enabled": false,
      "ttl-seconds": 5
    }
  }
}
```

and per entity alongside `source`, `graphql`, `rest`, and `permissions`

```json
{
  "myEntity": {
    "cache": {
      "enabled": false,
      "ttl-seconds": 5
    }
  }
}
```


schemas/dab.draft.schema.json
```diff
@@ -236,6 +236,22 @@
               }
             }
           }
+        },
+        "cache": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "enabled": {
```
Aniruddh25, line 244: with this new schema in place, will the `dab validate` command be successful if used on a config that contains the cache property?

Aniruddh25 (approved):
LGTM, after validation is done

aaronburtle (approved):
Looks good to me!

schemas/dab.draft.schema.json
seantleonard: Working on validating this. Odd error that states schema references were not all resolved properly. Doesn't seem related to `cache` because cli still says "json validated" but an exception is thrown from the validate function which checks config against schema

seantleonard: image

The JSON Schema validation tooling does not like that we have the following property and value in our `dab-config.json` file:
```json
```

Even when the link IS valid like `https://raw.githubusercontent.com/Azure/data-api-builder/37c6db7eb23edcd5b9398e4751951b79da26a73f/schemas/dab.draft.schema.json` the error is still encountered.

when i hard code DAB's json schema in code and exclude the `$schema` property and url value from the runtime config file, i get the expected behavior
```csharp
    public async Task<string?> GetJsonSchema(RuntimeConfig runtimeConfig)
    {
        // DEFAULT_CONFIG_SCHEMA_LINK is just a placeholder with no actual schema, hence should not be used to fetch the schema.
        if (!string.IsNullOrWhiteSpace(runtimeConfig.Schema) && !runtimeConfig.Schema.Equals(RuntimeConfig.DEFAULT_CONFIG_SCHEMA_LINK))
        {
            try
            {
                JsonSchema jsonSchema = await JsonSchema.FromUrlAsync("https://raw.githubusercontent.com/Azure/data-api-builder/37c6db7eb23edcd5b9398e4751951b79da26a73f/schemas/dab.draft.schema.json");
                return jsonSchema.ToJson();
            }
            catch (Exception e)
            {
                _logger!.LogError($"Failed to get schema from url: {runtimeConfig.Schema}\n{e}");
            }
        }
```

Success message: 
image

Information: The config satisfies the schema requirements.

- Error: Failed to get schema from url: https://raw.githubusercontent.com/Azure/data-api-builder/37c6db7eb23edcd5b9398e4751951b79da26a73f/schemas/dab.draft.schema.json

> System.InvalidOperationException: **Could not find the JSON path of a referenced schema: Manually referenced schemas must be added to the 'Definitions' of a parent schema.**

>    at NJsonSchema.JsonPathUtilities.**GetJsonPaths**(Object rootObject, IEnumerable`1 searchedObjects, IContractResolver contractResolver)
> ...
>    at Azure.DataApiBuilder.Core.Configurations.JsonConfigSchemaValidator.GetJsonSchema(RuntimeConfig runtimeConfig) in **C:\Users\seleonar\Documents\dev\dabquickpr\src\Core\Configurations\JsonConfigSchemaValidator.cs:line 75**

seantleonard: All that being said, we'll need to have a follow up deeper look into what if anything needs to be changed.

abhishekkumams: when the schema fetching from url step fails, it gets the schema from the package itself, and it should do the validation giving the expected result. apart from the url fetch failure, did you see any difference wrt to using the hard-coded json schema?

seantleonard: I have not been able to get URL schema fetching to succeed in any circumstances. 
However, pulling from assembly works. I'm not sure what's wrong with pulling from url. the ones I've used are accessible and not the unversioned sample link.

seantleonard: #2013

PR #2012, opened 2024-02-05
Cache Enablement check changes

seantleonard:
## Why make this change?

- Ensure cache is used when enabled and add additional checks to make this flow as expected.

## What is this change?

- Change cache check to ensure dbpolicies are NOT resolved for request in order to use cache
- Add check to ensure entity cache is enabled.

## How was this tested?

- TBD

### DIFF
--- a/src/Core/Resolvers/SqlQueryEngine.cs
+++ b/src/Core/Resolvers/SqlQueryEngine.cs
@@ -215,7 +215,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             {
                 bool dbPolicyConfigured = !string.IsNullOrEmpty(structure.DbPolicyPredicatesForOperations[EntityActionOperation.Read]);
 
-                if (dbPolicyConfigured)
+                if (!dbPolicyConfigured)
                 {
                     DatabaseQueryMetadata queryMetadata = new(queryText: queryString, dataSource: dataSourceName, queryParameters: structure.Parameters);
                     JsonElement result = await _cache.GetOrSetAsync<JsonElement>(queryExecutor, queryMetadata, cacheEntryTtl: runtimeConfig.GetEntityCacheEntryTtl(entityName: structure.EntityName));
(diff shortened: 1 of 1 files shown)


src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -211,11 +211,16 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
             // Open connection and execute query using _queryExecutor
             string queryString = queryBuilder.Build(structure);
 
+            // Global Cache enablement check
             if (runtimeConfig.CanUseCache())
             {
+                // Entity level cache behavior checks
                 bool dbPolicyConfigured = !string.IsNullOrEmpty(structure.DbPolicyPredicatesForOperations[EntityActionOperation.Read]);
+                bool entityCacheEnabled = runtimeConfig.Entities[structure.EntityName].IsCachingEnabled;
```
Aniruddh25, line 219: Is caching enabled check applicable only for parent level entities in GraphQL queries - for .e.g. there could be nested entities queried but we are not checking if caching is enabled for those or not. Is this by design?

What happens if 1 is enabled, 1 is not -> does it mean if 1 is not enabled we dont cache at all?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -211,11 +211,16 @@ public async Task<IActionResult> ExecuteAsync(StoredProcedureRequestContext cont
...
 
-                if (dbPolicyConfigured)
+                // If a db policy is resolved for the read operation, skip the cache. We want to avoid caching token metadata
```
Aniruddh25, line 221:
```suggestion
                // If a db policy is configured for the read operation, skip the cache. We want to avoid caching token metadata
```

Aniruddh25 (approved):
LGTM, pending answer to open questions. Aaron and I reviewed together.

aaronburtle (approved):
Looks good!

src/Core/Resolvers/SqlQueryEngine.cs
seantleonard: Only the top-level entity's `cache` setting is considered which is by-design. 

> What happens if 1 is enabled, 1 is not -> does it mean if 1 is not enabled we dont cache at all?
Request will be cached as long as the entity which has caching enabled, is the top-level entity.

PR #2015, opened 2024-02-06
Adding In-Memory cache support for CosmosDB

### LINKED ISSUE #1884: In Memory Cache - Cosmos
Update cosmos Query Engine to support in memory cache.

General design doc: [REST and GraphQL Query Caching Design Doc by seantleonard · Pull Request #1801 · Azure/data-api-builder (github.com)](https://github.com/Azure/data-api-builder/pull/1801)

SQL Implementation work: 
[ENGINE-CORE] [In-Memory Cache by seantleonard · Pull Request #1881 · Azure/data-api-builder (github.com)](https://github.com/Azure/data-api-builder/pull/1881)
[CONFIG] [Add Cache properties to runtime config file with JSON Serialize/Deserialize by seantleonard · Pull Request #1865 · Azure/data-api-builder (github.com)](https://github.com/Azure/data-api-builder/pull/1865)

neeraj-sharma2592:
## Why make this change?
Adding In-Memory cache support for CosmosDB. Closes #1884 

## What is this change?
Introduced a new generic method GetOrSetAsync  for getting and setting cached item.
Using GetOrSetAsync for caching Cosmosdb query result.

## How was this tested?

- [x] Integration Tests
Added integration tests using Mock Func execution by DabCacheService.GetOrSetAsync. 
Performed manual testing of caching behavior. Had to manually override enablecaching and TTL since config changes were throwing an error. Will perform another round of tests by enabling caching from configuration before merging the PR.

### DIFF
--- a/src/Config/ObjectModel/RuntimeConfig.cs
+++ b/src/Config/ObjectModel/RuntimeConfig.cs
@@ -374,7 +374,7 @@ public record RuntimeConfig
     public bool CanUseCache()
     {
         bool setSessionContextEnabled = DataSource.GetTypedOptions<MsSqlOptions>()?.SetSessionContext ?? true;
-        return IsCachingEnabled && (SqlDataSourceUsed || CosmosDataSourceUsed) && !setSessionContextEnabled;
+        return IsCachingEnabled && !setSessionContextEnabled;
     }
 
     /// <summary>
--- a/src/Config/ObjectModel/RuntimeConfig.cs
+++ b/src/Config/ObjectModel/RuntimeConfig.cs
@@ -374,7 +374,7 @@ public record RuntimeConfig
     public bool CanUseCache()
     {
         bool setSessionContextEnabled = DataSource.GetTypedOptions<MsSqlOptions>()?.SetSessionContext ?? true;
-        return IsCachingEnabled && SqlDataSourceUsed && !setSessionContextEnabled;
+        return IsCachingEnabled && (SqlDataSourceUsed || CosmosDataSourceUsed) && !setSessionContextEnabled;
     }
 
     /// <summary>
--- a/src/Core/Resolvers/CosmosQueryEngine.cs
+++ b/src/Core/Resolvers/CosmosQueryEngine.cs
@@ -121,12 +121,12 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         /// <summary>
         /// ExecuteQueryAsync Performs single parition and cross partition queries. 
         /// </summary>
-        /// <param name="structure"></param>
-        /// <param name="querySpec"></param>
-        /// <param name="queryRequestOptions"></param>
-        /// <param name="container"></param>
-        /// <param name="idValue"></param>
-        /// <param name="partitionKeyValue"></param>
+        /// <param name="structure">CosmosQueryStructure</param>
+        /// <param name="querySpec">QueryDefinition Define a Cosmos SQL Query</param>
+        /// <param name="queryRequestOptions">The Cosmos query request options</param>
+        /// <param name="container">CosmosDB Container</param>
+        /// <param name="idValue">Id param</param>
+        /// <param name="partitionKeyValue">PartitionKey Value</param>
         /// <returns>JObject</returns>
         private static async Task<JObject> ExecuteQueryAsync(CosmosQueryStructure structure, QueryDefinition querySpec, QueryRequestOptions queryRequestOptions, Container container, string idValue, string partitionKeyValue)
         {
--- a/src/Core/Resolvers/CosmosQueryEngine.cs
+++ b/src/Core/Resolvers/CosmosQueryEngine.cs
@@ -119,7 +119,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         }
 
         /// <summary>
-        /// ExecuteQueryAsync Performs single parition and cross partition queries. 
+        /// ExecuteQueryAsync Performs single partition and cross partition queries. 
         /// </summary>
         /// <param name="structure">CosmosQueryStructure</param>
         /// <param name="querySpec">QueryDefinition Define a Cosmos SQL Query</param>
--- a/src/Core/Resolvers/CosmosQueryEngine.cs
+++ b/src/Core/Resolvers/CosmosQueryEngine.cs
@@ -1,7 +1,7 @@
-    // Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-# nullable disable
+#nullable disable
 using System.Text;
 using System.Text.Json;
 using Azure.DataApiBuilder.Auth;
--- a/src/Core/Resolvers/CosmosQueryEngine.cs
+++ b/src/Core/Resolvers/CosmosQueryEngine.cs
@@ -107,7 +107,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 
                 DatabaseQueryMetadata queryMetadata = new(queryText: queryString, dataSource: dataSourceKey.ToString(), queryParameters: structure.Parameters);
 
-                executeQueryResult = await _cache.GetOrSetAsync<JObject>(async () => await ExecuteQueryAsync(structure, querySpec, queryRequestOptions, container, idValue, partitionKeyValue), queryMetadata, 10);//runtimeConfig.GetEntityCacheEntryTtl(entityName: structure.EntityName));
+                executeQueryResult = await _cache.GetOrSetAsync<JObject>(async () => await ExecuteQueryAsync(structure, querySpec, queryRequestOptions, container, idValue, partitionKeyValue), queryMetadata, runtimeConfig.GetEntityCacheEntryTtl(entityName: structure.EntityName));
             }
             else
             {
--- a/src/Core/Resolvers/CosmosQueryEngine.cs
+++ b/src/Core/Resolvers/CosmosQueryEngine.cs
@@ -63,7 +63,6 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             IDictionary<string, object> parameters,
             string dataSourceName)
         {
-            // TODO: fixme we have multiple rounds of serialization/deserialization JsomDocument/JObject
             // TODO: add support for join query against another container
             // TODO: add support for TOP and Order-by push-down
 
@@ -120,7 +119,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         }
 
         /// <summary>
-        /// ExecuteQueryAsync Perform single parition and cross partition query. 
+        /// ExecuteQueryAsync Performs single parition and cross partition queries. 
         /// </summary>
         /// <param name="structure"></param>
         /// <param name="querySpec"></param>
@@ -128,7 +127,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         /// <param name="container"></param>
         /// <param name="idValue"></param>
         /// <param name="partitionKeyValue"></param>
-        /// <returns>JsonDocument</returns>
+        /// <returns>JObject</returns>
         private static async Task<JObject> ExecuteQueryAsync(CosmosQueryStructure structure, QueryDefinition querySpec, QueryRequestOptions queryRequestOptions, Container container, string idValue, string partitionKeyValue)
         {
             string requestContinuation = null;
--- a/src/Core/Resolvers/CosmosQueryEngine.cs
+++ b/src/Core/Resolvers/CosmosQueryEngine.cs
@@ -92,7 +92,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 
             JObject executeQueryResult = null;
 
-            if (runtimeConfig.CanUseCache())
+            if (runtimeConfig.CanUseCache() && runtimeConfig.Entities[structure.EntityName].IsCachingEnabled)
             {
                 StringBuilder dataSourceKey = new(dataSourceName);
 
--- a/src/Core/Resolvers/CosmosQueryEngine.cs
+++ b/src/Core/Resolvers/CosmosQueryEngine.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 # nullable disable
-using System.Configuration;
 using System.Text;
 using System.Text.Json;
 using Azure.DataApiBuilder.Auth;
@@ -15,8 +14,6 @@ using Azure.DataApiBuilder.Core.Services.MetadataProviders;
 using Azure.DataApiBuilder.Service.GraphQLBuilder.Queries;
 using HotChocolate.Language;
 using HotChocolate.Resolvers;
-using HotChocolate.Utilities;
-using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Azure.Cosmos;
 using Newtonsoft.Json.Linq;
@@ -75,7 +72,6 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             CosmosQueryStructure structure = new(context, parameters, metadataStoreProvider, _authorizationResolver, _gQLFilterParser);
             RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
 
-            string requestContinuation = null;
             string queryString = _queryBuilder.Build(structure);
             QueryDefinition querySpec = new(queryString);
             QueryRequestOptions queryRequestOptions = new();
@@ -94,18 +90,22 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                 queryRequestOptions.PartitionKey = new PartitionKey(partitionKeyValue);
             }
 
-            if (structure.IsPaginated)
-            {
-                queryRequestOptions.MaxItemCount = (int?)structure.MaxItemCount;
-                requestContinuation = Base64Decode(structure.Continuation);
-            }
-
-
             if (runtimeConfig.CanUseCache())
             {
-                DatabaseQueryMetadata queryMetadata = new(queryText: queryString, dataSource: dataSourceName, queryParameters: structure.Parameters);
+                StringBuilder dataSourceKey = new( dataSourceName);
 
-                JObject result = await _cache.GetOrSetAsync2(async () => await ExecuteQueryAsync(structure, querySpec, queryRequestOptions, container, idValue, partitionKeyValue), queryMetadata, runtimeConfig.GetEntityCacheEntryTtl(entityName: structure.EntityName));
+                // to support caching for paginated query adding continuation token in the datasource
+                if (structure.IsPaginated)
+                {
+                    dataSourceKey.Append(":");
+                    dataSourceKey.Append(structure.IsPaginated);
+                    dataSourceKey.Append(":");
+                    dataSourceKey.Append(structure.Continuation);
+                }
+
+                DatabaseQueryMetadata queryMetadata = new(queryText: queryString, dataSource: dataSourceKey.ToString(), queryParameters: structure.Parameters);
+
+                JObject result = await _cache.GetOrSetAsync<JObject>(async () => await ExecuteQueryAsync(structure, querySpec, queryRequestOptions, container, idValue, partitionKeyValue), queryMetadata, runtimeConfig.GetEntityCacheEntryTtl(entityName: structure.EntityName));
 
                 JsonDocument cacheServiceResponse = JsonDocument.Parse(result.ToString());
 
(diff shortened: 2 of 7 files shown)


seantleonard:
There are some changes i need to cherry pick from 0.10 release branch to fix this behavior:
> Had to manually override enablecaching and TTL since config changes were not throwing an error.

seantleonard (commented):
still need to go through tests.

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -122,19 +174,19 @@ public class CosmosQueryEngine : IQueryEngine
...
 
                         // This extra deserialize/serialization will be removed after moving to Newtonsoft from System.Text.Json
-                        return new Tuple<JsonDocument, IMetadata>(JsonDocument.Parse(res.ToString()), null);
+                        return res;
                     }
 
                     if (page.Count > 0)
                     {
-                        return new Tuple<JsonDocument, IMetadata>(JsonDocument.Parse(page.First().ToString()), null);
+                        return page.First();
```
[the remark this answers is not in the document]
neeraj-sharma2592: Instead of returning the tuple directly from help function we are returning the JObject and constructing the tuple at later stage. This is done in order to pass this helper function to func which then get executed with in Cache context. 
Tuple here was taking the JsonDocument as one of the input which was throwing the object dispose error while caching.

https://github.com/Azure/data-api-builder/blob/2ae5c0d7ab2ee03906d8fa8e70af40c086229e15/src/Core/Resolvers/CosmosQueryEngine.cs#L119C1-L120C1

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -78,6 +78,35 @@ public DabCacheService(IFusionCache cache, ILogger<DabCacheService>? logger, IHt
...
+    /// <summary>
+    /// Attempts to fetch response from cache. If there is a cache miss, invoke executeQueryAsync Func to get a response
+    /// </summary>
+    /// <typeparam name="TResult">Response payload Type</typeparam>
+    /// <param name="executeQueryAsync">Func with a result of type TResult. Only executed after a cache miss.</param>
+    /// <param name="queryMetadata">Metadata used to create a cache key or fetch a response from the database.</param>
+    /// <param name="cacheEntryTtl">Number of seconds the cache entry should be valid before eviction.</param>
+    /// <returns>JSON Response</returns>
+    /// <exception cref="Exception">Throws when the cache-miss factory method execution fails.</exception>
+    public async ValueTask<TResult?> GetOrSetAsync<TResult>(Func<Task<TResult>> executeQueryAsync, DatabaseQueryMetadata queryMetadata, int cacheEntryTtl)
```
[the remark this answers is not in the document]
neeraj-sharma2592: I didn't try with Jsonelement. The current Cosmos response is received as a JObject and then gets converted to a JsonDocument. I made it as TResult for two reasons:

1) To decouple the type from the Cache, so that the cache can store any data type as long as it's serializable.
2) To use the same cache method for ExecuteListAsync, which returns a list of JObject/JsonDocument. However, it later turned out that Cosmos does not support the list operation. Nevertheless, to make it future-proof, I inclined towards a generic approach instead of providing a direct type.

https://github.com/Azure/data-api-builder/blob/2ae5c0d7ab2ee03906d8fa8e70af40c086229e15/src/Core/Resolvers/CosmosQueryEngine.cs#L228C1-L229C1

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -472,6 +479,152 @@ ... p
             Assert.AreEqual(id, response.GetProperty("id").GetString());
         }
 
+        [TestMethod]
+        public async Task QueryWithCacheEnabledShouldReturnCachedResponse()
```
seantleonard, line 483: need test comments describing what is being validated:

It looks like you're querying for a record , then changing a property in that record, performing a second query to ensure the cached record is returned because you are still within the TTL of the cache entry?

neeraj-sharma2592: Yes second query should return the cached entry instead of the modified one.  I'll add the comments for clarity.

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -122,19 +174,19 @@ public class CosmosQueryEngine : IQueryEngine
                             new JProperty(QueryBuilder.PAGINATION_FIELD_NAME, jarray));
 
                         // This extra deserialize/serialization will be removed after moving to Newtonsoft from System.Text.Json
-                        return new Tuple<JsonDocument, IMetadata>(JsonDocument.Parse(res.ToString()), null);
```
mbhaskar, line 125: Thanks for taking care of this. Removing additional serialization step helps improve perf too. You should probably delete the comment too.

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -472,6 +479,158 @@ ... p
...
+                // Second query - the data returned is from the cache, not the mutated one
+                JsonElement queryResponse2 = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
+                                             client,
+                                             server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                                             query: graphQLQuery,
+                                             queryName: queryName
+                                             );
+
+                // Asserting cached data is returned
+                Assert.IsFalse(queryResponse.GetProperty("name").GetString() != name, "Query didn't return cached value");
```
mbhaskar, line 613: nit: Instead of negation, Asserting with IsTrue improves the readability of the test. 
Are we testing that old value is returned from the cache even after updating the name?

mbhaskar (approved):

src/Service.Tests/CosmosTests/QueryTests.cs
neeraj-sharma2592: Yes, we are testing the same cached value is returned and not the modified one. 
Negative assertion is added to provide a clear error message when the expected result doesn't match.

src/Core/Resolvers/CosmosQueryEngine.cs
neeraj-sharma2592: sure.

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -472,6 +479,158 @@ ... p
...
+                // Clean up setting document name back to original value
+                _ = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
+                                client,
+                                server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                                query: mutation,
+                                queryName: "updatePlanet",
+                                variables: new() { { "id", id }, { "partitionKeyValue", id }, { "item", update } },
+                                authToken: AuthTestHelper.CreateStaticWebAppsEasyAuthToken(),
+                                clientRoleHeader: AuthorizationResolver.ROLE_AUTHENTICATED
+                                );
```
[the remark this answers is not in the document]
neeraj-sharma2592: Right. I have removed the cleanup that I was performing in my test. And yes, it makes sense to revisit the TestCleanUp function, which is creating the container multiple times. Perhaps we can perform cleanup at the assembly level instead for each test. This will require some analysis. I'll create a separate item to check on this. Additionally, we are inquiring with the Emulator team about the issues we are facing in Linux and macOS images. Once those are resolved, we can fully transition to the emulator. However, we can start looking into this before the emulator team provides a fix.

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -80,6 +90,48 @@ public class CosmosQueryEngine : IQueryEngine
...
+                // to support caching for paginated query adding continuation token in the datasource
+                if (structure.IsPaginated)
+                {
+                    dataSourceKey.Append(":");
+                    dataSourceKey.Append(structure.IsPaginated);
+                    dataSourceKey.Append(":");
+                    dataSourceKey.Append(structure.Continuation);
+                }
+
+                DatabaseQueryMetadata queryMetadata = new(queryText: queryString, dataSource: dataSourceKey.ToString(), queryParameters: structure.Parameters);
```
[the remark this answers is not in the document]
neeraj-sharma2592: `why is pagination metadata used for DatabaseQueryMetadata.dataSource as opposed to queryText?`
Cosmos DB supports paginated queries, where the query remains constant while the continuation token changes. Cosmos DB utilizes this continuation token to retrieve the next set of records. Due to this functionality, relying solely on the query text as the key is undesirable, as it would result in receiving the same response for each page.
Here, we anticipate that the user will provide the same token again. Given that the initial cached response will return the cached token value, subsequent calls will be handled by the cache. If the user attempts to tamper with the continuation token, cached data will not be returned, and later, when the SDK makes a call, it will fail at the Cosmos DB end as well.

` Also should the Continuation token be used in its raw form here within dataSourceKey? what about added it as a queryParameter to avoid "sql" injection?`
We are appending the continuation token to the dataSourceKey, which will serve as our cache key. At no point are we executing any operations on this text. Furthermore, any changes or injections in the continuation key are already handled at the Cosmos DB end. Therefore, we should be in good shape.

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -80,6 +90,48 @@ public class CosmosQueryEngine : IQueryEngine
...
+
+                executeQueryResult = await _cache.GetOrSetAsync<JObject>(async () => await ExecuteQueryAsync(structure, querySpec, queryRequestOptions, container, idValue, partitionKeyValue), queryMetadata, runtimeConfig.GetEntityCacheEntryTtl(entityName: structure.EntityName));
+            }
+            else
+            {
+                executeQueryResult = await ExecuteQueryAsync(structure, querySpec, queryRequestOptions, container, idValue, partitionKeyValue);
+            }
+
+            JsonDocument response = executeQueryResult != null ? JsonDocument.Parse(executeQueryResult.ToString()) : null;
```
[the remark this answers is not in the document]
neeraj-sharma2592: We are returning this in the response, so it will be disposed of similarly to how it's happening for other APIs.

https://github.com/Azure/data-api-builder/blob/99c052dbffca6d585e8fb256e12e586c13814fbf/src/Core/Services/ResolverMiddleware.cs#L170

src/Core/Resolvers/CosmosQueryEngine.cs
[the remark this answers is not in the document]
neeraj-sharma2592: Right, but we don't need to store JsonElement either. I have it as JObject. We were converting the JObject to JsonDocument, which is no longer needed here.

Aniruddh25 (changes requested):
Need to still review tests, posting comments so far.

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -251,6 +252,124 @@ public async Task CacheServiceFactoryInvocationThrowsException()
...
+            int cacheEntryTtl = 1;
+            // First call. Cache miss
+            _ = await dabCache.GetOrSetAsync<JObject>(executeQueryAsync: mockExecuteQuery.Object, queryMetadata: queryMetadata, cacheEntryTtl: cacheEntryTtl);
+
+            // Act
+            JObject? result = await dabCache.GetOrSetAsync<JObject>(executeQueryAsync: mockExecuteQuery.Object, queryMetadata: queryMetadata, cacheEntryTtl: cacheEntryTtl);
+
+            // Assert
+            Assert.AreEqual(expected: true, actual: mockExecuteQuery.Invocations.Count is 1, message: ERROR_UNEXPECTED_INVOCATIONS);
+            Assert.IsFalse(mockExecuteQuery.Invocations.Count is 2, message: "Expected a cache hit, but observed two cache misses.");
```
Aniruddh25, line 321: can we keep the condition to be `> 1`

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -251,6 +252,124 @@ public async Task CacheServiceFactoryInvocationThrowsException()
...
+            Assert.AreEqual(expected: true, actual: mockExecuteQuery.Invocations.Count is 1, message: ERROR_UNEXPECTED_INVOCATIONS);
```
Aniruddh25, line 322: Duplicate assert on line 322 and 320?

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -251,6 +252,124 @@ public async Task CacheServiceFactoryInvocationThrowsException()
...
+            _ = await dabCache.GetOrSetAsync<JObject>(executeQueryAsync: mockExecuteQuery.Object, queryMetadata: queryMetadata, cacheEntryTtl: cacheEntryTtl);
+
+            // Sleep for the amount of time the cache entry is valid to trigger eviction.
+            Thread.Sleep(millisecondsTimeout: cacheEntryTtl * 1000);
+
+            // Act
+            JObject? result = await dabCache.GetOrSetAsync<JObject>(executeQueryAsync: mockExecuteQuery.Object, queryMetadata: queryMetadata, cacheEntryTtl: cacheEntryTtl);
+
+            // Assert
+            Assert.IsFalse(mockExecuteQuery.Invocations.Count is 1, message: "QueryExecutor invocation count too low. A cache hit shouldn't have occurred since the entry should have expired.");
```
Aniruddh25, line 367: if we assert false on < 2, > 2. equality assert on 2 is not needed

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -472,6 +479,141 @@ ... p
...
+
+                // Asserting name is mutated
+                Assert.IsFalse(queryResponse.GetProperty("name").GetString().Equals(update.name), "Mutation didn't change the name successfully");
+
+                // Second query - the data returned is from the cache, not the mutated one
```
Aniruddh25, line 604: Wouldn't you want to invalidate the cache if it gets mutated? So, shouldn't name in queryResponse2 be the new value?

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -472,6 +479,141 @@ ... p
...
+                JsonElement queryResponse2 = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
+                                             client,
+                                             server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                                             query: graphQLQuery,
+                                             queryName: queryName
+                                             );
+
+                // Asserting cached data is returned
+                Assert.IsFalse(queryResponse.GetProperty("name").GetString() != name, "Query didn't return cached value");
+            }
```
Aniruddh25, line 614: If the query should be returning old value from the cache, it should be queryResponse2 compared with name which comes from queryResponse 1.

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -472,6 +479,141 @@ ... p
...
+        }
```
Aniruddh25, line 615: Should add a new test for query returning a pagination continuation token, and check if it gets, cache hits.

Aniruddh25 (changes requested):
Looking forward to fixed tests.

src/Core/Resolvers/CosmosQueryEngine.cs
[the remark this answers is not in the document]
neeraj-sharma2592: Right. So, this will be the sequence.
First query -> SDK -> Cache
Second query with continuation token -> SDK -> Cache
End

Now, if user again request make the same first query, we will return cached data which will have the token for second query. And the response of which is also cached.

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -78,6 +78,35 @@ public DabCacheService(IFusionCache cache, ILogger<DabCacheService>? logger, IHt
...
+    {
+
+        string cacheKey = CreateCacheKey(queryMetadata);
```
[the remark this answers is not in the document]
Aniruddh25: Please either create a tracking issue for it, or make sure to send a followup PR.

Aniruddh25 (changes requested):
The test for cache key including ispaginated continuation token need a way to verify response is being received from the cache.

src/Core/Services/Cache/DabCacheService.cs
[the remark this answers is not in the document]
neeraj-sharma2592: I can create a tracking item. Since I couldn't locate caching test cases for the SQL implementation, it would be preferable for the SQL team to handle any follow-up PR if they intend to implement this generic behavior. Integration tests can also be incorporated into the same PR for comprehensive coverage.

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -472,6 +479,282 @@ ... p
...
+                                       server.Services.GetRequiredService<RuntimeConfigProvider>(),
+                                       query: planetConnectionQueryStringFormat,
+                                       queryName: "planets"
+                                       );
+
+                    JsonElement after = page.GetProperty(QueryBuilder.PAGINATION_TOKEN_FIELD_NAME);
+                    afterToken = after.ToString();
+
+                    // Asserting continuation are returned in the same sequence as original query.
+                    Assert.IsTrue(continuationTokenList[tokenCount++].Equals(afterToken) , "ContinuationToken not found in the cached response");
```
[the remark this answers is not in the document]
neeraj-sharma2592: I have updated the test cases to perform a mutation between two queries. However, implementing mocking would necessitate the introduction of new interfaces and classes for QueryExecutor, which would expand the scope of this modification. We could contemplate incorporating this change in the future if we decide to invalidate the cache for mutation operations.

Aniruddh25 (approved):
Looks good overall, had forgotten to previously post the comment about need for `IsPaginated`. Feel free to merge once thats addressed, but approving to avoid further delay.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -374,7 +374,7 @@ public int GetEntityCacheEntryTtl(string entityName)
     public bool CanUseCache()
     {
         bool setSessionContextEnabled = DataSource.GetTypedOptions<MsSqlOptions>()?.SetSessionContext ?? true;
-        return IsCachingEnabled && SqlDataSourceUsed && !setSessionContextEnabled;
+        return IsCachingEnabled && (SqlDataSourceUsed || CosmosDataSourceUsed) && !setSessionContextEnabled;
```
seantleonard, line 377: `(SqlDataSourceUsed || CosmosDataSourceUsed)` can be removed fully now since all db types allow caching.

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -80,6 +89,54 @@ public class CosmosQueryEngine : IQueryEngine
...
+
+            return new Tuple<JsonDocument, IMetadata>(response, null);
+        }
+
+        /// <summary>
+        /// ExecuteQueryAsync Performs single parition and cross partition queries.
```
seantleonard, line 122:
```suggestion
        /// ExecuteQueryAsync Performs single partition and cross partition queries. 
```

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -80,6 +89,54 @@ public class CosmosQueryEngine : IQueryEngine
...
+            JsonDocument response = executeQueryResult != null ? JsonDocument.Parse(executeQueryResult.ToString()) : null;
+
+            return new Tuple<JsonDocument, IMetadata>(response, null);
+        }
+
+        /// <summary>
+        /// ExecuteQueryAsync Performs single parition and cross partition queries. 
+        /// </summary>
+        /// <param name="structure">CosmosQueryStructure</param>
+        /// <param name="querySpec">QueryDefinition Define a Cosmos SQL Query</param>
```
seantleonard, line 125:
```suggestion
        /// <param name="querySpec">QueryDefinition defining a Cosmos SQL Query</param>
```

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -78,10 +78,43 @@ public DabCacheService(IFusionCache cache, ILogger<DabCacheService>? logger, IHt
...
+               key: cacheKey,
+               async (FusionCacheFactoryExecutionContext<TResult> ctx, CancellationToken ct) =>
+               {
+                   TResult result = await executeQueryAsync();
+
+                   ctx.Options.SetSize(EstimateCacheEntrySize(cacheKey: cacheKey, cacheValue: JsonSerializer.Serialize(result?.ToString())));
+                   ctx.Options.SetDuration(duration: TimeSpan.FromSeconds(cacheEntryTtl));
+
+                   return result;
+
```
seantleonard, line 104:
```suggestion
```

src/Core/Services/Cache/DabCacheService.cs
```diff
@@ -78,10 +78,43 @@ public DabCacheService(IFusionCache cache, ILogger<DabCacheService>? logger, IHt
...
+               });
+
+        return result;
+    }
+
     /// <summary>
     /// Creates a cache key using the request metadata resolved from a built SqlQueryStructure
     /// Format: DataSourceName:QueryText:JSON_QueryParameters
     /// Example: 7a07f92a-1aa2-4e2a-81d6-b9af0a25bbb6:select * from MyTable where id = @param1 = :{"@param1":{"Value":"42","DbType":null}}
+    /// Format CosmosDB: DataSourceName:IsPaginated:ContinuationToken:QueryText:JSON_QueryParameters
```
seantleonard, line 114: why do we need bool `IsPaginated` in the cache key?

src/Service.Tests/Caching/DabCacheServiceIntegrationTests.cs
```diff
@@ -251,6 +252,120 @@ public async Task CacheServiceFactoryInvocationThrowsException()
...
+            {
+                {"param1", new DbConnectionParam(value: "param1Value") }
+            };
+
+            DatabaseQueryMetadata queryMetadata = new(queryText: "select c.name from c", dataSource: "dataSource1", queryParameters: parameters);
+            DabCacheService dabCache = CreateDabCacheService(cache);
+
+            int cacheEntryTtl = 1;
+
+            // Act
```
seantleonard, line 352:
```suggestion
```

src/Service.Tests/CosmosTests/QueryTests.cs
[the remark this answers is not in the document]
seantleonard: please comment here the link to the issue created for this task.

seantleonard (approved):
Please update according to Ani's last feedback for `IsPaginated` and some other formatting nits brought up from my feedback. 

Additionally, you'll need to run `dotnet format` in your local to fix formatting errors caught by unit test pipeline.

src/Core/Services/Cache/DabCacheService.cs
neeraj-sharma2592: removed the check.

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -80,6 +89,54 @@ public class CosmosQueryEngine : IQueryEngine
...
+
+            if (runtimeConfig.CanUseCache() && runtimeConfig.Entities[structure.EntityName].IsCachingEnabled)
+            {
+                StringBuilder dataSourceKey = new(dataSourceName);
+
+                // to support caching for paginated query adding continuation token in the datasource
+                if (structure.IsPaginated)
+                {
+                    dataSourceKey.Append(":");
+                    dataSourceKey.Append(structure.IsPaginated);
```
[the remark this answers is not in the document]
neeraj-sharma2592: removed ispaginated check

src/Service.Tests/CosmosTests/QueryTests.cs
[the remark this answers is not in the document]
neeraj-sharma2592: https://github.com/Azure/data-api-builder/issues/2082

PR #2034, opened 2024-02-13
Cosmos DB: Added Authorization related tests

sourabh1007:
## Why make this change?

No Code Change as of Now.

## What is this change?
Added tests to cover scenarios related to auth in cosmos DB, with below config changes:
a) Added different combinations of role for `Earth` entity
b) Removed `Anonymous` role for `Earth` entity

**NOTE:** Item level authorization tests are not covered here.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests


src/Service.Tests/CosmosTests/MutationTests.cs
seantleonard, line 1: For `[DataRow]` tests, I suggest adding a DisplayName instead of in-line comments which describe what is being tested:

```csharp
[DataRow("stringParam", <...>, DisplayName = "This displayname shows in Visual Studio Test Explorer")]
```

This improves readability and allows allows more easily distinguishing which test case failed, when failures occur.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -252,62 +253,76 @@ public async Task MutationMissingRequiredPartitionKeyValueReturnError()
...
-        /// for entity 'earth'
+        /// Update Mutation performed on the fields with different auth permissions
+        /// It throws permission denied error if role doesn't have permission to perform the operation
         /// </summary>
         [TestMethod]
-        public async Task CreateItemWithUnauthorizedFieldsReturnsError()
+        [DataRow("field-mutation-with-read-permission", DataApiBuilderException.GRAPHQL_MUTATION_FIELD_AUTHZ_FAILURE)] // exclude and include fields
+        [DataRow("authenticated", null)] // full permission
+        [DataRow("only-update-role", "The mutation operation updateEarth was successful but the current user is unauthorized to view the response due to lack of read permissions")] // if only update permission is there
+        [DataRow("wildcard-exclude-fields-role", DataApiBuilderException.GRAPHQL_MUTATION_FIELD_AUTHZ_FAILURE)] // exclude wildcard
```
seantleonard, line 302: can you clarify whether the test case with role == `wildcard-exclude-fields-role` means that the test will fail if the mutation input such as `id` and `name` below cause the authZ failure?:
```graphql
createEarth (item: {{ id: ""{id}"", name: ""{name}"" }}) {{
``` 

(Can also add test DisplayName to make test case self-explanatory)

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -176,7 +176,11 @@ public async Task TestStringFiltersOnTwoLevelNestedArrayType()
                 "JOIN d IN b.moreAttributes " +
                 "WHERE d.name = \"moonattr0\"";
 
-            await ExecuteAndValidateResult(_graphQLQueryName, gqlQuery, dbQueryWithJoin);
+            await ExecuteAndValidateResult(
+                _graphQLQueryName,
+                gqlQuery,
+                dbQueryWithJoin,
+                authToken: AuthTestHelper.CreateStaticWebAppsEasyAuthToken(specificRole: AuthorizationType.Authenticated.ToString()));
```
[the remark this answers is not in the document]
sourabh1007: `cosmosdb_nosql-commands.txt` changes are not yet aligned with actual config JSON as of now...I will make sure it is before I made this PR "Ready For Review"

src/Service.Tests/CosmosTests/MutationTests.cs
sourabh1007: yes..you are right...because in `wildcard-exclude-fields-role` role, permission is limited to fields.

tarazou9 (approved):

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -252,62 +255,79 @@ public async Task MutationMissingRequiredPartitionKeyValueReturnError()
         }
 
         /// <summary>
-        /// Mutation can be performed on the authorized fields because the
-        /// field `id` is an included field for the create operation on the anonymous role defined
-        /// for entity 'earth'
+        /// Create Mutation performed on the fields with different auth permissions
+        /// It throws permission denied error if role doesn't have permission to perform the operation
+        /// 'item-level-permission-role' is not applicable for create operation
```
seantleonard, line 260: unclear why you need to include note on 'item-level-permission-role' here. Out of context, doesn't make sense to reference.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -252,62 +255,79 @@ public async Task MutationMissingRequiredPartitionKeyValueReturnError()
...
         /// </summary>
         [TestMethod]
-        public async Task CanCreateItemWithAuthorizedFields()
+        [DataRow("field-mutation-with-read-permission", DataApiBuilderException.GRAPHQL_MUTATION_FIELD_AUTHZ_FAILURE, DisplayName = "When there is limited permission at field level but have full read permission")]
```
seantleonard, line 263:
```suggestion
        [DataRow("field-mutation-with-read-permission", DataApiBuilderException.GRAPHQL_MUTATION_FIELD_AUTHZ_FAILURE, DisplayName = "AuthZ failure for create mutation because of reference to excluded/disallowed fields.")]
```

nit: suggestion to make DisplayName more self explanatory on what's being validated.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -252,62 +255,79 @@ public async Task MutationMissingRequiredPartitionKeyValueReturnError()
...
+        [DataRow("authenticated", MutationTests.NO_ERROR_MESSAGE, DisplayName = "When CRUD permission is there without any restriction")]
```
seantleonard, line 264:
```suggestion
        [DataRow("authenticated", MutationTests.NO_ERROR_MESSAGE, DisplayName = "AuthZ success when role has no create/read operation restrictions.")]
```

nit: more self explanatory on the validation occurring.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -252,62 +255,79 @@ public async Task MutationMissingRequiredPartitionKeyValueReturnError()
...
+        [DataRow("only-create-role", "The mutation operation createEarth was successful " +
+            "but the current user is unauthorized to view the response due to lack of read permissions", DisplayName = "When ONLY create permission is there")]
+        [DataRow("wildcard-exclude-fields-role", DataApiBuilderException.GRAPHQL_MUTATION_FIELD_AUTHZ_FAILURE, DisplayName = "When create permission is there at entity level but all the fields are excluded using wildcard")]
+        [DataRow("only-update-role", MutationTests.USER_NOT_AUTHORIZED, DisplayName = "When create permission is NOT there")]
```
seantleonard, line 268: > When create permission is NOT there

To finish the thought: what is the expectation when create permission is not there?

seantleonard (commented):
some suggestions on more informative displaynames for tests.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -912,7 +913,7 @@ public async Task TestQueryFilterFieldAuth_UnauthorizedField()
                     }
                 }
             }";
-            string clientRoleHeader = AuthorizationType.Anonymous.ToString();
+            string clientRoleHeader = "limited-read-role";
```
seantleonard, line 916: by changing the role from anonymous to `limite-read-role` you're changing the behavior of the test. The test description looks to be describing this test as validating behavior for the `anonymous` role.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -970,7 +971,8 @@ public async Task TestQueryFilterNestedFieldAuth_AuthorizedNestedField()
                 }
             }";
 
-            JsonElement actual = await ExecuteGraphQLRequestAsync(_graphQLQueryName, query: gqlQuery);
+            string authToken = AuthTestHelper.CreateStaticWebAppsEasyAuthToken(specificRole: AuthorizationType.Authenticated.ToString());
+            JsonElement actual = await ExecuteGraphQLRequestAsync(_graphQLQueryName, query: gqlQuery, authToken: authToken);
```
seantleonard, line 975: by changing the role from anonymous to `authenticated` you're changing the behavior of the test. The test description looks to be describing this test as validating behavior for the anonymous role.

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForCosmos.verified.txt
```diff
@@ -271,18 +271,22 @@
         },
         Permissions: [
           {
-            Role: anonymous,
+            Role: field-mutation-with-read-permission,
```
seantleonard, line 274: why not keep `anonymous` role  and add `field-mutation-with-read-permission` as a distinct role? that way you preserve what some tests may be validating.

src/Service.Tests/CosmosTests/MutationTests.cs
seantleonard: can extrapolate comments on this test's displaynames to the other test's display names.

src/Service.Tests/CosmosTests/MutationTests.cs
sourabh1007: I added that while testing item level auth..but you are right not related to this PR.

src/Service.Tests/CosmosTests/MutationTests.cs
sourabh1007: Renamed the display name.

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForCosmos.verified.txt
sourabh1007: actually with `anonymous` role..test cases were failing over to that role if read permission is not there for Mutation.

sourabh1007: I can change the entity to something else..to preserve these scenarios. IMO anonymous has nothing special, its just a default role.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: This test validate if include/exclude works or not. I am not seeing any specific scenario for `anonymous` here. I think description is misleading.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: I changed anonymous to `limited-read-role`. I think due to positioning, comparison is coming like this. `authenticated` role is as it is.

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForCosmos.verified.txt
sourabh1007: please refer other comments.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -321,11 +340,87 @@ public async Task UpdateItemWithUnauthorizedWildCardReturnsError()
...
+            }
+        }
+
+        /// <summary>
+        /// Delete Mutation performed on the fields with different auth permissions
+        /// It throws permission denied error if role doesn't have permission to perform the operation
+        /// </summary>
+        [TestMethod]
+        [DataRow("field-mutation-with-read-permission", MutationTests.NO_ERROR_MESSAGE, DisplayName = "AuthZ success and blank response for delete mutation because of reference to excluded/disallowed fields.")]
+        [DataRow("authenticated", MutationTests.NO_ERROR_MESSAGE, DisplayName = "AuthZ success and blank response when role has no delete operation restrictions.")]
```
seantleonard, line 370: isn't clear how these two tests are different. Even though the comments allude to the role restrictions being different, the response is blank for both tests. Is the response being blank for both tests intended?

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -1009,6 +1011,32 @@ public async Task TestQueryFilterNestedFieldAuth_UnauthorizedNestedField()
             Assert.IsTrue(errorMessage.Contains(DataApiBuilderException.GRAPHQL_FILTER_FIELD_AUTHZ_FAILURE));
         }
 
+        [TestMethod]
```
seantleonard, line 1014: need comment elaborating on what is actually being validate: which nested query filter field is disallowed?

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -252,62 +255,78 @@ public async Task MutationMissingRequiredPartitionKeyValueReturnError()
...
-        /// for entity 'earth'
+        /// Create Mutation performed on the fields with different auth permissions
+        /// It throws permission denied error if role doesn't have permission to perform the operation
         /// </summary>
         [TestMethod]
-        public async Task CanCreateItemWithAuthorizedFields()
+        [DataRow("field-mutation-with-read-permission", DataApiBuilderException.GRAPHQL_MUTATION_FIELD_AUTHZ_FAILURE, DisplayName = "AuthZ failure for create mutation because of reference to excluded/disallowed fields.")]
+        [DataRow("authenticated", MutationTests.NO_ERROR_MESSAGE, DisplayName = "AuthZ success when role has no create/read operation restrictions.")]
+        [DataRow("only-create-role", "The mutation operation createEarth was successful " +
+            "but the current user is unauthorized to view the response due to lack of read permissions", DisplayName = "AuthZ failure  but sucessful operation  where role has ONLY create permission and NO read permission.")]
```
seantleonard, line 265:
```suggestion
            "but the current user is unauthorized to view the response due to lack of read permissions", DisplayName = "Successful create operation but AuthZ failure for read when role has ONLY create permission and NO read permission.")]
```

seantleonard (commented):
last nits, then should be good. Thanks for addressing feedback!

src/Service.Tests/CosmosTests/MutationTests.cs
sourabh1007: @sajeetharan is already reviewing these tests. These tests are showing current state of the authorization feature in cosmos db. If anything needs to be changed, I will change it is separate PR.

seantleonard: That's fine he is reviewing the tests. I'm not suggesting a change needs to be made here though would like to know how the tests are different. 

I don't understand what restrictions are present just from reading this:
>   [DataRow("field-mutation-with-read-permission", MutationTests.NO_ERROR_MESSAGE, DisplayName = "AuthZ success and blank response for delete mutation because of reference to excluded/disallowed fields.")]

are the "because of reference to excluded/disallowed fields." on the read action or the delete action?

sourabh1007: Yes..if `exclude`and `include` has any effect on operation

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: Added summary on the Test. I didn't put `display name` intentionally, because in this file, none of the test has display name.

seantleonard: makes sense. We only use `DisplayName` for `[DataRow]` tests.

seantleonard: The updated comment here is very helpful :)

seantleonard (approved):
Thanks for addressing feedback to make tests/comments add clarity. Clarity will be helpful for us to understand context when we or other devs come back to read this in future.

PR #2040, opened 2024-02-15
fix dataSourceName not found exception in `dab validate`

### LINKED ISSUE #2039: [Bug]: `dab validate` getting exception due to datasourceName mismatch
### What happened?

image

### Version

0.11.6-rc

Azure SQL

Local (including CLI)

_No response_

### Relevant log output

_No response_

### Code of Conduct

abhishekkumams:
## Why make this change?

-  Closes #2039 
- `dab validate` was throwing exception stating `dataSourcename is not found`. This PR aims to fix it.

## What is this change?
- RuntimeConfig object stores dataSourceName information.
- previously the RuntimeConfig object was created in CLI during deserialization and sent to Engine Code. And In engine code when we setup metadatprovider we again create the dataSourceName info, due to which there is a mismatch.
- So, we are not sending the RuntimeConfig object from CLI, and in Engine we are sourcing it directly from RuntimeConfigProvider. (similar to how it's done while starting the engine).
- Therefore, Doing the deserialization only in a single place in the engine itself, improving performance as well.

# Before
image

## After
image


src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -150,10 +150,20 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
     /// <returns>true if no validation failures, else false.</returns>
     public async Task<bool> TryValidateConfig(
         string configFilePath,
-        RuntimeConfig runtimeConfig,
```
seantleonard, line 153: need to update method summary with removed param

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -150,10 +150,20 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
...
         ILoggerFactory loggerFactory,
         bool isValidateOnly = false)
     {
+        RuntimeConfig runtimeConfig;
+        try
+        {
+            runtimeConfig = _runtimeConfigProvider.GetConfig();
```
seantleonard, line 159: `_runtimeConfigProvider.GetConfig()` parses the config and doesn't just returned an already processed runtimeconfig object?

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -60,4 +60,20 @@ public void TestConfigWithInvalidConfigProperties()
         bool isConfigValid = ConfigGenerator.IsConfigValid(validateOptions, _runtimeConfigLoader!, _fileSystem!);
         Assert.IsFalse(isConfigValid);
     }
+
+    /// <summary>
+    /// This method validates that the IsConfigValid method returns false when the config is empty.
+    /// This is to test no exception gets thrown with validate during failure config deserialization.
```
seantleonard, line 66: how are you testing that no exception is thrown? By ensuring `ConfigGenerator.IsConfigValid(...)` only returns a bool?

src/Cli/ConfigGenerator.cs
```diff
@@ -1091,24 +1091,14 @@ public static bool IsConfigValid(ValidateOptions options, FileSystemRuntimeConfi
...
-            }
+            _logger.LogInformation("Validating config file: {runtimeConfigFile}", runtimeConfigFile);
 
             RuntimeConfigProvider runtimeConfigProvider = new(loader);
 
             ILogger<RuntimeConfigValidator> runtimeConfigValidatorLogger = LoggerFactoryForCli.CreateLogger<RuntimeConfigValidator>();
             RuntimeConfigValidator runtimeConfigValidator = new(runtimeConfigProvider, fileSystem, runtimeConfigValidatorLogger, true);
 
-            return runtimeConfigValidator.TryValidateConfig(runtimeConfigFile, deserializedRuntimeConfig, LoggerFactoryForCli).Result;
+            return runtimeConfigValidator.TryValidateConfig(runtimeConfigFile, LoggerFactoryForCli).Result;
```
seantleonard, line 1101: In your PR desription:
> So, we are not sending the RuntimeConfig object from CLI, and in Engine we are sourcing it directly from RuntimeConfigProvider. (similar to how it's done while starting the engine).
Therefore, Doing the deserialization only in a single place in the engine itself, improving performance as well.

Those two changes in behavior hold true because `runtimeConfigValidator.TryValidateConfig(runtimeConfigFile, LoggerFactoryForCli).Result;` is "engine code" and CLI no longer passes its own `deserializedRuntimeConfig` object?  just confirming my understanding.

src/Cli.Tests/ValidateConfigTests.cs
seantleonard: in `IsConfigValid` there is a call to `runtimeConfigValidator.TryValidateConfig(...)` which calls `ValidateEntitiesMetadata()`. Could  `ValidateEntitiesMetadata()` throw an exception which resuls in a bool not being returned for `IsConfigValid`? I suppose this PR's change only affects the mismatched dataSourceName, but unclear if it prevents other exceptions from being printed to the console

src/Cli/ConfigGenerator.cs
abhishekkumams: correct.

src/Cli.Tests/ValidateConfigTests.cs
abhishekkumams: correct, this PR is just to fix the issue of dataSourceName not found. 
Since, I moved the deserialization logic to a single place, I added this test to make sure exceptions due to deserialization failures are captured correctly.

src/Core/Configurations/RuntimeConfigValidator.cs
abhishekkumams: it does if same instance is used. First, CLI was trying to deserialize the config which creates the dataSourceName. Now, In the engine during validation, when we create a new instance for MetadataProviderFactory, it also creates a mapping between DataSourceName and metadataProvider. This dataSourceName differs from the one created in CLI. According to our code structure `MetadataProviderFactory` should be setup first and then used to get the config which keeps thing in sync.

For your question, yes it won't parse the config again, but the mapping that exists in metadataProviderFactory will differ, if we don't first setup metadataProvider and then source config.
image

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -144,16 +144,25 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
...
+        }
```
Aniruddh25, line 159: Shouldn't we be using _runtimeConfigProvier.TryGetConfig?

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -60,4 +60,20 @@ public void TestConfigWithInvalidConfigProperties()
...
+    /// </summary>
+    [TestMethod]
+    public void TestValidateWithEmptyConfig()
```
Aniruddh25, line 69: Shouldn't we add another test case which would have failed with the datasourceName not found exception without your fix?

Aniruddh25 (changes requested):
Needs the right testing

src/Cli.Tests/ValidateConfigTests.cs
abhishekkumams: added a test which test the whole deserilization to validation.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1062,6 +1062,45 @@ public void TestConfigPropertiesAreValid()
             configValidator.ValidateConfigProperties();
         }
 
+        /// <summary>
+        /// This method tests that config file is validated correctly and no exceptions are thrown.
+        /// </summary>
+        [TestMethod("Validates the complete config."), TestCategory(TestCategory.MSSQL)]
+        public async Task TestConfigIsValid()
```
Aniruddh25, line 1069: Why is this not added to `ConfigValidationUnitTests.cs`?

Aniruddh25 (approved):

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: this is an integration test and all the integration tests related to validations are in this class.

src/Cli.Tests/ValidateConfigTests.cs
severussundar, line 66:
```suggestion
    /// This is to validate that no exceptions are thrown with validate for failures during config deserialization.
```

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -60,4 +60,20 @@ public void TestConfigWithInvalidConfigProperties()
...
+    {
+        // create an empty config file
+        ((MockFileSystem)_fileSystem!).AddFile(TEST_RUNTIME_CONFIG_FILE, string.Empty);
+
+        ValidateOptions validateOptions = new(TEST_RUNTIME_CONFIG_FILE);
+
+        bool isConfigValid = ConfigGenerator.IsConfigValid(validateOptions, _runtimeConfigLoader!, _fileSystem!);
```
severussundar, line 76: Maybe enclose this statement within a try-catch block to handle any exceptions (if they get thrown) and fail the test within the catch block?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1062,6 +1062,45 @@ public void TestConfigPropertiesAreValid()
...
+        {
+            TestHelper.SetupDatabaseEnvironment(MSSQL_ENVIRONMENT);
+            DataSource dataSource = new(DatabaseType.MSSQL,
```
severussundar, line 1072: Could we include a test case where the dataSource is invalid in the integration test as well

abhishekkumams: those tests already exists

src/Cli.Tests/ValidateConfigTests.cs
abhishekkumams: updated

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: i've also updated the test to use the main integration test config

PR #2042, opened 2024-02-16
Resolve `uniqueidentifier` database type from `dab-config.json`  defined stored proc parameter default values.

### LINKED ISSUE #2027: [Bug]: Parse sp parameter default value from runtime config when target type is UUID.
### What happened?

A bug happened!

When default sp param value is defined in dab-config.json, and then a guid is provided, DAB/HC does not resolve the value to a UUID. here are the at fault code sections which need to be addressed:

https://github.com/Azure/data-api-builder/blob/a102e44877b66ca78f6601d78f47dd02481c3416/src/Service.GraphQLBuilder/GraphQLUtils.cs#L243-L258

https://github.com/Azure/data-api-builder/blob/a102e44877b66ca78f6601d78f47dd02481c3416/src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs#L51-L56

### Version

main/latest

Azure SQL

Local (including CLI)

GraphQL

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
## Why make this change?

- Closes #2027. DAB would not start up correctly when runtime config took the following form:
```json
  "entities": {
    "SpUuidParam": {
      "source": {
        "object": "sp_uuid_param",
        "type": "stored-procedure",
        "parameters": {
          "param1": "f58b7b58-62c9-4b97-ab60-75de70793f66"
        }
      },
      "graphql": {
        "enabled": true,
        "operation": "Query",
        "type": {
          "singular": "SpUuidParam",
          "plural": "SpUuidParams"
        }
      },
      "rest": {
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [ "*" ]
        },
        {
          "role": "authenticated",
          "actions": [ "*" ]
        }
      ]
    }
```

And stored proc in tsql
```tsql
CREATE PROCEDURE [dbo].[sp_uuid_param] @param1 uniqueidentifier AS
      SELECT @param1 AS [ReturnPayload]
```

DAB was lacking the ability to handle the stored procedure having an input parameter with value type `uniqueidentifier`. When a default value was defined in the config, that value would fail conversion to the UUID type during GraphQL schema creation.

The call stack flows through:

```csharp
                    if (entity.Source.Parameters is not null && entity.Source.Parameters.TryGetValue(param, out object? value))
                    {
                        Tuple<string, IValueNode> defaultGraphQLValue = ConvertValueToGraphQLType(value.ToString()!, parameterDefinition: spdef.Parameters[param]);
                        defaultValueNode = defaultGraphQLValue.Item2;
                    }
```

where `ConvertValueToGraphQLType(...)` would attempt to convert the config defined value to a GraphQL type based on the type inferred from the parameter's SystemType (System.Guid). The parameter object has the following properties when the parameter is passed to that function:

- SystemType -> `System.Guid`
- DbType -> `Guid`
- ConfigDefaultValue -> `f58b7b58-62c9-4b97-ab60-75de70793f66`
- HasConfigDefault -> `true`

`ConvertValueToGraphQLType(...)` did not have the conversion needed to create a UUID type.

## What is this change?

- In the function called to process config defined default values for stored procedure parameters, `ConvertValueToGraphQLType(...)` add the conversion:

```csharp
UUID_TYPE => new(UUID_TYPE, new UuidType().ParseValue(Guid.Parse(defaultValueFromConfig))),
```

- Moved `ConvertValueToGraphQLType()` from `GraphQLUtils.cs` to `GraphQLStoredProcedureBuilder.cs` to align with usage and purpose of function. Reduces size of MEGA UTIL class. Also adds comment for `GraphQLUtils.BuiltInTypes` hashset entry 'ID' to inform that it enables CosmosDB functionality as only cosmos tests failed when that entry was commented out. 
- Reorganizes the ConvertValueToGraphQLType() order of types in switch statements to align with GraphQLUtils.BuiltInTypes hashset. That way it is easier to notice discrepancies in the two lists.

## How was this tested?

- [x] Integration Tests

## Sample Request(s)

- Use the db schema and entity config provided above. Startup will succeed without conversion errors when attempting to create the GraphQL schema.


src/Service.Tests/GraphQLBuilder/Sql/StoredProcedureBuilderTests.cs
```diff
@@ -56,6 +56,8 @@ public class StoredProcedureBuilderTests
         [DataRow(typeof(DateTimeOffset), DATETIME_TYPE, "11/19/2012 10:57:11 AM -08:00", false, DisplayName = "DateTimeOffset")]
         [DataRow(typeof(TimeOnly), LOCALTIME_TYPE, "10:57:11.0000", false, DisplayName = "LocalTime")]
         [DataRow(typeof(byte[]), BYTEARRAY_TYPE, "AgQGCAoMDhASFA==", false, DisplayName = "Byte[]")]
+        [DataRow(typeof(Guid), UUID_TYPE, "f58b7b58-62c9-4b97-ab60-75de70793f66", false, DisplayName = "GraphQL UUID/ SystemType GUID")]
```
severussundar, line 59: The [HotChocolate docs](https://chillicream.com/docs/hotchocolate/v12/defining-a-schema/scalars#uuid-type) specifies 5 supported formats for UUID type. 

It would be good to have test cases with all the formats.

severussundar (commented):
Suggestion to include test cases with different UUID supported formats. Otherwise, looks good.

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -248,6 +248,7 @@ public static ObjectType UnderlyingGraphQLEntityType(IType type)
                     INT_TYPE => new(INT_TYPE, new IntValueNode(int.Parse(defaultValueFromConfig))),
                     LONG_TYPE => new(LONG_TYPE, new IntValueNode(long.Parse(defaultValueFromConfig))),
                     STRING_TYPE => new(STRING_TYPE, new StringValueNode(defaultValueFromConfig)),
+                    UUID_TYPE => new(UUID_TYPE, new UuidType().ParseValue(Guid.Parse(defaultValueFromConfig))),
```
Aniruddh25, line 251: Are there any other expected **supported** types that might be missing handling in this conversion? We should at least identify what else is missing.. 

Its could be a different issue to add new unsupported types but need to ensure all types that we claim to support are supported everywhere - be it stored proc parameter or view column type

Aniruddh25 (approved):
LGTM, thanks for the quick fix!

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
seantleonard, line 1: this file is not new code. It is just moved from GraphQLUtils.cs because it is only applicable in this file and I wanted to reduce size of mega utility class. (God class antipattern).

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -122,5 +129,57 @@ public static FieldDefinitionNode GetDefaultResultFieldForStoredProcedure()
...
+                    LONG_TYPE => new(LONG_TYPE, new IntValueNode(long.Parse(defaultValueFromConfig))),
+                    SINGLE_TYPE => new(SINGLE_TYPE, new SingleType().ParseValue(float.Parse(defaultValueFromConfig))),
+                    FLOAT_TYPE => new(FLOAT_TYPE, new FloatValueNode(double.Parse(defaultValueFromConfig))),
+                    DECIMAL_TYPE => new(DECIMAL_TYPE, new FloatValueNode(decimal.Parse(defaultValueFromConfig))),
+                    STRING_TYPE => new(STRING_TYPE, new StringValueNode(defaultValueFromConfig)),
+                    BOOLEAN_TYPE => new(BOOLEAN_TYPE, new BooleanValueNode(bool.Parse(defaultValueFromConfig))),
+                    DATETIME_TYPE => new(DATETIME_TYPE, new DateTimeType().ParseResult(
+                        DateTime.Parse(defaultValueFromConfig, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal))),
+                    BYTEARRAY_TYPE => new(BYTEARRAY_TYPE, new ByteArrayType().ParseValue(Convert.FromBase64String(defaultValueFromConfig))),
+                    LOCALTIME_TYPE => new(LOCALTIME_TYPE, new LocalTimeType().ParseResult(LocalTimePattern.ExtendedIso.Parse(defaultValueFromConfig).Value)),
```
seantleonard, line 165: reordered these to be same order as GraphQLUtils.BuiltInTypes hashset.

src/Service.GraphQLBuilder/GraphQLUtils.cs
```diff
@@ -225,56 +219,6 @@ public static ObjectType UnderlyingGraphQLEntityType(IType type)
...
-                    BOOLEAN_TYPE => new(BOOLEAN_TYPE, new BooleanValueNode(bool.Parse(defaultValueFromConfig))),
-                    SINGLE_TYPE => new(SINGLE_TYPE, new SingleType().ParseValue(float.Parse(defaultValueFromConfig))),
-                    FLOAT_TYPE => new(FLOAT_TYPE, new FloatValueNode(double.Parse(defaultValueFromConfig))),
-                    DECIMAL_TYPE => new(DECIMAL_TYPE, new FloatValueNode(decimal.Parse(defaultValueFromConfig))),
-                    DATETIME_TYPE => new(DATETIME_TYPE, new DateTimeType().ParseResult(
-                        DateTime.Parse(defaultValueFromConfig, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal))),
-                    BYTEARRAY_TYPE => new(BYTEARRAY_TYPE, new ByteArrayType().ParseValue(Convert.FromBase64String(defaultValueFromConfig))),
-                    LOCALTIME_TYPE => new(LOCALTIME_TYPE, new LocalTimeType().ParseResult(LocalTimePattern.ExtendedIso.Parse(defaultValueFromConfig).Value)),
-                    _ => throw new NotSupportedException(message: $"The {defaultValueFromConfig} parameter's value type [{paramValueType}] is not supported.")
```
seantleonard, line 259: this code is not deleted fully, just moved to GraphQLStoredProcedureBuilder class.

src/Service.Tests/GraphQLBuilder/Sql/StoredProcedureBuilderTests.cs
seantleonard: I think this belongs in a separate work item because the tsql uuid type doesn't seem to support the other formats that HotChocolate specifies: https://learn.microsoft.com/en-us/sql/t-sql/data-types/uniqueidentifier-transact-sql?view=sql-server-ver16

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -122,5 +129,57 @@ public static FieldDefinitionNode GetDefaultResultFieldForStoredProcedure()
...
+        /// <summary>
+        /// Translates a JSON string or number value defined as a stored procedure's default value
+        /// within the runtime configuration to a GraphQL {Type}ValueNode which represents
+        /// the associated GraphQL type. The target value type is referenced from the passed in parameterDefinition which
+        /// holds database schema metadata.
+        /// </summary>
+        /// <param name="defaultValueFromConfig">String representation of default value defined in runtime config.</param>
+        /// <param name="parameterDefinition">Database schema metadata for stored procedure parameter which include value and value type.</param>
+        /// <returns>Tuple where first item is the string representation of a GraphQLType (e.g. "Byte", "Int", "Decimal")
+        /// and the second item is the GraphQL {type}ValueNode </returns>
```
aaronburtle, line 142:
```suggestion
        /// and the second item is the GraphQL {type}ValueNode.</returns>
```

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -122,5 +129,57 @@ public static FieldDefinitionNode GetDefaultResultFieldForStoredProcedure()
...
+                    DATETIME_TYPE => new(DATETIME_TYPE, new DateTimeType().ParseResult(
+                        DateTime.Parse(defaultValueFromConfig, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal))),
+                    BYTEARRAY_TYPE => new(BYTEARRAY_TYPE, new ByteArrayType().ParseValue(Convert.FromBase64String(defaultValueFromConfig))),
+                    LOCALTIME_TYPE => new(LOCALTIME_TYPE, new LocalTimeType().ParseResult(LocalTimePattern.ExtendedIso.Parse(defaultValueFromConfig).Value)),
+                    _ => throw new NotSupportedException(message: $"The {defaultValueFromConfig} parameter's value type [{paramValueType}] is not supported.")
+                };
+
+                return valueNode;
+            }
+            catch (Exception error) when (
```
aaronburtle, line 171: What happens if we get an exception that is not of one of these error types?

aaronburtle (approved):
Looks good!

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
seantleonard: I tried it out by stepping through debugger and throwing a different type of exception. This resulted in hard to decipher stack trace and uncaught error during dab startup. 
image

Good question and I modified this to no longer filter on certain exception types, because any issue with processing the value here should result in an easier to read exception.

When this catches all exception types, the message is easier to read: 
image

though i do admit we should probably be catching this exception elsewhere too so that we can eliminate long unnecessary stack traces printed to the console.

seantleonard: my intent with filtering certain exceptions is that I'd only want to emit the "the <param_name> parameter's value type [Type] is not supported." message when an actual parse error occurred. Other exceptions wouldn't necessarily fit to this message.

seantleonard: updated to no longer filter on specific exception type.

aaronburtle: thanks, good info here!

PR #2043, opened 2024-02-16
Fixing nit comments from previous DW mutation support prs.

rohkhann:
## Why make this change?

Fixing nit comments from earlier pr: https://github.com/Azure/data-api-builder/pull/1978

## What is this change?

There were some nit comments to fix while adding implementation for mutation support for DW.

## How was this tested?

Existing tests should cover as this pr only handles nit fixes.


Aniruddh25 (approved):
LGTM, thank you for following up and ensuring all the comments are resolved even after merging the PR!

seantleonard (approved):
thanks for the follow up!

Issue #1982, opened 2024-01-19
REST support for datawarehouse insert/put/patch requests.


rohkhann:
Expose support for rest calls to datawarehouse for create update and delete scenarios.

PR #2048, opened 2024-02-16
Stored proc support test cases REST DW.

rohkhann:
## Why make this change?
This change adds appropriate test cases that validates the REST support for stored procedures for Data warehouse.

## What is this change?
This change adds test cases needed to validate stored proc rest support in dab. It only adds test cases currently supported for other db types.
1. Find: For find, we remove the overriden test cases in the dab class. this means the code will execute the test cases in the base classes for DW. 
2. Delete: Add the delete class for dwtests that inherits off of the base. (delete support exists for rest dw case as for delete we return no content as opposed to insert). The specific test cases that need to be overriden due to DW specific implementations have been overriden and added to the delete test cases file. 
3. PUT/PATCH: There are no stored proc specific tests for PUT/PATCH for other db's and hence none added here. Tracked in this issue: https://github.com/Azure/data-api-builder/issues/2085
4. Insert: There is one test case pertaining to stored proc insert that is added for other db's. That is not included in this pr, as since we would want to inherit from the InsertTestBase class, other insert tests would also execute. As compared to delete, insert on rest, currently expects the inserted object to be returned. In DW, we have gone with option of only returning success/failure for mutation operations instead of the result and that logic + test cases will be implemented in https://github.com/Azure/data-api-builder/issues/1982

## How was this tested?
test cases added.


src/Service.Tests/SqlTests/RestApiTests/Delete/DwSqlDeleteApiTest.cs
```diff
@@ -0,0 +1,89 @@
...
+                $"WHERE id = 14"
+            }
+        };
+        #region Test Fixture Setup
+
+        /// <summary>
+        /// Sets up test fixture for class, only to be run once per test run, as defined by
+        /// MSTest decorator.
+        /// </summary>
+        /// <param name="context"></param>
```
aaronburtle, line 32: nit: if the param doesnt need a description it probably doesnt need to be listed since we can see it in the method sig.

aaronburtle (approved):
LGTM!

src/Service.Tests/SqlTests/RestApiTests/Find/DwSqlFindApiTests.cs
```diff
@@ -546,7 +545,15 @@ public class DwSqlFindApiTests : FindApiTestBase
                 $"FROM {_integrationTableName} " +
                 $"ORDER BY [publisher_id] ASC, [id] ASC " +
                 $"FOR JSON PATH, INCLUDE_NULL_VALUES"
-            }
+            },
```
seantleonard, line 548: nit: there's an extra newline on L341, can get that as part of this change.

src/Service.Tests/SqlTests/RestApiTests/Delete/DwSqlDeleteApiTest.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        #endregion
+
+        [TestMethod]
+        public async Task DeleteOneInViewBadRequestTest()
+        {
+            string expectedErrorMessage = $"View or function '{_defaultSchemaName}.{_composite_subset_bookPub}' is not updatable " +
+                                           "because the modification affects multiple base tables.";
+            await base.DeleteOneInViewBadRequestTest(expectedErrorMessage);
+        }
```
seantleonard, line 56: was this just a gap for the 'delete' case in testing REST DW? or do you also need to ensure DW view tests are added for find/insert/update?

seantleonard (commented):
you have Find,Delete, and Insert. Do you plan to add DW StoredProc tests for Patch/Put?

src/Service.Tests/SqlTests/RestApiTests/Delete/DwSqlDeleteApiTest.cs
```diff
@@ -0,0 +1,88 @@
...
+
```
Aniruddh25, line 57: It seems like DW tests are still missing for Patch/Put - those are required for complete REST APIs support for DW.

seantleonard: we don't have these types of tests for SQL either, I propose an issue to track and follow-up because as-is, this PR is in parity with SQL. 
@rohkhann , Please check at least what the behavior is when runtime config defines stored proc to be PUT/PATCH
```json
"rest": {
  "methods": [ "GET", "POST" ]
}
```
per https://learn.microsoft.com/en-us/azure/data-api-builder/views-and-stored-procedures#rest-support-for-stored-procedures

src/Service.Tests/SqlTests/RestApiTests/Find/DwSqlFindApiTests.cs
rohkhann: fixed.

src/Service.Tests/SqlTests/RestApiTests/Delete/DwSqlDeleteApiTest.cs
rohkhann: We didnt have delete tests for rest dw. I just added whatever coverage we have for the other db's.

seantleonard (approved):
thanks for making these changes and good to see follow up issue for PUT/PATCH.

Aniruddh25 (approved):

src/Service.Tests/SqlTests/RestApiTests/Find/DwSqlFindApiTests.cs
```diff
@@ -551,7 +549,15 @@ public class DwSqlFindApiTests : FindApiTestBase
...
+            {
+                "FindManyStoredProcedureTest",
```
Aniruddh25, line 554: Looks like some tests are failing for DW, since this method might be returning false: 

`RestService::IsHttpMethodAllowedForStoredProcedure`

You may need to modify the config file or check why this is returning false for StoredProcedure on DW.

Aniruddh25 (changes requested):
Need to fix test cases

src/Service.Tests/SqlTests/RestApiTests/Find/DwSqlFindApiTests.cs
rohkhann: Fixed this. seemed to be a missing entity. I added that to config and ran tests locally and all passed.

src/Service.Tests/SqlTests/RestApiTests/Delete/DwSqlDeleteApiTest.cs
rohkhann: We have one test for insert. Since the format we are using is insertTest will inherit from the base - this means all base tests are run. For DW, since insert does not return the resultant rows (we went with a DBOperationResult return type in GQL) Have to see similar for REST (how do we handle no data being returned from a mutation in rest scenario). Hence I am removing the insert test for now and will handle it once I add the logic for mutation support all up for dw for rest scenario. this should atleast unblock this pr and add required stored proc support

src/Service.Tests/SqlTests/RestApiTests/Insert/DwSqlInsertApiTests.cs
```diff
@@ -1,274 +0,0 @@
...
-using System.Collections.Generic;
-using System.Net;
-using System.Threading.Tasks;
-using Azure.DataApiBuilder.Config.ObjectModel;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-
-namespace Azure.DataApiBuilder.Service.Tests.SqlTests.RestApiTests.Insert
-{
-    /// <summary>
-    /// Test REST Apis validating expected results are obtained.
```
Aniruddh25: Why are we removing Insert tests for DW? In the config, you can keep the allowed rest methods to be `get`, `post` instead of just `get`, right?

seantleonard: Rohan's reply: https://github.com/Azure/data-api-builder/pull/2048#discussion_r1520500939

rohkhann: Please let me know if that is ok and we can move with completing this pr.

Aniruddh25 (approved):
LGTM, thanks for adding the test gap and uncovering additional missing gaps as tracked by the new issues

PR #2049, opened 2024-02-17
Create a unittest pipeline, moving common steps into templates, disable CosmosDB testing on linux

Aniruddh25:
## Why make this change?

- Trying to improve the pipeline triggering experience by moving all PR validation pipelines to public project. 
- Desired goal is to have all pipelines automatically trigger upon commenting `/azp run` on an approved PR by a team member

## What is this change?

- Create a unittest pipeline to run unittests
- Move common steps and variables from original build pipeline to templates
- Remove testing on the linux platform for CosmosDB in preparation to run this pipeline in the public project too. Its been difficult to start a CosmosDB emulator on linux, preventing us to test on this platform in the public project where pipeline secrets aren't allowed for PRs from forked repositories. (they are allowed in private projects, hence we had been able to test on linux when the pipeline was in the private project) 

## How was this tested?

- Temporarily renamed the unittest-pipelines.yml as build-pipelines.yml and ran the `Build` pipeline on this branch to verify a successful build

image


.pipelines/templates/build-pipelines.yml
abhishekkumams, line 1: since it only has unit tests. can we name this file to something like unit-tests.yml ?

Aniruddh25: its the common base build that the `build-pipelines.yml` in the `.pipelines` folder imports as well. I plan to rename the previous build-pipelines to release-pipelines.yml and keep this as the build-pipelines.yml

There is a separate unittest-pipeline.yml

.pipelines/templates/build-pipelines.yml
```diff
@@ -0,0 +1,106 @@
...
+# will be used to do code signing so we need to do this first.
+- task: DotNetCoreCLI@2
+  displayName: Build
+  inputs:
+    command: custom
+    custom: pack
+    projects: '**/*.sln'
+    arguments: '--configuration $(buildConfiguration) -p:Version=$(dabVersion)' 
+
+- task: DotNetCoreCLI@2
```
severussundar, line 71: The steps following this could be moved to `unittest-pipelines.yml` as they are related to unit tests - Running the unit tests, publishing snapshot files and code coverage report.

abhishekkumams: it is already running in unit tests only as it's included in task step of `unittest-pipelines.yml`

abhishekkumams (approved):

ayush3797 (approved):

.pipelines/templates/build-pipelines.yml
severussundar: I got that the tests are running in unit testing pipeline. I thought it would be better to have the unit test related steps in unit test related yaml file - this will avoid running unit tests in all the places (in future if the template is to be re-used in integration testing pipelines). 

However, if needed, this change seems small and can be incorporated in the future. This needn't block this PR from getting merged. So, resolving the conversation.

Issue #1940, opened 2023-12-21
[Bug]: T-SQL Generated for a Self-Joining Table is Invalid


CalvinQuark:
### What happened?

The T-SQL generated for a self-joining table relationship is invalid. Queries return `null` rather than the related field's value. For example, consider the `dbo.DimAccount` table from the [AdventureWorks](https://github.com/microsoft/sql-server-samples/tree/master/samples/databases/adventure-works) Microsoft sample database.

```sql
USE [AdventureWorksDW];
GO

CREATE TABLE [dbo].[DimAccount] (
    [AccountKey]                    [INT]           IDENTITY(1, 1) NOT NULL,
    [ParentAccountKey]              [INT]           NULL,
    [AccountCodeAlternateKey]       [INT]           NULL,
    [ParentAccountCodeAlternateKey] [INT]           NULL,
    [AccountDescription]            [NVARCHAR](50)  NULL,
    [AccountType]                   [NVARCHAR](50)  NULL,
    [Operator]                      [NVARCHAR](50)  NULL,
    [CustomMembers]                 [NVARCHAR](300) NULL,
    [ValueType]                     [NVARCHAR](50)  NULL,
    [CustomMemberOptions]           [NVARCHAR](200) NULL,
    CONSTRAINT [PK_DimAccount]
        PRIMARY KEY CLUSTERED ([AccountKey] ASC)
);
GO

ALTER TABLE [dbo].[DimAccount] WITH CHECK
ADD CONSTRAINT [FK_DimAccount_DimAccount]
    FOREIGN KEY ([ParentAccountKey])
    REFERENCES [dbo].[DimAccount] ([AccountKey]);
GO

ALTER TABLE [dbo].[DimAccount] CHECK CONSTRAINT [FK_DimAccount_DimAccount];
GO
```

When the `dab-config.json` file is configured as:

```json
{
    "$schema": "https://github.com/Azure/data-api-builder/releases/download/v0.9.7/dab.draft.schema.json",
    "data-source": {
        "database-type": "mssql",
        "connection-string": "Data Source=localhost;Initial Catalog=AdventureWorksDW;Integrated Security=True;TrustServerCertificate=True",
        "options": { "set-session-context": false }
    },
    "runtime": {
        "rest": { "enabled": false },
        "graphql": { "enabled": true, "path": "/graphql" },
        "host": {
            "cors": {
                "origins": [ "http://localhost:5500" ],
                "allow-credentials": false
            },
            "authentication": {
[...]

Issue #1935, opened 2023-12-20
Validations required for source/target fields in relationships


ayush3797:
Consider these scenarios for source/target fields defined for a relationship:

1. `SourceFields`: **NULL**, TargetFields: **Non-NULL** (and vice-versa)
2. `SourceFields.size() ! = TargetFields.size()`
3. `SourceFields`/`TargetFields` contains one or more fields/columns which do not belong to the source/target entities respectively.

All these scenarios should be caught during validation.

severussundar:
Validations can be added to check that each source, target field specified is a valid column in the underlying database object.

PR #2051, opened 2024-02-19
Removing wrong FK definition 

ayush3797:
## Why make this change?
If there is a relationship defined for a (source,target) in the database with right cardinality as `ONE`, we add the relationship in both the directions (source->target and target->source) to the source's `RelationshipMetadata` corresponding to the target entity's entry in `RelationshipMetadata.TargetEntityToFkDefinitionMap` dictionary. 

We do this because the relationship can be 1:1 or N:1. In case of N:1, the first FK (from source->target) would be present. In case of 1:1, either or both of them can be present. But all this is known to us only at a later stage when we collect metadata from the database and actually infer which of the two relationships (or both) is/are valid.

After validating the correct relationship, we don't remove the other relationship which we added earlier. The redundant relationship jadds a redundant `JOIN` condition (I think this is something which is visible in the sql query generated by DAB highlighted in this issue: https://github.com/Azure/data-api-builder/issues/1940).

## What is this change?
The functioning of `SqlMetadatProvider.FillInferredFkInfo()` is changed such that for all the FK definitions present for a target entity in the source entity's definition, the FK definitions for a target entity in the source entity's definition contain:
1. One entry for relationships defined between source to target in the database (via FK constraint) in the right order of referencing/referenced entity.
2. Two entries for relationships defined in the config between source to target (which don't exist in the database).

The above is achieved via a call to a newly added method `SqlMetadataProvider.GetValidatedFKs()` which  loops over all the foreign key definitions defined for the target entity in the source entity's definition and adds to the set of validated FK definitions:
1. All the FK definitions which actually map to a foreign key constraint defined in the database. In such a case, if the source/target fields are also provided in the config, they are given precedence over the FK constraint.
2. FK definitions for custom relationships defined by the user in the configuration file where no FK constrain t exists between the pair of (source, target) entities.

## How was this tested?
- Added tests to `SqlMetadataProviderUnitTests` for Pg/Ms/MySql for the cases when an FK constraint exists between the source/target entities.

## Manual Testing
1. Did manual testing to ensure the redundant join condition is not added in the query.
For a nested query like:

image

where the configuration for Book entity looks like:

image

the query generated before the fix (in main):

```tsql
SELECT TOP 100 
[table0].[title] AS [title], 
JSON_QUERY ([table1_subq].[data]) AS [publishers] 
FROM [dbo].[books] AS [table0] 
OUTER APPLY (
    SELECT TOP 1
    [table1].[name] AS [name] 
    FROM [dbo].[publishers] AS [table1] 
    WHERE [table0].[publisher_id] = [table1].[id] 
    AND [table1].[id] = [table0].[publisher_id] 
    ORDER BY [table1].[id] ASC 
    FOR JSON PATH, INCLUDE_NULL_VALUES,WITHOUT_ARRAY_WRAPPER
) AS [table1_subq]([data]) 
WHERE 1 = 1 
ORDER BY [table0].[id] ASC 
FOR JSON PATH, INCLUDE_NULL_VALUES
```

after the fix:

```tsql
SELECT TOP 100 
[table0].[title] AS [title],
JSON_QUERY ([table1_subq].[data]) AS [publishers] 
FROM [dbo].[books] AS [table0] 
OUTER APPLY (
    SELECT TOP 1 
    [table1].[name] AS [name] 
    FROM [dbo].[publishers] AS [table1] 
    WHERE [table0].[publisher_id] = [table1].[id] 
    ORDER BY [table1].[id] ASC 
    FOR JSON PATH, INCLUDE_NULL_VALUES,WITHOUT_ARRAY_WRAPPER
) AS [table1_subq]([data]) 
WHERE 1 = 1 
ORDER BY [table0].[id] ASC FOR JSON PATH, INCLUDE_NULL_VALUES
```
<Observe the removal of redundant join condition in the 4th line in the second query>

Similar observations for MySql:
```mysql
SELECT COALESCE(JSON_ARRAYAGG(JSON_OBJECT(@param3, `subq6`.`title`, @param4, `subq6`.`publishers`)), JSON_ARRAY()) AS `data`
FROM ( SELECT `table0`.`title` AS `title`, `table1_subq`.`data` AS `publishers` FROM `books` AS `table0` LEFT OUTER JOIN LATERAL (SELECT JSON_OBJECT(@param2, `subq5`.`name`) AS `data` 
FROM ( SELECT `table1`.`name` AS `name` FROM `publishers` AS `table1` 
WHERE `table0`.`publisher_id` = `table1`.`id` 
ORDER BY `table1`.`id` ASC LIMIT 1 ) AS `subq5`) AS `table1_subq` ON TRUE WHERE 1 = 1 ORDER BY `table0`.`id` ASC LIMIT 100 ) AS `subq6`
```

For pgSql:
```psql
SELECT COALESCE(jsonb_agg(to_jsonb("subq6")), '[]') AS "data" FROM ( SELECT "table0"."title" AS "title", "table1_subq"."data" AS "publishers" FROM "public"."books" AS "table0" 
LEFT OUTER JOIN LATERAL (SELECT to_jsonb("subq5") AS "data" FROM ( SELECT "table1"."name" AS "name" 
FROM "public"."publishers" AS "table1" 
WHERE "table0"."publisher_id" = "table1"."id" 
ORDER BY "table1"."id" ASC LIMIT 1 ) AS "subq5") AS "table1_subq" 
ON TRUE 
WHERE 1 = 1 ORDER BY "table0"."id" ASC LIMIT 100 ) AS "subq6"
```

2. Did manual testing to confirm that the relationship fields provided by user take precedence:
(Creating issue here to add a test later:https://github.com/Azure/data-api-builder/issues/2104)
Config:
image

GraphQL query:
image

Generated SQL query:
image

3. Manually tested that for relationships which are not backed by an FK constraint, we infer both the source/target entities as the referencing entity. Test to be added post the merge of : https://github.com/Azure/data-api-builder/pull/2056


src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,98 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                {
+                    // Being here indicates that we inferred an FK constraint for the current FK.
+                    // If the referencing and referenced columns count > 0,
+                    // we have already gathered this information from the runtime config.
+                    if (fKDefinitionToTarget.ReferencingColumns.Count > 0 && fKDefinitionToTarget.ReferencedColumns.Count > 0)
+                    {
+                        validatedFKDefinitionsToTarget.Add(fKDefinitionToTarget);
+                    }
+                    // Only add the referencing/referenced columns if they have not been
+                    // specified in the configuration file.
```
severussundar, line 1718:
```suggestion
                    // The count of referenced and referencing columns being = 0 indicates that source.fields and target.fields 
                    // are not configured through the config file. In this case, the FK fields inferred from the database is populated.
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,98 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
-                            // if the referencing and referenced columns count > 0,
-                            // we have already gathered this information from the runtime config.
-                            if (fk.ReferencingColumns.Count > 0 && fk.ReferencedColumns.Count > 0)
-                            {
-                                continue;
-                            }
+                        // List of validated FK definitions from source to target which contains:
+                        // 1. One entry for relationships defined between source to target in the database  (via FK constraint)
+                        // in the right order of referencing/referenced entity.
+                        // 2. Two entries for relationships defined in the config between source to target (which don't exist in the database).
```
severussundar, line 1679:
```suggestion
                        // 
                        // Scenario 1: When a FK constraint is defined between source and target entities
                        // In this case, there will be exactly one ForeignKeyDefinition with the right pair of Referencing and Referenced tables. 
                        // Scenario 2: When no FK constraint is defined between source and target entities, but the relationship fields are configured through config file
                        // In this case, two entries will be created. 
                        // First entry: Referencing table: Source entity, Referenced table: Target entity
                        // Second entry: Referencing table: Target entity, Referenced table: Source entity 
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,98 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                    // we have already gathered this information from the runtime config.
+                    if (fKDefinitionToTarget.ReferencingColumns.Count > 0 && fKDefinitionToTarget.ReferencedColumns.Count > 0)
+                    {
+                        validatedFKDefinitionsToTarget.Add(fKDefinitionToTarget);
+                    }
+                    // Only add the referencing/referenced columns if they have not been
+                    // specified in the configuration file.
+                    else
+                    {
+                        validatedFKDefinitionsToTarget.Add(inferredFKDefinition);
```
severussundar, line 1721: The current logic checks individually, for `ReferencingColumns.Count == 0` before adding `ReferencingColumn` information. Likewise, for `ReferencingColumns.Count`. 

AFAICT, this is possible only `source.fields` is empty when `target.fields` is non-empty (or vice-versa). Such a setup is invalid. It looks like it should be okay to populate the inferred details from the database without those checks. 
Do you see any side-effects of populating without those checks? @Aniruddh25

severussundar (commented):
1. Is there opportunity to have unit tests that validate the count of ForeignKeyDefinitions for various scenarios?

There are few unit tests in `SqlMetadaProviderUnitTests` that execute against `InitializeAsync()`. Ex: `CheckNoExceptionForNoForeignKey()`

2. At the moment, this PR is target towards `main`. But this needs to be ported to `dev/NestedMutations` as well - just calling it out.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Yeah I and Ani had a thorough discussion around this and we came to an agreement that this change makes logical sense and should be fine.

severussundar:
In [BaseSqlQueryStructure.AddJoinPredicatesForRelatedEntity()](https://github.com/Azure/data-api-builder/blob/main/src/Core/Resolvers/Sql%20Query%20Structures/BaseSqlQueryStructure.cs#L181), there is conditional logic to determine whether the foreign key definition is valid or not. 

With this change, can you please evaluate whether that can be removed?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
severussundar: Please add details around the same as comments or in your PR description

ayush3797: - > All the FK definitions which actually map to a foreign key constraint defined in the database. In such a case, the source/target fields are also provided in the config, they are given precedence over the FK constraint.

I had added this detail initially. Works?

ayush3797:
> 2. At the moment, this PR is target towards `main`. But this needs to be ported to `dev/NestedMutations` as well - just calling it out.

Yes, correct. This needs to be ported and we need to build on top of it to handle other invalid cases with nested insertion. I had a similar PR for that targeted to dev/NesredMutations but I have closed it temporarily and will reopen it once this PR is merged.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,101 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
-                        foreach (ForeignKeyDefinition fk in foreignKeysForTarget)
-                        {
-                            // if the referencing and referenced columns count > 0,
-                            // we have already gathered this information from the runtime config.
-                            if (fk.ReferencingColumns.Count > 0 && fk.ReferencedColumns.Count > 0)
-                            {
-                                continue;
-                            }
+                        // 
+                        // Scenario 1: When a FK constraint is defined between source and target entities
```
Aniruddh25, line 1677:
```suggestion
                        // Scenario 1: When a FK constraint is defined between source and target entities in the database
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,101 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                // Add the referencing and referenced columns for this foreign key definition for the target.
+                if (PairToFkDefinition is not null &&
+                    PairToFkDefinition.TryGetValue(fKDefinitionToTarget.Pair, out ForeignKeyDefinition? inferredFKDefinition))
+                {
+                    // Being here indicates that we inferred an FK constraint for the current FK.
+                    // The count of referencing and referenced columns being > 0 indicates that source.fields and target.fields 
+                    // have been specified in the config file. In this scenario, higher precedence is given to the fields configured through the config file. So, the existing FK definition is retained as is.   
+                    if (fKDefinitionToTarget.ReferencingColumns.Count > 0 && fKDefinitionToTarget.ReferencedColumns.Count > 0)
+                    {
+                        validatedFKDefinitionsToTarget.Add(fKDefinitionToTarget);
```
Aniruddh25, line 1718: if precedence is always going to be given to the config file fks, why do we need to do a lookup of inferredFKdefinition on line 1711. Why cant we directly add this fkDefinitionToTarget and skip the loop. 
Only if we dont find anything in the config is the need to lookup the inferred FK defintion or its inverse.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,101 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                    // while the 'publishers' is the referenced table.
+                    // (The definition for that constraint would be taken care of while adding database FKs above.)
+
+                    // So, before concluding that there is no FK constraint between the source, target entities, we need
+                    // to confirm absence of FK constraint from source->target and target->source tables.
+                    RelationShipPair inverseFKPair = new(fKDefinitionToTarget.Pair.ReferencedDbTable, fKDefinitionToTarget.Pair.ReferencingDbTable);
+
+                    // Add FK definition to the set of validated FKs only if no FK constraint is defined for the source and target entities
+                    // in the database, either from source -> target or target -> source.
+                    if (PairToFkDefinition is not null && !PairToFkDefinition.ContainsKey(inverseFKPair))
```
Aniruddh25, line 1756: If the inversePair does exist but hasnt been added yet to the validatedFKDefinitions, will it get added in next iteration of this loop?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,101 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                    {
+                        validatedFKDefinitionsToTarget.Add(fKDefinitionToTarget);
```
Aniruddh25, line 1758: If there is no FK definition either from source-> target or target->source, adding the fkDefinition here will mean its coming from config. But if it doesn't contain any referencing/referenced columns, adding it here is of no use. instead that should be a validation error.

Aniruddh25 (changes requested):
Can do some optimizations.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Like we discussed, this should be caught during config validation. Until we have that validation checked in (issue: https://github.com/Azure/data-api-builder/issues/1935), we can decide to have a validation in place here. What do you suggest? I'll ping Sean to see if this can be assigned to anyone for 0.12 in the meantime.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Consider declaring a relationship in the config where entity: `Book`, relationship: `publishers` and source/target fields are provided by the user:

image,

For such a relationship where right cardinality is 1, we add an FK definition from both direction. In one of them we keep `Book` as the `referencing` table, and `Publisher` in the other.
At the end of the `FillInferredFKInfo()` however, we only want validated FK definitions. And this would be the FK definition where Book is the referencing entity.
If we do what you suggest, we would end up adding the invalid FK definition as well.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Yes, if the inverse pair exists, it will eventually get added to the `validatedFKDefinitions`.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: > if precedence is always going to be given to the config file fks
And this behavior will change with support for `multiple-create`. I'll have to build on top of this PR.

src/Service.Tests/dab-config.PostgreSql.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json",
   "data-source": {
     "database-type": "postgresql",
-    "connection-string": "Host=localhost;Database=datagatewaytest;username=REPLACEME;password=REPLACEME",
+    "connection-string": "Host=localhost;Database=postgres;username=postgres;password=Srsral12345;Port=5433;",
```
severussundar, line 5: nit: please revert

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
severussundar: The check for count of referencing/referenced columns being 0 is removed before adding `ReferencingColumn`/`ReferencedColumn` information. Adding details about why it is okay to remove this check will be helpful in the future if we were to look at this logic in the future.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -339,5 +340,135 @@ public void ValidateGraphQLReservedNaming_DatabaseColumns(string dbColumnName, s
...
+            // Validate that when no custom source.fields/target.fields are defined in the config for a relationship of cardinality 1:1
+            // between Book - BookWebsitePlacement and an FK constraint exists from BookWebsitePlacement->Book,
+            // we successfully determine at the startup, that BookWebsitePlacement is the referencing entity.
+            ValidateReferencingEntitiesForRelationship("Book", "BookWebsitePlacement", new List<string>() { "BookWebsitePlacement" });
+        }
+
+        private static void ValidateReferencingEntitiesForRelationship(
+            string sourceEntityName,
+            string targetEntityName,
+            List<string> referencingEntityNames)
```
severussundar, line 434: nit: Becomes apparent when reading the test logic, but re-naming would make the logic easier to understand
```suggestion
            List<string> expectedReferencingEntityNames)
```

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -339,5 +340,135 @@ public void ValidateGraphQLReservedNaming_DatabaseColumns(string dbColumnName, s
...
+                }
+
+                DatabaseTable actualReferencingTable = foreignKey.Pair.ReferencingDbTable;
+                actualReferencingTables.Add(actualReferencingTable);
+            }
+
+            Assert.IsTrue(actualReferencingTables.SetEquals(expectedReferencingTables));
+        }
+
+        private static async Task InferMetadata()
```
severussundar, line 463:
```suggestion
        private static async Task InitializeTestSetupAndInferMetadata()
```

ayush3797:
In process of adding/analysing tests. Will mark the PR ready for review again when done. Can review non-testing changes meanwhile.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,101 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
-                                }
-                            }
-                        }
+                // Add the referencing and referenced columns for this foreign key definition for the target.
+                if (PairToFkDefinition is not null &&
+                    PairToFkDefinition.TryGetValue(fKDefinitionToTarget.Pair, out ForeignKeyDefinition? inferredFKDefinition))
+                {
+                    // Being here indicates that we inferred an FK constraint for the current FK.
+                    // The count of referencing and referenced columns being > 0 indicates that source.fields and target.fields 
+                    // have been specified in the config file. In this scenario, higher precedence is given to the fields configured through the config file. So, the existing FK definition is retained as is.
```
aaronburtle, line 1715: nit: line break for easier reading this comment

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1666,52 +1666,101 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                    // This code block adds FK definitions between source and target entities when there is no FK constraint defined
+                    // in the database, either from source->target or target->source entities.
+
+                    // Being here indicates that we did not find an FK constraint in the database for the current FK definition.
+                    // But this does not indicate absence of an FK constraint between the source, target entities yet.
+                    // This may happen when an FK constraint exists between two tables, but in an order opposite to the order
+                    // of referencing and referenced tables present in the current FK definition. This happens because for a relationship
+                    // with right cardinality as 1, we add FK definitons from both source->target and target->source to the source entity's definition.
+                    // because at that point we don't know if the relationship is an N:1 relationship or a 1:1 relationship.
+                    // So here, we need to remove the wrong FK definition for:
```
seantleonard, line 1743: One thing that tripped us up during review was the usage of 'wrong'. Is it accurate to say that 'wrong' can be replaced with either 'redundant' or 'duplicate' ? Because it looks like, without this check, the where clause had duplicate predicates, but wasn't exactly causing wrong results/crashing.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -339,5 +340,121 @@ public void ValidateGraphQLReservedNaming_DatabaseColumns(string dbColumnName, s
...
+            ValidateInferredRelationshipInfoForTables();
+        }
+
+        /// <summary>
+        /// Helper method for tests which validate that the relationship data is correctly inferred based on the info provided
+        /// in the config and the metadata collected from the database. It runs the test against various test cases verifying that
+        /// when a relationship is defined in the config between source and target entity and an FK constraint exists in the database between the two entities: We successfully determine which is the referencing
+        /// entity based on the FK constraint.
+        /// </summary>
+        private static void ValidateInferredRelationshipInfoForTables()
```
aaronburtle, line 386: do we have test coverage to handle this scenario?

// Scenario 2: When no FK constraint is defined between source and target entities, but the relationship fields are configured through config file
                        // In this case, two entries will be created. 
                        // First entry: Referencing table: Source entity, Referenced table: Target entity
                        // Second entry: Referencing table: Target entity, Referenced table: Source entity

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -339,5 +340,121 @@ public void ValidateGraphQLReservedNaming_DatabaseColumns(string dbColumnName, s
...
+        /// Test to validate successful inference of relationship data based on data provided in the config and the metadata
+        /// collected from the MsSql database.
+        /// </summary>
+        [TestMethod, TestCategory(TestCategory.MSSQL)]
+        public async Task ValidateInferredRelationshipInfoForMsSql()
+        {
+            DatabaseEngine = TestCategory.MSSQL;
+            await SetupTestFixtureAndInferMetadata();
+            ValidateInferredRelationshipInfoForTables();
+        }
```
seantleonard, line 369: This test has an issue with artifact config files. It always runs successfully the first time the test is ran or debugged. Any subsequent runs result in a "null reference exception" , and the console shows loading two different config files. Please check and make sure this test runs with a clean filesystem each time.

seantleonard: For some reason the first run always uses the file from the test project. then uses the config from the "service" project for subsequent runs (perhaps it is left as an artifact). -You can repro if from the service project config, you remove the "stock" entity.
-while the repo doesn't have "stock" entity removed, this is still a bug where the wrong config is retreived for subsequent test runs.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: So pending #1935 , I understand that certain invalid state should be caught during validation such that it isn't needed to be dealt with here:
1. as Ani mentioned, when no db fk exists, but relationship is defined in config.. and that relationship doesn't define source or target fields. OR source.fields.count != target.fields.count as mentioned by the linked validation issue.

Ayush, this is also a condition that is not hit in the integration tests. Can you add a test to exercise this condition? e.g. no db defined fk for N:1 relationship

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1671,52 +1671,101 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
-                                // specified in the configuration file.
-                                if (fk.ReferencedColumns.Count == 0)
-                                {
-                                    fk.ReferencedColumns.AddRange(inferredDefinition.ReferencedColumns);
-                                }
-                            }
-                        }
+                // Add the referencing and referenced columns for this foreign key definition for the target.
+                if (PairToFkDefinition is not null &&
+                    PairToFkDefinition.TryGetValue(fKDefinitionToTarget.Pair, out ForeignKeyDefinition? inferredFKDefinition))
```
seantleonard, line 1716: is this condition entered in any tests?

seantleonard (commented):
Looks good so far, but the "Scenario 2" case mentioned in code comments has no tests exercising those paths. that would help understand those code segments so we can step through debugger in the code.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: If you talk about Foreign key, they are from referencing table to referencing table. So when you add a foreign key from publisher as referencing table to book as referenced table, this is a wrong foreign key and not redundant.

I get why you have the idea of redundant - because for joins it does not matter which is the referencing entity. But going forward for mulitple-mutation, this matters.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Going inside the if block guarded by lines 1715-1716? Yes, in all the cases where relationship is backed by an Fk, the code hits this block.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: As I mentioned in the PR, I plan to add tests for custom relationships once this PR: https://github.com/Azure/data-api-builder/pull/2056 is in. Currently, there are no set entities present in the config which allow me to add meaningful custom relationships. Hence, I have manually tested the scenarios. 

Also for custom relationships, 1:1 or N:1 does not really matter as I explained during the session. Let me know if you want to get on a call and run through some manual tests which I did.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
ayush3797: Please refer: https://github.com/Azure/data-api-builder/pull/2051#discussion_r1533581943

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1671,52 +1671,102 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                else
+                {
+                    // This code block adds FK definitions between source and target entities when there is no FK constraint defined
+                    // in the database, either from source->target or target->source entities.
+
+                    // Being here indicates that we did not find an FK constraint in the database for the current FK definition.
+                    // But this does not indicate absence of an FK constraint between the source, target entities yet.
+                    // This may happen when an FK constraint exists between two tables, but in an order opposite to the order
+                    // of referencing and referenced tables present in the current FK definition. This happens because for a relationship
+                    // with right cardinality as 1, we add FK definitons from both source->target and target->source to the source entity's definition.
```
aaronburtle, line 1742:
```suggestion
                    // with right cardinality as 1, we add FK definitions from both source->target and target->source to the source entity's definition.
```

aaronburtle (approved):
This looks good and very thorough, please add an issue to track the addition of integration tests, since while everything looks good, these kinds of cases can be tricky to understand and no doubt in the future having good integration testing will be valuable for preventing regressions.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -354,5 +355,121 @@ public void ValidateGraphQLReservedNaming_DatabaseColumns(string dbColumnName, s
...
+
+        /// <summary>
+        /// Test to validate successful inference of relationship data based on data provided in the config and the metadata
+        /// collected from the MySql database.
+        /// </summary>
+        [TestMethod, TestCategory(TestCategory.MYSQL)]
```
severussundar, line 375: Optional: `FillInferredFkInfo()` and `GetValidatedFKs()` get executed inrrespective of the SQL database type. Having a test with a single SQL database type looks sufficient to exercise those code paths. Please let me know if I've missed something.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
```diff
@@ -354,5 +355,121 @@ public void ValidateGraphQLReservedNaming_DatabaseColumns(string dbColumnName, s
...
+        /// <summary>
+        /// Resets the database state and infers metadata for all the entities exposed in the config.
+        /// </summary>
+        private static async Task SetupTestFixtureAndInferMetadata()
+        {
+            TestHelper.SetupDatabaseEnvironment(DatabaseEngine);
+            RuntimeConfig runtimeConfig = SqlTestHelper.SetupRuntimeConfig();
+            RuntimeConfigProvider runtimeConfigProvider = TestHelper.GenerateInMemoryRuntimeConfigProvider(runtimeConfig);
+            SetUpSQLMetadataProvider(runtimeConfigProvider);
+            await ResetDbStateAsync();
```
severussundar, line 471: nit: suggestion to add a short comment about what `ResetDbStateAsync()` performs to avoid ambiguity on whether to invoke it before or after `_sqlMetadaProvider.InitializeAsync()` which is the method under test. 
```suggestion
           // The `ResetDbStateAsync()` method executes the .sql script of the respective database type. 
           // This serves as a setup phase for this test.   
            await ResetDbStateAsync();
```

severussundar (approved):
Thanks for making these changes :)

Few straight-forward suggestions, apart from that looks good.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1671,52 +1671,102 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
-                                {
-                                    fk.ReferencedColumns.AddRange(inferredDefinition.ReferencedColumns);
-                                }
-                            }
-                        }
+                // Add the referencing and referenced columns for this foreign key definition for the target.
+                if (PairToFkDefinition is not null &&
+                    PairToFkDefinition.TryGetValue(fKDefinitionToTarget.Pair, out ForeignKeyDefinition? inferredFKDefinition))
+                {
+                    // Being here indicates that we inferred an FK constraint for the current FK.
```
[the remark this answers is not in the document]
seantleonard: adding

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1671,52 +1671,102 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                    // are not configured through the config file. In this case, the FK fields inferred from the database are populated.
+                    else
+                    {
+                        validatedFKDefinitionsToTarget.Add(inferredFKDefinition);
+                    }
+                }
+                else
+                {
+                    // This code block adds FK definitions between source and target entities when there is no FK constraint defined
+                    // in the database, either from source->target or target->source entities.
```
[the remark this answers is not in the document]
seantleonard: modified.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
seantleonard: Added to method summary.

src/Service.Tests/Unittests/SqlMetadataProviderUnitTests.cs
seantleonard: fixed by Ayush in earlier iteration

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: the concern raised in this thread should be resolved by #2139 to catch mismatch in number of config defined referencing/referenced columns (target/source fields).

PR #2052, opened 2024-02-20
Fix Performance Issue for filtering on varchar columns in MS_SQL database

### LINKED ISSUE #1973: MSSQL API slow with large tables due to incorrect varchar/nvarchar in sp_executesql
### Discussed in https://github.com/Azure/data-api-builder/discussions/1964

I converted the discussion to an issue since I've been able to identify the problem. Please see my comment here: https://github.com/Azure/data-api-builder/discussions/1964#discussioncomment-8121313

<div type='discussions-op-text'>

<sup>Originally posted by **tdutch1** January 11, 2024</sup>
I'm working on a POC using DAB in a Static Web App. My data source is an Azure SQL table. It's taking anywhere from three to ten seconds for the DAB API to return a response from database queries. When capturing and running the same SQL in SSMS, the queries return results in less than one second.

Is it generally expected that DAB APIs, when running in a SWA, will take at least three seconds to return results? Also, what are some factors that would cause the range to go from three seconds to ten? The Azure SQL database is not heavily used so that does not appear to be the issue.</div>

abhishekkumams:
## Why make this change?

- Closes #1973 
  - Slow response of DAB when filtering on varchar columns in MS_SQL compared to other DBs. 
  - Same query in mssql taking around ~1sec, compared to <50ms in other DBs for large number of rows, eg. 20 million
  - This was happening because the query generated by DAB when runs on Sql Server for varchar columns, it automatically assumes the filter value of type nvarchar, due to which it does implicit cast leading to slowness.
  - To fix this we needed to identify if the filter is for varchar column, we explitily specify it to be varchar for faster response.

## What is this change?

- We are modifying the DbCommand and adding extra information that is sent to SqlClient.
- We introduced a new property called `SqlDbType`, By Default for String DbType, SqlDbType was considered as `nvarchar`. now we are updating this property to the correct sql type.
- This allows db to directly execute the command without type casting due to type mismatch.
- Carved out the code which creates DbCommand in `QueryExecuter`, and added an override method in `MsSqlQueryExecuter`. here we set the property SqlDbType for the Parameters.

  - Link to Discussion which talks about this highlighting the issue : https://github.com/Azure/data-api-builder/discussions/1964#discussioncomment-8121313
  - Added a new property SqlDbType in Column class, currently we used to have only DotNet DbType which doesn't differenciate between varchar and Nvarchar. So, adding this SqlDbType gave us the freedom to modify query based on the sql db type.

## How was this tested?

- Since, this is a performance issue and only a new column got introduced. The main aim here was to make sure excising functionality doesn't break. 
- Added few more extra tests.

The performance difference is more visible when there are large number of rows. So, I added 20 million records and ran perf test.
Below are the performance results with avg response time:

|  | varchar(before) | nvarchar(before |
|---|---|---|
| GraphQL | ~734ms | ~23ms |
| REST | ~655ms | ~24ms |

|  | varchar(after) | nvarchar(after) |
|---|---|---|
| GraphQL | ~18ms | ~23ms |
| REST | ~18ms | ~25ms |

#### VARCHAR
image
image

#### NVARCHAR
image
image

#### VARCHAR
image

#### NVARCHAR
image

#### VARCHAR
image
image

#### NVARCHAR
image
image

#### VARCHAR
image

#### NVARCHAR
image


seantleonard:
Based on your screenshots, it seems like "varchar" variant had the perf boost:

|  | varchar(before) | nvarchar(before |
|---|---|---|
| GraphQL | 734ms | 19ms |
| REST | 655ms | 24ms |

|  | varchar(after) | nvarchar(after) |
|---|---|---|
| GraphQL | 18ms | 23ms |
| REST | 18ms | 25ms |

Question about:
> CAST(@param1 AS VARCHAR(MAX))

why cast instead of setting the sqldbtype on the dbconnectionparameter that is created?

seantleonard:
> I have tried setting sqldbtype in DbConnectionparam but it dodn't work.

Please provide more context on what you tried and what exactly didn't work. Are you saying when we create parameters and send them to SQL, SQL isn't honoring the type specified? Are you suggesting that there is a bug with SqlClient?

src/Service.Tests/SqlTests/GraphQLFilterTests/GraphQLFilterTestBase.cs
seantleonard, line 1: Can you clarify how these tests are validating your changes? per pr description, the performance behavior was surfacing as long execution time and not "invalid results returned." how are you testing that perf is better here or that the appropriate CAST is occurring?

src/Service.Tests/DatabaseSchema-MySql.sql
```diff
@@ -152,7 +152,8 @@ CREATE TABLE trees (
 
 CREATE TABLE fungi (
     speciesid int PRIMARY KEY,
-    region text
+    region text,
+    habitat varchar(6)
```
seantleonard, line 156: are we encountering this perf issue on pg/mysql?

seantleonard (changes requested):
Need to elaborate "why" for the following two points in pr description 
> Cast the value to varchar(max) work on columns of type varchar(x). x is a finite number. 
> filtering on columns with varchar(x), when the value is >x. doesn't get truncated to match values. example: column of type varchar(4) has a value "sand". if i filter on this column with "sandman". It should not truncate it to "sand" and check. Here, the result will be empty.

It seems the root cause is how DAB uses the request provided metadata in SqlQueryStructure to create the SQL query sent to the database. All request provided values are put into parameters and those parameters need to have the correct SqlDbType. It would help to elaborate why evaluating the correct SqlDbType during parameter creation is not sufficient as opposed to having "resolveSqlDbTypeFromColumn" calls across the code base.

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -125,6 +125,7 @@ public class ParameterDefinition
 {
     public Type SystemType { get; set; } = null!;
     public DbType? DbType { get; set; }
+    public SqlDbType? SqlDbType { get; set; }
```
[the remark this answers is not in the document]
Aniruddh25: Also why is it nullable?

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -125,6 +125,7 @@ public class ParameterDefinition
 {
     public Type SystemType { get; set; } = null!;
     public DbType? DbType { get; set; }
```
Aniruddh25, line 127: What purpose does DbType have now as opposed to SqlDbType?

Do we need a MsSqlspecific derived Parameter Definition?

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -62,6 +69,51 @@ public string Build(SqlQueryStructure structure)
...
+                string rightOperand = ResolveOperand(predicate.Right);
+
+                // This is done to make sure there are performance issues due to implicit conversion of data types from nvarchar to varchar.
+                if (predicate.Left.AsColumn() is not null && predicate.Left.AsColumn()!.ColumnSqlDbType == SqlDbType.VarChar && !rightOperand.Equals("NULL", StringComparison.OrdinalIgnoreCase))
+                {
+                    // This operation is valid even when the column being filtered is not VARCHAR(MAX), but a smaller VARCHAR type such as VARCHAR(20).
+                    rightOperand = $"CAST({rightOperand} AS VARCHAR(MAX))";
+                }
+
+                // For Binary predicates:
```
Aniruddh25, line 98: Why cant the right side be a Column? What if both sides are columns i.e. its a Join Predicate?

src/Service.Tests/SqlTests/GraphQLFilterTests/GraphQLFilterTestBase.cs
```diff
@@ -71,6 +71,100 @@ public async Task TestStringFiltersEqWithMappings(string dbQuery)
...
+                        id
+                        journalname
+                        color
+                        ownername
+                    }
+                }
+            }";
+
+            string dbQuery = MakeQueryOn(
+                "journals",
```
Aniruddh25, line 94: How are we ensuring the perf issue identified doesnt regress again?

Aniruddh25 (changes requested):
Need better reasoning between DbType v.s SqlDbType..

src/Core/Resolvers/Sql Query Structures/SqlUpdateQueryStructure.cs
```diff
@@ -159,7 +160,15 @@ private Predicate CreatePredicateForParam(KeyValuePair<string, object?> param)
             Predicate predicate;
             // since we have already validated param we know backing column exists
             MetadataProvider.TryGetBackingColumn(EntityName, param.Key, out string? backingColumn);
-            if (param.Value is null && !sourceDefinition.Columns[backingColumn!].IsNullable)
+            if (backingColumn is null)
+            {
+                // If param.Key was not present in the ExposedToBackingColumnMap then provided param.Key is already the backing column
+                backingColumn = param.Key;
+            }
```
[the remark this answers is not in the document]
abhishekkumams: yep, while going through the code, i noticed we are already passing the function with backing column at several places. So, when we get the backing column already, and it calls a method to get the backing column from backing column, it returns null, which is why i added this condition.

src/Config/DatabasePrimitives/DatabaseObject.cs
abhishekkumams: DbType here is the .Net DbType, while SqlDbType is SqlDb specific dbType. For example: .Net DbType doesn't differentiate between varchar and nvarchar, it uses just String type for both. So, here's how SqlDbType is helpful.

abhishekkumams:
> > I have tried setting sqldbtype in DbConnectionparam but it dodn't work.
> 
> Please provide more context on what you tried and what exactly didn't work. Are you saying when we create parameters and send them to SQL, SQL isn't honoring the type specified? Are you suggesting that there is a bug with SqlClient?

So, what i did was, I added a new property called `SqlDbType` in DbConnectionParam, and then i was setting its value before execution. It didn't bring any perf improvement. Also, I checked the Class DbParameters.
image

which is required as part of command that we send to SqlClient, doesn't have SqlDbType property or any other relevant property that could be used to specify the underlying Sql DbType for the parameter.

src/Config/DatabasePrimitives/DatabaseObject.cs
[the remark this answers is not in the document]
abhishekkumams: for other Db's it will be null.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -305,12 +305,51 @@ public override string GetSessionParamsQuery(HttpContext? httpContext, IDictiona
...
+            cmd.CommandType = CommandType.Text;
+
+            // Add query to send user data from DAB to the underlying database to enable additional security the user might have configured
+            // at the database level.
+            string sessionParamsQuery = GetSessionParamsQuery(httpContext, parameters, dataSourceName);
+
+            cmd.CommandText = sessionParamsQuery + sqltext;
+            if (parameters is not null)
+            {
+                foreach (KeyValuePair<string, DbConnectionParam> parameterEntry in parameters)
```
seantleonard, line 326: Odd behavior seen in debugger: when you submit a GraphQL request like the following:
```graphql
query GetBooksPaged {
  books(first: 5) {
    items {
      id
      title
      publisher_id
    }
    endCursor
    hasNextPage
  }
}
```

the generated sql query looks like:
```tsql
SELECT TOP 6 [table0].[id] AS [id], [table0].[title] AS [title], [table0].[publisher_id] AS [publisher_id] FROM [dbo].[books] AS [table0] WHERE 1 = 1 ORDER BY [table0].[id] ASC FOR JSON PATH, INCLUDE_NULL_VALUES
```

yet the variable `parameters` has three entries. So there is work being done here for parameters that aren't even in the generated sql query. 

Only reason i saw this is because when creating `SqlParameter parameter = cmd.CreateParameter();` you can use the `SqlParameter` constructor which takes in a `SqlDbType`. If you provide that , then `SqlParameter` automatically resolves the `DbType` rendering the helper function `PopulateDbTypeForParameter(..)` as no longer needed.

seantleonard: While the issue i mentioned may not necessarily need to be addressed in this pr, please file an issue for it so we can track. 

Here are the three params that are present:
image

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -62,6 +69,51 @@ public string Build(SqlQueryStructure structure)
...
+                string leftOperand = ResolveOperand(predicate.Left);
+                string operation = Build(predicate.Op);
+                string rightOperand = ResolveOperand(predicate.Right);
+
+                // This is done to make sure there are performance issues due to implicit conversion of data types from nvarchar to varchar.
+                if (predicate.Left.AsColumn() is not null && predicate.Left.AsColumn()!.ColumnSqlDbType == SqlDbType.VarChar && !rightOperand.Equals("NULL", StringComparison.OrdinalIgnoreCase))
+                {
+                    // This operation is valid even when the column being filtered is not VARCHAR(MAX), but a smaller VARCHAR type such as VARCHAR(20).
+                    rightOperand = $"CAST({rightOperand} AS VARCHAR(MAX))";
+                }
```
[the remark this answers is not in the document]
abhishekkumams: not using casting anymore

src/Service.Tests/DatabaseSchema-MySql.sql
abhishekkumams: nope. this is specific to MsSql

src/Core/Resolvers/MsSqlQueryBuilder.cs
abhishekkumams: removed casting approach, the new approach works for both sides

src/Service.Tests/SqlTests/GraphQLFilterTests/GraphQLFilterTestBase.cs
abhishekkumams: since this perf issue was happening only for very large set of records ~20 million. we can't add a test for this atm.

src/Core/Resolvers/MsSqlQueryExecutor.cs
abhishekkumams: created an issue #2125 to track the same. Thanks for pointing out.

src/Service.Tests/SqlTests/GraphQLFilterTests/GraphQLFilterTestBase.cs
abhishekkumams: these tests are not perf tests. I added them because they're some extra cases for which i didn't find in the current test suite and it also validates that the changes are not breaking anything.

src/Config/DatabasePrimitives/DatabaseObject.cs
ayush3797: So, if we have SqlDbType available for an MsSql parameter, do we need the DbType as well?

ayush3797: Bloating a ParamDefinition/ColumnDefinition with fields not relevant for every database seems not to be the best approach here.

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -201,6 +202,16 @@ public bool IsAnyColumnNullable(List<string> columnsToCheck)
 
         return null;
     }
+
+    public virtual SqlDbType? GetSqlDbTypeForParam(string paramName)
```
ayush3797, line 206: Add method summary

src/Core/Models/DbConnectionParam.cs
```diff
@@ -10,10 +10,11 @@ namespace Azure.DataApiBuilder.Core.Models;
 /// </summary>
 public class DbConnectionParam
```
ayush3797, line 11: Might not be in this PR but something to consider: Can we create a sub-classes of DbConnectionParam for MsSql Vs Other databases (like `MsSqlDbConnectionParam`) and create parameter objects for sub-classes? Seems like for MsSql, just SqlDbType should be sufficient. Please open up an issue if this makes sense.

src/Core/Resolvers/BaseQueryStructure.cs
```diff
@@ -107,13 +107,16 @@ public class BaseQueryStructure
...
-        /// <param name="paramName"> The name of the parameter - column name for table/views or parameter name for stored procedures.</param>
+        /// <param name="paramName"> The name of the parameter - backing column name for table/views or parameter name for stored procedures.</param>
         public virtual string MakeDbConnectionParam(object? value, string? paramName = null)
         {
             string encodedParamName = GetEncodedParamName(Counter.Next());
             if (!string.IsNullOrEmpty(paramName))
             {
-                Parameters.Add(encodedParamName, new(value, GetUnderlyingSourceDefinition().GetDbTypeForParam(paramName)));
+                Parameters.Add(encodedParamName,
+                    new(value,
```
ayush3797, line 117: Nit: add named parameters for quick reference.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -305,12 +305,51 @@ public override string GetSessionParamsQuery(HttpContext? httpContext, IDictiona
...
+                {
+                    SqlParameter parameter = cmd.CreateParameter();
+                    parameter.ParameterName = parameterEntry.Key;
+                    parameter.Value = parameterEntry.Value.Value ?? DBNull.Value;
+                    PopulateDbTypeForParameter(parameterEntry, parameter);
```
ayush3797, line 331: Most of this method seems to be a duplicate of the base class' method definition. Can't just providing an overridden implementation for `PopulateDbTypeForParameter()` method suffice?

src/Core/Models/DbConnectionParam.cs
abhishekkumams: #2127

src/Core/Resolvers/MsSqlQueryExecutor.cs
abhishekkumams: No, because we need `SqlCommand` and not `DbCommand` which is there in the base implementation.

src/Config/DatabasePrimitives/DatabaseObject.cs
abhishekkumams: we would have to check it's where it is being used. Could be taken as a seperate clean up PR.

src/Service.Tests/SqlTests/GraphQLFilterTests/GraphQLFilterTestBase.cs
seantleonard: can't we test that the parameters are being created with the exact/expected type and that we're not losing information by using `DbType`?

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -246,13 +242,12 @@ public GQLFilterParser(RuntimeConfigProvider runtimeConfigProvider, IMetadataPro
...
-                                tableName: sourceName,
-                                tableAlias: sourceAlias,
-                                fieldSqlDbType: columnSqlDbType,
-                                processLiterals: queryStructure.MakeDbConnectionParam)));
+                                backingColumnName,
+                                subfields,
+                                schemaName,
+                                sourceName,
+                                sourceAlias,
+                                queryStructure.MakeDbConnectionParam)));
```
seantleonard, line 250: nit: didn't need to remove named parameters.

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -481,25 +476,23 @@ public HttpContext GetHttpContextFromMiddlewareContext(IMiddlewareContext ctx)
...
     /// <param name="schemaName">The db schema name to which the table belongs</param>
     /// <param name="tableName">The name of the table underlying the *FilterInput being processed</param>
     /// <param name="tableAlias">The alias of the table underlying the *FilterInput being processed</param>
-    /// <param name="fieldSqlDbType">The sql DbType of the field</param>
     /// <param name="processLiterals">Parametrizes literals before they are written in string predicate operands</param>
     private static Predicate ParseScalarType(
         IMiddlewareContext ctx,
         IInputField argumentSchema,
-        string fieldName,
+        string name,
```
seantleonard, line 488: seems like param name `fieldName` was more self explanatory.

src/Core/Resolvers/MsSqlQueryExecutor.cs
seantleonard, line 331: I don't understand why you need `PopulateDbTypeForParameter(parameterEntry, parameter);`

`SqlParameter parameter = cmd.CreateParameter(sqlDbType: parameterEntry.Value.SqlDbType);` achieves the same thing because it auto-resolves the `DbType` .

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -224,6 +224,12 @@ Mock<MsSqlQueryExecutor> queryExecutor
                 = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object);
 
             queryExecutor.Setup(x => x.ConnectionStringBuilders).Returns(new Dictionary<string, DbConnectionStringBuilder>());
+            queryExecutor.Setup(x => x.PrepareDbCommand(
+                It.IsAny<SqlConnection>(),
+                It.IsAny<string>(),
+                It.IsAny<IDictionary<string, DbConnectionParam>>(),
+                It.IsAny<HttpContext>(),
+                It.IsAny<string>())).CallBase();
```
seantleonard, line 232: when you use `CallBase()` doesn't that not run the MsSqlQueryExecutor specific implementation?

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
seantleonard, line 1: is there really no way to test the changes in this PR? you have the `SqlQueryExecutorUnitTests` file.. perhaps a validation that the invocation is provided  a dictionary of DbConnectionParams, which when casted to SqlConnectionParam.. is populated with the expected `SqlDbType`s. Before this change, I believe SqlDbType was not provided to the database.

seantleonard (changes requested):
last question about tests.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
abhishekkumams: yes, it runs `MsSqlQueryExecutor` specific implementation

src/Core/Resolvers/MsSqlQueryExecutor.cs
abhishekkumams: This method has some extra checks as well which validates before assigning values. Also it was added in an older PR. Btw, we do have an issue tracking whether we need DbType, now that we are introducing SqlDbType. So this will be touched in that issue to understand better usage.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
abhishekkumams: added a test for the same.

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -412,7 +412,7 @@ private static void VerifyColumnDefinitionSerializationDeserialization(ColumnDef
         {
             // test number of properties/fields defined in Column Definition
             int fields = typeof(ColumnDefinition).GetFields(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance).Length;
-            Assert.AreEqual(fields, 7);
+            Assert.AreEqual(fields, 8);
```
seantleonard, line 415: magic number. not sure why this value increased from 7 to 8. I understand that this was existing test, but none of the comments surrounding this code explain the number.

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -422,7 +422,7 @@ private static void VerifyParameterDefinitionSerializationDeserialization(Parame
         {
             // test number of properties/fields defined in Column Definition
             int fields = typeof(ParameterDefinition).GetFields(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance).Length;
-            Assert.AreEqual(fields, 4);
+            Assert.AreEqual(fields, 5);
```
seantleonard, line 425: magic number. not sure why this value increased from 4 to 5. I understand that this was existing test, but none of the comments surrounding this code explain the number.

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
seantleonard: new field in `ColumnDefinition` : `public SqlDbType? SqlDbType { get; set; }`

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
seantleonard: new field in `ColumnDefinition` : `public SqlDbType? SqlDbType { get; set; }`

seantleonard (approved):

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -130,6 +130,7 @@ public class ParameterDefinition
 {
     public Type SystemType { get; set; } = null!;
     public DbType? DbType { get; set; }
+    public SqlDbType? SqlDbType { get; set; }
```
aaronburtle, line 133: Can you add a comment that covers the discussion about this only being valid for MSSQL and null for other DBs, with future refactoring potentially to come?

seantleonard: https://github.com/Azure/data-api-builder/issues/2128
https://github.com/Azure/data-api-builder/issues/2127

contain references to the discussion for your question.

src/Core/Parsers/ODataASTVisitor.cs
```diff
@@ -277,6 +277,7 @@ private void PopulateDbTypeForProperty(BinaryOperatorNode nodeIn)
             string? paramName = BaseQueryStructure.GetEncodedParamName(_struct.Counter.Current() - 1);
             _metadataProvider.TryGetBackingColumn(_struct.EntityName, propertyNode.Property.Name, out string? backingColumnName);
             _struct.Parameters[paramName].DbType = _struct.GetUnderlyingSourceDefinition().Columns[backingColumnName!].DbType;
+            _struct.Parameters[paramName].SqlDbType = _struct.GetUnderlyingSourceDefinition().Columns[backingColumnName!].SqlDbType;
```
aaronburtle, line 280: .SqlDbType will always exist here?

src/Core/Resolvers/Sql Query Structures/SqlUpdateQueryStructure.cs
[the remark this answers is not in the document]
aaronburtle: Is there any way that param.Key here can be an invalid column name? Should we verify that it is a valid backing column before assigning?

aaronburtle (approved):
Looks good, the performance increases here are awesome! Just had a couple clarifying questions, a suggestion for some potential comment.

src/Core/Parsers/ODataASTVisitor.cs
abhishekkumams: it will be null for other db types.

tdutch1:
I'm finding the same varchar/nvarchar issue with Azure SQL views. Did the fix described in this issue cover views as well or only tables? If the fix was for tables only, is there a way (e.g., in dab-config.json) to declare a column's data type?

Issue #2019, opened 2024-02-07
Allow a referencing column to reference only one referenced column in relationships


ayush3797:
## Context:
When one column from a referencing table references two columns belonging to the same or different tables, it poses certain problems in dealing with nested insertions. Since this is a bad design practice, we are going to support nested mutation when a referencing column references only one referenced column from any other table.

## What needs to be done?
Ensure that one column in an entity/table is referencing only one column in another entity/table.

### 1. Relationship defined in the database:
This is the case when an FK constraint exist in the database. MsSql allows creating foreign key constraints where one column from a referencing table can be used in multiple Foreign key constraints i.e. one column in the referencing column can reference multiple columns in one or multiple referenced tables. Some simple sql to explain this:

**Case 1: One column in referencing table references two different columns in ONE referenced table.**

Consider these two tables A (_referenced table_) and B (_referencing table_) with 2 FK constraints defined:

```tsql
CREATE TABLE A(
    categoryid int NOT NULL,
    pieceid int NOT NULL,
       anothercol int unique,
    PRIMARY KEY(categoryid,pieceid)
);
```

```tsql
CREATE TABLE B(
    categoryid int NOT NULL,
    pieceid int NOT NULL,
    PRIMARY KEY(categoryid, pieceid)
);
```

```tsql
ALTER TABLE B
ADD CONSTRAINT B_A_fk
FOREIGN KEY (categoryid, pieceid)
REFERENCES A (categoryid, pieceid)
ON DELETE CASCADE;
```

```tsql
ALTER TABLE B
ADD CONSTRAINT B_A_fk2
FOREIGN KEY (categoryid)
REFERENCES A (anothercol)
ON DELETE NO ACTION;
```

Now `B.categoryid` is referencing two columns `categoryid`, `anothercol` from table A. While performing a nested insertion, the value a referencing column (to perform insertion on the referencing table) will come via insertion in the referenced table. However, for this case, where the value of a column B.categoryid can assume value of `A.categoryid` and `A.anothercol` does not seem logical. The foreign key constraints might be placed for some arbitrary real life use case and might still work fine if you’re using sql via say SSMS, but for nested insertion, this does not make logical sense because using nested insertion means the user is expecting a relation between the tables and such constraints will lead to confusions.

**Case 2: One column in referencing table references two different columns in the DIFFERENT referenced table.**

[...]

PR #2057, opened 2024-02-22
Request Validation for nested insertions

ayush3797:
## Why make this change?

  - Include links to any additional discussion threads related to this change.

## What is this change?

  - Links to relevant documentation / StackOverflow, if applicable.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)


seantleonard:
In PR description, starting with "- The method `ValidateObjectFieldNodes()`, I've modified the description to be clearer (at least from my perspective). Please see whether this would be okay to use in your pr description.
--------------------------------------------------------
- The method `ValidateObjectFieldNodes()` contains logic to ensure that when validating the `book` input objects, any (relationship) fields which reference the target entity **Publisher** such as `publisher_id`, should not be specified. In other words, the **Book** field `publisher_id` is a referencing field to the **Publisher** entity's referenced field `id`.
    ```graphql
    mutation {
        createPublisher(item: {
            name: "Penguin the Second"
            books: [
            { title: "magic tree house 1"}, 
            { title: "magic tree house 2"}
            ]
        }){
            id,
            name
        }
    }
    ```
  - The **Book** entity shouldn't include `publisher_id` because `publisher_id`'s value is supplied by the `id` returned from creating the new **Publisher** "Penguin the Second".
  - Note: DAB determines whether an entity is a referencing entity by calling `MultipleCreateOrderHelper.GetReferencingEntityName()` (added via: [Multiple-create: Order determination for insertions](https://github.com/Azure/data-api-builder/pull/2056).
  - How do we determine relationship fields for book? Using the following dab-config.json relationship metadata for book, there is a many-to-one relationship defined. *"Book(many)::Publisher(one)"* or *"many books can be associated to one publisher"*

    ```json
    "Book": {
        "relationships": {
            "publishers": {
            "cardinality": "one",
            "target.entity": "Publisher",
            "source.fields": [],
            "target.fields": [],
            "linking.source.fields": [],
            "linking.target.fields": []
            }
        }
    }
    ```

### `ValidateObjectFieldNodes()` makes calls to

1. `ValidateAbsenceOfReferencingColumnsInTargetEntity()`: Ensures that the entity input being validated does not specify referencing columns to its parent (referenced) entity.
    - For example in the mutation `mutation { Publisher { Book1, Book2, Book3}}`, each `book` entity that the user intends to create should not include **Book**'s *referencing* field `publisher_id` which maps to the (referenced) entity **Publisher**'s `id` field.
    - This validation iterates through each **Book** entity's defined relationships (from DAB's runtime config) to identify which fields are *referencing* fields.
1. `ValidateAbsenceOfRepeatedReferencingColumn()`: Ensures that referencing columns in the referencing entity (`publisher_id` in **Book**) do not reference more than one column in the referenced entity **Publisher**.  This method is called for each relationship field included in the request body for the given entity.
    - Issue tracking config validation for such an occurence [Allow a referencing column to reference only one referenced column in relationships](https://github.com/Azure/data-api-builder/issues/2019)
1. `ValidatePresenceOfRequiredColumnsForInsertion()`: Ensures that a request denotes one source of truth for referencing fields and their associated values.
1. `ValidateRelationshipFields()`: Recursively performs the same validations for all the relationships with the current entity included in the mutation request body.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -96,6 +96,22 @@ public class SqlMutationEngine : IMutationEngine
...
+                GraphQLRequestValidator.ValidateGraphQLValueNode(
+                    schemaForArgument,
+                    entityName,
+                    context,
+                    param,
+                    _runtimeConfigProvider.GetConfig(),
+                    new(),
+                    new(),
+                    0,
+                    string.Empty,
```
seantleonard, line 111: named params would be helpful for the `new()`'s ,`0`, and `string.Empty`

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -197,5 +197,7 @@ public DatabaseObject GetDatabaseObjectForGraphQLType(string graphqlType)
         /// mode, it returns false.
         /// </summary>
         public bool IsDevelopmentMode();
+
+        public ForeignKeyDefinition GetFKDefinition(string sourceEntityName, string targetEntityName, string referencedEntityName, string referencingEntityName) => throw new NotImplementedException();
```
seantleonard, line 201: nit: order of params would be more cohesive if source/target entityName order matched referenced/referencing order:

source
target
referencing
referenced

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -270,5 +270,123 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+
+        /*/// <inheritdoc/>
+        protected override void FillInferredFkInfo(
+            IEnumerable<SourceDefinition> dbEntitiesToBePopulatedWithFK)
+        {
+            // For each table definition that has to be populated with the inferred
+            // foreign key information.
+            foreach (SourceDefinition sourceDefinition in dbEntitiesToBePopulatedWithFK)
+            {
+                // For each source entities, which maps to this table definition
```
seantleonard, line 282: Remove commented code.

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
seantleonard: and need function comment.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1854,6 +1854,26 @@ public bool IsDevelopmentMode()
         {
             return _runtimeConfigProvider.GetConfig().IsDevelopmentMode();
         }
+
+        public ForeignKeyDefinition GetFKDefinition(string sourceEntityName, string targetEntityName, string referencedEntityName, string referencingEntityName)
+        {
+            if (GetEntityNamesAndDbObjects().TryGetValue(sourceEntityName, out DatabaseObject? sourceDbObject) &&
```
seantleonard, line 1860: aren't sourceEntityName and referencingEntityName the same entity?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1854,6 +1854,26 @@ public bool IsDevelopmentMode()
...
+            {
+                DatabaseTable referencingDbTable = (DatabaseTable)referencingDbObject;
+                DatabaseTable referencedDbTable = (DatabaseTable)referencedDbObject;
+                SourceDefinition sourceDefinition = sourceDbObject.SourceDefinition;
+                RelationShipPair referencingReferencedPair = new(referencingDbTable, referencedDbTable);
+                List<ForeignKeyDefinition> fKDefinitions = sourceDefinition.SourceEntityRelationshipMap[sourceEntityName].TargetEntityToFkDefinitionMap[targetEntityName];
+                return fKDefinitions.FirstOrDefault(
+                    fk => fk.Pair.Equals(referencingReferencedPair) &&
+                    fk.ReferencingColumns.Count > 0
+                    && fk.ReferencedColumns.Count > 0)!;
```
seantleonard, line 1872: we can guarantee that fkDefinitions.FirstOrDefault is not null? is that because default is empty FkDefinition?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1854,6 +1854,26 @@ public bool IsDevelopmentMode()
...
+            if (GetEntityNamesAndDbObjects().TryGetValue(sourceEntityName, out DatabaseObject? sourceDbObject) &&
+                GetEntityNamesAndDbObjects().TryGetValue(referencingEntityName, out DatabaseObject? referencingDbObject) &&
+                GetEntityNamesAndDbObjects().TryGetValue(referencedEntityName, out DatabaseObject? referencedDbObject))
+            {
+                DatabaseTable referencingDbTable = (DatabaseTable)referencingDbObject;
+                DatabaseTable referencedDbTable = (DatabaseTable)referencedDbObject;
+                SourceDefinition sourceDefinition = sourceDbObject.SourceDefinition;
+                RelationShipPair referencingReferencedPair = new(referencingDbTable, referencedDbTable);
+                List<ForeignKeyDefinition> fKDefinitions = sourceDefinition.SourceEntityRelationshipMap[sourceEntityName].TargetEntityToFkDefinitionMap[targetEntityName];
+                return fKDefinitions.FirstOrDefault(
```
seantleonard, line 1869: what is being check and returned here? you want to see whether a fkpair exists that matches the RelationshipPair you created above?

seantleonard (commented):
comments so far. Still going through tests and majority of `GraphQLRequestValidator`

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+        public static async Task SetupAsync(TestContext context)
+        {
+            DatabaseEngine = TestCategory.MSSQL;
+            await InitializeTestFixture();
+        }
+
+        /// <summary>
+        /// Test to validate that for 'create one' mutations, we throw an appropriate exception when the user request provides
+        /// no source of value for a non-nullable and non-default referencing column.
+        /// There are 3 sources of values for a referencing column:
```
seantleonard, line 28:
```suggestion
        /// The referencing column can get its value from 3 possible sources:
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+        /// 1. Explicit value assignment to the referencing column,
+        /// 2. Value derived from the parent referenced entity,
+        /// 3. Providing value for inserting record in the referenced target entity, in which case value for referencing column
```
seantleonard, line 31: "Providing value for inserting record in the referenced target entity," is weird phrasing. needs clarity.

perhaps this item number 3 just needs to read: "Insertion of referenced entity will produce a value that can be used in referencing field"

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+        public async Task AbsenceOfSourceOfTruthForReferencingColumnForCreateOneMutations()
+        {
+            // For a relationship between Book (Referencing) - Publisher (Referenced) defined as books.publisher_id -> publisher.id,
+            // in the request input:
+            // 1. No explicit value is provided for referencing field publisher_id
+            // 2. Book is the top-level entity - so there is no parent entity.
+            // 3. No value is assigned for the referenced entity via relationship field ('publisher').
+            string createOneBookMutationName = "createbook";
+            string multipleCreateOneBookWithoutPublisher = @"mutation {
+                    createbook(item: { title: ""My New Book"" }) {
```
seantleonard, line 45: if book.publisher_id were nullable... is that possible / what would happen? would multiple insert validation still reject the request?

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+            // Validate that an appropriate exception is thrown because we have two possibly conflicting sources of
+            // values for the referencing field 'publisher_id'.
+            SqlTestHelper.TestForErrorInGraphQLResponse(
+                    actual.ToString(),
+                    message: "Found conflicting sources of values for the field: publisher_id for entity: Book at level: 1." +
+                    "Source 1: entity: Book ,Source 2: Relationship: publishers.",
+                    statusCode: DataApiBuilderException.SubStatusCodes.BadRequest.ToString()
+                );
+
+            // For a relationship between Review (Referencing) - Book (Referenced) defined as reviews.book_id -> books.id,
```
seantleonard, line 152: When reading the mutation left to right, it would be helpful if this description were consistent with reading the mutation:
 For a relationship between Book (Referenced) - Review (Referencing)  defined as books.id <- reviews.book_id

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+            // values for the referencing field 'book_id'.
+            SqlTestHelper.TestForErrorInGraphQLResponse(
+                    actual.ToString(),
+                    message: "The field: book_id cannot be present for entity: Review at level: 2.",
+                    statusCode: DataApiBuilderException.SubStatusCodes.BadRequest.ToString()
+                );
+
+            string createOnePublisherName = "createPublisher";
+
+            // For a relationship between Book (Referencing) - Publisher (Referenced) defined as books.publisher_id -> publisher.id,
```
seantleonard, line 180:
```suggestion
            // For a relationship between Publisher (Referenced) - Book (Referencing) defined as publisher.id <- books.publisher_id,
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+                    statusCode: DataApiBuilderException.SubStatusCodes.BadRequest.ToString()
+                );
+        }
+
+        /// <summary>
+        /// Test to validate that for 'create one' mutations, we throw an appropriate exception when the user request provides
+        /// multiple sources of value for a referencing column. There are 3 sources of values for a referencing column:
+        /// -> Explicit value assignment to the referencing column,
+        /// -> Value derived from the parent referenced entity,
+        /// -> Providing value for inserting record in the referenced target entity, in which case value for referencing column
```
seantleonard, line 116: do the three test cases in this testmethod exercise different validation functions? if so, could they be split up into separate test methods where each testmethod comment gives context as to what validation code (validation function(s)) is(are) exercised?

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+            // Validate that an appropriate exception is thrown because we have two possibly conflicting sources of
+            // values for the referencing field 'publisher_id'.
+            SqlTestHelper.TestForErrorInGraphQLResponse(
+                    actual.ToString(),
+                    message: "Found conflicting sources of values for the field: publisher_id for entity: Book at level: 1." +
+                    "Source 1: entity: Book ,Source 2: Relationship: publishers.",
+                    statusCode: DataApiBuilderException.SubStatusCodes.BadRequest.ToString()
+                );
+
+            // For a relationship between Review (Referencing) - Book (Referenced) defined as reviews.book_id -> books.id,
```
seantleonard, line 259:
```suggestion
            // For a relationship between Book (Referenced) - Review (Referencing) defined as books.id <- reviews.book_id,
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+            // values for the referencing field 'book_id'.
+            SqlTestHelper.TestForErrorInGraphQLResponse(
+                    actual.ToString(),
+                    message: "The field: book_id cannot be present for entity: Review at level: 2.",
+                    statusCode: DataApiBuilderException.SubStatusCodes.BadRequest.ToString()
+                );
+
+            string createMultiplePublisherName = "createPublishers";
+
+            // For a relationship between Book (Referencing) - Publisher (Referenced) defined as books.publisher_id -> publisher.id,
```
seantleonard, line 289:
```suggestion
            // For a relationship between Publisher (Referenced) - Book (Referencing) defined as publisher.id <- books.publisher_id,
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
seantleonard, line 1: I found the examples in these test cases VERY VERY helpful. Thank you :)

seantleonard (changes requested):
need more clarity and i believe more test coverage because there are many instances of exceptions being thrown, but not being checked in test cases.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+            {
+                if (referencingFields.Contains(referencingColumn))
+                {
+                    sqlMetadataProvider.TryGetExposedColumnName(referencedEntityName, referencingColumn, out string? exposedReferencingColumnName);
+                    // This indicates one column is holding reference to multiple referenced columns in the related entity,
+                    // which leads to possibility of two conflicting sources of truth for this column.
+                    // This is an invalid use case for multiple-create.
+                    throw new DataApiBuilderException(
+                        message: $"The field: {exposedReferencingColumnName} in the entity: {referencedEntityName} references multiple fields in the" +
+                        $"related entity: {referencingEntityName} for the relationship: {relationshipName} at level: {nestingLevel}.",
```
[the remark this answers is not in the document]
ayush3797: Sorry not clear with what you mean by which entity's relationship is at fault? Aren't both the entities related by same relationship?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: No. Its not always the case. Source entity/target entity depend on the config how they are define. 
Eg. For a relationship between Book-Publisher defined in the Publisher's configuration:

source entity: Publisher
target entity: Book
Referencing entity: Book
Referenced entity: Publisher

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: This is because we are sure we always have a valid FK. We validate it during startup by using the method `SqlMetadatProvider.ValidateAllFkHaveBeenInferred()`.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
ayush3797: Yes. The Fk definition where referencingTable = referencingEntity's table and referencedTable = referencedEntity's table and where referencingColumns,referencedColumns count >0 (so that we have fields to relate the two entities).

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+            HashSet<string> columnsDerivedFromSourceEntity,
+            HashSet<string> columnsToBeDerivedFromThisEntity,
+            int nestingLevel,
+            string parentEntityName,
+            IMetadataProviderFactory sqlMetadataProviderFactory)
+        {
+            string dataSourceName = GraphQLUtils.GetDataSourceNameFromGraphQLContext(context, runtimeConfig);
+            ISqlMetadataProvider metadataProvider = sqlMetadataProviderFactory.GetMetadataProvider(dataSourceName);
+            SourceDefinition sourceDefinition = metadataProvider.GetSourceDefinition(entityName);
+            Dictionary<string, IValueNode?> backingColumnData = MultipleCreateOrderHelper.GetBackingColumnDataFromFields(context, entityName, objectFieldNodes, metadataProvider);
```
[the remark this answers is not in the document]
ayush3797: `GetBackingColumnDataFromFields` -> Gets the backing column data from fields. The column data can only be obtained for fields which represent a column. Hence, in your example, when this method is called for Publisher, it will return `name`, for Book, it will return `title`.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+            }
+
+            // For the relationships with the parent entity, where the current entity is a referenced entity,
+            // we need to make sure that we have non-null values for all the referenced columns - since the values for all those
+            // columns will be used for insertion in the parent entity. We can get the referenced column value:
+            // Case 1. Via an absolute value
+            // Case 2. Via another relationship where a referenced column by the parent of the current entity is a referencing column for a child of the current entity.
+
+            // Case 1: Remove from columnsToBeDerivedFromEntity, the columns which are autogenerated or
+            // have been provided a non-null value in the input for the current entity.
```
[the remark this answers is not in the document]
ayush3797: I would refrain using `non-null` columns as being null/non-null is a column property.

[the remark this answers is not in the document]
ayush3797: Added `why`

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+            {
+                (IValueNode? fieldValue, SyntaxKind fieldKind) = GraphQLUtils.GetFieldDetails(fieldNode.Value, context.Variables);
+                if (fieldKind is SyntaxKind.NullValue || GraphQLUtils.IsScalarField(fieldKind))
+                {
+                    // If the current field is a column/scalar field or has a null value, continue.
+                    continue;
+                }
+
+                string relationshipName = fieldNode.Name.Value;
+                string targetEntityName = runtimeConfig.Entities![entityName].Relationships![relationshipName].TargetEntity;
```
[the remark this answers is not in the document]
ayush3797: For non-relationship fields, we continue from the code just above this line.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+                string? linkingObject = runtimeConfig.Entities![entityName].Relationships![relationshipName].LinkingObject;
+                if (!string.IsNullOrWhiteSpace(linkingObject))
```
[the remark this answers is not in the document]
ayush3797: I think rather than using all those nested dictionary references altogether in the if check, it makes it easier to use a variable telling us what the if check is performed on.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+                                message: $"Found conflicting sources of values for the field: {exposedColumnName} for entity: {entityName} at level: {nestingLevel}." +
+                                $"Source 1: {conflictingSource} ,Source 2: Relationship: {relationshipName}.",
+                                statusCode: HttpStatusCode.BadRequest,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                        }
+
+                        // Case 2: When a column whose value is to be derived from the insertion in current entity
+                        // (happens when the parent entity is a referencing entity in a relationship with current entity),
+                        // is a referencing column in the current relationship, we pass on the responsibility of getting the value
+                        // of such a column to the child entity in the current relationship.
```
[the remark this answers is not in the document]
ayush3797: No, I mean pass on the responsibility to the child entity. When the parent entity is the referencing entity, the current entity needs to return values of the referenced columns. But if those referenced columns are themselves the referencing columns for the relationship of current entity with the child entity, the values for those referenced columns are actually derived from the insertion in child entity. 
Consider 3 entities A,B,C appearing in the same order in the mutation:

A.aid references B.bid
B.bid references C.cid

Insertion order -> C followed by B followed by A
value of C.cid will be allotted to B.bid 
and then value of B.bid will be allotted to A.aid.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+                Tuple<IValueNode?, SyntaxKind> fieldDetails = GraphQLUtils.GetFieldDetails(field.Value, context.Variables);
+                SyntaxKind fieldKind = fieldDetails.Item2;
+
+                // For non-scalar fields, i.e. relationship fields, we have to recurse to process fields in the relationship field -
+                // which represents input data for a related entity.
+                if (fieldKind is not SyntaxKind.NullValue && !GraphQLUtils.IsScalarField(fieldKind))
+                {
+                    string relationshipName = field.Name.Value;
+                    string targetEntityName = runtimeConfig.Entities![entityName].Relationships![relationshipName].TargetEntity;
+                    HashSet<string>? derivedColumnsForEntity, columnsToBeDerivedFromEntity;
```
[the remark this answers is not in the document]
ayush3797: IMHO, this seems less intuitive I changed the name to `columnsToDerive` which now makes more sense and updated the param summary as well.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+        /// <param name="targetEntityName">Name of the target entity.</param>
+        /// <param name="metadataProvider">Metadata provider.</param>
+        private static void ValidateAbsenceOfReferencingColumnsInTargetEntity(
+            Dictionary<string, string> columnsInTargetEntity,
+            HashSet<string> derivedColumnsFromSourceEntity,
+            int nestingLevel,
+            string targetEntityName,
+            ISqlMetadataProvider metadataProvider)
+        {
+            foreach (string derivedColumnFromParentEntity in derivedColumnsFromSourceEntity)
```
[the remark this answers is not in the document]
ayush3797: Source means the upper level entity in the mutation. It cannot be child. Child is always the target entity in a relationship. Basically child = target. I will update referenced to child by target entity since this seems to be causing confusion.

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
ayush3797: updated as appropriate

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
ayush3797: No, it would not reject. If there are no sources of value for a referencing field, it means you are not performing a related insertion. You are just performing an insertion on the independent Book entity. And it is fine to insert null value for publisher_id then - because it is not used to relate the records with records inserted in publisher entity.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+                    }
+                }
+                else
+                {
+                    columnsDerivedForRelationships.Add(relationshipName, new(fkDefinition.ReferencingColumns));
+                }
+            }
+
+            // After determining all the fields that can be derived for the current entity either:
+            // 1. Via absolute value,
```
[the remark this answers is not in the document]
ayush3797: yes updated

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+        /// <param name="relationshipName">Name of the relationship.</param>
+        /// <param name="nestingLevel">Current depth of nesting in the multiple-create request.</param>
+        /// <param name="sqlMetadataProvider">Metadata provider.</param>
+        private static void ValidateAbsenceOfRepeatedReferencingColumn(
+            List<string> referencingColumns,
+            string referencingEntityName,
+            string referencedEntityName,
+            string relationshipName,
+            int nestingLevel,
+            ISqlMetadataProvider sqlMetadataProvider)
```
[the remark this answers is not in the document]
ayush3797: Yeah that's better. I'll update.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+            foreach (string columnDerivedFromSourceEntity in columnsDerivedFromSourceEntity)
+            {
+                derivableColumnsFromRequestBody.TryAdd(columnDerivedFromSourceEntity, parentEntityName);
+            }
+
+            // For the relationships with the parent entity, where the current entity is a referenced entity,
+            // we need to make sure that we have non-null values for all the referenced columns - since the values for all those
+            // columns will be used for insertion in the parent entity. We can get the referenced column value:
+            // Case 1. Via an absolute value
+            // Case 2. Via another relationship where a referenced column by the parent of the current entity is a referencing column for a child of the current entity.
```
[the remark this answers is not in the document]
ayush3797: Refer to this comment: https://github.com/Azure/data-api-builder/pull/2057#discussion_r1533468575. In our current config there is no such set of entities which will enable me to give an example from there.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+                {
+                    columnsDerivedForRelationships.Add(relationshipName, new(fkDefinition.ReferencingColumns));
+                }
+            }
+
+            // After determining all the fields that can be derived for the current entity either:
+            // 1. Via absolute value,
+            // 2. Via an autogenerated value from the database,
+            // 3. Via Insertion in a referenced child entity in a relationship,
+            // if there are still columns which are yet to be derived, this means we don't have sufficient data to perform insertion.
```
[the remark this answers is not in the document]
ayush3797: > 3. Via Insertion in a referenced child entity in a relationship,

An example of creating book with related/nested publisher is a typical example of this where insertion in referenced entity (Publisher) gives Book back the publisher_id.

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,425 @@
...
+        /// <summary>
+        /// Test to validate that when one referencing column references multiple referenced columns in multiple referenced entities,
+        /// we fail the request during request validation because then we have conflicting sources of truth for values for
+        /// the referencing column. In such cases, the referencing column can assume the value of any referenced column in the referenced entities
+        /// which leads to amibugites as to what value to assign to the referencing column.
+        /// </summary>
+        [TestMethod]
+        public async Task InvalidateRepeatedReferencingColumnToMultipleReferencedEntitiesInReferencingEntity()
+        {
+            // The entity createUserProfile_RepeatedRelationshipColumnToTwoEntities acts as the referencing entity the two relationships:
```
ayush3797, line 385:
```suggestion
            // The entity createUserProfile_RepeatedRelationshipColumnToTwoEntities acts as the referencing entity for the two relationships:
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,425 @@
...
+            // 1. UserProfile_RepeatedReferencingColumnToTwoEntities.userid = books.id
+            // 2. UserProfile_RepeatedReferencingColumnToTwoEntities.userid = publishers.id,
```
ayush3797, line 387:
```suggestion
            // 1. UserProfile_RepeatedReferencingColumnToTwoEntities.userid -> books.id
            // 2. UserProfile_RepeatedReferencingColumnToTwoEntities.userid -> publishers.id,
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,425 @@
...
+            // due to which we have two sources of truth for the column userid - books.id, publishers.id
+            // which causes ambiguity as to what value should be assigned to the createUserProfile_RepeatedRelationshipColumnToTwoEntities.userid
+            // field after performing insertion in the referenced UserProfile entity.
```
ayush3797, line 390:
```suggestion
            // field after performing insertion in the two referenced entities, i.e. Book and Publisher.
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1909,6 +1909,35 @@ public bool IsDevelopmentMode()
...
+                // At this point, we are sure that a valid foreign key definition would exist from the referencing entity
+                // to the referenced entity because we validate it during the startup that the Foreign key information
+                // has been inferred for all the relationships.
+                return fKDefinitions.FirstOrDefault(
+                    fk => fk.Pair.Equals(referencingReferencedPair) &&
+                    fk.ReferencingColumns.Count > 0
+                    && fk.ReferencedColumns.Count > 0)!;
+            }
+
+            return new();
```
severussundar, line 1939: Why not throw an exception be when the database object for source entity, Referencing entity or referenced entity does not exist?

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -208,5 +208,28 @@ public DatabaseObject GetDatabaseObjectForGraphQLType(string graphqlType)
         void InitializeAsync(
             Dictionary<string, DatabaseObject> entityToDatabaseObject,
             Dictionary<string, string> graphQLStoredProcedureExposedNameToEntityNameMap);
+
+        /// <summary>
+        /// Helper method to get the Foreign Key definition in the object definition of the source entity which relates it
+        /// with the target entity. In the Foreign key definition, the referencing entity acts as the referencing table and the
```
severussundar, line 214:
```suggestion
        /// with the target entity. In the Foreign key definition, the table backing the referencing entity acts as the referencing table and the
```

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -208,5 +208,28 @@ public DatabaseObject GetDatabaseObjectForGraphQLType(string graphqlType)
...
+        /// referenced entity acts as the referenced table.
```
severussundar, line 215:
```suggestion
        /// the table backing the referenced entity acts as the referenced table.
```

src/Core/Resolvers/MultipleCreateOrderHelper.cs
```diff
@@ -347,7 +347,8 @@ public class MultipleCreateOrderHelper
             {
                 Tuple<IValueNode?, SyntaxKind> fieldDetails = GraphQLUtils.GetFieldDetails(field.Value, context.Variables);
                 SyntaxKind fieldKind = fieldDetails.Item2;
-                if (GraphQLUtils.IsScalarField(fieldKind) && metadataProvider.TryGetBackingColumn(entityName, field.Name.Value, out string? backingColumnName))
+                if ((GraphQLUtils.IsScalarField(fieldKind) || fieldKind is SyntaxKind.NullValue)
```
severussundar, line 350: Can the condition to check `fieldKind is SyntaxKind.NullValue` be part of the implementation of the function `GraphQLUtils.IsScalkarField(fieldKind)` ?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -96,6 +96,23 @@ public class SqlMutationEngine : IMutationEngine
             // If authorization fails, an exception will be thrown and request execution halts.
             AuthorizeMutation(context, parameters, entityName, mutationOperation);
 
+            string inputArgumentName = IsPointMutation(context) ? MutationBuilder.ITEM_INPUT_ARGUMENT_NAME : MutationBuilder.ARRAY_INPUT_ARGUMENT_NAME;
```
severussundar, line 99:
```suggestion
           // Multiple create mutation request is validated to ensure that the request is valid semantically. 

            string inputArgumentName = IsPointMutation(context) ? MutationBuilder.ITEM_INPUT_ARGUMENT_NAME : MutationBuilder.ARRAY_INPUT_ARGUMENT_NAME;
```

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -208,5 +208,28 @@ public DatabaseObject GetDatabaseObjectForGraphQLType(string graphqlType)
...
+        /// </summary>
+        /// <param name="sourceEntityName">Source entity name.</param>
+        /// <param name="targetEntityName">Target entity name.</param>
+        /// <param name="referencedEntityName">Referenced entity name.</param>
+        /// <param name="referencingEntityName">Referencing entity name.</param>
+        /// <returns></returns>
```
severussundar, line 221: nit:
```suggestion
        /// <returns>Foreign Key Definition between the given pair of source and target entities with the specified referencing and referenced entities.</returns>
```

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+        /// <param name="nestingLevel">Current depth of nesting in the multiple-create request.</param>
+        /// <param name="parentEntityName">Parent entity's name.</param>
+        /// <param name="sqlMetadataProviderFactory">Metadata provider factory.</param>
+        /// <example>       1. mutation {
+        ///                 createbook(
+        ///                     item: {
+        ///                         title: "book #1",
+        ///                         reviews: [{ content: "Good book." }, { content: "Great book." }],
+        ///                         publisher: { name: "Macmillan publishers" },
+        ///                         authors: [{ birthdate: "1997-09-03", name: "Red house authors", author_name: "Dan Brown" }]
```
severussundar, line 43: nit: If we go by the existing schema and config setup, the field name would be `name`
```suggestion
        ///                         authors: [{ birthdate: "1997-09-03", name: "Red house authors", name: "Dan Brown" }]
```

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+        /// <param name="runtimeConfig">Runtime config</param>
+        /// <param name="columnsDerivedFromParentEntity">Set of columns in this entity whose values should be
+        /// derived from insertion in the source entity (i.e. source entity would have been the referenced entity).</param>
+        /// <param name="columnsToBeDerivedFromEntity">Set of columns in this entity whose values are to be
+        /// derived from insertion in the this entity and returned to the source entity so as to provide values for
+        /// the corresponding referencing fields (i.e. source entity would have been the referencing entity).</param>
+        /// <param name="nestingLevel">Current depth of nesting in the multiple-create request.</param>
+        /// <param name="parentEntityName">Parent entity's name.</param>
+        /// <param name="sqlMetadataProviderFactory">Metadata provider factory.</param>
+        /// <example>       1. mutation {
```
severussundar, line 37: These examples look like valid requests. If the intention was to outline few examples of invalid requests which the method will catch successfully, adding a one-liner as to why the request is invalid will help.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+                    entityName: entityName,
+                    context: context,
+                    parameters: objectValueNodeInListValueNode,
+                    runtimeConfig: runtimeConfig,
+                    columnsDerivedFromParentEntity: columnsDerivedFromParentEntity,
+                    columnsToBeDerivedFromEntity: columnsToBeDerivedFromEntity,
+                    nestingLevel: nestingLevel,
+                    parentEntityName: parentEntityName,
+                    sqlMetadataProviderFactory: sqlMetadataProviderFactory));
+            }
```
severussundar, line 143: Add an else case that throws an exception with the message along the lines of parameters not being expected to be of any other type

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+            // c. Insertion in the target referenced entity (if the current entity is a referencing entity for a relationship with its target entity)
+            //
+            // derivableColumnsFromRequestBody is initialized with the set of columns whose value is specified in the input (a).
+            Dictionary<string, string> derivableColumnsFromRequestBody = new();
+            foreach ((string backingColumnName, _) in backingColumnData)
+            {
+                derivableColumnsFromRequestBody.TryAdd(backingColumnName, entityName);
+            }
+
+            // When the parent entity is a referenced entity in a relationship, the values of the referencing columns
```
severussundar, line 192: How can we be sure at this point that the parent entity is the referenced entity?

This validation runs when the parent entity is the referencing as well as the referenced entity right? Should this be run after making sure that parent entity is the referenced entity?

severussundar (commented):
Posting comments so far, need to look at tests

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+            foreach (string columnDerivedFromSourceEntity in columnsDerivedFromSourceEntity)
+            {
+                derivableColumnsFromRequestBody.TryAdd(columnDerivedFromSourceEntity, parentEntityName);
+            }
+
+            // For the relationships with the parent entity, where the current entity is a referenced entity,
+            // we need to make sure that we have non-null values for all the referenced columns - since the values for all those
+            // columns will be used for insertion in the parent entity. We can get the referenced column value:
+            // Case 1. Via a scalar value provided in the input for the current entity
+            // Case 2. Via another relationship where a column referenced by the parent of the current entity is a
```
severussundar, line 212: For Case 2, supplementing it with an example would make it easier to understand

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+            foreach (string columnToBeDerivedFromEntity in columnsToDerive)
+            {
+                if (sourceDefinition.Columns[columnToBeDerivedFromEntity].IsAutoGenerated)
+                {
+                    // The value for an autogenerated column is derivable.
+                    columnsToDerive.Remove(columnToBeDerivedFromEntity);
+                }
+                else if (backingColumnData.TryGetValue(columnToBeDerivedFromEntity, out IValueNode? value))
+                {
+                    metadataProvider.TryGetExposedColumnName(entityName, columnToBeDerivedFromEntity, out string? exposedColumnName);
```
severussundar, line 231: Fetching the exposed column name can be moved inside the `if` as the exposed column name is used only for the error message

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+            // Case 1: Remove from columnsToBeDerivedFromEntity, the columns which are autogenerated or
+            // have been provided a non-null value in the input for the current entity.
+
+            // As we iterate through columns which can derived from the current entity, we keep removing them from the
+            // columnsToDerive because we know the column value can be derived.
+            // At the end, if there are still columns yet to be derived, i.e.columnsToDerive.count > 0,
+            // we throw an exception.
+            foreach (string columnToBeDerivedFromEntity in columnsToDerive)
+            {
+                if (sourceDefinition.Columns[columnToBeDerivedFromEntity].IsAutoGenerated)
```
severussundar, line 224: How about columns with default values?

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+        /// </summary>
+        /// <param name="schemaObject">Input object type for the field.</param>
+        /// <param name="entityName">Current entity's name.</param>
+        /// <param name="context">Middleware Context.</param>
+        /// <param name="objectFieldNodes">List of ObjectFieldNodes for the the input field.</param>
+        /// <param name="runtimeConfig">Runtime config</param>
+        /// <param name="columnsDerivedFromSourceEntity">Set of columns in this entity whose values should be
+        /// derived from insertion in the source entity (i.e. source entity would have been the referenced entity).</param>
+        /// <param name="columnsToDerive">Set of columns in this entity whose values are to be
+        /// derived from insertion in the this entity or its subsequent referenced entities and returned to the source entity
```
severussundar, line 158:
```suggestion
        /// <param name="columnsToDerive">Set of columns in this entity whose values are to be
        /// derived from insertion in its referenced entities and returned to the source entity
```

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+                            subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                    }
+
+                    columnsToDerive.Remove(columnToBeDerivedFromEntity);
+                }
+            }
+
+            // Dictionary storing the mapping from relationship name to the set of referencing fields for all
+            // the relationships for which the current entity is the referenced entity.
+            Dictionary<string, HashSet<string>> columnsDerivedForRelationships = new();
```
severussundar, line 246:
```suggestion
            Dictionary<string, HashSet<string>> fieldsToBeSuppliedForReferencingEntities = new();
```
or something along these lines.. 

Suggestion to remove `derived` from this variable name as `derived` has an undertone of fetching PKs from the creation of other entities but the intent of the variable here is to hold PKs to supply to other entities

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+            SourceDefinition sourceDefinition = metadataProvider.GetSourceDefinition(entityName);
+            Dictionary<string, IValueNode?> backingColumnData = MultipleCreateOrderHelper.GetBackingColumnDataFromFields(context, entityName, objectFieldNodes, metadataProvider);
+
+            // Set of columns in the current entity whose values can be derived via:
+            // a. User input
+            // b. Insertion in the source referenced entity (if current entity is a referencing entity for a relationship with its parent entity)
+            // c. Insertion in the target referenced entity (if the current entity is a referencing entity for a relationship with its target entity)
+            //
+            // derivableColumnsFromRequestBody is initialized with the set of columns whose value is specified in the input (a).
+            Dictionary<string, string> derivableColumnsFromRequestBody = new();
```
severussundar, line 186:
```suggestion
            Dictionary<string, string> currentEntityFieldsPresentInTheRequestBody = new();
```

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+        /// validate that the presence/absence of fields make logical sense for the multiple-create request.
+        /// </summary>
+        /// <param name="schemaObject">Input object type for the field.</param>
+        /// <param name="entityName">Current entity's name.</param>
+        /// <param name="context">Middleware Context.</param>
+        /// <param name="objectFieldNodes">List of ObjectFieldNodes for the the input field.</param>
+        /// <param name="runtimeConfig">Runtime config</param>
+        /// <param name="columnsDerivedFromSourceEntity">Set of columns in this entity whose values should be
+        /// derived from insertion in the source entity (i.e. source entity would have been the referenced entity).</param>
+        /// <param name="columnsToDerive">Set of columns in this entity whose values are to be
```
severussundar, line 157:
```suggestion
        /// <param name="columnsReferencedBySourceEntity">Set of columns in this entity whose values are to be
```
(or)
```suggestion
        /// <param name="columnsToBeSuppliedToSourceEntity">Set of columns in this entity whose values are to be
```

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,575 @@
...
+                        // Case 2: When a column whose value is to be derived from the insertion in current entity
+                        // (happens when the parent entity is a referencing entity in a relationship with current entity),
+                        // is a referencing column in the current relationship, we pass on the responsibility of getting the value
+                        // of such a column to the target entity in the current relationship.
+                        if (columnsToDerive.Contains(referencingColumn))
+                        {
+                            // We optimistically assume that we will get the value of the referencing column
+                            // from the insertion in the target entity.
+                            columnsToDerive.Remove(referencingColumn);
+                        }
```
severussundar, line 362: This is when the parent entity turns out to be the referencing entity. But, we already know that the current entity is the referencing entity ---> L322.

So, can we move this out of the `if` block --> block that checks that the current entity is the referencing entity

src/Core/Resolvers/MultipleCreateOrderHelper.cs
ayush3797: NullValue can be for a relationship field as well as for a column field. The value `NULL` does not fit into the definition of either being a complex (relationship) field or a scalar field.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: This example is to show the catgeorisation as to which input will be processed by which IF/ELSEIF block. Similar to what was done in authz processing. There are a lot of invalid scenarios which can occur and to provide examples for each of them here will be an overkill I think because I will then have to explain as well, why the request is invalid which I think is best handled when each of the invalid scenarios are actually validated.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: Supporting column with default values does not make sense because:

1. We only know the column has a default value. But we need to get the exact value to be sure that the column value is not null. This requires additional metadata collection.
2. Is not a very useful/intuitive/probable to occur scenario. If you're doing a multiple-create, what makes sense is to provide data/values for fields which are related. (obviously when the field is not autogenerated).
3. The order determination logic gets unnecessarily complicated.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: The current summary I think better explains what I am trying to convey here. Please let me know if you have any questions/confusions regarding this.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: I think the current variable name is apt to represent what it intends to store. Please refer to the comments preceding this variable explaining what all it can store.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: added

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: This is the case when parent is referencing child and child is referencing its child.

A->B->C. I have added an example here: https://github.com/Azure/data-api-builder/blob/e0fbd36687017f4a289d3143bbd09364f94d325d/src/Core/Services/MultipleMutationInputValidator.cs#L218. The code block which you have highlighted specifically covers this case.

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -319,7 +317,58 @@ public async Task PresenceOfMultipleSourcesOfTruthForReferencingColumnForCreateM
...
+                queryName: createUserRepeatedRelationshipColumn,
+                isAuthenticated: false,
+                variables: null,
+                clientRoleHeader: "anonymous");
+
+            // Validate that an appropriate exception is thrown because the value for the referencing field
+            // publisher_id cannot be derived from any of the two possible sources mentioned above.
+            SqlTestHelper.TestForErrorInGraphQLResponse(
+                    actual.ToString(),
+                    message: "The fields: {username} in the entity: User_RepeatedReferencingColumn references multiple fields in the " +
```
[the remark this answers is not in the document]
ayush3797: There can be more than 1 field here and hence used `fields`.

src/Core/Services/GraphQLRequestValidator.cs
```diff
@@ -0,0 +1,538 @@
...
+                // all the fields present for item namely- title, reviews, publisher, authors are interpreted as ObjectFieldNode.
+                ValidateObjectFieldNodes(
+                    schemaObject: ResolverMiddleware.InputObjectTypeFromIInputField(schema),
+                    entityName: entityName,
+                    context: context,
+                    objectFieldNodes: listOfObjectFieldNode,
+                    runtimeConfig: runtimeConfig,
+                    columnsDerivedFromSourceEntity: columnsDerivedFromParentEntity,
+                    columnsToBeDerivedFromThisEntity: columnsToBeDerivedFromEntity,
+                    nestingLevel: nestingLevel + 1,
```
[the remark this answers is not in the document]
ayush3797: Added as explanation in the param summary.

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
ayush3797: I think the current setup makes sense in terms of keeping the tests with same concern together. I will add a comment to the test case which excercises a particular method.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+                Tuple<IValueNode?, SyntaxKind> fieldDetails = GraphQLUtils.GetFieldDetails(field.Value, context.Variables);
+                SyntaxKind fieldKind = fieldDetails.Item2;
+
+                // For non-scalar fields, i.e. relationship fields, we have to recurse to process fields in the relationship field -
+                // which represents input data for a related entity.
+                if (fieldKind is not SyntaxKind.NullValue && !GraphQLUtils.IsScalarField(fieldKind))
+                {
+                    string relationshipName = field.Name.Value;
+                    string targetEntityName = runtimeConfig.Entities![entityName].Relationships![relationshipName].TargetEntity;
+                    HashSet<string>? derivedColumnsForEntity, columnsToBeDerivedFromEntity;
```
seantleonard, line 553:
```suggestion
                    HashSet<string>? columnValuePairsSourcedFromCurrentEntity, columnsToBeDerivedFromEntity;
```

- does `columnsToBeDerivedFromEntity` mean that the child entity must resolve values for these columns from within it's own input/autogenerated column values, or even a further nested "referenced" entity? Could this be named: `columnValuePairsToBeResolved`(+ `byChild`)?

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+        /// derived from insertion in the source entity (i.e. source entity would have been the referenced entity).</param>
+        /// <param name="columnsToBeSuppliedToSourceEntity">Set of columns in this entity whose values are to be
+        /// derived from insertion in this entity or its subsequent referenced entities and returned to the source entity
+        /// so as to provide values for the corresponding referencing fields (i.e. source entity would have been the referencing entity).</param>
+        /// <param name="nestingLevel">Current depth of nesting in the multiple-create request. As we go down in the multiple mutation from the source
+        /// entity input to target entity input, the nesting level increases by 1. This helps us to throw meaningful exception messages to the user
+        /// as to at what level in the multiple mutation the exception occurred.</param>
+        /// <param name="parentEntityName">Parent entity's name.</param>
+        /// <param name="sqlMetadataProviderFactory">Metadata provider factory.</param>
+        private static void ValidateObjectFieldNodes(
```
seantleonard, line 173: this was a huge function to read through/evaluate. Was there no way to break this up + have functions with smaller method signatures?

If this were a service instantiated at startup as non static and `scoped`(per request) service lifetime per https://learn.microsoft.com/dotnet/core/extensions/dependency-injection#scoped, you could use DI to resolve `RuntimeConfigProvider`/`RuntimeConfig`, `IMetadataProviderFactory`, reducing your parameter count everywhere in this file by 2 and allowing you to create short/low param count helper functions that don't need runtimeconfig/metadataprovider passed to them.

You may be able to further reduce parameter count (which increases readability) by introducing a `MultipleMutationRequestValidationContext` object which contains relevant/related properties such as
- `currentEntityName`
- `parentEntityName`
- `columnsDerivedFromSourceEntity`
- `columnsToBeSuppliedToSourceEntity`
- nestingLevel

but that may be for another PR. I just worry that in this may be tough for others to come in and understand without indepth exposure/understanding of our design discussions.

seantleonard: and IMiddlewareContext is stored once per-request into a private var. That would require you to have a separate function that is recursed .

src/Service.Tests/Unittests/MultipleCreateUnitTests/MySqlMultipleCreateOrderHelperUnitTests.cs
seantleonard, line 1: comment about whether Multiple inserts is supported for pg/mysql

src/Service.Tests/Unittests/MultipleCreateUnitTests/PgSqlMultipleCreateOrderHelperUnitTests.cs
seantleonard, line 1: comment about whether Multiple inserts is supported for pg/mysql

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,429 @@
...
+                queryName: createmultipleBooksMutationName,
+                isAuthenticated: true,
+                variables: null,
+                clientRoleHeader: "authenticated");
+
+            // Validate that an appropriate exception is thrown because the value for the referencing field
+            // publisher_id cannot be derived from any of the two possible sources mentioned above.
+            SqlTestHelper.TestForErrorInGraphQLResponse(
+                    actual.ToString(),
+                    message: "Missing value for required column: publisher_id for entity: Book at level: 1.",
```
seantleonard, line 104: nit: we say required column here when in the GraphQL schema the field isn't required. Wouldn't it be more actionable to say that either `publisher_id` OR `publishers` must supply a value?

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,429 @@
...
+                queryName: createOneBookMutationName,
+                isAuthenticated: true,
+                variables: null,
+                clientRoleHeader: "authenticated");
+
+            // Validate that an appropriate exception is thrown because we have two possibly conflicting sources of
+            // values for the referencing field 'publisher_id'.
+            SqlTestHelper.TestForErrorInGraphQLResponse(
+                    actual.ToString(),
+                    message: "Found conflicting sources of values for the field: publisher_id for entity: Book at level: 1." +
```
seantleonard, line 146: perhaps:
 - "Found conflicting sources providing a value for the field..."
 - "Found conflicting value sources for the field..."
 
 "sources of values" reads awkwardly to me.

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,429 @@
...
+                );
+        }
+
+        /// <summary>
+        /// Test to validate that when one referencing column references multiple referenced columns in one referenced entity,
+        /// we fail the request during request validation because then we have conflicting sources of truth for values for
+        /// the referencing column. In such cases, the referencing column can assume the value of any referenced column which
+        /// leads to amibugites as to what value to assign to the referencing column.
+        /// </summary>
+        [TestMethod]
```
seantleonard, line 335:
```suggestion
        /// Test to validate that when one referencing column references multiple referenced columns in one referenced entity,
        /// we fail the request during request validation because then we have conflicting sources of truth for values for
        /// the referencing column. In these instances, the column that is being referenced can take on the value from any
        /// column it references. This can create uncertainty about which value should be assigned to the column being referenced.
        /// </summary>
        [TestMethod]
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,429 @@
...
+        public async Task InvalidateRepeatedReferencingColumnToOneReferencedEntityInReferencingEntity()
+        {
+            // For a relationship between User_RepeatedReferencingColumnToOneEntity(username,username) - UserProfile (username, profilepictureurl),
+            // when the User_RepeatedReferencingColumnToOneEntity entity acts as the referencing entity, there are two sources
+            // of truth for the value of User_RepeatedReferencingColumnToOneEntity.username:
+            // 1. UserProfile.username
+            // 2. UserProfile.profilepictureurl
+            // which causes ambiguity as to what value should be assigned to the User_RepeatedReferencingColumnToOneEntity.username
```
seantleonard, line 343:
```suggestion
            // which leads to uncertainty about the value that should be assigned to the User_RepeatedReferencingColumnToOneEntity.username
```

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
```diff
@@ -0,0 +1,429 @@
...
+            string createmultipleBooksMutationName = "createbooks";
+            string createMultipleBooksWithoutPublisher = @"mutation {
+                    createbooks(items: [{ title: ""My New Book"" }]) {
+                        items{
+                           id
+                           title
+                        }
+                    }
+                }";
+
```
seantleonard, line 92: need some tests that cover `createbooks(items:[{...},{...}])` where `items` has more than one item. This will theoretically further exercise the recursive validation functions.

seantleonard (commented):
Added feedback to clarify intent / comments/ add some more tests.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+        /// In case when a value for referencing column is also specified in the referencing entity, there can be two conflicting sources of truth,
+        /// which we don't want to allow. In such a case, we throw an appropriate exception.
+        /// </summary>
+        /// <param name="columnsInTargetEntity">Columns in the target (referencing) entity.</param>
+        /// <param name="derivedColumnsFromSourceEntity">Foreign key columns to be derived from the source (referenced) entity.</param>
+        /// <param name="nestingLevel">Current depth of nesting in the multiple-create GraphQL request.</param>
+        /// <param name="targetEntityName">Name of the target entity.</param>
+        /// <param name="metadataProvider">Metadata provider.</param>
+        private static void ValidateAbsenceOfReferencingColumnsInTargetEntity(
+            Dictionary<string, string> columnsInTargetEntity,
```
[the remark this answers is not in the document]
ayush3797: The `columnsInTargetEntity` contains mapping from column names in target entity to the source of the values for those columns. The source can be user provided scalar input or insertion in the source(referenced) entity. Adding more descriptive param summary. Renaming variable to `backingColumnToSourceMapInTargetEntity` to be indicative of what it stores.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+            Dictionary<string, HashSet<string>> fieldsToDeriveFromReferencedEntities = new();
+
+            // When the source of a referencing field in current entity is a relationship, the relationship's name is added to the value in
+            // the KV pair of (referencing column, source) in derivableColumnsFromRequestBody with a prefix '$' so that if the relationship name
+            // conflicts with the current entity's name or the parent entity's name, we are able to distinguish
+            // with the help of this identifier. It should be noted that the identifier is not allowed in the names
+            // of entities exposed in DAB.
+            string relationshipSourceIdentifier = "$";
+            // Loop over all the relationship fields input provided for the current entity.
+            foreach (ObjectFieldNode fieldNode in objectFieldNodes)
```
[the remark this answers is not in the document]
ayush3797: No, they cannot be renamed to `relationshipFieldNodes` because we are iterating over all the fields but processing only the `relationshipFields` in the loop. I can, at best rename it to `currentEntityInputFields`.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+            {
+                (IValueNode? fieldValue, SyntaxKind fieldKind) = GraphQLUtils.GetFieldDetails(fieldNode.Value, context.Variables);
+                if (fieldKind is SyntaxKind.NullValue || GraphQLUtils.IsScalarField(fieldKind))
+                {
+                    // If the current field is a column/scalar field or has a null value, continue.
```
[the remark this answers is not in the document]
ayush3797: A relationship field is not scalar. But this loop traverses over all the fields in the input for current entity and hence may encounter column fields or null fields, in which case we just skip the field.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+                    continue;
+                }
+
+                string relationshipName = fieldNode.Name.Value;
+                string targetEntityName = runtimeConfig.Entities![entityName].Relationships![relationshipName].TargetEntity;
+                string? linkingObject = runtimeConfig.Entities![entityName].Relationships![relationshipName].LinkingObject;
+                if (!string.IsNullOrWhiteSpace(linkingObject))
+                {
+                    // When a linking object is present, it indicates an M:N relationship between current entity and the target entity.
```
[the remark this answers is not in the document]
ayush3797: No, it's not with parent, but rather with the child entity.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+                if (referencingEntityName.Equals(entityName))
+                {
+                    for (int idx = 0; idx < fkDefinition.ReferencingColumns.Count; idx++)
+                    {
+                        string referencingColumn = fkDefinition.ReferencingColumns[idx];
+                        string referencedColumn = fkDefinition.ReferencedColumns[idx];
+
+                        // The input data for current entity should not specify a value for a referencing column -
+                        // as it's value will be derived from the insertion in the referenced (target) entity.
+                        if (derivableColumnsFromRequestBody.TryGetValue(referencingColumn, out string? referencingColumnSource))
```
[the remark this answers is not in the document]
ayush3797: `fieldsInRequestBodyForCurrentEntity` - I feel this is not indicative of the various sources which can provide a value for a record in the current entity. This variable name is more inclined towards indicating that it stores the set of fields for which user has provided scalar input value.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+                    targetEntityName: targetEntityName,
+                    referencingEntityName: referencingEntityName,
+                    referencedEntityName: referencedEntityName,
+                    foreignKeyDefinition: out ForeignKeyDefinition? fkDefinition))
+                {
+                    // This should not be hit ideally.
+                    throw new DataApiBuilderException(
+                        message: $"Could not determine definition for source: {entityName} and target: {targetEntityName} entities for " +
+                        $"relationship: {relationshipName} at level: {nestingLevel}",
+                        statusCode: HttpStatusCode.NotFound,
```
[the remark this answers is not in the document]
ayush3797: That makes sense. I'll update it to 400.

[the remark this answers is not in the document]
ayush3797: As for BCP, there is no change as to how other exceptions bubble up vs the ones in input validation.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+        /// to the set of referenced fields for all the relationships for which the current entity is the referencing entity.</param>
+        /// <param name="nestingLevel">Current depth of nesting in the multiple-create request. As we go down in the multiple mutation from the source
+        /// entity input to target entity input, the nesting level increases by 1. This helps us to throw meaningful exception messages to the user
+        /// as to at what level in the multiple mutation the exception occurred.</param>
+        /// <param name="sqlMetadataProviderFactory">Metadata provider factory.</param>
+        private static void ValidateRelationshipFields(
+            InputObjectType schemaObject,
+            string entityName,
+            IMiddlewareContext context,
+            IReadOnlyList<ObjectFieldNode> objectFieldNodes,
```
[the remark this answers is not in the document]
ayush3797: Renamed to `currentEntityInputFields`.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+                        fieldsToDeriveFromReferencedEntities.TryAdd(relationshipName, new());
+                        fieldsToDeriveFromReferencedEntities[relationshipName].Add(referencedColumn);
+
+                        // All the referencing columns in the current entity can get a value via insertion in the target entity (c).
+                        derivableColumnsFromRequestBody.TryAdd(referencingColumn, relationshipSourceIdentifier + relationshipName);
+                    }
+                }
+                else
+                {
+                    fieldsToSupplyToReferencingEntities.Add(relationshipName, new(fkDefinition.ReferencingColumns));
```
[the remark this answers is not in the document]
ayush3797: No. We want to keep track of the referencing columns in the target (referencing) entity which will get their values sourced from the insertion in the current entity.  Adding the same comment in the code.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+                        // (happens when the parent entity is a referencing entity in a relationship with current entity),
+                        // is a referencing column in the current relationship, we pass on the responsibility of getting the value
+                        // of such a column to the target entity in the current relationship.
+                        if (columnsToBeSuppliedToSourceEntity.Contains(referencingColumn))
+                        {
+                            // We optimistically assume that we will get the value of the referencing column
+                            // from the insertion in the target entity.
+                            columnsToBeSuppliedToSourceEntity.Remove(referencingColumn);
+                        }
+
```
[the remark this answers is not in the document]
ayush3797: The explanation of this case is provided here: https://github.com/Azure/data-api-builder/blob/bfe0a2e300a481c7c8f1315914b64f96a27277e7/src/Core/Services/MultipleMutationInputValidator.cs#L222

I think after reading this, the paragraph should make perfect sense.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: Your understanding is correct. Renaming variables as suggested.

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
ayush3797: I think the current message is already actionable enough where we say that we need a value for `publisher_id` to create a record in book entity. Also, there can be multiple relationships with `publisher_id` as the referencing field. So, I don't think we need to tell the user that they have the option provide input for any one of those relationships so that we can get a value for `publisher_id`. They might just want to create a record in the book entity and don't want to perform a multiple-insertion at all. So I think it would be an overkill in us telling them the anymore than what the current exception message tells.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -580,25 +586,29 @@ public class MultipleMutationInputValidator
         /// In case when a value for referencing column is also specified in the referencing entity, there can be two conflicting sources of truth,
         /// which we don't want to allow. In such a case, we throw an appropriate exception.
         /// </summary>
-        /// <param name="columnsInTargetEntity">Columns in the target (referencing) entity.</param>
+        /// <param name="backingColumnToSourceMapInTargetEntity">Dictionary storing mapping from names of backing columns in the target (referencing) entity to the source
+        /// of the corresponding column's value The source can either be:
+        /// 1. User provided scalar input for a column
+        /// 2. Insertion in the source (referenced) entity.</param>
```
seantleonard, line 592: nit: `backingColumnToSourceMapInTargetEntity` I'm not clear on this naming.. 
do you mean `columnNameToSourcedValue`? can you give possible values for a dictionary value?

seantleonard (approved):
Pending more tests and further resolution of open feedback threads + resolving merge conflicts.

severussundar (approved):
Nit suggestion to re-name `GraphQLRequestValidator` to `MultipleMutationInputValidator` in the PR description. 

LGTM, after addressing this and pending suggestions. 

Thank you for adding logic to validate the Multiple Create mutation request :)

src/Service.Tests/SqlTests/GraphQLMutationTests/MultipleMutationIntegrationTests.cs
ayush3797: modified existing tests as that covers both.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: > You may be able to further reduce parameter count (which increases readability) by introducing a MultipleMutationRequestValidationContext object which contains relevant/related properties such as

Introduced class `MultipleMutationEntityInputValidationContext`.

> this was a huge function to read through/evaluate. Was there no way to break this up + have functions with smaller method signatures?

Moved relationship processing in the method ValidateObjectFieldNodes() to a new method ProcessRelationshipField() to break up function into smaller ones.

> If this were a service instantiated at startup as non static and scoped(per request) 

For now, instantiating a multiple mutation input validator instance on a per request basis inside `SqlMutationEngine`. Not using DI right now as that might involve some more changes. Avoiding that at the moment.

src/Core/Services/MultipleMutationInputValidator.cs
ayush3797: Added comment to both the cases' source values.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -0,0 +1,608 @@
...
+            // the value for A.id, in effect, will be derived from C.id.
+
+            // Case 1: Remove from columnsToBeDerivedFromEntity, the columns which are autogenerated or
+            // have been provided a non-null value in the input for the current entity.
+
+            // As we iterate through columns which can derived from the current entity, we keep removing them from the
+            // columnsToDerive because we know the column value can be derived.
+            // At the end, if there are still columns yet to be derived, i.e.columnsToDerive.count > 0,
+            // we throw an exception.
+            foreach (string columnToBeDerivedFromEntity in columnsToBeSuppliedToSourceEntity)
```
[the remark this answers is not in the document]
ayush3797: > In the context of processing a GraphQL request, does it make more sense to be consistently saying currentEntity or targetEntity?

I think when we are in this method, everything is in the context of current entity. Parent entity is current entity's parent. Target entity is current entity's child related by some relationship.

>  This inconsistency makes this harder to read because that's even more to keep track of.
Updated/outdated.

PR #2060, opened 2024-02-26
Cosmos DB: Refactors tests GraphQL queries

sourabh1007:
## Why make this change?
Making tests to reflect true nature of functionality.

## What is this change?
1. Updated source name in db-config for all the entities. Actually, all the entities belong to the same container i.e. `planet`. This was already happening in our test cases ref https://github.com/Azure/data-api-builder/blob/main/src/Service.Tests/CosmosTests/TestBase.cs#L135
I just updated the config file, to get it highlighted.

2. Realized few tests were asserting on blank results so, added a flag to indicate if response is empty intentionally.
3. Changed graphql query in all the test cases (without changing the nature of the test) to indicate that we can always query at container level i.e. "planets". Previously, tests were querying `sun`, `earth` object directly, which is not possible actually.
4. Added logic to generate table alias in case of nested json entities

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests


src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -1071,15 +1083,15 @@ public async Task TestQueryFieldAuthConflictingWithFilterFieldAuth_Unauthorized(
 
         /// <summary>
         /// Tests that the field level query filter succeeds requests
-        /// when GraphQL is set to true without setting singular type in runtime config and
+        /// when GraphQL i  s set to true without setting singular type in runtime config and
```
seantleonard, line 1086:
```suggestion
        /// when GraphQL is set to true without setting singular type in runtime config and
```

src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -178,7 +186,7 @@
     },
     "TagAlias": {
       "source": {
-        "object": "graphqldb.tag"
+        "object": "graphqldb.planet"
```
seantleonard, line 189: I don't understand why these changes are being made. The top-level entities are Character, Star, Tag, etc. and now you're setting the backing object to be planet for all of the entities? that seems incorrect and possibly breaking for many cosmos tests.

Can you elaborate more on what you're changing here within your PR description?

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -303,6 +303,7 @@ public GQLFilterParser(RuntimeConfigProvider runtimeConfigProvider, IMetadataPro
                 counter: queryStructure.Counter);
 
         comosQueryStructure.DatabaseObject.SchemaName = queryStructure.SourceAlias;
+        comosQueryStructure.DatabaseObject.Name = entityName;
```
seantleonard, line 306: this seems incorrect.. The databaseobject name differ from the entityName used in the config.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -89,7 +89,7 @@ public async Task TestStringMultiFiltersWithAndCondition()
                         and: [
                             { additionalAttributes: {name: {eq: ""volcano1""}}}
                             { moons: {name: {eq: ""1 moon""}}}
-                            { moons: {details: {contains: ""v1""}}}
+                            { moons: {details: {contains: ""11""}}}
```
seantleonard, line 92: why this change?

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -693,7 +701,7 @@ public async Task TestGetNullStringFields()
             }";
 
             string dbQuery = "select c.name, c.age from c where IS_NULL(c.name)";
-            await ExecuteAndValidateResult(_graphQLQueryName, gqlQuery, dbQuery);
+            await ExecuteAndValidateResult(_graphQLQueryName, gqlQuery, dbQuery, true);
```
seantleonard, line 704: nit: used named param to help with readability.

`ignoreBlankResults: true`

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -861,19 +869,16 @@ public async Task TestFilterOnInnerNestedFields()
         public async Task TestFilterWithEntityNameAlias()
         {
             string gqlQuery = @"{
-                stars(first: 1, " + QueryBuilder.FILTER_FIELD_NAME + @" : {tag : {name : {eq : ""test name""}}})
+                planets(first: 1, " + QueryBuilder.FILTER_FIELD_NAME + @" : {stars : {tag: {name : {eq : ""tag1""}}}})
```
seantleonard, line 872: why does the order of the query/filtered entities need to change? please add comments in code to clarify intent.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -1101,18 +1113,20 @@ public async Task TestQueryFilterFieldAuthWithoutSingularType()
         public async Task TestQueryFilterFieldAuth_ExcludeTakesPredecence()
         {
             string gqlQuery = @"{
-                suns(first: 1, " + QueryBuilder.FILTER_FIELD_NAME + @" : {name : {eq : ""test name""}})
+                planets(first: 1, " + QueryBuilder.FILTER_FIELD_NAME + @" : {suns: { name : {eq : ""test name""}}})
```
seantleonard, line 1116: please update test summary to indicate what behavior is being validated. is authZ being performed for `planets` or `suns` or both? which fields are excluded.

seantleonard (changes requested):
Need more comments in code/clarity in PR description about why certain changes are being made. need help understanding why various entities are having their backing database objects point to the same entity 'planets' instead of their respective types.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: This query was returning blank results so changed this condition to get proper results. This test was added by me in my last PR.

src/Core/Models/GraphQLFilterParsers.cs
sourabh1007: Database Object name in config represents the container name. Once we are in nested query logic, then container will be changed to the entity.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: here order is not changed...now this graphql query represent you can query at container level only i.e. `planet`.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: Nature of the test is same. We are testing it for `sun`.... only graphql query is changed to represent that we can query `sun` through `planet` only, thats where `sun` actually resides.

src/Service.Tests/dab-config.CosmosDb_NoSql.json
sourabh1007: Just put it in the PR description. This is existing functionality, but this config file was not representing it.

config-generators/cosmosdb_nosql-commands.txt
```diff
@@ -2,13 +2,14 @@ init --config "dab-config.CosmosDb_NoSql.json" --database-type "cosmosdb_nosql"
...
-add StarAlias --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.star" --permissions "anonymous:create,read,update,delete" --graphql "Star:Stars"
-update StarAlias --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.star" --permissions "authenticated:create,read,update,delete"
-add TagAlias --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.tag" --permissions "anonymous:create,read,update,delete" --graphql "Tag:Tags"
-add Moon --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.moon" --permissions "anonymous:create,read,update,delete" --graphql true
-update Moon --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.moon" --permissions "authenticated:create,read,update,delete"
-add Earth --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.earth" --permissions "field-mutation-with-read-permission:read" --graphql "Earth:Earths"
+update Planet --config "dab-config.CosmosDb_NoSql.json" --permissions "limited-read-role:read"
+add Character --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.planet" --permissions "anonymous:create,read,update,delete" --graphql "Character:Characters"
```
Aniruddh25, line 6: as per our discussion, the entity level source.object value of the container should be taking precedence. So, we should rather fix the test at https://github.com/Azure/data-api-builder/blob/fa419cd46fe12d16633ce9bb317140c342d38183/src/Service.Tests/CosmosTests/TestBase.cs#L135 to keep separate container and check the functionality as long as there's no cross container join

src/Core/Models/GraphQLFilterParsers.cs
Aniruddh25: Should this be queryStructure.SourceAlias?

src/Service.Tests/dab-config.CosmosDb_NoSql.json
Aniruddh25: I think we should actually modify existing test, to verify source.object value i.e. container name provided at the entity level overrides the global container. And that if none is provided at entity level, the global continues to be used.

src/Core/Resolvers/CosmosQueryBuilder.cs
```diff
@@ -162,7 +162,7 @@ private static string Build(Stack<CosmosJoinStructure> joinstructure)
                     continue;
                 }
 
-                joinBuilder.Append($" JOIN {structure.TableAlias} IN {structure.DbObject.SchemaName}.{structure.DbObject.Name}");
+                joinBuilder.Append($" JOIN {structure.TableAlias} IN {structure.DbObject.FullName}");
```
Aniruddh25, line 165: Were there any tests failing afresh that we are required to fix this?

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -262,18 +262,26 @@ public async Task TestStringMultiFiltersOnArrayTypeWithOrCondition()
...
 
-        private static void ValidateResults(JsonElement actual, JsonElement expected)
+        private static void ValidateResults(JsonElement actual, JsonElement expected, bool ignoreBlankResults)
         {
             Assert.IsNotNull(expected);
             Assert.IsNotNull(actual);
+
+            if (!ignoreBlankResults)
+            {
+                // Making sure we are not asserting emply results
```
Aniruddh25, line 280:
```suggestion
                // Making sure we are not asserting empty results
```

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -262,18 +262,26 @@ public async Task TestStringMultiFiltersOnArrayTypeWithOrCondition()
...
+                Assert.IsFalse(expected.ToString().Equals("[]"), "Expected  Response is Empty.");
```
Aniruddh25, line 281: Do you mean to expected.ToString() should be string.empty? If yes, can't we use string.IsNullOrEmpty check on it and the actual string?

Aniruddh25 (changes requested):
Posting comments so far.. need to review QueryFilterTests yet,

src/Core/Resolvers/CosmosQueryBuilder.cs
sourabh1007: No, tests were not failing due to this one, I found that `FullName` already has this information which I am constructing here.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: No it means, if I am getting "empty" result from the query then, it should be considered as a valid result until unless I am mentioning it explicitly (by passing the flag) in my test.

config-generators/cosmosdb_nosql-commands.txt
sourabh1007: This can be covered as separate PR, I am noting down all these scenarios in the same sheet to include them later.

src/Service.Tests/dab-config.CosmosDb_NoSql.json
seantleonard: Per pr description:
> Updated source name in db-config for all the entities. Actually, all the entities belong to the same container i.e. planet. This was already happening in our test cases ref https://github.com/Azure/data-api-builder/blob/main/src/Service.Tests/CosmosTests/TestBase.cs#L135
I just updated the config file, to get it highlighted.

The file referenced shows that _containerName is a GUID, not `planet`:
```csharp
    internal string _containerName = Guid.NewGuid().ToString();
```

sourabh1007: Yes, I reverted it to GUID because, pipeline started failing, as these tests run parallelly, and it was trying to delete and create the same container again and again due to which tests were failing (only in pipelines) with "not found" error.
In this highlighted statement, I am saying that even override is happening in test base, still I have updated config to indicate, what is expected so that nobody should have to go through this logic, like I had to do to understand the behavior.

src/Core/Models/GraphQLFilterParsers.cs
sourabh1007: it should be unique _tablealias_. Made the code change. here I was using entity name as table alias.

config-generators/cosmosdb_nosql-commands.txt
Aniruddh25: Sure, you can open a tracking issue as well.

Aniruddh25 (approved):
LGTM, please create a tracking issue to add tests where entity specific container name is different than global container name

seantleonard:
>  Previously, tests were querying sun, earth object directly, which is not possible actually.

Were you able to identify why tests weren't failing in this case?

config-generators/cosmosdb_nosql-commands.txt
seantleonard: @sourabh1007 please open GitHub issues to track

sourabh1007: already included in my tracking sheet.

PR #2065, opened 2024-03-01
[Cherry pick] Table name collision bug fix (#1990)

### LINKED ISSUE #1908: [Bug]: DAB seems to have issues with tables in different schemas that have the same name (SQL Server)

seantleonard:
Cherry picks to `release/0.11` from `main`

## Why make this change?

Closes #1908

## What is this change?

We were previously filling the schema information for the entity data set but only using the table name. We add in the schema name to deconflict between two tables with the same name in different schemas. This means that we retrieve information from the entity data set by using schema and tablename when schema is available, and just the table name otherwise.

The reason that we do not need to worry about collisions at the schema and table name level (eg. where we have the same schema and table names within different databases) is because this entity data set is a part of the SqlMetadataProvider, and we have a SqlMetadataProvider for each database (not database type but actual DBs).

## How was this tested?

A regression test with a collision between table names in different schemas was added. We also update the unit test for SqlMetadatProvider.


Aniruddh25 (approved):
LGTM, thanks for the port

aaronburtle (approved):

PR #2069, opened 2024-03-02
Maintain Default Data Source Name from Original Config after Hot Reloading

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/1967

We maintain the same default data source name after hot reloading.

## What is this change?

When hot reloading, pass along the original default data source name, and after creating the runtimeconfig object, update that name and the dependent data structures to use the original data source name from before the hot reload.

## How was this tested?

- [ ] Unit Tests
Added validation that we maintain the same default data source name to the unit tests.

## Sample Request(s)


src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -130,7 +130,7 @@ public bool AllowIntrospection
         }
     }
 
-    private string _defaultDataSourceName;
+    public string DefaultDataSourceName;
```
Aniruddh25, line 133: Need to add jsonignore so that it doesn't write to file when serialized

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -273,6 +273,26 @@ public void UpdateDataSourceNameToDataSource(string dataSourceName, DataSource d
...
+    /// </summary>
+    /// <param name="defaultDataSourceName">The name used to update the dictionaries.</param>
+    public void UpdateDefaultDataSourceNameDependantDictionaries(string defaultDataSourceName)
+    {
+        _dataSourceNameToDataSource.Remove(DefaultDataSourceName);
+        _dataSourceNameToDataSource.Add(defaultDataSourceName, this.DataSource);
+        foreach (KeyValuePair<string, Entity> entity in Entities)
+        {
+            _entityNameToDataSourceName.Remove(entity.Key);
+            _entityNameToDataSourceName.TryAdd(entity.Key, DefaultDataSourceName);
```
Aniruddh25, line 292: You are removing the entity.key but readding it with the old value here. Did you test this?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -311,7 +331,7 @@ public bool CheckDataSourceExists(string dataSourceName)
     public string GetDefaultDataSourceName()
 #pragma warning restore CA1024 // Use properties where appropriate
     {
-        return _defaultDataSourceName;
+        return DefaultDataSourceName;
```
Aniruddh25, line 334: This function comment clearly says they preferred to use private field. instead of making it a public field, can we create a public property since you need not restrict the set?

If you do that, this function is no longer useful. We can get rid of it and the comments. 

When making changes to an existing function, always read the context surrounding it and understand how best to integrate your changes into the function such that it is still meaningful. If its no longer meaningful, feel empowered to make the necessary changes that make it meaningful.

src/Core/Configurations/ConfigFileWatcher.cs
```diff
@@ -38,7 +38,8 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
 
             if (!_configProvider.IsLateConfigured && _configProvider.GetConfig().IsDevelopmentMode())
             {
-                _configProvider.HotReloadConfig();
+                // pass along the original default data source name for consistency within runtime config's data structures
+                _configProvider.HotReloadConfig(_configProvider.GetConfig().DefaultDataSourceName);
```
Aniruddh25, line 42: Why do you need to pass along this from the ConfigFileWatcher when you call HotReloadConfig? Doesnt _configProvider already have reference to the old config? So, can't you simply do GetConfig().DefaultDataSourceName inside the configProvider.HotReloadConfig function?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -149,9 +149,9 @@ public bool TryGetLoadedConfig([NotNullWhen(true)] out RuntimeConfig? runtimeCon
     /// Hot Reloads the runtime config when the file watcher
     /// is active and detects a change to the underlying config file.
     /// </summary>
-    public void HotReloadConfig()
+    public void HotReloadConfig(string dataSourceName)
     {
-        ConfigLoader.TryLoadKnownConfig(out _runtimeConfig, replaceEnvVar: true);
+        ConfigLoader.TryLoadKnownConfig(out _runtimeConfig, replaceEnvVar: true, dataSourceName: dataSourceName);
```
Aniruddh25, line 154: cant you do GetConfig().DefaultDataSourceName here?

Aniruddh25 (changes requested):
Needs better integration into existing code.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: i did but I think I got the names mixed up, I renamed the argument to try and be more clear about which data source names are which.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: because we need to update the DefaultDataSourceName from outside of the Runtimeconfig, I think it must be made public. I will remove this function.

src/Core/Configurations/ConfigFileWatcher.cs
aaronburtle: this is the case, so I will just pass along the DefaultDataSourceName from the reference we already have available in the config provider.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -98,13 +98,26 @@ public string GetConfigFileName()
...
+                json: json,
+                config: out config,
+                connectionString: _connectionString,
+                replaceEnvVar: replaceEnvVar);
+
+            if (isParsed && !string.IsNullOrEmpty(dataSourceName))
+            {
+                config!.UpdateDefaultDataSourceNameDependantDictionaries(dataSourceName);
+                config!.DefaultDataSourceName = dataSourceName;
+            }
```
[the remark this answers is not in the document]
aaronburtle: Yes, it requires the TryParse to be in the condition to stop complaining. Good catch.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -273,6 +273,26 @@ public void UpdateDataSourceNameToDataSource(string dataSourceName, DataSource d
...
 
+    /// <summary>
+    /// In a Hot Reload scenario we should maintain the same default data source
+    /// name before the hot reload as after the hot reload. This method takes
+    /// a default data source name, such as the one from before hot reload, and
+    /// replaces the current dictionary entries of this RuntimeConfig that were
+    /// built using a new, unique guid during the construction of this RuntimeConfig
+    /// with entries using the provided default data source name.
+    /// </summary>
+    /// <param name="defaultDataSourceName">The name used to update the dictionaries.</param>
```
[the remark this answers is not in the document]
aaronburtle: this was just a mix up from bad aming, I renamed the argument to try and mae it clearer. We want to update using the initial data source name, and remove the stale entries using the (stale, but current until hot reloading is finished) current default data source name.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -120,6 +114,7 @@ public void HotReloadConfigRestRuntimeOptions()
             RuntimeConfigProvider configProvider = new(configLoader);
             // Must GetConfig() to start file watching
             RuntimeConfig runtimeConfig = configProvider.GetConfig();
+            string initialDefaultDataSourceName = runtimeConfig.DefaultDataSourceName;
```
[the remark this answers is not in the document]
aaronburtle: `GetConfig()` will never return a null config. The asserts are simply to validate that we have certain values before we do the hot reload. The test flow is then.

// Arrange
1.Setup the strings that will be turned into our initial runtimeconfig and our reloaded runtimeconfig.
2.Generate the initial runtimeconfig, start file watching, and assert we have the initial value.


// Act
1. Hot reload the runtime config.

// Assert
1. Assert we have the correct values after a hot reload.

I've added some spacing to try and make it more easily readable, please let me know if there are any other reorgs of the code you think would make it better.

seantleonard (commented):
Some questions about approach in latest changeset

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -275,21 +276,24 @@ public void UpdateDataSourceNameToDataSource(string dataSourceName, DataSource d
...
+    public void UpdateDefaultDataSourceNameDependantDictionaries(string initialDefaultDataSourceName)
     {
         _dataSourceNameToDataSource.Remove(DefaultDataSourceName);
-        _dataSourceNameToDataSource.Add(defaultDataSourceName, this.DataSource);
+        _dataSourceNameToDataSource.Add(initialDefaultDataSourceName, this.DataSource);
         foreach (KeyValuePair<string, Entity> entity in Entities)
         {
             _entityNameToDataSourceName.Remove(entity.Key);
-            _entityNameToDataSourceName.TryAdd(entity.Key, DefaultDataSourceName);
+            _entityNameToDataSourceName.Add(entity.Key,initialDefaultDataSourceName);
```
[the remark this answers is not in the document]
aaronburtle: can just update as well.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -275,21 +276,24 @@ public void UpdateDataSourceNameToDataSource(string dataSourceName, DataSource d
...
+    /// for lookups. To correctly retrieve this information after a hot reload
+    /// we need the data source name to say the same after hot reloading. This method takes
     /// a default data source name, such as the one from before hot reload, and
     /// replaces the current dictionary entries of this RuntimeConfig that were
     /// built using a new, unique guid during the construction of this RuntimeConfig
     /// with entries using the provided default data source name.
     /// </summary>
     /// <param name="defaultDataSourceName">The name used to update the dictionaries.</param>
-    public void UpdateDefaultDataSourceNameDependantDictionaries(string defaultDataSourceName)
+    public void UpdateDefaultDataSourceNameDependantDictionaries(string initialDefaultDataSourceName)
```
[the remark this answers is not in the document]
aaronburtle: makes sense, I added the updating of the DefaultDataSourceName to this method, renamed it, and made that a private property again.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -269,8 +269,8 @@ public void HotReloadConfig()
                 DatabaseType.CosmosDB_NoSQL => HandleCosmosNoSqlConfiguration(graphQLSchema, runtimeConfig, connectionString),
                 _ => runtimeConfig with { DataSource = runtimeConfig.DataSource with { ConnectionString = connectionString } }
             };
-            ManagedIdentityAccessToken[_runtimeConfig.DefaultDataSourceName] = accessToken;
-            _runtimeConfig.UpdateDataSourceNameToDataSource(_runtimeConfig.DefaultDataSourceName, _runtimeConfig.DataSource);
+            ManagedIdentityAccessToken[_runtimeConfig.GetDefaultDataSourceName()] = accessToken;
+            _runtimeConfig.UpdateDataSourceNameToDataSource(_runtimeConfig.GetDefaultDataSourceName(), _runtimeConfig.DataSource);
```
seantleonard, line 273: Doesn't `UpdateDefaultDataSourceName(...)` handle updating the runtimeconfig object's relevant internal data structures, rendering   `_runtimeConfig.UpdateDataSourceNameToDataSource(...)` redundant?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -130,8 +130,7 @@ public bool AllowIntrospection
         }
     }
 
-    [JsonIgnore]
```
seantleonard: Don't you still need this JsonIgnore? what does the output config look like?

seantleonard (changes requested):

src/Config/ObjectModel/RuntimeConfig.cs
seantleonard: This field now shows up in output config files which isn't intended: 
image

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -130,8 +130,7 @@ public bool AllowIntrospection
...
-    public string DefaultDataSourceName;
+    public string _defaultDataSourceName;
```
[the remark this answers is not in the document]
aaronburtle: I am changing this member to private so that we only update through the public method. should be pushed now.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -327,6 +327,17 @@ public bool CheckDataSourceExists(string dataSourceName)
         return _dataSourceNameToDataSource.ContainsKey(dataSourceName);
     }
 
+    /// <summary>
+    /// Get the default datasource name.
+    /// </summary>
+    /// <returns>default datasourceName.</returns>
+#pragma warning disable CA1024 // Use properties where appropriate. Reason: Do not want datasource serialized and want to keep it private to restrict set;
+    public string GetDefaultDataSourceName()
+#pragma warning restore CA1024 // Use properties where appropriate
```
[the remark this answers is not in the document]
aaronburtle: changing that field to private so we can just internally set it, and then still use the getter here.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: changed to private so this should be OK.

src/Core/Configurations/RuntimeConfigProvider.cs
aaronburtle: This is used slightly differently, where a datasource name and a data source are provided and then this function will update that data source name to match the provided data source.

this function is only called from the initialization function that is used in a hosted scenario, and so will not be effected by the changes that I made, which will only happen when we are loading from an actual file.

src/Config/ObjectModel/RuntimeConfig.cs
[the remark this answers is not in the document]
seantleonard: Can you utilize this syntax so that we don't have to ignore StyleCop warning? https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties . To address your comment "Do not want datasource serialized" wouldn't adding `[JsonIgnore]` mitigate?

Do any of the exceptions in the C1024 article apply here? https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1024
if so, can the exception text be added to your comment?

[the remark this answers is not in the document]
aaronburtle: Yes, I can refactor this, but just to be clear, this was already existing code I originally removed because I was going to make that field public, then reverted.

seantleonard (changes requested):
pending fixes to unit tests

src/Core/Configurations/RuntimeConfigProvider.cs
seantleonard: thanks for elaborating, understood now.

src/Config/ObjectModel/RuntimeConfig.cs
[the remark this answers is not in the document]
seantleonard: thanks for cleaning this up. The getter and setter are applicable to what you're working on so it's helpful to have made this change.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -123,19 +128,27 @@ public void HotReloadConfigRestRuntimeOptions()
...
             System.Threading.Thread.Sleep(1000);
-            // Hot Reloaded config
+
+            // Act
+            // 1. Hot reload the runtime config
             runtimeConfig = configProvider.GetConfig();
             string updatedDefaultDataSourceName = runtimeConfig.DefaultDataSourceName;
+
+            // Assert
+            // 1. Assert we have the correcr values after a hot reload.
```
Aniruddh25, line 144:
```suggestion
            // 1. Assert we have the correct values after a hot reload.
```

Aniruddh25 (approved):
LGTM, thanks for addressing all comments

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -275,22 +276,26 @@ public void UpdateDataSourceNameToDataSource(string dataSourceName, DataSource d
...
+    /// with entries using the provided default data source name. We then update the DefaultDataSourceName.
     /// </summary>
-    /// <param name="defaultDataSourceName">The name used to update the dictionaries.</param>
-    public void UpdateDefaultDataSourceNameDependantDictionaries(string defaultDataSourceName)
+    /// <param name="initialDefaultDataSourceName">The name used to update the dictionaries.</param>
+    public void UpdateDefaultDataSourceName(string initialDefaultDataSourceName)
     {
         _dataSourceNameToDataSource.Remove(DefaultDataSourceName);
-        _dataSourceNameToDataSource.Add(defaultDataSourceName, this.DataSource);
+        _dataSourceNameToDataSource.Add(initialDefaultDataSourceName, this.DataSource);
```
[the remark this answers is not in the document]
aaronburtle: Added just for extra safety.

seantleonard (approved):

PR #2072, opened 2024-03-02
Condition OpenAPI document creation to non-CosmosDB database types.

### LINKED ISSUE #2045: [Bug]: Console logged "error" when DBtype is Cosmos and OpenAPI service fails to generate doc
### What happened?

A bug happened!

Cosmos doesn't support openapi/rest so there should be no reference to openapi in dab logs when using cosmos.

### Version

`main`

CosmosDB NoSQL

Local (including CLI)

GraphQL

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
## Why make this change?

- Closes #2045. Any time that CosmosDB is the configured database, the codepath that attempts to init the OpenAPI documentor service is run and will always fail for CosmosDB. As a result, a "Fail" message always appears on the console which can be misinterpreted as a user fixable error and is a false positive notification of an issue.

## What is this change?

- OpenAPI documentor service init only occurs for non-CosmosDB database types because db types like MSSQL , PG, MySQL are the only db types that support the REST endpoint in DAB.

## How was this tested?

- [x] Run DAB with CosmosDB set as db type and ensure OpenAPI init failure is not logged to console.
image


src/Service/Startup.cs
```diff
@@ -624,18 +624,23 @@ private async Task<bool> PerformOnConfigChangeAsync(IApplicationBuilder app)
...
+                    // because OpenAPI document creation is not required for the engine to operate.
+                    // Errors will be logged.
+                    try
+                    {
+                        IOpenApiDocumentor openApiDocumentor = app.ApplicationServices.GetRequiredService<IOpenApiDocumentor>();
+                        openApiDocumentor.CreateDocument();
+                    }
+                    catch (DataApiBuilderException dabException)
+                    {
+                        _logger.LogError(exception: dabException, message: "OpenAPI Documentor initialization failed.");
```
Aniruddh25, line 642: Can we change this to LogWarning, so that even for SQL it doesnt show up as `Fail`? Otherwise the successfully completed message below doesn't do justice.

seantleonard: sure

seantleonard: Made it a warning and updating message to: `"OpenAPI Documentor initialization failed. This will not affect DAB engine startup.");`

severussundar (approved):
LGTM!

src/Service/Startup.cs
```diff
@@ -624,18 +624,23 @@ private async Task<bool> PerformOnConfigChangeAsync(IApplicationBuilder app)
...
+                    // because OpenAPI document creation is not required for the engine to operate.
+                    // Errors will be logged.
+                    try
+                    {
+                        IOpenApiDocumentor openApiDocumentor = app.ApplicationServices.GetRequiredService<IOpenApiDocumentor>();
+                        openApiDocumentor.CreateDocument();
+                    }
+                    catch (DataApiBuilderException dabException)
+                    {
+                        _logger.LogWarning(exception: dabException, message: "OpenAPI Documentor initialization failed. This will not affect DAB engine startup.");
```
Aniruddh25, line 642:
```suggestion
                        _logger.LogWarning(exception: dabException, message: "OpenAPI Documentor initialization failed. This will not affect dab startup.");
```

since in CLI, we say `dab start`

Aniruddh25 (approved):

PR #2073, opened 2024-03-04
Cosmos DB: Upgrade cosmos DB SDK to latest version i.e. 3.38.1

sourabh1007:
## Why make this change?

We want to use SDK latest features like:
1. Patch API
2. Ability to disable certificate check by including a flag in connection string.
3. Hierarchal Partition Key

etc.

## What is this change?
Upgrade Cosmos to the latest version`3.38.1`. It's a non-breaking change

## How was this tested?

NA

## Sample Request(s)

NA

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -22,7 +22,7 @@
     <PackageVersion Include="Microsoft.AspNetCore.Http" Version="2.2.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.14" />
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="6.0.14" />
-    <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.20.0" />
+    <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.38.1" />
     <!--When updating Microsoft.Data.SqlClient, update license URL in scripts/notice-generation.ps1-->
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.1.4" />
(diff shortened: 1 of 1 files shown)


seantleonard:
There may not be tests for this.. how do we know DAB isn't affected by at least one of the breaking changes for CosmosSDK listed in : https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/changelog.md since we are upgrading from `3.2.0 2021-06-21` (released in 2021)

> ⚠️ Note: Starting with version 3.38.0, the .NET SDK enables the ODE feature by default. This can potentially cause a new type of continuation token to be generated. Such a token is not recognized by the older SDKs by design and this could result in a Malformed Continuation Token Exception. If you have a scenario where tokens generated from the newer SDKs are used by an older SDK, we recommend a 2 step approach to upgrade:
>
> Upgrade to the new SDK and disable ODE, both together as part of a single deployment. Wait for all nodes to upgrade.
In order to disable ODE, set EnableOptimisticDirectExecution to false in the QueryRequestOptions.
Enable ODE as part of second deployment for all nodes.

Aniruddh25 (approved):
LGTM

Aniruddh25:
> If you have a scenario where tokens generated from the newer SDKs are used by an older SDK, we recommend a 2 step approach to upgrade:

Good question! but I don't think we have a scenario where tokens generated by newer DAB(that uses the new SDK) will be used by  older DAB (that uses the old SDK). Because when client apps target a newer DAB, older DAB shouldn't be in picture.

seantleonard (approved):
thanks for updating the sdk!

Aniruddh25:
FYI @sourabh1007 merging this, if you have concerns regarding breaking changes please address in a separate PR

PR #2084, opened 2024-03-08
Update DAB CLI to return appropriate exit codes based upon exit reason

### LINKED ISSUE #1777: [Bug]: Dab start command exited with code 0
### What happened?

When trying to run the dab tool programatically, the tool returns status 0 when an error occurred. 
Run the "dab start" tool from location "C:\Path\with\config" and then run it again programatically.
See the following program repro (wait some time to give the dab tool to exit):
```csharp
static void Main(string[] args)
        {
            ProcessStartInfo processStartInfo = new("dab", "start")
            {
                UseShellExecute = true,
                WorkingDirectory = @"C:\Path\with\config",
                WindowStyle = ProcessWindowStyle.Minimized
            };
            var process = new Process
            {
                StartInfo = processStartInfo,
                EnableRaisingEvents = true
            };

            process.Exited += (sender, args) =>
            {
                Process process = (Process)sender;
                if (process != null)
                {
                    Console.WriteLine("Exited: ExitCode " + process.ExitCode);
                }
            };

            Console.WriteLine("Started");
            process.Start();

            if (process.HasExited)
            {
                Console.WriteLine("ExitCode " + process.ExitCode);
            }
            else
            {
                Console.WriteLine("Did not exit");
            }

            Console.ReadKey();
        }
```

which leads to the following output (after some time):
```
Started
Did not exit
Exited: ExitCode 0
```

According to the good practices, the “return 0” implies that the code has executed successfully without any error.
Here, the program exited due to some error (we wanted to run the host twice from the same location).
**Expected**: error code different than 0.

### Version

0.8.52

Azure SQL

REST

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
## Why make this change?

- Closes #1777
  - Certain failures that occurred within CLI execution resulted in the CLI returning exit code 0 for success, even though there were failures.
     - For example, `dab start {options}` may contain fully valid values, but when used as arguments to start the DAB engine process within CLI, engine startup fails. The failure didn't get reflected as error code -1.
     - For example, `dab init {options}` may reference a file that already exists in the path that the CLI uses to read/write config files, this results in an error that is not properly reflected in the exit code.

## What is this change?

- Any failures that occur in CLI usage will result in the proper exit codes:
    - `0` success
    - `-1` failure
- The change updates the usage of CommandLineParser. Instead of swallowing exceptions and the status of `isSuccess` within the `options.Handler()` methods, the handler methods all return the int error code now.
- e.g.  the following now captures the return codes after command line parsing.
```csharp
parser.ParseArguments<InitOptions, AddOptions, UpdateOptions, StartOptions, ValidateOptions, ExportOptions, AddTelemetryOptions>(args)
                .MapResult(
```

## How was this tested?

- [x] Integration Tests


src/Cli/Exporter.cs
```diff
@@ -23,7 +23,7 @@ public static void Export(ExportOptions options, ILogger logger, FileSystemRunti
             if (!TryGetConfigFileBasedOnCliPrecedence(loader, options.Config, out string runtimeConfigFile))
             {
                 logger.LogError("Failed to find the config file provided, check your options and try again.");
-                return;
+                return -1;
```
Aniruddh25, line 26: Can we define the value -1 as a const error code somewhere?

src/Cli/Exporter.cs
```diff
@@ -65,6 +67,7 @@ public static void Export(ExportOptions options, ILogger logger, FileSystemRunti
             }
 
             cancellationTokenSource.Cancel();
+            return isSuccess ? 0 : -1;
```
Aniruddh25, line 70: Similar to -1, can we define 0 as a const success code somewhere?

src/Cli/Commands/InitOptions.cs
```diff
@@ -120,18 +120,22 @@ public class InitOptions : Options
         [Option("rest.request-body-strict", Required = false, HelpText = "(Default: true) Allow extraneous fields in the request body for REST.")]
         public CliBool RestRequestBodyStrict { get; }
 
-        public void Handler(ILogger logger, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
+        // public int returnCode;
```
Aniruddh25, line 123: nit: remove extra code

Aniruddh25 (changes requested):
Const codes to indicate success or error

src/Cli/Exporter.cs
seantleonard: Yes, introduced constants folder with CliReturnCode class with two constants CliReturnCode.GENERAL_ERROR and CliReturnCode.SUCCESS

src/Cli/Exporter.cs
seantleonard: Yes, introduced constants folder with CliReturnCode class with two constants CliReturnCode.GENERAL_ERROR and CliReturnCode.SUCCESS

src/Cli/Commands/InitOptions.cs
seantleonard: Done.

Aniruddh25 (approved):
LGTM.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -607,20 +609,48 @@ public void TestEngineStartUpWithVerboseAndLogLevelOptions(string logLevelOption
...
+    /// Validates that invalid verbs and options produce exit code -1 (CliReturnCode.GENERAL_ERROR).
+    /// </summary>
+    /// <param name="cliArguments">cli verbs, options, and option values</param>
+    [DataTestMethod]
+    [DataRow(new string[] { "--remove-telemetry" }, DisplayName = "Usage of non-existent verb remove-telemetry")]
+    [DataRow(new string[] { "--initialize" }, DisplayName = "Usage of invalid verb (longform of init not supported) initialize")]
+    [DataRow(new string[] { "init", "--database-name", "mssql" }, DisplayName = "Invalid init options database-name")]
+    public void InvalidVerbsAndOptionsReturnNonZeroExitCode(string[] cliArguments)
+    {
+        Assert.AreEqual(expected: CliReturnCode.GENERAL_ERROR, Program.Execute(cliArguments, _cliLogger!, _fileSystem!, _runtimeConfigLoader!));
```
aaronburtle, line 638: nit: you labeled `expected` here, may want to label `actual` as well in this case for symmetry.

aaronburtle: same in others if you decide to add

src/Cli/DabCliParserErrorHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <summary>
+    /// Processes errors that occur during parsing of CLI verbs (start, init, export, add, update, etc) and their arguments.
+    /// </summary>
+    public class DabCliParserErrorHandler
+    {
+        /// <summary>
+        /// Processes errors accumulated by each parser in parser.ParseArguments<parsers>().
+        /// For DAB CLI, this only includes scenarios where the user provides invalid DAB CLI input.
+        /// e.g. incorrectly formed or missing options and parameters.
+        /// Additionally, an error is tracked if the user uses an unsupported CLI verb or uses --help.
```
aaronburtle, line 17:
```suggestion
        /// Additionally, an error is tracked if the user uses an unsupported CLI verb or uses --help or --version.
```

Just out of curiosity, why are we tracking errors for invoking help or version?

aaronburtle (approved):
Looks good, just a nit.

Issue #1456, opened 2023-04-26
Release branch's commit ID is included in the version Assembly metadata


severussundar:
Enabling deterministic builds appends the release branch's git commit ID to the assembly metadata. 

So running `dab --version` outputs the below

![image](https://user-images.githubusercontent.com/11196553/234464975-b8191a84-d936-4b2e-a781-ceaf7fbe638c.png)

Displaying the git commit ID is unnecessary from the user's perspective. 

Investigate and come up with a fix to output just the <major.minor.patch> (without including the git commit id hash)

Some options to consider:

1. Possibility of just printing the <major.minor.patch> yet have the git commit id in the assembly metadata.
2. Possiblity of including the commit id in the assembly metadata in field other than the version number

seantleonard:
This is due to commandlineparser's AutoHelper. The changes needed can be inferred from the suggestions here: https://github.com/commandlineparser/commandline/issues/757#issuecomment-1005505288

Docs with examples where we would implement our own method which printed out the help text. I didn't want to bloat #2086 even further. This may not be that big of a change, but requires its own troubleshooting.
https://github.com/commandlineparser/commandline/wiki/HelpText-Configuration

PR #2086, opened 2024-03-08
Improved Health Endpoint Metadata (DAB Version/App User Agent)

### LINKED ISSUE #1971: Return deployed version at root path
Return the current deployed version on the “Is Healthy?” root path (“/”) - today we just say – “Healthy”.

seantleonard:
## Why make this change?

- Closes #1971, where the health point only returned a plain text `Healthy`. There was no additional information that would be useful such as DAB version. 

## What is this change?

- Updates the ProductInfo functions to return either a version number with `major.minor.patch` or `major.minor.patch+<commit_hash>`

```json
{
    "status": "Healthy",
    "version": "0.12.0",
    "appName": "dab_oss_0.12.0"
}
```

- Added the classes
    1. `DabHealthCheck : IHealthCheck` 
    2. `HealthReportResponseWriter` -> uses Health Check metadata to fill health check endpoint response contents (JSON String). Because the writer writes raw json and uses values that stay constant through the lifetime of the dab process, the created json is cached after first creation and returned whenever the health endpoint is hit.

- Updates `Directory.Build.props` with `<Version>0.12</Version>` to avoid having to hardcode version numbers in all .csproj files like in #1995 

## How was this tested?

- [X] Integration Tests
    - CLI tests checking version number
    - [IN PROGRESS], actual test for health check classes
    - [IN PROGRESS], validate Build pipeline and modify directory build props file with `*.*.*` wildcard so that pipeline variable takes precedence: see discussion: https://github.com/Azure/data-api-builder/pull/1995#issuecomment-1920491707

## Sample Request(s)

```https
GET https://[localhost:5001
```

### DIFF
--- a/src/Cli.Tests/EndToEndTests.cs
+++ b/src/Cli.Tests/EndToEndTests.cs
@@ -120,7 +120,7 @@ public class EndToEndTests
             replaceEnvVar: true));
 
         SqlConnectionStringBuilder builder = new(runtimeConfig.DataSource.ConnectionString);
-        Assert.AreEqual(ProductInfo.GetDataApiBuilderApplicationName(), builder.ApplicationName);
+        Assert.AreEqual(ProductInfo.GetDataApiBuilderUserAgent(includeCommitHash: false), builder.ApplicationName);
 
         Assert.IsNotNull(runtimeConfig);
         Assert.AreEqual(DatabaseType.MSSQL, runtimeConfig.DataSource.DatabaseType);
@@ -657,11 +657,11 @@ public class EndToEndTests
 
     /// <summary>
     /// Test to verify that help writer window generates output on the console.
+    /// For every test, validates that the first line of the output contains the product name and version.
     /// </summary>
     [DataTestMethod]
     [DataRow("", "", new string[] { "ERROR" }, DisplayName = "No flags provided.")]
     [DataRow("initialize", "", new string[] { "ERROR", "Verb 'initialize' is not recognized." }, DisplayName = "Wrong Command provided.")]
-    [DataRow("", "--version", new string[] { "Microsoft.DataApiBuilder 1.0.0" }, DisplayName = "Checking version.")]
     [DataRow("", "--help", new string[] { "init", "add", "update", "start" }, DisplayName = "Checking output for --help.")]
     public void TestHelpWriterOutput(string command, string flags, string[] expectedOutputArray)
     {
@@ -682,24 +682,25 @@ public class EndToEndTests
         process.Kill();
     }
 
-    [DataRow("", "--version", DisplayName = "Checking dab version with --version.")]
-    [DataTestMethod]
-    public void TestVersionHasBuildHash(
-        string command,
-        string options
-    )
+    /// <summary>
+    /// When CLI is started via: dab --version, it should print the version number
+    /// which includes the commit hash. For example:
+    /// Microsoft.DataApiBuilder 0.12+2d181463e5dd46cf77fea31b7295c4e02e8ef031
+    /// </summary>
+    [TestMethod]
+    public void TestVersionHasBuildHash()
     {
         _fileSystem!.File.WriteAllText(TEST_RUNTIME_CONFIG_FILE, INITIAL_CONFIG);
 
         using Process process = ExecuteDabCommand(
-            command: $"{command} ",
-            flags: $"--config {TEST_RUNTIME_CONFIG_FILE} {options}"
+            command: string.Empty,
+            flags: $"--config {TEST_RUNTIME_CONFIG_FILE} --version"
         );
 
         string? output = process.StandardOutput.ReadLine();
         Assert.IsNotNull(output);
 
-        // Check that build hash is returned as part of  version number
+        // Check that the build hash is returned as part of the version number.
         string[] versionParts = output.Split('+');
         Assert.AreEqual(2, versionParts.Length, "Build hash not returned as part of version number.");
         Assert.AreEqual(40, versionParts[1].Length, "Build hash is not of expected length.");
--- a/src/Product/ProductInfo.cs
+++ b/src/Product/ProductInfo.cs
@@ -8,22 +8,44 @@ namespace Azure.DataApiBuilder.Product;
 
 public static class ProductInfo
 {
-    public const string DEFAULT_VERSION = "0.0.0";
     public const string DAB_APP_NAME_ENV = "DAB_APP_NAME_ENV";
-    public static readonly string DEFAULT_APP_NAME = $"dab_oss_{ProductInfo.GetProductVersion()}";
-    public static readonly string ROLE_NAME = "DataApiBuilder";
+    public static readonly string DAB_USER_AGENT = $"dab_oss_{GetMajorMinorPatchVersion()}";
+    public static readonly string DEFAULT_APP_NAME = $"dab_oss_{GetProductVersion()}";
+    public static readonly string CLOUD_ROLE_NAME = "DataApiBuilder";
 
     /// <summary>
     /// Reads the product version from the executing assembly's file version information.
+    /// Includes commit hash.
     /// </summary>
+    /// <param name="includeCommitHash">True by default: the user agent string will include the commit hash.</param>
     /// <returns>Product version if not null, default version 0.0.0 otherwise.</returns>
-    public static string GetProductVersion()
+    public static string GetProductVersion(bool includeCommitHash = true)
     {
+        if (!includeCommitHash)
+        {
+            return GetMajorMinorPatchVersion();
+        }
+ 
         Assembly assembly = Assembly.GetExecutingAssembly();
         FileVersionInfo fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
         string? version = fileVersionInfo.ProductVersion;
 
-        return version ?? DEFAULT_VERSION;
+        return version ?? "DAB UNVERSIONED OSS";
+    }
+
+    /// <summary>
+    /// Returns the Product version in Major.Minor.Patch format without a commit hash.
+    /// FileVersionInfo.ProductBuildPart is used to represent the Patch version.
+    /// FileVersionInfo is used to retrieve the version information from the executing assembly
+    /// set by the Version property in Directory.Build.props
+    /// </summary>
+    /// <returns>Version string "Major.Minor.Patch"</returns>
+    public static string GetMajorMinorPatchVersion()
+    {
+        Assembly assembly = Assembly.GetExecutingAssembly();
+        FileVersionInfo fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
+        string version = fileVersionInfo.ProductMajorPart + "." + fileVersionInfo.ProductMinorPart + "." + fileVersionInfo.ProductBuildPart;
+        return version;
     }
 
     /// <summary>
@@ -31,21 +53,16 @@ public static class ProductInfo
     /// DAB_APP_NAME_ENV environment variable. If the environment variable is not set,
     /// it returns a default value indicating connections from open source.
     /// </summary>
-    public static string GetDataApiBuilderUserAgent()
+    /// <param name="includeCommitHash">True by default: the user agent string will include the commit hash.</param>
+    /// <returns></returns>
+    public static string GetDataApiBuilderUserAgent(bool includeCommitHash = true)
     {
-        return Environment.GetEnvironmentVariable(DAB_APP_NAME_ENV) ?? DEFAULT_APP_NAME;
-    }
+        if (includeCommitHash)
+        {
+            return Environment.GetEnvironmentVariable(DAB_APP_NAME_ENV) ?? DEFAULT_APP_NAME;
+        }
 
-    /// <summary>
-    /// Returns the application name to be used for database connections for the DataApiBuilder.
-    /// It strips the hash value from the user agent string to only return the application name and the version.
-    /// The method serves as a means of identifying the source of connections made through the DataApiBuilder.
-    /// </summary>
-    public static string GetDataApiBuilderApplicationName()
-    {
-        string dabVersion = ProductInfo.GetDataApiBuilderUserAgent();
-        int hashStartPosition = dabVersion.LastIndexOf('+');
-        return hashStartPosition != -1 ? dabVersion[..hashStartPosition] : dabVersion;
+        return Environment.GetEnvironmentVariable(DAB_APP_NAME_ENV) ?? DAB_USER_AGENT;
     }
 }
 
--- a/src/Service/HealthCheck/HealthReportResponseWriter.cs
+++ b/src/Service/HealthCheck/HealthReportResponseWriter.cs
@@ -1,0 +1,84 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using System.IO;
+using System.Text.Json;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.DataApiBuilder.Service.HealthCheck
+{
+    /// <summary>
+    /// Creates a JSON response for the health check endpoint using the provided health report.
+    /// If the response has already been created, it will be reused.
+    /// </summary>
+    public class HealthReportResponseWriter
+    {
+        private byte[]? _responseBytes;
+        private const string JSON_CONTENT_TYPE = "application/json; charset=utf-8";
+        private const string DAB_VERSION_KEY = "version";
+        private const string DAB_APPNAME_KEY = "appName";
+
+        /// <summary>
+        /// Function provided to the health check middleware to write the response.
+        /// </summary>
+        /// <param name="context">HttpContext for writing the response.</param>
+        /// <param name="healthReport">Result of health check(s).</param>
+        /// <returns>Writes the http response to the http context.</returns>
+        public Task WriteResponse(HttpContext context, HealthReport healthReport)
+        {
+
+            context.Response.ContentType = JSON_CONTENT_TYPE;
+
+            if (_responseBytes is null)
+            {
+                _responseBytes = CreateResponse(healthReport);
+            }
+
+            return context.Response.WriteAsync(Encoding.UTF8.GetString(_responseBytes));
+        }
+
+        /// <summary>
+        /// Using the provided health report, creates the JSON byte array to be returned and cached.
+        /// Currently, checks for the custom DabHealthCheck result and adds the version and app name to the response.
+        /// The result of the response returned for the health endpoint would be:
+        /// {
+        ///     "status": "Healthy",
+        ///     "version": "Major.Minor.Patch",
+        ///     "appName": "dab_oss_Major.Minor.Patch"
+        /// }
+        /// </summary>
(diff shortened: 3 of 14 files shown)


.pipelines/templates/build-pipelines.yml
```diff
@@ -14,6 +14,22 @@ steps:
 # generate the prerelease nuget version.
 # We cannot set this in variables section above because $(isNugetRelease) 
 # is not available at pipeline compilation time.
+- task: PowerShell@2
+  inputs:
+    targetType: 'inline'
+    script: |
+      [xml]$directoryBuildProps = Get-Content -Path $(Build.SourcesDirectory)/src/Directory.Build.props
+      # Get Version from Directory.Build.props
+      # When you access this XML property, it returns an array of elements (even if there�s only one element with that name).
```
Aniruddh25, line 23:
```suggestion
      # When you access this XML property, it returns an array of elements (even if there's only one element with that name).
```

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -682,24 +679,25 @@ public void TestHelpWriterOutput(string command, string flags, string[] expected
...
 
-    [DataRow("", "--version", DisplayName = "Checking dab version with --version.")]
-    [DataTestMethod]
-    public void TestVersionHasBuildHash(
-        string command,
-        string options
-    )
+    /// <summary>
+    /// When CLI is started via: dab --version, it should print the version number
+    /// which includes the commit hash. For example:
```
Aniruddh25, line 684: We have an issue that contradicts this test: according to the issue, the commit hash should NOT be included in the version: https://github.com/Azure/data-api-builder/issues/1456 that is displayed via CLI. 

Can we get clarity on this? to remove the hash altogether or keep it in CLI but not in appname?

src/Product/ProductInfo.cs
```diff
@@ -8,44 +8,61 @@ namespace Azure.DataApiBuilder.Product;
...
+    public static readonly string CLOUD_ROLE_NAME = "DataApiBuilder";
 
     /// <summary>
     /// Reads the product version from the executing assembly's file version information.
+    /// Includes commit hash.
     /// </summary>
+    /// <param name="includeCommitHash">True by default: the user agent string will include the commit hash.</param>
     /// <returns>Product version if not null, default version 0.0.0 otherwise.</returns>
-    public static string GetProductVersion()
+    public static string GetProductVersion(bool includeCommitHash = true)
```
Aniruddh25, line 22: Why do we need separate logic for GetProductVersion() v/s GetMajorMinorPatchVersion()? 

I dont think we ever desire the commit hash to be returned anywhere. 
We dont need includeCommitHash argument.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -514,37 +514,80 @@ public void CleanupAfterEachTest()
...
+                out RuntimeConfig updatedRuntimeConfig,
+                replaceEnvVar: true));
+
+            string actualUpdatedConnectionString = updatedRuntimeConfig.DataSource.ConnectionString;
+
+            Assert.AreEqual(actualUpdatedConnectionString, expectedUpdatedConnectionString);
+        }
+
+        /// <summary>
+        /// Checks if the connection string provided in the config is remains unchanged for MySQL, PostgreSQL, CosmosDB_PostgreSQl, CosmosDB_NoSQL.
```
Aniruddh25, line 566:
```suggestion
        /// Checks if the connection string provided in the config remains unchanged for MySQL, PostgreSQL, CosmosDB_PostgreSQl, CosmosDB_NoSQL.
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2687,6 +2730,65 @@ public async Task OpenApi_GlobalEntityRestPath(bool globalRestEnabled, bool expe
...
+
+            // Validate response body
+            Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+
+            Assert.IsTrue(
+                condition: responseProperties.TryGetValue(key: "status", out JsonElement statusValue) && statusValue.ToString() == "Healthy",
+                message: "Expected endpoint to be healthy.");
+
+            Assert.IsTrue(
+                condition: responseProperties.TryGetValue(key: "version", out JsonElement versionValue) && versionValue.ToString() == ProductInfo.GetMajorMinorPatchVersion(),
```
Aniruddh25, line 2784: can reuse the constant DAB_VERSION_KEY here to lookup version

Aniruddh25 (changes requested):
Looking for clarity on need for inclusion of commit hash in CLI returned version

src/Service/HealthCheck/HealthReportResponseWriter.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <summary>
+    /// Creates a JSON response for the health check endpoint using the provided health report.
+    /// If the response has already been created, it will be reused.
+    /// </summary>
+    public class HealthReportResponseWriter
+    {
+        private byte[]? _responseBytes;
+        private const string JSON_CONTENT_TYPE = "application/json; charset=utf-8";
+        private const string DAB_VERSION_KEY = "version";
+        private const string DAB_APPNAME_KEY = "appName";
```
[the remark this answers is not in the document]
seantleonard: done. updated to public const and moved from this class to DabHealthCheck.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: yep, did the same with app-name const too and updated both this test file and all other healthcheck related files.

src/Cli.Tests/EndToEndTests.cs
seantleonard: ah per that issue, the problem was that we were relying on FileVersionInfo.ProductVersion which included the hash. We can instead concatenate the following to enable returning a version without the hash and still allowing the hash value to be present in the assembly metadata for NuGet usage.

```csharp
fileVersionInfo.ProductMajorPart + "." + fileVersionInfo.ProductMinorPart + "." + fileVersionInfo.ProductBuildPart;
```

src/Product/ProductInfo.cs
seantleonard: consolidated to GetProductVersion and removed any usage of commit hash.

src/Cli.Tests/EndToEndTests.cs
seantleonard: I need to keep this for CLI. We can have follow up item track reimplementing the health screen for this limitation: https://github.com/commandlineparser/commandline/issues/757#issuecomment-1005505288

seantleonard: the follow up item and existing issue is the one you linked: https://github.com/Azure/data-api-builder/issues/1456. I'll add my suggestions for how to implement this there.

src/Product/ProductInfo.cs
seantleonard: Revised: Dab health endpoint does not reference commit hash. Commit hash only referenced in CLI due to how we use CommandLineParser dependency. #1456 tracks the fix. But I still consolidated GetProductVersion() v/s GetMajorMinorPatchVersion() as you suggested here.

src/Product/ProductInfo.cs
```diff
@@ -8,44 +8,42 @@ namespace Azure.DataApiBuilder.Product;
...
+    /// <param name="includeCommitHash">If true, returns the version string with the commit hash</param>
+    /// <returns>Version string "Major.Minor.Patch"</returns>
+    public static string GetProductVersion(bool includeCommitHash = false)
     {
         Assembly assembly = Assembly.GetExecutingAssembly();
         FileVersionInfo fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
-        string? version = fileVersionInfo.ProductVersion;
 
-        return version ?? DEFAULT_VERSION;
+        if (includeCommitHash)
```
Aniruddh25, line 28: #1456 can be used to track the issue with `dab --version` alone. We can still prevent the hash from occurring in the first line of individual commands like `dab update` if we dont add this `includeCommitHash` parameter altogether.

src/Cli/Commands/ValidateOptions.cs
```diff
@@ -26,7 +26,7 @@ public ValidateOptions(string config)
         /// </summary>
         public void Handler(ILogger logger, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
         {
-            logger.LogInformation("{productName} {version}", PRODUCT_NAME, ProductInfo.GetProductVersion());
+            logger.LogInformation("{productName} {version}", PRODUCT_NAME, ProductInfo.GetProductVersion(includeCommitHash: true));
```
Aniruddh25, line 29: we dont need the hash even on the first line of the CLI command, so we can at least avoid the hash here by not changing this file altogether.

seantleonard: Thanks for pointing out. I too eagerly reverted to using commit hash. Commit hash will only be printed when: CLI usage with `--help`, `--version` , or unknown verbs/options are used.
Tests and CLI options parsing now properly only prints version without commit hash where possible.

src/Product/ProductInfo.cs
seantleonard: Thanks for pointing out. I too eagerly reverted to using commit hash. Commit hash will only be printed when: CLI usage with `--help`, `--version` , or unknown verbs/options are used.
Tests and CLI options parsing now properly only prints version without commit hash where possible.

seantleonard:
Updated the PR description to call out the specific circumstances where version includes the commit hash.

src/Product/ProductInfo.cs
```diff
@@ -8,44 +8,42 @@ namespace Azure.DataApiBuilder.Product;
...
+        {
+            return fileVersionInfo.ProductVersion ?? "DAB_UNVERSIONED";
```
abhishekkumams, line 30: will there be any scenario where it will return DAB_UNVERSIONED? Also why did we choose this over "0.0.0"?

seantleonard: Agree unversioned is not helpful. I'll come up with better value. I added that value because `fileVersionInfo.ProductVersion` is nullable.

 Not using `0.0.0` because that does not accurately reflect the version.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -514,37 +515,80 @@ public void CleanupAfterEachTest()
...
         [DataRow(DatabaseType.PostgreSQL, "Something;Application Name=CustAppName;", "Something;Application Name=CustAppName;", true, DisplayName = "[PGSQL]:No Change in connectionString containing customer Application name for DAB hosted.")]
         [DataRow(DatabaseType.CosmosDB_PostgreSQL, "Something;Application Name=CustAppName;", "Something;Application Name=CustAppName;", true, DisplayName = "[COSMOSDB_PGSQL]:No Change in connectionString containing customer Application name for DAB hosted.")]
         [DataRow(DatabaseType.CosmosDB_NoSQL, "Something;Application Name=CustAppName;", "Something;Application Name=CustAppName;", true, DisplayName = "[COSMOSDB_NOSQL]:No Change in connectionString containing customer Application name for DAB hosted.")]
-        [DataRow(DatabaseType.MSSQL, "Data Source=<>;App=CustAppName;User ID=<>", "Data Source=<>;User ID=<>;Application Name=CustAppName,dab_oss_1.0.0", false, DisplayName = "[MSSQL]:Updating connectionString containing `App` for customer Application name with dab_oss app name.")]
```
abhishekkumams, line 547: why are these tests removed?

abhishekkumams:
can we add a test where we see the app_name as `dab_hosted_1.0.0` and for oss `dab_oss_1.0.0` ?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -746,6 +744,40 @@ string options
...
     }
 
+    [DataRow("", "--version", false, DisplayName = "Checking dab version with --version.")]
+    [DataRow("", "--help", false, DisplayName = "Checking version through --help option.")]
+    [DataRow("edit", "--new-option", false, DisplayName = "Version printed with invalid command edit.")]
+    [DataTestMethod]
+    public void InvalidCliVerbsAndOptions_DisplayVersionWithCommitHashAndConfigFileName(
+    string command,
+    string options,
+    bool isParsableDabCommandName)
```
aaronburtle, line 754: since all the cases have `isParsableDabCommandName` as `false` do we really need it as a variable in the data rows?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -746,6 +744,40 @@ string options
...
+        string? output = process.StandardOutput.ReadLine();
+        Assert.IsNotNull(output);
+
+        // Version Info logged by dab irrespective of commands being parsed correctly.
+        // When DAB CLI (CommandLineParser) detects that usage includes --version, --help, or invalid verbs/options,
+        // CommandLineParser's default HelpWriter is used to display the version and help information.
+        // Because the HelpWriter uses fileVersionInfo.ProductVersion, the version includes the commit hash.
+        StringAssert.Contains(output, $"{Program.PRODUCT_NAME} {ProductInfo.GetProductVersion(includeCommitHash: true)}", StringComparison.Ordinal);
+
+        if (isParsableDabCommandName)
```
aaronburtle, line 772: when is this conditional evaluating to true?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -514,37 +515,80 @@ public void CleanupAfterEachTest()
         }
 
         /// <summary>
-        /// Checks if the connection string provided in the config is correctly updated for MSSQL.
+        /// Checks if the connection string provided in the config is correctly updated for MSSQL by checking for the injected User Agent string w/ format:
```
aaronburtle, line 518: nit: some of these lines in the summary comment are a bit long and would benefit from being broken up.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -514,37 +515,80 @@ public void CleanupAfterEachTest()
...
         /// If the connection string already contains the `Application Name` property, it should append the DataApiBuilder Application Name to the existing value.
         /// If not, it should append the property `Application Name` to the connection string.
+        /// Also validates that any custom set environment variable(e.g. hosting) for DAB_APP_NAME_ENV is used in the connection string for Application Name (or appended to the existing value).
+        /// </summary>
+        /// <param name="databaseType">database type.</param>
+        /// <param name="providedConnectionString">connection string provided in the config.</param>
+        /// <param name="expectedUpdatedConnectionString">Updated connection string with Application Name.</param>
+        /// <param name="isHostedScenario">If Dab is hosted or OSS.</param>
+        [DataTestMethod]
+        [DataRow(DatabaseType.MSSQL, "Data Source=<>;", "Data Source=<>;Application Name=", false, DisplayName = "[MSSQL]:Adding Application Name property to connectionString with dab_oss app name.")]
```
aaronburtle, line 529: nit: readability may be improved by moving the display name or more to new lines

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -514,37 +515,80 @@ public void CleanupAfterEachTest()
...
+            {
+                Environment.SetEnvironmentVariable(ProductInfo.DAB_APP_NAME_ENV, "dab_hosted_1.0.0");
+            }
+            else
+            {
+                Environment.SetEnvironmentVariable(ProductInfo.DAB_APP_NAME_ENV, null);
+            }
+
+            // Resolve assembly version. Not possible to do in DataRow as DataRows expect compile-time constants.
+            string assemblyResolvedVersion = ProductInfo.GetDataApiBuilderUserAgent();
```
aaronburtle, line 551: nit: if we are resolving the assembly version, the name `resolvedAssemblyVersion` may be more descriptive.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2687,6 +2731,65 @@ public async Task OpenApi_GlobalEntityRestPath(bool globalRestEnabled, bool expe
...
+            Dictionary<string, JsonElement> responseProperties = JsonSerializer.Deserialize<Dictionary<string, JsonElement>>(responseBody);
+
+            // Validate response body
+            Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+
+            Assert.IsTrue(
+                condition: responseProperties.TryGetValue(key: "status", out JsonElement statusValue) && statusValue.ToString() == "Healthy",
+                message: "Expected endpoint to be healthy.");
+
+            Assert.IsTrue(
```
aaronburtle, line 2784: nit: some of these lines are a bit long, breaking up the conditional might help readability (i just hate scrolling to the right in the editor :)

src/Service/HealthCheck/HealthReportResponseWriter.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            JsonWriterOptions options = new() { Indented = true };
+
+            using MemoryStream memoryStream = new();
+            using (Utf8JsonWriter jsonWriter = new(memoryStream, options))
+            {
+                jsonWriter.WriteStartObject();
+                jsonWriter.WriteString("status", healthReport.Status.ToString());
+
+                if (healthReport.Entries.TryGetValue(key: typeof(DabHealthCheck).Name, out HealthReportEntry healthReportEntry))
```
aaronburtle, line 63: if these lookups fails, do we want to do anything about it? Print or log error, warning, etc?

aaronburtle (approved):
It looks good. Few comments about formatting, and some clarifying questions.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: new lines in my opinion, for this test specifically, would make the datarows harder to read. Harder to distinguish between actual and expected connection strings. I updated the datarows to have more spacing to be clearer. This will be readable in Visual Studio or in CodeSpaces. Not necessarily github.
image

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: Updated these to be broken up into if statements for the "TryGetValue" which then allowed me to break the conditions up to new lines and be more readable.

seantleonard: image

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: MSSQL specific tests are moved to their own testmethod for readability since the behavior for MSSQL is different from all other db's.

src/Service/HealthCheck/HealthReportResponseWriter.cs
seantleonard: Good point. Adding logger.

src/Cli.Tests/EndToEndTests.cs
seantleonard: nope, good catch, fixed.

abhishekkumams (approved):

JerryNixon:
This is awesome!

PR #2099, opened 2024-03-13
Fix dab validate schema fetch Error from URL

### LINKED ISSUE #2013: `dab validate` fetch schema from url error
URL referenced json schemas fail to resolve when using `dab validate`

_Originally posted by @Aniruddh25 in https://github.com/Azure/data-api-builder/pull/2011#discussion_r1478927487_

abhishekkumams:
## Why make this change?

- Closes #2013
  - An error is thrown when NJsonSchema tries to fetch the schema from url, due to failing to add refrences.

## What is this change?

- since, we only need to fetch the data from the schema file, we can simply make a get call to the url, instead of using NJsonSchema.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)


src/Core/Configurations/JsonConfigSchemaValidator.cs
```diff
@@ -28,6 +29,19 @@ public JsonConfigSchemaValidator(ILogger<JsonConfigSchemaValidator> jsonSchemaVa
         _fileSystem = fileSystem;
     }
 
+    /// <summary> 
+    /// Sets the logger, file system, and HttpClient for the JSON config schema validator. 
+    /// </summary> 
+    /// <param name="jsonSchemaValidatorLogger">The logger to use for the JSON schema validator.</param>
```
seantleonard, line 35: Note in PR description why we need second constructor versus just adding this new param to the original constructor.
backwards compat?

seantleonard: If you use constructor without httpClient, won't the call to `GetJsonSchema` fail because client is null.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1323,6 +1324,55 @@ public async Task TestEngineCanStartConfigWithCustomProperties()
...
+        /// <summary>
+        /// This test checks that the GetJsonSchema method of the JsonConfigSchemaValidator class
+        /// correctly downloads a JSON schema from a given URL, and that the downloaded schema matches the expected schema.
+        /// </summary>
+        [TestMethod]
+        public async Task GetJsonSchema_DownloadsSchemaFromUrl()
+        {
+            // Arrange
+            Mock<HttpMessageHandler> handlerMock = new(MockBehavior.Strict);
+            string jsonSchemaContent = "{\"type\": \"object\", \"properties\": {\"property1\": {\"type\": \"string\"}}}";
```
aaronburtle, line 1336: It sounds like the schema validation was never really working, but now that we use our own http client to get get from the url, it should function. However, if we are mocking the value returned, are we really testing the case that was failing previously? Is there a way to actually get the file that we want, or perhaps have it available locally, so that the test validates we have fixed the scenario?

src/Core/Configurations/JsonConfigSchemaValidator.cs
```diff
@@ -71,8 +85,10 @@ public async Task<JsonSchemaValidationResult> ValidateJsonConfigWithSchemaAsync(
         {
             try
             {
-                JsonSchema jsonSchema = await JsonSchema.FromUrlAsync(runtimeConfig.Schema);
-                return jsonSchema.ToJson();
+                // Send a GET request to the URL specified in runtimeConfig.Schema to get the JSON schema.
+                HttpResponseMessage response = await _httpClient.GetAsync(runtimeConfig.Schema);
```
seantleonard, line 89: This restricts us to using a file hosted behind a url. Is there any way to support a user providing their own local file to use for schema validation? e.g. environment with no internet connection or our own local testing of dab: we don't have a valid url pointing to a JsonSchema (`https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json`) until an official release is published on github and the pipeline substitutes the major/minor/patch values.

So in addition to `dab -validate dab-config.json`, the user can also do `dab -validate dab-config.json C:\files\dab-config-schema.json`

src/Core/Configurations/JsonConfigSchemaValidator.cs
```diff
@@ -71,8 +85,10 @@ public async Task<JsonSchemaValidationResult> ValidateJsonConfigWithSchemaAsync(
...
+                string jsonSchema = await response.Content.ReadAsStringAsync();
+                return jsonSchema;
```
seantleonard, line 91: please include the error message that was appearing before this change to illustrate the URL connectivity was to blame and that there wasn't a problem with the Schema file that was downloaded by Newtonsoft jsonSchema

seantleonard (changes requested):

src/Core/Configurations/JsonConfigSchemaValidator.cs
seantleonard: allowing us to use a local file can be separate work (open an issue and respond with the issue number here to close out)

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: it was working, we have a fail safe method, if it was not able to fetch from the url due to any reasons, it gets the schema from the package itself to verify. 
We were using a different method before and then casting it to string, it was failing before because some refrences were not resolved when reading it as JsonSchema, since we need the data as string, this new approach is much better because it gives the data in the format we need.

This test here is to verify that we are able to fetch the data from the url.

src/Core/Configurations/JsonConfigSchemaValidator.cs
abhishekkumams: client is never null. it is initialized in the begining itself. But good point. I can actually have just one constructor as well.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: > it was failing before because some refrences were not resolved when reading it as JsonSchema, 

This wasn't in pr description. can you elaborate since this sounds like it is more than just URL fetching that was an issue, and instead, was an issue with how the fetched JSON schema was processed by newtonsoft. Do unresolved references mean that there is something wrong with our JSON schema?

aaronburtle: > it was working, we have a fail safe method, if it was not able to fetch from the url due to any reasons, it gets the schema from the package itself to verify. We were using a different method before and then casting it to string, it was failing before because some refrences were not resolved when reading it as JsonSchema, since we need the data as string, this new approach is much better because it gives the data in the format we need.
> 
> This test here is to verify that we are able to fetch the data from the url.

So if we were to have this same test with the mocked data being returned, but with the previous code from before the fix, would it have passed?

My concern is just that we dont have a test that would fail before and pass now to prevent regression, if that makes sense.

src/Core/Configurations/JsonConfigSchemaValidator.cs
abhishekkumams: in case the fetch from url fails we get the jsonSchema from the package itself. So we always have schema to validate from

src/Core/Configurations/JsonConfigSchemaValidator.cs
abhishekkumams: created issue to track it #2124

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: we didn't had a test before for testing the url section, the previous tests were passing because all of them were fetching the schema from the package and not the url. this time i added a test that tests the url. So now we have tests that covers both scenarios.

src/Core/Configurations/JsonConfigSchemaValidator.cs
```diff
@@ -87,8 +76,15 @@ public async Task<JsonSchemaValidationResult> ValidateJsonConfigWithSchemaAsync(
...
-                return jsonSchema;
+                if (response.IsSuccessStatusCode)
+                {
+                    string jsonSchema = await response.Content.ReadAsStringAsync();
+                    return jsonSchema;
+                }
+                else
+                {
+                    _logger!.LogError($"Failed to get schema from url: {runtimeConfig.Schema}");
+                }
```
seantleonard, line 87: nit: not blocking this PR, but this error is no different than the error printed in the `catch` block. The `catch` block is catching specific errors like no connect, bad uri, timeouts, but this else statement catches an http response not equal to 2XX. We can give user more informative message to get this working.

Also, any errors sourced from failure to resolve schema file from URI should be logged as warnings because we have a fallback method of reading the schema file from the local repo. Any failure from the fallback method should definitely still be logged as "error"

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1373,6 +1373,47 @@ public async Task GetJsonSchema_DownloadsSchemaFromUrl()
...
+                Schema: url,
+                DataSource: new(DatabaseType.MSSQL, "connectionString", null),
+                new RuntimeEntities(new Dictionary<string, Entity>())
+            );
+
+            // Act
+            string recievedJsonSchema = await jsonConfigSchemaValidator.GetJsonSchema(runtimeConfig);
+
+            // Assert
+            Assert.IsFalse(string.IsNullOrEmpty(recievedJsonSchema));
```
seantleonard, line 1414:
```suggestion
            string receivedJsonSchema = await jsonConfigSchemaValidator.GetJsonSchema(runtimeConfig);

            // Assert
            Assert.IsFalse(string.IsNullOrEmpty(receivedJsonSchema));
```

seantleonard: This test doesn't really validate that we actually received a json schema.. maybe best to also includes a sanity check `contents.StartsWith("{ "my json text":....")`

seantleonard (approved):
little nits but can be merged. thanks for making this update.

src/Core/Configurations/JsonConfigSchemaValidator.cs
abhishekkumams: makes sense. thanks!

aaronburtle (approved):
LGTM!

seantleonard (approved):

PR #2100, opened 2024-03-13
fix relationship exception on dab validate

### LINKED ISSUE #2037: [Bug]: `dab validate` returns an Unhandled Exception if a table definition is not inferred
### What happened?

Provided a config where the source object for the `Book` entity was not present in the file. 
Expected dab validate to just error out saying unable to fetch schema for `Book`.
Instead, it hangs as follows throwing an unhandled exception:
image

### Version

0.11.6-rc

Azure SQL

Local (including CLI)

REST

### Relevant log output

```Text
Information: Microsoft.DataApiBuilder 0.11.6-rc+f557b4da4c350f878d9a44632037daef9349f769
Information: Environment variable DAB_ENVIRONMENT is (null)
Loading config file from dab-config.json.
Error: Failed to get schema from url: https://dataapibuilder.azureedge.net/schemas/v0.11.6-rc/dab.draft.schema.json
System.Net.Http.HttpRequestException: Response status code does not indicate success: 404 (Not Found).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at CallSite.Target(Closure , CallSite , Object )
   at System.Dynamic.UpdateDelegates.UpdateAndExecuteVoid1[T0](CallSite site, T0 arg0)
   at NJsonSchema.Infrastructure.DynamicApis.HttpGetAsync(String url, CancellationToken cancellationToken)
   at NJsonSchema.JsonSchema.FromUrlAsync(String url, Func`2 referenceResolverFactory, CancellationToken cancellationToken)
   at Azure.DataApiBuilder.Core.Configurations.JsonConfigSchemaValidator.GetJsonSchema(RuntimeConfig runtimeConfig)
Loading config file from dab-config.json.
Information: [Book] REST path: /api/Book
Unhandled exception. System.AggregateException: One or more errors occurred. (Table Definition for Book has not been inferred.)
 ---> Azure.DataApiBuilder.Service.Exceptions.DataApiBuilderException: Table Definition for Book has not been inferred.
   at Azure.DataApiBuilder.Core.Services.SqlMetadataProvider`3.GetSourceDefinition(String entityName)
   at Azure.DataApiBuilder.Core.Services.SqlMetadataProvider`3.LogPrimaryKeys()
   at Azure.DataApiBuilder.Core.Services.SqlMetadataProvider`3.InitializeAsync()
   at Azure.DataApiBuilder.Core.Services.MetadataProviders.MetadataProviderFactory.InitializeAsync()
   at Azure.DataApiBuilder.Core.Configurations.RuntimeConfigValidator.ValidateEntitiesMetadata(RuntimeConfig runtimeConfig, ILoggerFactory loggerFactory)      
   at Azure.DataApiBuilder.Core.Configurations.RuntimeConfigValidator.TryValidateConfig(String configFilePath, RuntimeConfig runtimeConfig, ILoggerFactory loggerFactory, Boolean isValidateOnly)
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at Cli.ConfigGenerator.IsConfigValid(ValidateOptions options, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
   at Cli.Commands.ValidateOptions.Handler(ILogger logger, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
   at Cli.Program.<>c__DisplayClass2_0.<Execute>b__5(ValidateOptions options)
   at CommandLine.ParserResultExtensions.WithParsed[T](ParserResult`1 result, Action`1 action)
   at Cli.Program.Execute(String[] args, ILogger cliLogger, IFileSystem fileSystem, FileSystemRuntimeConfigLoader loader)
   at Cli.Program.Main(String[] args)
```

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #2037 

## What is this change?

- A lot of places in our code, we have had assumptions of happy paths and no failures, no null values, and it was working fine as of now because when starting the engine, any exception once thrown usually exits the engine. But `dab validate` tries to collect all the exception, so some part of code might be triggered which was excepting some values. This PR addresses that and put more conservative approach to some of the code we have, by adding null checks and try/catch block.

## How was this tested?

- [X] Integration Tests


src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -323,26 +323,37 @@ private void LogPrimaryKeys()
...
             foreach ((string entityName, Entity _) in _entities)
             {
-                SourceDefinition sourceDefinition = GetSourceDefinition(entityName);
-                _logger.LogDebug("Logging primary key information for entity: {entityName}.", entityName);
-                foreach (string pK in sourceDefinition.PrimaryKey)
+                try
                 {
-                    column = sourceDefinition.Columns[pK];
-                    if (TryGetExposedColumnName(entityName, pK, out string? exposedPKeyName))
+                    SourceDefinition sourceDefinition = GetSourceDefinition(entityName);
```
aaronburtle, line 328: Can this call to `GetSourceDefinition` throw an exception, and if so, does the message about logging primary keys make sense?

If it can't throw an exception, does that mean that the only call in here that can is the call to TryGetExposedColumnName? If so, would it help to tailor the exception messaging toward that?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -720,6 +739,14 @@ private static EntitySourceType GetEntitySourceType(string entityName, Entity en
                     throw new InvalidOperationException($"Target Entity {targetEntityName} should be one of the exposed entities.");
                 }
 
+                if (targetEntity.Source.Object is null)
+                {
+                    throw new DataApiBuilderException(
+                                message: $"Target entity {entityName} does not have a valid source object.",
+                                statusCode: HttpStatusCode.NotAcceptable,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
+                }
```
seantleonard, line 748: Would this  ever be true? Wouldn't the error check you added earlier catch this? `entity.Source.Object is null)` in `GenerateDatabaseObjectForEntities()` which i assume runs first? 

Or is this needed because: yes you do hit the check i just mentioned, but need to continue execution for Dab validate to catch any other errors?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -953,7 +980,16 @@ private async Task PopulateObjectDefinitionForEntities()
...
+            try
+            {
+                await PopulateForeignKeyDefinitionAsync();
+            }
+            catch (Exception e)
+            {
+                HandleOrRecordException(e);
+            }
+
+            // await PopulateForeignKeyDefinitionAsync();
```
seantleonard, line 992:
```suggestion
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1200,6 +1200,94 @@ public async Task TestSqlMetadataForInvalidConfigEntities()
...
+                await configValidator.ValidateEntitiesMetadata(configProvider.GetConfig(), mockLoggerFactory);
+            }
+            catch
+            {
+                Assert.Fail("The validation should not throw an exception.");
+            }
+
+            Assert.IsTrue(configValidator.ConfigValidationExceptions.Any());
+            List<string> exceptionMessagesList = configValidator.ConfigValidationExceptions.Select(x => x.Message).ToList();
+            Assert.IsTrue(exceptionMessagesList.Contains("The entity Book does not have a valid source object."));
```
aaronburtle, line 1283: Should we confirm that we only return these messages once? What if the validation is adding erroneous messaging, should we confirm that this is the limit of the messaging?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1200,6 +1200,94 @@ public async Task TestSqlMetadataForInvalidConfigEntities()
             Assert.AreEqual("No stored procedure definition found for the given database object publishers", exceptionsList[1].Message);
         }
 
+        /// <summary>
+        /// This Test validates that when the entities in the runtime config have source object as null,
+        /// the validation exception handler collects the message and exits gracefully.
+        /// </summary>
+        [TestMethod("Validate Exception handling for Entities with Source object as null."), TestCategory(TestCategory.MSSQL)]
+        public async Task TestSqlMetadataValidationForEntitiesWithInvalidSource()
```
aaronburtle, line 1208: Should we add a valid entity and then confirm that we don't have error messaging for the entity that should not be throwing any errors?

seantleonard (commented):
some feedback for making error messages a little clearer

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
abhishekkumams: this is to catch other errors. Also, the previous one was for source entity, this is for target entity.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1200,6 +1200,94 @@ public async Task TestSqlMetadataForInvalidConfigEntities()
...
+
+            ILoggerFactory mockLoggerFactory = TestHelper.ProvisionLoggerFactory();
+
+            try
+            {
+                await configValidator.ValidateEntitiesMetadata(configProvider.GetConfig(), mockLoggerFactory);
+            }
+            catch
+            {
+                Assert.Fail("The validation should not throw an exception.");
```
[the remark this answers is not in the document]
abhishekkumams: Yes, makes sense. thanks!

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
abhishekkumams: `GetSourceDefinition` can throw exception. The message about logging primary keys still makes sense, because that is what we are trying to achieve and failure to do so should be logged with appropriate message. So if exception is thrown at `GetSourceDefinition`, it will give failure to  log primary keys because of SourceDefinition error.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -323,26 +323,37 @@ private void LogPrimaryKeys()
...
+                                column.SystemType.Name,
+                                column.IsNullable,
+                                column.IsAutoGenerated);
+                        }
                     }
                 }
+                catch (Exception ex)
+                {
+                    HandleOrRecordException(new DataApiBuilderException(
+                        message: $"Failed to log primary key information for entity: {entityName} due to: {ex.Message}",
```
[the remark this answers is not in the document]
abhishekkumams: one of the error message can be `Table Definition for {entityName} has not been inferred.` 
which tells that user need to verify if the config has correct object name mentioned in the config for that entityName

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: there are already tests from previous PRs that verify valid entities with no exception messages. for example, `TestSqlMetadataForValidConfigEntities` in ConfigurationTests.cs

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -806,15 +806,15 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
                 && entity.Relationships.Count > 0)
             {
                 HandleOrRecordException(new DataApiBuilderException(
-                        message: $"Cannot define relationship for entity: {entity}",
```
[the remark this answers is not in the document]
abhishekkumams: Yeah, the whole json object, which bloated the error message.

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: some messages can occur more than once if certain methods are called from multiple places or if different methods throw same error. for example `GetDatabaseObjectName` and `GetSchemaName` in SqlMetadatProvider are throwing the same error in case they dont find the mapping. 

This is because we try to capture all the error at once, this can be part of seperate PR where we refactor our code with better error messages.

abhishekkumams: For this test, i asserted on all the unique exceptions thrown. 2 errors repeated because they're called from the above two methods i mentioned.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -835,8 +835,33 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
 
-                DatabaseTable sourceDatabaseObject = (DatabaseTable)sqlMetadataProvider.EntityToDatabaseObject[entityName];
-                DatabaseTable targetDatabaseObject = (DatabaseTable)sqlMetadataProvider.EntityToDatabaseObject[relationship.TargetEntity];
+                // Validation to ensure DatabaseObject is correctly inferred from the entity name.
+                DatabaseObject? sourceObject, targetObject;
+                if (!sqlMetadataProvider.EntityToDatabaseObject.TryGetValue(entityName, out sourceObject))
+                {
+                    sourceObject = null;
+                    HandleOrRecordException(new DataApiBuilderException(
+                        message: $"Could not infer database object for entity: {entityName} in relationship: {relationshipName}.",
```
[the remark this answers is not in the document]
abhishekkumams: updated the error message.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: When i run the cli using the config file created in this test, i see duplicates of the errors Aaron mentions. While I understand that the errors may be sourced from different places, this makes the errors pretty noisy if there are duplicates.. 
- "Table Definition for Book has not been inferred" -> appears twice
image

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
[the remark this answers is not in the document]
seantleonard: closing because message now provides actionable response to check runtime config for that entity.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: > This is because we try to capture all the error at once, this can be part of seperate PR where we refactor our code with better error messages.

understood that we can always add better messages.. but duplicate messages seem like a problem that can be addressed sooner?

seantleonard: I suppose this behavior is consistent with how `dab validate` has operated thus far. open an issue to track enhancements such as this if not opened already.

seantleonard (approved):

src/Service.Tests/Configuration/ConfigurationTests.cs
abhishekkumams: created #2131 . Don't think it would be a big change but yes out of scope of this PR.

aaronburtle (approved):
Looks good, thanks for addressing all comments so quickly and opening new issues when needed.

PR #2101, opened 2024-03-13
Introduce new metadata provider initialize method 

rusamant:
## Why make this change?
GraphQL engine supports multi-tenancy (handles call for different tenants/artifacts in the same process). In Graphql workload, we are using dab as a nuget and every time we get a query request from a tenant, we use the config file of that tenant to set up the dab classes to respond to that particular request.  
In each request we call metadataprovider.initializeAsync() which is an expensive call and adds overhead in graphql workload for query execution.

## What is this change?
This code adds a new initialize method which takes in the necessary objects which are required for query execution. GraphQL workload will now call this new method in its query flow. 
We also have added different converters to serialize and deserialize the different objects that get set, as Graphql workload will be storing these values in serialized format on its end.

Use of System.Text.Json serializer and deserializer
I am using System.Text.Json serializer- deserializer as compared to NewtonSoft, as System.Text.Json focuses primarily on performance, security and standards compliance. [Migrate from Newtonsoft.Json to System.Text.Json](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-6-0)
System.Text.Json doesnot support TypeNameHandling.All, that helps in handling abstract classes and its objects serialization and deserialization. Adding TypeName property on DatabaseObject level to differentiate different types of DatabaseObject as Dictionary EntityToDatabaseObject consists of different DatabaseObject objects : DatabaseTable, DatabaseView etc and eventually we need to deserialize into DatabaseObject.

System.Test.Json does not support direct serialization and deserialization of System.Type, one needs to add a customer converter to support this. Currently DatabaseObject contains SourceDefinition which contains ColumnDefinition, ColumnDefinition has property of type System.Type hence need added TypeConveter to explicitly handle serialization and deserialization.
image
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/docs/ThreatModel.md#mitigation-3

Similarly for object type - ColumnDefinition property Default Value I had to explicitly create an converter. Lets say there is a default value of type string is set of an column: firstname. The default value set is John. If I do not add an explicit deserialization then instead of "('John')" in original EntityDatabaseObject the deserialized value was looking as ValueKind = String : "('John')" which threw exeception.

How did I solve my initial infinite loop issue?
I was passing wrong element in line "DatabaseObject)JsonSerializer.Deserialize(document, concreteType, options)!;" of converter. I was passing root. Hence on correcting the object being passed it solved the infinite loop. My earlier issues discussed in first run, were not due to lack of converter, but were related to not correct object being passed, some properties on the Database have private set, hence during deserialization they were not getting set.( solution add : JsonInclude tag on property, https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonincludeattribute?view=net-8.0)

## How was this tested?
- Unit Tests
manually running local instance of DAB :
image


src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.DatabasePrimitives;
+
+namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
```
rusamant, line 9: @Aniruddh25/ @seantleonard is this the right place to keep the converters or you want me to move to different location ?

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -42,7 +42,9 @@ public bool Equals(DatabaseObject? other)
     {
         return other is not null &&
                SchemaName.Equals(other.SchemaName) &&
-               Name.Equals(other.Name);
+               Name.Equals(other.Name) &&
+               FullName.Equals(other.FullName) &&
+               ((SourceDefinition == null && other.SourceDefinition == null) || SourceDefinition?.Equals(other.SourceDefinition) == true);
```
rohkhann, line 47: if source definition is null and other.sourcedefinition is null, wont this condition evaluate to true? Is that the expectation.

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -79,6 +81,12 @@ public DatabaseTable(string schemaName, string tableName)
 
     public DatabaseTable() { }
     public SourceDefinition TableDefinition { get; set; } = null!;
+
+    public virtual bool Equals(DatabaseTable other)
+    {
+        return
+            base.Equals(other) && ((TableDefinition == null && other.TableDefinition == null) || TableDefinition?.Equals(other.TableDefinition) == true);
```
rohkhann, line 88: nit: indentation

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
```diff
@@ -0,0 +1,73 @@
...
+                return objA;
+            }
+        }
+
+        public override void Write(Utf8JsonWriter writer, DatabaseObject? value, JsonSerializerOptions options)
+        {
+            writer.WriteStartObject();
+
+            // Add TypeName
+            writer.WriteString("TypeName", value!.GetType().AssemblyQualifiedName);
```
rohkhann, line 43: nit: I see TypeName being used in multiple places in this class. can we move it to a constant?

rohkhann, line 43: Didnt understand this part where databaseObject is a nullable type but we explicitly say not null on line 43. Could we add a comment here explaining why it cant be null?

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            // Add other properties of DatabaseObject
+            foreach (PropertyInfo prop in value.GetType().GetProperties())
+            {
+                // Skip the TypeName property, as it has been handled above
+                if (prop.Name == "TypeName")
+                {
+                    continue;
```
rohkhann, line 51: Could you add a comment on why TypeName cannot be handled as part of this loop only and why it needs to be seperate as above?

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
```diff
@@ -0,0 +1,73 @@
...
+                writer.WritePropertyName(prop.Name);
+                JsonSerializer.Serialize(writer, prop.GetValue(value), options);
+            }
+
+            writer.WriteEndObject();
+        }
+
+        private static Type GetTypeFromName(string typeName)
+        {
+            Type type = Type.GetType(typeName)!;
```
rohkhann, line 63: If you are doing the null check below, then this should not have the not null ! right? you should probably do Type? type = Type.GetType(typeName);

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
seantleonard: Converters go in `Azure.DataApiBuilder.Config` project in `Converters` folder

seantleonard: Though actually, since these converters don't apply to config, and instead apply to objects defined in the MetadataProviders, perhaps create a folder here called `Converters`

src/Config/DatabasePrimitives/DatabaseObject.cs
rusamant: yes that is the expectation, both the object should be equal if they are null or not.

rohkhann (commented):
Some initial comments. Will try to test out locally and will give more feedback if applicable.

seantleonard:
include in your pr description what you identified to help you use system.text.json over newtonsoft. How did you get around the infinite loops you initially experienced, etc.

src/Config/DatabasePrimitives/DatabaseObject.cs
rohkhann: I meant is not null sorry. Like one is null and the other is not.

rusamant: if one is null and other is not then second half of the condition will return false.

SourceDefinition?.Equals(other.SourceDefinition) == true
scenario 1 : SourceDefinition is null and other.SourceDefinition is not null
the above will translate to null == true which will return false

scenario 2: Source Definition is not null and other.SourceDefinition is null
SourceDefinition.Equals(null) will return false.

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
rusamant: Updated the comment in code

This line will : 
Add TypeName property in DatabaseObject object that we are serializing based on its type. The DatabaseObject can be of many types : DatabaseTable, DatabaseView etc. We add this property to differentiate between them in the dictionary. This extra property gets used in deserialization above. For example if object is DatabaseTable then we need to add "TypeName": "Azure.DataApiBuilder.Config.DatabasePrimitives.DatabaseTable, Azure.DataApiBuilder.Config, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null we cannot perform GetType() on a null object hence the check.

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
rusamant: it doesnot allow me to, it gives me build error. even if I add multiple null checks above.

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
rusamant: I have updated above what the above means and what typename means.

As TypeName is the property that we are adding explicitly in serialized way above we skip it and iterate over other properties of DatabaseObject. The above line comment self explains why this skip, hence not adding more

src/Core/Services/MetadataProviders/DatabaseObjectConverter.cs
rusamant: Added a folder : "Convertors"  inside Metadataproviders folder and moved all the converters inside it.

src/Config/DatabasePrimitives/DatabaseObject.cs
rusamant: tested with a unit test and sent screenshot offline

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
```diff
@@ -0,0 +1,134 @@
...
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
+{
+    /// <summary>
+    /// This is a converter to serialize and deserialize object type
+    /// can object can be of different types int, long, string, datetime. this converter handles different basic object types supported by
+    /// Hotchocolate
```
vadeveka, line 12: Can we document the reference for this?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -0,0 +1,403 @@
...
+            // and deserialization test.
+            int fields = typeof(DatabaseTable).GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance).Length;
+            Assert.AreEqual(fields, 6);
+
+            string serializedDatabaseTable = JsonSerializer.Serialize(_databaseTable, _options);
+            DatabaseTable deserializedDatabaseTable = JsonSerializer.Deserialize<DatabaseTable>(serializedDatabaseTable, _options)!;
+
+            Assert.AreEqual(deserializedDatabaseTable.SourceType, _databaseTable.SourceType);
+            Assert.AreEqual(deserializedDatabaseTable.FullName, _databaseTable.FullName);
+            deserializedDatabaseTable.Equals(_databaseTable);
```
vadeveka, line 53: Are we able to leverage reflection to compare properties?
Relying on developer to update Equals on new props sounds unreliable.

src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                throw new ArgumentNullException("Database Object being serialised cannot be null");
+            }
+
+            writer.WriteStartObject();
+
+            // Add TypeName property in DatabaseObject object that we are serializing based on its type. (DatabaseTable, DatabaseView)
+            // We add this property to differentiate between them in the dictionary. This extra property gets used in deserialization above.
+            // for example if object is DatabaseTable then we need to add
+            // "TypeName": "Azure.DataApiBuilder.Config.DatabasePrimitives.DatabaseTable, Azure.DataApiBuilder.Config, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",
```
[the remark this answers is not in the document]
vadeveka: Would it also fail if namespace was changed?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: yes I tried, but in reflection also we need to use equals at one point. and we cannot equals on all objects. Hence I use reflections atleast to catch the property change so that people are diverted to this test.

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
```diff
@@ -0,0 +1,134 @@
...
+                    else if (reader.TryGetDouble(out double doubleValue))
+                    {
+                        return doubleValue;
+                    }
+                    else
+                    {
+                        throw new JsonException($"Unable to deserialize number value. Token: {reader.TokenType}");
+                    }
+
+                case JsonTokenType.String:
```
vadeveka, line 46: To verify reliability of approach as discussed offline

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -192,5 +192,14 @@ public DatabaseObject GetDatabaseObjectForGraphQLType(string graphqlType)
         /// mode, it returns false.
         /// </summary>
         public bool IsDevelopmentMode();
+
+        /// <summary>
+        /// Initializes this metadata provider for the runtime.
+        /// This method will take in various objects which we can use directly rather than recreating the objects
+        /// using multiple tsql query
```
seantleonard, line 199:
```suggestion
        /// using multiple tsql queryies
```
nit

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -285,6 +285,16 @@ public async Task InitializeAsync()
             _logger.LogTrace($"Done inferring Sql database schema in {timer.ElapsedMilliseconds}ms.");
         }
 
+        /// <inheritdoc />
+        public void InitializeAsync(
+            Dictionary<string, DatabaseObject> entityDatabaseObject,
+            Dictionary<string, string> graphQLStoredProcedureExposedNameToEntityNameMap)
```
seantleonard, line 291: can you do this same organization in Factory/Metadataprovider interfaces? params on separate lines for readability

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -285,6 +285,16 @@ public async Task InitializeAsync()
...
+        {
+            GenerateDatabaseObjectForEntities();
+            EntityToDatabaseObject = entityDatabaseObject ?? EntityToDatabaseObject;
```
seantleonard, line 294: so it wouldn't be an error if `entityDatabaseObject` was null? Whichever assembly is calling this initialize should be expected to provide non-null params, right?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -285,6 +285,16 @@ public async Task InitializeAsync()
             _logger.LogTrace($"Done inferring Sql database schema in {timer.ElapsedMilliseconds}ms.");
         }
 
+        /// <inheritdoc />
+        public void InitializeAsync(
+            Dictionary<string, DatabaseObject> entityDatabaseObject,
+            Dictionary<string, string> graphQLStoredProcedureExposedNameToEntityNameMap)
+        {
+            GenerateDatabaseObjectForEntities();
```
seantleonard, line 293: can you add comments for why this is necessary? arent the databse objects provided in the parameters via `entityDatabaseObject`

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -285,6 +285,16 @@ public async Task InitializeAsync()
             _logger.LogTrace($"Done inferring Sql database schema in {timer.ElapsedMilliseconds}ms.");
         }
 
+        /// <inheritdoc />
+        public void InitializeAsync(
+            Dictionary<string, DatabaseObject> entityDatabaseObject,
```
seantleonard, line 290:
```suggestion
            Dictionary<string, DatabaseObject> entityDatabaseObjects,
```

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -192,5 +192,14 @@ public DatabaseObject GetDatabaseObjectForGraphQLType(string graphqlType)
...
+        /// </summary>
+        /// <param name="entityDatabaseObject">maps entity to DatabaseObject</param>
+        /// <param name="graphQLStoredProcedureExposedNameToEntityNameMap">Dictionary containing mapping of graphQL stored procedure exposed query/mutation name to their corresponding entity names defined in the config.</param>
+        void InitializeAsync(Dictionary<string, DatabaseObject> entityDatabaseObject, Dictionary<string, string> graphQLStoredProcedureExposedNameToEntityNameMap);
```
seantleonard, line 203:
```suggestion
        void InitializeAsync(Dictionary<string, DatabaseObject> entityDatabaseObjects, Dictionary<string, string> graphQLStoredProcedureExposedNameToEntityNameMap);
```

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
+{
+    /// <summary>
+    /// This is a converter to serialize and deserialize object type
+    /// can object can be of different types int, long, string, datetime. this converter handles different basic object types supported by
+    /// Hotchocolate
+    /// </summary>
+    public class ObjectConverter : JsonConverter<object>
```
seantleonard, line 14: Do you not have to register this converter anywhere to ensure it is only used for `object`'s within the `DatabaseObject` class hierarchy? e.g. register like the converters in Config project: `RuntimeConfigLoader.cs` `GetSerializationOptions(...)`

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
```diff
@@ -0,0 +1,134 @@
...
+                        return floatValue;
+                    }
+                    else if (reader.TryGetDouble(out double doubleValue))
+                    {
+                        return doubleValue;
+                    }
+                    else
+                    {
+                        throw new JsonException($"Unable to deserialize number value. Token: {reader.TokenType}");
+                    }
```
seantleonard, line 44: Seems like order matters here, right?  because you'd have to attempt each and every conversion if, let's say, you were trying to parse out a `double`

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
```diff
@@ -0,0 +1,134 @@
...
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
+{
+    /// <summary>
+    /// This is a converter to serialize and deserialize object type
+    /// can object can be of different types int, long, string, datetime. this converter handles different basic object types supported by
```
seantleonard, line 11: is this applicable to hot chocolate at all? Isn't this to deserialize default values for database columns? this should also be documented as the use case here.

src/Core/Services/MetadataProviders/Converters/TypeConverter.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
+{
+    /// <summary>
+    /// This is a converter to serialize and deserialize the a object of Type : System.Type
```
seantleonard, line 10:
```suggestion
    /// This is a converter to serialize and deserialize that an object of Type : System.Type
```

src/Core/Services/MetadataProviders/Converters/TypeConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// for example currently in ColumnDefiniton object, one of the property type is of System.Type property
```
seantleonard, line 11:
```suggestion
    /// For example, the ColumnDefiniton object's default value property of type System.Type.
```

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -127,6 +132,15 @@ public class ParameterDefinition
     public DbType? DbType { get; set; }
     public bool HasConfigDefault { get; set; }
     public object? ConfigDefaultValue { get; set; }
+
+    public bool Equals(ParameterDefinition other)
```
seantleonard, line 136: what are the equals overloads used for? can you add comments

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -216,6 +232,7 @@ public class RelationshipMetadata
     /// <summary>
     /// Dictionary of target entity name to ForeignKeyDefinition.
     /// </summary>
+    [JsonInclude]
```
seantleonard, line 235: do you not need to add a constructor for this class as well?

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -98,6 +101,8 @@ public class DatabaseStoredProcedure : DatabaseObject
 {
     public DatabaseStoredProcedure(string schemaName, string tableName)
         : base(schemaName, tableName) { }
+
+    public DatabaseStoredProcedure() { }
```
seantleonard, line 105: What case is it required to add these constructors?

seantleonard (commented):
still need to review tests. may not get to today.

src/Config/DatabasePrimitives/DatabaseObject.cs
rusamant: I need during deserialization to create the objects accordingly. without this my deserialization fails

src/Config/DatabasePrimitives/DatabaseObject.cs
rusamant: nope. Without constructor it works

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: as discussed offline, will come back as improvement for next round

src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
```diff
@@ -0,0 +1,84 @@
...
+            writer.WriteString(TYPE_NAME, value.GetType().AssemblyQualifiedName);
```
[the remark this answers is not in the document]
rusamant: If I use FullName, during deserialization I get JsonException cannot find the fullname, hence have to go AssemblyQualifiedName

Did a quick test again : 
image

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rusamant: yes technically it shouldnt be null, but this is just check I added if null then it remains unchanged. Currently in SqlMetadtaProvider 

it assigned :         public Dictionary<string, DatabaseObject> EntityToDatabaseObject { get; set; } =
            new(StringComparer.InvariantCulture);
           Hence if incoming is null then we donot assign null but keep it as  empty()

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rusamant: renamed to entityToDatabaseObject to match with what is the file currently

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
rusamant: same as above

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
rusamant: updated the comment

src/Config/DatabasePrimitives/DatabaseObject.cs
rusamant: not required, removed

src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
```diff
@@ -0,0 +1,84 @@
...
+        public override DatabaseObject Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            using (JsonDocument document = JsonDocument.ParseValue(ref reader))
+            {
+                JsonElement root = document.RootElement;
+                string typeName = root.GetProperty(TYPE_NAME).GetString() ?? throw new JsonException("TypeName is missing");
+
+                Type concreteType = GetTypeFromName(typeName);
+
+                if (concreteType == null)
```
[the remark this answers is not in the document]
rusamant: removed

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
rusamant: implemented as discussed offline

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
rusamant: I have updated the implementation of object converter itself, to increase the reliability, rather than  maintaining the switch for each case. I am leveraging similar approach as that of the DatabaseObjectOnverter to add extra property for Type

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
rusamant: currently converters are in core and dataobject in config, we donot have access to core in config project, do you want me to move the converters to config folder instead ? 
I had in Metadataprovider folder as suggested earlier

seantleonard: I'm not suggesting you move any files. 

I'm asking whether there is any "registration" process for the converters you wrote.
The pre-existing converters were "registered" via the following code in `RuntimeConfigLoader::GetSerializationOptions(...)`:
```csharp
JsonSerializerOptions options = new()
{
    PropertyNameCaseInsensitive = false,
    WriteIndented = true,
    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
    Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
};
```

I don't see anything similar in this PR. Is registration not necessary?

rusamant: closing this out as discussed offline

src/Core/Services/MetadataProviders/Converters/TypeConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+                throw new JsonException();
+            }
+
+            string typeName = reader.GetString()!;
+            return Type.GetType(typeName)!;
+        }
+
+        public override void Write(Utf8JsonWriter writer, Type value, JsonSerializerOptions options)
+        {
+            writer.WriteStringValue(value.FullName);
```
rohkhann, line 28: Earlier, I observed you used AssemblyQualifiedName but here you are using fullName, could you elaborate in comment here why fullname is chosen?

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
```diff
@@ -88,5 +89,19 @@ public IEnumerable<ISqlMetadataProvider> ListMetadataProviders()
         {
             return _metadataProviders.Values;
         }
+
+        /// <inheritdoc />
+        public void InitializeAsync(
```
rohkhann, line 94: Is it possible to add a test case for this to ensure that the correct metadataprovider is getting picked up by this function.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -285,6 +285,16 @@ public async Task InitializeAsync()
             _logger.LogTrace($"Done inferring Sql database schema in {timer.ElapsedMilliseconds}ms.");
         }
 
+        /// <inheritdoc />
+        public void InitializeAsync(
+            Dictionary<string, DatabaseObject> entityToDatabaseObject,
+            Dictionary<string, string> graphQLStoredProcedureExposedNameToEntityNameMap)
+        {
+            EntityToDatabaseObject = entityToDatabaseObject ?? EntityToDatabaseObject;
+            GraphQLStoredProcedureExposedNameToEntityNameMap = graphQLStoredProcedureExposedNameToEntityNameMap ?? GraphQLStoredProcedureExposedNameToEntityNameMap;
```
rohkhann, line 294: Could you add a test case for this initializeAsync as well? Also I think another test that would be useful would be query execution scenario as this is eventually going to be used to execute queries. You could use existing integration tests for query execution and replace the old initializeasync with the call to this.

rohkhann (commented):
Overall looks good to me. Will approve after these 3 comments + pending comments from vaibhav and Sean are addressed.

src/Core/Services/MetadataProviders/Converters/TypeConverter.cs
rusamant: AssemblyQualified Name brings in version name, whole path and project path. In one of the comments from Sean he suggested if we can use the FullName instead which is shorter. I tested with TypeConverter and feels like for .Net types we should be good to use Full Name instead.

the same doesnot apply to DatabaseObject, it failed and gives error on "Unable to find DatabaseTable".

rohkhann (approved):
Discussed offline. the other test cases will be added as part of a follow up pr.

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
rusamant: created a work item : "https://dev.azure.com/powerbi/AppDev/_backlogs/backlog/AppDev%20Team/Epics?workitem=1356906" as discussed offline

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
rusamant: Created a work item : "https://dev.azure.com/powerbi/AppDev/_backlogs/backlog/AppDev%20Team/Epics?workitem=1356906:"

src/Core/Services/MetadataProviders/Converters/TypeConverter.cs
rusamant: added documentation like discussed offline

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
```diff
@@ -7,128 +7,40 @@
 namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
 {
     /// <summary>
-    /// This is a converter to serialize and deserialize object type
-    /// can object can be of different types int, long, string, datetime. this converter handles different basic object types supported by
-    /// Hotchocolate
+    /// This is a converter to serialize and deserialize default
+    /// can object can be of different types int, long, string, datetime.
```
seantleonard, line 11: phrasing

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -247,6 +254,63 @@ public void TestColumnDefinitionNegativeCases()
...
+            Dictionary<string, DatabaseObject> dict = new() { { "person", _databaseTable } };
+
+            string serializedDict = JsonSerializer.Serialize(dict, _options);
+            Dictionary<string, DatabaseObject> deserializedDict = JsonSerializer.Deserialize<Dictionary<string, DatabaseObject>>(serializedDict, _options)!;
+
+            DatabaseTable deserializedDatabaseTable = (DatabaseTable)deserializedDict["person"];
+
+            Assert.AreEqual(deserializedDatabaseTable.SourceType, _databaseTable.SourceType);
+            Assert.AreEqual(deserializedDatabaseTable.FullName, _databaseTable.FullName);
+            deserializedDatabaseTable.Equals(_databaseTable);
```
seantleonard, line 275: i saw you removed  the overrides for `Equals`, does that apply here/ what determines whether the two databasetables are equal?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -247,6 +254,63 @@ public void TestColumnDefinitionNegativeCases()
...
+            VerifySourceDefinitionSerializationDeserialization(deserializedDatabaseTable.SourceDefinition, _databaseTable.SourceDefinition, "FirstName");
+            VerifySourceDefinitionSerializationDeserialization(deserializedDatabaseTable.TableDefinition, _databaseTable.TableDefinition, "FirstName");
+        }
+
+        /// <summary>
+        /// Test to check the impact of version change in TypeName
```
seantleonard, line 281: any examples to provide of what version string looks like before and after an update? 
The description says you check the impact of version change.. what is the expected impact when a version is updated? e.g. exception or your code gracefully handles the new version?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -247,6 +254,63 @@ public void TestColumnDefinitionNegativeCases()
...
+            DatabaseTable deserializedDatabaseTable = (DatabaseTable)deserializedDict["person"];
+            Assert.AreEqual(deserializedDatabaseTable.SourceType, _databaseTable.SourceType);
+
+            // Define the replacement version number
+            replacement = "Version=0.0.9";
+
+            // Use Regex.Replace to replace the version number in the JSON string
+            modifiedJsonString = Regex.Replace(serializedDict, pattern, replacement);
+
+            deserializedDict = JsonSerializer.Deserialize<Dictionary<string, DatabaseObject>>(modifiedJsonString, _options);
```
seantleonard, line 309: If the version number changes, I'd expect that more aspects of the object change too. For example, when `DatabaseObject` is upgraded to a new version, there is a possibility that new properties are added. Would new properties break this (de)serialization? Wouldn't the old and new objects no longer be equal?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -305,6 +369,45 @@ private void InitializeObjects()
             _databaseStoredProcedure.StoredProcedureDefinition.Parameters.Add("Id", _parameterDefinition);
         }
 
+        /// <summary>
+        /// During serialization we add TypeName to the DatabaseObject based on its type.
+        /// If the code refactor then this change will catch that as GraphQL workload would be impacted.
```
seantleonard, line 374: This sentence is unclear.
- If which code refactors? do you mean the JsonConverter you wrote for `System.Type`?
- and which change will catch the refactor? do you mean this test will catch a refactor?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -305,6 +369,45 @@ private void InitializeObjects()
...
+        /// We will have artifacts with TypeName with old project directory/ location of this class, hence during deserialization if object is not in
+        /// that location it will fail.
+        /// [Note] : is this test fails notify the graphql workload team.
```
seantleonard, line 377:
```suggestion
        /// [Note] : If this test fails, notify the GraphQL workload team.
```

Since this project is open source, this statement doesn't translate well or at the very least is unactionable by an external contributor. I'd suggest that devs should open a GitHub issue and add a specific tag that you specify -> when they encounter a failure in this test.

seantleonard: e.g. tag such as `GraphQLWorkloadBreakingChange` or `SDK-Compat`

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -305,6 +369,45 @@ private void InitializeObjects()
...
+            dict.Add("person", databaseobject);
+            string jsonString = JsonSerializer.Serialize(dict, _options);
+
+            // Find the start index of TypeName
+            int typeNameIndex = jsonString.IndexOf("\"TypeName\":\"", StringComparison.Ordinal);
+
+            if (typeNameIndex != -1)
+            {
+                // Find the end index of TypeName
+                int endIndex = jsonString.IndexOf("\",", typeNameIndex + 11, StringComparison.Ordinal);
```
seantleonard, line 391: why `11`?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -305,6 +369,45 @@ private void InitializeObjects()
...
+        private void TestTypeNameChanges(DatabaseObject databaseobject, string objectName)
+        {
+            Dictionary<string, DatabaseObject> dict = new();
+            dict.Add("person", databaseobject);
+            string jsonString = JsonSerializer.Serialize(dict, _options);
+
+            // Find the start index of TypeName
+            int typeNameIndex = jsonString.IndexOf("\"TypeName\":\"", StringComparison.Ordinal);
+
+            if (typeNameIndex != -1)
```
seantleonard, line 388: what happens when `typeNameIndex` is 1? should that result in a failure?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -305,6 +369,45 @@ private void InitializeObjects()
...
+            {
+                // Find the end index of TypeName
+                int endIndex = jsonString.IndexOf("\",", typeNameIndex + 11, StringComparison.Ordinal);
+
+                if (endIndex != -1)
```
seantleonard, line 393: what happens when `endIndex` is -1, should that result in failure? is that indicative of some error in the converter or serialization process?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -305,6 +369,45 @@ private void InitializeObjects()
...
+                {
+                    // Extract TypeName
+                    string typeName = jsonString.Substring(typeNameIndex + 11, endIndex - typeNameIndex - 11).TrimStart('\"');
+
+                    // Split TypeName into two halves
+                    string[] halves = typeName.Split(',');
```
seantleonard, line 399: example of what "halves" are?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -305,6 +369,45 @@ private void InitializeObjects()
...
         }
 
+        /// <summary>
+        /// During serialization we add TypeName to the DatabaseObject based on its type.
+        /// If the code refactor then this change will catch that as GraphQL workload would be impacted.
+        /// We will have artifacts with TypeName with old project directory/ location of this class, hence during deserialization if object is not in
+        /// that location it will fail.
+        /// [Note] : is this test fails notify the graphql workload team.
+        /// </summary>
+        private void TestTypeNameChanges(DatabaseObject databaseobject, string objectName)
```
seantleonard, line 379: please provide an example or two of what this test validates and what it is searching for in serialized object json string

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class ObjectConverter : JsonConverter<object>
+    {
+        public override object Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            using (JsonDocument doc = JsonDocument.ParseValue(ref reader))
+            {
+                JsonElement obj = doc.RootElement;
+                string typeName = obj.GetProperty("Type").GetString()!;
+                Type type = Type.GetType(typeName)!;
+                object value = JsonSerializer.Deserialize(obj.GetProperty("Value").GetRawText(), type, options)!;
```
seantleonard, line 26: now that the code that checked the type and attempted casting was removed, what is the effect? How are the types resolved now?

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
rusamant: addressed, correct the documentation

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: I have updated the documentation here now - added how typename currently looks. 
by code refactors I mean DAB codebase changes and changes the location of DatabaseObject.

image

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: added

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: reworded and added reference

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: added a variable instead, 11 is the length of TypeName": "

this is how  it looks like in serialization string
 "TypeName": "Azure.DataApiBuilder.Config.DatabasePrimitives.DatabaseTable, Azure.DataApiBuilder.Config, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",
 
 the test tried to fetch the value part:
Azure.DataApiBuilder.Config.DatabasePrimitives.DatabaseTable, Azure.DataApiBuilder.Config, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: updated the note section as suggested.

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
rusamant: as discussed offline

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: removed this test, the test passes in DAB code, but same version upgrade fails in workload. We have added a fallback code on our end if deserialization fail.

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: good catch, yes, if TypeNameIndex == -1 means TypeName substring not found in the serialized string, which is incorrect, I have added the failure else case - will just fail the test.

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: yes if any of the if fails in this test, that means during serialization we didnt add the TypeName property in DatabaseObject correctly. Failing the test in else case.

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: currently equals in DatabaseObject only checks for Name and SchemaName, hence in line 275 that will be part of equals, for rest I have added different asserts. 
I also have properties count check. I have a seperate test in file handling equality of DatabaseTable:
image

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
rusamant: yes there is a possibility of DatabaseObject having more properties, I have added check in test to compare number of properties for most of the objects I am comparing. 
Now I agree this is a bit risk as contributor needs to update the object in test and also compare that property. 

there is an idea to use reflections, but in first pass when I checked it failed. I need to spend more time in investigations but I am going to in next round.

src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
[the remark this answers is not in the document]
rusamant: yes it would break, closing this thread as discussed. Added test for namespace change.

src/Core/Services/MetadataProviders/Converters/ObjectConverter.cs
rusamant: This is the serialized version : image
deserialized back to '('John')' the original format

seantleonard (approved):
Thanks for addressing feedback!

vadeveka (approved):

src/Core/Services/MetadataProviders/Converters/TypeConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
+{
+    /// <summary>
+    /// This is a converter to serialize and deserialize an object of Type : System.Type
+    /// For example, the ColumnDefiniton object's default value property of type System.Type.
```
vadeveka, line 11: Could we add a sample serialized obejct for each converter in description?

rusamant: will add during integration test PR

Issue #2090, opened 2024-03-11
Rename Nested Mutations/Nested Create feature


severussundar:
Rename nested mutations and nested create references in the all the relevant places as this issue https://github.com/graphql/graphql-spec/issues/252 in the graphql specs contradicts with what we are calling as nested mutations.

PR #2103, opened 2024-03-13
Rename nested-mutations and nested-create to multiple-mutations and multiple-create respectively

severussundar:
## Why make this change?

-  Closes https://github.com/Azure/data-api-builder/issues/2090
  - Renames `nested-mutations` and `nested-create` to `multiple-mutations` and `multiple-create` respectively.
  - Nested Mutations has a significantly different meaning in the graphql specs (as explained in issue https://github.com/graphql/graphql-spec/issues/252) compared to what we are trying to achieve. Hence, the decision to rename. 

## What is this change?

- All references to nested mutations/nested create - option name, field names in config JSON, class names, comments in code have been renamed.

## How was this tested?

- Unit and Integration tests added as part of https://github.com/Azure/data-api-builder/pull/1983 has been updated to validate the rename of CLI option, property in the config file, etc. 

## Sample Commands

image


aaronburtle (approved):
LGTM!

seantleonard (approved):
thanks for making these changes. Also, i didn't see anything missed when searching for nested mutations/inserts in Codespaces :)

PR #2106, opened 2024-03-14
Cosmos DB: Adds Item Level Auth support using DB Policy

sourabh1007:
## Why make this change?

We need to support Item Level DB policies for cosmosdb.

```json
{
  "role": "item-level-permission-role", // Role Name
  "actions": [
    {
      "action": "read",
      "policy": { 
        "database": "@item.type eq 'earth0'"  // DB policy
      }
    }
  ]
}
```
## How is it different from Relational DB?
In case of Relational Database like SQL, each entity indicates a separate table which can be queried independently.
But in comosdb, there is nothing like table, Here we have a container which contains data in JSON format consists of entities defined in the configuration file.

## What is this change?

**Context:**  We always run a GraphQL query against a container. Hence, we have information about container and we run only those filters which are defined in the GraphQL filter.
Item level policy should always applied (no matter if entity is a part of query or not)
e.g 
**Schema:** We have a container (`Planet`) with 2 other sub entities `Character`, `Star`
``` gql
type Character {
    id : ID,
    name : String,
    type: String,
    homePlanet: Int,
    primaryFunction: String,
    star: Star
}

type Planet @model {
    id : ID!,
    name : String,
    character: Character,
    stars: [Star],
}

type Star {
    id : ID,
    name : String,
    tag: Tag
}
```

**DB Config:** In DB config, DB Policy is defined for `star`
```json
  "Planet": {
  },
  "Character": {
  },
  "Star": {
    "permissions": [
     {
        "role": "item-level-permission-role",
        "actions": [
           {
              "action": "read",
              "policy": {
                     "database": "@item.name eq 'earth0'"
                }
            }
        ]
     }]
  }
```

**GraphQL Query:** Query an entity for which policy is not defined

```graphql
string gqlQuery = @"{
    planets(first: 1, " + QueryBuilder.FILTER_FIELD_NAME + @" : { character: {type: {eq: ""Mars""}}})
    {
        items {
            id
        }
    }
}";
```

**Generated CosmosDB Query:** It should fire a query with conditions from policies along with JOIN if entity is of array type.
``` sql
SELECT c.id 
FROM c 
WHERE c.character.type = 'Mars'  ## From filter
AND c.character.star.name = 'earth0'  ## From DB Policy
AND EXISTS ( SELECT VALUE 1 
                      FROM table1 IN c.stars 
                      WHERE (table1.name = 'earth0'))  ## From DB Policy
```
### Defining Container
~~[Breaking Change: Need Documentation Update]_ Since, CosmosDB natively supports one entity per container, hence `@model` will indicate container level entity and other entities should be defined as `type` only.~~

### Process Configuration During Load 
_(src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs)_
1. Generate EDM model. https://github.com/Azure/data-api-builder/blob/8b07ad7522c7ffe2432e10acdf68889daf7792c3/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs#L94
2. Collecting all the paths for an Entity from container. https://github.com/Azure/data-api-builder/blob/8b07ad7522c7ffe2432e10acdf68889daf7792c3/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs#L101
    a) If given entity is on Array format, creating alias for the same. https://github.com/Azure/data-api-builder/blob/8b07ad7522c7ffe2432e10acdf68889daf7792c3/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs#L177
    b) Generate JOIN statement https://github.com/Azure/data-api-builder/blob/8b07ad7522c7ffe2432e10acdf68889daf7792c3/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs#L196

### Generate and Append conditions during query.
1. Use above models, generate filter clause for each entity (including resolving `@claim` also). https://github.com/Azure/data-api-builder/blob/8b07ad7522c7ffe2432e10acdf68889daf7792c3/src/Core/Resolvers/AuthorizationPolicyHelpers.cs#L85
2. Generate where conditions for an entity using prefixes collected_(at point 2 in "Process Configuration during load" section). https://github.com/Azure/data-api-builder/blob/8b07ad7522c7ffe2432e10acdf68889daf7792c3/src/Core/Resolvers/AuthorizationPolicyHelpers.cs#L113

### Other Changes:
1. I realized during implementation of this, in cosmos DB it is recommended to use subquery instead of join if entity type is of Array type. https://github.com/Azure/data-api-builder/blob/8b07ad7522c7ffe2432e10acdf68889daf7792c3/src/Core/Models/GraphQLFilterParsers.cs#L272
2. Fix tests to use model entity instead of direct type entity as it is not recommended to use `@model` for all the entities
(_src/Service.Tests/CosmosTests/MutationTests.cs_ and other test files)

### Clean up
Removed all the JOIN related changes

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests


seantleonard (commented):
Still reviewing (need another day Redmond time) but some questions that have come up so far.

src/Service.Tests/CosmosTests/TestBase.cs
```diff
@@ -56,51 +56,64 @@ public class TestBase
...
 }
 
-type MoreAttribute @model(name:""MoreAttrAlias"") {
+type MoreAttribute {
+    id : ID,
+    name : String,
+    type: String @authorize(roles: [""authenticated""])
+}
+
+type InvalidAuthModel @model @authorize(policy: ""Crater"") {
```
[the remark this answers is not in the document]
sourabh1007: where exactly, it is saying `@model` is invalid syntax. Although, I can see this statement _The directive is formatted as: @model(name:"<Entity_Name>")_but it doesn't say only `@model` is in valid.

src/Config/ObjectModel/RuntimeEntities.cs
```diff
@@ -143,4 +143,5 @@ private static Entity ProcessRestDefaults(Entity nameCorrectedEntity)
 
         return nameCorrectedEntity;
     }
+
```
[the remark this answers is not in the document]
sourabh1007: I generally fix whatever fails in pipeline..this one was passed in the pipeline. Hence it made its way to here.

Anyway, I am reverting this file.

src/Service.GraphQLBuilder/GraphQLNaming.cs
```diff
@@ -28,6 +28,11 @@ public static class GraphQLNaming
         /// <seealso cref="https://spec.graphql.org/October2021/#sec-Names.Reserved-Names"/>
         public const string INTROSPECTION_FIELD_PREFIX = "__";
 
+        /// <summary>
+        /// For any CosmosDB Query, the default alias for the container is 'c'
+        /// </summary>
+        public const string COSMOSDB_CONTAINER_DEFAULT_ALIAS = "c";
```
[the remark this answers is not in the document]
sourabh1007: I was looking for a place to put this constant. I think this place is relevant because container alias is always used while reading graphql schema.
I am open to move it to better place.

src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -48,6 +48,7 @@
       "rest": {
         "enabled": false
       },
+      "paths": [],
```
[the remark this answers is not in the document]
sourabh1007: I didn't add this config, I just generated it using config generator and push the file..

src/Core/Resolvers/CosmosQueryBuilder.cs
```diff
@@ -29,9 +30,26 @@ public string Build(CosmosQueryStructure structure)
...
             {
-                queryStringBuilder.Append($" WHERE {predicateString}");
+                queryStringBuilder.Append($" {predicateString}");
+            }
+
+            if (!string.IsNullOrEmpty(predicateString) && !string.IsNullOrEmpty(policy))
+            {
+                queryStringBuilder.Append(" AND ");
+            }
+
```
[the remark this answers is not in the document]
sourabh1007: refactored

src/Config/ObjectModel/EntityPrefix.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel
+{
+    public record EntityPrefix
```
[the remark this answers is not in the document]
sourabh1007: added documentation, open to change it to some other more meaningful name.

ayush3797:
From the PR description:

> SELECT c.id 
FROM c 
WHERE c.character.type = 'Mars' 
AND EXISTS (SELECT VALUE 1 
                      FROM c.earth " +
                       WHERE (c.earth.type = 'earth0'))  // From DB Policy
                       
                     
Do you mean `c.name` instead of `c.earth`?

sourabh1007:
> From the PR description:
> 
> > SELECT c.id
> > FROM c
> > WHERE c.character.type = 'Mars'
> > AND EXISTS (SELECT VALUE 1
> > FROM c.earth " +
> > WHERE (c.earth.type = 'earth0'))  // From DB Policy
> 
> Do you mean `c.name` instead of `c.earth`?

Thanks for pointing it out. I have updated the query in description

seantleonard:
You mention 
> [Breaking Change: Need Documentation Update] 

Please describe how this affects customers upgrading from earlier versions of DAB.
1. What changes do they need to make in their configs/apps to ensure transition between dab versions? 
    - Depending on how this impacts customers, we may not be able to merge right away without mitigating the breaking change. 
      - e.g. Customers should be able to upgrade DAB versions without error and must opt-in to this specific change by use of feature flag. 
2. Does upgrading to this version of DAB (which includes this PR's changes), do old configs cause DAB startup to fail? Do old configs cause incorrect results to be returned?

src/Config/ObjectModel/EntityDbPolicyCosmosModel.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// </summary>
+        public string? Alias { get; }
+
+        /// <summary>
+        /// Entity Name
+        /// </summary>
+        public string? EntityName { get; }
+
+        /// <summary>
+        /// Pre-generated JOIN statement for the entity
```
seantleonard, line 33: only pregenerated in some circumstances

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -272,12 +271,12 @@ public GQLFilterParser(RuntimeConfigProvider runtimeConfigProvider, IMetadataPro
...
-    /// non-scalar filter input. This function:
-    /// 1. Recursively parses any more(possibly nested) filters.
-    /// 2. Adds join predicates between the related entities.
-    /// 3. Adds the subquery to the existing list of predicates.
+    /// For CosmosDB, a nested filter represents an EXISTS clause with a subquery.
+    /// This function:
+    /// 1. Defines the Exists Query structure
+    /// 2. Recursively parses any more(possibly nested) filters on the Exists sub query.
+    /// 3. Adds join predicates between the related entities to the Exists sub query.
+    /// 4. Adds the Exists subquery to the existing list of predicates.
```
seantleonard, line 279: can you also note which parameter is modified as a result of this function? even though its return type is void, there are side effects on some of the passed in args.

src/Core/Resolvers/AuthorizationPolicyHelpers.cs
```diff
@@ -46,23 +48,154 @@ public static class AuthorizationPolicyHelpers
...
+            }
+        }
+
+        /// <summary>
+        /// Generate Cosmos DB Query for the given fromClause and predicates.
+        /// </summary>
+        /// <param name="fromClause"></param>
+        /// <param name="predicates"></param>
+        /// <returns></returns>
+        private static string GetExistQueryForCosmos(string? fromClause, string? predicates)
```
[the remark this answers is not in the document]
seantleonard: also why wouldn't this function be in CosmosQueryBuilder?

src/Core/Resolvers/AuthorizationPolicyHelpers.cs
```diff
@@ -46,23 +48,154 @@ public static class AuthorizationPolicyHelpers
...
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="authorizationResolver"></param>
+        /// <param name="sqlMetadataProvider"></param>
+        /// <param name="clientRoleHeader"></param>
+        /// <param name="elementalOperations"></param>
+        /// <param name="entityName"></param>
+        /// <param name="entityDBObject"></param>
+        /// <param name="postProcessCallback"></param>
+        /// <returns></returns>
```
seantleonard, line 170: incomplete

src/Core/Resolvers/AuthorizationPolicyHelpers.cs
```diff
@@ -46,23 +48,154 @@ public static class AuthorizationPolicyHelpers
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Generate Cosmos DB Query for the given fromClause and predicates.
+        /// </summary>
+        /// <param name="fromClause"></param>
+        /// <param name="predicates"></param>
+        /// <returns></returns>
```
seantleonard, line 141: incomplete comments

src/Core/Resolvers/AuthorizationPolicyHelpers.cs
```diff
@@ -46,23 +48,154 @@ public static class AuthorizationPolicyHelpers
...
+            {
+                existQuery += ")";
+            }
+
+            return existQuery;
+        }
+
+        /// <summary>
+        /// It process the filter clauses for the given elemental operation.
+        /// </summary>
```
seantleonard, line 161: Can you elaborate on the side effects of "processing the filter"?
- which passed in args are modified (side effects), if any
- example of what a returned filter clause list may look like?

src/Service.GraphQLBuilder/GraphQLNaming.cs
```diff
@@ -30,6 +30,11 @@ public static class GraphQLNaming
 
         public const string LINKING_OBJECT_PREFIX = "linkingObject";
 
+        /// <summary>
+        /// For any CosmosDB Query, the default alias for the container is 'c'
```
seantleonard, line 34: Consider making this a public constant in CosmosQueryStructure. Then it can be used instead of this and can also be used instead of the private var in CosmosQueryBuilder.

src/Service.Tests/CosmosTests/QueryTests.cs
```diff
@@ -307,11 +306,11 @@ public async Task GetByPrimaryKeyWhenEntityNameDoesntMatchGraphQLType()
             string id = _idList[0];
             string query = @$"
 query {{
-    star_by_pk (id: ""{id}"", _partitionKeyValue: ""{id}"") {{
+    planet_by_pk (id: ""{id}"", _partitionKeyValue: ""{id}"") {{
```
seantleonard, line 309: test method comments are now out of date. Did `star_by_pk` not work here? why can't we keep it how it was?

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -881,6 +897,50 @@ public async Task TestFilterWithEntityNameAlias()
             await ExecuteAndValidateResult(_graphQLQueryName, gqlQuery, dbQuery);
         }
 
+        /// <summary>
+        /// MoonAdditionalAttributes (array inside moon object which is an array in container): "@item.name eq 'moonattr0'"
+        /// Earth(object in object): "@item.type eq 'earth0'"
+        /// AdditionalAttribute (array in container): "@item.type eq 'volcano0'"
```
seantleonard, line 903: I don't understand what you're explaining in this comment. Explain relevance of the Policy text you're referencing.

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForCosmos.verified.txt
```diff
@@ -282,42 +298,32 @@
...
                   Exclude: [
-                    name
+                    *
                   ],
                   Include: [
-                    id,
-                    type
+                    *
                   ]
                 }
```
seantleonard, line 317: where is this used? is this a test validating that `Exclude: ["*"]` takes precedence over included fields?

src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -34,7 +34,7 @@
     }
   },
   "entities": {
-    "Planet": {
+    "PlanetAlias": {
```
seantleonard, line 37: in tests it seems like you're removing "Alias" from the entity names, but here you're adding it? I'm not following the naming changes here.

src/Core/Resolvers/CosmosQueryBuilder.cs
```diff
@@ -153,29 +164,56 @@ protected override string Build(Predicate? predicate)
...
+            }
+            else if ((s = operand.AsString()) != null)
+            {
+                return s;
+            }
+            else if ((p = operand.AsPredicate()) != null)
+            {
+                return Build(p);
+            }
+            else if ((sqlQueryStructure = operand.AsCosmosQueryStructure()) is not null
```
seantleonard, line 193: When would `sqlQueryStructure` not be of type `CosmosQueryStructure` when this code executes?

config-generators/cosmosdb_nosql-commands.txt
```diff
@@ -1,33 +1,48 @@
 init --config "dab-config.CosmosDb_NoSql.json" --database-type "cosmosdb_nosql" --connection-string "AccountEndpoint=https://localhost:8081/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==" --cosmosdb_nosql-database "graphqldb" --cosmosdb_nosql-container "planet" --graphql-schema "schema.gql" --host-mode Development --cors-origin "http://localhost:5000"
-add Planet --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.planet" --permissions "anonymous:create,read,update,delete" --graphql "Planet:Planets"
-update Planet --config "dab-config.CosmosDb_NoSql.json" --permissions "anonymous:read" --fields.include "*"
-update Planet --config "dab-config.CosmosDb_NoSql.json" --permissions "authenticated:create,read,update,delete"
-update Planet --config "dab-config.CosmosDb_NoSql.json" --permissions "limited-read-role:read"
+add PlanetAlias --config "dab-config.CosmosDb_NoSql.json" --source "graphqldb.planet" --permissions "anonymous:create,read,update,delete" --graphql "Planet:Planets"
```
seantleonard, line 2: confused why tests and sample configs differ in entity naming. Same say "PlanetAlias" and others say "Planet"

seantleonard (changes requested):

src/Core/Resolvers/AuthorizationPolicyHelpers.cs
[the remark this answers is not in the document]
sourabh1007: Added this as utility function

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -65,6 +80,130 @@ public CosmosSqlMetadataProvider(RuntimeConfigProvider runtimeConfigProvider, IF
...
+
+            Dictionary<string, ObjectTypeDefinitionNode> schemaDefinitions = new();
+            foreach (ObjectTypeDefinitionNode typeDefinition in GraphQLSchemaRoot.Definitions)
+            {
+                schemaDefinitions.Add(typeDefinition.Name.Value, typeDefinition);
+            }
+
+            foreach (IDefinitionNode typeDefinition in GraphQLSchemaRoot.Definitions)
+            {
+                if (typeDefinition is ObjectTypeDefinitionNode node)
```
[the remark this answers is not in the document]
sourabh1007: Added documentation to explain this.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -65,6 +80,130 @@ public CosmosSqlMetadataProvider(RuntimeConfigProvider runtimeConfigProvider, IF
...
+                            });
+
+                        ProcessSchema(node.Fields, schemaDefinitions, GraphQLNaming.COSMOSDB_CONTAINER_DEFAULT_ALIAS, tableCounter);
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Once container is found, it will traverse the fields and inner fields to get the paths for each entity.
```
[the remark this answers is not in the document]
sourabh1007: Added example in doc

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -65,6 +80,130 @@ public CosmosSqlMetadataProvider(RuntimeConfigProvider runtimeConfigProvider, IF
...
+        /// 1. If the entity is not in the runtime config, skip it.
+        /// 2. If the field is an array type, we need to add an alias to the table because we need to a JOIN to the same table.
+        /// 3. Create a new EntityDbPolicyCosmosModel object and add it to the EntityWithJoins dictionary.
+        /// 4. Check if we get previous entity with join information, if yes append it to the current entity also
+        /// 5. Recursively call this function, to process the schema
+        /// </summary>
+        /// <param name="fields"></param>
+        /// <param name="schemaDocument"></param>
+        /// <param name="currentPath"></param>
+        /// <param name="previousEntity"></param>
```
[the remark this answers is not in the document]
sourabh1007: This function is recursive call to traverse the schema. (mentioned in doc).

src/Core/Resolvers/CosmosQueryStructure.cs
```diff
@@ -138,6 +138,17 @@ private void Init(IDictionary<string, object?> queryParams)
                 Container = MetadataProvider.GetDatabaseObjectName(entityName);
             }
 
+            HttpContext httpContext = GraphQLFilterParser.GetHttpContextFromMiddlewareContext(_context);
+            if (httpContext is not null)
```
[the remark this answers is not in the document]
sourabh1007: it is already handled in `GraphQLFilterParser.GetHttpContextFromMiddlewareContext`

src/Core/Resolvers/CosmosQueryBuilder.cs
```diff
@@ -153,29 +164,56 @@ protected override string Build(Predicate? predicate)
...
-
-                joinBuilder.Append($" JOIN {structure.TableAlias} IN {structure.DbObject.FullName}");
-                tableNames.Add(structure.DbObject);
+                throw new ArgumentNullException(nameof(operand));
             }
 
-            return joinBuilder.ToString();
+            Column? c;
+            string? s;
+            Predicate? p;
```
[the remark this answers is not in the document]
sourabh1007: I believe it was existing code, I just copied it here from `BaseSqlQueryBuilder.cs`. Anyway I have renamed the variables.

src/Core/Parsers/EdmModelBuilder.cs
```diff
@@ -37,6 +38,36 @@ public EdmModelBuilder BuildModel(ISqlMetadataProvider sqlMetadataProvider)
...
+        private EdmModelBuilder BuildEdmModelsForCosmos(DocumentNode graphQLSchemaRoot)
+        {
+            EdmEntityContainer container = new(DEFAULT_NAMESPACE, DEFAULT_CONTAINER_NAME);
+
+            foreach (ObjectTypeDefinitionNode typeDefinition in graphQLSchemaRoot.Definitions)
+            {
+                EdmEntityType edmEntity = new(DEFAULT_NAMESPACE, typeDefinition.Name.Value);
+                foreach (FieldDefinitionNode field in typeDefinition.Fields)
+                {
+                    edmEntity.AddStructuralProperty(field.Name.Value, EdmPrimitiveTypeKind.String, false);
```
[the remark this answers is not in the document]
sourabh1007: no added function to do this conversion.

src/Core/Parsers/ODataASTCosmosVisitor.cs
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.OData.UriParser;
+
+/// <summary>
+/// This class is a visitor for an AST generated when parsing a $filter query string
+/// with the OData Uri Parser for Cosmos DB Db Policies
+/// </summary>
```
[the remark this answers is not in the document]
sourabh1007: Right now, only difference is, Cosmos Visitor is simple, just need "prefix" information to get appended to column,  but SQL visitor needs more information.
Second thing, tomorrow, you might want to handle different other kind of queries differently for cosmos and sql, this would give you this flexibility.

src/Core/Models/GraphQLFilterParsers.cs
sourabh1007: tried to mention it.

src/Core/Resolvers/AuthorizationPolicyHelpers.cs
sourabh1007: 1.nothing is modified, it returns OData filter clause.
2. I am not sure what example, you are looking for, it totally depends on the postProcessCallback

src/Service.Tests/CosmosTests/QueryTests.cs
sourabh1007: it gives you wrong impression that, you can query entity directly which is not true, in case of CosmosDB. In case of CosmosDB, you can query only at container level entity.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
sourabh1007: added.

src/Core/Resolvers/CosmosQueryBuilder.cs
sourabh1007: It is a safety similar to what we have in `BaseSqlQueryBuilder`

src/Service.Tests/dab-config.CosmosDb_NoSql.json
sourabh1007: Yes because Alias is only applicable on "Model" and for Cosmos DB only container level entity can be a model. Hence, putting alias for other entities, can be confusing in test, for Cosmos DB perspective. 
Hence, I modified the tests and config to reflect the same intention.

config-generators/cosmosdb_nosql-commands.txt
sourabh1007: it is same, can you please highlight, which config is exactly different.

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForCosmos.verified.txt
sourabh1007: this permission was not getting used. Hence removed it https://github.com/Azure/data-api-builder/blob/072c910133d6b86360072aaf3554bfade42da4ac/src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForCosmos.verified.txt#L199

config-generators/cosmosdb_nosql-commands.txt
sourabh1007: here "Planet" is renamed to "PlanetAlias" to make sure aliasing is not broken.

sourabh1007:
> You mention
> 
> > [Breaking Change: Need Documentation Update]
> 
> Please describe how this affects customers upgrading from earlier versions of DAB.
> 
> 1. What changes do they need to make in their configs/apps to ensure transition between dab versions?
>    
>    * Depending on how this impacts customers, we may not be able to merge right away without mitigating the breaking change.
>      
>      * e.g. Customers should be able to upgrade DAB versions without error and must opt-in to this specific change by use of feature flag.
>      * See documentation about `@model(name:"EntityNameInConfig")` https://learn.microsoft.com/azure/data-api-builder/database-specific-features#user-provided-graphql-schema
> 2. Does upgrading to this version of DAB (which includes this PR's changes), do old configs cause DAB startup to fail? Do old configs cause incorrect results to be returned?

1. Customer should be able to upgrade without any error. Old configs will work as it is. Customer should not see any problem with that.
2. We want to publish new recommendation for customers, that is, they should avoid having multiple `@model` in one schema. `@model` should represent container level entity only and other entities should be represented as `type`.
Why?
a) Query Definitions are generated for all the models which gives a wrong impression to customer that they can query with any level of entity. which is not true. in Cosmos DB, you can query only at container level.
b) As part of this PR, I added a logic which loads all the entities in memory and pre-generate JOINs for DB Policy defined, considering `@model` entity as container level entity. 
Having multiple models in the schema, will make this logic to run multiple times, as a side effect customer may experience high loading time.

cc @sajeetharan

sajeetharan (approved):
LGTM

sourabh1007:
Synced up with @sajeetharan, It is not a breaking change and it is already documented that, container level entity should be model.

seantleonard (approved):

Issue #2064, opened 2024-02-29
[Bug]: Unable to hot reload runtime section since file is being used by another process


Aniruddh25:
### What happened?

With dab 0.10.23, I tested hot reload of the runtime config section. 

I changed the _runtime.rest.path_ value from "/api" to "/rest".

Was expecting the route changed to a different prefix. Instead, I saw this message in the logs and the new path route couldn't be found on querying.

<img width="959" alt="image" src="https://github.com/Azure/data-api-builder/assets/3513779/5e94f969-1e3f-4a23-8ab6-80238b837de7">

### Version

0.10.23

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

REST

### Relevant log output

```Text
Loading config file from dab-config.json.
Unable to hot reload configuration file due to The process cannot access the file 'dab-config.json' because it is being used by another process.
Loading config file from dab-config.json.
Unable to hot reload configuration file due to The process cannot access the file 'dab-config.json' because it is being used by another process.
Loading config file from dab-config.json.

....

 Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker[3]
      Route matched with {action = "Find", controller = "Rest"}. Executing controller action with signature System.Threading.Tasks.Task`1[Microsoft.AspNetCore.Mvc.IActionResult] Find(System.String) on controller Azure.DataApiBuilder.Service.Controllers.RestController (Azure.DataApiBuilder.Service).
fail: Azure.DataApiBuilder.Service.Controllers.RestController[0]
      b16e74da-4022-4880-a2c6-50e1cce3c8c6 Error handling REST request.
      Azure.DataApiBuilder.Service.Exceptions.DataApiBuilderException: Invalid Path for route: rest/Book.
         at Azure.DataApiBuilder.Core.Services.RestService.GetRouteAfterPathBase(String route)
         at Azure.DataApiBuilder.Service.Controllers.RestController.HandleOperation(String route, EntityActionOperation operationType)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

seantleonard:
This task needs resolution in .12rc milestone. not moving.

aaronburtle:
Some additional context as to what is happening here:

[...]

PR #2108, opened 2024-03-14
Only instantiate the file watcher in a hot reloadable scenario

aaronburtle:
## Why make this change?

Related to https://github.com/Azure/data-api-builder/issues/2064

We are getting error messaging associated with making edits to the config file due to a file watcher that is constantly looking for changes to the config file as part of the initial pieces of hot reload. Until hot reload is ready this is not error messaging that we want to be made visible.

## What is this change?

Add a function to the `runtimeconfig` that checks if we are in a hot reloadable scenario. For now, this will always return false, and when we want to make the hot reload feature available, then the logic of this function can change to accommodate those updates.

## How was this tested?

Run manually to verify we are not file watching.

NOTE: testing for hot reload will need to be turned back on before we enable it as a feature, we have an issue to track the unit tests here: https://github.com/Azure/data-api-builder/issues/2109

## Sample Request(s)

No request is needed, simply start dab, make a change to the config file and save it, you will no longer see the error messages.

To verify the differences in the error messages and validate that they are consistent and what we expect, you can do the following:

1. Start dab (using the current version without this PR's changes) as normal, noting which config file you are using and its file location.
2. Make a rest request using the path from the config during startup (ie: "path": "/api") `https://localhost:5001/api/...`
3. While dab is running, open that same config file in an editor that will only hold a write lock while saving, such as Notepad++ and change the path (ie: "path": "/rest")
4. Save your changes
5. Note the error messages that indicate that hot reloading is not possible due to the file being in use by another process
6. Make a rest request using the new path from the saved config file `https://localhost:5001/rest/...`
7. Note that hot reloading did in fact work and your rest request works with the new path
8. Open the config file in another editor which will hold the write lock while the file is open, such as Visual Studio Code and change the path (ie: change it back to "path": "/api")
9. Note the many error messages that indicate hot reloading is not possible due to the file being in use by another process
10. Make a rest request using the newest path from the changes you just made. `https://localhost:5001/api/...`
11. Note that hot reloading can not work since the write lock was never relinquished by the editor.

### DIFF
--- a/src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
+++ b/src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
@@ -65,8 +65,11 @@ namespace Azure.DataApiBuilder.Service.Tests.Unittests
         /// Use the file system (not mocked) to create a hot reload
         /// scenario of the REST runtime options and verify that we
         /// correctly hot reload those options.
+        /// NOTE: This test is ignored until we have the possibility of turning
+        /// on hot reload.
         /// </summary>
         [TestMethod]
+        [Ignore]
         public void HotReloadConfigRestRuntimeOptions()
         {
             // Arrange
--- a/src/Config/ObjectModel/RuntimeConfig.cs
+++ b/src/Config/ObjectModel/RuntimeConfig.cs
@@ -440,4 +440,17 @@ public record RuntimeConfig
         CosmosDataSourceUsed = _dataSourceNameToDataSource.Values.Any
             (x => x.DatabaseType is DatabaseType.CosmosDB_NoSQL);
     }
+
+    /// <summary>
+    /// Handles the logic for determining if we are in a scenario where hot reload is possible.
+    /// Hot reload is currently not available, and so this will always return false. When hot reload
+    /// becomes an available feature this logic will change to reflect the correct state based on
+    /// the state of the runtime config and any other relevant factors.
+    /// </summary>
+    /// <returns>True in a scenario that support hot reload, false otherwise.</returns>
+    public static bool IsHotReloadable()
+    {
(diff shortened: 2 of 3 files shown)


src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -65,8 +65,11 @@ HostMode mode
         /// Use the file system (not mocked) to create a hot reload
         /// scenario of the REST runtime options and verify that we
         /// correctly hot reload those options.
+        /// NOTE: This test is ignored until we have the possibility of turning
```
seantleonard, line 68: On one of your hot reload issues (or a new one that tracks enabling hot reload with a feature switch), point to this file::line::testmethod to ensure that this test is re-enabled when appropriate.

seantleonard (approved):
Thanks for updating this to ensure we don't experience the file contention error when not using hot reload yet. Small favor to ask before you merge in:
- Were you able to get consistent repro of the error message you get before this change?
- Can you provide those repro steps in the PR description so that we can check locally to confirm that the error no longer appears?

aaronburtle:
> * Were you able to get consistent repro of the error message you get before this change?
> * Can you provide those repro steps in the PR description so that we can check locally to confirm that the error no longer appears?

Yes, I was, and I will add those steps.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
aaronburtle: issue here https://github.com/Azure/data-api-builder/issues/2109

abhishekkumams (approved):
Thanks for the quick fix.

PR #2112, opened 2024-03-18
Making datasourcename mandatory during executequeryasync for multi-db scenario.

rohkhann:
## Why make this change?
During the call to executeQuery, if DataSource Name is not specified, the query will execute against the default dB. Currently this option is an optional argument. This can lead to caller not sending the DataSource Name as the method does not require it. 

If used incorrectly, it will then execute against the default db which can lead to errors during multi-db scenario. For example, in SQLMetadataProvider, we have the queryexecutor call executequeryAsync for readonly columns. Currently, that does not specify the datasource as it is not required by the method. This means that will always execute against the default db, even though the datasourceName property is available on the SQLMetadataprovider which tells you which db to execute this query against.

 currently we execute the query against the default db (this was done to maintain backward compatibility for single source scenarios). In this change, we make it mandatory to specify the datasource name during a call to executequeryAsync. String.Empty or null can be sent if query is to be executed against default db.

## What is this change?
1. Change the executeAsync call to explicitly require for the the dbname. The executeAsync in the queryExecutor is the final call made to actually execute the query against the underlying db. By this time in both rest and GQL, we have either determined the dbname or gone with the default db name. 
2. Updates to all callers to specify dbname explicitly.

## How was this tested?
1. Existing test cases are updated to account for mandatory passing of the datasourceName.
3. Integration test done.

### DIFF
--- a/src/Cli.Tests/ModuleInitializer.cs
+++ b/src/Cli.Tests/ModuleInitializer.cs
@@ -51,6 +51,8 @@ static class ModuleInitializer
         VerifierSettings.IgnoreMember<RuntimeConfig>(config => config.Schema);
         // Ignore the message as that's not serialized in our config file anyway.
         VerifierSettings.IgnoreMember<DataSource>(dataSource => dataSource.DatabaseTypeNotSupportedMessage);
+        // Ignore DefaultDataSourceName as that's not serialized in our config file.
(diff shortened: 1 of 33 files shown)


aaronburtle (approved):
Looks good to me!

rusamant (approved):

Issue #1951, opened 2024-01-03
Separate engine workflow depending on whether nested mutations are on/off


ayush3797:
The purpose of the issue is to track the efforts needed for 
 a) Consuming the feature flag fields in the config file 
 b) Adding different code paths in the engine based on whether the feature flag is enabled or disabled

JerryNixon:
The content of this issue has no grammatical errors! ;)

![image](https://github.com/Azure/data-api-builder/assets/1749983/54fa387c-8c26-49d1-9771-c8509224ea70)

JerryNixon:
The reasoning around this issue includes giving the ability for developers to avoid any startup cost associated with nested mutations. Developers can already change read/write permissions of entities which can effectively disable mutations. The nuance here is that disabling them globally can mean the engine doesn't have to compute the schema at startup. Please refer to #1950

severussundar:
> The reasoning around this issue includes giving the ability for developers to avoid any startup cost associated with nested mutations. Developers can already change read/write permissions of entities which can effectively disable mutations. The nuance here is that disabling them globally can mean the engine doesn't have to compute the schema at startup. Please refer to #1950

Thank you for adding this. Updated the description.

When the feature is disabled, the schema will be generated the way it is right now; thus avoiding additional computations, memory utilized etc.

seantleonard:
Follows up the work from #1950

PR #1902, opened 2023-11-28
Multiple-create: Schema generation

ayush3797:
## Why make this change?
Currently, for a create mutation (or any graphql mutation essentially), we can only do insertion/mutation in one table and not in another table which is related to this table via some relationship (either defined in config or in the database). This PR aims to:
1. Extend the functionality of the existing create mutations (which are basically point insertions) to allow insertions in related table.
2. To generate additional `createMultiple` mutations which will allow multiple nested insertions starting from the top-level entity.

**Note:** _All changes made in this PR are specific to MsSql. Nothing changes for the other database flavors._

## Quick glossary:
1. **ObjectTypeDefinitionNode:** Represents a table/view/stored-procedure. For a table, this contains all the column fields which belong to the table and the relationship fields which are defined in this table's entity configuration in the config file.
2. **InputObjectTypeDefinitionNode:** Represents an input object type we generate for mutations.
## What is this change?
1. **We no more ignore relationship fields while generating the input types for a create mutation**: Till date, we were only considering the column fields in an object (of type `ObjectTypeDefinitionNode`) to generate the input type (of type `InputObjectTypeDefinitionNode`) for a table entity. But to support nested insertion, we would now also iterate over relationship fields to generate the input type for a **create**/**createMultiple** mutation.
2. **Addition of linking entities at the backend:** To support nested insertions in tables which have a relationship with cardinality N:N, the user can provide a linking table with the source defined in `linking.object` field which we need the user to provide. We need to generate an object type (of type `ObjectTypeDefinitionNode`) for this linking entity. This object type is later be used to generate the input object type for the linked tables.
Additional details about linking entities:

   -> **A boolean property `IsLinkingEntity` is added with a _default value of `false`_** to the `Entity` record. This ensures that we are backwards compatible, _and all the entities provided in the config are not considered as linking entities. For all the linking entities, this boolean property will be set to `true`._
   -> For a linking entity, **_the GraphQL and REST endpoints are disabled (set to `false`) by default_**. This ensures that we don't accidentally expose the linking entity to the user.
[...]

PR #2116, opened 2024-03-19
Enable or Disable Multiple Create operation based on feature flag value

severussundar:
## Why make this change?

- Closes https://github.com/Azure/data-api-builder/issues/1951
  - PR https://github.com/Azure/data-api-builder/pull/1983, https://github.com/Azure/data-api-builder/pull/2103 add CLI options to enable or disable multiple mutation/multiple create operation through CLI. With the changes introduced in the mentioned PRs, the configuration properties successfully gets written to the config file. Also, during deserialization, the properties are read and the `MultipleMutationOptions`, `MultipleCreateOptions`, `GraphQLRuntimeOptions` objects are created accordingly.
  -  The above-mentioned PRs do not introduce any change in DAB engine behavior depending on the configuration property values.
  - This PR introduces changes to read these fields and enable/disable multiple create operation depending on whether the feature is enabled/disabled through the config file. This is achieved by introducing behavior changes in the schema generation.

## What is this change?
- This PR builds on top of a) Schema generation PR https://github.com/Azure/data-api-builder/pull/1902  b) Rename nested-create to multiple create PR https://github.com/Azure/data-api-builder/pull/2103
- When multiple create operation is disabled, 
 > i) Fields belonging to the related entities are not created in the input object type are not created.
 > ii) Many type multiple create mutation nodes (ex:  `createbooks`, `createpeople_multiple` ) are not created. 
 > iii) ReferencingField directive is not applied on relationship fields, so they continue to remain required fields for the create mutation operation.
 > iv) Entities for linking objects are not created as they are relevant only in the context of multiple create operations.

## How was this tested?

- [x] Unit Tests and Integration Tests 
- [x] Manual Tests 

**Note:** At the moment, multiple create operation is disabled in the config file generated for integration tests. This is because of the plan to merge in the Schema generation, AuthZ/N branches separately to the main branch. With just these 2 PRs, a multiple create operation will fail, hence, the disabling multiple create operation.  At the moment, tests that perform validations specific to multiple create feature enable it by i) updating the runtime object (or) ii) creating a custom config in which the operation is enabled.   

#### Related entity fields are created in the input object type
image

#### Multiple type create operation is created in addition to point create operation
image

#### Querying related entities continue to work successfully
image

#### Only fields belonging to the given entity are created in the input object type
image

#### Only fields belonging to the given entity are created in the input object type
image


src/Service.GraphQLBuilder/Mutations/MutationBuilder.cs
```diff
@@ -37,7 +37,8 @@ public static class MutationBuilder
             Dictionary<string, DatabaseType> databaseTypes,
             RuntimeEntities entities,
             Dictionary<string, EntityMetadata>? entityPermissionsMap = null,
-            Dictionary<string, DatabaseObject>? dbObjects = null)
+            Dictionary<string, DatabaseObject>? dbObjects = null,
+            bool isMultipleCreateOperationEnabled = false)
```
ayush3797, line 41: Add param summary

src/Service.GraphQLBuilder/Mutations/MutationBuilder.cs
```diff
@@ -120,7 +121,8 @@ public static class MutationBuilder
             DatabaseType databaseType,
             RuntimeEntities entities,
             List<FieldDefinitionNode> mutationFields,
-            string returnEntityName
+            string returnEntityName,
+            bool isMultipleCreateOperationEnabled = false
```
ayush3797, line 125: nit: param summary

src/Service.GraphQLBuilder/Mutations/MutationBuilder.cs
```diff
@@ -130,7 +132,7 @@ string returnEntityName
                 {
                     case EntityActionOperation.Create:
                         // Get the create one/many fields for the create mutation.
-                        IEnumerable<FieldDefinitionNode> createMutationNodes = CreateMutationBuilder.Build(name, inputs, objectTypeDefinitionNode, root, databaseType, entities, dbEntityName, returnEntityName, rolesAllowedForMutation);
+                        IEnumerable<FieldDefinitionNode> createMutationNodes = CreateMutationBuilder.Build(name, inputs, objectTypeDefinitionNode, root, databaseType, entities, dbEntityName, returnEntityName, rolesAllowedForMutation, isMultipleCreateOperationEnabled);
```
ayush3797, line 135: nit: split long line

src/Service.GraphQLBuilder/Mutations/MutationBuilder.cs
ayush3797: Please do the same in all other places.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -539,21 +557,22 @@ public static NameNode GenerateInputTypeName(string typeName)
 
             createMutationNodes.Add(createOneNode);
 
-            if (IsRelationalDb(databaseType) && DoesRelationalDBSupportMultipleCreate(databaseType))
+            // Multiple create node is created in the schema only when multiple create operation is enabled.
+            if (isMultipleCreateOperationEnabled && IsRelationalDb(databaseType) && DoesRelationalDBSupportMultipleCreate(databaseType))
```
ayush3797, line 561: Rather than using independent conditions: `isMultipleCreateOperationEnabled &&  DoesRelationalDBSupportMultipleCreate(databaseType)`, can we create one method which takes care of everything? Here we have two variables:

1. Database type
2. Feature-flag: set to true/false.

Obviously, I don't anticipate multiple-mutations to be supported for multiple database scenario soon. So both the above two things are known to RuntimeConfig. So maybe that might be a good place to have this method?

On a side note, I see that the RuntimeConfig.Entities are also passed on throughout all these method calls. So maybe if we can just pass the runtime config through these methods, we can take care of both the things.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2030,6 +2030,155 @@ public async Task ValidateErrorMessageForMutationWithoutReadPermission()
...
+                Policy: new());
+
+            EntityAction readAction = new(
+                Action: EntityActionOperation.Read,
+                Fields: null,
+                Policy: new());
+
+            EntityPermission[] permissions = new[] { new EntityPermission(Role: AuthorizationResolver.ROLE_ANONYMOUS, Actions: new[] { readAction, createAction }) };
+
+            EntityRelationship bookRelationship = new(Cardinality: Cardinality.One, TargetEntity: "Publisher", SourceFields: new string[] { }, TargetFields: new string[] { }, LinkingObject: null, LinkingSourceFields: null, LinkingTargetFields: null);
```
ayush3797, line 2068: nit: split long line

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2030,6 +2030,155 @@ public async Task ValidateErrorMessageForMutationWithoutReadPermission()
...
+                              Mappings: null);
+
+            string bookEntityName = "Book";
+
+            Dictionary<string, Entity> entityMap = new()
+            {
+                { bookEntityName, bookEntity }
+            };
+
+            EntityRelationship publisherRelationship = new(Cardinality: Cardinality.Many, TargetEntity: "Book", SourceFields: new string[] { }, TargetFields: new string[] { }, LinkingObject: null, LinkingSourceFields: null, LinkingTargetFields: null);
```
ayush3797, line 2084: nit: split long line

src/Service.Tests/GraphQLBuilder/MultipleMutationBuilderTests.cs
```diff
@@ -349,9 +350,28 @@ public static async Task InitializeAsync()
...
+            {
+                Runtime = new RuntimeOptions(Rest: runtimeConfig.Runtime.Rest,
+                                                                GraphQL: new GraphQLRuntimeOptions(MultipleMutationOptions: new MultipleMutationOptions(new MultipleCreateOptions(enabled: true))),
+                                                                Host: runtimeConfig.Runtime.Host,
+                                                                BaseRoute: runtimeConfig.Runtime.BaseRoute,
+                                                                Telemetry: runtimeConfig.Runtime.Telemetry,
+                                                                Cache: runtimeConfig.Runtime.Cache)
+            };
+
+            System.IO.Abstractions.TestingHelpers.MockFileSystem fileSystem = new();
```
ayush3797, line 370: It would be helpful to add a comment as to what the next few lines of code are doing.

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -60,6 +61,18 @@ public class GraphQLSchemaCreator
         {
             RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
 
+            if (runtimeConfig.Runtime is not null &&
```
ayush3797, line 64: Can all this code can be moved to `RuntimeConfig` (or any other appropriate place) and reused wherever required?

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -245,8 +262,14 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
 
             // Create ObjectTypeDefinitionNode for linking entities. These object definitions are not exposed in the schema
             // but are used to generate the object definitions of directional linking entities for (source, target) and (target, source) entities.
-            Dictionary<string, ObjectTypeDefinitionNode> linkingObjectTypes = GenerateObjectDefinitionsForLinkingEntities();
-            GenerateSourceTargetLinkingObjectDefinitions(objectTypes, linkingObjectTypes);
+            // However, ObjectTypeDefinitionNode for linking entities are need only for multiple create operation. So, creating these only when multiple create operations are
+            // enabled.
+            if (_isMultipleCreateOperationEnabled)
+            {
+
```
ayush3797, line 269: nit: remove blank line

ayush3797: I don't think this needs to be put under a conditional. The conditional logic is already there in SqlMetadataProvider to  create linking entities when multiple create is enabled: https://github.com/Azure/data-api-builder/blob/814bc34f8e6f3a79e194af62f4e71a1137e5482b/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs#L761

Trying to put as less conditionals/branches as we can - will make life easier when we/other devs add support for other db types.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2030,6 +2030,155 @@ public async Task ValidateErrorMessageForMutationWithoutReadPermission()
...
+        /// Multiple mutation operations are disabled through the configuration properties.
+        ///
+        /// 1. Executing the following multiple create mutation operations,
+        ///     a. Point multiple create operation
+        ///     b. Many type multiple create operation
+        ///    should result in failure as both these operations are disabled.
+        ///
+        /// 2. Executing a create mutation operation should succeed.
+        /// 
+        /// </summary>
```
ayush3797, line 2043:
```suggestion
       /// <summary>
/// Test to validate that when multiple-create is disabled:
/// 1. Including a relationship field in the input for create mutation for an entity returns an exception as when multiple mutations are disabled, we don't add fields for relationships in the input type and hence users should not be able to do insertion in the related entities.
/// 2. Excluding all the relationship fields i.e.performing insertion in just the top-level entity executes successfully.
/// </summary>
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2030,6 +2030,155 @@ public async Task ValidateErrorMessageForMutationWithoutReadPermission()
...
+        [TestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        public async Task ValidateMultipleCreateAndCreateMutationWhenMultipleCreateOperationIsDisabled()
+        {
+            // Multiple create operations are disabled.
+            GraphQLRuntimeOptions graphqlOptions = new(Enabled: true, MultipleMutationOptions: new(new(enabled: false)));
+
+            RestRuntimeOptions restRuntimeOptions = new(Enabled: false);
+
+            DataSource dataSource = new(DatabaseType.MSSQL,
```
ayush3797, line 2053: super-nit: Will be more clear to either write all of the params together or in separate lines.

src/Service.Tests/GraphQLBuilder/MultipleMutationBuilderTests.cs
severussundar: Added

src/Core/Services/GraphQLSchemaCreator.cs
severussundar: Moved this logic to a helper method within `RuntimeConfig`

src/Core/Services/GraphQLSchemaCreator.cs
severussundar: ObjectTypeDefinitionNode objects are created for linking tables only when multiple create is enabled. Adding a conditional in `GraphQLSchemaCreator` makes it easier to follow rather than letting the flow go through and a condition failing at a later stage preventing the creation of  ObjectTypeDefinitionNode objects.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
severussundar: Ack. Evaluating this.

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -1079,7 +1095,7 @@ string[] expectedNames
             int totalExpectedMutations = 0;
             foreach ((_, DatabaseType dbType) in entityNameToDatabaseType)
             {
-                if (GraphQLUtils.DoesRelationalDBSupportMultipleCreate(dbType))
+                if (isMultipleCreateOperationEnabled && GraphQLUtils.DoesRelationalDBSupportMultipleCreate(dbType))
```
seantleonard, line 1098: nit: once `GraphQLUtils.DoesRelationalDBSupportMultipleCreate(dbType)` becomes implementation detail of `isMultipleCreateOperationEnabled` this can be modified.

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -1012,34 +1012,49 @@ public static ObjectTypeDefinitionNode GetMutationNode(DocumentNode mutationRoot
...
+            DisplayName = "Mutation name and description validation for multiple entities with singular plural not defined - Multiple Create Operation disabled")]
         public void ValidateMutationsAreCreatedWithRightName(
             string gql,
             string[] entityNames,
             string[] singularNames,
             string[] pluralNames,
-            string[] expectedNames
+            string[] expectedNames,
+            string[] expectedCreateMultipleMutationNames,
+            bool isMultipleCreateOperationEnabled
```
seantleonard, line 1057: i understand there would be code duplication if you had a separate test was explicitly for multiple mutations. The benefits outweigh the code duplication in my opinion:
- easer to read the DataRows with way fewer parameters
- fewer conditionals in test logic that make this harder to follow
- don't have to confuse test logic that explicitly tests point mutation schema creation.

seantleonard (approved):
pending some suggestions about combining conditionals behind "ismultiplemutationenabled" flag. and nit about test separation. Otherwise looks good and approving so you can get those changes and merge on monday your time.

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
severussundar: I considered these factors but somehow I was leaning more towards performing validations for scenarios with feature flag on/off in the same test.

I agree with the reasoning - I've split this in two different tests. One for validating scenarios with multiple create enabled, the other with multiple create disabled.

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
severussundar: I've re-written this

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2030,6 +2030,165 @@ public async Task ValidateErrorMessageForMutationWithoutReadPermission()
...
+                Assert.IsNotNull(mutationResponse);
+                SqlTestHelper.TestForErrorInGraphQLResponse(mutationResponse.ToString(),
+                                                            message: "The field `createbooks` does not exist on the type `Mutation`.");
+
+                string pointCreateOperation = @"mutation createbook{
+                                                            createbook(item: { title: ""Book #1"", publisher_id: 1234 }) {
+                                                                title
+                                                                publisher_id
+                                                            }
+                                                        }";
```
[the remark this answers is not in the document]
severussundar: Thanks for the suggestion.

I've added two additional tests: 
1. Point create mutation with missing `publisher_id` field with multiple create operation disabled. This should be caught by hotchocolate since a required field is missing. 
2. Point create mutation with missing `publisher_id` field with multiple create operation enabled. This should also lead to a failure. But, the error gets thrown from the database layer since there is no sufficient data to successfully execute the insert statement.  
Note: Once request validation PR gets merged, these types of requests will fail at the request validation stage (the flow will not reach till the database layer). @ayush3797, the error message that is checked in the test will have to be updated when checking-in the request validation PR.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
severussundar: I've introduced a method called `IsMultipleCreateOperationSupportedAndEnabled()` in `RuntimeConfig` class that essentially combines the checks: a) Whether multiple create is enabled through the config file and b) Whether the database type supports multiple create operation. 

I'm still continuing to pass just the `RuntimeEntities` and `IsMultipleCreateOperationSupportedAndEnabled` parameters instead of the complete `Runtime` object as a parameter - this is because, only these 2 fields from the `RuntimeConfig` object are required. The other fields are not necessary, and hence we don't necessarily need the entire `RuntimeConfig` object.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -270,8 +269,8 @@ private static bool DoesFieldHaveReferencingFieldDirective(FieldDefinitionNode f
         /// <param name="name">Name of the field.</param>
         /// <param name="fieldDefinition">Field definition.</param>
         /// <param name="databaseType">Database type</param>
-        /// <param name="isMultipleCreateOperationEnabled">Indicates whether multiple create operation is enabled</param>
-        private static InputValueDefinitionNode GenerateScalarInputType(NameNode name, FieldDefinitionNode fieldDefinition, DatabaseType databaseType, bool isMultipleCreateOperationEnabled = false)
+        /// <param name="IsMultipleCreateOperationSupportedAndEnabled">Indicates whether multiple create operation is enabled</param>
+        private static InputValueDefinitionNode GenerateScalarInputType(NameNode name, FieldDefinitionNode fieldDefinition, bool isMultipleCreateOperationSupportedAndEnabled = false)
```
seantleonard, line 273: naming

src/Service.Tests/GraphQLBuilder/Helpers/GraphQLTestHelpers.cs
```diff
@@ -43,6 +43,8 @@ public static class GraphQLTestHelpers
                     }
                     ";
 
+        public static HashSet<DatabaseType> RELATIONAL_DBS_SUPPORTING_MULTIPLE_CREATE = new() { DatabaseType.MSSQL };
```
seantleonard, line 46: this hashset is now here and runtimeconfig? is the duplication necessary?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2167,25 +2112,96 @@ public async Task ValidateMultipleCreateAndCreateMutationWhenMultipleCreateOpera
...
+                                                                                        queryName: "createbook",
+                                                                                        variables: null,
+                                                                                        clientRoleHeader: null);
 
                 string expectedResponse = @"{ ""title"":""Book #1"",""publisher_id"":1234}";
 
                 Assert.IsNotNull(mutationResponse);
                 SqlTestHelper.PerformTestEqualJsonStrings(expectedResponse, mutationResponse.ToString());
+
+                // When  a create multiple operation is enabled, the "publisher_id" field will generated as an optional field in the schema. But, when multiple create operation is disabled,
```
seantleonard, line 2137:
```suggestion
                // When  a create multiple operation is enabled, the "publisher_id" field will be generated as an optional field in the schema. But, when multiple create operation is disabled,
```

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -1007,14 +1007,15 @@ public static ObjectTypeDefinitionNode GetMutationNode(DocumentNode mutationRoot
         }
 
         /// <summary>
+        ///
```
seantleonard, line 1010:
```suggestion
```

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -1100,6 +1192,15 @@ string[] expectedNames
                 FieldDefinitionNode createMutation = mutation.Fields.First(f => f.Name.Value == expectedCreateMutationName);
                 Assert.AreEqual(expectedCreateMutationDescription, createMutation.Description.Value);
 
+                if (GraphQLTestHelpers.DoesRelationalDBSupportMultipleCreate(entityNameToDatabaseType[entityNames[i]]))
```
seantleonard, line 1195: why is this check needed when the mutation root is provided with: `IsMultipleCreateOperationSupportedAndEnabled: true)`?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -455,15 +459,20 @@ public static bool IsHotReloadable()
...
-    /// Helper function to check if multiple create option is enabled.
+    /// Helper method to check if multiple create option is supported and enabled.
+    /// 
+    /// Returns true when
+    /// 1. Multiple create operation is supported by the database type and
+    /// 2. Multiple create operation is enabled in the runtime config.
+    /// 
     /// </summary>
-    public bool IsMultipleCreateOptionEnabled()
+    public bool IsMultipleCreateOperationSupportedAndEnabled()
```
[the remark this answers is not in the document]
severussundar: Good catch..

Agree that the caller of the function only needs to worry about whether the result is `true`/`false`. It does not matter if it is false because of underlying db type not supporting the operation or the operation not being enabled in the config file.

I've updated all the function and variable names to `IsMultipleCreateEnabled`

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
severussundar: Not needed, removed it.

src/Service.Tests/GraphQLBuilder/Helpers/GraphQLTestHelpers.cs
severussundar: With the removal of conditional in `MutationBuilderTests`, this hashset is not needed anymore. Removed this.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -30,6 +30,10 @@ public record RuntimeConfig
     [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
     public bool SqlDataSourceUsed { get; private set; }
 
+    // HashSet of database types that support multiple create operation.
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
+    public static HashSet<DatabaseType> RELATIONAL_DBS_SUPPORTING_MULTIPLE_CREATE = new() { DatabaseType.MSSQL };
```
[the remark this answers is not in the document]
severussundar: Thanks for the suggestion.

I've added a new enum name `MultipleCreateSupportingDatabaseType` - purpose is to hold all the database types that support multiple create. 

Removed the hashset from `RuntimeConfig` class.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -82,8 +84,10 @@ public static class CreateMutationBuilder
             // we find that the input object has already been created for the entity.
             inputs.Add(input.Name, input);
 
-            // Generate fields for related entities only if multiple mutations are supported for the database flavor.
-            if (DoesRelationalDBSupportMultipleCreate(databaseType))
+            // Generate fields for related entities when
+            // 1. Multiple mutation operations are supported for the database type.
+            // 2. Multiple mutation operations are enabled.
+            if (isMultipleCreateOperationEnabled && DoesRelationalDBSupportMultipleCreate(databaseType))
```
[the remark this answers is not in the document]
severussundar: Since, https://github.com/Azure/data-api-builder/pull/2116#discussion_r1536631102 and this comment are about the same thing, we can have furhter discussion on that thread. 

Resolving this one.

src/Config/ObjectModel/MultipleCreateSupportingDatabaseType.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel
+{
+    public enum MultipleCreateSupportingDatabaseType
+    {
+        MSSQL
+    }
+
```
seantleonard, line 10:
```suggestion
```

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -466,9 +462,9 @@ public static bool IsHotReloadable()
     /// 2. Multiple create operation is enabled in the runtime config.
     /// 
     /// </summary>
-    public bool IsMultipleCreateOperationSupportedAndEnabled()
+    public bool IsMultipleCreateOperationEnabled()
     {
-        return (RELATIONAL_DBS_SUPPORTING_MULTIPLE_CREATE.Contains(DataSource.DatabaseType)) &&
+        return Enum.GetNames(typeof(MultipleCreateSupportingDatabaseType)).Any(x => x.Equals(DatabaseType.MSSQL.ToString(), StringComparison.OrdinalIgnoreCase)) &&
```
seantleonard, line 467: Can you elaborate on what is happening here? You're checking that `MultipleCreateSupportingDatabaseType` has `DatabaseType.MSSQL` defined.. which it does. This will always return true, even if the runtime config were set for DW/My/PG/Cosmos?

seantleonard (changes requested):
unclear on check for MultipleCreateSupportingDatabaseType

src/Config/ObjectModel/RuntimeConfig.cs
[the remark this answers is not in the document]
seantleonard: the new implementation requires parsing an enum and iterating through MultipleCreateSupportingDatabaseType every time `IsMulitpleCreateOperationEnabled()`... seems more expensive than to just check whether configured database == mssql

src/Config/ObjectModel/RuntimeConfig.cs
severussundar: Thanks for spotting this...Miss on my end.

Updated to `DataSource.DatabaseType.ToString()`

src/Config/ObjectModel/RuntimeConfig.cs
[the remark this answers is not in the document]
severussundar: In the future, when the feature is supported by more database types, the enum can be updated to accommodate the new database types that support the feature.

seantleonard (approved):

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
severussundar: Changing `CreateMutationBuilder.Build()`  to accept `RuntimeObject` is purely a refactor change. IMO, we don't need the complete `RuntimeConfig` object, as only `RuntimeEntities` and `IsMultipleCreateOperationEnabled` fields are relevant for schema generation; hence, we can continue to pass them separately. 

I've created the https://github.com/Azure/data-api-builder/issues/2132 for further evaluation/discussion on this.

Since, this is purely a refactor change, this can be picked up in a follow-up PR (if we make the decision to go ahead with the refactor). This need not hold this PR from merging. 

Resolving this thread

Issue #2115, opened 2024-03-19
[Bug]: Cannot filter with custom role (access forbidden...)


lionelschiepers:
### What happened?

Hi, I have a config file that is defining a custom role (let say role-a) but I can't filter anonymous entites when that role is used. I'm receiving this error: 
```
{
  "errors": [
    {
      "message": "Access forbidden to a field referenced in the filter.",
      "extensions": {
        "code": "AuthorizationCheckFailed"
      }
    }
  ]
}
```

To reproduce the problem, you have to query the server with X-MS-API-ROLE = role-a + a principal that defines that role. You receive that error when you are filtering an anonymous entity.

When I look at the code, the authenticated role derives from anonymous and a custom role derives from authenticated role.

I think that the problem is in AuthorizationResolver near line 135 in the method AreColumnsAllowedForOperation. When a custom role is used and that role is not defined at the entity level, the method should check the role the custom role is inheriting from, authenticated in this case.

The test should looks like:
```
        RoleMetadata? roleMetadata;
        if (!EntityPermissionsMap[entityName].RoleToOperationMap.TryGetValue(roleName, out roleMetadata) && roleMetadata is null)
        {
            if (ROLE_ANONYMOUS.Equals(roleName, StringComparison.OrdinalIgnoreCase) ||
                ROLE_AUTHENTICATED.Equals(roleName, StringComparison.OrdinalIgnoreCase))
            {
                return false;
            }
            // roleName is a custom role and derives from authenticated role. Let's try with authenticated role.
            if (!EntityPermissionsMap[entityName].RoleToOperationMap.TryGetValue(ROLE_AUTHENTICATED, out roleMetadata) && roleMetadata is null)
            {
                return false;
            }
        }
```
instead of 

```
        if (!EntityPermissionsMap[entityName].RoleToOperationMap.TryGetValue(roleName, out RoleMetadata? roleMetadata) && roleMetadata is null)
        {
            return false;
        }
```
 
a work around could be to define the custom role in all entities but it's not possible to do that with the cli. You can only set one single permission by entity.

another work around could be to update the EntityPermissionsMap collections to include all custom roles defined in the configuration file and derive them from the authenticated role like the authenticated role is copied from the anonymous role (method SetEntityPermissionMap in AuthorizationResolver near line 357 when calling CopyOverPermissionsFromAnonymousToAuthenticatedRole).

I tried to do a pull request but I don't have a write access to the repository.

Is there any chance to find a correction for this?

[...]

PR #2117, opened 2024-03-19
Enable filtering with custom roles

lionelschiepers:
## Why make this change?
This pull request is related to issue 2115 (https://github.com/Azure/data-api-builder/issues/2115) that reports that it's not
possible to filter entities anonymously accessible when the caller is using a custom role.

## What is this change?
When filtering is disallowed for a custom role because the role is not defined for that entity, it is then checked for the authenticated role the custom role inherits from. Please note that when no authenticated role is defined for an entity, it is automatically defined using the definition of the anonymous role.

## How was this tested?
I've tested in our live environment.

### DIFF
--- a/src/Core/Authorization/AuthorizationResolver.cs
+++ b/src/Core/Authorization/AuthorizationResolver.cs
@@ -136,7 +136,17 @@ public class AuthorizationResolver : IAuthorizationResolver
 
         if (!EntityPermissionsMap[entityName].RoleToOperationMap.TryGetValue(roleName, out RoleMetadata? roleMetadata) && roleMetadata is null)
         {
-            return false;
+            if (ROLE_ANONYMOUS.Equals(roleName, StringComparison.OrdinalIgnoreCase) ||
+                ROLE_AUTHENTICATED.Equals(roleName, StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            // roleName is a custom role and derives from authenticated role. Let's try with authenticated role.
+            if (!EntityPermissionsMap[entityName].RoleToOperationMap.TryGetValue(ROLE_AUTHENTICATED, out roleMetadata) && roleMetadata is null)
+            {
+                return false;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


seantleonard:
Thank you for your contribution, I will take a look soon. But in the meantime, please add tests that exercise this scenario that you are fixing.

JerryNixon:
@Aniruddh25 we should review this this semester.

JerryNixon:
+1

JerryNixon:
## This behavior described is *by design*, not a bug.

Here’s the key:

Data API builder **does not perform role inheritance**. Each role must be explicitly defined per entity in the configuration.

So when you set `X-MS-API-ROLE = role-a`, DAB looks for an explicit `role-a` permission block on that entity. If it doesn’t find one, it does *not* automatically inherit permissions from `authenticated` or `anonymous`. That’s why the filter authorization check fails.

The apparent inconsistency—being able to read fields but not filter—comes from how GraphQL and REST differ in execution paths. In the read operation, the engine evaluates “access to entity” (which passes through because of anonymous read permissions), but filtering checks column-level permission for the *active role*, which is undefined for `role-a`. Since DAB sees `role-a` but can’t find corresponding metadata, it correctly blocks the filter for security reasons.

Aniruddh25:
> Please note that when no authenticated role is defined for an entity, it is automatically defined using the definition of the anonymous role.

While this is true, the same rule does not apply to custom roles this `authenticated` and `anonymous` are special roles.

Aniruddh25:
As discussed in the issue here: https://github.com/Azure/data-api-builder/issues/2115 
this behavior is by design hence closing this PR.

But really appreciate and thank you for your deep investigation and contributions, @lionelschiepers!

Refer to this section of doc to add multiple permissions to the same entity, example:
`dab add Book --source dbo.Books --permissions "anonymous:read" --permissions "authenticated:create,read,update,delete"`
https://learn.microsoft.com/en-us/azure/data-api-builder/command-line/dab-add#--permissions

Issue #2030, opened 2024-02-09
[Bug]: Inconsistency with endCursor


pingu2k4:
### What happened?

With SQL as the backing DB, `endCursor` will not be null and will provide a value when `hasNextPage` is false. I would expect `endCursor` to be null here.

When using Cosmos as the DB, and when not encountering the bug in #2029, when `hasNextPage` is false, `endCursor` will also be null, as expected.

### Version

0.9.7+e560142426d1c080b9fd7b7fabff51a276f6bf61

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

GraphQL

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

seantleonard:
hi @pingu2k4 thanks for reporting, just to clarify, in this issue `hasNextPage` works but the `endCursor` is improperly populated for MSSQL?

pingu2k4:
Yes, thats correct, and thanks. :)

PR #2121, opened 2024-03-20
Ensure endCursor is NULL when hasNextPage is FALSE in GraphQL response

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2030

Closes https://github.com/Azure/data-api-builder/issues/2130

Fixes a bug where we would have a `non null` `endCursor` in GraphQL responses, despite `hasNextPage` being false. In other words, providing a bookmark to the next page of results when no such page exists.

And also, when we do not include `hasNextPage` but we are still on the last page, `endCursor` should be `null`.

 

## What is this change?

We now check if we have an extra element or not before we create the `endCursor`. We define having a next page of results as having an extra element in the result set that is returned from our query. This is because when we make a paginated query we add 1 to our limit, that way we can tell if there is another page of results based on if that extra element returns or not. We now include this logic if either `hasNextPage`, or `endCursor` is in the request so that we will correctly follow the logic for any pagination scenario. That means adding 1 to the limit and checking if there is another page of records beyond the number in the request.

## How was this tested?

Updated the pagination tests to have null `endCursor` expected when `hasNextPage` is false.

Added cases to the `RequestAfterTokenOnly` test cases to include both being on the final page and not, this verifies we get a correctly built `non null` and correctly built `null` `endCursor` for those respective cases.

## Sample Request(s)
```
{
	reviews(first: 2, after: "W3siRW50aXR5TmFtZSI6IlJldmlld3MiLCJGaWVsZE5hbWUiOiJib29rX2lkIiwiRmllbGRWYWx1ZSI6MSwiRGlyZWN0aW9uIjowfSx7IkVudGl0eU5hbWUiOiJSZXZpZXdzIiwiRmllbGROYW1lIjoiaWQiLCJGaWVsZFZhbHVlIjo1NjcsIkRpcmVjdGlvbiI6MH1d") {
		items {
			id
            content
        }
        hasNextPage
        endCursor
    }
}
```

Before the fix, this request would result in a response that would include an endCursor to a page that does not exist.

```{
  "data": {
    "reviews": {
      "items": [
        {
          "id": 568,
          "content": "I loved it"
        },
        {
          "id": 569,
          "content": "best book I read in years"
        }
      ],
      "hasNextPage": false,
      "endCursor": "W3siRW50aXR5TmFtZSI6IlJldmlldyIsIkZpZWxkTmFtZSI6ImJvb2tfaWQiLCJGaWVsZFZhbHVlIjoxLCJEaXJlY3Rpb24iOjB9LHsiRW50aXR5TmFtZSI6IlJldmlldyIsIkZpZWxkTmFtZSI6ImlkIiwiRmllbGRWYWx1ZSI6NTY5LCJEaXJlY3Rpb24iOjB9XQ=="
    }
  }
}
```

`W3siRW50aXR5TmFtZSI6IlJldmlldyIsIkZpZWxkTmFtZSI6ImJvb2tfaWQiLCJGaWVsZFZhbHVlIjoxLCJEaXJlY3Rpb24iOjB9LHsiRW50aXR5TmFtZSI6IlJldmlldyIsIkZpZWxkTmFtZSI6ImlkIiwiRmllbGRWYWx1ZSI6NTY5LCJEaXJlY3Rpb24iOjB9XQ==` decodes to `[{"EntityName":"Review","FieldName":"book_id","FieldValue":1,"Direction":0},{"EntityName":"Review","FieldName":"id","FieldValue":569,"Direction":0}]` and if we look at the data in our table we can see that an endCursor that indicates the next page of data starts after id=569 will not be providing a link to a page at all as there are no more records after id=569
image

After adding this fix we can see that instead we will have a null endCursor, as expected

```{
  "data": {
    "reviews": {
      "items": [
        {
          "id": 568,
          "content": "I loved it"
        },
        {
          "id": 569,
          "content": "best book I read in years"
        }
      ],
      "hasNextPage": false,
      "endCursor": null
    }
  }
}
```

Similar behavior can be seen for only request an `endCursor` but not including `hasNextPage`. If you simply remove `hasNextPage` from the above requests, you will see the same inconsistent behavior before the fix, and the same consistency now after the fix.

So looking at the request

```
{
	reviews(first: 2, after: "W3siRW50aXR5TmFtZSI6IlJldmlld3MiLCJGaWVsZE5hbWUiOiJib29rX2lkIiwiRmllbGRWYWx1ZSI6MSwiRGlyZWN0aW9uIjowfSx7IkVudGl0eU5hbWUiOiJSZXZpZXdzIiwiRmllbGROYW1lIjoiaWQiLCJGaWVsZFZhbHVlIjo1NjcsIkRpcmVjdGlvbiI6MH1d") {
		items {
			id
            content
        }
        endCursor
    }
}
```

Before the fix we would see

```
  "data": {
    "reviews": {
      "items": [
        {
          "id": 568,
          "content": "I loved it"
        },
        {
          "id": 569,
          "content": "best book I read in years"
        }
      ],
      "endCursor": "W3siRW50aXR5TmFtZSI6IlJldmlldyIsIkZpZWxkTmFtZSI6ImJvb2tfaWQiLCJGaWVsZFZhbHVlIjoxLCJEaXJlY3Rpb24iOjB9LHsiRW50aXR5TmFtZSI6IlJldmlldyIsIkZpZWxkTmFtZSI6ImlkIiwiRmllbGRWYWx1ZSI6NTY5LCJEaXJlY3Rpb24iOjB9XQ=="
    }
  }
}
```

But after the fix we have

```
  "data": {
    "reviews": {
      "items": [
        {
          "id": 568,
          "content": "I loved it"
        },
        {
          "id": 569,
          "content": "best book I read in years"
        }
      ],
      "endCursor": null
    }
  }
}
```


seantleonard:
can you show example of contents of last page before and after your change?

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -102,7 +102,9 @@ private static JsonObject CreatePaginationConnection(JsonElement root, Paginatio
             {
                 // parse *Connection.endCursor if there are no elements
                 // if no after is added, but it has been requested HotChocolate will report it as null
-                if (returnedElemNo > 0)
+                // Need to validate we have an extra element, because otherwise there is no next page
```
seantleonard, line 105: The comment above this doesn't really make sense. Is it still applicable?

> // parse *Connection.endCursor if there are no elements
> // if no after is added, but it has been requested HotChocolate will report it as null

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
seantleonard, line 1: What was the purpose of the `after` and `endCursor` values that you removed and replaced with `null`? were we erroneously adding an end cursor value to "expectedResponse" when we shoudn't have been expecting an end cursor?

seantleonard (commented):
Few questions. Can you also point out in PR description some test that exercise pagination to the last page? so we can look locally to see what the response contents as a result of this change?

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: I think what this comment is trying to say is that we you don't enter this conditional endCursor will be returned as null automatically, which is why we dont have to manually add it to the connection. I can change that language to try and be more clear.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
aaronburtle: we were effectively providing 1 more book mark than you'd expect. Consider a pagination response with an endCursor as providing you a book mark to the last record you visited. Then you can say after=bookmark and you will get the next page. What we were doing was providing that book mark even when you were already on the last page. So youd make a new request saying after the last page, which doesn't make any sense, there is no "after the last page." If you followed that book mark youd end up in a final state where everything was null, has next, items, endCursor, because there was no page after the last. Now we just stop providing a book mark for the last page and instead return null for endCursor in this case (where endCursor is our book mark).

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -102,7 +102,9 @@ private static JsonObject CreatePaginationConnection(JsonElement root, Paginatio
...
+                // and endCursor should be left as null.
+                if (returnedElemNo > 0 && hasExtraElement)
```
seantleonard, line 107: This conditional introduces a regression that also needs an added test to protect against once you identify a fix:

This query works as expected in `main` but fails with this change:

`main`
```graphql
query GetBooksPaged {
  books(first: 5) {
    items {
      id
      title
      publisher_id
    }
    endCursor
  }
}
```
result which includes endCursor
``` json
{
  "data": {
    "books": {
      "items": [
        { "<items>":"<here>"}
      ],
      "endCursor": "W3siRW50aXR5TmFtZSI6IkJvb2siLCJGaWVsZE5hbWUiOiJpZCIsIkZpZWxkVmFsdWUiOjUsIkRpcmVjdGlvbiI6MH1d"
    }
  }
}
```

However, with this PR's changes,
the same query has a `null` for `endCursor`. Weird, because including the field `hasNextPage` results in the `endCursor` being properly supplied (non-null)

``` json
{
  "data": {
    "books": {
      "items": [
        { "<items>":"<here>"}
      ],
      "endCursor": null
    }
  }
}
```

This works fine when `hasNextPage` field is included .. which is odd:
```graphql
query GetBooksPaged {
  books(first: 5) {
    items {
      id
      title
      publisher_id
    }
    endCursor
    hasNextPage
  }
}
```
result:
```json
{
  "data": {
    "books": {
      "items": [
        { "<items>":"<here>"}
      ],
      "endCursor": "W3siRW50aXR5TmFtZSI6IkJvb2siLCJGaWVsZE5hbWUiOiJpZCIsIkZpZWxkVmFsdWUiOjUsIkRpcmVjdGlvbiI6MH1d",
      "hasNextPage": true
    }
  }
}
```

seantleonard (changes requested):
Need to prevent the regression introduced and add a test to protect against it.

src/Core/Resolvers/SqlPaginationUtil.cs
aaronburtle: it is because the check for pagination (looking at if we have extra elements after the current page) that sets the value for `hasExtraElement` was hidden inside of a check looking for if we have included `hasNextPage`, I've moved the setting of this value outside of that conditional which should fix everything.

aaronburtle: > This conditional introduces a regression that also needs an added test to protect against once you identify a fix:
> 
> This query works as expected in `main` but fails with this change:


More context for what changed, and how we are handling this.

Main is handling this specific query correctly, but only because it happens to fulfill a scenario where the behavior is correct, not because we have the correct logic for handling pagination with only endCursor in the request.

This is because we would include an endCursor anytime there were elements returned greater than 0, and so in this case, that is met, and the correct endCursor comes back. However, we could have the case where more than 0 elements are returned by an endCursor should not be included.

To properly handle these cases in their entirety, we need our logic for correctly handling pagination scenarios to extend to requests that include either endCursor, hasNextPage, or both.

I have made changes to accommodate this and will add the details to the PR description.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -62,26 +62,33 @@ private static JsonObject CreatePaginationConnection(JsonElement root, Paginatio
                 root = document.RootElement.Clone();
             }
 
-            IEnumerable<JsonElement> rootEnumerated = root.EnumerateArray();
+            // If the request includes either hasNextPage or endCursor then to correctly return those
+            // values we need to determine the correct pagination logic
+            bool paginationScenario = paginationMetadata.RequestedHasNextPage || paginationMetadata.RequestedEndCursor;
```
seantleonard, line 67: nit: `isPaginationRequested`

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -62,26 +62,33 @@ private static JsonObject CreatePaginationConnection(JsonElement root, Paginatio
...
 
+            IEnumerable<JsonElement> rootEnumerated = root.EnumerateArray();
+            int returnedElemNo = rootEnumerated.Count();
```
seantleonard, line 70: nit: since touching this area of code, perhaps `returnedElemNo` can have a more descriptive name like `returnedElementCount`

src/Core/Resolvers/SqlPaginationUtil.cs
seantleonard: > However, we could have the case where more than 0 elements are returned by an endCursor should not be included.

Can you identify any examples of this? would be good to include in pr description to illustrate your findings.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -100,11 +107,13 @@ private static JsonObject CreatePaginationConnection(JsonElement root, Paginatio
...
             if (paginationMetadata.RequestedEndCursor)
             {
-                // parse *Connection.endCursor if there are no elements
-                // if no after is added, but it has been requested HotChocolate will report it as null
-                if (returnedElemNo > 0)
+                // Note: if we do not add endCursor to the connection but it was in the request, its value will
+                // automatically be populated as null.
+                // Need to validate we have an extra element, because otherwise there is no next page
+                // and endCursor should be left as null.
+                if (returnedElementCount > 0 && hasExtraElement)
```
seantleonard, line 114: Does `hasExtraElement` imply `returnedElementCount > 0`?
Because
```csharp
hasExtraElement = returnedElementCount == paginationMetadata.Structure!.Limit();
```

seantleonard: actually, i suppose `returnedElementCount` could get to zero via the code:

```csharp
            if (isPaginationRequested)
            {
                // structure.Limit() is first + 1 for paginated queries where hasNextPage or endCursor is requested
                hasExtraElement = returnedElementCount == paginationMetadata.Structure!.Limit();
                if (hasExtraElement)
                {
                    // In a pagination scenario where we have an extra element, this element
                    // must be removed since it was only used to determine if there are additional
                    // records after those requested.
                    rootEnumerated = rootEnumerated.Take(rootEnumerated.Count() - 1);
                    --returnedElementCount;
                }
            }
```

src/Service.Tests/SqlTests/GraphQLPaginationTests/MySqlGraphQLPaginationTests.cs
```diff
@@ -25,43 +25,68 @@ public static async Task SetupAsync(TestContext context)
...
-        [DataRow("boolean_types", "false", "true", 2, 4,
+        [DataRow("decimal_types", 0.333333, "", 1, "", true,
+            DisplayName = "Test after token for decimal values for last page.")]
+        [DataRow("boolean_types", "false", "true", 2, 3, false,
             DisplayName = "Test after token for boolean values.")]
+        [DataRow("boolean_types", "true", "", 3, "", true,
+            DisplayName = "Test after token for boolean values for last page.")]
         [DataRow("datetime_types", "\"1753-01-01T00:00:00.000\"",
-            "\"9999-12-31 23:59:59.000000\"", 3, 4,
+            "\"1999-01-08T10:23:54\"", 3, 1, false,
```
seantleonard, line 65: nit: please open issue to track fixing this (most likely) bug. When datetime was "\"9999-12-31 23:59:59.000000\"" the test fails locally for us (UTC-7), but not in UTC+5:30

seantleonard: I'm assuming that's why this number was changed

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -212,14 +213,16 @@ public async Task RequestItemsOnly()
             }
 
             string graphQLQuery = @"{
-                supportedTypes(first: 3," + $"after: \"{after}\" " +
+                supportedTypes(first: 2," + $"after: \"{after}\" " +
```
seantleonard, line 216: why does `first` need to be changed from 3 to 2?

src/Service.Tests/SqlTests/GraphQLPaginationTests/MsSqlGraphQLPaginationTests.cs
```diff
@@ -25,55 +25,92 @@ public static async Task SetupAsync(TestContext context)
 
         /// <inheritdoc />
         [DataTestMethod]
-        [DataRow("typeid", 1, 4, "", "",
+        [DataRow("typeid", 1, 4, "", "", false,
```
seantleonard, line 28: How did you determine whether to assign true/false to `isLastPage` given just these param values?

seantleonard:
for your pr description: 
> We not include this logic if either hasNextPage, or endCursor is in the request so that we will correctly follow the logic for any pagination scenario.

do you mean that "this logic" for adding to the limit() value is excluded when both `hasNextPage` and `endCursor` are excluded from the GraphQL request? 

Does this change look to affect REST API at all, with your latest commits?

Would be good to include a copy/paste of the example request/response for this: 
> Similar behavior can be seen for only request an `endCursor` but not including `hasNextPage`. If you simply remove `hasNextPage` from the above requests, you will see the same inconsistent behavior before the fix, and the same consistency now after the fix.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
aaronburtle: so that we can have tests that both land on the last page and land before the next page. Since there are only 4 records to look at, we need a value <3 to have another page available and get a non null endCursor. This was easier than adding new records and doesnt impact the test in any other way.

src/Service.Tests/SqlTests/GraphQLPaginationTests/MsSqlGraphQLPaginationTests.cs
aaronburtle: By looking at the records we populate the DB with, if the request will land on the last record, we set last page to true, otherwise there is another record available and we set it to false.

aaronburtle:
> for your pr description:
> 
> > We not include this logic if either hasNextPage, or endCursor is in the request so that we will correctly follow the logic for any pagination scenario.
> 
> do you mean that "this logic" for adding to the limit() value is excluded when both `hasNextPage` and `endCursor` are excluded from the GraphQL request?


Yes, I added that to the PR description now to be more clear.


> Does this change look to affect REST API at all, with your latest commits?

No. Nothing on the REST side is changing.

> Would be good to include a copy/paste of the example request/response for this:
> > Similar behavior can be seen for only request an `endCursor` but not including `hasNextPage`. If you simply remove `hasNextPage` from the above requests, you will see the same inconsistent behavior before the fix, and the same consistency now after the fix.

Included!

src/Service.Tests/SqlTests/GraphQLPaginationTests/MySqlGraphQLPaginationTests.cs
aaronburtle: I adjusted these just to be sure we'd land on either the last page or before the last page. I did not notice the bug youre referencing, but can take a look and open an issue once I understand it.

src/Service.Tests/SqlTests/GraphQLPaginationTests/MySqlGraphQLPaginationTests.cs
```diff
@@ -41,11 +41,11 @@ public static async Task SetupAsync(TestContext context)
             DisplayName = "Test after token for long values.")]
         [DataRow("long_types", 1, "", 1, "", true,
             DisplayName = "Test after token for long values for last page.")]
-        [DataRow("string_types", "\"\"", "\"null\"", 1, 3, false,
+        [DataRow("string_types", "\"\"", "\"lksa;jdflasdf;alsdflksdfkldj\"", 1, 2, false,
```
seantleonard, line 44: `"\"lksa;jdflasdf;alsdflksdfkldj\""` is an end cursor value?

seantleonard: nvm. I see it here: 
image

seantleonard (approved):
Thank you for making these changes, adding tests for coverage of both bugs identified and having pr description outlining all context and before/after

severussundar (approved):
Thanks for making these changes. This will help bring consistency in the value of `endCursor`  across Cosmos and SQL database types, GraphQL and REST.

aaronburtle:
connection issue caused test to fail, re-running

Issue #1953, opened 2024-01-04
[Bug]: Unable to use the generated eq filters on string or ID fields in graphql


pingu2k4:
### What happened?

Hey.
We are using DAB with cosmos DB. Any time that we attempt to filter a string or ID field by a given value, where said value is also the partition key, we get no data returned.

Let me provide a minimal example...
We have a Roles container, containing simple roles with an id, tenandId, roleName. Currently, all items have the tenantId of `9cbe5961-7060-479f-a4f4-fd4cb03d6550`, but of course this won't always be the case.

If we run the following query, everything works as expected and we get our data out:

```graphql
query Roles {
    roles(filter: { tenantId: { eq: "9cbe5961-7060-479f-a4f4-fd4cb03d6550" } }) {
        items {
            id
            tenantId
            roleName
        }
    }
}
```

The problem arises where we attempt to parameterise the filter value. Using the following query, and variables gets us no data:
```graphql
query Roles($tenantId: ID!) {
    roles(filter: { tenantId: { eq: $tenantId } }) {
        items {
            id
            tenantId
            roleName
        }
    }
}
```
```json
{
    "tenantId": "9cbe5961-7060-479f-a4f4-fd4cb03d6550"
}
```

With the parameterised query, if we change the filter from `eq` to `neq`, then we still get no data but this is correct behaviour of course. However when we then change the tenantId in the variables json, keeping the filter to a `neq` filter... we then get all the data out. So, the `neq` behaves correctly when parameterised, however in all instances, where the field we are filtering is the partition key, is a string or ID, and we are using the `eq` filter - we will get no data back 100% of the time.

For clarity, when I say that we get no data back, we do get a response but the items are empty, which isn't expected. There are no errors returned or anything. Example response:
```json
{
    "data": {
        "roles": {
            "items": []
        }
    }
}
```

### Version

0.9.7+e560142426d1c080b9fd7b7fabff51a276f6bf61

### What database are you using?

CosmosDB NoSQL

### What hosting model are you using?

Local (including CLI)
Static Web Apps (SWA)

### Which API approach are you accessing DAB through?

GraphQL

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

abhishekkumams:
when running locally using v0.9.7 and 0.10.11-rc, i was able to fetch the data using both `eq` and `neq` for sql as well cosmosDB.
[...]

PR #2134, opened 2024-03-26
Cosmos DB: Add variable support for filter by Id Query

sourabh1007:
## Why make this change?

ref. https://github.com/Azure/data-api-builder/issues/1953

## What is this change?

Added variable support if query filter has ID

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests


neeraj-sharma2592 (approved):
LGTM

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -396,7 +400,7 @@ private static async Task<(string? idValue, string? partitionKeyValue)> GetIdAnd
         /// </summary>
         /// <param name="parameter"></param>
         /// <returns></returns>
-        private static string? GetIdValue(object? parameter)
+        private static string? GetIdValue(IMiddlewareContext context, object? parameter)
```
ayush3797, line 403: Any advantage of passing the entire `context` instead of just `context.Variables`?

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -322,7 +323,7 @@ private static async Task<string> GetPartitionKeyPath(Container container, ISqlM
         }
 
 #nullable enable
-        private static async Task<(string? idValue, string? partitionKeyValue)> GetIdAndPartitionKey(IDictionary<string, object?> parameters, Container container, CosmosQueryStructure structure, ISqlMetadataProvider metadataStoreProvider)
+        private static async Task<(string? idValue, string? partitionKeyValue)> GetIdAndPartitionKey(IMiddlewareContext context, IDictionary<string, object?> parameters, Container container, CosmosQueryStructure structure, ISqlMetadataProvider metadataStoreProvider)
```
ayush3797, line 326: Please add method/param summaries to whatever new/existing methods modified.

ayush3797 (approved):
Some questions. LGTM otherwise!

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -1137,6 +1137,28 @@ public async Task TestQueryFilterFieldAuth_ExcludeTakesPredecence()
             Assert.IsTrue(errorMessage.Contains(DataApiBuilderException.GRAPHQL_FILTER_FIELD_AUTHZ_FAILURE));
 
         }
+
+        /// <summary>
+        /// Tests that the pk level query filter is working variables
```
Aniruddh25, line 1142: nit: `is working with variables`

src/Core/Resolvers/CosmosQueryEngine.cs
```diff
@@ -405,7 +409,18 @@ private static async Task<(string? idValue, string? partitionKeyValue)> GetIdAnd
                     if (string.Equals(item.Name.Value, "id", StringComparison.OrdinalIgnoreCase))
                     {
                         IList<ObjectFieldNode>? idValueObj = (IList<ObjectFieldNode>?)item.Value.Value;
-                        return idValueObj?.FirstOrDefault(x => x.Name.Value == "eq")?.Value?.Value?.ToString();
+
+                        ObjectFieldNode? itemToResolve = idValueObj?.FirstOrDefault(x => x.Name.Value == "eq");
```
Aniruddh25, line 413: Can there be a different operator than `eq` here? like `gt` or `lt`

Aniruddh25 (approved):
Few questions, LGTM otherwise!

src/Core/Resolvers/CosmosQueryEngine.cs
sourabh1007: yes it we are using `selection` from context also https://github.com/Azure/data-api-builder/blob/25012aa14fd2c72e9239b179d6d4f4acda2e4dfb/src/Core/Resolvers/CosmosQueryEngine.cs#L421

src/Core/Resolvers/CosmosQueryEngine.cs
sourabh1007: I don't think we support  `gt` `lt` with partition key.

PR #2135, opened 2024-03-27
Cosmos DB: Adds support for string array search

sourabh1007:
## Why make this change?

ref. https://github.com/Azure/data-api-builder/discussions/1897#discussioncomment-8761872

## What is this change?
Added `contains` and `notContains` support for string Array Type.

1. Data schema:
``` gql
type Planet @model {
    id : ID!,
    name : String,
    tags: [String!]
}
```

2. GraphQl :
```graphql
{
      planets(" + QueryBuilder.FILTER_FIELD_NAME + @" : {tags: { contains : ""tag1""}})
      {
          items {
              id
              name
          }
      }
}
```

4. Cosmos DB Query:
```sql
SELECT c.id, c.name FROM c where ARRAY_CONTAINS(c.tags, 'tag1')
```

5. GraphQl :
```graphql
{
    planets(" + QueryBuilder.FILTER_FIELD_NAME + @" : {tags: { notContains : ""tag3""}})
    {
        items {
            id
            name
        }
    }
}
```

7. Cosmos DB Query:
```sql
SELECT c.id, c.name FROM c where NOT ARRAY_CONTAINS(c.tags, 'tag3')
```

## How was this tested?

- [x] Integration Tests
- [ ] Unit Tests


src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -237,17 +237,33 @@ public GQLFilterParser(RuntimeConfigProvider runtimeConfigProvider, IMetadataPro
...
-                            ParseScalarType(
-                                ctx,
-                                argumentSchema: filterArgumentObject.Fields[name],
-                                fieldName: backingColumnName,
-                                fields: subfields,
-                                schemaName: schemaName,
-                                tableName: sourceName,
-                                tableAlias: sourceAlias,
-                                processLiterals: queryStructure.MakeDbConnectionParam)));
+                   new PredicateOperand(
```
ayush3797, line 256: Nit: Please use named params for quick reference.

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -490,11 +506,12 @@ public HttpContext GetHttpContextFromMiddlewareContext(IMiddlewareContext ctx)
         string schemaName,
         string tableName,
         string tableAlias,
-        Func<object, string?, string> processLiterals)
+        Func<object, string?, string> processLiterals,
+        bool isListType = false)
```
ayush3797, line 510: nit: add param summary

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -616,7 +633,8 @@ public static class FieldFilterParser
         IInputField argumentSchema,
         Column column,
         List<ObjectFieldNode> fields,
-        Func<object, string?, string> processLiterals)
+        Func<object, string?, string> processLiterals,
+        bool isListType = false)
```
ayush3797, line 637: Add in the param summary as to what this is and how this helps in deciding between using LIKE/NOT_LIKE Vs ARRAY_CONTAINS/NOT_CONTAINS.

src/Core/Resolvers/CosmosQueryBuilder.cs
```diff
@@ -125,7 +129,11 @@ protected override string Build(Predicate? predicate)
             }
 
             string predicateString;
-            if (ResolveOperand(predicate.Right).Equals(GQLFilterParser.NullStringValue))
+            if (predicate.Op == PredicateOperation.ARRAY_CONTAINS || predicate.Op == PredicateOperation.NOT_ARRAY_CONTAINS)
```
ayush3797, line 132: Are there potentially more ARRAY operators which might need to be added soon? If so, should we add a helper which classifies an operator as an ARRAY operator rather than appending an `||` for each of them?

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -1137,6 +1137,46 @@ public async Task TestQueryFilterFieldAuth_ExcludeTakesPredecence()
             Assert.IsTrue(errorMessage.Contains(DataApiBuilderException.GRAPHQL_FILTER_FIELD_AUTHZ_FAILURE));
 
         }
+
+        /// <summary>
+        /// Tests that the field level query filter work with list type
```
ayush3797, line 1142:
```suggestion
        /// Tests that the field level query filter work with list type for 'contains' operator
```

ayush3797 (approved):
Left some nits/suggestions. LGTM otherwise!

src/Core/Models/GraphQLFilterParsers.cs
sourabh1007: I have mentioned the purpose of this flag. I think code is straightforward and explaining how things are changing with this flag.

src/Core/Resolvers/CosmosQueryBuilder.cs
sourabh1007: I don't think there are more. if in future, I need to add more such condition, I will definitely consider it to move to a helper function.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -1139,7 +1139,47 @@ public async Task TestQueryFilterFieldAuth_ExcludeTakesPredecence()
...
+                planets(" + QueryBuilder.FILTER_FIELD_NAME + @" : {tags: { notContains : ""tag3""}})
+                {
+                    items {
+                        id
+                        name
+                    }
+                }
+            }";
+
+            string dbQuery = $"SELECT c.id, c.name FROM c where NOT ARRAY_CONTAINS(c.tags, 'tag3')";
```
Aniruddh25, line 1177: Can you please update the PR description that Integration tests have been added - Mark the check box to yes.

Also, provide a sample GQL query and corresponding Cosmos DB query that will be generated.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -1139,7 +1139,47 @@ public async Task TestQueryFilterFieldAuth_ExcludeTakesPredecence()
...
+            await ExecuteAndValidateResult(_graphQLQueryName, gqlQuery, dbQuery);
+        }
+
+        /// <summary>
+        /// Tests that the pk level query filter is working with variables
```
Aniruddh25, line 1182:
```suggestion
        /// Tests that the pk level query filter is working with variables.
```

neeraj-sharma2592 (approved):
LGTM

Issue #1718, opened 2023-09-15
[Bug]: Creating a relationship to the same table twice causes null data.


julianadormon:
### What happened?

Love this project!
Though I am encountering an issue.

I have an Orders table which has a two columns CountryTypeId and ShippingCountryTypeId.
I have a CountryTypes table which lists countries.

My C# Order model has a CountryType member and a ShippingCountryType member.

```
CountryType () {
        int CountryTypeId
        string CountryTypeName
        string CountryTypeCode
}
```

The goal is to fill these two members from a graphQl request. I do not have a join table in the database.
I want to pull data from CountryTypes table twice (2 joins), essentially, once for each the matching CountryTypeId and another for the ShippingCountryTypeId.

My dab.config looks like this:

```
"Order": {
      "source": {
          ..... for brevity
      },
      "graphql": {
        "enabled": true,
        ..... for brevity
        }
      },
      
"CountryType": {
          "cardinality": "one",
          "target.entity": "CountryType",
          "source.fields": [
            "CountryTypeId"
          ],
          "target.fields": [
            "CountryTypeId"
          ],
          "linking.source.fields": ["CountryTypeId"],
          "linking.target.fields": ["CountryTypeId"]
        },
        "ShippingCountryType": {
          "cardinality": "one",
          "target.entity": "CountryType",
          "source.fields": [
            "ShippingCountryTypeId"
          ],
          "target.fields": [
            "CountryTypeId"
          ],
          "linking.source.fields": ["ShippingCountryTypeId"],
          "linking.target.fields": ["CountryTypeId"]
        }
}
```

Graphql retuned:  

```
"CountryType": null,
 "ShippingCountryType": null
```

If I do:

```
"CountryType": {
          "cardinality": "one",
          "target.entity": "CountryType",
          "source.fields": [
            "CountryTypeId"
          ],
          "target.fields": [
            "CountryTypeId"
          ]
        }
}
```

It returns the proper data.

### Version

0.8.50+69f45352c90fae016c33c4e6c1294105898269cc

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

GraphQL

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

seantleonard:
## generated sqltext

```tsql
 SELECT TOP 100 
 [table0].[id] AS [id], 
 [table0].[countrytypeid] AS [countrytypeid],
[...]

PR #2138, opened 2024-03-27
GraphQL Queries - Entities with self-joining relationships get expected responses

### LINKED ISSUE #1940: [Bug]: T-SQL Generated for a Self-Joining Table is Invalid

seantleonard:
# Self Join

## Why this change?

- Closes #1940 

When users defined self-joining relationships between entities (SourceEntity == TargetEntity), DAB couldn't properly process the relationship at request-time and returned null/unexpected results due to incorrectly generated sqltext.

### Primary Changes
The primary changes:
- An additional member of SqlMetadataProvider named **RelationshipToFkDefinition** (`Dictionary<EntityRelationshipKey, ForeignKeyDefinition>`) 
- Two new **ForeignKeyDefinition** fields: 
  - `ReferencedEntityRole` -> indicates whether the entity in a self-joined relationship is the "Target" or "Source" entity.
  - `ReferencingEntityRole` -> indicates whether the entity in a self-joined relationship is the "Target" or "Source" entity.

### Example `dab-config.json` -> in-engine data structure

Below is an example translation between a dab-config.json relationship and the RelationshipToFkDefinition:

#### dab-config.json

```json
"relationships": {
  "parent_account": {
    "cardinality": "one",
    "target.entity": "dbo_DimAccount",
    "source.fields": [
      "ParentAccountKey"
    ],
    "target.fields": [
      "AccountKey"
    ]
  },
  "child_accounts": {
    "cardinality": "many",
    "target.entity": "dbo_DimAccount",
    "source.fields": [
      "AccountKey"
    ],
    "target.fields": [
      "ParentAccountKey"
    ]
  }
}
```

#### in-engine data structure

The following table illustrates how the dictionary `RelationshipToFkDefinition` is populated.

| EntityRelationshipKey {EntityName-RelationshipName} | FKD - Referenced Entity (Column) [Role]  | FKD Referencing Entity (Column) [Role]     |
|-----------------------------------------------------|------------------------------------------|--------------------------------------------|
| dbo_DimAccount - parent_account                     | dbo_DimAccount (AccountKey) [Target]     | dbo_DimAccount (ParentAccountKey) [Source] |
| dbo_DimAccount - child_accounts                     | dbo_DimAccount (AccountKey) [Source]     | dbo_DimAccount (ParentAccountKey) [Target] |

The enhanced metadata collection listed above enables DAB to pick out 
1. The correct `ForeignKeyDefinition` created to represent a config relationship
2. Which self-join relationship columns to set for `leftColumnNames` (Source Fields) and `rightColumnNames` (Target Fields).
    - The `ForeignKeyDefinition` object (prior to this pr) only tracked "referencing" and "referenced" columns and DAB didn't know whether to map
    Referencing - Source; Referenced - Target; OR Referencing - Target; Referenced - Source;
    - With this PR, DAB now determines the correct mapping which is critical to creating correct join statements in the sqltext.

DAB startup uses the relationship cardinality value (`one`, `many`) to determine whether
1. The source entity is the "referenced" or "referencing" entity.
2. The target entity is the "referenced" or "referencing" entity.

##### Cardinality: `one` (many:one; one:one)

Example: books(Many) - publisher(One) where books.publisher_id is referenc**ing** publisher.id

Referencing Entity | Referenced Entity
-------------------|-------------------
Source Entity      | Target Entity

##### Cardinality: `many` (one:many)

Example: publisher(One)-books(Many) where publisher.id is referenc**ed** by books.publisher_id

Referencing Entity | Referenced Entity
-------------------|-------------------
Target Entity      | Source Entity

### Example GraphQL Query

In a GraphQL query, the top level entity is always the **source** entity and the nested entity is the **target** entity.
- Before this PR, DAB didn't know whether the **source** entity was the referenced or referencing entity. 
  - Now DAB knows and finds out via querying:
- **dbo_DimAccount - parent_account** (entity - relationshipName) -> top level entity in query is **Source** and nested entity is **Target**
  - `ForeignKeyDefinition.ResolveSourceColumns()` -> check `ReferencingEntityRole` to return expected columns (referencing/referenced)
  - `ForeignKeyDefinition.ResolveTargetColumns()` -> checks `ReferencedEntityRole` to return expected columns (referencing/referenced)

```csharp
// BaseSqlQueryStructure.AddJoinPredicatesForSelfJoinedEntity(...)
            if (MetadataProvider.RelationshipToFkDefinition.TryGetValue(key: fkLookupKey, out ForeignKeyDefinition? fkDef))
            {
                subQuery.Predicates.AddRange(
                        CreateJoinPredicates(
                            leftTableAlias: SourceAlias,
                            leftColumnNames: fkDef.ResolveSourceColumns(),
                            rightTableAlias: subqueryTargetTableAlias,
                            rightColumnNames: fkDef.ResolveTargetColumns()));
            }
```

The following example requires DAB to utilize both entries in the `RelationshipToFkDefinition` table above:
 
```graphql
query DboDimAccountsByPk
 {
  dbo_DimAccount_by_pk(AccountKey: 2) {
    AccountKey
    ParentAccountKey
    child_accounts {
      items {
        AccountKey
      }
    }
    parent_account{
      AccountKey
    }
  }
}
```

- `parent_account` field configured with relationship cardinatlity (`one`) which means:
  - Source Entity: Referencing
  - Target Entity: Referenced
- `child_accounts` field configured with relationship cardinality (`many`) which means:
  - Source Entity: Referenced
  - Target Entity: Referencing

## How was this tested?

-[] Unit tests
-[x] Integration tests

### GraphQL Parent and Child Self-Join with tsql
sample data

| AccountKey | ParentAccountKey |
|------------|------------------|
| 1          | NULL             |
| 2          | 1                |
| 3          | 2                |
| 4          | 2                |

```graphql
            query queryAccountAndParent{
                dbo_DimAccounts(first: 3, filter: {ParentAccountKey: { isNull: false}}, orderBy: { AccountKey: ASC}) {
                    items {
                        AccountKey
                        ParentAccountKey
                      parent_account {
                            AccountKey
                        }
                        child_accounts{
                          items{
                            AccountKey
                            ParentAccountKey
                          }
                        }
                    }
                }
            }
```

```tsql
 SELECT TOP 3 
 [table0].[AccountKey] AS [AccountKey], 
 [table0].[ParentAccountKey] AS [ParentAccountKey], 
 JSON_QUERY ([table1_subq].[data]) AS [parent_account], 
 JSON_QUERY (COALESCE([table3_subq].[data], '[]')) AS [child_accounts] 
 FROM [dbo].[DimAccount] AS [table0] 
 OUTER APPLY (
	SELECT TOP 1 [table1].[AccountKey] AS [AccountKey] 
	FROM [dbo].[DimAccount] AS [table1] 
	WHERE [table1].[AccountKey] = [table0].[ParentAccountKey] 
	ORDER BY [table1].[AccountKey] ASC FOR JSON PATH, INCLUDE_NULL_VALUES,WITHOUT_ARRAY_WRAPPER) 
AS [table1_subq]([data]) 
OUTER APPLY (
	SELECT TOP 100 [table3].[AccountKey] AS [AccountKey], 
	[table3].[ParentAccountKey] AS [ParentAccountKey] 
	FROM [dbo].[DimAccount] AS [table3] 
	WHERE [table3].[ParentAccountKey] = [table0].[AccountKey] 
	ORDER BY [table3].[AccountKey] ASC FOR JSON PATH, INCLUDE_NULL_VALUES) 
AS [table3_subq]([data]) WHERE [table0].[ParentAccountKey] IS NOT NULL 
ORDER BY [table0].[AccountKey] ASC FOR JSON PATH, INCLUDE_NULL_VALUES
```
```json
{
  "data": {
    "dbo_DimAccounts": {
      "items": [
        {
          "AccountKey": 2,
          "ParentAccountKey": 1,
          "parent_account": {
            "AccountKey": 1
          },
          "child_accounts": {
            "items": [
              {
                "AccountKey": 3,
                "ParentAccountKey": 2
              },
              {
                "AccountKey": 4,
                "ParentAccountKey": 2
              }
            ]
          }
        },
        {
          "AccountKey": 3,
          "ParentAccountKey": 2,
          "parent_account": {
            "AccountKey": 2
          },
          "child_accounts": {
            "items": []
          }
        },
        {
          "AccountKey": 4,
          "ParentAccountKey": 2,
          "parent_account": {
            "AccountKey": 2
          },
          "child_accounts": {
            "items": []
          }
        }
      ]
    }
  }
}
```


src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -67,6 +67,13 @@ public interface ISqlMetadataProvider
         // of referencing and referenced tables.
         public Dictionary<RelationShipPair, ForeignKeyDefinition>? PairToFkDefinition { get; }
 
+        /// <summary>
+        /// Maps {entityName, relationshipName} to the foreign key definition defined for the relationship.
+        /// The fk definition denotes referencing/referenced fields and whether the referencing/refereced fields
```
ayush3797, line 72:
```suggestion
        /// The fk definition denotes referencing/referenced fields and whether the referencing/referenced fields
```

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -671,7 +671,11 @@ private void AddGraphQLFields(IReadOnlyList<ISelectionNode> selections, RuntimeC
                     string targetEntityName = MetadataProvider.GetEntityName(subunderlyingType.Name);
                     string subtableAlias = subquery.SourceAlias;
 
-                    AddJoinPredicatesForRelatedEntity(targetEntityName, subtableAlias, subquery);
+                    AddJoinPredicatesForRelatedEntityNEW(
```
ayush3797, line 674: Better method name.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -150,41 +150,92 @@ public Type GetColumnSystemType(string columnName)
...
-                // Identify the side of the relationship first, then check if its valid
+                // Identify the side of the relationship first, then check if it's valid
                 // by ensuring the referencing and referenced column count > 0
                 // before adding the predicates.
                 foreach (ForeignKeyDefinition foreignKeyDefinition in foreignKeyDefinitions)
                 {
                     // First identify which side of the relationship, this fk definition
                     // is looking at.
-                    if (foreignKeyDefinition.Pair.ReferencingDbTable.Equals(DatabaseObject))
+                    if (EntityName == targetEntityName)
```
ayush3797, line 225: Why is self-joining case considered under `LinkingTableProcessing()`?

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -150,41 +150,92 @@ public Type GetColumnSystemType(string columnName)
...
+        public void AddJoinPredicatesForRelatedEntityNEW(
+            string relationshipName,
+            string targetEntityName,
+            string relatedSourceAlias,
+            BaseSqlQueryStructure subQuery)
+        {
+            EntityRelationshipKey selfJoinKey = new(EntityName, relationshipName);
+            if (MetadataProvider.RelationshipToFkDefinitions.TryGetValue(key: selfJoinKey, out ForeignKeyDefinition? fkDef))
+            {
+                if (fkDef.ReferencedEntityRole is not RelationshipRole.Linking && fkDef.ReferencingEntityRole is not RelationshipRole.Linking)
```
ayush3797, line 167: When is ReferencedEntityRole ever set to `Linking`?

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -150,41 +150,92 @@ public Type GetColumnSystemType(string columnName)
...
         /// <param name="relatedSourceAlias">The alias assigned for the underlying source of this related entity.</param>
         /// <param name="subQuery">The subquery to which the join predicates are to be added.</param>
-        public void AddJoinPredicatesForRelatedEntity(
+        public void AddJoinPredicatesForRelatedEntityNEW(
+            string relationshipName,
+            string targetEntityName,
+            string relatedSourceAlias,
+            BaseSqlQueryStructure subQuery)
+        {
+            EntityRelationshipKey selfJoinKey = new(EntityName, relationshipName);
```
ayush3797, line 164: It's not really a selfJoin key but more like `currentEntityRelationshipKey`

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
seantleonard: fixed this in a second spot as well.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
seantleonard: agreed.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
seantleonard: good catch, not intended. removed.

ayush3797:
Still not clear with the purpose of inclusion of `RelationshipRole` for an entity. Why does it matter when you have EntityRelationshipKey ?

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
seantleonard: Set when a linking object is detected in the relationship within `private void AddForeignKeysForRelationships()`
https://github.com/Azure/data-api-builder/blob/5a09b5aafef7bbe44dcc74bd82a475cf7ea56be7/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs#L771-L798

ayush3797: Here the referencingentityrole is set to Linking right?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -815,52 +830,68 @@ private static EntitySourceType GetEntitySourceType(string entityName, Entity en
...
                     // and the foreign key is defined in the source of targetEntity.
                     // This foreign key WILL NOT exist if its a Many-One relationship.
-                    AddForeignKeyForTargetEntity(
-                        targetEntityName,
-                        referencingDbTable: targetDbTable,
-                        referencedDbTable: databaseTable,
-                        referencingColumns: relationship.TargetFields,
-                        referencedColumns: relationship.SourceFields,
-                        relationshipData);
+                    // Skips this fk when target and source entities are the same (self-referencing).
```
severussundar, line 863: Adding a note about why the addition of foreign key is skipped (for self-referencing relationship) will be helpful when going through the logic in the future

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -815,52 +830,68 @@ private static EntitySourceType GetEntitySourceType(string entityName, Entity en
...
                 }
                 else if (relationship.Cardinality is Cardinality.Many)
                 {
-                    // Case of publisher(One)-books(Many)
-                    // we would need to obtain the foreign key information from the books table
+                    // Example: publisher(One)-books(Many)
+                    // Obtain the foreign key information from the books table
                     // about the publisher id so we can do the join.
-                    // so, the referencingTable is the source of the target entity.
+                    // The referencingTable (books) is the source of the target entity (publisher). //confusing language.
```
severussundar, line 884: Needs to re-phrased

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1802,73 +1856,90 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                    if (!doesFkExistInDatabase)
                     {
-                        validatedFKDefinitionsToTarget.Add(fKDefinitionToTarget);
+                        validatedFKDefinitionsToTarget.Add(configResolvedFkDefinition);
+
+                        // The following operation generates FK metadata for use when processing requests on self-joined/referencing entities.
+                        EntityRelationshipKey key = new(entityName: configResolvedFkDefinition.SourceEntityName, configResolvedFkDefinition.RelationshipName);
+                        if (!RelationshipToFkDefinitions.TryAdd(key, configResolvedFkDefinition))
+                        {
+                            Console.WriteLine("for config-only defined relationship, no db fk, dab is failing to keep track of both 2 generated fk definitions.");
```
severussundar, line 1941: nit: update L1941 to throw an exception.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1877,8 +1948,26 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
         }
 
         /// <summary>
-        /// For the given two database objects, returns true if a foreignKey exists between them.
-        /// Else returns false.
+        /// When a foreign key relationship is defined in the runtime config, the user may configure
```
severussundar, line 1951:
```suggestion
        /// When a relationship is defined in the runtime config, the user may configure
```

severussundar (commented):
Posting comments so far...still need to look at more files.

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1564,6 +1564,116 @@ public async Task GraphQLQueryWithMultipleOfTheSameFieldReturnsFieldOnce()
...
+        /// <summary>
+        /// Validates that DAB evaluates the self-joining relationship "child_accounts "for the entity dbo_DimAccounts.
+        /// The field child_accounts represents the one-to-many relationship entry:
+        /// - source entity: dbo_DimAccount | source.fields: AccountKey
+        /// - target entity: dbo_DimAccount | target.fields: ParentAccountKey
+        /// In plain language: one (parent) account whose referenced field AccountKey maps
+        /// to many child records' referencing field ParentAccountKey.
+        /// </summary>
+        [TestMethod]
+        public async Task QueryById_SelfReferencingRelationship_ReturnsExpectedChildre()
```
severussundar, line 1576: nit:
```suggestion
        public async Task QueryById_SelfReferencingRelationship_ReturnsExpectedChildren()
```

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1564,6 +1564,116 @@ public async Task GraphQLQueryWithMultipleOfTheSameFieldReturnsFieldOnce()
...
+            Assert.AreEqual(2, response.GetProperty("AccountKey").GetInt32());
+            Assert.AreEqual(1, response.GetProperty("ParentAccountKey").GetInt32());
+            Assert.AreEqual(2, response.GetProperty("child_accounts").GetProperty("items").GetArrayLength());
+            List<JsonElement> childAccounts = response.GetProperty("child_accounts").GetProperty("items").EnumerateArray().ToList();
+            Assert.IsTrue(childAccounts[0].GetProperty("AccountKey").GetInt32() == 3);
+            Assert.IsTrue(childAccounts[1].GetProperty("AccountKey").GetInt32() == 4);
+        }
+
+        /// <summary>
+        /// Validates that DAB evaluates the self-joining relationship "parent_account "for the entity dbo_DimAccounts.
```
severussundar, line 1624: nit:
```suggestion
        /// Validates that DAB evaluates the self-joining relationship "parent_account" for the entity dbo_DimAccounts.
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -53,6 +53,13 @@ public abstract class SqlMetadataProvider<ConnectionT, DataAdapterT, CommandT> :
         // of referencing and referenced tables.
         public Dictionary<RelationShipPair, ForeignKeyDefinition>? PairToFkDefinition { get; set; }
 
+        /// <summary>
+        /// Maps {entityName, relationshipName} to the foreign key definition defined for the relationship.
+        /// The fk definition denotes referencing/referenced fields and whether the referencing/referenced fields
+        /// apply to the target or source entity as defined in the relationshp config.
```
severussundar, line 59: nit:
```suggestion
        /// apply to the target or source entity as defined in the relationship in the config file.
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -777,19 +784,27 @@ private static EntitySourceType GetEntitySourceType(string entityName, Entity en
                     (linkingTableSchema, linkingTableName) = ParseSchemaAndDbTableName(relationship.LinkingObject)!;
                     DatabaseTable linkingDbTable = new(linkingTableSchema, linkingTableName);
                     AddForeignKeyForTargetEntity(
+                        sourceEntityName: entityName,
+                        relationshipName: relationshipName,
                         targetEntityName,
```
severussundar, line 789:
```suggestion
                        targetEntityName: targetEntityName,
```

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -777,19 +784,27 @@ private static EntitySourceType GetEntitySourceType(string entityName, Entity en
...
                         referencingColumns: relationship.LinkingSourceFields,
                         referencedColumns: relationship.SourceFields,
+                        referencingEntityRole: RelationshipRole.Linking,
+                        referencedEntityRole: RelationshipRole.Source,
                         relationshipData);
 
                     AddForeignKeyForTargetEntity(
+                        sourceEntityName: entityName,
+                        relationshipName: relationshipName,
                         targetEntityName,
```
severussundar, line 801: targetEntityName: targetEntityName

severussundar (commented):
Looks good overall

1. Some suggestions to incorporate
2. For self-joining relationships, even when the relationship cardinality is one, the foreign keys are added only in one direction ~ source --> referencing, target ---> referenced. Suggestion to add a unit test in `SqlMetadataProviderUnitTests` to check the count of foreign keys for a self-join relationship

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1564,6 +1564,116 @@ public async Task GraphQLQueryWithMultipleOfTheSameFieldReturnsFieldOnce()
...
+                        parent_account {
+                            AccountKey
+                        }
+                    }
+                }
+            }";
+
+            JsonElement response = await ExecuteGraphQLRequestAsync(query, "dbo_DimAccounts", false);
+
+            /* Expected Resposne
```
abhishekkumams, line 1649:
```suggestion
            /* Expected Response
```

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -72,6 +72,13 @@ public interface ISqlMetadataProvider
         // of referencing and referenced tables.
         public Dictionary<RelationShipPair, ForeignKeyDefinition>? PairToFkDefinition { get; }
 
+        /// <summary>
+        /// Maps {entityName, relationshipName} to the foreign key definition defined for the relationship.
+        /// The fk definition denotes referencing/referenced fields and whether the referencing/referenced fields
+        /// apply to the target or source entity as defined in the relationshp config.
```
abhishekkumams, line 78:
```suggestion
        /// apply to the target or source entity as defined in the relationship config.
```

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -290,6 +303,38 @@ public class ForeignKeyDefinition
...
+        {
+            return ReferencingColumns;
+        }
+        else if (ReferencedEntityRole == RelationshipRole.Target)
+        {
+            return ReferencedColumns;
+        }
+        else
+        {
+            throw new Exception("Unable to resolve target columns");
```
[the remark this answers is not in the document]
seantleonard: in this case, i think it's more accurate to say:
> message: "Unable to resolve target columns because this ForeignKeyDefinition relates a source entity and linking object.");

while the relationship in runtime config may denote the target columns, the specific ForeignKeyDefinition may only represent the relationship between the target entity and linking object.

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -290,6 +303,38 @@ public class ForeignKeyDefinition
...
+        {
+            return ReferencingColumns;
+        }
+        else if (ReferencedEntityRole == RelationshipRole.Source)
+        {
+            return ReferencedColumns;
+        }
+        else
+        {
+            throw new Exception("Unable to resolve source columns");
```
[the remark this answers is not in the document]
seantleonard: in this case, i think it's more accurate to say:
> message: "Unable to resolve source columns because this ForeignKeyDefinition relates a source entity and linking object.");

while the relationship in runtime config may denote the target columns, the specific ForeignKeyDefinition may only represent the relationship between the source entity and linking object.

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -310,14 +355,27 @@ public override int GetHashCode()
...
+    DatabaseTable referencedDbObject)
+    {
+        ReferencingDbTable = referencingDbObject;
+        ReferencedDbTable = referencedDbObject;
+    }
+
+    public RelationShipPair(
+        string relationshipName,
         DatabaseTable referencingDbObject,
         DatabaseTable referencedDbObject)
```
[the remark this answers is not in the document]
seantleonard: we can make your suggested change when relevant to addressing that issue.

src/Config/ObjectModel/FkDefSource.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
+
+/// <summary>
+/// Identifies the source of the foreign key definition.
+/// </summary>
+public enum ForeignKeyDefinitionSource
```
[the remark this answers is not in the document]
seantleonard: Changed to RelationshipDefinitionSource

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: updated comments in whole section.

ayush3797:
From the code, looks like `RelationshipToFKDefinition` is only populated for self-referencing relationships? Accordingly, should we update the data structure's name to indicate exactly what it is storing?

src/Config/ObjectModel/EntityRelationshipKey.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <summary>
+    /// Identifies a specific value pair:
+    /// 1. entity name
+    /// 2. relationship name
+    /// Which can be used to uniquely identify a relationship (ForeignKeyDefinition object(s)).
+    /// </summary>
+    [DebuggerDisplay("{EntityName} - {RelationshipName}")]
+    public class EntityRelationshipKey
+    {
+        public string EntityName { get; set; }
```
ayush3797, line 17: From the usages, looks like EntityName is the source entity name for the relationship. Helpful to add a comment here or in the summary above.

src/Config/ObjectModel/EntityRelationshipKey.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public bool Equals(EntityRelationshipKey? other)
+        {
+            if (other == null)
+            {
+                return false;
+            }
+
+            return EntityName == other.EntityName && RelationshipName == other.RelationshipName;
```
ayush3797, line 38: For string comparisons, can we use Equals() ?

src/Config/ObjectModel/RelationshipDefinitionSource.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
+
+/// <summary>
+/// Identifies the source of the relationship.
+/// </summary>
+public enum RelationshipDefinitionSource
```
ayush3797, line 9: I see multiple references to this where you're setting the `ForeignKeyDefinition.Source` in the logic, those references are never used.

src/Core/Services/MetadataProviders/ISqlMetadataProvider.cs
```diff
@@ -72,6 +72,13 @@ public interface ISqlMetadataProvider
         // of referencing and referenced tables.
         public Dictionary<RelationShipPair, ForeignKeyDefinition>? PairToFkDefinition { get; }
 
+        /// <summary>
+        /// Maps {entityName, relationshipName} to the ForeignKeyDefinition defined for the relationship.
```
ayush3797, line 76: might need to update the comment to indicate that this only stores relevant data for self-referencing relationships.

seantleonard:
@ayush3797 
> From the code, looks like RelationshipToFKDefinition is only populated for self-referencing relationships?

For now yes, the idea is that as we update DAB, we'd look into using this datastructure for tracking other relationships too. Essentially enhancing how we keep track of relationships. 

> Accordingly, should we update the data structure's name to indicate exactly what it is storing?

I don't want to change the name so that we can keep the interface constant when we do adopt this datastructure for other scenarios in the future.

src/Config/ObjectModel/RelationshipDefinitionSource.cs
seantleonard: good catch. I wanted to use `RelationshipDefinitionSource` to supplement what we know about a particular FKDefinition. ultimately, didn't use it. Removing. But may be good to add in future.

src/Config/ObjectModel/EntityRelationshipKey.cs
seantleonard: added comment

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -3674,6 +3674,58 @@
...
+          "target.fields": [
+            "ParentAccountKey"
+          ],
+          "linking.source.fields": [],
+          "linking.target.fields": []
+        }
+      }
     }
   }
 }
```
ayush3797, line 3731: Add a new line here?

ayush3797 (approved):
LGTM! Thanks for fixing this!

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: went with using "tryadd" and conditioned this section (adding entityrelationship keys) to self-join scenario only.

src/Config/ObjectModel/RelationshipRole.cs
```diff
@@ -0,0 +1,17 @@
...
+
+/// <summary>
+/// Denotes the role of a referencing or referenced table in a relationship.
+/// Default value is None because the role is only resolved in some scenarios
+/// such as self-referencing relationships.
+/// </summary>
+public enum RelationshipRole
+{
+    None,
+    Source,
```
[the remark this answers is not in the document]
seantleonard: clarified comment:
```csharp
/// Denotes the role of a referencing or referenced table in a relationship.
/// Default value is None because the role is only resolved to help DAB
/// create join predicates for self-referencing relationships.
/// In non self-join relationships, DAB uses the ForeignKeyDefinition.RelationShipPair
/// to determine the referencing/referenced entity. RelationShipPair isn't sufficient
/// for self-join relationships because it the DatabaseObjects used to represent the pair
/// reference the same object: e.g. both Referenced/Referencing entity would point to 'dbo.MyTable'
```

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
seantleonard: yes

src/Service.Tests/dab-config.MsSql.json
seantleonard: added back. Probably removed by dotnet format.

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -3674,6 +3674,58 @@
...
+            {
+              "action": "*"
+            }
+          ]
+        }
+      ],
+      "relationships": {
+        "parent_account": {
+          "cardinality": "one",
+          "target.entity": "dbo_DimAccount",
```
Aniruddh25, line 3706: From the cardinality, how can you infer that the source is `ReferencingEntity` for sure? I get this is possible in many:one relationship. 

How about one:one relationships? In those either side can be the referencing entity, no? depending upon which entity contains the foreign key..

src/Config/ObjectModel/RelationshipRole.cs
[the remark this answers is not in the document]
Aniruddh25: Would be useful to identify what a source means? what a target means? IIUC, its with respect to how the configuration is layed out.

Aniruddh25 (commented):
Posting comments so far. Few more files to go.

src/Config/ObjectModel/RelationshipRole.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
+
+/// <summary>
+/// Denotes the role of a referencing or referenced table in a relationship.
+/// Default value is None because the role is only resolved to help DAB
+/// create join predicates for self-referencing relationships.
+/// In non self-join relationships, DAB uses the ForeignKeyDefinition.RelationShipPair
+/// to determine the referencing/referenced entity. RelationShipPair isn't sufficient
+/// for self-join relationships because it the DatabaseObjects used to represent the pair
```
severussundar, line 12: nit:
```suggestion
/// for self-join relationships because the DatabaseObjects used to represent the pair
```

src/Service.Tests/dab-config.MsSql.json
severussundar: For queries, we do not need to accurately identify the referencing and referencing entities. 

For relationships with cardinality `one`, DAB adds foreign key definition in both entities. Eventually, this translates to an additional incorrect condition during the join operation. This extra condition is the cause for returning no results getting returned (as both the conditions are never satisfied). 

When the foreign key definition is added in both entities, the database query is of the form,
```sql
WHERE [table0].[ParentAccountKey] = [table1].[AccountKey]
      AND [table0].[AccountKey]       = [table1].[ParentAccountKey]
```

but, we would want only the right equality check.

```sql
WHERE [table0].[ParentAccountKey] = [table1].[AccountKey]
```
Generating the right condition is only dependent on identifying and matching the top-level entity and the nested entity with their corresponding relationship fields. 

So, essentially, for self-join, 
-  Foreign key definition should be populated only with one combination (source --> referencing, target --> referenced or vice-versa. This doesn't have to be accurate) to avoid the extra incorrect condition. 
- `ForeignKeyDefinition` data structure contains `ReferencingColumns`, `ReferencedColumns` and `RelationShipPair`. We do not immediately know which is the referenced table and match it with referenced columns (same for referencing columns) when generating the sql query. In different places, the database object is compared to see if it is the referencedtable/referencingtable before fetching referencing/referenced columns. With self-join relationships, this is a challenge, because the database object is the same for referencing and referenced tables, hence the need for relationship roles and helper methods such as `ResolveSourceColumns()`, `ResolveTargetColumns()` which enable matching the top-level entity/nested entity with their associated relationship fields.

severussundar (approved):
LGTM!

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -316,12 +387,27 @@ public override int GetHashCode()
...
 public class RelationShipPair
 {
+    /// <summary>
+    /// Relationship name specified in the runtime config.
+    /// </summary>
+    public string RelationshipName { get; set; } = string.Empty;
+
     public RelationShipPair() { }
 
     public RelationShipPair(
```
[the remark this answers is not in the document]
seantleonard: Yes, there are still references to all constructors (during validations andoptimistic foreignkeydefinition creation scenarios). It will be a gradual adoption through the codebase to use the constructor that allows the dev to provide the relationship name. However, at this time, not all locations have a reference to the relationship name when `RelationshipPair` is created :

- `RelationShipPair(
        string relationshipName,
        DatabaseTable referencingDbObject,
        DatabaseTable referencedDbObject)`
-   `public RelationShipPair() { }`
- `public RelationShipPair(
DatabaseTable referencingDbObject,
DatabaseTable referencedDbObject)`

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1590,6 +1590,120 @@ public async Task GraphQLQueryWithMultipleOfTheSameFieldReturnsFieldOnce()
...
+            "AccountKey": 4
+          }
+        ]
+      }
+    }
+  }
+}
+             */
+
+            Assert.AreEqual(expected: 2, actual: response.GetProperty("AccountKey").GetInt32());
```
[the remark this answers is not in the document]
seantleonard: I didn't explicitly decide not to run a SQL query against the DB's. The GraphQLQueryTestBase has a mix of using DB results to validate and explicit validation of GraphQL result field contents.

When writing the test, I wanted to check the following boxes:
- Results matched what I expected to be returned for a self-join (e.g. parentAccountKey actually matched the parent record resolved via the relationship). 
- Readability: see exactly what the test is validating in the test without switching between files and parsing out 4 types of sqltext syntax
- Database agnostic.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1779,34 +1837,31 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
                     foreach ((string targetEntityName, List<ForeignKeyDefinition> fKDefinitionsToTarget) in relationshipData.TargetEntityToFkDefinitionMap)
                     {
-                        // 
-                        // Scenario 1: When a FK constraint is defined between source and target entities in the database.
-                        // In this case, there will be exactly one ForeignKeyDefinition with the right pair of Referencing and Referenced tables. 
-                        // Scenario 2: When no FK constraint is defined between source and target entities, but the relationship fields are configured through config file
-                        // In this case, two entries will be created. 
-                        // First entry: Referencing table: Source entity, Referenced table: Target entity
-                        // Second entry: Referencing table: Target entity, Referenced table: Source entity 
+                        // fkDefinitionsToTarget is a List that is hydrated differently depending
```
Aniruddh25, line 1857: do you mean this hydration has already happened before calling the GetValidatedFKs call below or is that state after the call is complete?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1818,73 +1873,99 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                        }
                     }
-                    // The count of referenced and referencing columns being = 0 indicates that source.fields and target.fields 
-                    // are not configured through the config file. In this case, the FK fields inferred from the database are populated.
                     else
                     {
-                        validatedFKDefinitionsToTarget.Add(inferredFKDefinition);
+                        validatedFKDefinitionsToTarget.Add(databaseResolvedFkDefinition);
+
+                        // Save additional metadata for use when processing requests on self-joined/referencing entities.
```
Aniruddh25, line 1908: Is this additional metadata not stored if its not a self joining relationship? Do we need to make it a special case?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: The hydration has already happened before calling the GetValidatedFKs. The comment describes the possible states of `fkDefinitionsToTarget` that `GetValidatedFKs()` needs to handle.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: I intend that this is a special case because the additional metadata is only stored for self-joining relationships. My intention is that the new way of storing metadata is gradually adopted other types of relationships. This PR would be an order of magnitude larger  if I didn't special case the new metadata and saved it for every type of relationship + strip out whatever old relationship metadata we tracked.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1830,73 +1885,99 @@ IEnumerable<List<ForeignKeyDefinition>> foreignKeysForAllTargetEntities
...
+                        {
+                            EntityRelationshipKey entityToFkDefKey = new(
+                                entityName: configResolvedFkDefinition.SourceEntityName,
+                                relationshipName: configResolvedFkDefinition.RelationshipName);
+                            RelationshipToFkDefinition.TryAdd(entityToFkDefKey, configResolvedFkDefinition);
+                        }
                     }
-                    // The count of referenced and referencing columns being = 0 indicates that source.fields and target.fields 
-                    // are not configured through the config file. In this case, the FK fields inferred from the database are populated.
                     else
```
Aniruddh25, line 1916: why dont we add the `configResolvedFkDefinition` to validatedFKdefinitions in this case since that has more metadata?

Aniruddh25 (commented):
Pending to look at QueryStructures.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
seantleonard: Good catch, both the configResolvedFkDefinition and databaseResolvedFkDefinition had important metadata that would be beneficial to consolidation.

image

I added this change to consolidate the referencing/referenced columns from `databaseResolvedFkDefinition` into `configResolvedFkDefinition` because `configResolvedFkDefinition` was missing that metadata, but did have the metadata ( the "more metadata" you called out) for relationshipName/sourceEntityName.

```csharp
                        // When the configured relationship doesn't override the database FK constraint,
                        // DAB can consolidate the referenced and referencing columns from the database FK definition
                        // into the configResolvedFkDefinition object.
                        configResolvedFkDefinition.ReferencedColumns = databaseResolvedFkDefinition.ReferencedColumns;
                        configResolvedFkDefinition.ReferencingColumns = databaseResolvedFkDefinition.ReferencingColumns;
                        validatedFKDefinitionsToTarget.Add(configResolvedFkDefinition);

                        // Save additional metadata for use when processing requests on self-joined/referencing entities.
                        // Since the configResolvedFkDefinition has additional metadata populated, DAB supplements that
                        // object by using the inferred FK definition's referenced/referencing columns.
                        if (IsSelfJoiningRelationship(databaseResolvedFkDefinition))
                        {
                            EntityRelationshipKey entityToFkDefKey = new(
                                entityName: configResolvedFkDefinition.SourceEntityName,
                                relationshipName: configResolvedFkDefinition.RelationshipName);
                            RelationshipToFkDefinition.TryAdd(entityToFkDefKey, configResolvedFkDefinition);
                        }
```

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -143,8 +143,85 @@ public Type GetColumnSystemType(string columnName)
...
+        public void AddJoinPredicatesForRelationship(
+            EntityRelationshipKey fkLookupKey,
+            string targetEntityName,
+            string subqueryTargetTableAlias,
+            BaseSqlQueryStructure subQuery)
+        {
+            if (string.Equals(fkLookupKey.EntityName, targetEntityName))
+            {
+                AddJoinPredicatesForSelfJoinedEntity(
+                    fkLookupKey: fkLookupKey,
```
Aniruddh25, line 165: Is this fkLookupKey is not useful today when its not a self joined relationship? If we start using it for relationships between different entities, can we for e.g. solve the problem of having multiple relationships between same pair of different entities?

seantleonard: With this PR, `fkLookupKey` is only useful for self-joined relationships. I initially authored this PR to utilize the `fkLookupKey` for all types of relationships, but more issues came up which required even more cascading changes make work. As a result, I'm introducing these changes incrementally. First -> to enable self-join relationships. In the future, can adopt this new way of persisting relationshipName-> fkMetadata to non-self-join relationships.

src/Service.Tests/dab-config.MsSql.json
Aniruddh25: I understand this reasoning. What confused me is the fact that the description doesn't say, we need to use source entity as referencing + target entity as referenced when cardinality is `one` ONLY when the db doesn't have a foreign key defined in it.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
Aniruddh25: Is there an issue tracking this work to allow multiple relationships between same pair of different entities? 

Also, will this PR solve this issue as well: https://github.com/Azure/data-api-builder/issues/1718 ?

seantleonard: By:
> this work to allow multiple relationships between same pair of different entities?

do you mean like the following? I'm not seeing how that is different from #1718:

```json
{
  "entities" {
    "MyEntity" {
      "relationships": {
         "Relationship1": {
             "targetEntity": "YourEntity", // 
             "sourceCols": ["col1"],
             "targetCols": [ "colA"]
           },
         "Relationship2" {
             "targetEntity": "YourEntity",
             "sourceCols": ["col2"],
             "targetCols": [ "colb"]
}}

seantleonard: 1718 is not fixed via this change.

src/Config/ObjectModel/RelationshipRole.cs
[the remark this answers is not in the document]
seantleonard: You're right, Ani. The concept of 'source' originates from the DAB config. 
- **Source** -> entity which defines a relationship between it (the source) and a target entity (defined by "targetentity" property).

In GraphQL query, source is the parent entity and target is the sub-entity.
image

PR #2139, opened 2024-03-28
Add validation for source and target fields in relationships

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/1935

Closes https://github.com/Azure/data-api-builder/issues/2145

By adding validation for source and target fields and linking source and linking target fields in the `RuntimeConfigValidator`.

## What is this change?

In the `ValidateRelationshipsInConfig` method within the `RuntimeConfigValidator`, we add logic to cover the cases from https://github.com/Azure/data-api-builder/issues/1935 such that we properly validate the relationships in the config.

It should be noted that this behavior aligns directly with the published documentation that describes how to use configs with dab. However, it also changes some behavior of the engine. That is, before this validation existed, it was possible to provide a config that was not valid as per our documentation, but dab would sort out the relationships on its own using foreign key information. For most versions this means merging the unknown values with the foreign key information, and in the most recent version it meant over riding entirely with the foreign key version in most invalid scenarios.

But we want the process of generating and using a config to be as clear as possible, with as few confusing grey areas as we can manage. To do this, we now enforce the validation steps that align with our documentation, and do not allow for badly formed configs to be used. We no longer will try to fix a bad config with the engine but will instead clearly communicate to the user that their config does not pass validation, and why.

There are two main differences in the validation logic that we add for this PR. They differ when we are in a one to many or many to one versus when we are in a many to many scenario.

In either one to many or many to one, there exists a relationship directly between source and target entities. This means that source and target fields need to match. However, in a many to many relationship, the source and target entities are related through a linking object. When that linking object exists in the config we know we are in a many to many relationship and use the separate validation logic. In this case, is is the source and linkingSource fields that must match, as well as the target and linkingTarget fields that must match. See: https://learn.microsoft.com/en-us/azure/data-api-builder/relationships for more information.

We will enforce the same principles for one to many, many to one, and many to many, such that the entire set of required fields must be included in any of those scenarios, but at the `foreign key` level. This means that in many to many you can have the source and linkingSource fields existing, but null target and linkingTarget fields, so long as the foreign key information provides the relationship information needed on the missing side. So for missing target and linkingTarget fields that would be a foreign key relationship between the target entity and linking object, and if source and linkingSource are missing, then between the source entity and linking object.

We also remove, in its entirety, `TestRelationshipForCorrectPairingOfLinkingObjectWithSourceAndTarget` because the cases that it tests are no longer valid. This test was checking for foreign key relationships when there were missing fields for one of source/target or linkingSource/linkingTarget, which is no longer valid in our config. Related cases which are valid such that both source/linkingSource or both target/linkingTarget are null are included within `TestRelationshipWithLinkingObjectNotHavingRequiredFields`. Because of this, we remove `TestRelationshipForCorrectPairingOfLinkingObjectWithSourceAndTarget`, and will add in additional unit testing for many to many relationships in the follow up PR related to https://github.com/Azure/data-api-builder/issues/2144 as needed.

## How was this tested?

- [X] Unit Tests

We add to the `ConfigValidationUnitTests` class all of the cases from the above issue. We assert that the correct kind of exception is thrown, with the correct message, status code, and sub status code.

## Sample Request(s)

You can manually test this by creating configs that violate the rules outlined in the above issue. Or by stepping through the new unit tests which do the same, but programmatically in the tests.


src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -403,6 +401,9 @@ string relationshipEntity
             _sqlMetadataProvider.Setup(x =>
                 x.ParseSchemaAndDbTableName("TEST_SOURCE_LINK")).Returns(("dbo", "TEST_SOURCE_LINK"));
 
+            string discard;
+            _sqlMetadataProvider.Setup(x => x.TryGetBackingColumn(It.IsAny<string>(), It.IsAny<string>(), out discard)).Returns(true);
+
```
seantleonard, line 406: assuming you need `discard` because `_` isn't valid syntax for Moq?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -437,21 +438,21 @@ string relationshipEntity
...
             DisplayName = "LinkingTargetField is null, only ForeignKeyPair between LinkingObject and source is present. Invalid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, false, false,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, false, false,
             DisplayName = "TargetField is null, , only ForeignKeyPair between LinkingObject and source is present. Invalid Case.")]
-        [DataRow(new string[] { "id" }, null, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity1", true, false, true,
             DisplayName = "LinkingSourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity1", true, false, true,
```
seantleonard, line 451: for this test, your modifications null out what was previously an array with a column name. Does that require any modification to the datarow display names because the tests are now different?

seantleonard (commented):
comments so far. Still have more of the test file to review. `RuntimeConfigValidator.cs` looks good though!

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
aaronburtle: Right, I can't put it directly into the MOQ call.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
aaronburtle: These cases are not really valid, where we have only target or source fields as null while the other is non null, because this PR adds validation to ensure that this throws an exception, and we handle that case in the unit tests I added.

I am trying to modify these tests minimally to avoid collisions with that new validation logic.

aaronburtle: I am still going through the updated tests and trying to get the language in its best form to properly describe the cases, since theyve been modified a bit.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,129 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                // Validation to ensure that if target fields exist, source fields exist as well.
+                if (relationship.TargetFields is not null && relationship.SourceFields is null)
+                {
+                    HandleOrRecordException(new DataApiBuilderException(
+                        message: $"Entity: {entityName} has target fields that are not null, but source fields that are null.",
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                }
+
+                if (relationship.SourceFields is not null && relationship.TargetFields is not null)
```
ayush3797, line 856: This check is redundant I guess.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,129 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                        }
+                    }
+
+                    foreach (string linkingTargetField in relationship.LinkingTargetFields)
+                    {
+                        if (!sqlMetadataProvider.TryGetBackingColumn(relationship.TargetEntity, linkingTargetField, out _))
+                        {
+                            // Validation to ensure that entities that have linking target fields defined, define linking target fields
+                            // that are valid columns in the target entity.
```
ayush3797, line 946:
```suggestion
                            // Validation to ensure that linking target fields defined for a target entity, are valid columns present in the target entity.
```

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -631,6 +633,323 @@ string linkingObject
...
         }
 
+        [DataRow(new[] { "non null" }, null, "Entity: SampleEntity1 has source fields that are not null, but target fields that are null.",
+            DisplayName = "sourceFields exist but targetFields are null")]
+        [DataRow(null, new[] { "non null" }, "Entity: SampleEntity1 has target fields that are not null, but source fields that are null.",
+            DisplayName = "targetFields exist but sourceFields are null")]
+        [DataRow(new[] { "A", "B", "C" }, new[] { "1", "2" }, "Entity: SampleEntity1 has 3 source fields defined, but 2 target fields defined.",
+            DisplayName = "sourceFields and targetFields have different length.")]
+        [DataTestMethod]
+        public void TestRelationshipWithoutSourceAndTargetFieldsMatching(
```
ayush3797, line 643: Missing test summary

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -631,6 +633,323 @@ string linkingObject
...
+            // Exception is thrown since sourceFields and targetFields do not match in either their existence,
+            // or their length.
+            DataApiBuilderException ex = Assert.ThrowsException<DataApiBuilderException>(() =>
+                configValidator.ValidateRelationshipsInConfig(runtimeConfig, _metadataProviderFactory.Object));
+            Assert.AreEqual(expectedExceptionMessage, ex.Message);
+            Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
+            Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ConfigValidationError, ex.SubStatusCode);
+        }
+
+        [DataRow(new[] { "field without backing column" }, new[] { "field with backing column" },
```
ayush3797, line 703: nit: Please use a single literal for backing column names, something like - `NonExistentBackingColumn` which is more like an actual backing column name rather than a string.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -631,6 +633,323 @@ string linkingObject
...
+        [DataRow(
+            new string[] { "sourceField" },
+            new string[] { "A", "B", "C" },
+            new string[] { "targetField" },
+            new string[] { "1", "2" },
+            "SampleEntity2",
+            "Entity: SampleEntity1 has 3 linking source fields defined, but 2 linking target fields defined.",
+            DisplayName = "Linking source fields and linking target fields are different length.")]
+        [DataTestMethod]
+        public void TestRelationshipWithoutLinkingSourceAndTargetFieldsMatching(
```
ayush3797, line 801: Test summary here and for other tests missing it.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -631,6 +633,323 @@ string linkingObject
...
+            // Exception thrown as linking source fields exist, but linking target fields are null.
+            DataApiBuilderException ex = Assert.ThrowsException<DataApiBuilderException>(() =>
+                configValidator.ValidateRelationshipsInConfig(runtimeConfig, _metadataProviderFactory.Object));
+            Assert.AreEqual(expectedExceptionMessage, ex.Message);
+            Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
+            Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ConfigValidationError, ex.SubStatusCode);
+        }
+
+        [DataRow(
+            new string[] { "field with backing column" },
```
ayush3797, line 868: Same suggestion to use a single literal for backing column names everywhere.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,129 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
                 {
                     HandleOrRecordException(new DataApiBuilderException(
-                        message: $"entity: {relationship.TargetEntity} is disabled for GraphQL.",
+                        message: $"Entity: {relationship.TargetEntity} is disabled for GraphQL.",
                         statusCode: HttpStatusCode.ServiceUnavailable,
                         subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
                 }
 
+                // Validation to ensure that if source fields exist, target fields exist as well.
+                if (relationship.SourceFields is not null && relationship.TargetFields is null)
```
ayush3797, line 839: Nit/suggestion: The number of && conditionals in the IF statements can be significantly reduced if we start with the condition:

`if (relationship.SourceFields is not null || relationship.TargetFields is not null)`, because then only we need to do the validations. All the other cases will go inside this if condition. If this condition is returns false, we don't need to do any validations at all.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -631,6 +633,323 @@ string linkingObject
...
+                sourceEntity: "SampleEntity1",
+                targetEntity: "SampleEntity2",
+                sourceFields: sourceFields,
+                targetFields: targetFields,
+                linkingObject: "TEST_SOURCE_LINK",
+                linkingSourceFields: linkingSourceFields,
+                linkingTargetFields: linkingTargetFields
+            );
+
+            RuntimeConfig runtimeConfig = new(
```
ayush3797, line 820: The code for creating a runtimeconfig seems to be duplicate across tests. Can this be moved to some common helper? Also, other existing tests which also create config in a similar manner can reuse that helper.

ayush3797 (commented):
Looks very good functionally. Left some nits/suggestions to consider.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: It is a good suggestion, and I considered some other options, but I think this way is most readable, and the conditionals end up describing the cases well.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: I added some more clarification. But I want to keep the language clear that the `linking target fields` are being defined in the relationship of the current entity we are looking at. We don't look at the linking target fields as defined in the relationship section of the target entity in the config, but rather the linking target fields that are defined in the current entity we are validating, and then verify they exist as backing columns for the source of the target entity.

Just want to be as clear as possible here because the usage of "target" exists in both the source and target entity, so it can be easy to confuse the two (for me at least).

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
aaronburtle: There is a lot of code duplication in our tests, which is OK, but also should be refactored out for more readability. I will either refactor this out, or maybe include this in a larger issue that tracks readability for tests more generally, as this is one of many, many cases and as you point out includes other tests in this class too.

ayush3797: Makes sense. Please open up an issue tracking this refactor.

ayush3797 (approved):
Thanks for the quick fix!

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
aaronburtle: issue here https://github.com/Azure/data-api-builder/issues/2143

seantleonard: #2143

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -437,21 +438,21 @@ string relationshipEntity
...
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity1", true, true, true,
             DisplayName = "LinkingSourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity1", true, true, true,
             DisplayName = "SourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, new string[] { "num" }, null, "SampleEntity2", false, true, true,
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity2", true, true, true,
             DisplayName = "LinkingTargetField is null, only ForeignKeyPair between LinkingObject and source is present. Valid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", false, true, true,
```
seantleonard, line 454: "TargetField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case."

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -437,21 +438,21 @@ string relationshipEntity
...
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity1", false, true, false,
             DisplayName = "SourceField is null, only ForeignKeyPair between LinkingObject and target is present.  Invalid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, new string[] { "num" }, null, "SampleEntity2", true, false, false,
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity2", true, false, false,
             DisplayName = "LinkingTargetField is null, only ForeignKeyPair between LinkingObject and source is present. Invalid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, false, false,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, false, false,
             DisplayName = "TargetField is null, , only ForeignKeyPair between LinkingObject and source is present. Invalid Case.")]
-        [DataRow(new string[] { "id" }, null, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
```
seantleonard, line 448: LinkingSourceField is null, only ForeignKeyPair between LinkingObject and source is present. Valid Case

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -437,21 +438,21 @@ string relationshipEntity
...
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity1", true, true, true,
             DisplayName = "LinkingSourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity1", true, true, true,
             DisplayName = "SourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, new string[] { "num" }, null, "SampleEntity2", false, true, true,
```
seantleonard, line 452: DisplayName = "LinkingTargetField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -437,21 +438,21 @@ string relationshipEntity
...
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity2", true, false, false,
             DisplayName = "LinkingTargetField is null, only ForeignKeyPair between LinkingObject and source is present. Invalid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, false, false,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, false, false,
             DisplayName = "TargetField is null, , only ForeignKeyPair between LinkingObject and source is present. Invalid Case.")]
-        [DataRow(new string[] { "id" }, null, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity1", true, true, true,
             DisplayName = "LinkingSourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
```
seantleonard, line 450: DisplayName = "SourceField is null, only ForeignKeyPair between LinkingObject and source is present. Valid Case.")]

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -437,21 +438,21 @@ string relationshipEntity
...
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity1", false, true, false,
             DisplayName = "LinkingSourceField is null, only ForeignKeyPair between LinkingObject and target is present. Invalid Case.")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", false, true, false,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity1", false, true, false,
             DisplayName = "SourceField is null, only ForeignKeyPair between LinkingObject and target is present.  Invalid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, new string[] { "num" }, null, "SampleEntity2", true, false, false,
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity2", true, false, false,
             DisplayName = "LinkingTargetField is null, only ForeignKeyPair between LinkingObject and source is present. Invalid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, false, false,
```
seantleonard, line 446: TargetField is null, only ForeignKeyPair between LinkingObject and source is present. Invalid Case.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,109 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                            // Validation to ensure that entities that have target fields defined, define target fields
+                            // that are valid columns in the target entity.
+                            HandleOrRecordException(new DataApiBuilderException(
+                                message: $"Entity: {entityName} has a relationship: {relationshipName} with target field: {targetField} that " +
+                                    $"does not exist as a column in target entity: {relationship.TargetEntity}.",
+                                statusCode: HttpStatusCode.ServiceUnavailable,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                        }
+                    }
+                }
```
seantleonard, line 901: these may need to fall under new condition for linking metadata null

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,109 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+
+                    // Validation to ensure that if target fields exist, source fields exist as well.
+                    if (relationship.TargetFields is not null && relationship.SourceFields is null)
+                    {
+                        HandleOrRecordException(new DataApiBuilderException(
+                            message: $"Entity: {entityName} has target fields that are not null, but source fields that are null.",
+                            statusCode: HttpStatusCode.ServiceUnavailable,
+                            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                    }
+
```
seantleonard, line 860:
```suggestion
```

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,108 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                {
+                    // Validation to ensure that if linking target fields exist that linking source fields exist as well.
+                    HandleOrRecordException(new DataApiBuilderException(
+                        message: $"Entity: {entityName} has a relationship: {relationshipName} with linking target fields that are not null, " +
+                            $"but linking source fields that are null.",
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                }
+
+                if (relationship.LinkingSourceFields is not null && relationship.LinkingTargetFields is not null)
```
seantleonard, line 922: asymmetrical existence to be moved to rc2 because this info may be resolved from FKDefinition resolved from db

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -792,33 +794,15 @@ string linkingObject
...
-            DisplayName = "Linking source fields are non null, but linking target fields are null.")]
-        [DataRow(
-            new string[] { "sourceField" },
-            null,
-            new string[] { "targetField" },
-            new string[] { "nonNull" },
-            "SampleEntity2",
-            "Entity: SampleEntity1 has a relationship: rname1 with linking target fields that are not null, " +
-                "but linking source fields that are null.",
-            DisplayName = "Linking target fields are non null, but linking source fields are null.")]
```
seantleonard: with this removal, test no longer needs to be datatestmethod

src/Core/Configurations/RuntimeConfigValidator.cs
seantleonard: fixed by removal. deferred to rc2

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -545,27 +551,20 @@ bool isValidScenario
 
         /// <summary>
         /// Test method to check that an exception is thrown when LinkingObject is null
-        /// and either of SourceFields and TargetFields is null in the config.
+        /// and SourceFields and TargetFields are null in the config.
         /// And the foreignKey pair between source and target is not defined in the database as well.
         /// Also verify that after adding foreignKeyPair in the Database, no exception is thrown.
         /// </summary>
-        [DataRow(null, new string[] { "das" }, null, DisplayName = "SourceFields is null")]
-        [DataRow(new string[] { "id" }, null, null, DisplayName = "TargetFields is null")]
```
seantleonard, line 553: these tests need to be added back. Their presence for this specific test implies that a mismatch in sourcefield and target field count is fine as long as a FK is defined in the database which would ultimately provide the missing metadata to dab.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -545,27 +551,20 @@ bool isValidScenario
...
-        [DataRow(null, null, null, DisplayName = "both source and targetFields are null")]
         [DataTestMethod]
-        public void TestRelationshipWithNoLinkingObjectAndEitherSourceOrTargetFieldIsNull(
-            string[] sourceFields,
-            string[] targetFields,
-            string linkingObject
-        )
+        public void TestRelationshipWithNoLinkingObjectAndEitherSourceOrTargetFieldIsNull()
```
seantleonard, line 559: Test description should be updated along with test method name. this test doesn't look like it has anything to do with "linkingObject" since `linkingObject` , `linkingSourceFields`, and `linkingTargetFields` are null for all tests. This seems to be specifically address:

1. `sourceFields` and `targetFields` having mismatched column counts - without FK relating source and target entity.
2. When FK is added to relate source and target entity, DAB then has metadata to fill in the gaps and allow the scenario.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
seantleonard: This config is valid, test in your environment:

```json
{
  "data-source": {
    "database-type": "mssql",
    "connection-string": "<CONN_STRING>",
    "options": {
      "set-session-context": true
    }
  },
  "runtime": {
    "rest": {
      "enabled": true,
      "path": "/api",
      "request-body-strict": true
    },
    "graphql": {
      "enabled": true,
      "path": "/graphql",
      "allow-introspection": true,
      "multiple-mutations": {
        "create": {
          "enabled": true
        }
      }
    },
    "host": {
      "cors": {
        "origins": [
          "http://localhost:5000"
        ],
        "allow-credentials": false
      },
      "authentication": {
        "provider": "Simulator"
      },
      "mode": "development"
    }
  },
  "entities": {
    "Publisher": {
      "source": {
        "object": "publishers",
        "type": "table"
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "Publisher",
          "plural": "Publishers"
        }
      },
      "rest": {
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            "*"
          ]
        }
      ],
      "relationships": {
        "books": {
          "cardinality": "many",
          "target.entity": "Book",
          "source.fields": ["id"],
          "target.fields": [],
          "linking.source.fields": [],
          "linking.target.fields": []
        }
      }
    },
    "Book": {
      "source": {
        "object": "books",
        "type": "table"
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "book",
          "plural": "books"
        }
      },
      "rest": {
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            "*"
          ]
        }
      ],
      "mappings": {
        "id": "id",
        "title": "title"
      },
      "relationships": {
        "publishers": {
          "cardinality": "one",
          "target.entity": "Publisher",
          "source.fields": [],
          "target.fields": [],
          "linking.source.fields": [],
          "linking.target.fields": []
        }
      }
    }
  }
}

```

note: `publisher.relationship.books` has the following contents for source/target fields are mismatched count:
```json
      "relationships": {
        "books": {
          "cardinality": "many",
          "target.entity": "Book",
          "source.fields": ["id"],
          "target.fields": [],
          "linking.source.fields": [],
          "linking.target.fields": []
        }
```

The DB has a FK for this relationship, which dab resolves and then allows this config because the db metadata makes up for the missing `target.fields` value.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
seantleonard: correction: test looks to directly exercise this validation when all three referenced values are null (this is the test case that you left in)
```csharp
                if (relationship.LinkingObject is null
                    && (relationship.SourceFields is null || relationship.TargetFields is null))
                {
                    if (!sqlMetadataProvider.VerifyForeignKeyExistsInDB(sourceDatabaseObject, targetDatabaseObject))
                    {
                        HandleOrRecordException(new DataApiBuilderException(
                            message: $"Could not find relationship between entities: {entityName} and {relationship.TargetEntity}.",
                            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
                    }
                }
```

The next section of code is what the two deleted test cases exercise: essentially supplementing the user provided source/target fields with the source/target fields defined in the FK definition. this code doesn't throw any exceptions and looks to simply log what it finds.
```csharp
if (relationship.LinkingObject is null && !_runtimeConfigProvider.IsLateConfigured)
{
    RelationShipPair sourceTargetRelationshipPair = new(sourceDatabaseObject, targetDatabaseObject);
    RelationShipPair targetSourceRelationshipPair = new(targetDatabaseObject, sourceDatabaseObject);
    string sourceDBOName = sqlMetadataProvider.EntityToDatabaseObject[entityName].FullName;
    string targetDBOName = sqlMetadataProvider.EntityToDatabaseObject[relationship.TargetEntity].FullName;
    string cardinality = relationship.Cardinality.ToString().ToLower();
    ForeignKeyDefinition? fKDef;
    string sourceColumns = relationship.SourceFields is not null ? string.Join(",", relationship.SourceFields) :
        sqlMetadataProvider.PairToFkDefinition!.TryGetValue(sourceTargetRelationshipPair, out fKDef) ?
        string.Join(",", fKDef.ReferencingColumns) :
        sqlMetadataProvider.PairToFkDefinition!.TryGetValue(targetSourceRelationshipPair, out fKDef) ?
        string.Join(",", fKDef.ReferencedColumns) : string.Empty;
    string targetColumns = relationship.TargetFields is not null ? string.Join(",", relationship.TargetFields) :
        sqlMetadataProvider.PairToFkDefinition!.TryGetValue(sourceTargetRelationshipPair, out fKDef) ?
        string.Join(",", fKDef.ReferencedColumns) :
        sqlMetadataProvider.PairToFkDefinition!.TryGetValue(targetSourceRelationshipPair, out fKDef) ?
        string.Join(",", fKDef.ReferencingColumns) : string.Empty;
```

seantleonard (changes requested):
Identified additional test cases that need to be restored. Additionally, the conditionals that gate your validations may need to consider whether existing FK definitions exist to supplement the user provided data because it seems to be supported today. By adding validation to block a currently supported config, we'd be introducing a regression.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
seantleonard: intent is to use config target/source only and if not defined use db metadata. DB metadata will not fill in the blanks in config where fields are erroneously left out.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
seantleonard: db metadata should NOT supplement config data. config relationship data (source/target) fields OVERRIDE database fk metadata. This is what our documentation implies and instructs users to fill out their config if they dont want to use what the db fk metadata defines.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -437,21 +438,21 @@ string relationshipEntity
...
             DisplayName = "LinkingSourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", true, false, true,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity1", true, true, true,
             DisplayName = "SourceField is null, only ForeignKeyPair between LinkingObject and target is present. Valid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, new string[] { "num" }, null, "SampleEntity2", false, true, true,
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity2", true, true, true,
             DisplayName = "LinkingTargetField is null, only ForeignKeyPair between LinkingObject and source is present. Valid Case.")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", false, true, true,
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", true, true, true,
```
seantleonard, line 455: These display names aren't accurate because they don't align with your updates. They state "only ForeignKeyPair between" even when BOTH `isForeignKeyPairBetSourceAndLinkingObject` and `isForeignKeyPairBetTargetAndLinkingObject` are true.

With your new validations, can you clarify the purpose of this test by updating this test method's summary? 
The testmethod name is `TestRelationshipForCorrectPairingOfLinkingObjectWithSourceAndTarget` so a summary that answers:
1. what is a "correct pairing"
2. what is the purpose of declaring `isForeignKeyPairBet{Target/Source}AndLinkingObject` as true? 
    - why does false affect the test results?  
    - what effect do your new validations have on this?

Everything else looks good. Let's use this opportunity to make this test more readable for everyone.

aaronburtle: We've been discussing this offline, but I wanted to follow up with the information here in this PR. This test doesn't seem to be valid given the behavior we have decided on.

In every case, this test presumes that we have a state where we are missing some collection of fields, which according to our docs, and according to the behavior of the engine, is not valid. We can't have missing linkingSource/linkingTarget fields, and we can't have missing source/Target fields, nor can we have a case where all of that is missing. We must include those fields for the engine to be able to correctly form the relationships in our own necessary data models.

Because of that I think we should remove this test entirely, and then add back in the many to many validations that make sense in the follow up PR that covers the linking object being added to our engine's data model prior to validation of the config

seantleonard: agreed

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -430,39 +431,17 @@ string relationshipEntity
...
 
         /// <summary>
-        /// Test method to check that an exception is thrown when relationship exists between the wrong pair. i.e.,
-        /// LinkingSourceField OR SourceFields are null and relationship between target and linking object exists, no relationship between source and linking object
-        /// LinkingTargetField OR TargetField are null but relationship between source and linking object exists, no relationship between target and linking object
-        /// Also Test when the foreignKeyPair exist for correct pair, no exception is thrown. i.e.,
-        /// LinkingSourceField OR SourceFields are null but relationship between source and linking object exists, no relationship between target and linking object
-        /// LinkingTargetField OR TargetField are null but relationship between target and linking object exists, no relationship between source and linking object
+        /// Test method to check that an exception is thrown when LinkingObject is null
+        /// and SourceFields and TargetFields are null in the config.
```
seantleonard, line 435: Test doesn't need to be data row test anymore

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -332,20 +332,18 @@ public void TestAddingRelationshipWithDisabledGraphQL()
...
+        /// or LinkingTargetField is null, while the relationship is not defined in the database.
         /// Also verify that after adding foreignKeyPair in the Database, no exception is thrown.
         /// </summary>
-        [DataRow(new string[] { "id" }, null, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", DisplayName = "LinkingSourceField is null")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", DisplayName = "SourceField is null")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, new string[] { "num" }, null, "SampleEntity2", DisplayName = "LinkingTargetField is null")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", DisplayName = "TargetField is null")]
+        [DataRow(new string[] { "id" }, null, new string[] { "num" }, null, "SampleEntity1", DisplayName = "LinkingSourceField is null")]
+        [DataRow(null, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity1", DisplayName = "SourceField is null")]
```
seantleonard, line 346: need updated displaynames.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -430,50 +431,23 @@ string relationshipEntity
...
-            string[] sourceFields,
-            string[] linkingSourceFields,
-            string[] targetFields,
-            string[] linkingTargetFields,
-            string relationshipEntity,
-            bool isForeignKeyPairBetSourceAndLinkingObject,
-            bool isForeignKeyPairBetTargetAndLinkingObject,
-            bool isValidScenario
-        )
+        public void TestRelationshipWithNoLinkingObjectAndEitherSourceOrTargetFieldIsNull()
```
seantleonard, line 440: test method attribute should be `[TestMethod]`

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -430,50 +431,23 @@ string relationshipEntity
...
-        /// LinkingSourceField OR SourceFields are null and relationship between target and linking object exists, no relationship between source and linking object
-        /// LinkingTargetField OR TargetField are null but relationship between source and linking object exists, no relationship between target and linking object
-        /// Also Test when the foreignKeyPair exist for correct pair, no exception is thrown. i.e.,
-        /// LinkingSourceField OR SourceFields are null but relationship between source and linking object exists, no relationship between target and linking object
-        /// LinkingTargetField OR TargetField are null but relationship between target and linking object exists, no relationship between source and linking object
+        /// Test method to check that an exception is thrown when LinkingObject is null
+        /// and SourceFields and TargetFields are null in the config.
+        /// And the foreignKey pair between source and target is not defined in the database as well.
+        /// Also verify that after adding foreignKeyPair in the Database, no exception is thrown.
```
seantleonard, line 437: adding ForeignKeyPair for which tables? between source and target? between Source/Linking Object and Target/Linking Object?

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -430,50 +431,23 @@ string relationshipEntity
         }
 
         /// <summary>
-        /// Test method to check that an exception is thrown when relationship exists between the wrong pair. i.e.,
-        /// LinkingSourceField OR SourceFields are null and relationship between target and linking object exists, no relationship between source and linking object
-        /// LinkingTargetField OR TargetField are null but relationship between source and linking object exists, no relationship between target and linking object
-        /// Also Test when the foreignKeyPair exist for correct pair, no exception is thrown. i.e.,
-        /// LinkingSourceField OR SourceFields are null but relationship between source and linking object exists, no relationship between target and linking object
-        /// LinkingTargetField OR TargetField are null but relationship between target and linking object exists, no relationship between source and linking object
+        /// Test method to check that an exception is thrown when LinkingObject is null
```
seantleonard, line 434: It would be clearer to explicitly call out that this test validates config written for 1:many, many:1, 1:1 and NOT many:many. Many:Many is excluded from this test because we intentionally don't define a linking object/linking fields. 

You're essentially testing that this config is valid when a fk def exists in the database relating series and book. and that this config results in startup failure when a fk def doesn't exist in the database relating series and book.

```json
"Series": {
  "source": "dbo.series",
  "relationships": {
    "books": {
      "cardinality": "many",
      "target.entity": "Book",
      "source.fields": null,
      "target.fields": null
    }    
  }
}
```

seantleonard: in other words, it's a little confusing when the first condition that this comment describes is the lacking of **linking** object.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -332,20 +332,18 @@ public void TestAddingRelationshipWithDisabledGraphQL()
...
+            Assert.AreEqual($"Entity: {sampleRelationship.TargetEntity} is disabled for GraphQL.", ex.Message);
             Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
         }
 
         /// <summary>
         /// Test method to check that an exception is thrown when LinkingObject was provided
-        /// while either LinkingSourceField or SourceField is null, and either targetFields or LinkingTargetField is null.
-        /// And the relationship is not defined in the database.
+        /// while either LinkingSourceField or SourceField is null, and the corresponding targetFields
+        /// or LinkingTargetField is null, while the relationship is not defined in the database.
```
seantleonard, line 342: which relationship? the fkdefinition for source to target or between linkingobject::source and linkingobject::target?

aaronburtle:
Some more clarifications in some design choices we recently made.

Using the config to define the relationship via the fields that are related to one another happens at the **foreign key level**. This means that in a many to many relationship you can 

* define the entire relationship with the fields in the config (ie: providing all of source/target and linkingSource/linkingTarget)
* define only one side of the relationship (ie: providing only source and linkingSource or target and linkingTarget)
* define neither sides of the relationship

However, it must be the case that for any missing fields in the config, that a foreign key relationship exists for us to correctly interpret what those fields ought to be.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
                         statusCode: HttpStatusCode.ServiceUnavailable,
                         subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
                 }
 
+                // Linking object is null and therefore we have either a many to one or a one to many relationship. These relationships
+                // must be validated separately from many to many relationships. In one to many and many to one relationships, source
+                // fields and target fields need to match since they define the relationship between the two entities in its entirely.
+                // If both of these sets of fields are null, foreign key information will be used to define the relationship instead.
+                // see: https://learn.microsoft.com/en-us/azure/data-api-builder/relationships
+                if (relationship.LinkingObject is null)
```
seantleonard, line 843: Should this be string.IsNullorWhitespace? e.g. `linkingObject: ""` wouldn't meet this condition, would it?

severussundar (changes requested):
The validation logic seems comprehensive. Kudos for identifying and incorporating that in the case of linking field validations, it is okay to have one of (<source.fields, linking.sourcefields>, <target.fields, linking.targetfields>) specified through the config while the other can be inferred from the database metadata because there are 2 foreign keys invloved - <source table, linking table> and <target table, linking table>.

- In the `*-commands.txt` file, the command for creating the `authors` relationship in the `Book` entity, does not use `--relationship.fields` but uses `--linking.source.fields` and `--linking.target.fields` options. As a result of this, in the config file, the `sourceFields` and `targetFields` would be empty while `linking.source-fields` and `linking.target-fields` would be non-empty.  

This should've errored out during engine startup because of the validations added. One possibility could be that these validations run only when `dab validate` is executed. We should run these validations during engine start-up too to catch errors when users hand-edit the config file.

- The reference config files that are present in our repo - `dab-config.*.json` also needs to be updated to reflect equal number of source fields, linking source fields

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
                     HandleOrRecordException(new DataApiBuilderException(
-                        message: $"entity: {relationship.TargetEntity} is disabled for GraphQL.",
+                        message: $"Entity: {relationship.TargetEntity} is disabled for GraphQL.",
                         statusCode: HttpStatusCode.ServiceUnavailable,
                         subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
                 }
 
+                // Linking object is null and therefore we have either a many to one or a one to many relationship. These relationships
+                // must be validated separately from many to many relationships. In one to many and many to one relationships, source
+                // fields and target fields need to match since they define the relationship between the two entities in its entirely.
```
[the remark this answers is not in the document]
seantleonard: or:
> In one-many and many-one relationships, the count of source and target fields needs to match.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                    // Validation to ensure that if target fields exist, source fields exist as well.
+                    if (relationship.TargetFields is not null && relationship.SourceFields is null)
+                    {
+                        HandleOrRecordException(new DataApiBuilderException(
+                            message: $"Entity: {entityName} has target fields that are not null, but source fields that are null.",
+                            statusCode: HttpStatusCode.ServiceUnavailable,
+                            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                    }
+
+                    // In one:many or many:one relationships when both source and target are non null their size must match.
```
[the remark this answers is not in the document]
seantleonard: Aaron, you can pick which usage you like better:  "one:one" or "one-one". Though be consistent throughout PR.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                        // Validation to ensure that if target fields exist that linking target fields exist as well.
+                        HandleOrRecordException(new DataApiBuilderException(
+                            message: $"Entity: {entityName} has a many to many relationship: {relationshipName} with target fields that are not null, " +
+                                $"but linking target fields that are null.",
+                            statusCode: HttpStatusCode.ServiceUnavailable,
+                            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                    }
+
+                    // Validation to ensure that if linking target fields exist that targets fields exist as well.
+                    if (relationship.LinkingTargetFields is not null && relationship.TargetFields is null)
```
Aniruddh25, line 949: All these validations are between 2 pairs of fields:
1) SourceFields, TargetFields
2) SourceFields, LinkingSourceFields
3) TargetFields, LinkingTargetFields 

- there can easily be defined an helper function that takes a pair to validate. That helper can be called 3 times to reduce code duplication.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                        {
+                            // Validation to ensure that if source and linking source fields exist, that they have the same number of fields. 
+                            HandleOrRecordException(new DataApiBuilderException(
+                                message: $"Entity: {entityName} has a many to many relationship: {relationshipName} with {relationship.TargetFields.Length} " +
+                                    $"target fields defined, but {relationship.LinkingTargetFields.Length} linking target fields defined.",
+                                statusCode: HttpStatusCode.ServiceUnavailable,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                        }
+                    }
+                }
```
Aniruddh25, line 986: Given that we have added new validations here, the validations that are happening below this code on line 1022 seem unnecessary.. Can you please check if they are still necessary, remove if not?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                    if (relationship.SourceFields is not null && relationship.TargetFields is null)
+                    {
+                        HandleOrRecordException(new DataApiBuilderException(
+                            message: $"Entity: {entityName} has source fields that are not null, but target fields that are null.",
+                            statusCode: HttpStatusCode.ServiceUnavailable,
+                            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                    }
+
+                    // Validation to ensure that if target fields exist, source fields exist as well.
+                    if (relationship.TargetFields is not null && relationship.SourceFields is null)
```
Aniruddh25, line 855: We can do an EXOR operation to check if the nullability of TargetFields/SourceFields doesnt match. This will reduce the number of separate `if` conditions.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -403,6 +409,9 @@ string relationshipEntity
             _sqlMetadataProvider.Setup(x =>
                 x.ParseSchemaAndDbTableName("TEST_SOURCE_LINK")).Returns(("dbo", "TEST_SOURCE_LINK"));
 
+            string discard;
+            _sqlMetadataProvider.Setup(x => x.TryGetBackingColumn(It.IsAny<string>(), It.IsAny<string>(), out discard)).Returns(true);
```
Aniruddh25, line 413: if we are discarding, doesn't `out _` work here?

Aniruddh25 (dismissed):
Posting comments so far. Still need to review the test

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                                    $"target fields defined, but {relationship.LinkingTargetFields.Length} linking target fields defined.",
+                                statusCode: HttpStatusCode.ServiceUnavailable,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                        }
+                    }
+                }
+
                 // Validation to ensure DatabaseObject is correctly inferred from the entity name.
                 DatabaseObject? sourceObject, targetObject;
                 if (!sqlMetadataProvider.EntityToDatabaseObject.TryGetValue(entityName, out sourceObject))
```
abhishekkumams, line 990: line 1046, where we are just doing _logger.Debug, can we move this out of this method? this method is too big now.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                    }
+                }
+
+                // Linking object exists and we therefore have a many to many relationship. Validation here differs from one to many and many to one in that
+                // the source and target fields are now only indirectly related through the linking object. Therefore, it is the source and linkingSource
+                // along with the target and linkingTarget fields that must match, respectively. Source and linkingSource fields provide the relationship
+                // from the source entity to the linkingObject while target and linkingTarget fields provide the relationship from the target entity to the
+                // linkingObject.
+                // see: https://learn.microsoft.com/en-us/azure/data-api-builder/relationships#many-to-many-relationship
+                if (relationship.LinkingObject is not null)
```
abhishekkumams, line 916: line 916 to line 986, can be shortened. there is many repetative code.
```
if (relationship.LinkingObject is not null)
{
    ValidateFields(relationship.SourceFields, relationship.LinkingSourceFields, "source", entityName, relationshipName);
    ValidateFields(relationship.TargetFields, relationship.LinkingTargetFields, "target", entityName, relationshipName);
}

private void ValidateFields(string[]? fields, string[]? linkingFields, string fieldType, string entityName, string relationshipName)
{
    if (fields is not null && linkingFields is null)
    {
        HandleOrRecordException(CreateException($"{fieldType} fields", "linking " + fieldType + " fields", entityName, relationshipName));
    }

    if (linkingFields is not null && fields is null)
    {
        HandleOrRecordException(CreateException("linking " + fieldType + " fields", $"{fieldType} fields", entityName, relationshipName));
    }

    if (fields is not null && linkingFields is not null && fields.Length != linkingFields.Length)
    {
        HandleOrRecordException(new DataApiBuilderException(
            message: $"Entity: {entityName} has a many to many relationship: {relationshipName} with {fields.Length} " +
                $"{fieldType} fields defined, but {linkingFields.Length} linking {fieldType} fields defined.",
            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
    }
}

private DataApiBuilderException CreateException(string nonNullFields, string nullFields, string entityName, string relationshipName)
{
    return new DataApiBuilderException(
        message: $"Entity: {entityName} has a many to many relationship: {relationshipName} with {nonNullFields} that are not null, " +
            $"but {nullFields} that are null.",
}
```

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: They are needed, although our validation makes the conditional a bit redundant. To add context, we were checking `if relationship.LinkingSourceFields is null || relationship.SourceFields is null` before verifying foreign key information exists in the DB because we would "fix" a config that only provided one of the linking source fields or the source fields. We now validate that if one exists, so does the other, but we still want to check that the foreign key information is in the DB if one of them (which now guarantees so is the other) is null.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: thanks for these suggestions, Ive added the helper, but made the error message a little different so that we can use XOR in the comparison.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: I am not sure what you mean, this seems to log information about the relationship in non-hosted scenarios, do you want to remove the logging entirely?

image

aaronburtle: I am not sure what you mean by moving this out of this method, the information that is being logged here is taken from the calls made just previous to the logging. Where would it go if not here?

image

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +833,107 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+
+                        if (invalidColumns.Count > 0)
+                        {
+                            HandleOrRecordException(new DataApiBuilderException(
+                                message: $"Entity: {entityName} has a relationship: {relationshipName} with source fields: {string.Join(",", invalidColumns)} that " +
+                                    $"do not exist as columns in {entityName}.",
+                                statusCode: HttpStatusCode.ServiceUnavailable,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                            invalidColumns.Clear();
+                        }
```
seantleonard, line 889: When you clear `invalidColumns` after each occurence of detecting an invalid column, you'll never have `invalidColumns.Count` be greater than 1. So there doesn't seem to be a benefit to `string.Join(",", invalidColumns)`. Did you mean to have this if block be outside the for loop? 

If you intended the `if` block to be outside the for loop, you can further simplify the code checking for valid source and valid target fields.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -830,11 +830,161 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+                                statusCode: HttpStatusCode.ServiceUnavailable,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                        }
+                    }
+                }
+
+                // In all kinds of relationships, if targetFields are included they must be valid columns in the backend.
+                if (relationship.TargetFields is not null)
+                {
+                    foreach (string targetField in relationship.TargetFields)
```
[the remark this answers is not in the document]
seantleonard: how was this resolved? Shyam's ask is similiar to my comment above where placing the check for invalid columns outside the for loop will allow for simplification.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -332,20 +332,26 @@ public void TestAddingRelationshipWithDisabledGraphQL()
...
+        /// from that foreign key.
         /// </summary>
-        [DataRow(new string[] { "id" }, null, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", DisplayName = "LinkingSourceField is null")]
-        [DataRow(null, new string[] { "token_id" }, new string[] { "num" }, new string[] { "book_num" }, "SampleEntity1", DisplayName = "SourceField is null")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, new string[] { "num" }, null, "SampleEntity2", DisplayName = "LinkingTargetField is null")]
-        [DataRow(new string[] { "id" }, new string[] { "token_id" }, null, new string[] { "book_num" }, "SampleEntity2", DisplayName = "TargetField is null")]
+        [DataRow(null, null, new string[] { "targetFields" }, new string[] { "linkingTargetFields" }, "SampleEntity1",
+            DisplayName = "sourceFields and LinkingSourceFields are null")]
+        [DataRow(new string[] { "sourceFields" }, new string[] { "linkingSourceFields" }, null, null, "SampleEntity2",
+            DisplayName = "targetFields and LinkingTargetFields are null")]
```
seantleonard, line 354:
```suggestion
        [DataRow(null, null, new string[] { "targetField" }, new string[] { "linkingTargetField" }, "SampleEntity1",
            DisplayName = "sourceFields and LinkingSourceFields are null")]
        [DataRow(new string[] { "sourceField" }, new string[] { "linkingSourceField" }, null, null, "SampleEntity2",
            DisplayName = "targetFields and LinkingTargetFields are null")]
```

super nit: because you're only defining one field per array.

seantleonard, line 354:
```suggestion
        [DataRow(null, null, new string[] { "targetField" }, new string[] { "linkingTargetField" }, "SampleEntity1",
            DisplayName = "sourceFields and LinkingSourceFields are null")]
        [DataRow(new string[] { "sourceField" }, new string[] { "linkingSourceField" }, null, null, "SampleEntity2",
            DisplayName = "targetFields and LinkingTargetFields are null")]
```

super nit: because you're only defining one field per array.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -509,63 +492,142 @@ bool isValidScenario
...
+            Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ConfigValidationError, ex.SubStatusCode);
         }
 
         /// <summary>
-        /// Test method to check that an exception is thrown when LinkingObject is null
-        /// and either of SourceFields and TargetFields is null in the config.
-        /// And the foreignKey pair between source and target is not defined in the database as well.
-        /// Also verify that after adding foreignKeyPair in the Database, no exception is thrown.
+        /// This test methods ensures that our validation code catches the case where the listed source or target fields
+        /// are not valid backing columns in either the source or target entity respectively.
```
seantleonard, line 601:
```suggestion
        /// This test method ensures that our validation code catches the case where the listed source or target fields
        /// are not valid backing columns in either the source or target entity, respectively.
```

seantleonard (commented):
One last suggestion on simplification and question about organization of column validity check placement.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: Yes, this was supposed to be outside the loop, per our discussion, ive moved it all a helper function.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -977,6 +1059,83 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+        ISqlMetadataProvider sqlMetadataProvider)
+    {
+        invalidColumns.Clear();
+        foreach (string field in fields)
+        {
+            if (!sqlMetadataProvider.TryGetBackingColumn(entityName, field, out _))
+            {
+                invalidColumns.Add(field);
+            }
+        }
```
seantleonard, line 1093: most of these params are passed in just to be used in an error message.. any way that this helper just return a list of collected invalid columns? then the caller can see if there are any contents in teh invalid columns list and raise an exception accordingly.

aaronburtle: Good points, based on our call we have made this a lot simpler.

seantleonard (approved):
thanks for pushing through the iterations and feedback!

src/Core/Configurations/RuntimeConfigValidator.cs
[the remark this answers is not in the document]
seantleonard: this was resolved with code deduping.

src/Core/Configurations/RuntimeConfigValidator.cs
seantleonard: if there is cleanup needed outside these changes we can add to backlog.

PR #2153, opened 2024-04-04
Adding pagination limits.

rohkhann:
## Why make this change?

This change allows users to use the runtime config to provide limits on the number of records that can be retrieved through paginated calls. It also gives them the ability to define a default size which will be the size returned when no pagination input is given (no first in case of graphql and no limit in case of rest calls).

## What is this change?
It adds the nullable paginationoptions property to runtimeoptions of runtimeConfig. 
Default page size is set to 100. 
Max page size is set to 100,000 ( can be altered by the user in their runtimeconfig).

A call with -1 pagination input will result in max page size records being returned. Any call with a pagination number higher than max page size will be rejected with bad request.

In config:
default-page-size is an integer
default-page-size default is 100
default-page-size value -1 means "same as max-page-size"
default-page-size value 0 is an error
default-page-size value less than -1 is an error
default-page-size value more than than max-page-size is an error
max-page-size is an integer
max-page-size default is 100,000
max-page-size value -1 means "same as int.MaxValue"
max-page-size value 0 is an error
max-page-size value less than -1 is an error
max-page-size value more than int.MaxValue is an error

In a query (REST or GQL):
$first=-1 means "whatever the max-page-size value is"
$first=less than -1 is an error
$first=0 is an error
$first=(any value more than max-page-size) is an error

Sample configuration file:
```json
{
  "runtime": {
    "pagination": {
      "default-page-size": -1,
      "max-page-size": 1233
    }
  }
}
```

## How was this tested?
1. For default page value of 100, invalid page value of 0 or <-1 the existing test cases should cover. 
2. Added tests for both GQL and REST for above conditions.


src/Service.GraphQLBuilder/Queries/QueryBuilder.cs
```diff
@@ -24,6 +24,8 @@ public static class QueryBuilder
         public const string ORDER_BY_FIELD_NAME = "orderBy";
         public const string PARTITION_KEY_FIELD_NAME = "_partitionKeyValue";
         public const string ID_FIELD_NAME = "id";
+        public const int MAX_PAGE_LIMIT = 1000;
```
vadeveka, line 27: Can we reference constants from single place? I see PaginationOptions also has these.
Also, isnt MAX_PAGE_LIMIT suppoesed to be 10000?

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -577,12 +578,14 @@ public static JsonElement CreateNextLink(string path, NameValueCollection? query
         /// </summary>
         /// <param name="jsonResult">Results plus one extra record if more exist.</param>
         /// <param name="first">Client provided limit if one exists, otherwise 0.</param>
+        /// <param name="Default">Default limit if none provided.</param>
         /// <returns>Bool representing if more records are available.</returns>
-        public static bool HasNext(JsonElement jsonResult, uint? first)
+        public static bool HasNext(JsonElement jsonResult, int? first, uint defaultPageSize = PaginationOptions.DEFAULT_PAGE_SIZE, uint maxPageSize = PaginationOptions.MAX_PAGE_SIZE)
```
vadeveka, line 583: is this logic being repeated in src/Core/Resolvers/SqlResponseHelpers.cs? Could just keep either version

src/Service.GraphQLBuilder/Queries/QueryBuilder.cs
rohkhann: fixed. this was old un-referenced code that i missed removing. thanks for catching it.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -577,12 +577,14 @@ public static JsonElement CreateNextLink(string path, NameValueCollection? query
         /// </summary>
         /// <param name="jsonResult">Results plus one extra record if more exist.</param>
         /// <param name="first">Client provided limit if one exists, otherwise 0.</param>
+        /// <param name="Default">Default limit if none provided.</param>
```
vadeveka, line 580: Update documented params to match actual

vadeveka (approved):

seantleonard:
pr description says this works for both rest/graphql but only states config property is added for runtime.graphql. Is there also a config for runtime.rest?

Would also be good to know how in our pagination logic you convert $first=-1 into proper sql query. Just  value detection->replacement?

seantleonard:
please add example configuration json  to include new property in runtime section.

like this:

```json
{
  "runtime": {
    "pagination": {
      "default-page-size": -1,
      "max-page-size": 1233
    }
  }
}
```

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -886,6 +972,25 @@ public async Task RequestInvalidZeroFirst()
...
+            string graphQLQuery = @"{
+                books(first: 100001) {
+                    items {
+                        id
+                    }
+                }
+            }";
+
+            JsonElement result = await ExecuteGraphQLRequestAsync(graphQLQuery, graphQLQueryName, isAuthenticated: false);
+            SqlTestHelper.TestForErrorInGraphQLResponse(result.ToString(), statusCode: $"{DataApiBuilderException.SubStatusCodes.BadRequest}");
```
seantleonard, line 991: substaus code? do you specify a clarifying code that we can associate with invalid `first` value?

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1573,7 +1573,7 @@ public virtual async Task TestInvalidFirstParamQuery()
         {
             string graphQLQueryName = "books";
             string graphQLQuery = @"{
-                books(first: -1) {
+                books(first: -2) {
```
seantleonard, line 1576: add test comment that describes that this test ensures `first` value of <-1 results in request failure. may also need to add assert on substatus code if you emit a substatus code aligning with bad `first` param.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -762,6 +762,23 @@ public async Task FindTest_NoQueryParams_PaginationNextLink()
             );
         }
 
+        /// <summary>
+        /// Validates that a proper nextLink is created for FindMany requests which do not
+        /// restrict results with query parameters. Engine default paging mechanisms are used
+        /// when > 100 records will be present in result set.
```
seantleonard, line 768: what is the significance of 100 here? because you're setting new values for min max page size, let's specify which constant /configured property is referenced to assert on page size.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
```diff
@@ -1553,7 +1570,26 @@ public async Task FindTestWithFirstZeroSingleKeyPagination()
...
                 exceptionExpected: true,
-                expectedErrorMessage: "Invalid number of items requested, $first must be an integer greater than 0. Actual value: 0",
+                expectedErrorMessage: "Invalid number of items requested, $first must be -1 or an integer greater than 0. Actual value: 0",
+                expectedStatusCode: HttpStatusCode.BadRequest
+            );
+        }
+
+        /// <summary>
+        /// Tests the REST Api for Find operation using $first=100001
+        /// to request > max records of 100000 records, which should throw a DataApiBuilder
```
seantleonard, line 1580: nit: specify that this test depends on runtime config default of min/max page size property being set to xyz.

src/Core/Services/RequestValidator.cs
seantleonard, line 1: Please add a few unit tests to `ConfigValidationUnitTests.cs` to add coverage for the validations you add here. e.g. unexpected min/max page size. max< min etc.

seantleonard (changes requested):
some suggestions and comment clarifications requested.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -138,6 +139,28 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
...
+    public void ValidatePaginationOptions(RuntimeConfig runtimeConfig)
+    {
+        if (runtimeConfig.Runtime is not null && runtimeConfig.Runtime.Pagination is not null)
+        {
+            PaginationOptions paginationOptions = runtimeConfig.Runtime.Pagination;
+            if (paginationOptions.DefaultPageSize < -1 ||
+                paginationOptions.DefaultPageSize == 0 ||
+                paginationOptions.MaxPageSize < -1 ||
+                paginationOptions.MaxPageSize == 0 ||
+                paginationOptions.DefaultPageSize > paginationOptions.MaxPageSize)
```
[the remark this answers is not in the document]
ayush3797: Why not break this up into multiple more-specific conditions/exceptions?

src/Core/Resolvers/BaseQueryStructure.cs
```diff
@@ -181,5 +185,42 @@ internal static IObjectField ExtractItemsSchemaField(IObjectField connectionSche
         {
             return GraphQLUtils.UnderlyingGraphQLEntityType(connectionSchemaField.Type).Fields[QueryBuilder.PAGINATION_FIELD_NAME];
         }
+
+        internal static int GetPaginationLimit(RuntimeConfigProvider runtimeConfigProvider, int? userInputValue)
```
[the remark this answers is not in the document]
ayush3797: It might be bloating the `RuntimeConfig` but helper methods which deal with all the intricacies and return the maxpagesize/defaultpagesize can be placed in the `RuntimeConfig`.

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// The default page size for pagination.
+    /// </summary>
+    [JsonPropertyName("default-page-size")]
+    public uint? DefaultPageSize { get; init; } = null;
+
+    /// <summary>
+    /// The max page size for pagination.
+    /// </summary>
+    [JsonPropertyName("max-page-size")]
+    public uint? MaxPageSize { get; init; } = null;
```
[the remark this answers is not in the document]
rohkhann: Wont whatever the user put pass through the constructor and hence be cast to an uint?

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+            UserProvidedMaxPageSize = true;
+        }
+        else
+        {
+            this.MaxPageSize = MAX_PAGE_SIZE;
+        }
+
+        if (DefaultPageSize is not null)
+        {
+            this.DefaultPageSize = DefaultPageSize == -1 ? (uint)this.MaxPageSize : (uint)DefaultPageSize;
```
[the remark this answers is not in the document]
rohkhann: Yes.

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// <summary>
+    /// Flag which informs CLI and JSON serializer whether to write max-page-size
+    /// property and value to the runtime config file.
+    /// When user doesn't provide the max-page-size property/value, which signals DAB to use the default,
+    /// the DAB CLI should not write the default value to a serialized config.
+    /// This is because the user's intent is to use DAB's default value which could change
+    /// and DAB CLI writing the property and value would lose the user's intent.
+    /// This is because if the user were to use the CLI created config, a max-page-size
+    /// property/value specified would be interpreted by DAB as "user explicitly set ttl."
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+        else
+        {
+            this.DefaultPageSize = DEFAULT_PAGE_SIZE;
+        }
+    }
+
+    /// <summary>
+    /// Flag which informs CLI and JSON serializer whether to write ttl-seconds
+    /// property and value to the runtime config file.
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/BaseQueryStructure.cs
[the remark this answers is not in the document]
rohkhann: This makes sense. moved it.

src/Core/Resolvers/SqlPaginationUtil.cs
rohkhann: fixed.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -198,7 +192,8 @@ public class SqlQueryStructure : BaseSqlQueryStructure
             }
 
             AddColumnsForEndCursor();
-            _limit = context.First is not null ? context.First + 1 : DEFAULT_LIST_LIMIT + 1;
+            _limit = (uint?)GetPaginationLimit(runtimeConfigProvider, (int?)context.First) + 1;
```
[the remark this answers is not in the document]
rohkhann: Modified the code to make it easier to read.

src/Config/ObjectModel/PaginationOptions.cs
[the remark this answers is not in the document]
seantleonard: Yeah, i see that now.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -64,6 +64,92 @@ public async Task RequestFullConnection()
...
+
+            JsonElement actual = await ExecuteGraphQLRequestAsync(graphQLQuery, graphQLQueryName, isAuthenticated: false);
+            string expected = @"{
+              ""items"": [
+                {
+                  ""id"": 1,
+                  ""title"": ""Awesome book""
+                },
+                {
+                  ""id"": 2,
```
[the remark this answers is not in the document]
rohkhann: updated with comment. On the second point, arent all the other tests in this file susceptible to the same issue on if you change underlying db contents they will get affected as well? If we do a number check also with max page size  (something like number of records returned are <= maxpagesize), we would still need to check that the number of records returned == 12 so that we ensure that it did not end up throwing less rows and that would be susceptible if you add more books to the table

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
rohkhann: So I've followed the convention here of other test cases that treat invalid request with substatus code of bad request. This is how it was configured for negative/0 values in other test cases as well.

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
rohkhann: fixed this.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
rohkhann: Fixed the comment. Had copied from earlier test case and missed changing it.

src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
rohkhann: fixed.

src/Core/Configurations/RuntimeConfigValidator.cs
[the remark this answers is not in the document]
rohkhann: moved this to paginationOptions.

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// </summary>
+    [JsonPropertyName("max-page-size")]
+    public uint? MaxPageSize { get; init; } = null;
+
+    [JsonConstructor]
+    public PaginationOptions(int? DefaultPageSize = null, int? MaxPageSize = null)
+    {
+        if (MaxPageSize is not null)
+        {
+            this.MaxPageSize = MaxPageSize == -1 ? UInt32.MaxValue : (uint)MaxPageSize;
```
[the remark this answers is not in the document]
rohkhann: Added appropriate validations.

src/Core/Resolvers/BaseQueryStructure.cs
```diff
@@ -181,5 +185,42 @@ internal static IObjectField ExtractItemsSchemaField(IObjectField connectionSche
...
+                else
+                {
+                    return (int)(userInputValue == -1 ? maxPageSize : userInputValue);
+                }
+            }
+            else
+            {
+                return defaultPageSize;
+            }
+        }
```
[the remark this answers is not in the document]
rohkhann: fixed this.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -59,15 +58,10 @@ public class SqlQueryStructure : BaseSqlQueryStructure
...
-        /// <summary>
-        /// Default limit when no first param is specified for list queries
-        /// </summary>
-        private const uint DEFAULT_LIST_LIMIT = 100;
-
         /// <summary>
         /// The maximum number of results this query should return.
         /// </summary>
-        private uint? _limit = DEFAULT_LIST_LIMIT;
+        private uint? _limit = PaginationOptions.DEFAULT_PAGE_SIZE;
```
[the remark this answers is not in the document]
rohkhann: its not really a breaking change right? as this is backward compatible. if user does not specifiy pagination options this will still be 100 as was before.

src/Core/Services/RequestValidator.cs
rohkhann: Added more tests.

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,113 @@
...
+            ValidatePageSize((int)DefaultPageSize);
+            this.DefaultPageSize = DefaultPageSize == -1 ? (uint)this.MaxPageSize : (uint)DefaultPageSize;
+            UserProvidedDefaultPageSize = true;
+        }
+        else
+        {
+            this.DefaultPageSize = DEFAULT_PAGE_SIZE;
+        }
+
+        if (this.DefaultPageSize > this.MaxPageSize)
```
aaronburtle, line 65: I think this part of the validation could be added to the `ValidatePageSize` method, since a max page size can't be greater than itself, the comparison would only fail in the case of a default page size given that is larger than the max page size. Lets all the validation happen inside the validation method, and makes the name of that method more descriptive (as of now the validation in that method matches validation for max page size but not default)

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// This is because the user's intent is to use DAB's default value which could change
+    /// and DAB CLI writing the property and value would lose the user's intent.
+    /// This is because if the user were to use the CLI created config, a max-page-size
+    /// property/value specified would be interpreted by DAB as "user explicitly max-page-size."
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
+    [MemberNotNullWhen(true, nameof(MaxPageSize))]
+    public bool UserProvidedMaxPageSize { get; init; } = false;
+
+    private static void ValidatePageSize(int pageSize)
```
aaronburtle, line 102: nit: can you add a summary to this helper method?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -278,6 +279,38 @@ public void HotReloadConfig()
...
+        this.TryGetConfig(out RuntimeConfig? runtimeConfig);
+        uint defaultPageSize = runtimeConfig!.DefaultPageSize();
+        uint maxPageSize = runtimeConfig.MaxPageSize();
+
+        if (first is not null)
+        {
+            if (first < -1 || first == 0 || first > maxPageSize)
+            {
+                throw new DataApiBuilderException(
+                message: $"Invalid number of items requested, {QueryBuilder.PAGE_START_ARGUMENT_NAME} argument must be either -1 or a positive number within the max page size limit of {maxPageSize}. Actual value: {first}",
```
aaronburtle, line 299: super nit:
```suggestion
                message: $"Invalid number of items requested, {QueryBuilder.PAGE_START_ARGUMENT_NAME} argument must be either -1 or a positive integer within the max page size limit of {maxPageSize}. Actual value: {first}",
```

src/Core/Services/RequestValidator.cs
```diff
@@ -546,17 +546,17 @@ private static StoredProcedureDefinition TryGetStoredProcedureDefinition(string
         /// </summary>
```
aaronburtle, line 546: This summary should be updated since we no longer are checking that it can parse to a uint > 0.

src/Core/Services/RequestValidator.cs
```diff
@@ -546,17 +546,17 @@ private static StoredProcedureDefinition TryGetStoredProcedureDefinition(string
...
         /// <returns>uint > 0 representing $first</returns>
-        public static uint CheckFirstValidity(string first)
+        public static int CheckFirstValidity(string first)
         {
-            if (!uint.TryParse(first, out uint firstAsUint) || firstAsUint == 0)
+            if (!int.TryParse(first, out int firstAsInt) || firstAsInt == 0)
             {
                 throw new DataApiBuilderException(
-                        message: $"Invalid number of items requested, $first must be an integer greater than 0. Actual value: {first}",
+                        message: $"Invalid number of items requested, $first must be -1 or an integer greater than 0. Actual value: {first}",
```
aaronburtle, line 554: `(!int.TryParse(first, out int firstAsInt) || firstAsInt == 0)` seems to allow for values below -1, do we need to change the responsibility of this helper function?

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1568,12 +1568,16 @@ public async Task GraphQLQueryWithMultipleOfTheSameFieldReturnsFieldOnce()
 
         #region Negative Tests
 
+        /// <summary>
+        /// This test checks the failure on providing invalid nfirst parameter in graphQL Query.
```
aaronburtle, line 1572: nit:
```suggestion
        /// This test checks the failure on providing invalid first parameter in graphQL Query.
```

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -278,6 +279,38 @@ public void HotReloadConfig()
...
+    /// <summary>
+    /// Get the pagination limit from the runtime configuration.
+    /// </summary>
+    /// <param name="first">The pagination input from the user. Example: $first=10</param>
+    /// <returns></returns>
+    /// <exception cref="DataApiBuilderException"></exception>
+    public uint GetPaginationLimit(int? first)
+    {
+        this.TryGetConfig(out RuntimeConfig? runtimeConfig);
+        uint defaultPageSize = runtimeConfig!.DefaultPageSize();
```
aaronburtle, line 291:
```
        uint defaultPageSize = runtimeConfig!.DefaultPageSize();
        uint maxPageSize = runtimeConfig.MaxPageSize();
```

This section seems to be throwing a null reference exception in the tests sometimes when the runtimeconfig is null here, which is happening when the `ODataASTVisitorUnitTests` are running. Those unit tests are SqlTestBase type and so the initialization of the test fixture should not allow this, and when you manually step through one test at a time, at least for me locally, it always passes. But when running the entire test suite, it can happen.

I got this to always pass locally by changing the above to

```
        uint defaultPageSize = runtimeConfig is null ? 100 : runtimeConfig!.DefaultPageSize();
        uint maxPageSize = runtimeConfig is null ? 100000 : runtimeConfig.MaxPageSize();
```

though it would be nice to know why the test is being flakey here.

JerryNixon:
Documentation https://github.com/MicrosoftDocs/data-api-builder-docs-pr/pull/149

src/Config/ObjectModel/PaginationOptions.cs
rohkhann: The idea was to have two seperate exceptions for size values being incorrect for the default page and for the max page size respectively. so the validate page size method, validates a single page size. Here we want to throw a different exception text, rather than just grouping it in one. I feel it is more clear this way.

src/Config/ObjectModel/PaginationOptions.cs
rohkhann: Followed practice in code base of private methods not having summaries.

src/Core/Configurations/RuntimeConfigProvider.cs
rohkhann: Thanks for helping with this. Let me fix this.

src/Core/Services/RequestValidator.cs
rohkhann: Thanks for catching this. Let me fix.

src/Config/ObjectModel/PaginationOptions.cs
aaronburtle: Two exceptions make sense for clarity, but you might want to rename this so that it reflects that it only validates the max size case.

src/Core/Services/RequestValidator.cs
rohkhann: fixed.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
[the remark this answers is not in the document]
seantleonard: Gotcha, Resolving because of:
> if user does not specifiy pagination options this will still be 100 as was before.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -886,6 +974,25 @@ public async Task RequestInvalidZeroFirst()
             SqlTestHelper.TestForErrorInGraphQLResponse(result.ToString(), statusCode: $"{DataApiBuilderException.SubStatusCodes.BadRequest}");
         }
 
+        /// <summary>
+        /// Request an invalid number of entries for a pagination page
+        /// </summary>
+        [TestMethod]
+        public async Task RequestInvalidMaxSize()
```
seantleonard, line 981: do we also want a test that validates `100000` explicitly renders a request successful? that would be "regression detection" test because if someone inadvertently changes maxpagesize from a value of 100000, we'd catch that before allowing a  breaking change to be committed.

seantleonard: Your tests `RequestMaxUsingNegativeOne()` only ensures that `-1` translates into whatever value happens to be set as maxpagesize.

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
```diff
@@ -1568,12 +1568,16 @@ public async Task GraphQLQueryWithMultipleOfTheSameFieldReturnsFieldOnce()
 
         #region Negative Tests
 
+        /// <summary>
+        /// This test checks the failure on providing invalid first parameter in graphQL Query.
+        /// We only allow -1 or positive integers for first parameter.-1 means max page size.
```
seantleonard, line 1573: is there a test that validates behavior when user provides `first:0`? 

I added some other comments calling out first:0 translates to default page size prior to this change. If we now error out, that 's a breaking change. would be worth having a test to either assert failure or ensure success (to be "regression detection").

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2362,6 +2362,67 @@ public void TestRuntimeConfigSetupWithNonJsonConstructor()
...
+        /// <param name="expectedMaxPageSize">expected max page size from config.</param>
+        [DataTestMethod]
+        [DataRow(false, null, null, "", (int)PaginationOptions.DEFAULT_PAGE_SIZE, (int)PaginationOptions.MAX_PAGE_SIZE,
+            DisplayName = "MaxPageSize should be 100,000 and DefaultPageSize should be 100 when no value provided in config.")]
+        [DataRow(false, 1000, 10000, "", 1000, 10000,
+            DisplayName = "Valid inputs of MaxPageSize and DefaultPageSize must be accepted and set in the config.")]
+        [DataRow(false, -1, 10000, "", 10000, 10000,
+            DisplayName = "DefaultPageSize should be the same as MaxPageSize when input is -1")]
+        [DataRow(false, 100, -1, "", 100, Int32.MaxValue,
+            DisplayName = "MaxPageSize should be UInt32.MaxValue when input is -1")]
```
seantleonard, line 2382: need to be specific about what field you are referring to when you say "input". The two displaynames for these tests aren't consistent because "input" may be referring to `defaultPageSize` or `maxPageSize`

seantleonard (changes requested):
Last nits and should be good to merge once addressed. Still one last remaining question for how we want to handle user provided First:0 since before this change, that resolved to default page size.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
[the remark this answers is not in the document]
rohkhann: Yes but adding is one type of change that we are talking about to a table right? Similarly if I were to remove books which could be another type of change, that would break this case then.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -198,7 +192,9 @@ public class SqlQueryStructure : BaseSqlQueryStructure
             }
 
             AddColumnsForEndCursor();
-            _limit = context.First is not null ? context.First + 1 : DEFAULT_LIST_LIMIT + 1;
+            runtimeConfigProvider.TryGetConfig(out RuntimeConfig? runtimeConfig);
+            _limit = runtimeConfig?.GetPaginationLimit((int?)context.First) + 1;
```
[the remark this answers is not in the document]
rohkhann: Can runtimeconfig even be null?

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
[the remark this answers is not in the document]
seantleonard: we've typically only ever added records to tables that are used in our tests. That should be expected to happen in the future. We don't remove records from tests because of the implications you mentioned, breaking tests which expect certain records to be present.

src/Core/Services/RequestValidator.cs
```diff
@@ -542,21 +542,21 @@ private static StoredProcedureDefinition TryGetStoredProcedureDefinition(string
...
+        /// to be sure that it can parse to a int > 0 or -1.
         /// </summary>
         /// <param name="first">String representing value associated with $first</param>
-        /// <returns>uint > 0 representing $first</returns>
-        public static uint CheckFirstValidity(string first)
+        /// <returns>int > 0 or int == -1 representing $first</returns>
+        public static int CheckFirstValidity(string first)
         {
-            if (!uint.TryParse(first, out uint firstAsUint) || firstAsUint == 0)
+            if (!int.TryParse(first, out int firstAsInt) || firstAsInt == 0 || firstAsInt < -1)
```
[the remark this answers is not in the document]
rohkhann: According to previous code, if first was 0, this would have thrown.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
rohkhann: Yes let me add this test.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -577,12 +577,33 @@ public static JsonElement CreateNextLink(string path, NameValueCollection? query
...
         /// <param name="jsonResult">Results plus one extra record if more exist.</param>
         /// <param name="first">Client provided limit if one exists, otherwise 0.</param>
+        /// <param name="defaultPageSize">Default limit for page size.</param>
+        /// <param name="maxPageSize">Maximum limit for page size.</param>
         /// <returns>Bool representing if more records are available.</returns>
-        public static bool HasNext(JsonElement jsonResult, uint? first)
+        public static bool HasNext(JsonElement jsonResult, int? first, uint defaultPageSize, uint maxPageSize)
         {
-            // When first is 0 we use default limit of 100, otherwise we use first
+            // When first is null we use default limit from runtime config, otherwise we use first
```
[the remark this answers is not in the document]
rohkhann: Please see this test case validating the 0 scenario and also below test comment: https://github.com/Azure/data-api-builder/blob/a23c590ceb45dbac416f9d038aa9d8a6e7d85317/src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs#L1548 in main validating 0 being invalid.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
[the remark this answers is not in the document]
rohkhann: Makes sense will make this fix.

src/Service.Tests/SqlTests/GraphQLQueryTests/GraphQLQueryTestBase.cs
rohkhann: Yes see this from main: https://github.com/Azure/data-api-builder/blob/a23c590ceb45dbac416f9d038aa9d8a6e7d85317/src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs#L1548

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
rohkhann: Oh i meant to refer to both here. this is the valid scenario where both are correct. Hence valid inputs of both should be accepted in config.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -886,6 +974,25 @@ public async Task RequestInvalidZeroFirst()
             SqlTestHelper.TestForErrorInGraphQLResponse(result.ToString(), statusCode: $"{DataApiBuilderException.SubStatusCodes.BadRequest}");
         }
 
+        /// <summary>
+        /// Request an invalid number of entries for a pagination page
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Core/Resolvers/CosmosQueryStructure.cs
```diff
@@ -142,7 +147,9 @@ private void Init(IDictionary<string, object?> queryParams)
             // TODO: Revisit 'first' while adding support for TOP queries
             if (queryParams.ContainsKey(QueryBuilder.PAGE_START_ARGUMENT_NAME))
             {
-                MaxItemCount = (int?)queryParams[QueryBuilder.PAGE_START_ARGUMENT_NAME];
+                RuntimeConfigProvider.TryGetConfig(out RuntimeConfig? runtimeConfig);
+                MaxItemCount = runtimeConfig?.GetPaginationLimit((int)queryParams[QueryBuilder.PAGE_START_ARGUMENT_NAME]!);
```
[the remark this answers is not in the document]
rohkhann: Added null check similar to order by code below.

seantleonard (approved):
Thanks for working through feedback! Pending test of first: 0 for GraphQL in main, and then closing out remainder of discussions since we discussed over a call.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
rohkhann: Added a comment as discussed on call in configvalidator with note that changing these values could be a breaking change.

src/Core/Services/RequestValidator.cs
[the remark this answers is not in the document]
rohkhann: Validated with GQL that 0 fails with code on main:

src/Core/Resolvers/SqlPaginationUtil.cs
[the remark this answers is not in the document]
rohkhann: Validated for 0 case for gql as well from main branch code: 
image

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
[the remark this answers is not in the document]
rohkhann: Changed to ! to indicate not null as discussed on call that runtime config cant be null when it reaches here.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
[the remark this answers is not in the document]
rohkhann: Discussed on call. resolving due to large refactoring required to do this specific implementation.

aaronburtle (approved):
LGTM! Thanks for quickly addressing comments.

src/Config/ObjectModel/PaginationOptions.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// <summary>
+    /// The max page size for pagination.
+    /// </summary>
+    [JsonPropertyName("max-page-size")]
+    public uint? MaxPageSize { get; init; } = null;
+
+    [JsonConstructor]
+    public PaginationOptions(int? DefaultPageSize = null, int? MaxPageSize = null)
+    {
+        if (MaxPageSize is not null)
```
rohkhann, line 43: Need to validate type mismatch

seantleonard (approved):

PR #2161, opened 2024-04-12
Cosmos DB: Adds Patch Support

sourabh1007:
## Why make this change?
Today, in cosmos DB, we have ability to `update` the item which is actually `replace` the item. Adding new operation i.e. `patch` where customer would have ability to "patch" an item.
`PATCH` would be available only for Cosmos DB.

## What is this change?
Before going further, it is highly recommended to go through below docs:
1. How patch works in Cosmos DB https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update#supported-operations
2. Limitations: https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update#supported-modes
3. How it works with SDK: https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update-getting-started?tabs=dotnet#prerequisites

To Summarize, here is a simple example of patch:
```
List<PatchOperation> operations = new ()
{
    PatchOperation.Add("/color", "silver"),
    PatchOperation.Remove("/used"),
    PatchOperation.Increment("/price", 50.00),
    PatchOperation.Add("/tags/-", "featured-bikes")
};

ItemResponse<Product> response = await container.PatchItemAsync<Product>(
    id: "e379aea5-63f5-4623-9a9b-4cd9b33b91d5",
    partitionKey: new PartitionKey("road-bikes"),
    patchOperations: operations
);
```
You need to generate `PatchOpertation` which need below information:
a) Operation type i.e. Set, Add, Remove etc
b) Attribute Path where operation needs to be applied i.e "/color", "/used" in above example
c) New value

**What DAB supports?**
1. We decided to support only `Set` operation, as it solves the purpose to update an item. It means, you cannot perform any specific operations like remove, move etc.
2. There is no special handling for an array.
3. If the target path specifies an element that doesn't exist, it's added.
4. If the target path specifies an element that already exists, its value is replaced.

**Changes as part of this PR:**
1. Generate `patch` operation for given entities
image
5. Generate `PatchPlanetInput` without _id_ field as using patch operation you can not update an `Id`
image
6. Implement patch operation 
a) It translates the passed item into "patchoperation" by traversing the item.
b) Checks if number of patch operations are less than 10 or more than 10 (as cosmsodb supports at max 10 patch operations at a time)
c) If it is less than or equal to 10, it fires patch call with patch operations
d) If it is greater than 10, then it creates a transaction batch of patch call, with 10 patch operations in each patch call. (_RU exhaustive but functionally it works_)

**Pictorial Overview of the implementation**
```mermaid
flowchart TD
    User[fa:fa-user User]-->| patch operation |DAB
    subgraph DAB[DAB]
        Authorization[Authorization]-->Patch
        subgraph Patch[Patch Operation]
        PatchOperation[Generate Patch 'Set' Operations with passed item] -->CheckCount{Number of Patch Operation > 10}
        CheckCount --> |No| SDKOperation[SDK Patch call]
        CheckCount --> |Yes| TransactionBatch[Create a batch of patch calls with max 10 patch operations ]-->SDKBatchOperation[SDK ExecuteBatch call]
        end
    end
```

**Authorization for this new Mutation Operation:**
`update` access would be applied for `patch` and `update` operations both.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)

Refer test cases.


src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -69,6 +69,9 @@
             },
             {
               "action": "delete"
+            },
```
Aniruddh25, line 72: do you need to introduce another action? Isnt a combination of create + update sufficient to determine the permissions for `patch`?

src/Config/ObjectModel/EntityAction.cs
```diff
@@ -5,6 +5,18 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 
 public record EntityAction(EntityActionOperation Action, EntityActionFields? Fields, EntityActionPolicy? Policy)
 {
-    public static readonly HashSet<EntityActionOperation> ValidPermissionOperations = new() { EntityActionOperation.Create, EntityActionOperation.Read, EntityActionOperation.Update, EntityActionOperation.Delete };
+    public static readonly HashSet<EntityActionOperation> ValidPermissionOperationsForCosmos = new() {
```
Aniruddh25, line 8: it doesn't make sense to introduce a special action for CosmosDB if the same behavior can be achieved through other actions.

Aniruddh25 (changes requested):
Waiting on clarification on the need to introduce a `patch` permission..

src/Config/ObjectModel/EntityAction.cs
sourabh1007: Clarified on scrum, we agreed to, not to have separate `patch` permission. if `update` operation is there, then it should allow `patch` operation.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -729,6 +730,252 @@ public async Task ValidateInheritanceOfReadPermissionFromAnonymous()
...
+
+            const string newName = "new_name";
+            mutation = $@"
+mutation {{
+    patchPlanet (id: ""{id}"", _partitionKeyValue: ""{id}"", item: {{name: ""{newName}"", stars: [{{ id: ""{id}"" }}] }}) {{
+        id
+        name
+    }}
+}}";
+            JsonElement response = await ExecuteGraphQLRequestAsync("patchPlanet", mutation, variables: new());
```
seantleonard, line 756: I see the request is different. Are you testing whether nested patch works? could that be it's own test case with appropriate comment?

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -729,6 +730,252 @@ public async Task ValidateInheritanceOfReadPermissionFromAnonymous()
...
+            };
+            _ = await ExecuteGraphQLRequestAsync("createPlanet", _createPlanetMutation, new() { { "item", input } });
+
+            const string newName = "new_name";
+            string mutation = @"
+mutation ($id: ID!, $partitionKeyValue: String!, $item: PatchPlanetInput!) {
+    patchPlanet (id: $id, _partitionKeyValue: $partitionKeyValue, item: $item) {
+        id
+        name
+     }
```
seantleonard, line 781: Shouldn't validation select all fields for the entity record? that way you can validate that ONLY the intended fields are "SET" (Patched) and no other fields are modified?

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -729,6 +730,252 @@ public async Task ValidateInheritanceOfReadPermissionFromAnonymous()
...
+            // Validate results
+            Assert.AreEqual(newName, response.GetProperty("name").GetString());
+            Assert.AreNotEqual(input.name, response.GetProperty("name").GetString());
+        }
+
+        /// <summary>
+        ///  Patch Operation has limitation of executing/patching only 10 attributes at a time, internally which is 10 patch operation.
+        ///  In DAB, we are supporting multiple patch operations in a single patch operation by executing them in a Transactional Batch.
+        /// </summary>
+        /// <returns></returns>
```
seantleonard, line 800:
```suggestion
```

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -419,17 +419,31 @@ public void SetEntityPermissionMap(RuntimeConfig runtimeConfig)
...
-    public static IEnumerable<EntityActionOperation> GetAllOperationsForObjectType(EntityActionOperation operation, EntitySourceType? sourceType)
+    public static IEnumerable<EntityActionOperation> GetAllOperationsForObjectType(DatabaseType databaseType, EntityActionOperation operation, EntitySourceType? sourceType)
     {
         if (sourceType is EntitySourceType.StoredProcedure)
         {
             return new List<EntityActionOperation> { EntityActionOperation.Execute };
         }
 
-        return operation is EntityActionOperation.All ? EntityAction.ValidPermissionOperations : new List<EntityActionOperation> { operation };
+        IEnumerable<EntityActionOperation> validOperation =
```
seantleonard, line 433:
```suggestion
        IEnumerable<EntityActionOperation> validOperations =
```

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -345,7 +346,7 @@ public void ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(RuntimeEntit
             }
             else
             {
-                // For entities (table/view) that have graphQL exposed, two queries and three mutations would be generated.
+                // For entities (table/view) that have graphQL exposed, two queries and four mutations would be generated.
```
seantleonard, line 349: only four mutations for CosmosDB though. So you need to make that distinction in this comment.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1284,6 +1287,7 @@ private void HandleOrRecordException(Exception ex)
     /// Valid stored procedure  actions:
     /// - Execute
     /// </summary>
+    /// <param name="databaseType">Database Type</param>
```
seantleonard, line 1290: outdated param. remove.

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -65,13 +65,23 @@ private async Task<JObject> ExecuteAsync(IMiddlewareContext context, IDictionary
...
-            ItemResponse<JObject>? response = resolver.OperationType switch
+            JObject result;
+            if (resolver.OperationType == EntityActionOperation.Patch)
             {
-                EntityActionOperation.UpdateGraphQL => await HandleUpdateAsync(queryArgs, container),
-                EntityActionOperation.Create => await HandleCreateAsync(queryArgs, container),
-                EntityActionOperation.Delete => await HandleDeleteAsync(queryArgs, container),
-                _ => throw new NotSupportedException($"unsupported operation type: {resolver.OperationType}")
-            };
+                result = await HandlePatchAsync(queryArgs, container);
```
seantleonard, line 71: why is patch response different? is there no "result.Resource" for patch response? is it because it's batch patch operation per https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update-getting-started?tabs=dotnet#prerequisites:~:text=Combine%20multiple%20patch%20operations%3A

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -234,6 +255,118 @@ private static async Task<ItemResponse<JObject>> HandleUpdateAsync(IDictionary<s
...
 
+        /// <summary>
+        /// Refer https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update for more details on patch operations
+        /// </summary>
+        /// <param name="queryArgs"></param>
+        /// <param name="container"></param>
+        /// <returns></returns>
+        /// <exception cref="InvalidDataException"></exception>
+        /// <exception cref="DataApiBuilderException"></exception>
+        private static async Task<JObject> HandlePatchAsync(IDictionary<string, object?> queryArgs, Container container)
```
seantleonard, line 266:
```suggestion
        /// <summary>
        /// Refer to https://learn.microsoft.com/azure/cosmos-db/partial-document-update for more details on patch operations
        /// </summary>
        /// <param name="queryArgs"></param>
        /// <param name="container"></param>
        /// <returns></returns>
        /// <exception cref="InvalidDataException"></exception>
        /// <exception cref="DataApiBuilderException"></exception>
        private static async Task<JObject> HandlePatchAsync(IDictionary<string, object?> queryArgs, Container container)
```
incomplete param/exception comments

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -234,6 +255,118 @@ private static async Task<ItemResponse<JObject>> HandleUpdateAsync(IDictionary<s
...
+            if (input is null)
+            {
+                throw new InvalidDataException("Input Item field is invalid");
+            }
+
+            // This would contain the patch operations to be applied on the document
+            List<PatchOperation> patchOperations = new();
+            GeneratePatchOperations(input, "", patchOperations);
+
+            int patchOperationsLimit = 10;
```
seantleonard, line 314: can make const for entire CosmosMutationEngine?

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -234,6 +255,118 @@ private static async Task<ItemResponse<JObject>> HandleUpdateAsync(IDictionary<s
...
+            TransactionalBatchResponse response = await batch.ExecuteAsync();
+            if (!response.IsSuccessStatusCode)
+            {
+                throw new DataApiBuilderException(
+                        message: "Failed to patch the item",
+                        statusCode: HttpStatusCode.InternalServerError,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.DatabaseOperationFailed);
+            }
+
+            return response.GetOperationResultAtIndex<JObject>(numberOfBatches).Resource;
```
seantleonard, line 341: if you don't add `.Resource` here, the caller need not have conditional if/else block on operation?

```csharp
JObject result;
ItemResponse<JObject>? response = resolver.OperationType switch
                {
                    EntityActionOperation.Patch => await HandlePatchAsync(queryArgs, container),
                    EntityActionOperation.UpdateGraphQL => await HandleUpdateAsync(queryArgs, container),
                    EntityActionOperation.Create => await HandleCreateAsync(queryArgs, container),
                    EntityActionOperation.Delete => await HandleDeleteAsync(queryArgs, container),
                    _ => throw new NotSupportedException($"unsupported operation type: {resolver.OperationType}")
                };

result = response.Resource;
```

seantleonard (commented):
some suggestions/questions.

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -729,6 +730,252 @@ public async Task ValidateInheritanceOfReadPermissionFromAnonymous()
...
+    patchPlanet (id: ""{id}"", _partitionKeyValue: ""{id}"", item: {{name: ""{newName}"", stars: [{{ id: ""{id}"" }}] }}) {{
+        id
+        name
+    }}
+}}";
+            JsonElement response = await ExecuteGraphQLRequestAsync("patchPlanet", mutation, variables: new());
+
+            // Validate results
+            Assert.AreEqual(newName, response.GetProperty("name").GetString());
+            Assert.AreNotEqual(name, response.GetProperty("name").GetString());
```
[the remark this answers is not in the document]
sourabh1007: what do you mean by "test execution would never get to this assert'?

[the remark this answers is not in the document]
seantleonard: If code execution ever gets to the assert on line 760, the assert will never fail. Its purpose is redundant. 
```csharp
759:: Assert.AreEqual(newName, response.GetProperty("name").GetString());
760:: Assert.AreNotEqual(name, response.GetProperty("name").GetString());
```
If L759 assert passes, then that implies the patch was successful and it's impossible that L760 would fail. If `name` ==, `response.GetProperty("name").GetString()`, then L759 would always fail.

src/Service.Tests/CosmosTests/MutationTests.cs
sourabh1007: Added separate test to cover nested patch and added comments and more assertions. I hope it makes more sense now.

src/Service.Tests/CosmosTests/MutationTests.cs
sourabh1007: selected all the fields

src/Core/Resolvers/CosmosMutationEngine.cs
sourabh1007: Because `TransactionalBatchResponse` does not return `Itemresponse` like other mutation operations.

src/Core/Resolvers/CosmosMutationEngine.cs
sourabh1007: No,` response.GetOperationResultAtIndex<JObject>(numberOfBatches)` is of type `TransactionalBatchOperationResult` which is different than `ItemResponse`.

src/Service.Tests/CosmosTests/MutationTests.cs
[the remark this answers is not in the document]
sourabh1007: got it, I have removed the other assertion and added more assertions.

src/Config/ObjectModel/EntityActionOperation.cs
```diff
@@ -19,7 +19,7 @@ public enum EntityActionOperation
     Delete, Read,
 
     // cosmosdb_nosql operations
-    Upsert, Create,
+    Upsert, Create, Patch,
```
Aniruddh25, line 22: If we dont need a separate permission for Patch, do we still need a separate operation in EntityActionOperation?

src/Cli/Utils.cs
```diff
@@ -96,8 +96,13 @@ public static EntityAction[] CreateOperations(string operations, EntityActionPol
...
-                        EntityAction.ValidStoredProcedurePermissionOperations :
-                        EntityAction.ValidPermissionOperations;
+                        EntityAction.ValidStoredProcedurePermissionOperations : EntityAction.ValidPermissionOperations;
+
+                    if (databaseType is DatabaseType.CosmosDB_NoSQL && resolvedOperations.Contains(EntityActionOperation.Update))
+                    {
+                        resolvedOperations.Add(EntityActionOperation.Patch);
+                    }
+
                     // Expand wildcard to all valid operations (except execute)
```
Aniruddh25, line 106: If there are no separate permissions for CosmosDB to accommodate the Patch operation, why do we still have a separate function `ValidatePermissionOperationsForCosmosDB`? Subsequently, if its not needed, there's no need to input database type as an additional argument to the various functions.

src/Config/ObjectModel/RuntimeEntities.cs
```diff
@@ -67,7 +68,9 @@ public bool ContainsKey(string key)
             }
             else
             {
-                throw new ApplicationException($"The entity '{key}' was not found in the dab-config json");
```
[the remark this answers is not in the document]
sourabh1007: Till now, it was "keyNotFoundException" which got thrown when it tries to get an element from the dictionary. this code will just change the exception type and message. 
It came as feedback from an old PR. Since it is a small change, making it as part of this PR.

src/Config/ObjectModel/EntityActionOperation.cs
sourabh1007: Yes, `Patch` is a separate operation and it should be reflected in the DAB code.

src/Cli/Utils.cs
sourabh1007: Sorry, I couldn't find the reference of  `ValidatePermissionOperationsForCosmosDB`.. am I missing something?

seantleonard: In AuthorizationResolver, you're 1:1 mapping `Operation.Update` to `Operation.Patch`.
```csharp
// <AuthorizationResolver.cs>
        IEnumerable<EntityActionOperation> validOperation =
            operation is EntityActionOperation.All ? EntityAction.ValidPermissionOperations : new List<EntityActionOperation> { operation };

        //For CosmosDB, add Patch operation to the list of valid operations if Update operation is present.
        if (databaseType is DatabaseType.CosmosDB_NoSQL &&
                    validOperation.Contains(EntityActionOperation.Update))
        {
            List<EntityActionOperation> tempOperationList = validOperation.ToList<EntityActionOperation>();
            tempOperationList.Add(EntityActionOperation.Patch);

            validOperation = tempOperationList;
        };
```

 Back to Ani's comment, it doesn't look like AuthorizationResolver needs to pass DatabaseType==CosmosDB (or pass dbtype at all) and all you need to do is check whether the `operation.Update` permission is allowed.

```csharp
//<CosmosMutationEngine.cs>
 EntityActionOperation.Patch =>
                    _authorizationResolver.AreColumnsAllowedForOperation(entityName,
                        roleName: clientRole,
                        operation: EntityActionOperation.Patch, // --> This could just be EntityActionOperation.Update ?
                        columns: inputArgumentKeys),
```

src/Config/ObjectModel/EntityActionOperation.cs
seantleonard: I think this represents high level of operation type needed for permissions versus MutationEngine code path execution:

| GraphQL Mutation | Permission    | DAB OperationType / CosmosMutationEngine code path |
|------------------|---------------|----------------------------------------------------|
| patchEntityByPk  | Update        | Patch                                              |
| updateEntityByPk | Create,Update | Update                                             |

src/Cli/Utils.cs
sourabh1007: Thanks for the review @Aniruddh25 and Thanks for the explanation @seantleonard  I have made this changes.

src/Core/Services/TypeHelper.cs
```diff
@@ -293,5 +293,36 @@ public static bool TryGetDbTypeFromSqlDbDateTimeType(SqlDbType sqlDbType, [NotNu
...
+                SyntaxKind.FloatValue => Convert.ToDouble(node.Value), // spec
+                SyntaxKind.BooleanValue => Convert.ToBoolean(node.Value), // spec
+                SyntaxKind.StringValue => Convert.ToString(node.Value), // spec
+                SyntaxKind.NullValue => null, // spec
+                _ => Convert.ToString(node.Value)
+            };
+        }
+
+        /// <summary>
+        /// This function identifies is the value type is primitive or not.
```
Aniruddh25, line 317:
```suggestion
        /// This function identifies if the value type is primitive or not.
```

Aniruddh25:
previous comments have been addressed, but I need more time to look at few other files.. so reset my review. If this gets approved before that, feel free to merge it in.

seantleonard (approved):

neeraj-sharma2592 (approved):
LGTM

src/Core/Resolvers/CosmosMutationEngine.cs
```diff
@@ -65,13 +65,23 @@ private async Task<JObject> ExecuteAsync(IMiddlewareContext context, IDictionary
             string entityName = metadataProvider.GetEntityName(graphQLType);
             AuthorizeMutation(context, queryArgs, entityName, resolver.OperationType);
 
-            ItemResponse<JObject>? response = resolver.OperationType switch
+            JObject result;
+            if (resolver.OperationType == EntityActionOperation.Patch)
```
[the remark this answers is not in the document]
sourabh1007: discussed offline, closing on this.

Issue #1753, opened 2023-09-27
[Improvement] Multi-Source Support for REST scenario.


rohkhann:
with RFC: https://github.com/Azure/data-api-builder/issues/1638, we have added support for multiple datasources for graphql scenario. Pr:https://github.com/Azure/data-api-builder/pull/1709. This issue tracks adding support for multiple datasources for REST scenario.

seantleonard:
Make sure to consider @kliszaq 's points raised in #1761 with regards to how `RestService` maps requests to additional datasources.

PR #2169, opened 2024-04-16
Multiple db support for REST scenario

rohkhann:
## Why make this change?
This change introduces REST support for multiple DB Scenario.
Issue: https://github.com/Azure/data-api-builder/issues/1753
## What is this change?
1. Instead of using the default db name to dispatch queries, we are using the db name based on entity and then dispatching queries and mutations.

## How was this tested?
1. Existing unit tests should cover backward compatibility scenarios of all single db rest support.
2. Existing unit tests cover the flow once the query is dispatched to the queryEngine/query executor. We need to add the tests for the integration scenario for other calls in rest.
3. Added test for rest scenario dispatch routing.

Integration test:
1. Two db's one has publishers entity and one has books:
Get publishers
image

Get books
image


src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
```diff
@@ -215,5 +280,56 @@ public async Task TestMultiSourceTokenSet()
...
+            Dictionary<string, string> entityNameToDataSourceName = new()
+            {
+                { ENTITYNAME1, DATASOURCENAME1 },
+                { ENTITYNAME2, DATASOURCENAME2 }
+            };
+
+            RuntimeConfig mockConfig1 = new(
+               Schema: "",
+               DataSource: new(DatabaseType.MySQL, "Server =<>;Database=<>;User=xyz;Password=xxx", new()),
+               Runtime: new(
```
seantleonard, line 308: maybe i'm misremembering how we configure multi db scenario: don't we need two runtime config files? one for each source?

seantleonard (commented):
few questions, otherwise looks good. Have you tested this in end2end scenario outside of unit test? I understand that testing that isn't doable/straighforward in pipelines for cross-db types (e.g. sql -> cosmos) because we only have one db type per pipeline. But for multiple sql sources, that may be possible? e.g. different db's on same server and each db is represented as two distinct datasources.

src/Core/Services/RestService.cs
```diff
@@ -212,17 +212,18 @@ RequestValidator requestValidator
         private async Task<IActionResult> DispatchQuery(RestRequestContext context, DatabaseType databaseType)
         {
             IQueryEngine queryEngine = _queryEngineFactory.GetQueryEngine(databaseType);
-            string defaultDataSourceName = _runtimeConfigProvider.GetConfig().DefaultDataSourceName;
+
+            string dataSourceName = _runtimeConfigProvider.GetConfig().GetDataSourceNameFromEntityName(context.EntityName);
```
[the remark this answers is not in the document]
rohkhann: Yes.

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
rohkhann: We do but it all merges into one runtime config eventually. We merge the entities from both datasources into a single dictionary and take the runtime settings from the base config. Hence in test case, I directly just set the resultant runtimeconfig.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -154,6 +156,16 @@ public IEnumerable<DataSource> ListAllDataSources()
         return _dataSourceNameToDataSource.AsEnumerable();
     }
 
+    public bool TryAddEntityPathNameToEntityName(string entityPathName, string entityName)
```
aaronburtle, line 159: I noticed these and a couple other methods come before the constructor in this class. Normally the order is contructors before methods, was there a reason in particular that these (and the others) come first?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -154,6 +156,16 @@ public IEnumerable<DataSource> ListAllDataSources()
...
+    {
+        return _entityPathNameToEntityName.TryAdd(entityPathName, entityName);
+    }
+
+    public bool TryGetEntityNameFromPath(string entityPathName, out string? entityName)
```
aaronburtle, line 164: we now have two methods of the same name, one in runtimeconfig and one in sqlmetdataprovider. Should we really maintain two methods with the same name in different classes that seem to have the same responsibility? Can we refactor this to just be the responsibility of a single method in a single class?

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
```diff
@@ -41,6 +48,13 @@ public class MultiSourceQueryExecutionUnitTests
                 }
             }";
 
+        private const string DATASOURCENAME1 = "db1";
```
aaronburtle, line 51: nit: underscore should separate words for constants, ie: DATA_SOURCE_NAME_1

aaronburtle (commented):
Looks good so far, still need to finish going through the tests.

ayush3797 (approved):
LGTM after addressing Aaron's comments!

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
```diff
@@ -146,15 +121,108 @@ public async Task TestMultiSourceQuery()
...
 
             // validate that the data returend for the queries we did matches the moq data we set up for the respective query engines.
-            Assert.AreEqual("db1", queryMap1[0].Value, $"Data returned for {queryName1} is incorrect for multi-source query");
-            Assert.AreEqual("db2", queryMap2[0].Value, $"Data returned for {queryName2} is incorrect for multi-source query");
+            Assert.AreEqual("db1", queryMap1[0].Value, $"Data returned for {QUERYNAME1} is incorrect for multi-source query");
+            Assert.AreEqual("db2", queryMap2[0].Value, $"Data returned for {QUERYNAME2} is incorrect for multi-source query");
+        }
+
+        [TestMethod]
+        public async Task TestMultiSourceQueryRest()
```
seantleonard, line 136: nit: clarify in a summary what you're validating. e.g.
```csharp
/// Validates that the REST endpoint returns data from the proper datasource configured for an entity:
/// Request: /api/entity1 | Entity: Entity1 | DataSource: CosmosDB
/// Request: /api/entity2 | Entity: Entity2 | DataSource: MsSql
```

src/Core/Services/RestService.cs
```diff
@@ -451,15 +451,15 @@ public bool TryGetRestRouteFromConfig([NotNullWhen(true)] out string? configured
             string entityPath = entityPathAndPKRoute[0];
             string primaryKeyRoute = entityPathAndPKRoute.Length == maxNumberOfElementsFromSplit ? entityPathAndPKRoute[1] : string.Empty;
 
-            if (!sqlMetadataProvider.TryGetEntityNameFromPath(entityPath, out string? entityName))
+            if (!runtimeConfig.TryGetEntityNameFromPath(entityPath, out string? entityName))
```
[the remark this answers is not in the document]
rohkhann: Removed data structure from the sqlMetadataProvider.

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
rohkhann: fixed.

src/Service.Tests/Unittests/RestServiceUnitTests.cs
```diff
@@ -145,16 +145,10 @@ public static void InitializeTest(string restRoutePrefix, string entityName)
...
-            string outParam;
-            sqlMetadataProvider.Setup(x => x.TryGetEntityNameFromPath(It.IsAny<string>(), out outParam)).Returns(true);
-            Dictionary<string, string> _pathToEntityMock = new() { { entityName, entityName } };
-            sqlMetadataProvider.Setup(x => x.TryGetEntityNameFromPath(It.IsAny<string>(), out outParam))
-                               .Callback(new metaDataCallback((string entityPath, out string entity) => _ = _pathToEntityMock.TryGetValue(entityPath, out entity)))
-                               .Returns((string entityPath, out string entity) => _pathToEntityMock.TryGetValue(entityPath, out entity));
-
-            metadataProviderFactory.Setup(x => x.GetMetadataProvider(It.IsAny<string>())).Returns(sqlMetadataProvider.Object);
+            RuntimeConfig loadedConfig = provider.GetConfig();
+            loadedConfig.TryAddEntityPathNameToEntityName(entityName, entityName);
```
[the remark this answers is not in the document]
rohkhann: This is because the metadataprovider was being used for the trygetentitynamefrom path function. since that logic is now in runtime config we no longer need it here.

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
rohkhann: Added.

src/Config/ObjectModel/RuntimeConfig.cs
rohkhann: fixed this!

src/Config/ObjectModel/RuntimeConfig.cs
rohkhann: I think this was a styling thing missed in earlier pr's. I dont want to bloat this pr, will fix this styling issue in a follow up pr that addresses this style concern.

aaronburtle (approved):
LGTM, thanks for addressing things quickly!

JerryNixon:
Some smart people on this PR thread

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
```diff
@@ -146,15 +121,110 @@ public async Task TestMultiSourceQuery()
...
             // validate that the data returend for the queries we did matches the moq data we set up for the respective query engines.
-            Assert.AreEqual("db1", queryMap1[0].Value, $"Data returned for {queryName1} is incorrect for multi-source query");
-            Assert.AreEqual("db2", queryMap2[0].Value, $"Data returned for {queryName2} is incorrect for multi-source query");
+            Assert.AreEqual("db1", queryMap1[0].Value, $"Data returned for {QUERY_NAME_1} is incorrect for multi-source query");
+            Assert.AreEqual("db2", queryMap2[0].Value, $"Data returned for {QUERY_NAME_2} is incorrect for multi-source query");
+        }
+
+        /// <summary>
+        /// Validates successful execution of a query against multiple sources for graphql scenario.
+        /// 1. Tries to use a built schema to execute a query against multiple sources.
```
seantleonard, line 137: built schema? this sounds like graphql again

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
seantleonard: nit: below describes test in more explicit terms.
```csharp
/// Validates that the REST endpoint returns data from the proper datasource configured for an entity:
/// Request: /api/entity1 -> Resolves data from: Entity: Entity1 | DataSource: CosmosDB | QueryEngine: CosmosQueryEngine
/// Request: /api/entity2 -> Resolves data from: Entity: Entity2 | DataSource: MsSql    | QueryEngine: MsSqlQueryEngine
```

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
```diff
@@ -146,15 +121,110 @@ public async Task TestMultiSourceQuery()
...
+            Assert.AreEqual(0, cosmosQueryEngine.Invocations.Count, "Cosmos query engine should not be invoked for multi-source query as entity belongs to sql db.");
+            sqlQueryEngine.Verify(x => x.ExecuteAsync(It.Is<FindRequestContext>(ctx => ctx.EntityName == ENTITY_NAME_1)), Times.Once);
+            cosmosQueryEngine.Verify(x => x.ExecuteAsync(It.IsAny<FindRequestContext>()), Times.Never);
+            IActionResult result = await restService.ExecuteAsync(ENTITY_NAME_2, EntityActionOperation.Read, null);
+            Assert.AreEqual(1, cosmosQueryEngine.Invocations.Count, "Cosmos query engine should be invoked for multi-source query as entity belongs to cosmos db.");
+            Assert.AreEqual(1, sqlQueryEngine.Invocations.Count, "Sql query engine should not be invoked again for multi-source query as entity2 belongs to cosmos db.");
+            sqlQueryEngine.Verify(x => x.ExecuteAsync(It.Is<FindRequestContext>(ctx => ctx.EntityName == ENTITY_NAME_1)), Times.Once);
+            cosmosQueryEngine.Verify(x => x.ExecuteAsync(It.Is<FindRequestContext>(ctx => ctx.EntityName == ENTITY_NAME_2)), Times.Once);
```
seantleonard, line 227: nit add newline between Asserts from first "Act" operation. It's not easy to read that there are, in fact, two calls to the system under test (SUT) here. 
1. First test -> entity1
2. Second test -> entity2

In the first test, you check CosmosQueryEngine is never invoked, then you verify if CosmosQueryEngine.ExecuteAsync is never invoked. Isn't the second check redundant since zero invocations of CosmosQueryEngine implies zero invocations of CosmosQueryEngine.ExecuteAsync?

Same situation for the second test but you're redundantly checking that SqlQueryEngine.ExecuteAsync is not invoked again after already ensuring SqlQueryENgine isn't invoked again.

seantleonard (commented):
additional questions about tests.

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
rohkhann: Makes sense. fixed

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
rohkhann: Missed this sorry when taking the summary over from previous function. fixed.

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
seantleonard: Does my understanding of your validations reflect what you intended to test? If I'm incorrect, that's fine, just go ahead and add extra comments if needed.

seantleonard (approved):

PR #2181, opened 2024-04-22
.NET 8 Upgrade (Multi Framework targetting)

### LINKED ISSUE #2140: .NET 8 upgrade
Upgrade project to target .NET8

seantleonard:
## Why make this change?

- Closes #2140 

## What is this change?
- Updates all `.csproj` files to target both .NET8 and .NET6. (source: [How to multi target](https://learn.microsoft.com/dotnet/core/tutorials/libraries#how-to-multitarget)
```xml
    <TargetFrameworks>net8.0;net6.0</TargetFrameworks>
```
- Update **Directory.Packages.Props** to include framework specific sections targeting dependency versions compatible with .NET6
  - Ref: https://josef.codes/multi-targeting-your-nuget-packages/
```xml
<ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
  <PackageVersion Include="Microsoft.AspNetCore.Authorization" Version="8.0.4" />
</ItemGroup>
<ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
  <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="6.0.29" />
</ItemGroup>
```
- Added framework specific constructors for EasyAuth/Simulator AuthenticationHandlers because `ISystemClock` was deprecated in .NET8 and is a required constructor parameter when overriding `AuthenticationHandler` dotnet class used for custom authentication handlers.
  - https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/8.0/isystemclock-obsolete
  - [Multi target dotnet project && code](https://learn.microsoft.com/nuget/create-packages/multiple-target-frameworks-project-file#create-a-project-that-supports-multiple-net-framework-versions)

- Updated instances of pulling client role header value (GraphQLFilterParser, SqlMutationEngine) due to stricter nullability checks (now errors) in .NET 8. The following original code "possibly returns null" and needs to be handled.
```csharp
GetHttpContext().Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER];
```
Changed to be a conditional check:
```csharp
if (!GetHttpContext().Request.Headers.TryGetValue(AuthorizationResolver.CLIENT_ROLE_HEADER, out StringValues headerValues) && headerValues.Count != 1)
```
- Due to stricter .NET 8 nullability checks, DabCacheService specification of return type now specifies nullable jsonElement? to be consistent with method signature.
```csharp
async (FusionCacheFactoryExecutionContext<JsonElement?> ctx, CancellationToken ct) =>
```
 - Startup class also gets a fix for stricter nullability when processing config file name. Previously a null value was possible according to the compiler:
 OLD:
```csharp
string configFileName = Configuration.GetValue<string>("ConfigFileName", FileSystemRuntimeConfigLoader.DEFAULT_CONFIG_FILE_NAME);
```
NEW
```csharp
string configFileName = Configuration.GetValue<string>("ConfigFileName") ?? FileSystemRuntimeConfigLoader.DEFAULT_CONFIG_FILE_NAME;
```
## How was this tested?
This isn't a feature addition, but ensure that all existing tests pass.
- [x] Integration Tests
- [x] Unit Tests


JerryNixon:
We know there are performance improvements from this even if we do not measure for them. This includes improvements to JIT compilation and reduced memory consumption. The .NET team has bragged about this too much for us not to see those benefits coming for free. But the best part is the improved developer experience with the latest C# language features—that's the win! 

By the way, .NET 8 is set to be EOL in 2026 (compared to .NET 7 in 2024). The team alternates LTS with only even versions (similar to DAB), which means our next upgrade will be .NET 10, not .NET 9 - which is considered a maintenance version. I bring this up because .NET 9 is right around the corner, and I don't want the team to feel discouraged to see .NET 9 coming out at the same time we upgrade to .NET 8. We're upgrading to the right version, LTS version of .NET, and it's perfect.

.pipelines/templates/build-pipelines.yml
```diff
@@ -50,11 +50,14 @@ steps:
     echo "##vso[task.setvariable variable=\$id]$schemaId"
   displayName: Set dab version
 
+# The .NET CLI commands in proceeding tasks use the .NET SDK version specified here.
+# Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from
```
Aniruddh25, line 54: What does "selecting the version" mean here?

src/Directory.Packages.props
```diff
@@ -1,57 +1,66 @@
...
+        <PackageVersion Include="MSTest.TestFramework" Version="3.2.2" />
+        <PackageVersion Include="MySqlConnector" Version="2.1.5" />
+        <PackageVersion Include="Newtonsoft.Json" Version="13.0.2" />
+        <PackageVersion Include="Npgsql" Version="7.0.1" />
+        <PackageVersion Include="Polly" Version="7.2.3" />
+        <PackageVersion Include="NJsonSchema" Version="10.9.0" />
+        <PackageVersion Include="StyleCop.Analyzers" Version="1.1.118" />
+        <PackageVersion Include="Swashbuckle.AspNetCore.SwaggerUI" Version="6.5.0" />
+        <PackageVersion Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
+        <PackageVersion Include="System.Drawing.Common" Version="8.0.3" />
```
Aniruddh25, line 44: Does our NOTICE file mention the updated version? We might have to test the official pipeline with this PR to find that info

seantleonard:
Issue: https://github.com/dotnet/aspnetcore/issues/53332
> System.InvalidOperationException: Unable to resolve service for type 'Microsoft.AspNetCore.Routing.EndpointDataSource' while attempting to activate 'Microsoft.AspNetCore.Authorization.Policy.AuthorizationPolicyCache'.

Present in ClientRoleHeaderAuthorizationMiddlewareTests

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -161,7 +161,7 @@ public GQLFilterParser(RuntimeConfigProvider runtimeConfigProvider, IMetadataPro
 
                     bool columnAccessPermitted = queryStructure.AuthorizationResolver.AreColumnsAllowedForOperation(
                         entityName: originalEntityName,
-                        roleName: GetHttpContextFromMiddlewareContext(ctx).Request.Headers[CLIENT_ROLE_HEADER],
+                        roleName: GetHttpContextFromMiddlewareContext(ctx).Request.Headers[CLIENT_ROLE_HEADER].ToString(),
```
Aniruddh25, line 164: is it guaranteed that client role header is a non-null? In other words, are we sure ToString will not cause null dereference

Nuget.config
```diff
@@ -2,7 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="data_api_builder_build_packages" value="https://pkgs.dev.azure.com/sqldab/fcb212b3-b288-4c9e-b55a-5842a268b16d/_packaging/data_api_builder_build_packages/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
```
Aniruddh25, line 5: We shouldnt use nuget.org directly, should upstream the necessary packages to our central packages feed

src/Core/AuthenticationHelpers/AppServiceAuthentication.cs
```diff
@@ -69,11 +69,11 @@ public class AppServiceClaim
     {
         ClaimsIdentity? identity = null;
 
-        if (context.Request.Headers.TryGetValue(AuthenticationOptions.CLIENT_PRINCIPAL_HEADER, out StringValues header))
+        if (context.Request.Headers.TryGetValue(AuthenticationOptions.CLIENT_PRINCIPAL_HEADER, out StringValues headerValues) && headerValues.Count == 1)
```
Aniruddh25, line 72: Should it be `out StringValues?` since the value may not exist?

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -94,7 +94,7 @@ public async Task InvokeAsync(HttpContext httpContext)
             // in an authenticated requests.
             if (isAuthenticatedRequest)
             {
-                clientDefinedRole = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER];
+                clientDefinedRole = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER].ToString();
```
Aniruddh25, line 97: are we sure the client role header will exist? What if the TryAdd operation above failed due to reasons other than key being already present?

src/Core/AuthenticationHelpers/EasyAuthAuthenticationHandler.cs
```diff
@@ -21,22 +21,40 @@ namespace Azure.DataApiBuilder.Core.AuthenticationHelpers;
 /// </summary>
 public class EasyAuthAuthenticationHandler : AuthenticationHandler<EasyAuthAuthenticationOptions>
 {
+#if NET8_0_OR_GREATER
```
Aniruddh25, line 24: where is this macro defined? NET8_0_OR_GREATER?

Aniruddh25 (changes requested):
Waiting for usage of nuget.org

seantleonard:
> Issue: [dotnet/aspnetcore#53332](https://github.com/dotnet/aspnetcore/issues/53332)
> 
> > System.InvalidOperationException: Unable to resolve service for type 'Microsoft.AspNetCore.Routing.EndpointDataSource' while attempting to activate 'Microsoft.AspNetCore.Authorization.Policy.AuthorizationPolicyCache'.
> 
> Present in ClientRoleHeaderAuthorizationMiddlewareTests

Fixed this via: https://github.com/dotnet/aspnetcore/issues/53332#issuecomment-2091861884

src/Core/AuthenticationHelpers/EasyAuthAuthenticationHandler.cs
seantleonard: Defined in the .NET8 SDK per: https://learn.microsoft.com/en-us/dotnet/standard/frameworks#preprocessor-symbols

in row for "Target Frameworks" : ".NET 5+ (and .NET Core)"

Nuget.config
seantleonard: ah my bad. That was to unblock me from testing this without going through manually adding all those package upstream. Reverting.

.pipelines/templates/build-pipelines.yml
seantleonard: clarifying with comment. "selecting the version" means installing a specific SDK version as done in this task.
```csharp
# The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
# Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
# specifying the runtime version a project targets."
```

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
seantleonard: I see that the only other way `TryAdd` fails is if the header dictionary is null:
```csharp
        public static bool TryAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, TValue value)
        {
            if (dictionary is null)
            {
                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.dictionary);
            }

            if (!dictionary.ContainsKey(key))
            {
                dictionary.Add(key, value);
                return true;
            }

            return false;
        }
```

Then, to answer "can the headerDictionary be null"? It seems the answer is "No" per source code:
 - [DefaultHttpRequest.cs](https://github.com/dotnet/aspnetcore/blob/6ef0fe5cffc59fec5a68275b7178db877d8b9396/src/Http/Http/src/Internal/DefaultHttpRequest.cs#L133-L136)
 ```csharp
    public override IHeaderDictionary Headers
    {
        get { return HttpRequestFeature.Headers; }
    }
```
  - [HttpRequestFeature.cs] referenced by DefaultHttpRequest(https://github.com/dotnet/aspnetcore/blob/main/src/Http/Http/src/Features/HttpRequestFeature.cs)
```csharp
 public HttpRequestFeature()
    {
        Headers = new HeaderDictionary(); // <-- always created.
        Body = Stream.Null;
        Protocol = string.Empty;
        Scheme = string.Empty;
        Method = string.Empty;
        PathBase = string.Empty;
        Path = string.Empty;
        QueryString = string.Empty;
        RawTarget = string.Empty;
    }
```

src/Core/AuthenticationHelpers/AppServiceAuthentication.cs
seantleonard: `out StringValues?` is a compiler error as the nullable type violates out param expected type
image

seantleonard: if the value doesn't exist, and empty StringValues collection is returned per source code:
https://github.com/dotnet/aspnetcore/blob/8486d31e24f30e3fa1809a95699a0adc16f448d7/src/Http/Http.Features/src/IHeaderDictionary.cs#L14

```csharp
    /// <summary>
    /// IHeaderDictionary has a different indexer contract than IDictionary,
 where it will return StringValues.Empty for missing entries.
    /// </summary>
    /// <param name="key"></param>
    /// <returns>The stored value, or StringValues.Empty if the key is not present.</returns>
```

seantleonard: this then begs the question: " so if an empty stringvalues collection is returned, how do you protect against processing an empty collection?"

- immediately after trygetvalue(), have to check that there is exactly one element in the collection.

src/Core/Models/GraphQLFilterParsers.cs
seantleonard: Yes, guaranteed that client role header is non-null:
- ClientRoleHeaderAuthenticationMiddleware -> guaranteed that x-ms-api-role injected, non-authN case: value will be `anonymous`

```mermaid
  graph TD;
      Req_Received_by_DAB --> ClientRoleHeaderAuthenticationMiddleware
      ClientRoleHeaderAuthenticationMiddleware -->HotChocolateMiddleware;
      HotChocolateMiddleware-->GraphQLFilterParser;

File: src/Directory.Packages.props (hunk 2 above)
Comment (seantleonard) [reply src/Directory.Packages.props:44]: updated the notice file creation script and will check output.

File: Nuget.config (hunk 4 above)
Comment (seantleonard) [reply Nuget.config]: reverted.

Comment (seantleonard) [reply Nuget.config]: and also updated upstream feed with latest versions of packages.

Microsoft.AspNetCore.Authorization
Microsoft.AspNetCore.Authentication.JwtBearer
Microsoft.AspNetCore.Mvc.Testing
Microsoft.AspNetCore.TestHost
6.0.29
8.0.4

Verify.MsTest
23.6.0

Verify.DiffPlex
2.3.0

ZiggyCreatures.FusionCache
1.1.0
1.0.0

System.IO.Abstractions.TestingHelpers
System.IO.Abstractions
21.0.2

System.IdentityModel.Tokens.Jwt
7.5.1
7.5.0

System.Drawing.Common
8.0.4

MSTest.TestFramework
MSTest.TestAdapter
3.3.1
3.2.2

Microsoft.OpenApi.Readers
Microsoft.OpenApi
1.6.14

Microsoft.OData.Core
Microsoft.OData.Edm
7.21.1
7.20.0

Microsoft.NET.Test.Sdk
17.9.0

Microsoft.IdentityModel.JsonWebTokens
7.5.1
7.5.0

Microsoft.Extensions.Logging.ApplicationInsights
Microsoft.ApplicationInsights.AspNetCore
2.22.0

Microsoft.Extensions.Configuration.Json
Microsoft.Extensions.Configuration.Binder
8.0.0

Microsoft.Data.SqlClient
5.2.0

Microsoft.AspNetCore.Http
2.2.2

DotNetEnv
3.0.0

Azure.Identity
1.11.2

File: Dockerfile [hunk 8]
```diff
@@ -1,8 +1,11 @@
 # Version values referenced from https://hub.docker.com/_/microsoft-dotnet-aspnet
 
-FROM mcr.microsoft.com/dotnet/sdk:8.0-cbl-mariner2.0. AS build
+FROM mcr.microsoft.com/dotnet/sdk:6.0-cbl-mariner2.0. AS build
+
+
 WORKDIR /src
-FROM mcr.microsoft.com/dotnet/aspnet:8.0-cbl-mariner2.0 AS runtime
+COPY [".", "./"]
+RUN dotnet build "./src/Service/Azure.DataApiBuilder.Service.csproj" -c Docker -o /out -r linux-x64
```
Aniruddh25, line 8: are we bringing this -r `linux-x64` back? so rebuilding it intentionally as opposed to copying from the built artifacts?

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
```diff
@@ -75,6 +74,12 @@ public static class WebHostBuilderHelper
 
                             services.AddSingleton(runtimeConfigProvider);
 
+                            // https://github.com/dotnet/aspnetcore/issues/53332#issuecomment-2091861884
+                            // AddRouting() adds required services that .NET8 does not add by default
+                            // Without this, .NET8 fails to resolve the required services for the middleware
+                            // and tests fail.
+                            services.AddRouting();
```
Aniruddh25, line 81: Does doing this on .NET 6 explicitly cause any side effects?

Dockerfile
seantleonard: I didn't make changes to the Dockerfile... I don't see that in the changeset listed by GitHub either:
image

src/Directory.Packages.props
seantleonard: noticefile generates successfully with newest sqlclient version specified.

Dockerfile
seantleonard: this was merging from main. updated in #2182 . This is the Dockerfile that had been historically present in our repo and was restored from adopting any OB related changes.

seantleonard: The dockerfile in this repo does not contribute to the binaries built and released. Separate internal dockerfile is used.

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
seantleonard: No observable side effects. The usage of `services.AddRouting()` adds relevant services to the DI container. This is slightly different from `app.UseRouting()` which adds middleware to the request execution pipeline.

Aniruddh25 (approved):
LGTM

src/Service.Tests/Authentication/JwtTokenAuthenticationUnitTests.cs
```diff
@@ -215,22 +215,36 @@ public async Task TestInvalidToken_InvalidSigningKey()
...
         [TestMethod("JWT signed with unrecognized/unconfigured key, results in signature key not found")]
         public async Task TestInvalidToken_InvalidSignature()
         {
-            RsaSecurityKey key = new(RSA.Create(2048));
-            string token = CreateJwt(audience: AUDIENCE, issuer: LOCAL_ISSUER, signingKey: key);
-            string tokenForgedSignature = ModifySignature(token, removeSig: false);
-
-            HttpContext postMiddlewareContext = await SendRequestAndGetHttpContextState(key, tokenForgedSignature);
+            // Arrange
+            RsaSecurityKey tokenIssuerSigningKey = CreateRsaSigningKeyForTest();
```
Aniruddh25, line 227: Which function necessitated this test change for .NET 8 upgrade?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -256,8 +256,11 @@ public void TestColumnDefinitionNegativeCases()
         /// <summary>
         /// Validates serialization and deserilization of Dictionary containing DatabaseTable
         /// this is how we serialize and deserialize metadataprovider.EntityToDatabaseObject dict
+        /// [Test Ignored] See https://github.com/dotnet/runtime/issues/41920#issuecomment-688497802
```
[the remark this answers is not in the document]
seantleonard: Yes, removed because this breaks in .net8. 

The custom DatabaseObjectConverter experiences deserialization issue, I believe, due to 
- Breaking Change: https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/8.0/metadata-resolving

Here is a diff of the deserialized json text that the test creates and feeds to the converter:
image

image

Our tests run for both .net8 and .net6 target frameworks. However, locally in visual studio , even though the TestExplorer shows tests for both target frameworks, this particular test fails for .net6 when I have dotnet8 installed. I don't know why that is because i'd expect .net6 tests to use .net6 runtime/sdk.

Here is sample pipeline logs that show tests running for both target frameworks

> Test run for D:\a\1\s\src\out\tests\net8.0\Azure.DataApiBuilder.Service.Tests.dll (.NETCoreApp,Version=v8.0)
> Passed!  - Failed:     0, Passed:  1162, Skipped:     2, Total:  1164, Duration: 8 m 59 s - Azure.DataApiBuilder.Service.Tests.dll (net8.0)

> Test run for D:\a\1\s\src\out\tests\net6.0\Azure.DataApiBuilder.Service.Tests.dll (.NETCoreApp,Version=v6.0)
> Passed!  - Failed:     0, Passed:  1162, Skipped:     2, Total:  1164, Duration: 9 m 10 s - Azure.DataApiBuilder.Service.Tests.dll (net6.0)

src/Cli.Tests/Cli.Tests.csproj
```diff
@@ -14,7 +14,10 @@
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="coverlet.msbuild" />
+        <PackageReference Include="coverlet.msbuild">
+          <PrivateAssets>all</PrivateAssets>
```
ayush3797, line 18: why was this change required?

ayush3797 (approved):
LGTM!

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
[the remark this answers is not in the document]
seantleonard: Updated so that the objectConverter isn't used in the .net 8 test. I now don't need to ignore the test at all. Test passes for both .net6 and .net8.

src/Service.Tests/Authentication/JwtTokenAuthenticationUnitTests.cs
seantleonard: Upgrading to dotnet8 required that I update `System.IdentityModel.Tokens.Jwt`, which transitively depends on newer version of `Microsoft.IdentityModel.Tokens.Jwt` that had a breaking change requiring stricter parameter values on the RsaKey object created for signing jwt tokens in the test fixture.

src/Cli.Tests/Cli.Tests.csproj
seantleonard: this change seems to mirror that sub-properties of the other package reference: `coverlet.collector` 
The intent is to limit the content sent to the build output folder from this dependency: https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

Aniruddh25 (approved):

PR #2183, opened 2024-04-23
Update Dockerfile to copy Linux instead of Windows build contents

abhishekkumams:
## Why make this change?

- Currently it is copying the windows build file. which is breaking a container scenario as it expects linux build files.

## What is this change?

- updated the onebranch dockerfile to not copy the windows build.
- instead it should copy the all the files from the directory specified in the docker task in our one branch pipeline.
- we have updated the docker task to use only the linux build while runnning the docker build task.

## How was this tested?
- verified the files in the container
image


seantleonard (approved):

Aniruddh25 (commented):
Fix description

abhishekkumams:
not required anymore

Issue #887, opened 2022-10-17
DAB Engine Cosmos Improvement : Given key 'x' was not present in the dicitonary


sajeetharan:
Engine should throw a valid error to the user regarding the configuration file.

Scenario : If user does not have entities in the entities section and run the engine. It throws an error as 
"Given key 'x' was not present in the dicitonary" 

Expected result : It should give a proper error saying the entities are not present in the configuration file or something.

PR #2186, opened 2024-04-24
Changed Error Message if Entity is not configured in dab-config json

sourabh1007:
## Why make this change?

As mentioned here #887, if entity is not configured in DAB Config JSON file, it throws a generic exception i.e. keyNotFoundException.

As part of this PR, changed this message to make it clearer.

## What is this change?

If Entity is not configured then DAB will throw `System.ApplicationException` with error message as _"The entity '<EntityName>' was not found in the dab-config json"_

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)


sajeetharan (approved):
LGTM

neeraj-sharma2592 (approved):

src/Config/ObjectModel/RuntimeEntities.cs
```diff
@@ -56,7 +56,21 @@ public bool ContainsKey(string key)
...
+        get
+        {
+            Entity? entity;
+            if (Entities.TryGetValue(key, out entity) && entity is not null)
+            {
+                return entity;
+            }
+            else
+            {
+                throw new ApplicationException($"The entity '{key}' was not found in the dab-config json");
```
seantleonard, line 70: why ApplicationException and not DataApiBuilderException which is used throughout the project?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2937,6 +2938,63 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
...
+
+                object payload = new { query };
+
+                HttpRequestMessage graphQLRequest = new(HttpMethod.Post, "/graphql")
+                {
+                    Content = JsonContent.Create(payload)
+                };
+
+                ApplicationException ex = await Assert.ThrowsExceptionAsync<ApplicationException>(async () => await client.SendAsync(graphQLRequest));
+                Assert.AreEqual(ex.Message, "The entity 'Planet' was not found in the dab-config json");
```
seantleonard, line 2994: why is this test sending a GraphQL request? if the entity doesn't exist in the config, shouldn't there be a startup error? 

Alternatively, if the intent is to capture some sort of error that occurs during runtime, then erroring on a non-existent entity should be handled by HotChocolate which detects that the requested entity is not part of the GraphQL schema -> HTTP400 bad request/GraphQL error because request doesn't adhere to schema.

seantleonard:
PR description nor the linked issue describe where this issue was surfacing: as a response to a request, or during DAB startup. Please include a before/after request that shows how your fix appears to the end user.

src/Config/ObjectModel/RuntimeEntities.cs
sourabh1007: It makes sense, Since, it is a small change, making this change in patch PR

sourabh1007: @sajeetharan what do you think?

src/Service.Tests/Configuration/ConfigurationTests.cs
sourabh1007: It is not schema error, it is config error, if there is an entity which is defined in schema but not defined in config then, `Hotchocolate` wouldn't even know about it since schema has this information. It will fail when DAB check for the entity in the config.

seantleonard: can you add more details to the test description then? there's no context in the test describing what mismatch exists between schema.gql and dab-config.json. The request is querying "EntityName", but the error message says "Planet". That's not really actionable by the end user because it's not immediately clear what the relation is (or isn't) between "EntityName" and "Planet"

src/Config/ObjectModel/RuntimeEntities.cs
seantleonard: also message should be something like:
"... was not found in the runtime config."

using "dab-config json" is too close to "dab-config.json" which is an explicit file name. End-users most likely aren't using `dab-config.json` as the file name (more likely dab-config.mssql.json, dab-config.cosmos.json etc.) so this may add confusion.

Issue #2157, opened 2024-04-10
Support multiple create for self-referencing relationships


severussundar:
PR https://github.com/Azure/data-api-builder/pull/2138 adds support for querying relationships that reference the same entity (self-join tables).

Extend multiple create functionality to support relationships that reference itself. 

Some things to consider are:
- Accurate determination of referencing/referenced entity. Though, referenced and referencing entity are both the same, we need to be able to identify whether **top level item** (or) **related item** should be created first.
- The current data structure maintains the relationship metadata at the entity level. PR https://github.com/Azure/data-api-builder/pull/2138 introduces changes to maintain the relationship metadata at the granularity level of relationship. This new data structure should be used during query generation and rest of the mutation flows.

PR #2189, opened 2024-04-26
Validation to disallow multiple create on related entities backed by same database table

ayush3797:
## Why make this change?

DAB doesn't support multiple-create on related entities which are backed by same database table. This PR adds a validation to catch that scenario and throw a meaningful exception to the end user. Issue to track support: #2157 

## What is this change?

- Adds a validation during order determination logic in `MultipleCreateOrderHelper.GetReferencingEntityName()` method to catch the scenario.
- In the method signature for `MultipleCreateOrderHelper.GetReferencingEntityName()`, a new parameter isMNRelationship is added with default value of false. The reasoning is explained in the section.

#### Reason for why `MultipleCreateOrderHelper.GetReferencingEntityName()` is called even for M:N relationships when we know that we are always going to return an empty string?
1. For M:N relationships, we assume one of the source/target entity as the referencing entity when we do insertion during query execution because any of the entity can be considered as the referencing/referenced entity. However this does not hold true when the source and target entities are backed by same database tables - in which case we cannot consider any of the entity as the referenced/referencing entity.
2. The not-supported use case - that of not allowing multiple-create via relationships in which source and target entities are backed by same database tables is because DAB cannot determine a valid order of insertion. Since this concerns order determination, it is better to keep the code inside the `MultipleCreateOrderHelper.GetReferencingEntityName()` method.
3. Keeping this in `MultipleCreateOrderHelper` provides unit test coverage even for the other 2 db types i.e. PgSql,MySql.
 
## How was this tested?
 - Added test MultipleCreateOrderHelperUnitTests.TestExceptionForSelfReferencingRelationships() to test the change for the 3 dbs - MySql/MsSql/PgSql.

## Sample Request(s)

1. Config:
 
image

Request:

image


src/Core/Resolvers/MultipleCreateOrderHelper.cs
```diff
@@ -61,6 +61,17 @@ public class MultipleCreateOrderHelper
                     subStatusCode: DataApiBuilderException.SubStatusCodes.EntityNotFound);
             }
 
+            DatabaseTable sourceDbTable = (DatabaseTable)sourceDbObject;
+            DatabaseTable targetDbTable = (DatabaseTable)targetDbObject;
+            if (sourceDbTable.Equals(targetDbTable))
+            {
+                throw new DataApiBuilderException(
+                message: $"Multiple-create for relationship: {relationshipName} at level: {nestingLevel} having the " +
+                $"source entity: {sourceEntityName} and target entity: {targetEntityName} backed by same database table is not yet supported.",
```
seantleonard, line 70: $"Multiple-create for relationship: {relationshipName} at level: {nestingLevel} is not supported "+
$" because the source entity: {sourceEntityName} and target entity: {targetEntityName} are backed by same database table.",

nit: for clarity/conciseness

seantleonard (commented):
are tests dependent on #2138

ayush3797:
> are tests dependent on #2138

Are not dependent. But will require additional entities in the config and some more changes. Keeping bug-bash in mind, trying to get the logic changes in and test can follow.

ayush3797:
Now I'm onto adding test.

src/Service.Tests/Unittests/MultipleCreateUnitTests/MultipleCreateOrderHelperUnitTests.cs
```diff
@@ -374,6 +375,23 @@ public void ValidateReferencingEntityBasedOnEntityMetadata()
 
         #endregion
 
+        #region Order determination test for relationships having source/target entities backed by same database table
+
+        /// <summary>
+        /// Test to validate that when multiple-create is executed for a relationship for which source and target entities are backed by the
+        /// same database tables, we throw an appropriate exception because DAB currently does not support multiple-create for such relationships.
```
ayush3797, line 382:
```suggestion
        /// same database table, we throw an appropriate exception because DAB currently does not support multiple-create for such relationships.
```

src/Service.Tests/Unittests/MultipleCreateUnitTests/MultipleCreateOrderHelperUnitTests.cs
```diff
@@ -374,6 +375,23 @@ public void ValidateReferencingEntityBasedOnEntityMetadata()
...
+        /// </summary>
+        [TestMethod]
+        public void TestExceptionForSelfReferencingRelationships()
+        {
+            // Identical source and target entities backed by the database table 'books'.
```
ayush3797, line 387:
```suggestion
            // Identical source and target entities backed by the same database table 'books'.
```

config-generators/mysql-commands.txt
```diff
@@ -2,6 +2,7 @@ init --config "dab-config.MySql.json" --database-type mysql --connection-string
 add Publisher --config "dab-config.MySql.json" --source publishers --permissions "anonymous:read"
 add Stock --config "dab-config.MySql.json" --source stocks --permissions "anonymous:create,read,update,delete"
 add Book --config "dab-config.MySql.json" --source books --permissions "anonymous:create,read,update,delete" --graphql "book:books"
+add BookNF --config "dab-config.MySql.json" --source books --permissions "anonymous:*" --rest true --graphql "bookNF:booksNF"
```
Aniruddh25, line 5: what does the suffix NF stand for?

src/Core/Resolvers/MultipleCreateOrderHelper.cs
```diff
@@ -24,13 +24,14 @@ public class MultipleCreateOrderHelper
...
         /// When visualized as a graphQL mutation request, 
-        ///   Source entity refers to the top level entity
+        ///   Source entity refers to the top level entity in whose configuration the relationship is defined.
         ///   Target entity refers to the related entity.
         ///   
         /// This method handles the logic to determine the referencing entity for relationships from (source, target) with cardinalities:
         /// 1. 1:N - Target entity is the referencing entity
         /// 2. N:1 - Source entity is the referencing entity
         /// 3. 1:1 - Determined based on foreign key constraint/request input data.
+        /// 4. M:N - None of the source/target entity is the referencing entity. Instead, linking table act as the referencing table.
```
Aniruddh25, line 34:
```suggestion
        /// 4. M:N - None of the source/target entity is the referencing entity. Instead, linking table acts as the referencing entity.
```

src/Core/Resolvers/MultipleCreateOrderHelper.cs
```diff
@@ -61,6 +64,26 @@ public class MultipleCreateOrderHelper
                     subStatusCode: DataApiBuilderException.SubStatusCodes.EntityNotFound);
             }
 
+            DatabaseTable sourceDbTable = (DatabaseTable)sourceDbObject;
```
Aniruddh25, line 67: How can we be sure the sourceDbObject is a DatabaseTable and not a view/stored proc? We should check for it before casting.

src/Core/Resolvers/MultipleCreateOrderHelper.cs
```diff
@@ -61,6 +64,26 @@ public class MultipleCreateOrderHelper
...
+            DatabaseTable targetDbTable = (DatabaseTable)targetDbObject;
+            if (sourceDbTable.Equals(targetDbTable))
+            {
+                //  DAB does not yet support multiple-create for self referencing relationships where both the source and
+                //  target entities are backed by same database table.
+                throw new DataApiBuilderException(
+                message: $"Multiple-create for relationship: {relationshipName} at level: {nestingLevel} is not supported because " +
```
Aniruddh25, line 74: Are queries supported when there is a relationship between entities backed by same tables? 

And the PR title only says mutation is not allowed. We should be specific in the title to qualify that "multiple" mutation isnt allowed.

src/Core/Resolvers/MultipleCreateOrderHelper.cs
```diff
@@ -61,6 +64,26 @@ public class MultipleCreateOrderHelper
...
+                $"source entity: {sourceEntityName} and target entity: {targetEntityName} are backed by same database table.",
+                statusCode: HttpStatusCode.BadRequest,
+                subStatusCode: DataApiBuilderException.SubStatusCodes.NotSupported);
+            }
+
+            if (isMNRelationship)
+            {
+                // For M:N relationships,'neither the source nor the target entity act as the referenced entity.
```
Aniruddh25, line 82:
```suggestion
                // For M:N relationships, neither the source nor the target entity act as the referencing entity.
```

Aniruddh25: why is this change part of this validation PR, it seems like this is a bug fix

src/Service.Tests/Unittests/MultipleCreateUnitTests/MultipleCreateOrderHelperUnitTests.cs
```diff
@@ -374,6 +375,23 @@ public void ValidateReferencingEntityBasedOnEntityMetadata()
 
         #endregion
 
+        #region Order determination test for relationships having source/target entities backed by same database table
+
+        /// <summary>
+        /// Test to validate that when multiple-create is executed for a relationship for which source and target entities are backed by the
+        /// same database table, we throw an appropriate exception because DAB currently does not support multiple-create for such relationships.
```
Aniruddh25, line 382: do we expect to add such a support in the future? Is it even possible?

src/Service.Tests/Unittests/MultipleCreateUnitTests/MultipleCreateOrderHelperUnitTests.cs
```diff
@@ -374,6 +375,23 @@ public void ValidateReferencingEntityBasedOnEntityMetadata()
...
+        /// </summary>
+        [TestMethod]
+        public void TestExceptionForSelfReferencingRelationships()
+        {
+            // Identical source and target entities backed by the same database table 'books'.
+            ValidateExceptionForSelfReferencingRelationship(sourceEntityName: "Book", targetEntityName: "Book");
+
+            // Different source and target entities backed by the same database table 'books'.
+            ValidateExceptionForSelfReferencingRelationship(sourceEntityName: "Book", targetEntityName: "BookNF");
+        }
```
Aniruddh25, line 392:
```suggestion
        }
        
```

src/Service.Tests/Unittests/MultipleCreateUnitTests/MultipleCreateOrderHelperUnitTests.cs
```diff
@@ -395,6 +413,35 @@ public void ValidateReferencingEntityBasedOnEntityMetadata()
...
+                columnDataInSourceBody: new(),
+                targetNodeValue: null,
+                nestingLevel: 1));
+
+            // Assert that the exception is as expected.
+            Assert.AreEqual(HttpStatusCode.BadRequest, ex.StatusCode);
+            Assert.AreEqual(DataApiBuilderException.SubStatusCodes.NotSupported, ex.SubStatusCode);
+            Assert.AreEqual($"Multiple-create for relationship: testRelationship at level: 1 is not supported because " +
+                $"source entity: {sourceEntityName} and target entity: {targetEntityName} are backed by same database table.", ex.Message);
+        }
```
Aniruddh25, line 444:
```suggestion
        }
        
```

Aniruddh25 (dismissed):
Waiting on safe typecast to `DatabaseTable`

src/Core/Resolvers/MultipleCreateOrderHelper.cs
```diff
@@ -48,7 +50,8 @@ public class MultipleCreateOrderHelper
             ISqlMetadataProvider metadataProvider,
             Dictionary<string, IValueNode?> columnDataInSourceBody,
             IValueNode? targetNodeValue,
-            int nestingLevel)
+            int nestingLevel,
+            bool isMNRelationship = false)
```
seantleonard, line 54: if a caller knows `isMNRelationship` is true, then they shouldn't call this function to begin with.

src/Core/Services/MultipleMutationInputValidator.cs
```diff
@@ -364,7 +356,18 @@ public MultipleMutationInputValidator(IMetadataProviderFactory sqlMetadataProvid
...
+
+            if (isMNRelationship)
+            {
+                // The presence of a linking object indicates that an M:N relationship exists between the current entity and the target/child entity.
+                // The linking table acts as a referencing table for both the source/target entities which act as
+                // referenced entities. Consequently:
+                // - Column values for the target entity can't be derived from insertion in the current entity.
+                // - Column values for the current entity can't be derived from the insertion in the target/child entity.
+                return;
+            }
```
seantleonard, line 370: this check for `isMNRelationship` should occur before the call to `MultipleCreateOrderHelper.GetReferencingEntityName(...)` so that you can return early.

seantleonard (changes requested):
dont need `isMNRelationship` as a param to `getreferencingEntityName` otherwise looks good. As Ani mentioned, casting feedback needs addressing

seantleonard (approved):
need not block on my feedback, only Ani's. Go ahead and make those changes and you should be good from my end. Let me know if you think we really need to keep `isMNRelationship` param.

src/Core/Resolvers/MultipleCreateOrderHelper.cs
ayush3797: Not a bug fix. Previously, this method was not even called for M:N relationships. Reasoning added to the PR description.

src/Core/Resolvers/MultipleCreateOrderHelper.cs
ayush3797: Discussed over call. Reasoning added to the PR description.

src/Service.Tests/Unittests/MultipleCreateUnitTests/MultipleCreateOrderHelperUnitTests.cs
ayush3797: I think it will be possible to add such a support maybe for some use cases. But for now, we can wait for any such customer use cases.

src/Core/Resolvers/MultipleCreateOrderHelper.cs
ayush3797: Added

src/Core/Resolvers/MultipleCreateOrderHelper.cs
ayush3797: @seantleonard, could you help answer the first question? I'm assuming we will have the support atleast for self-referencing relationships, right?

abhishekkumams (approved):

src/Core/Resolvers/MultipleCreateOrderHelper.cs
seantleonard: for question 1: 
- #2138 handles a single entity which defines "self-joining" relationships
- TBD on two separate entities which define relationships to each other. TBD because I think there is an open issue and then I need to test within pr 2138 if the scenario works. 
Shouldn't block this pr though.

PR #2192, opened 2024-04-29
Cosmos DB: Adds circular reference check for entities in graphQL schema

sourabh1007:
## Why make this change?
Today, if graphQL schema has circular reference of entities, DAB won't be able to load the schema and die with stackoverflow exception when it tries to traverse the schema.
With JSON data, it can get very complicated to handle all the scenarios with circular reference.

## What is this change?
Adding a proper exception during the load, if DAB identifies that schema has circular reference.

If schema is found with circular reference, DAB will throw below exception.

**DAB Exception**
``` c#
DataApiBuilderException(
                        message: $"Circular reference detected in the schema for entity '{entityType}'.",
                        statusCode: System.Net.HttpStatusCode.InternalServerError,
```

**Schema with circular reference**
```gql

type Character {
    id : ID,
    name : String,
    moons: [Moon],    // Character has Moon Reference 
}
type Planet @model(name:""Planet"") {
    id : ID!,
    name : String,
    character: Character
}
type Moon {
    id : ID,
    name : String,
    details : String,
    character: Character // Moon has Character Reference
}
```

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests


src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -195,18 +195,40 @@ private void ParseSchemaGraphQLFieldsForJoins()
             Dictionary<string, ObjectTypeDefinitionNode> schemaDocument,
             string currentPath,
             IncrementingInteger tableCounter,
-            EntityDbPolicyCosmosModel? previousEntity = null)
+            EntityDbPolicyCosmosModel? previousEntity = null,
+            HashSet<string>? visitedEntity = null)
```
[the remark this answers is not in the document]
seantleonard: unclear what `previousEntity` means and how it relates to `EntityDbPolicyCosmosModel`

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2938,6 +2959,34 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
             }
         }
 
+        [TestMethod, TestCategory(TestCategory.COSMOSDBNOSQL)]
+        public void ValidateGraphQLSchemaForCircularReference()
```
seantleonard, line 2963: need test summary. What is the circular reference?  which GraphQL type is processing expected to fail and throw exception? (character.moons or moons.character?)

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -425,6 +425,27 @@ public class ConfigurationTests
...
+    name : String,
+    character: Character
+}
+
+type Moon {
+    id : ID,
+    name : String,
+    details : String,
+    character: Character
+}
```
seantleonard, line 447: because Character.moons seems to be an array type (`[Moon]`), does your code work differently if Moon.character were processed last vs. Character.moons?

seantleonard: any other examples of circular reference or does this cover all circumstances?

seantleonard (changes requested):
some questions before merging.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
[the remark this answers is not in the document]
sourabh1007: Updated summary, fixed formatting and added more doc.

src/Service.Tests/Configuration/ConfigurationTests.cs
sourabh1007: Circular Reference is a general term, I think it is not really required to explain explicitly what it really means. Anyway, I tried to put more information around.

src/Service.Tests/Configuration/ConfigurationTests.cs
sourabh1007: yes, it should cover all circumstances. Let me know if want me to cover any particular scenario. I added one more test to cover array and no-array object in circular reference separately.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: I didn't mean to explain the term circular reference. I meant for you to point out, in your sample `schema.gql` file, which properties are you expected to be detected as circular references.

neeraj-sharma2592 (approved):
LGTM

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -187,26 +187,49 @@ private void ParseSchemaGraphQLFieldsForJoins()
...
+        /// <param name="visitedEntity"> Keeps a track of the path in an entity, to detect circular reference</param>
+        /// <remarks>It detects the circular reference in the schema while processing the schema and throws <seealso cref="DataApiBuilderException"/> </remarks>
+        private void ProcessSchema(
+            IReadOnlyList<FieldDefinitionNode> fields,
             Dictionary<string, ObjectTypeDefinitionNode> schemaDocument,
             string currentPath,
             IncrementingInteger tableCounter,
-            EntityDbPolicyCosmosModel? previousEntity = null)
+            EntityDbPolicyCosmosModel? parentEntity = null,
+            HashSet<string>? visitedEntity = null)
```
Aniruddh25, line 204: nit: since it is a set, the name of this variable should be `visitedEntities`

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -187,26 +187,49 @@ private void ParseSchemaGraphQLFieldsForJoins()
...
+                {
+                    trackerForField = visitedEntity;
+                }
+
                 string entityType = field.Type.NamedType().Name.Value;
+
                 // If the entity is not in the runtime config, skip it
                 if (!_runtimeConfig.Entities.ContainsKey(entityType))
                 {
                     continue;
```
Aniruddh25, line 221: if the entity is not present in the runtime config, the check for circular reference below will be skipped. Is that the intention? Will we allow circular references if the entity is not present in the config?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -425,6 +425,48 @@ public class ConfigurationTests
             }
         }";
 
+        internal const string GRAPHQL_SCHEMA_WITH_CYCLE_ARRAY = @"
```
Aniruddh25, line 428: why is this a configuration test? It is specific to cosmosdb, should be in one of the cosmos db tests.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2938,6 +2980,42 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
             }
         }
 
+        /// <summary>
+        /// In CosmosDB NoSQL, we store data in the form of JSON.Practically, JSON can be very complex.
```
Aniruddh25, line 2984:
```suggestion
        /// In CosmosDB NoSQL, we store data in the form of JSON. Practically, JSON can be very complex.
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2938,6 +2980,42 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
...
+        /// But DAB doesn't support JSON with circular references e.g if 'Character.Moon' is a valid JSON Path, the
```
Aniruddh25, line 2985:
```suggestion
        /// But DAB doesn't support JSON with circular references e.g if 'Character.Moon' is a valid JSON Path, then
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2938,6 +2980,42 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
...
+            if (!baseLoader.TryLoadKnownConfig(out RuntimeConfig baseConfig))
+            {
+                throw new ApplicationException("Failed to load the default CosmosDB_NoSQL config and cannot continue with tests.");
+            }
+
+            // Setup a mock file system, and use that one with the loader/provider for the config
+            MockFileSystem fileSystem = new(new Dictionary<string, MockFileData>()
+            {
+                { @"../schema.gql", new MockFileData(schema) },
+                { FileSystemRuntimeConfigLoader.DEFAULT_CONFIG_FILE_NAME,
```
Aniruddh25, line 3006: nit: `FileSystemRuntimeConfigLoader` seems not needed.. I see that in light font

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2938,6 +2980,42 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
...
+                    new MockFileData(baseConfig.ToJson()) }
+            });
+            FileSystemRuntimeConfigLoader loader = new(fileSystem);
+            RuntimeConfigProvider provider = new(loader);
+
+            DataApiBuilderException exception =
+                Assert.ThrowsException<DataApiBuilderException>(() => new CosmosSqlMetadataProvider(provider, fileSystem));
+            Assert.AreEqual("Circular reference detected in the provided GraphQL schema for entity 'Character'.", exception.Message);
+            Assert.AreEqual(System.Net.HttpStatusCode.InternalServerError, exception.StatusCode);
```
Aniruddh25, line 3015: `System.Net` seems to not be needed.

Aniruddh25 (changes requested):
Requires circular condition check even if runtime config doesnt contain the entity

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
sourabh1007: It would fail during operation when customer try to use that entity. (existing behavior)

src/Service.Tests/Configuration/ConfigurationTests.cs
sourabh1007: This file contains test specific to "Configuration". Database type doesn't matter. I can see cosmos related code in this file already.

src/Service.Tests/Configuration/ConfigurationTests.cs
sourabh1007: it is required to

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -193,24 +193,24 @@ private void ParseSchemaGraphQLFieldsForJoins()
...
             // Traverse the fields and add them to the path
             foreach (FieldDefinitionNode field in fields)
             {
                 // Create a tracker to keep track of visited entities to detect circular references
                 HashSet<string> trackerForField = new();
-                if (visitedEntity is not null)
+                if (visitedEntities is not null)
                 {
-                    trackerForField = visitedEntity;
+                    trackerForField = visitedEntities;
```
Aniruddh25, line 213: should be `trackerForFields`

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
Aniruddh25: it may not fail during operation if the entity - which has the circular reference but not in runtime Config is present as a child of another entity - present in the config.

src/Service.Tests/Configuration/ConfigurationTests.cs
Aniruddh25: cosmos related code is there but that is testing the configuration. Your test is checking schema - which is only specific to cosmosdb.

sourabh1007: I considered passing schema is part of "configuring" cosmos DB.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
sourabh1007: as discussed, used buildin function to check the type.

Aniruddh25 (approved):
LGTM, thanks for addressing all comments!

PR #2200, opened 2024-05-02
Fix issue with schema validation for config with env variables

### LINKED ISSUE #1979: [Bug]: Schema Validation is not considering environment variables
### What happened?

Currently when we do schema validation, it checks on the basis of given config, without replacing the environment variables.
For Example:
```
"data-source": {
      "database-type": "@env('DATABASE_TYPE')",
      "options": {
        "database": "@env('DATABASE_NAME')",
        "schema": "@env('GRAPHQL_SCHEMA_PATH')"
      },
      "connection-string": "@env('CONN_STRING')"
    }
```

Schema validation will give error stating `data-source.database-type is not an enum`. But if schema validation would be done considering the value after replacing env, it would show no error.

### Version

0.10.11-rc

Azure SQL

Local (including CLI)

_No response_

### Relevant log output

_No response_

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #1979 
  - the Json sent to schema validation was validated without replacing environment variables, due to which it started throwing error incase of type mismatch.

## What is this change?

- Added a method that takes in Json as string, and returns the new Json with replaced environment variables as string.

## NOTE:
- There were certain methods which was doing replacement, but it was written in a way to align with the RuntimeConfig Object. So, it couldn't be used. for example, different converterFactory classes were calling this StringJsonConvertorFactory and then converting it to required type based on the definition in the schema.
- Also, trying to deserialize it and then serialize it again changes the formatting. for example:
no by invalid i mean.
if i deserialize the below json 
```
{
"database-type": "mssql"
}
```

and then deserialize it
```
{
"database-type" : {
"DatabaseType": 2
}
}
```
- Also, we need the original conversion mechanism to stay since it will be utilized for the workflow outside of `dab validate`.
- Since, the new method takes string as input, it can be used in the future for the hosting scenario as the config file is provided as string.

## How was this tested?
- [X] Unit Tests

## Sample Request(s)
```
"data-source": {
      "database-type": "@env('DATABASE_TYPE')",
      "options": {
        "database": "@env('DATABASE_NAME')",
        "schema": "@env('GRAPHQL_SCHEMA_PATH')"
      },
      "connection-string": "@env('CONN_STRING')"
    }
```

### Before update
image

### After Update
image


src/Config/Converters/Utf8JsonReaderExtensions.cs
```diff
@@ -41,4 +42,115 @@ static internal class Utf8JsonReaderExtensions
 
         return JsonSerializer.Deserialize<string>(ref reader, options);
     }
+
+    /// <summary>
+    /// Replaces placeholders for environment variables in the given JSON string and returns the new JSON string.
+    /// </summary>
+    /// <param name="inputJson">The JSON string to process.</param>
+    /// <exception cref="JsonException">Thrown when the input is null or when an error occurs while processing the JSON.</exception>
+    public static string? ReplaceEnvVarsInJson(string? inputJson)
```
seantleonard, line 51: why do we need this function when `StringJsonConverterFactory()` already exists which has helpers to parse out envvironment variable placeholders? that class also uses regex to detect instances of `@env()`

why not something like:

```csharp
TryParseConfig(
                json: json,
                config: out config,
                connectionString: _connectionString,
                replaceEnvVar: true)
```

seantleonard: Is there something about `dab validate` that requires a completely different code path + new code?
if so , please add to pr description and code comments.

seantleonard (changes requested):
I don't follow why we need new environment variable parsing logic as opposed to using pre-existing code.

src/Config/Converters/Utf8JsonReaderExtensions.cs
seantleonard: please elaborate in PR description. this doesn't really tell us much about "why" it can't be used.

> There were certain methods which was doing replacement, but it was written in a way to align with the RuntimeConfig Object. So, it couldn't be used.

seantleonard: > Also, we need the original conversion mechanism to stay since it will be utilized for the workflow outside of dab validate.

Not clear why original conversion would need to change.

abhishekkumams:
> I don't follow why we need new environment variable parsing logic as opposed to using pre-existing code.

Because the original logic is written as part of Deserialization logic for the Runtime config, and here we don't need RuntimeConfig Object.

src/Config/Converters/Utf8JsonReaderExtensions.cs
```diff
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Text;
 using System.Text.Json;
 
 namespace Azure.DataApiBuilder.Config.Converters;
 
-static internal class Utf8JsonReaderExtensions
+static public class Utf8JsonReaderExtensions
```
Aniruddh25, line 9: why are we making public? To be used in tests? Cant we do internals visible to test project?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -193,7 +195,19 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
     /// </summary>
     public async Task<JsonSchemaValidationResult> ValidateConfigSchema(RuntimeConfig runtimeConfig, string configFilePath, ILoggerFactory loggerFactory)
     {
-        string jsonData = _fileSystem.File.ReadAllText(configFilePath);
+        string? jsonData = _fileSystem.File.ReadAllText(configFilePath);
+
+        // The config file may contain some environment variables that need to be replaced before validation.
+        try
+        {
+            jsonData = Utf8JsonReaderExtensions.ReplaceEnvVarsInJson(jsonData);
```
Aniruddh25, line 203: We already have the `runtimeConfig` deserialized from the `configFilePath`. 

Wouldnt that mean we already had the jsonData with the environment variables replaced? 
Or does the runtimeConfig object doesnt have the environment variable values? 
Instead of reading from the file path, cant we get the string from which the `runtimeConfig` object was formed - to do the schema validation on?

Aniruddh25 (changes requested):
Some questions on the approach

src/Core/Configurations/RuntimeConfigValidator.cs
abhishekkumams: no we cant. schemaValidator require a json string. i tried converting the deserialized runtimeConfig to string and then passing it to the schema validator but it was not forming a valid json string.

seantleonard: can you provide an example? invalid json string because of extra escape characters? is 'invalid json string' indicative of a different bug?

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -276,6 +278,120 @@ public void TestDictionaryDatabaseObjectSerializationDeserialization()
...
+            // Call the method under test and expect a JsonException
+            try
+            {
+                Utf8JsonReaderExtensions.ReplaceEnvVarsInJson(inputJson);
+                Assert.Fail("Expected a JsonException to be thrown");
+            }
+            catch (JsonException ex)
+            {
+                Assert.AreEqual("Failed to replace environment variables in given JSON. "
+                    + "Environment variable 'MISSING_ENV_VAR' is not set.", ex.Message);
```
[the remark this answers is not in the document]
abhishekkumams: well, actually not. environment files are read as test.  environment values are read as string,i.e. "null", and then added to the json as a null value. So, if a user wants to set a null value. it will be assigned as null and no error should be thorwn. there is a test for the same.

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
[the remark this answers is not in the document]
abhishekkumams: Okay, With the name of the file, it was not very evident that it will be used only for DW

src/Config/Converters/Utf8JsonReaderExtensions.cs
abhishekkumams: no, it's also being called up by RuntimeCongigValidator

src/Core/Configurations/RuntimeConfigValidator.cs
abhishekkumams: no by invalid i mean.
if i deserialize the below json 
```
{
"database-type": "mssql"
}
```

and then deserialize it
```
{
"database-type" : {
"DatabaseType": 2
}
}
```

seantleonard: that seems like another bug

abhishekkumams: it's just how enums are serialized i believe, don't think it's a bug

aaronburtle: > it's just how enums are serialized i believe, don't think it's a bug

I think we would need to have custom (de)serialization that handles the case of these enums so that they go back to the intended format. Deserialization/Serialization ought to be reversable so this does seem like a bug that just hasnt arisen yet.

src/Config/Converters/Utf8JsonReaderExtensions.cs
```diff
@@ -41,4 +42,115 @@ static internal class Utf8JsonReaderExtensions
...
+                {
+                    // Recursively process each item of the array
+                    ReplaceEnvVarsAndWrite(item, writer);
+                }
+
+                writer.WriteEndArray();
+                break;
+            case JsonValueKind.String:
+                string? value = element.GetString();
+                if (value is not null && value.StartsWith("@env('") && value.EndsWith("')"))
```
aaronburtle, line 110: Is this assuming that the start of the string is the beginning of the `env('<match>')`?

What happens if the `env('<match>')` happens in the middle of the string?

For example, if the person had 

{ 
  "database-type": "env('db_type')env('db_lang')" 
}

where their environment variable are db_type=ms, and db_lang=sql

will this replace `env('db_lang')` with `sql`?

aaronburtle: If you look at the JsonStringConverter the `StringJsonConverterFactory` there is logic that handles all of our edge cases with regex match and replace. If you can't use that class explicitly, perhaps you can factor out the logic that is being used for parsing the env variables to be shared?

src/Config/Converters/Utf8JsonReaderExtensions.cs
```diff
@@ -41,4 +42,115 @@ static internal class Utf8JsonReaderExtensions
...
+                    }
+                }
+                else
+                {
+                    writer.WriteStringValue(value);
+                }
+
+                break;
+            default:
+                // Write the JSON element to the writer as is
```
aaronburtle, line 151: So if the env('<match>') doesnt start the string we end up here correct, but then we don't do any replacements?

aaronburtle (commented):
Some questions about how and when we are matching the environment variables, and if we can share some code.

src/Core/Configurations/RuntimeConfigValidator.cs
seantleonard: using `string jsonData = runtimeConfig.ToJson()` instead of all the custom parsing code and then provided `jsonData` to:
`await jsonConfigSchemaValidator.ValidateJsonConfigWithSchemaAsync(jsonSchema, jsonData!);`

works great and doesn't present any of the issues mentioned earlier in this thread. Am I missing something?

Aniruddh25: do the tests added in this PR pass with your proposed changes? If yes, we may not need the custom parsing code mentioned here.

PR #2201, opened 2024-05-02
Ensure environment variables works for child files.

rohkhann:
## Why make this change?
When we are loading child config files, we were not setting replace env variable to true and hence child config environment variables were not being obeyed. This fix ensures that environment variables can be used for multi-db scenario.

## What is this change?
Adding true to runtimeconfig file.

## How was this tested?
Ran an integration test with two config files and an environment variable in the child file.

### DIFF
--- a/src/Config/ObjectModel/RuntimeConfig.cs
+++ b/src/Config/ObjectModel/RuntimeConfig.cs
@@ -196,7 +196,7 @@ public record RuntimeConfig
 
             foreach (string dataSourceFile in DataSourceFiles.SourceFiles)
             {
-                if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config))
+                if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config, true))
                 {
                     try
                     {
--- a/src/Config/ObjectModel/RuntimeConfig.cs
+++ b/src/Config/ObjectModel/RuntimeConfig.cs
@@ -196,7 +196,7 @@ public record RuntimeConfig
 
             foreach (string dataSourceFile in DataSourceFiles.SourceFiles)
             {
-                if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config, true))
+                if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config, replaceEnvVar: true))
                 {
                     try
                     {


src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -196,7 +196,7 @@ public RuntimeConfig(string? Schema, DataSource DataSource, RuntimeEntities Enti
 
             foreach (string dataSourceFile in DataSourceFiles.SourceFiles)
             {
-                if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config))
+                if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config, true))
```
seantleonard, line 199: named params

seantleonard:
> Ran an integration test with two config files and an environment variable in the child file.

can this be added as integration test?

src/Config/ObjectModel/RuntimeConfig.cs
rohkhann: fixed this.

rohkhann:
> > Ran an integration test with two config files and an environment variable in the child file.
> 
> can this be added as integration test?

Maybe bit tricky as we dont have any test structure in the pipelines right now that can take 2 config files. Given dab's ga deadline and we want this fix to go in, maybe better to invest in that testing all up for multi-db post GA.

ayush3797 (approved):

seantleonard:
PR description needs more details.

1. Can you add a minimum reproducible example parent `dab-config.json` and child `dab-child-config.json`? (perhaps these files are already in the test project, perhaps just sample `dab start` command? 
2. Since there isn't an open issue tracking this, can you also note what error messages appeared when env_vars were not processed in the child config files? 

Need to open issue to track adding a test: here are some notes for whoever picks it up:
1. Test for this use case. This should be straightforward in the `ConfigurationTests.cs` class:
  - Set test-specific environment variable `child-conf-connstring` and put a value
  - create parent runtime config with data=sources property and child runtime config and write to disk (there are examples of writing custom config to disk in ConfigurationTests.cs
  -  Start dab which implies that this constructor is called in RuntimeConfig.cs: `public RuntimeConfig(string? Schema, DataSource DataSource, RuntimeEntities Entities, RuntimeOptions? Runtime = null, DataSourceFiles? DataSourceFiles = null)`
  - Validate that whatever error message was encountered previous, is no longer experienced when not processing env vars in the child file.

seantleonard (approved):

Issue #2184, opened 2024-04-23
[Bug]: Validating entity relationships throws unhandled exception


Aniruddh25:
### What happened?

Cannot figure out what's wrong with this sample config. Doing `dab validate` causes an `Unhandled Exception` with the stack trace as provided in the log output.

```json
{
  "$schema": "https://dataapibuilder.azureedge.net/schemas/v0.5.34/dab.draft.schema.json",
  "data-source": {
    "database-type": "mssql",
    "options": {
      "set-session-context": false
    },
    "connection-string": "Server=XXXXX;Persist Security Info=False;User ID=<USERHERE>;Password=<PWD HERE> ;MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;"
  },
  "runtime": {
    "rest": {
      "enabled": true,
      "path": "/api"
    },
    "graphql": {
      "allow-introspection": true,
      "enabled": true,
      "path": "/graphql"
    },
    "host": {
      "mode": "development",
      "cors": {
        "origins": [],
        "allow-credentials": false
      },
      "authentication": {
        "provider": "StaticWebApps"
      }
    }
  },
  "entities": {
    "Publisher": {
      "source": {
        "object": "publishers",
        "type": "table",
        "key-fields": ["id"]
      },
      "graphql": {
        "enabled": true,
        "type": {
          "singular": "Publisher",
          "plural": "Publishers"
        }
      },
      "rest": {
        "enabled": true
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            {
              "action": "read"
            }
          ]
        },
        {
          "role": "authenticated",
          "actions": [
            {
              "action": "create"
            },
            {
              "action": "read"
            },
            {
              "action": "update"
            },
            {
              "action": "delete"
            }
          ]
        },
        {
          "role": "policy_tester_01",
          "actions": [
            {
              "action": "read",
              "fields": {
                "exclude": [],
                "include": [
                  "*"
                ]
              },
              "policy": {
                "database": "@item.id eq 1940"
              }
            },
            {
              "action": "update",
              "fields": {
                "exclude": [],
                "include": [
                  "*"
                ]
              }
            },
            {
              "action": "create"
            },
            {
              "action": "delete"
            }
          ]
        },
        {
          "role": "policy_tester_02",
          "actions": [
            {
              "action": "read",
              "fields": {
                "exclude": [],
                "include": [
                  "*"
                ]
              },
[...]

Issue #2180, opened 2024-04-22
[Bug]: New relationship validations causing unexpected exception being thrown


abhishekkumams:
### What happened?

we recently introduced some extra validations in our relationship validation workflow. `dab validate` fails on these new validations when database connection fails.

The issue is being caused because now some of the validations required DB metadatas, and when there is connection failure, it throws exception.

### Version

latest

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2202, opened 2024-05-03
Handle Exception during dab validate when SqlMetadataProvider has not correctly initialized

aaronburtle:
## Why make this change?
Closes https://github.com/Azure/data-api-builder/issues/2184
Closes https://github.com/Azure/data-api-builder/issues/2180

## What is this change?

When we don't properly initialize the `SqlMetadataProvider`, certain functions will not behave as intended. This change adds some defensive programming around this fact so that we have more readable exceptions in these cases, as well as restructuring the validation to avoid the potential of such a circumstance.

We refactor the `ValidateEntitiesMetadata` function into 2 functions, one called `ValidateRelationshipConfigCorrectness` that validates the relationships in the config without needing to cross reference database metadata, and leaving the remaining validations that do require such metadata within the original function. This validation is then handled by a mix of the initialization of the metadata provider (this init process does some validation), and the call to `ValidateRelationships` which happens from within `ValidateEntitiesMetadata` only when the initialization of the metadata provider resulted in no connection errors.

Therefore, the normal code path when doing `dab validate` will be

1. Call `ValidateRelationshipConfigCorrectness`
2. Call `ValidateEntitiesMetadata`
    2a. If there were no connection errors Call `ValidateRelationships`

image

## How was this tested?

Manually, against our test suite, and we add a regression test that looks for the correct error messaging when the config that caused the error from the finding of the bug is used.

## Sample Request(s)

This can be reprod by using the command `dab validate -c <path to config>


src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1075,6 +1057,67 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+        List<string> invalidColumns,
+        string[]? sourceFields,
+        string[]? targetFields,
+        string relationshipName,
+        string entityName,
+        string targetEntityName,
+        ISqlMetadataProvider sqlMetadataProvider)
+    {
+        // In all kinds of relationships, if sourceFields are included they must be valid columns in the backend.
+        if (sourceFields is not null)
```
seantleonard, line 1083: is there another place in validation code that checks whether the targetEntity is valid (metadata acquired too)? 
- perhaps you should do that here, there are a few datastructures that can give you this detail.
- Fail early with a precise error message that allows you to leave SqlMetadataProvider as-is to allow for refactoring later.

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -46,6 +46,27 @@ public void TestConfigWithCustomPropertyAsInvalid()
...
+    /// <summary>
+    /// This method validates that the correct error messaging occurs when a config
+    /// with an invalid connection string is used that has relationship fields defined.
+    /// This case is a regression test due to the metadata needed not always being
+    /// populated in the SqlMetadataProvider if for example a bad connection string
+    /// is given. The error messages occur then when functions that depend on this
+    /// metadata are called.
+    /// </summary>
+    [TestMethod]
+    public void TestErrorMessageFromConfigWithNonWorkingConnectionString()
```
seantleonard, line 58: Function name is too vague to represent the specific issue of validating relationship metadata when database metadata can't be fetched.

For the summary, do you mean: "This test ensures that relationship validation doesn't fail due to unhandled exceptions thrown as side effects of missing database metadata. A specific error message is expected which informs the user which entity/relationship caused validation to fail."?

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -46,6 +46,27 @@ public void TestConfigWithCustomPropertyAsInvalid()
...
+    {
+        ((MockFileSystem)_fileSystem!).AddFile(TEST_RUNTIME_CONFIG_FILE, CONFIG_WITH_NON_WORKING_CONN_STRING);
+        ValidateOptions validateOptions = new(TEST_RUNTIME_CONFIG_FILE);
+        StringWriter writer = new();
+        // Capture console output to get error messaging.
+        Console.SetOut(writer);
+        ConfigGenerator.IsConfigValid(validateOptions, _runtimeConfigLoader!, _fileSystem!);
+        string errorMessage = writer.ToString();
+        Assert.IsTrue(errorMessage.Contains("Unable to validate source and target fields as backing columns in the DB due to"));
```
seantleonard, line 67: nit: split into "arrange act assert"

seantleonard (changes requested):

src/Cli.Tests/TestHelper.cs
```diff
@@ -1065,6 +1065,105 @@ public static Process ExecuteDabCommand(string command, string flags)
         }
       }";
 
+        public const string CONFIG_WITH_NON_WORKING_CONN_STRING = @"
```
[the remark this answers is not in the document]
aaronburtle: I looked at constructing from the already created constants, but they almost all exclusively have entities already that are either empty or simply dont have a relationship section. So we would only be able to reuse a small portion of these in creating the config we want. Seemed most natural to simply create this constant on its own that has what we need and then also directly matches the config that caused the bug to be found. I could replace the data source section alone with already created constants, but thats a small portion of the config we need.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -873,42 +873,24 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
-                            message: $"Entity: {entityName} has a relationship: {relationshipName} with source fields: {string.Join(",", invalidColumns)} that " +
-                                $"do not exist as columns in entity: {entityName}.",
-                            statusCode: HttpStatusCode.ServiceUnavailable,
-                            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
-                    }
                 }
-
-                // In all kinds of relationships, if targetFields are included they must be valid columns in the backend.
-                if (relationship.TargetFields is not null)
+                catch (KeyNotFoundException ex)
```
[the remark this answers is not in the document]
aaronburtle: The `DataApiBuilderException` is already handled by the internal calls that can throw them. KeyNotFound comes from an outside class that doesn't have this same logic for handling the exception which is why we catch it here.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -873,42 +873,24 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
-                    if (invalidColumns.Count > 0)
-                    {
-                        HandleOrRecordException(new DataApiBuilderException(
-                            message: $"Entity: {entityName} has a relationship: {relationshipName} with target fields: {string.Join(",", invalidColumns)} that " +
-                                $"do not exist as columns in entity: {relationship.TargetEntity}.",
-                            statusCode: HttpStatusCode.ServiceUnavailable,
-                            subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
-                    }
+                    HandleOrRecordException(new DataApiBuilderException(
+                        message: $"Unable to validate source and target fields as backing columns in the DB due to: {ex.Message} \n" +
```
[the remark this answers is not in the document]
aaronburtle: We talked about this offline, but just to give context here, we will use a custom message for the key not found so that it is useful to the end user, but remove the connection string messaging because it is redundant to other exceptions that will already communicate this information in our exception/error handling.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: We do indeed have validation code for checking the metadata (which in this case is stored as a DatabaseObject) on both the source and target entities in the outer `ValidateRelationshipsInConfig` function.

seantleonard (approved):
few nits, but looks good!

seantleonard (approved):

src/Cli.Tests/TestHelper.cs
```diff
@@ -1065,6 +1065,104 @@ public static Process ExecuteDabCommand(string command, string flags)
         }
       }";
 
+        public const string COMPLETE_CONFIG_WITH_RELATIONSHIPS_NON_WORKING_CONN_STRING = @"
+        {
+  ""$schema"": ""https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json"",
```
Aniruddh25, line 1070: are there other const strings in this test code that this const can build up on?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1075,6 +1054,63 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
         }
     }
 
+    /// <summary>
+    /// Handles the validation of source and target fields as valid backing columns in the DB.
+    /// We put the logic for both source and target fields into a single function here for readability
+    /// when catching exceptions in the caller related to being unable to lookup the backing columns.
+    /// This happens when validation is called with a mal formed connection string.
```
Aniruddh25, line 1061: Why cant we skip such a validation when we know its a malformed connection string? 
`if (!ConfigValidationExceptions.Any(x => x.Message.StartsWith(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE)))` => cant we do this check?

Aniruddh25 (changes requested):
Why do we validate backing columns if we know connection string is incorrect?

src/Cli.Tests/TestHelper.cs
aaronburtle: we'd have to modify them because the entities are not formed propery, with the exception of the datasource section

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: This can work, but i will need to change the validation we use for the connection string. Currently, that list of errors only contains the CONNECTION_STRING_ERROR_MESSAGE when the connection string is empty.


image


So instead we will need better validation logic to throw those errors for any malformed connection string. I am adding that logic now.

aaronburtle: Yes, this should work. This check happens before we do any meta data validation, but it turns out that the DataSource validation that we do is only checking if the connection string is empty, not if it is invalid in any way.

image


But then later we add the exceptions from the meta data provider, and these include errors that are done by checking the validity of the actual connection to the backend DB, so we can use this check later on inside of the validation

image


image

aaronburtle: To give a little more context on how this will ultimately be handled (new changes pending). The ValidateEntitiesMetadata function is validating entities in ways that both do and do not require metadata surrounding those entities from the backend. Rather than having to separately check for metadata initialization errors for each of this checks individually, we are going to split this function into 2 separate functions. One that doesn't rely on metadata from the backend, and one that does. We will then use the conditional above to make sure we dont attempt any checks that require this metadata in the case where we have those initialization errors that prevented that data from populating in the metadata provider, while still continuing with the function that validates information that doesnt depend on that meta data.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -209,9 +207,117 @@ public async Task<JsonSchemaValidationResult> ValidateConfigSchema(RuntimeConfig
...
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// This method validates the entities relationships against the database objects using
+    /// metadata from the backend DB generated by this function.
+    /// </summary>
+    public async Task ValidateEntityRelationshipsInConfigWithGeneratedMetadata(RuntimeConfig runtimeConfig, ILoggerFactory loggerFactory)
```
seantleonard, line 320:
```suggestion
    public async Task ValidateEntityRelationshipsInConfigWithDbMetadataAsync(RuntimeConfig runtimeConfig, ILoggerFactory loggerFactory)
```
name should be consistent with other validation "With/Without + DbMetadata" and also have "Async Suffix because it's an async method.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -225,11 +331,15 @@ public async Task ValidateEntitiesMetadata(RuntimeConfig runtimeConfig, ILoggerF
...
-
         await metadataProviderFactory.InitializeAsync();
+
         ConfigValidationExceptions.AddRange(metadataProviderFactory.GetAllMetadataExceptions());
 
-        ValidateRelationshipsInConfig(runtimeConfig, metadataProviderFactory);
+        // Validation below relies on metadata being populated from backend, only execute when there were no connection errors
+        if (!ConfigValidationExceptions.Any(x => x.Message.StartsWith(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE)))
+        {
+            ValidateEntityRelationshipsInConfigWithProvidedMetadata(runtimeConfig, metadataProviderFactory);
```
seantleonard, line 341: The method names seems redundant:
- ValidateEntityRelationshipsInConfigWithGeneratedMetadata
- ValidateEntityRelationshipsInConfigWithProvidedMetadata
- ValidateEntityRelationshipsInConfigWithoutDBMetadata

Some of the methods say "Metadata" while one says "DbMetadata" and then you also have "Generated" and "Provided" metadata. Naming needs to be updated with comments to clarify what the difference is. The first two seem to be the same.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -162,12 +162,10 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
...
 
-        // If the ConfigValidationExceptions list doesn't contain a DataApiBuilderException with connection string error message,
-        // then only we run the metadata validation.
-        if (!ConfigValidationExceptions.Any(x => x.Message.Equals(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE)))
-        {
-            await ValidateEntitiesMetadata(runtimeConfig, loggerFactory);
-        }
+        _logger.LogInformation("Validating entity relationships.");
+        ValidateEntityRelationshipsInConfigWithoutDBMetadata(runtimeConfig);
+        await ValidateEntityRelationshipsInConfigWithGeneratedMetadata(runtimeConfig, loggerFactory);
```
seantleonard, line 167: Ideas for shorter and more explicit naming: 
- `ValidateEntityRelationshipsWithDatabaseMetadata`
- `ValidateRelationshipsWithDatabaseMetadata`
- `ValidateRelationshipConfigWithDatabase`

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -162,12 +162,10 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
         ValidateConfigProperties();
         ValidatePermissionsInConfig(runtimeConfig);
 
-        // If the ConfigValidationExceptions list doesn't contain a DataApiBuilderException with connection string error message,
-        // then only we run the metadata validation.
-        if (!ConfigValidationExceptions.Any(x => x.Message.Equals(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE)))
-        {
-            await ValidateEntitiesMetadata(runtimeConfig, loggerFactory);
-        }
+        _logger.LogInformation("Validating entity relationships.");
```
seantleonard, line 165: can this logger be in the relationship validation function?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -209,9 +207,117 @@ public async Task<JsonSchemaValidationResult> ValidateConfigSchema(RuntimeConfig
     }
 
     /// <summary>
-    /// This method runs validates the entities metadata against the database objects.
+    /// Validates the semantic correctness of an Entity's relationships in the
+    /// runtime configuration when metadata from the backend DB is not required.
+    /// Validating Cases:
```
seantleonard, line 212:
```suggestion
    /// runtime configuration without crossreferencing database metadata.
    /// Validating Cases:
```

"not required" is ambiguous. It could imply:
- when database metadata is available it could be used
- user can supply database metadata somehow to alter how this function validates config.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -209,9 +207,117 @@ public async Task<JsonSchemaValidationResult> ValidateConfigSchema(RuntimeConfig
...
+                        entityName: entityName,
+                        relationshipName: relationshipName);
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// This method validates the entities relationships against the database objects using
+    /// metadata from the backend DB generated by this function.
```
seantleonard, line 318: what do you mean by "from the backend DB generated by this function." ?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -162,12 +162,10 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
         ValidateConfigProperties();
         ValidatePermissionsInConfig(runtimeConfig);
 
-        // If the ConfigValidationExceptions list doesn't contain a DataApiBuilderException with connection string error message,
-        // then only we run the metadata validation.
-        if (!ConfigValidationExceptions.Any(x => x.Message.Equals(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE)))
-        {
-            await ValidateEntitiesMetadata(runtimeConfig, loggerFactory);
```
seantleonard, line 169: why are you no longer calling `ValidateEntitiesMetadata`? aren't there additional validations there that aren't covered by your entity relationship validation?

src/Core/Configurations/RuntimeConfigValidator.cs
seantleonard: Also, this method doesn't just validate entity relationships. there are validations in the metadataprovider.initialize that aren't related to relationships

seantleonard (changes requested):
some additional changes needed.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: That was the intention of the summaries

* ValidateEntityRelationshipsInConfigWithGeneratedMetadata
    /// This method validates the entities relationships against the database objects using
    /// metadata from the backend DB generated by this function.
    
* ValidateEntityRelationshipsInConfigWithProvidedMetadata
    /// Validates the semantic correctness of an Entity's relationships in the
    /// runtime configuration against the metadata collected from the backend DB
    /// that is provided to this function via use of the MetadataProviderFactory

* ValidateEntityRelationshipsInConfigWithoutDBMetadata
    /// Validates the semantic correctness of an Entity's relationships in the
    /// runtime configuration when metadata from the backend DB is not required.


To summarize, the "generated" function will generate the metadata by creating a meta data factory and then initializing a metadata provider. This will go to the DB and populate the provider with metadata, that provider is then used to call the provided function.

the reason we have the matadata-generated function calling the metadata-provided function is that we already have code that is trying to validate using their own meta data provider, so we should keep that as an option.

I can add more details if you think its worth it.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: there are 2 relationship validation functions now, so if we don't want this message to potentially appear twice, probably better to just put it before both calls right?

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: the function will create its own metadata provider and generate the metadata during the metadata providers initalization. then that metadata is used to call the provided function. so this means that the metadata we use will be generated by this function at the time of calling.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: I am not sure if its because the intention of that function changed at some point, but this is what it looks like before my changes.

image


As you can see it really is just validating the entity relationships with the ValidateRelationshipInConfig call.

So to more easily understand, and I will add this to the PR description, what is going on here, you can think of `ValidateEntitiesMetadata` as being replaced by the `ValidateEntityRelationshipsInConfigWithGeneratedMetadata` which has an internal call to the replacement for `ValidateRelationshipsInConfig` which is now the `ValidateEntityRelationshipsInConfigWithProvidedMetadata`, and then we also add separately to these functions `ValidateEntityRelationshipsInConfigWithoutDBMetadata` which runs on its own to do the validation that doesnt need a metadataprovider. So in the case that we have DB errors trying to connect, we only run the function that has no need for a metadata provider, and then in the other cases we still call the same equivalent functions they just have different names now that more closely describe their behavior and responsibilities.

src/Core/Configurations/RuntimeConfigValidator.cs
seantleonard: yeah this should be fine then.

 If you did put a logging message in both validation functions, they'd need to be more specific about what they're validating. e.g. validating required relationship config values are present (no mismatching counts of source/target fields etc) and then validating relationship config aligns with database schema. (backing column validation).

src/Core/Configurations/RuntimeConfigValidator.cs
seantleonard: It's a bit confusing with those names being so verbose: `ValidateEntityRelationshipsInConfig...` where both "generated" and "provided" are public methods.
- (public) `ValidateEntityRelationshipsInConfigWithoutDBMetadata()`
- (public) `ValidateEntityRelationshipsInConfigWithGeneratedMetadataAsync()`
  - (public*) `ValidateEntityRelationshipsInConfigWithProvidedMetadata()`
    - (private) `ValidateSourceAndTargetFieldsAsBackingColumns()`

* Could all functions listed in sub-bullets of `ValidateEntityRelationshipsInConfigWithGeneratedMetadataAsync()` be private? The top level validation function can be generic: ValidateEntityRelationshipsAgainstDb()
and then whatever helper functions get called internally to that function can be named to more specifically denote what they're validating.

src/Core/Configurations/RuntimeConfigValidator.cs
seantleonard: `await metadataProviderFactory.InitializeAsync()` initializes a metadataprovider, and the metadataprovider initialization executes validations. The other validations are abstracted away by the Init() but they are happening.

ValidationRelationshipsInConfig is just another validation that occurs.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: We synced offline, but just to give context here. Since the metadata provider initialization is handling validation, we have the outer call which handles validation of entity relationships also handling validation of the entities, and is named `ValidateEntitiesMetadata`, this function then calls into a function that only validated the relationship information that requires metadata to proceed, `ValidateRelationships`, and the relationship validation that does not need metadata happens on its own and is named `ValidateRelationshipConfigCorrectness`.

`ValidateEntitiesMetadata` is always called during dab validate, as is `ValidateRelationshipConfigCorrectness`, however `ValidateRelationships` is only called from withing `ValidateEntitiesMetadata` when there are no connection errors during the initialization of the metadata provider.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -209,7 +206,115 @@ public async Task<JsonSchemaValidationResult> ValidateConfigSchema(RuntimeConfig
...
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+            }
+
+            string databaseName = runtimeConfig.GetDataSourceNameFromEntityName(entityName);
+
+            foreach ((string relationshipName, EntityRelationship relationship) in entity.Relationships!)
+            {
+                // Validate if entity referenced in relationship is defined in the config.
+                if (!runtimeConfig.Entities.ContainsKey(relationship.TargetEntity))
```
ayush3797, line 242: Should use TryGetValue here since we anyways fetch the value later at L251.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -810,128 +916,17 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
-                        fieldType: "target",
-                        entityName: entityName,
-                        relationshipName: relationshipName);
-                }
+                ValidateSourceAndTargetFieldsAsBackingColumns(
+                    entityName: entityName,
+                    relationshipName: relationshipName,
+                    relationship: relationship,
+                    sqlMetadataProvider: sqlMetadataProvider,
+                    invalidColumns: invalidColumns);
```
ayush3797, line 929: Need not have `invalidColumns` here. Just referred to in the downstream.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1075,6 +1070,60 @@ public void ValidateRelationshipsInConfig(RuntimeConfig runtimeConfig, IMetadata
...
+        string entityName,
+        string relationshipName,
+        EntityRelationship relationship,
+        ISqlMetadataProvider sqlMetadataProvider,
+        List<string> invalidColumns)
+    {
+        // In all kinds of relationships, if sourceFields are included they must be valid columns in the backend.
+        if (relationship.SourceFields is not null)
+        {
+            GetFieldsNotBackedByColumnsInDB(
```
ayush3797, line 1091: As a side effect of this method, you are populating the invalidColumns list which is not evident by the function name `GetFieldsNotBackedByColumnsInDB` which suggests getting back list of invalid fields. I think returning a list from this method call would be a better option.

ayush3797:
Might not need this `clear`: https://github.com/Azure/data-api-builder/blob/e895f356f2105132b9ea2cc6e3e396555f09b632/src/Core/Configurations/RuntimeConfigValidator.cs#L1143 based on the resolution of previous comments

ayush3797 (approved):
Some suggestions. LGTM otherwise! Thanks for fixing the dab validate command!

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: we dont want to create and return a new list because we dont have any guarantees about garbage collection, so we need to pass a reference to the list we created before the loop.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: I should also mention, this code is from a previous PR, but it has been moved as part of this PR.

seantleonard (approved):

PR #2205, opened 2024-05-08
Fix support for integer values in  float/decimal type columns

### LINKED ISSUE #1629: Error when receiving Decimal value without a decimal present
### What happened?

When trying to update a Numeric or Decimal column in an Azure SQL database using GraphQL, the update will work **only if a decimal value is present**.  If an integer is present, e.g. 123, I receive the following error from the server:

"Unable to cast object of type 'HotChocolate.Language.IntValueNode' to type 'HotChocolate.Language.FloatValueNode'."

If I change that updated value to 123.45, the update will go through successfully.  This behavior is consistent across both create and update operations (both with the field being the only one included, or other fields being updated along with it).

[Error.txt](https://github.com/Azure/data-api-builder/files/12324905/Error.txt)
[Success.txt](https://github.com/Azure/data-api-builder/files/12324906/Success.txt)

The problem that I am running into is that I am using the URQL client to access the DAB GraphQL server, and it does not include the decimal portion of the number if it is zero (123.0 is sent as 123, resulting in the error above).

I would expect DAB to accept an integer value for a Decimal field.

### Version

Microsoft.DataApiBuilder 0.7.5+2f938a6dfe4b90c26d9e24352b38280cb6e66f15

Azure SQL

Static Web Apps (SWA)

GraphQL

### Relevant log output

```Text
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 POST http://localhost:5000/graphql application/json 436
info: Microsoft.AspNetCore.Cors.Infrastructure.CorsService[4]
      CORS policy execution successful.
dbug: Azure.DataApiBuilder.Service.AuthenticationHelpers.ClientRoleHeaderAuthenticationMiddleware[0]       
      657e14e3-16dd-4c65-bbfb-ed63ad19d1de: Request authentication state: Anonymous.
dbug: Azure.DataApiBuilder.Service.AuthenticationHelpers.ClientRoleHeaderAuthenticationMiddleware[0]       
      657e14e3-16dd-4c65-bbfb-ed63ad19d1de: The request will be executed in the context of Anonymous role  
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'Hot Chocolate GraphQL Pipeline'
fail: Azure.DataApiBuilder.Service.Startup[0]
      Unable to cast object of type 'HotChocolate.Language.IntValueNode' to type 'HotChocolate.Language.FloatValueNode'.
fail: Azure.DataApiBuilder.Service.Startup[0]
         at Azure.DataApiBuilder.Service.Services.ResolverMiddleware.ExtractValueFromIValueNode(IValueNode 
value, IInputField argumentSchema, IVariableValueCollection variables)
         at Azure.DataApiBuilder.Service.Resolvers.BaseSqlQueryStructure.GQLMutArgumentToDictParams(IMiddlewareContext context, String fieldName, IDictionary`2 mutationParameters)
         at Azure.DataApiBuilder.Service.Resolvers.SqlUpdateStructure..ctor(IMiddlewareContext context, String entityName, ISqlMetadataProvider sqlMetadataProvider, IAuthorizationResolver authorizationResolver, GQLFilterParser gQLFilterParser, IDictionary`2 mutationParams, HttpContext httpContext)
         at Azure.DataApiBuilder.Service.Resolvers.SqlMutationEngine.PerformMutationOperation(String entityName, Operation operationType, IDictionary`2 parameters, IMiddlewareContext context)
         at Azure.DataApiBuilder.Service.Resolvers.SqlMutationEngine.ExecuteAsync(IMiddlewareContext context, IDictionary`2 parameters)
         at Azure.DataApiBuilder.Service.Services.ResolverMiddleware.InvokeAsync(IMiddlewareContext context)
         at HotChocolate.Utilities.MiddlewareCompiler`1.ExpressionHelper.AwaitTaskHelper(Task task)        
         at HotChocolate.Execution.Processing.Tasks.ResolverTask.ExecuteResolverPipelineAsync(CancellationToken cancellationToken)
         at HotChocolate.Execution.Processing.Tasks.ResolverTask.TryExecuteAsync(CancellationToken cancellationToken)
      Request finished HTTP/1.1 POST http://localhost:5000/graphql application/json 436 - 200 - application/json;+charset=utf-8 1614.2282ms
```

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #1629
  - Unable to cast object of type 'HotChocolate.Language.IntValueNode' to type 'HotChocolate.Language.FloatValueNode', when using Integer values for float/decimal/single column types

## What is this change?

- using `float.Parse()` and `decimal.Parse()` which automatically takes care of int values.

## How was this tested?

- Added new test data points in the existing tests for insertion/updation of float/decimal/single type columns with Integer values

### Before:
image

image

### After:
image

image


src/Core/Services/ExecutionHelper.cs
```diff
@@ -319,8 +321,8 @@ private static void SetContextResult(IMiddlewareContext context, JsonDocument? r
                 SupportedHotChocolateTypes.INT_TYPE => ((IntValueNode)value).ToInt32(),
                 SupportedHotChocolateTypes.LONG_TYPE => ((IntValueNode)value).ToInt64(),
                 SupportedHotChocolateTypes.SINGLE_TYPE => ((FloatValueNode)value).ToSingle(),
-                SupportedHotChocolateTypes.FLOAT_TYPE => ((FloatValueNode)value).ToDouble(),
-                SupportedHotChocolateTypes.DECIMAL_TYPE => ((FloatValueNode)value).ToDecimal(),
+                SupportedHotChocolateTypes.FLOAT_TYPE => value is IntValueNode intValueNode ? intValueNode.ToDouble() : ((FloatValueNode)value).ToDouble(),
+                SupportedHotChocolateTypes.DECIMAL_TYPE => value is IntValueNode intValueNode ? intValueNode.ToDecimal() : ((FloatValueNode)value).ToDecimal(),
```
seantleonard, line 325: How does `value` come to be of type `IntValueNode`:
- Does DAB attempt to cast variables to certain types?
- Is the conversion done by HotChocolate during request processing prior to DAB code?

Depending on the answer to the above question: Is this the right place for a fix or should a fix be added earlier in request processing? The fact that `value`'s type doesn't match `argumentSchema.Type.TypeName().Value` seems weird to me.

seantleonard (changes requested):
Before merging, would like additional perspective from you about your approach.

src/Core/Services/ExecutionHelper.cs
abhishekkumams: HotChocolate does this conversion giving the type of input value and then at this place we try to do the casting based on the column type.

This is the earliest place where we check and caste the input values from user.

ayush3797: Why put this check? If we refer here: https://chillicream.com/docs/hotchocolate/v13/defining-a-schema/scalars#net-scalars, it states that hotchocolate decimal type is equivalent to .net float type. So instead, we just use methods like float.parse() or double.parse() ? (Reference: https://learn.microsoft.com/en-us/dotnet/api/system.double.parse?view=net-8.0)

abhishekkumams: I didn't understand how we would use that here.  the problem was hotchocolate thinking of the value provided by user as IntNodeType, and we were incorrectly casting it before.

Aniruddh25, line 325: This will fix the issue if value is considered IntValueNode, but what if it is another valid float number yet, of a different value type? Would `((FloatValueNode)value)` cast fail again?

src/Core/Services/ExecutionHelper.cs
```diff
@@ -319,8 +321,8 @@ private static void SetContextResult(IMiddlewareContext context, JsonDocument? r
...
                 SupportedHotChocolateTypes.UUID_TYPE => Guid.TryParse(value.Value!.ToString(), out Guid guidValue) ? guidValue : value.Value,
```
Aniruddh25, line 326: Can we do `Float.Parse(value.Value)` so that all kinds of valid float values get parsed?

src/Core/Services/ExecutionHelper.cs
ayush3797: If we use the parsing functions, we don't need to worry about the casting done by hotchocolate

ayush3797:
Can you try with SINGLE_TYPE as well? Supply an int value for single types and see what happens?

ayush3797:
And add tests for all
Relevant types as well.

src/Core/Services/ExecutionHelper.cs
abhishekkumams: Thanks, Ani, I tried the suggested change but the .Net float type has different range than Hotchocolate floatNodeType, which is breaking existing tests.
The float type in .NetC# has a maximum value of approximately 3.4E38. Any value larger than this will exceed the range of float, and when we try to parse such a value, C# will return Infinity, and we have a test for 2E150 which started failing.
float.Parse() for any larger value gives `Infinity`
image

abhishekkumams: Also noticed other tests failing like below:
image

src/Core/Services/ExecutionHelper.cs
abhishekkumams: Thanks, Ani, I tried the suggested change but the .Net float type has different range than Hotchocolate floatNodeType, which is breaking existing tests.
The float type in .NetC# has a maximum value of approximately 3.4E38. Any value larger than this will exceed the range of float, and when we try to parse such a value, C# will return Infinity, and we have a test for 2E150 which started failing.
float.Parse() for any larger value gives `Infinity`
image

abhishekkumams: Also noticed other tests failing like below:
image

src/Core/Services/ExecutionHelper.cs
abhishekkumams: what are the other possibilities? i tried providing string and it gave error in hotchocolate query itself.

Aniruddh25: how about boolean? long? short? decimal which can be parsed as float?

aaronburtle: I have been playing around with this today, and one option that might handle all of these cases gracefully is to have helper parsing functions for each number type that we support that needs careful conversion.

so something like

return argumentSchema.Type.TypeName().Value switch
{
    SupportedHotChocolateTypes.BYTE_TYPE => ((IntValueNode)value).ToByte(),
    SupportedHotChocolateTypes.SHORT_TYPE => ((IntValueNode)value).ToInt16(),
    SupportedHotChocolateTypes.INT_TYPE => ((IntValueNode)value).ToInt32(),
    SupportedHotChocolateTypes.LONG_TYPE => ((IntValueNode)value).ToInt64(),
    SupportedHotChocolateTypes.SINGLE_TYPE => TryParseToSingle(value),
    SupportedHotChocolateTypes.FLOAT_TYPE => TryParseToDouble(value),
    SupportedHotChocolateTypes.DECIMAL_TYPE => TryParseToDecimal(value),
    SupportedHotChocolateTypes.UUID_TYPE => Guid.TryParse(value.Value!.ToString(), out Guid guidValue) ? guidValue : value.Value,
    _ => value.Value
};

and then the `TryParseTo` helper functions can take in an `IValueNode` and based on its type, do the relevant conversion to pass back to the caller and be cast to the type we want.

seantleonard: HotChocolate validation occurs before this code that ensures that the user-provided input object field value value is coercible to the type defined on the GraphQL schema. 

Using the `int_types` (integer) field as an example if the user provides:
- `"this is a string"` or `"123"` (string)
- `123.45` (decimal/float)
- `2147483648` (too large a number -> Int.MAX+1)
- `true` (boolean)

-> both BCP and the GraphQL engine will respond with
  >  "message": "The specified value type of field `int_types` does not match the field type."

seantleonard: Looks like HC has only two nodes for "numbers":
- `IntValueNode` https://github.com/ChilliCream/graphql-platform/blob/12.22.4/src/HotChocolate/Language/src/Language.SyntaxTree/IntValueNode.cs
- `FloatValueNode` https://github.com/ChilliCream/graphql-platform/blob/12.22.4/src/HotChocolate/Language/src/Language.SyntaxTree/FloatValueNode.cs

This PR should be scoped to coercing userinput to the type defined for the field on the GraphQL schema. e.g when HotChocolate resolves userinput to `IntValueNode`: try to coerce that value to the appropriate `float` or `decimal` type.

I don't think we should be trying to convert a string "123.23" to a float/decimal in this PR.

abhishekkumams: we shouldn't convert string to float/decimal, hotchocolate has its own validation based on schema generated which provide suggestions and give error when creating queries itself if we try to use different type then the target type.
Also
decimal and float are same as float in graphQL Node type
byte/short/int/long all are IntNodeType
giving boolean value will throw error for float/decimal.

seantleonard (approved):

Aniruddh25 (approved):
LGTM. Thank you for addressing all concerns.

Issue #2206, opened 2024-05-10
npgsql 7.0.7 for .net6


seantleonard:
Set npgsql in `directory.Packages.props` to 7.0.7 for .net6

PR #2207, opened 2024-05-10
Update Npgsql

seantleonard:
## Why make this change?

- `dotnet restore` (via `dotnet format` implicitly) raises NU1903 warning as an error because older versions of npgsql are tagged as vulnerable due to a high severity vulnerability.

## What is this change?
- Updates build-pipeline yaml to include `-v d` for detailed verbose logging in dotnet restore step so actual error messages are shown instead of generic failure.
- Sets separate versions of npgsql for .net6 and .net8:
  - .NET6 -> Npgsql 7.0.7 (set to 8.0.3, until nuget no longer reports 7.0.7 as vulnerable) tracked via https://github.com/Azure/data-api-builder/issues/2206
  - .NET8 -> Npgsql 8.0.3 
- ignores `TestDictionaryDatabaseObjectSerializationDeserialization()` for .net6 until #2206 is addressed because npgsql 8.0.3 depends on system.text.json from .net8 which breaks behavior in the test.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -33,7 +33,6 @@
     <PackageVersion Include="MSTest.TestFramework" Version="3.3.1" />
     <PackageVersion Include="MySqlConnector" Version="2.1.5" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.2" />
-    <PackageVersion Include="Npgsql" Version="7.0.1" />
     <PackageVersion Include="Polly" Version="7.2.3" />
     <PackageVersion Include="NJsonSchema" Version="10.9.0" />
     <PackageVersion Include="StyleCop.Analyzers" Version="1.1.118" />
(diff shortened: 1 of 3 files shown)


Aniruddh25 (approved):
LGTM!

src/Directory.Packages.props
```diff
@@ -64,5 +64,7 @@
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="6.0.29" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="6.0.0" />
+    <!--Once dotnet restore properly resolves that 'Npgsql' 7.0.7 is not vulnerable, set this ref to 7.0.7.-->
```
aaronburtle, line 67: Why do we want to revert back to 7.0.7 after this resolution instead of just keeping it at 8.0.3?

seantleonard: this update utilizes JsonSerialization changes from .NET8 and causes SerializeDeserialize tests to fail in .NET6.

aaronburtle (approved):
Looks good I just had a question about the versions.

PR #2211, opened 2024-05-13
CosmosDB : Schema Generation

sourabh1007:
## Why make this change?
Today customer or developer has to feed-in schema information in the form of `schema.gql` file while setting up DAB.

## What is this change?
As part of this feature, Customer would have option to let DAB generates the schema on startup.

So it means:
1. Customer can pass `schema.gql` file, as they pass today
image

2. If `schema.gql` file is not passed, DAB will generate the same schema and load it in during startup itself.
   a) By default, it will take the `top 1` record from the container.
   b) It would be configurable in config file, where developer can mention the sample size _(min: 10, max: 20)_.
   c) Developer can write a custom Cosmos DB query which will be used to generate the schema.
image

### What if both configs are there?
user passed schema i.e. `schema.gql` file, will get more preference than generating schema.

### High Level Design

```mermaid


flowchart TD
    Engine[fa:fa-cogs DAB Engine]-->| when initialized | CheckConfig{fa:fa-user Is user provided graphql.gql schema?}
    CheckConfig -->|yes| LoadConfig
    CheckConfig -->|no| GenerateConfig:::bar
    subgraph GenerateConfig[Schema Generation]
       b[fa:fa-sitemap Read Schema Analyzer Config from db-config.json] --> userconfigured{fa:fa-user is user configured?}
       userconfigured -->|yes| LoadFromConfigFile[fa:fa-file-alt Load config from db-config.json]
       userconfigured -->|No| LoadDefaultConfig[fa:fa-file-alt Load default config i.e. select top 10 data]
       LoadDefaultConfig --> c
       LoadFromConfigFile --> c[fa:fa-file-excel Query Cosmos DB get Sample data]
       c --> d[fa:fa-tools Schema Generator]
       g[fa:fa-database Database configured in db-config.json] --> c
    end
    subgraph LoadConfig[fa:fa-upload Load GraphQL.gql]
       f[fa:fa-file Read schema file]
    end
    d -->|Returns GQL schema in string format| GenerateSchema
    LoadConfig -->|Returns GQL schema in string format| GenerateSchema[fa:fa-spinner Generate and Cache Graphql schema]
    GenerateSchema --> MemoryData[fa:fa-file In-Memory Store]
    classDef bar stroke:#0f0

```

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)


JerryNixon:
Would this auto-populate the `mapping` property for the entity and allow for field aliases?

sourabh1007:
Closing this, as approach changed.

PR #2228, opened 2024-05-21
Streaming results in query flow to protect application memory

rohkhann:
## Why make this change?
1. Without adding limits on the amount of data that can be read from the db, the current reads could overload the memory of the hosting infrastructure of data-api-builder. Because of this, we need to have a limit on the amount of data that can be pulled from memory.

## What is this change?
1. This change uses Sequential access and buffers to buffer data in from the db. 
2. It throws exception if - there is more data available to read from the db for this particular query but we have already read the max buffer size available.

## How was this tested?

- Integration Tests
All existing integration tests execute successfully ensuring new code is only executed if feature flag is enabled.
- Unit Tests
added unit test for streaming logic.

### DIFF
--- a/src/Config/Converters/DataSourceConverterFactory.cs
+++ b/src/Config/Converters/DataSourceConverterFactory.cs
@@ -103,14 +103,6 @@ internal class DataSourceConverterFactory : JsonConverterFactory
                                         {
                                             optionsSubpropertyValue = null;
                                         }
-                                        else if (reader.TokenType is JsonTokenType.Number)
-                                        {
-                                            optionsSubpropertyValue = reader.GetInt32();
-                                            if (optionsSubpropertyValue is not null && (int)optionsSubpropertyValue < 1)
-                                            {
-                                                throw new ArgumentOutOfRangeException(nameof(MsSqlOptions.MaxDbResponseSizeMb), "MaxDbResponseSizeMB invalid. Max db response size argument cannot be less than 1MB");
-                                            }
-                                        }
                                         else
                                         {
                                             throw new JsonException($"Unexpected value for options {optionsSubproperty} while deserializing DataSource options.");
--- a/src/Config/Converters/DataSourceConverterFactory.cs
+++ b/src/Config/Converters/DataSourceConverterFactory.cs
@@ -103,6 +103,14 @@ internal class DataSourceConverterFactory : JsonConverterFactory
                                         {
                                             optionsSubpropertyValue = null;
                                         }
+                                        else if (reader.TokenType is JsonTokenType.Number)
+                                        {
+                                            optionsSubpropertyValue = reader.GetInt32();
+                                            if (optionsSubpropertyValue is not null && (int)optionsSubpropertyValue < 1)
+                                            {
+                                                throw new ArgumentOutOfRangeException(nameof(MsSqlOptions.MaxDbResponseSizeMb), "MaxDbResponseSizeMB invalid. Max db response size argument cannot be less than 1MB");
+                                            }
+                                        }
                                         else
                                         {
                                             throw new JsonException($"Unexpected value for options {optionsSubproperty} while deserializing DataSource options.");
--- a/src/Config/ObjectModel/HostOptions.cs
+++ b/src/Config/ObjectModel/HostOptions.cs
@@ -1,6 +1,80 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics.CodeAnalysis;
+using System.Net;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Service.Exceptions;
+
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
-public record HostOptions(CorsOptions? Cors, AuthenticationOptions? Authentication, HostMode Mode = HostMode.Production);
+public record HostOptions
+{
+    /// <summary>
+    /// Dab engine can at maximum handle 158 MB of data in a single response from a source.
+    /// Json deserialization of a response into a string has a limit of 166,666,666 bytes which when converted to MB is 158 MB.
+    /// ref: enforcing code:
+    /// .net8: https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.Text.Json/src/System/Text/Json/Writer/JsonWriterHelper.cs#L80
+    /// .net6: https://github.com/dotnet/runtime/blob/v8.0.0/src/libraries/System.Text.Json/src/System/Text/Json/Writer/JsonWriterHelper.cs#75
+    /// ref: Json constant: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/JsonConstants.cs#L80
+    /// </summary>
+    private const int MAX_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
+
+    /// <summary>
+    /// Dab engine default response length. As of now this is same as max response length.
+    /// </summary>
+    private const int DEFAULT_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
+
+    [JsonPropertyName("cors")]
+    public CorsOptions? Cors { get; init; }
+
+    [JsonPropertyName("authentication")]
+    public AuthenticationOptions? Authentication { get; init; }
+
+    [JsonPropertyName("mode")]
+    public HostMode Mode { get; init; }
+
+    [JsonPropertyName("max-response-size-mb")]
+    public int? MaxResponseSizeMB { get; init; } = null;
+
+    public HostOptions(CorsOptions? Cors, AuthenticationOptions? Authentication, HostMode Mode = HostMode.Production, int? MaxResponseSizeMB = null)
+    {
+        this.Cors = Cors;
+        this.Authentication = Authentication;
+        this.Mode = Mode;
+        this.MaxResponseSizeMB = MaxResponseSizeMB;
+
+        if (this.MaxResponseSizeMB is not null)
+        {
+            this.MaxResponseSizeMB = this.MaxResponseSizeMB == -1 ? MAX_RESPONSE_LENGTH_DAB_ENGINE_MB : (int)this.MaxResponseSizeMB;
+            if (this.MaxResponseSizeMB < 1 || this.MaxResponseSizeMB > MAX_RESPONSE_LENGTH_DAB_ENGINE_MB)
+            {
+                throw new DataApiBuilderException(
+                    message: $"{nameof(RuntimeConfig.Runtime.Host.MaxResponseSizeMB)} cannot be 0, exceed {MAX_RESPONSE_LENGTH_DAB_ENGINE_MB}MB or be less than -1",
+                    statusCode: HttpStatusCode.ServiceUnavailable,
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
+            }
+
+            UserProvidedMaxResponseSizeMB = true;
+        }
+        else
+        {
+            this.MaxResponseSizeMB = DEFAULT_RESPONSE_LENGTH_DAB_ENGINE_MB;
+        }
+    }
+
+    /// <summary>
+    /// Flag which informs CLI and JSON serializer whether to write MaxResponseSizeMB.
+    /// property and value to the runtime config file.
+    /// When user doesn't provide the MaxResponseSizeMB property/value or provides a null value, which signals DAB to use the default,
+    /// the DAB CLI will not write the default value to a serialized config.
+    /// This is because the user's intent is to use DAB's default value which could change
+    /// and DAB CLI writing the property and value would lose the user's intent.
+    /// This is because if the user were to use the CLI created config, MaxResponseSizeMB
+    /// property/value specified would be interpreted by DAB as "user explicitly set MaxResponseSizeMB.
+    /// UserProvidedMaxResponseSizeMB is true only when a user provides a non-null value
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
+    [MemberNotNullWhen(true, nameof(MaxResponseSizeMB))]
+    public bool UserProvidedMaxResponseSizeMB { get; init; } = false;
+}
--- a/src/Config/ObjectModel/HostOptions.cs
+++ b/src/Config/ObjectModel/HostOptions.cs
@@ -18,12 +18,12 @@ public record HostOptions
     /// .net6: https://github.com/dotnet/runtime/blob/v8.0.0/src/libraries/System.Text.Json/src/System/Text/Json/Writer/JsonWriterHelper.cs#75
     /// ref: Json constant: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/JsonConstants.cs#L80
     /// </summary>
-    private const int MAX_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
+    public const int MAX_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
 
     /// <summary>
     /// Dab engine default response length. As of now this is same as max response length.
     /// </summary>
-    private const int DEFAULT_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
+    public const int DEFAULT_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
 
     [JsonPropertyName("cors")]
     public CorsOptions? Cors { get; init; }
--- a/src/Core/Resolvers/QueryExecutor.cs
+++ b/src/Core/Resolvers/QueryExecutor.cs
@@ -36,8 +36,8 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 
         private RetryPolicy _retryPolicy;
 
-        private int? _maxDbResponseSizeMB;
-        private long? _maxDbResponseSizeBytes;
+        private int? _maxResponseSizeMB;
+        private long? _maxResponseSizeBytes;
         private bool _maxResponseSizeLogicEnabled;
 
         /// <summary>
@@ -55,10 +55,9 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             ConnectionStringBuilders = new Dictionary<string, DbConnectionStringBuilder>();
             ConfigProvider = configProvider;
             HttpContextAccessor = httpContextAccessor;
-            _maxDbResponseSizeMB = configProvider.GetConfig().MaxResponseSizeMB();
+            _maxResponseSizeMB = configProvider.GetConfig().MaxResponseSizeMB();
             _maxResponseSizeLogicEnabled = configProvider.GetConfig().MaxResponseSizeLogicEnabled();
-            // long value is larger than max int value * 1024 * 1024
-            _maxDbResponseSizeBytes = _maxDbResponseSizeMB * 1024 * 1024;
+            _maxResponseSizeBytes = _maxResponseSizeMB ?? _maxResponseSizeMB * 1024 * 1024;
 
             _retryPolicyAsync = Policy
                 .Handle<DbException>(DbExceptionParser.IsTransientException)
@@ -670,9 +669,9 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 
         /// <summary>
         /// This function reads the data from the DbDataReader and returns the JSON string.
-        /// 1. MaxDbResponse is used like a feature flag.
-        /// 2. If MaxDbResponse is not specified by the customer, getString is used and entire data is read into memory. No protections
-        /// 3. If MaxDbResponse is specified by the customer, getChars is used. GetChars tries to read the data in chunks and if the data is more than the specified limit, it throws an exception.
+        /// 1. MaxResponseSizeLogicEnabled is used like a feature flag.
+        /// 2. If MaxResponseSize is not specified by the customer or is null, getString is used and entire data is read into memory.
+        /// 3. If MaxResponseSize is specified by the customer, getChars is used. GetChars tries to read the data in chunks and if the data is more than the specified limit, it throws an exception.
         /// </summary>
         /// <param name="dbDataReader"></param>
         /// <returns></returns>
@@ -698,7 +697,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             else
             {
                 // max buffer size will always be smaller than the 
-                long availableSize = (long)_maxDbResponseSizeBytes!;
+                long availableSize = (long)_maxResponseSizeBytes!;
                 while (await ReadAsync(dbDataReader))
                 {
                     availableSize -= StreamData(dbDataReader, (int)availableSize, jsonString);
@@ -713,7 +712,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             if (availableSize - sizeToBeRead < 0)
             {
                 throw new DataApiBuilderException(
-                    message: $"The JSON result size exceeds max result size of {_maxDbResponseSizeMB}MB. Please use pagination to reduce size of result.",
+                    message: $"The JSON result size exceeds max result size of {_maxResponseSizeMB}MB. Please use pagination to reduce size of result.",
                     statusCode: HttpStatusCode.RequestEntityTooLarge,
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorProcessingData);
             }
--- a/src/Core/Resolvers/QueryExecutor.cs
+++ b/src/Core/Resolvers/QueryExecutor.cs
@@ -4,11 +4,13 @@
 using System.Data;
 using System.Data.Common;
 using System.Net;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 using Azure.DataApiBuilder.Core.Configurations;
 using Azure.DataApiBuilder.Core.Models;
+using Azure.DataApiBuilder.Core.Services;
 using Azure.DataApiBuilder.Service.Exceptions;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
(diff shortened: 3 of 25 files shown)


src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -103,6 +103,14 @@ public DataSourceConverter(bool replaceEnvVar)
                                         {
                                             optionsSubpropertyValue = null;
                                         }
+                                        else if (reader.TokenType is JsonTokenType.Number)
+                                        {
+                                            optionsSubpropertyValue = reader.GetInt32();
```
seantleonard, line 108: what happens if this line fails? e.g. when user sets value to 123.45 -> a JSON Number but not an integer.

rohkhann: Resolving as no longer valid and we moved to host options.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
         }
 
+        /// <summary>
+        /// Reads data into jsonString.
+        /// </summary>
+        /// <param name="dbDataReader">DbDataReader.</param>
+        /// <param name="availableSize">Available buffer.</param>
+        /// <param name="jsonString">jsonString to read into.</param>
+        /// <returns>size of data read in bytes.</returns>
+        public int StreamData(DbDataReader dbDataReader, int availableSize, StringBuilder jsonString)
```
seantleonard, line 655: Access modifier should be private. Callers only need to use one of the `Get*Result*()` functions.

If you need this for testing, you can reference other parts of the code base which use "MakeInternalsVisibleTo"

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
+
+            // read entire field into buffer and reduce available size.
+            dbDataReader.GetChars(0, 0, buffer, 0, fieldSize);
+
+            jsonString.Append(buffer[0..(int)(fieldSize)]);
+            return fieldSize;
+        }
+
+        /// <summary>
+        /// This function reads the data from the DbDataReader and returns the JSON string.
```
seantleonard, line 673:
```suggestion
        /// This function reads the data from the DbDataReader and returns a JSON string.
```

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -51,6 +55,10 @@ public class QueryExecutor<TConnection> : IQueryExecutor
             ConnectionStringBuilders = new Dictionary<string, DbConnectionStringBuilder>();
             ConfigProvider = configProvider;
             HttpContextAccessor = httpContextAccessor;
+            _maxResponseSizeMB = configProvider.GetConfig().MaxResponseSizeMB();
+            _maxResponseSizeLogicEnabled = configProvider.GetConfig().MaxResponseSizeLogicEnabled();
```
seantleonard, line 59: considering we are looking into enabling "hot reload" for dab. It would make more sense to directly use `configProvider.GetConfig().MaxResponseSizeLogicEnabled();` when determining whether to use new code path. since this constructor is only used once.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
+        /// <summary>
+        /// Reads data into jsonString.
+        /// </summary>
+        /// <param name="dbDataReader">DbDataReader.</param>
+        /// <param name="availableSize">Available buffer.</param>
+        /// <param name="jsonString">jsonString to read into.</param>
+        /// <returns>size of data read in bytes.</returns>
+        public int StreamData(DbDataReader dbDataReader, int availableSize, StringBuilder jsonString)
+        {
+            int fieldSize = (int)dbDataReader.GetChars(0, 0, null, 0, 0);
```
seantleonard, line 657: named params for GetChars(...)
```suggestion
            int resultFieldSize = (int)dbDataReader.GetChars(0, 0, null, 0, 0);
```

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
+
+            // if the size of the field is less than available size, then we can read the entire field.
+            // else we throw exception.
+            ValidateSize(availableSize, fieldSize);
+
+            char[] buffer = new char[fieldSize];
```
seantleonard, line 663: do we need to wrap this in try/catch to gracefully handle OOM case?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
+
+            // read entire field into buffer and reduce available size.
+            dbDataReader.GetChars(0, 0, buffer, 0, fieldSize);
+
+            jsonString.Append(buffer[0..(int)(fieldSize)]);
```
seantleonard, line 668: Casting fieldSize to int seems redundant as mentioned by intellisense.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -646,12 +688,36 @@ private async Task<string> GetJsonStringFromDbReader(DbDataReader dbDataReader)
...
+                while (await ReadAsync(dbDataReader))
+                {
+                    availableSize -= StreamData(dbDataReader, (int)availableSize, jsonString);
+                }
             }
 
             return jsonString.ToString();
         }
+
+        private void ValidateSize(long availableSize, long sizeToBeRead)
```
seantleonard, line 712: What do you mean by "validate size"as the function name is vague? 
- validate size is less than a certain amount?
- validate sizeToBeRead > 0 ? 

why are the params of type `long` when callers pass `int`?

need method summary and either:
- param name specificity: availableSizeMB, sizeToBeRead{unit}
- param description -> describes expected units of arg values.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -646,12 +688,36 @@ private async Task<string> GetJsonStringFromDbReader(DbDataReader dbDataReader)
...
+        {
+            if (availableSize - sizeToBeRead < 0)
```
seantleonard, line 714: why not `sizeToBeRead <= availableSize` ?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -646,12 +688,36 @@ private async Task<string> GetJsonStringFromDbReader(DbDataReader dbDataReader)
...
+            {
+                while (await ReadAsync(dbDataReader))
+                {
+                    jsonString.Append(dbDataReader.GetString(0));
+                }
+            }
+            else
             {
-                jsonString.Append(dbDataReader.GetString(0));
+                // max buffer size will always be smaller than the
```
seantleonard, line 701: incomplete comment

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
+
+            jsonString.Append(buffer[0..(int)(fieldSize)]);
+            return fieldSize;
+        }
+
+        /// <summary>
+        /// This function reads the data from the DbDataReader and returns the JSON string.
+        /// 1. MaxResponseSizeLogicEnabled is used like a feature flag.
+        /// 2. If MaxResponseSize is not specified by the customer or is null, getString is used and entire data is read into memory.
+        /// 3. If MaxResponseSize is specified by the customer, getChars is used. GetChars tries to read the data in chunks and if the data is more than the specified limit, it throws an exception.
```
seantleonard, line 676: nit: break up these lines. try to keep ~120 -130 char width per line.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
+        /// </summary>
+        /// <param name="dbDataReader"></param>
+        /// <returns></returns>
```
seantleonard, line 679: empty

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +645,39 @@ DbResultSet dbResultSet
...
+        /// <exception cref="DataApiBuilderException"></exception>
```
seantleonard, line 680: empty explanation

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -646,12 +688,36 @@ private async Task<string> GetJsonStringFromDbReader(DbDataReader dbDataReader)
...
+                }
+            }
+            else
             {
-                jsonString.Append(dbDataReader.GetString(0));
+                // max buffer size will always be smaller than the 
+                long availableSize = (long)_maxResponseSizeBytes!;
+                while (await ReadAsync(dbDataReader))
+                {
+                    availableSize -= StreamData(dbDataReader, (int)availableSize, jsonString);
```
seantleonard, line 705: what happens when availableSize (long) is out of range of int param -> `StreamData(.., int availableSize, ..)`?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -36,6 +36,10 @@ public class QueryExecutor<TConnection> : IQueryExecutor
 
         private RetryPolicy _retryPolicy;
 
+        private int? _maxResponseSizeMB;
+        private long? _maxResponseSizeBytes;
```
seantleonard, line 40: these aren't used in pre-existing code path.. do they need to be nullable? you've calculate what the max value can be, can just set that as default. Then you don't have to have all the explicit casts and `!` null forgiving operators in your code.

src/Core/Resolvers/QueryExecutor.cs
seantleonard: what happens when `dbDataReader.GetChars` returns a value outside the range of `int`?

src/Core/Resolvers/QueryExecutor.cs
seantleonard: if dab is processing >1 request, `availableSize` may be out of date by the time you're creating your buffer and could result in buffer alloc failure, couldn't it?

src/Core/Resolvers/QueryExecutor.cs
seantleonard: can also make update param name `jsonString` to better indicate that its value is affected by executing this function. e.g. `resultJsonString`.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
             Assert.AreEqual(3, queryExecutorLogger.Invocations.Count);
         }
 
+        /// <summary>
+        /// Validates that a query successfully executes within two retries by checking that the SqlQueryExecutor logger
+        /// was invoked the expected number of times.
+        /// </summary>
```
seantleonard, line 324: this description doesn't seem applicable to your new code. are you testing a specific function (and related downstream calls) within QueryExecutor? are you testing behavior interacting with a single specific helper function?

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+        [DataTestMethod, TestCategory(TestCategory.MSSQL)]
+        [DataRow(false, 4, DisplayName = "Data should be successfully read when amount of data being read is less than max-db-response-size-mb")]
+        [DataRow(false, 5, DisplayName = "Data should be successfully read when amount of data being read is less than max-db-response-size-mb")]
+        [DataRow(true, 6, DisplayName = "Exception should be thrown when amount of data read is more than max-db-response-size-mb")]
```
seantleonard, line 328: why is magic number `6` input result in a value more than `max-db-response-size-mb`?

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
             Assert.AreEqual(3, queryExecutorLogger.Invocations.Count);
         }
 
+        /// <summary>
+        /// Validates that a query successfully executes within two retries by checking that the SqlQueryExecutor logger
+        /// was invoked the expected number of times.
+        /// </summary>
+        [DataTestMethod, TestCategory(TestCategory.MSSQL)]
+        [DataRow(false, 4, DisplayName = "Data should be successfully read when amount of data being read is less than max-db-response-size-mb")]
+        [DataRow(false, 5, DisplayName = "Data should be successfully read when amount of data being read is less than max-db-response-size-mb")]
```
seantleonard, line 327: data row displaynames must be unique. Otherwise they don't render as expected in test explorer in visual studio.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+        [DataRow(true, 6, DisplayName = "Exception should be thrown when amount of data read is more than max-db-response-size-mb")]
+        public void ValidateStreamingLogicAsync(bool exceptionExpected, int readDataLoops)
```
seantleonard, line 329: in my opinion, order of params should be swapped. Reads as "input x results in Y {exception/no exception}"

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+                }
+
+            }
+            catch (DataApiBuilderException ex)
+            {
+                Assert.IsTrue(exceptionExpected);
+                Assert.AreEqual(HttpStatusCode.RequestEntityTooLarge, ex.StatusCode);
+                Assert.AreEqual("The JSON result size exceeds max result size of 5MB. Please use pagination to reduce size of result.", ex.Message);
+            }
+
```
seantleonard, line 374:
```suggestion
```

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+            DbExceptionParser dbExceptionParser = new MsSqlDbExceptionParser(runtimeConfigProvider);
+
+            // Instantiate the MsSqlQueryExecutor and Setup parameters for the query
+            MsSqlQueryExecutor msSqlQueryExecutor = new(runtimeConfigProvider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object);
+
+            try
+            {
+                /// In this test the DbDataReader.GetChars method is mocked to return 1024*1024 bytes (1MB) of data.
+                /// Max available size is set to 5 MB.
+                /// Based on number of loops, the data read will be 1MB * readDataLoops.Exception should be thrown in test cases where we go above 5MB.
```
seantleonard, line 357:
```suggestion
                /// Based on number of loops, the data read will be 1MB * readDataLoops. Exception should be thrown in test cases where we go above 5MB.
```

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+                Mock<DbDataReader> dbDataReader = new();
+                dbDataReader.Setup(d => d.HasRows).Returns(true);
+                dbDataReader.Setup(x => x.GetChars(It.IsAny<int>(), It.IsAny<long>(), It.IsAny<char[]>(), It.IsAny<int>(), It.IsAny<int>())).Returns(1024 * 1024);
+                int availableSize = (int)runtimeConfig.MaxResponseSizeMB() * 1024 * 1024;
+                for (int i = 0; i < readDataLoops; i++)
+                {
+                    availableSize -= msSqlQueryExecutor.StreamData(dbDataReader.Object, availableSize, new());
```
seantleonard, line 364: named params

seantleonard: when real QueryExecutor.StreamData(..) is called, would you be passing a new StringBuilder multiple times in a single request? or reusing the same stringbuilder?

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,62 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+                }
+
```
seantleonard, line 366:
```suggestion
```

src/Core/Resolvers/QueryExecutor.cs
seantleonard: do you also need to accommodate for the out of range exception that this can throw?

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
seantleonard: Looks like "system under test (SUT)" is `msSqlQueryExecutor.StreamData(dbDataReader.Object, availableSize, new());` and you're just ensuring the SUT raises an exception when you need to stream data but availableSize is less than the amount you need to stream.

seantleonard (changes requested):

seantleonard:
Unclear which Integration tests apply to this PR as mentioned in PR description:
> Integration Tests

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Available size is the size available per request. So we are not keeping track of it across requests. You would have to have separate throttling limits for high cost requests.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed this to make it long. In the validate, we should handle if the result returned by getchars is greater than max int number of bytes (roughly 2GB) as our maximum limit in any case would be 158MB.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: availableSize is on a per request basis. We are trying to limit the amount of data that can be pulled in a single request. I think solution for multiple requests would be to appropriate throtting of requests at api layer and then appropriately scaling instances.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed this. though technically this can never happen as we have a max supported input of 158MB which when multiplied *1024*1024 is within int limits.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: fixed.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
rohkhann: fixed description.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
rohkhann: This was explained in lines 357-359. Ive moved it to the summary description. hopefully should be clearer.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
rohkhann: fixed.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
rohkhann: fixed. We would be reusing the same stringbuilder.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Changed a bit of the logic here and fixed.

vadeveka (approved):

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -318,6 +318,63 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+        /// Based on number of loops, the data read will be 1MB * readDataLoops.Exception should be thrown in test cases where we go above 5MB.
+        /// This will be in cases where readDataLoops > 5.
+        /// </summary>
+        [DataTestMethod, TestCategory(TestCategory.MSSQL)]
+        [DataRow(4, false,
+            DisplayName = "Max available size is set to 5MB.Reading 1MB of data over 4 loops should succeed as amount of data being read is less than max-db-response-size-mb")]
+        [DataRow(5, false,
+            DisplayName = "Max available size is set to 5MB.Reading 1MB of data over 5 loops should succeed as amount of data being read is equal to max-db-response-size-mb")]
+        [DataRow(6, true,
+            DisplayName = "Max available size is set to 5MB.Reading 1MB of data over 6 loops should fail as amount of data being read exceeds max-db-response-size-mb")]
```
seantleonard, line 334: Super nit for next future consideration. Easy to immediately identify what DataRow param means.
```suggestion
            DisplayName = "6 data read loop iterations * 1MB -> Fails to read 6MB because max-db-response-size-mb is 5MB")]
```

seantleonard (approved):

src/Core/Resolvers/QueryExecutor.cs
seantleonard: It's still possible for ` char[] buffer = new char[fieldSize];` to raise an exception though.

seantleonard: but i suppose we don't want to be wrapping all `new` usage in try/catch

src/Config/ObjectModel/HostOptions.cs
```diff
@@ -18,12 +18,12 @@ public record HostOptions
...
     /// ref: Json constant: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/JsonConstants.cs#L80
     /// </summary>
-    private const int MAX_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
+    public const int MAX_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
 
     /// <summary>
     /// Dab engine default response length. As of now this is same as max response length.
     /// </summary>
-    private const int DEFAULT_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
+    public const int DEFAULT_RESPONSE_LENGTH_DAB_ENGINE_MB = 158;
```
aaronburtle, line 26: this 158 was picked because its slightly lower than the limit imposed by JSON serializing if I recall correctly, should we add that little detail here so in the future this magic number isn't so mysterious?

src/Core/Resolvers/QueryExecutor.cs
rohkhann: It would be very unlikely for buffer to raise an OOM exception though as we validate against the 158MB limit set (in the worst case) in validate size call and the memory should be available for GC after this function finishes.

src/Config/ObjectModel/HostOptions.cs
rohkhann: The description for why 158 is above max response size. line 18

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -636,6 +643,38 @@ DbResultSet dbResultSet
...
+        /// <summary>
+        /// Reads data into jsonString.
+        /// </summary>
+        /// <param name="dbDataReader">DbDataReader.</param>
+        /// <param name="availableSize">Available buffer.</param>
+        /// <param name="resultJsonString">jsonString to read into.</param>
+        /// <returns>size of data read in bytes.</returns>
+        internal int StreamData(DbDataReader dbDataReader, long availableSize, StringBuilder resultJsonString)
+        {
+            long resultFieldSize = dbDataReader.GetChars(0, 0, null, 0, 0);
```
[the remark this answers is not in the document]
rohkhann: fixed.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
rohkhann: fixed.

PR #2231, opened 2024-05-23
Fixing DateTime null parse issue in datawarehouse queries.

rohkhann:
## Why make this change?
When we moved to the new form of query execution: https://github.com/Azure/data-api-builder/pull/2068, we got rid of the resolver middleware that had the function: 
`        public static bool RepresentsNullValue(JsonElement element)
        {
            return (string.IsNullOrEmpty(element.ToString()) && element.GetRawText() == "null");
        }`

which would check if the returned string was null. This was useful for datetime values in datawarehouse where because of the use of string agg we have to return null as a string. 

The error that is seen is:         "The string 'null' was not recognized as a valid DateTime. There is an unknown word starting at index '0'.
## What is this change?
This change switches to leveraging tryparse for the datetime field value. This allows us to handle the null case, as if the string is null, tryparse will just fail and return null for date. 

## How was this tested?
A test has been added to ensure that datetime null case is handled.

### DIFF
--- a/src/Core/Services/ExecutionHelper.cs
+++ b/src/Core/Services/ExecutionHelper.cs
@@ -165,8 +165,8 @@ namespace Azure.DataApiBuilder.Service.Services
                     FloatType => fieldValue.GetDouble(), // spec
                     SingleType => fieldValue.GetSingle(),
                     DecimalType => fieldValue.GetDecimal(),
-                    DateTimeType => DateTimeOffset.Parse(fieldValue.GetString()!, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
-                    DateType => DateTimeOffset.Parse(fieldValue.GetString()!),
+                    DateTimeType => DateTimeOffset.TryParse(fieldValue.GetString()!, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal, out DateTimeOffset date) ? date : null, // for DW when datetime is null it will be in "" due to stringagg parsing and hence we need to ensure parsing is correct.
+                    DateType => DateTimeOffset.TryParse(fieldValue.GetString()!, out DateTimeOffset date) ? date : null,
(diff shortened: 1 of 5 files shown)


abhishekkumams:
is this issue only happening for DWSQL or also for other DBs?

src/Core/Services/ExecutionHelper.cs
```diff
@@ -165,8 +165,8 @@ public async ValueTask ExecuteMutateAsync(IMiddlewareContext context)
                     FloatType => fieldValue.GetDouble(), // spec
                     SingleType => fieldValue.GetSingle(),
                     DecimalType => fieldValue.GetDecimal(),
-                    DateTimeType => DateTimeOffset.Parse(fieldValue.GetString()!, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal),
-                    DateType => DateTimeOffset.Parse(fieldValue.GetString()!),
+                    DateTimeType => DateTimeOffset.TryParse(fieldValue.GetString()!, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal, out DateTimeOffset date) ? date : null, // for DW when datetime is null it will be in "" due to stringagg parsing and hence we need to ensure parsing is correct.
```
ayush3797, line 168:
```suggestion
                    DateTimeType => DateTimeOffset.TryParse(fieldValue.GetString()!, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal, out DateTimeOffset date) ? date : null, // for DW when datetime is null it will be in "" (double quotes) due to stringagg parsing and hence we need to ensure parsing is correct.
```

ayush3797 (approved):

seantleonard:
> Please add before/after bug fix screenshots. That'll help.

sample requests in text format preferred so that way we can copy/paste into our own GraphQL requests

vadeveka (approved):

rusamant (approved):

PR #2242, opened 2024-05-31
Add MaxResponseSizeMB to runtime config.

rohkhann:
## Why make this change?
This change adds in a new property to the host parameter of runtime config called MaxResponseSizeMB. This property will allow users to configure the amount of data that their host platform's memory can handle when streaming data from the underlying datasources.

## What is this change?
This change adds the maxresponseSizeMB property to the runtime config. In addition, it adds an accessor for that property which would be needed to implement the actual logic of streaming. 

## How was this tested?
Unit tests are added.

### DIFF
--- a/src/Config/Converters/HostOptionsConverterFactory.cs
+++ b/src/Config/Converters/HostOptionsConverterFactory.cs
@@ -1,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.ObjectModel;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+
+/// <summary>
+/// Defines how DAB reads and writes an entity's cache options (JSON).
+/// </summary>
+internal class HostOptionsConvertorFactory : JsonConverterFactory
+{
+    /// <inheritdoc/>
+    public override bool CanConvert(Type typeToConvert)
+    {
+        return typeToConvert.IsAssignableTo(typeof(HostOptions));
+    }
+
+    /// <inheritdoc/>
+    public override JsonConverter? CreateConverter(Type typeToConvert, JsonSerializerOptions options)
+    {
+        return new HostOptionsConverter();
+    }
+
+    private class HostOptionsConverter : JsonConverter<HostOptions>
+    {
+        /// <summary>
+        /// Defines how DAB reads host options and defines which values are
+        /// used to instantiate HostOptions.
+        /// </summary>
+        /// <exception cref="JsonException">Thrown when improperly formatted cache options are provided.</exception>
+        public override HostOptions? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            return JsonSerializer.Deserialize<HostOptions>(ref reader, options);
+        }
+
+        /// <summary>
+        /// When writing the HostOptions back to a JSON file, only write the MaxResponseSizeMB property
+        /// if the property is user provided. This avoids polluting the written JSON file with a property
+        /// the user most likely ommitted when writing the original DAB runtime config file.
+        /// This Write operation is only used when a RuntimeConfig object is serialized to JSON.
+        /// </summary>
+        public override void Write(Utf8JsonWriter writer, HostOptions value, JsonSerializerOptions options)
+        {
+            writer.WriteStartObject();
+            writer.WritePropertyName("cors");
+            JsonSerializer.Serialize(writer, value.Cors, options);
+            writer.WritePropertyName("authentication");
+            JsonSerializer.Serialize(writer, value.Authentication, options);
+            writer.WritePropertyName("mode");
+            JsonSerializer.Serialize(writer, value.Mode, options);
+
+            if (value?.UserProvidedMaxResponseSizeMB is true)
+            {
+                writer.WritePropertyName("max-response-size-mb");
+                JsonSerializer.Serialize(writer, value.MaxResponseSizeMB, options);
+            }
+
+            writer.WriteEndObject();
+        }
+    }
+}
--- a/src/Config/Converters/HostOptionsConverterFactory.cs
+++ b/src/Config/Converters/HostOptionsConverterFactory.cs
@@ -8,7 +8,7 @@ using Azure.DataApiBuilder.Config.ObjectModel;
 namespace Azure.DataApiBuilder.Config.Converters;
 
 /// <summary>
-/// Defines how DAB reads and writes an entity's cache options (JSON).
+/// Defines how DAB reads and writes host options.
 /// </summary>
 internal class HostOptionsConvertorFactory : JsonConverterFactory
 {
@@ -29,8 +29,9 @@ internal class HostOptionsConvertorFactory : JsonConverterFactory
         /// <summary>
         /// Defines how DAB reads host options and defines which values are
         /// used to instantiate HostOptions.
+        /// Uses default deserialize.
         /// </summary>
-        /// <exception cref="JsonException">Thrown when improperly formatted cache options are provided.</exception>
+        /// <exception cref="JsonException">Thrown when improperly formatted host options are provided.</exception>
         public override HostOptions? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
             return JsonSerializer.Deserialize<HostOptions>(ref reader, options);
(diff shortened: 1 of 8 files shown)


src/Config/ObjectModel/HostOptions.cs
```diff
@@ -1,6 +1,45 @@
...
+using Azure.DataApiBuilder.Service.Exceptions;
+
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
-public record HostOptions(CorsOptions? Cors, AuthenticationOptions? Authentication, HostMode Mode = HostMode.Production);
+public record HostOptions
+{
+    /// <summary>
+    /// Dab engine can at maximum handle 187 MB of data in a single response from a source.
+    /// Json deserialization of a response into a string has a limit of 197020041 bytes which when converted to MB is 187 MB.
```
vadeveka, line 14: Where does 197020041 number come from? Limit in System.Text.Json dserializer ? Can we be more specific?

vadeveka (approved):

src/Config/ObjectModel/HostOptions.cs
seantleonard: Rohan, you can link to Microsoft Learn docs and to a section of code in .NET's github repo.

seantleonard (changes requested):
When i have a dab-config.json where i set `"max-response-size-mb": null` and then run DAB's CLI -> `dab add Book --config "path/to/config.json" --source books --permissions "anonymous:create,read,update,delete" --graphql "book:books"`

The field `"max-response-size-mb"` disappears from my config. 
The use case is this: I am explicitly setting the property and value in my config. I am saying: "i want no limit to max response size, do NOT give me DAB's default value as that could change in the future. My config is explicitly telling dab what i want and the CLI shouldn't remove my explicitly defined field."

src/Config/ObjectModel/HostOptions.cs
rohkhann: Added.

src/Service.Tests/Unittests/ConfigValidationUnitTests.cs
```diff
@@ -2414,6 +2414,48 @@ public void TestRuntimeConfigSetupWithNonJsonConstructor()
...
+            DisplayName = "Valid inputs of MaxResponseSize must be accepted and set in the config.")]
+        [DataRow(true, -1, $"{nameof(RuntimeConfig.Runtime.Host.MaxResponseSizeMB)} must be greater than 0 and <= 187 MB.", null,
+            DisplayName = "Valid inputs of MaxResponseSize must be accepted and set in the config.")]
+        [DataRow(true, 188, $"{nameof(RuntimeConfig.Runtime.Host.MaxResponseSizeMB)} must be greater than 0 and <= 187 MB.", null,
+            DisplayName = "Valid inputs of MaxResponseSize must be accepted and set in the config.")]
+        public void ValidateMaxResponseSizeInConfig(
+            bool exceptionExpected,
+            int? maxDbResponseSizeMB,
+            string expectedExceptionMessage,
+            int? expectedMaxResponseSize = null)
```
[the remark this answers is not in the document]
rohkhann: Hmm, I see similar practice for other test cases in dab :https://github.com/Azure/data-api-builder/blob/05588152e8c1c0a4ea86230c95b4460bf126b97f/src/Service.Tests/Unittests/ConfigValidationUnitTests.cs#L2190 and other similar test cases in ConfigValidationUnitTests.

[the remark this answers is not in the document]
seantleonard: The displayNames must all be unique too. 

> Hmm, I see similar practice for other test cases in dab :

Sure, would those tests benefit from any changes to be more readable? In my opinion with this specific test, I'm merely aiming for improved readability. If not breaking up the test, perhaps order and consistent names of parameters make readability better?
- `providedMaxResponseSizeMB` next to `expectedMaxResponseSizeMB`
- proceeded by `exceptionExpected` and `expectedExceptionMessage`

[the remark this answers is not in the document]
rohkhann: Fixed!

rohkhann:
> When i have a dab-config.json where i set `"max-response-size-mb": null` and then run DAB's CLI -> `dab add Book --config "path/to/config.json" --source books --permissions "anonymous:create,read,update,delete" --graphql "book:books"`
> 
> The field `"max-response-size-mb"` disappears from my config. The use case is this: I am explicitly setting the property and value in my config. I am saying: "i want no limit to max response size, do NOT give me DAB's default value as that could change in the future. My config is explicitly telling dab what i want and the CLI shouldn't remove my explicitly defined field."

Added similar flow to pagination options. Added all the things discussed in design doc.

src/Config/Converters/HostOptionsConverterFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// the user most likely ommitted when writing the original DAB runtime config file.
+        /// This Write operation is only used when a RuntimeConfig object is serialized to JSON.
+        /// </summary>
+        public override void Write(Utf8JsonWriter writer, HostOptions value, JsonSerializerOptions options)
+        {
+            writer.WriteStartObject();
+            writer.WritePropertyName("cors");
+            JsonSerializer.Serialize(writer, value.Cors, options);
+            writer.WritePropertyName("authentication");
+            JsonSerializer.Serialize(writer, value.Authentication, options);
```
seantleonard, line 55: `HostOptions` method signature includes: `CorsOptions? Cors, AuthenticationOptions? Authentication`. Will this error out when `value.Cors` or `value.Authentication` is null?

src/Config/ObjectModel/HostOptions.cs
```diff
@@ -1,6 +1,77 @@
...
+        else
+        {
+            this.MaxResponseSizeMB = DEFAULT_RESPONSE_LENGTH_DAB_ENGINE_MB;
+        }
+    }
+
+    /// <summary>
+    /// Flag which informs CLI and JSON serializer whether to write MaxResponseSizeMB.
+    /// property and value to the runtime config file.
+    /// When user doesn't provide the MaxResponseSizeMB property/value, which signals DAB to use the default,
```
seantleonard, line 67: Should clarify that `UserProvidedMaxResponseSizeMB` is true only when a user provides a non-null value. Dab will ignore the property even if the user writes the following in their config:
```json
{
    "max-response-size-mb": null
}
```

rohkhann: Ok let me add that in.

src/Config/Converters/HostOptionsConverterFactory.cs
rohkhann: I validated that it would not by sending those in as null: 
image
image

src/Config/ObjectModel/HostOptions.cs
rohkhann: Added in.

src/Config/ObjectModel/HostOptions.cs
```diff
@@ -1,6 +1,78 @@
...
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
-public record HostOptions(CorsOptions? Cors, AuthenticationOptions? Authentication, HostMode Mode = HostMode.Production);
+public record HostOptions
+{
+    /// <summary>
+    /// Dab engine can at maximum handle 158 MB of data in a single response from a source.
+    /// Json deserialization of a response into a string has a limit of 166,666,666 bytes which when converted to MB is 158 MB.
+    /// ref: enforcing code: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Writer/JsonWriterHelper.cs#L103
+    /// ref: Json constant: https://github.com/stephentoub/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/JsonConstants.cs
```
seantleonard, line 17: nit: https://github.com/dotnet/runtime link is for latest runtime code, you'd want to ensure you select tag for .net8 and/or .net6 to demonstrate the code responsible for behavior we see today.

nit: https://github.com/stephentoub is .net engineer's personal repo `System.Text.Json/src/System/Text/Json/JsonConstants.cs`.. any chance you can make this direct link to dotnet/runtime repo?

seantleonard (approved):
last two nits on comments. Thanks for making these changes!

src/Config/ObjectModel/HostOptions.cs
rohkhann: Yes let me fix this

rohkhann: fixed.

Issue #2249, opened 2024-06-06
Add a CLI command to configure GraphQL depth limit in the config file.


abhishekkumams:
Related #1577.
This is required to give users flexibility to add/update depth limit for graphQL queries.

JerryNixon:
Related to #2251

PR #2250, opened 2024-06-06
Add new cli command to update runtime with graphql depth limit

### LINKED ISSUE #2243: Add a new CLI command to update values of the runtime section of the config
# Background
Currently, the properties in the runtime section of the config file are only setup once, and the values are specified with the `dab init` command. 

# Problem
As we add new properties in the runtime section, the `init` command is growing fast and becoming more complex. 
Furthermore, the CLI currently does not support updating these properties post-initialization. This means that if a user wants to change a property in the runtime section after running the init command, they have to manually edit the config file. This process is not only inconvenient but also prone to errors, especially for users who are not familiar with the JSON syntax or the structure of the config file.

Moreover, re-initializing the config just to update a single property is not a practical solution. It's a time-consuming process that requires the user to re-enter all the configuration details, even if they only want to change one property. This can be particularly frustrating for users who have a complex configuration setup.

This lack of support for updating the runtime properties post-initialization is a significant limitation of the current CLI. It makes the process of managing and updating the configuration more difficult and less flexible than it needs to be. This can be a barrier to efficient usage and can negatively impact the user experience.

# Proposal
There are 3 major parts of our config file in high-level : `data-source`,`runtime`,`entity`. `data-source` doesn't require change often and is usually an initialization section. For `entities`, we already have `add` and `update` command. So, I propose to add a new CLI command for updating just the runtime section of the config. This is how our runtime section of the config file looks like:

```json
"runtime": {
    "rest": {
      "enabled": true,
      "path": "/api",
      "request-body-strict": true
    },
    "graphql": {
      "enabled": true,
      "path": "/graphql",
      "allow-introspection": true
    },
    "host": {
      "cors": {
        "origins": [
          "http://localhost:5000"
        ],
        "allow-credentials": false
      },
      "authentication": {
        "provider": "StaticWebApps"
      },
      "mode": "development"
    }
  }
```

Now, during initialization our runtime section will be setup with the user provided value or default values. But later using the below command they can be updated.

`dab update-runtime [options]`

## Available options
it will have all the options to edit the runtime section of the config.
| Options | Option Required | Default Value | Value Required | Value Type | Description |
| --- | --- | --- | --- | --- | --- |
| **--host-mode** | ❌ No | `production` | ✔️ Yes | string | Specify the Host mode - development or production |
| **--cors-origin** | ❌ No | `""` | ✔️ Yes | string | Specify the list of allowed origins. |
| **--auth.provider** | ❌ No | `StaticWebApps` | ✔️ Yes | string | Specify the Identity Provider. |
| **--rest.path** | ❌ No | `/api` | ✔️ Yes | string | Specify the REST endpoint's prefix. |
| **--rest.disabled** | ❌ No | `false` | ❌ No | bool | Disables REST endpoint for all entities. |
| **--rest.enabled** | ❌ No | `true` | ✔️ Yes | bool | Enables REST endpoint for all entities. |
| **--rest.request-body-strict** | ❌ No | `true` | ✔️ Yes | bool | Doesn't allow extraneous fields in request body. |
| **--graphql.path** | ❌ No | `/graphql` | ✔️ Yes | string | Specify the GraphQL endpoint's prefix. |
| **--graphql.disabled** | ❌ No | `false` | ❌ No | bool | Disables GraphQL endpoint for all entities. |
| **--graphql.enabled** | ❌ No | `true` | ✔️ Yes | bool | Enables GraphQL endpoint for all entities. |
| **--graphql.multiple-create.enabled** | ❌ No | `false` | ✔️ Yes | bool | Enables multiple create functionality in GraphQL. |
| **--auth.audience** | ❌ No | | ✔️ Yes | string | Identifies the recipients that the Json Web Token (JWT) is intended for. |
| **--auth.issuer** | ❌ No | | ✔️ Yes | string | Specify the party that issued the JWT token. |
| **-c,--config** | ❌ No | `dab-config.json` | ✔️ Yes | string | Path to config file. |

abhishekkumams:
## Why make this change?

- Closes #2243, #2249
  - Currently CLI doesn't support updating runtime settings.
  - Need a flexible way to add/update graphQL depth limit

## What is this change?

- This PR contains changes related to CLI.
- Updated the draft schema to add the new property in the runtime section.
- A new command is added `dab configure`, which will allow updates for non-entity based config settings.
- We have added an optional flag `--runtime.graphql.depth-limit`, which will be used to specify the depth limit for GraphQL queries and mutation.
- By Default, No limit is imposed.
- values supported are only integer values [1, Int32.MaxValue].
example: `dab configure --runtime.graphql.depth-limit 20`.

## NOTE:
if limit is not already specified, it will add this property to config, else update it.
- Additionally, users can use -1 to remove limit.
`dab configure --runtime.graphql.depth-limit -1`

## How was this tested?
- [X] Unit Tests

## Sample Request(s)

initial config
```json
"graphql": {
    "path": "/graphql",
    "enabled": true,
    "allow-introspection": true
}
```

command: `dab configure --runtime.graphql.depth-limit 20`
```json
"graphql": {
    "path": "/graphql",
    "enabled": true,
    "allow-introspection": true,
    "depth-limit": 20
}
```

command: `dab configure --runtime.graphql.depth-limit 8`
```json
"graphql": {
    "path": "/graphql",
    "enabled": true,
    "allow-introspection": true,
    "depth-limit": 8
}
```

command: `dab configure --runtime.graphql.depth-limit -1`
```json
"graphql": {
    "path": "/graphql",
    "enabled": true,
    "allow-introspection": true,
    "depth-limit": -1
}
```

Special Case:
if value of "depth-limit" is null in the config. it will stay as is, if not updated.
```json
"graphql": {
    "path": "/graphql",
    "enabled": true,
    "allow-introspection": true,
    "depth-limit": null
}
```

command: `dab configure`

No change:
```json
"graphql": {
    "path": "/graphql",
    "enabled": true,
    "allow-introspection": true,
    "depth-limit": null
}
```


src/Cli.Tests/TestHelper.cs
```diff
@@ -1226,6 +1227,43 @@ public static Process ExecuteDabCommand(string command, string flags)
...
+            ""runtime"": {{
+                ""rest"": {{
+                    ""path"": ""/api"",
+                    ""enabled"": true,
+                    ""request-body-strict"": true
+                }},
+                ""graphql"": {{
+                    ""path"": ""/graphql"",
+                    ""enabled"": true,
+                    ""allow-introspection"": true
```
seantleonard, line 1249:
```suggestion
                    ""allow-introspection"": true,
```

needs comma?

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -114,6 +114,37 @@ public void TestValidateWithEmptyConfig()
...
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("null", true, false, DisplayName = "Invalid Value: 'null'")]
+    [DataRow("20", true, true, DisplayName = "Valid Value: 20")]
+    [DataRow(-1, false, false, DisplayName = "Invalid Value: -1")]
+    [DataRow(0, false, false, DisplayName = "Invalid Value: 0")]
+    [DataRow(-2, false, false, DisplayName = "Invalid Value: -2")]
+    [DataRow(2147483648, false, false, DisplayName = "Invalid Value: 2147483648")]
+    [DataRow("seven", true, false, DisplayName = "Invalid Value: 'seven'")]
+    public void TestValidateConfigFailsWithInvalidGraphQLDepthLimit(object? depthLimit, bool isStringValue, bool isSuccessExpected)
```
seantleonard, line 128: The data type on the runtime options property enforces that this value must be an integer.. i think it would be better to exclude the last test case of "seven" and make the test method paramter input data type an `int?`

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -114,6 +114,37 @@ public void TestValidateWithEmptyConfig()
...
+    /// <summary>
+    /// This Test is used to verify that the validate command is able to catch invalid values for the depth-limit property.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("null", true, false, DisplayName = "Invalid Value: 'null'")]
+    [DataRow("20", true, true, DisplayName = "Valid Value: 20")]
+    [DataRow(-1, false, false, DisplayName = "Invalid Value: -1")]
+    [DataRow(0, false, false, DisplayName = "Invalid Value: 0")]
+    [DataRow(-2, false, false, DisplayName = "Invalid Value: -2")]
+    [DataRow(2147483648, false, false, DisplayName = "Invalid Value: 2147483648")]
```
seantleonard, line 126: Display names would be more description if you listed why the value is invalid. isn't `-1` valid to denote "no depth limit".

Unclear why the string value `"20"` is a valid input for query-depth, that isn't an integer. That test case also contradicts the test method name which ensures invalid values result in failure

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Cli.Tests
+{
+    /// <summary>
+    /// Tests to validate cli commands to update runtime settings.
```
seantleonard, line 7:
```suggestion
    /// Validates that `dab runtime [options]` updates runtime settings with the user-provided options.
```

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// </summary>
+    [TestClass]
+    public class UpdateRuntimeTests : VerifyBase
+    {
+        private MockFileSystem? _fileSystem;
+        private FileSystemRuntimeConfigLoader? _runtimeConfigLoader;
+        private static string _testConfigFile = "test-update-runtime-setting.json";
```
seantleonard, line 14: const to be consistent with other code.

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            ILoggerFactory loggerFactory = TestLoggerSupport.ProvisionLoggerFactory();
+
+            SetLoggerForCliConfigGenerator(loggerFactory.CreateLogger<ConfigGenerator>());
+            SetCliUtilsLogger(loggerFactory.CreateLogger<Utils>());
+        }
+
+        /// <summary>
+        /// Test to validate when update runtime setting command is used with no options
+        /// no changes are made to the config file.
```
seantleonard, line 31:
```suggestion
        /// Validates that `dab runtime` is a no-op because the user provided no options to update
        /// the config file. The config file must not change.
```

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// </summary>
+        [TestMethod]
+        public void TestNoUpdateRuntimeSettings()
+        {
+            _fileSystem!.AddFile(_testConfigFile, new MockFileData(INITIAL_CONFIG));
+
+            UpdateRuntimeOptions options = new(
```
seantleonard, line 38: what happens when user runs: `dab runtime --query-depth null`?

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            string updatedConfig = _fileSystem!.File.ReadAllText(_testConfigFile);
+            // Assert that INITIAL_CONFIG is same as the updated config
+            Assert.IsTrue(JToken.DeepEquals(JObject.Parse(INITIAL_CONFIG), JObject.Parse(updatedConfig)));
+        }
+
+        /// <summary>
+        /// Test to validate when no limit was specified, it gets added to the config file
+        /// using update runtime command. Here we are adding depth limit of 8 for GraphQL.
+        /// Once added it should be present in the config file.
+        [TestMethod]
```
seantleonard, line 55: this summary is worded weird. "when no limit was specified," not specified when? when running `dab runtime --query-depth` ?

perhaps a clearer summary would state that: running "dab runtime --query-depth 8" on a config with no depth previously specified, results
in runtime.graphql.querydepthlimit property being set with the value 8.

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
+            Assert.IsNotNull(config);
+            Assert.IsNull(config!.Runtime!.GraphQL!.DepthLimit);
+
+            // Add Depth Limit
+            UpdateRuntimeOptions options = new(
+                depthLimit: maxDepthLimit,
+                config: _testConfigFile
+            );
+            Assert.IsTrue(TryUpdateRuntimeSettings(options, _runtimeConfigLoader!, _fileSystem!));
```
seantleonard, line 73: Nit: clarity between Arrange, Act, and Assert sections of this test. 
```suggestion
            // Arrange 
            // {setup code}

            // Act -> attempts to update query depth property
            bool didUpdateOperationSucceed = TryUpdateRuntimeSettings(options, _runtimeConfigLoader!, _fileSystem!));

            //Assert -> validate state and side effects of operation.
            Assert.IsTrue(didUpdateOperationSucceed);
```

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
+        {
+            int currentDepthLimit = 8;
+
+            // Generate Config with Depth Limit
+            RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config);
+            Assert.IsNotNull(config);
+            config = config with
+            {
+                Runtime = config!.Runtime! with
```
seantleonard, line 98:
```suggestion
                Runtime = config.Runtime! with
```
config confirmed to be not null already. does intellisense still complain?

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    GraphQL = config!.Runtime!.GraphQL! with
```
seantleonard, line 100:
```suggestion
                    GraphQL = config.Runtime!.GraphQL! with
```

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            );
+            Assert.IsTrue(TryUpdateRuntimeSettings(options, _runtimeConfigLoader!, _fileSystem!));
+
+            // Assert after adding Depth Limit
+            string updatedConfig = _fileSystem!.File.ReadAllText(_testConfigFile);
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out config));
+            Assert.IsNotNull(config);
+
+            int? expectedDepthLimit = newDepthLimit == -1 ? null : newDepthLimit;
+            Assert.AreEqual(expectedDepthLimit, config!.Runtime!.GraphQL!.DepthLimit);
```
seantleonard, line 122: if user explictly sets depth-limit to -1, that value should be present in the config and written to disk.

src/Cli.Tests/UpdateRuntimeTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            if (expectedDepthLimit is null)
+            {
+                Console.WriteLine(JObject.Parse(INITIAL_CONFIG));
+                Console.WriteLine(JObject.Parse(updatedConfig));
+                Assert.IsTrue(JToken.DeepEquals(JObject.Parse(INITIAL_CONFIG), JObject.Parse(updatedConfig)));
```
seantleonard, line 128: a comment to add a little more detail to what you're validating here would be helpful. perhaps add `message:` param to `Assert.IsTrue` so the dev knows where to start investigating a fix if this test ever fails.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -296,6 +296,34 @@ public void TestAddTelemetry(string? appInsightsEnabled, string appInsightsConnS
         Assert.AreEqual("InstrumentationKey=00000000", updatedConfig.Runtime.Telemetry.ApplicationInsights.ConnectionString);
     }
 
+    /// <summary>
+    /// Test to verify sucess or failure based on different values for
+    /// depth limit in GraphQL runtime settings.
```
seantleonard, line 301: Needs more specificity: this test checks behavior of executing `dab runtime --query-depth {value}`

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -296,6 +296,34 @@ public void TestAddTelemetry(string? appInsightsEnabled, string appInsightsConnS
...
+    {
+        string[] initArgs = { "init", "-c", TEST_RUNTIME_CONFIG_FILE, "--host-mode", "development", "--database-type",
+            "mssql", "--connection-string", TEST_ENV_CONN_STRING };
+        Program.Execute(initArgs, _cliLogger!, _fileSystem!, _runtimeConfigLoader!);
+
+        Assert.IsTrue(_runtimeConfigLoader!.TryLoadConfig(TEST_RUNTIME_CONFIG_FILE, out RuntimeConfig? runtimeConfig));
+        Assert.IsNotNull(runtimeConfig);
+
+        string[] runtimeArgs = { "runtime", "-c", TEST_RUNTIME_CONFIG_FILE, "--depth-limit", depthLimit };
+        int isError = Program.Execute(runtimeArgs, _cliLogger!, _fileSystem!, _runtimeConfigLoader!);
```
seantleonard, line 322: probably worth denoting that this line is your actual test: the `// Act` section.

seantleonard (changes requested):

src/Cli/ConfigGenerator.cs
```diff
@@ -510,6 +510,83 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
...
+        /// A depth limit of -1 is considered a special case that disables the GraphQL depth limit.
+        /// </summary>
+        /// <param name="options">Options including the new depth limit.</param>
+        /// <param name="runtimeConfig">Current config, updated if method succeeds.</param>
+        /// <returns>True if the update was successful, false otherwise.</returns>
+        private static bool TryUpdateDepthLimit(
+            UpdateRuntimeOptions options,
+            [NotNullWhen(true)] ref RuntimeConfig runtimeConfig)
+        {
+            if (options.DepthLimit is not null)
```
aaronburtle, line 550: we've started setting the standard that null values in the config means `default`. In the PR description it mentions that by default we have no depth limit. If someone had set a depth limit to some value, say 3, and then they want to go back to the default value, would their only option be to pass in the value that we have set as the default (currently int.MAX I presume).

Perhaps passing in null here should just go back to default behavior, and then if default behavior is a different value in the future, user doesnt have to know about this detail, just follow the pattern of null = default.

abhishekkumams: if users wants to go back to the default value, i.e, remove current depth limit, they need use the command
`dab runtime --depth-limit -1`.

src/Cli.Tests/TestHelper.cs
```diff
@@ -111,7 +111,8 @@ public static Process ExecuteDabCommand(string command, string flags)
           ""runtime"": {
               ""rest"": {
                   ""path"": ""/api"",
-                  ""enabled"": true
+                  ""enabled"": true,
+                  ""request-body-strict"": true
```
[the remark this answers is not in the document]
abhishekkumams: yes, because generating a new config adds this property to the file by default.

src/Cli.Tests/TestHelper.cs
abhishekkumams: comma is added automatically if the {depthLimitSection} is not empty.

src/Config/ObjectModel/GraphQLRuntimeOptions.cs
```diff
@@ -6,6 +6,7 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 public record GraphQLRuntimeOptions(bool Enabled = true,
                                     string Path = GraphQLRuntimeOptions.DEFAULT_PATH,
                                     bool AllowIntrospection = true,
+                                    int? DepthLimit = null,
```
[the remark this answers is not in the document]
abhishekkumams: yes, it affects mutation input as well.

src/Cli.Tests/UpdateRuntimeTests.cs
abhishekkumams: CLI will give error stating invalid value along with help text on valid values. because --depth-limit accepts integer value.

src/Cli.Tests/ValidateConfigTests.cs
abhishekkumams: Thanks for catching that. Actually the parameter `isSuccess` is not being used and we are only checking for failure. that's why both "20" as well as "-1" failed. "-1" failed because of invalid connection string as these are unit tests so we don't have valid connection string in CLI tests. Once the engine work is done we can have tests in engine which will validate the complete flow.

src/Cli/ConfigGenerator.cs
```diff
@@ -510,6 +510,83 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
...
+        /// Attempts to update the depth limit in the GraphQL runtime settings based on the provided value.
+        /// If a depth limit is provided in the options, it validates the limit to ensure it's an integer and within the valid range of 1 to Int32.MaxValue.
+        /// A depth limit of -1 is considered a special case that disables the GraphQL depth limit.
+        /// </summary>
+        /// <param name="options">Options including the new depth limit.</param>
+        /// <param name="runtimeConfig">Current config, updated if method succeeds.</param>
+        /// <returns>True if the update was successful, false otherwise.</returns>
+        private static bool TryUpdateDepthLimit(
+            UpdateRuntimeOptions options,
+            [NotNullWhen(true)] ref RuntimeConfig runtimeConfig)
```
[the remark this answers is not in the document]
abhishekkumams: without using ref, i don't see it getting updated.

src/Cli/ConfigGenerator.cs
aaronburtle: > if users wants to go back to the default value, i.e, remove current depth limit, they need use the command `dab runtime --depth-limit -1`.

I see, so if in the future we have some default depth limit that is an actual value (not just removing any depth limit), -1 will still be used to go back to that default value?

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            _fileSystem!.AddFile(TEST_RUNTIME_CONFIG_FILE, new MockFileData(config.ToJson()));
+
+            // Act: Update Depth Limit
+            ConfigureOptions options = new(
+                depthLimit: newDepthLimit,
+                config: TEST_RUNTIME_CONFIG_FILE
+            );
+            Assert.IsTrue(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+
```
seantleonard, line 115: nit: `ConfigureOptions options = new( ...` is part of "arrange" since it's setup code.

Act is just `bool configureResult = TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!)`

"Assert" runs all the asserts

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            // Assert: Validate the Depth Limit is added
+            string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out config));
+            Assert.IsNotNull(config.Runtime!.GraphQL!.DepthLimit);
+            Assert.AreEqual(maxDepthLimit, config!.Runtime!.GraphQL!.DepthLimit);
+        }
+
+        /// <summary>
+        /// Test to update the current depth limit for GraphQL and removal the depth limit using -1.
```
seantleonard, line 84: nit: more detail about validation.

---
When runtime.graphql.depth-limit has an initial value of 8, validates that 
"dab configure --runtime.graphql.depth-limit {value}" sets the expected depth limit.
---

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -296,6 +296,36 @@ public void TestAddTelemetry(string? appInsightsEnabled, string appInsightsConnS
...
+    /// <summary>
+    /// This test checks behavior of executing `dab configure --runtime.graphql.depth-limit {value}`.
+    /// Valid values are [1, INT32.MAX_VALUE], and -1 to remove depth limit.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("8", true, DisplayName = "Succesful update with a valid value for depth limit")]
+    [DataRow("0", false, DisplayName = "Failure as depth limit cannot be set to 0.")]
+    [DataRow("-1", true, DisplayName = "Succesful update to to remove depth limit using -1.")]
+    [DataRow("-15", false, DisplayName = "Failure as negative value other than -1 is invalid")]
+    [DataRow("2147483647", true, DisplayName = "Succesful update setting value to INT32_MAX")]
```
seantleonard, line 308:
```suggestion
    [DataRow("2147483647", true, DisplayName = "Successful update setting value to INT32_MAX")]
```

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -296,6 +296,36 @@ public void TestAddTelemetry(string? appInsightsEnabled, string appInsightsConnS
         Assert.AreEqual("InstrumentationKey=00000000", updatedConfig.Runtime.Telemetry.ApplicationInsights.ConnectionString);
     }
 
+    /// <summary>
+    /// This test checks behavior of executing `dab configure --runtime.graphql.depth-limit {value}`.
+    /// Valid values are [1, INT32.MAX_VALUE], and -1 to remove depth limit.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("8", true, DisplayName = "Succesful update with a valid value for depth limit")]
```
seantleonard, line 304:
```suggestion
    [DataRow("8", true, DisplayName = "Successful update with a valid value for depth limit")]
```

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -296,6 +296,36 @@ public void TestAddTelemetry(string? appInsightsEnabled, string appInsightsConnS
...
+    [DataRow("0", false, DisplayName = "Failure as depth limit cannot be set to 0.")]
+    [DataRow("-1", true, DisplayName = "Succesful update to to remove depth limit using -1.")]
```
seantleonard, line 306:
```suggestion
    [DataRow("-1", true, DisplayName = "Successful update to to remove depth limit using -1.")]
```

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -114,6 +114,37 @@ public void TestValidateWithEmptyConfig()
         }
     }
 
+    /// <summary>
+    /// This Test is used to verify that the validate command is able to catch invalid values for the depth-limit property.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("null", true, DisplayName = "Invalid Value: 'null'. Only integer values are allowed.")]
+    [DataRow("20", true, DisplayName = "Invalid Value: '20'. Integer value provided as string are not allowed.")]
```
seantleonard, line 122:
```suggestion
    [DataRow("20", true, DisplayName = "Invalid Value: '20'. Integer values provided as strings are not allowed.")]
```

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -114,6 +114,37 @@ public void TestValidateWithEmptyConfig()
         }
     }
 
+    /// <summary>
+    /// This Test is used to verify that the validate command is able to catch invalid values for the depth-limit property.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("null", true, DisplayName = "Invalid Value: 'null'. Only integer values are allowed.")]
```
seantleonard, line 121: is there a test where you have null without quotes?
```json
{
    "max-depth": null
}
```

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            ILoggerFactory loggerFactory = TestLoggerSupport.ProvisionLoggerFactory();
+
+            SetLoggerForCliConfigGenerator(loggerFactory.CreateLogger<ConfigGenerator>());
+            SetCliUtilsLogger(loggerFactory.CreateLogger<Utils>());
+        }
+
+        /// <summary>
+        /// Validates that `dab configure` is a no-op because the user provided no options to update
+        /// the config file. The config file must not change.
```
seantleonard, line 31: > The config file must not change.

This is valid when the config has `{"max-depth": null}` or `{"max-depth": -1}` . For any other valid integer [1,Int.max], the config will change

seantleonard (approved):
few nits but once cleaned up, looks good!

src/Cli/ConfigGenerator.cs
```diff
@@ -510,6 +510,72 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
...
+        /// <param name="runtimeConfig">Current config, updated if method succeeds.</param>
+        /// <returns>True if the update was successful, false otherwise.</returns>
+        private static bool TryUpdateDepthLimit(
+            ConfigureOptions options,
+            [NotNullWhen(true)] ref RuntimeConfig runtimeConfig)
+        {
+            // check if depth limit is within the valid range of 1 to Int32.MaxValue
+            int? newDepthLimit = options.DepthLimit;
+            if (newDepthLimit < 1)
+            {
```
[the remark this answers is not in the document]
abhishekkumams: No, because it wouldn't come to this statement if options.DepthLimit was null in the first place. because this method is only called when some integer value was specified.

src/Cli.Tests/ValidateConfigTests.cs
abhishekkumams: added a test for that.

src/Cli/ConfigGenerator.cs
```diff
@@ -510,6 +510,72 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
...
+
+        /// <summary>
+        /// Attempts to update the depth limit in the GraphQL runtime settings based on the provided value.
+        /// Validates that any user-provided depth limit is an integer within the valid range of [1 to Int32.MaxValue] or -1.
+        /// A depth limit of -1 is considered a special case that disables the GraphQL depth limit.
+        /// </summary>
+        /// <param name="options">Options including the new depth limit.</param>
+        /// <param name="runtimeConfig">Current config, updated if method succeeds.</param>
+        /// <returns>True if the update was successful, false otherwise.</returns>
+        private static bool TryUpdateDepthLimit(
```
aaronburtle, line 546: What happens if someone calls this function and the options.DepthLimit is null?

abhishekkumams: we already had a check before calling this method, that only calls this when depthLimit is not null. But in future if some other method wants to call it, for them to be aware, i will update the method summary.

src/Cli.Tests/ConfigureOptionsTests.cs
abhishekkumams: thanks, updated.

aaronburtle (approved):
Thanks for the quick responses.

Issue #951, opened 2022-11-04
Automate Package Dependency Updates


seantleonard:
As a follow up to the improvements from the PR addressing #948 and Ani's suggestion, we could add automated dependency upgrades (minor/patch versions to start with).

We can use GitHub's Dependabot if compatible with our private NuGet feed:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates

Local tooling possibilities:

- https://github.com/dotnet-outdated/dotnet-outdated

seantleonard:
This seems to become a need every patch tuesday because the upstream feed needs to be updated with latest dependencies to not fail nuget restore.

PR #2264, opened 2024-06-12
Bump Azure.Identity to 1.11.4

### LINKED ISSUE #2263: Bump Azure.Identity version
Get latest release per GitHub Security Advisory: https://github.com/advisories/GHSA-m5vv-6r4h-3vj9
image

seantleonard:
## Why make this change?

- Closes #2263
- Issue alerted to us via failed Dotnet format (which runs dotnet restore) and the GitHubAdvisory fails the pipeline step.

## What is this change?

- Updates Directory.Packages.props to point to newest version after adding it to the Nuget Feed DAB references.

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -3,7 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <ItemGroup>
-    <PackageVersion Include="Azure.Identity" Version="1.11.2" />
+    <PackageVersion Include="Azure.Identity" Version="1.11.4" />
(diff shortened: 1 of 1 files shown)


aaronburtle (approved):
thanks for the quick fix, are we going to have to keep manually bumping this?

seantleonard:
> thanks for the quick fix, are we going to have to keep manually bumping this?

Item on backlog which would auto-upgrade packages using dependabot: https://github.com/Azure/data-api-builder/issues/951

Would still require PR to change but ideally would be automated. Still a manual process to add the newest package version to our feed though.

abhishekkumams (approved):
Thanks for figuring out and fixing it.

PR #2265, opened 2024-06-13
Fix MySQL test execution time + Fix DateTime GraphQL input handling

seantleonard:
## Why make this change?

I discovered another bug while working on this change #2268, which will be addressed separately.

## What is this change?

- Fixes a bug where GraphQL DateTime input with Timezone (TZ) offset was not honored. 
    - e.g. `12-31-2024T12:00:00+03:00` was stored as `12-31-2024T12:00:00` in the backend database instead of `12-31-2024T09:00:00`
    - This bug was discovered because some SupportedTypes tests involving dates always failed locally but not in the integration test pipeline. My local machine in UTC-8 while pipelines are UTC. Additionally, some test input `9999-12-31 23:59:59.9999999` was run for MySql where that value is an invalid date (max decimal is .499999).
- Fixes false negative (passing) test cases in **GraphQLSupportedTypesTests** by adding more comprehensive asserts, updating DataRow test input, and ensuring tests specific to a database type are correctly executed. 
    - As a result, MySQL tests in this specific test suite run way faster (~minutes down to less than 10 seconds because the conclusion of each data row execution doesn't reset the MySql database contents which takes a long time via `ResetDbStateAsync()`.)
    - Updates test code queries and mutations to include the primary key (typeid) of the table **type_table** so that tests interact with determinate records:
    ```csharp
    string gqlQuery = "{ supportedType_by_pk(typeid: " + id + ") { typeid, " + field + " } }";
    ```

- A bug in the test 
```csharp
// Old test
[DataRow(TIME_TYPE, "\"23:59:59.9999999\"")]

// Fixed Test
[DataRow(TIME_TYPE, "23:59:59")]
public async Task InsertIntoTypeColumnWithArgument(string type, object value)
```
This test shows as green in main branch. However, looking at the log of the test, I saw this error:
>     GraphQL error: [{"message":"The variable `param` is not compatible with the type of the current location.","locations":[{"line":1,"column":65}],"path":["createSupportedType"],"extensions":{"variable":"param","variableType":"Time","locationType":"LocalTime","specifiedBy":"http://spec.graphql.org/October2021/#sec-All-Variable-Usages-are-Allowed"}}]

There are two layers of changes here:
1. TIME_TYPE is technically not a GraphQL data type recognized or defined by DAB's generated GraphQL schema/ HotChocolate -> LocalTime is. So when the data row for TIME_TYPE comes in this code resolves the expected type:
```csharp
        private static string TypeNameToGraphQLType(string typeName)
        {
            return typeName switch
            {
                DATETIMEOFFSET_TYPE => DATETIME_TYPE,
                TIME_TYPE => LOCALTIME_TYPE,
                _ => typeName
            };
        }
```

2. Leaving quotes on the value -> `\"23:59:59.9999999\"` in results in the error: 
>     GraphQL error: [{"message":"Unable to deserialize string to LocalTime","path":["param"]}]

why keep data row as "TIME_TYPE"? well, our test code resolves that type name not only as the GraphQL type but also as a way to determine which column in the data source to select which is named time_types. So we'd need to decouple the GraphQL Data Type value from being used to determine the SQL table column name.

## How was this tested?

- [x] Integration Tests - Fixed tests in GraphQLSupportedTypesTests


src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -215,8 +217,8 @@ public async Task QueryTypeColumnFilterAndOrderBy(string type, string filterOper
             DisplayName = "datetime type filter and orderby test with neq operator and specific value '1999-01-08 10:23:54'.")]
         [DataRow(DATETIMEOFFSET_TYPE, "neq", "'1999-01-08 10:23:54.9999999-14:00'", "\"1999-01-08 10:23:54.9999999-14:00\"", "!=",
             DisplayName = "datetimeoffset type filter and orderby test with neq operator")]
-        [DataRow(DATETIMEOFFSET_TYPE, "lt", "'9999-12-31 23:59:59.9999999'", "\"9999-12-31 23:59:59.9999999\"", "<",
-            DisplayName = "datetimeoffset type filter and orderby test with lt operator and max value for datetimeoffset.")]
+        [DataRow(DATETIMEOFFSET_TYPE, "lt", "'9999-12-31 23:59:59.9999999'", "\"9999-12-31 23:59:59.9999999Z\"", "<",
+            DisplayName = "datetimeoffset type filter and orderby test with lt operator and max value (with UTC offset specified) for datetimeoffset.")]
```
aaronburtle, line 221: Maybe im misunderstanding what you mean by UTC offset specified, what is the UTC offset for this data row?

seantleonard: Z -> +00:00 UTC

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -426,8 +461,7 @@ public async Task InsertInvalidTimeIntoTimeTypeColumn(string type, string value)
         [DataRow(BOOLEAN_TYPE, true)]
         [DataRow(DATETIMEOFFSET_TYPE, "1999-01-08 10:23:54+8:00")]
         [DataRow(DATETIME_TYPE, "1999-01-08 10:23:54")]
-        [DataRow(TIME_TYPE, "\"23:59:59.9999999\"")]
-        [DataRow(TIME_TYPE, "null")]
+        [DataRow(TIME_TYPE, "23:59:59")]
```
abhishekkumams, line 464: why this change?

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/MySqlGQLSupportedTypesTests.cs
```diff
@@ -24,27 +26,135 @@ public static async Task SetupAsync(TestContext context)
...
+        {
+            // Arrange
+            expectedResult = DateTime.UtcNow.ToString("yyyy-MM-ddT") + expectedResult;
+            await InsertMutationInput_DateTimeTypes_ValidRange_ReturnsExpectedValues(dateTimeGraphQLInput, expectedResult);
+        }
+
+        /// <summary>
+        /// MySql Single Type Tests.
+        /// </summary>
+        /// <param name="type">GraphQL Type</param>
```
abhishekkumams, line 115: maybe GraphQL DataType, to not get confused with GraphQL request type:query/mutation

ayush3797 (approved):
Neat! LGTM!

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
seantleonard: This test shows as green in main branch. However, looking at the log of the test, I saw this error:
>     GraphQL error: [{"message":"The variable `param` is not compatible with the type of the current location.","locations":[{"line":1,"column":65}],"path":["createSupportedType"],"extensions":{"variable":"param","variableType":"Time","locationType":"LocalTime","specifiedBy":"http://spec.graphql.org/October2021/#sec-All-Variable-Usages-are-Allowed"}}]

There are two layers of changes here:
1. TIME_TYPE is technically not a GraphQL data type recognized or defined by DAB's generated GraphQL schema/ HotChocolate -> LocalTime is. So when the data row for TIME_TYPE comes in this code resolves the expected type:
```csharp
        private static string TypeNameToGraphQLType(string typeName)
        {
            return typeName switch
            {
                DATETIMEOFFSET_TYPE => DATETIME_TYPE,
                TIME_TYPE => LOCALTIME_TYPE,
                _ => typeName
            };
        }
```
2. Leaving quotes on the value -> `\"23:59:59.9999999\"` in results in the error: 
>     GraphQL error: [{"message":"Unable to deserialize string to LocalTime","path":["param"]}]

seantleonard: why keep data row as "TIME_TYPE"? well, our test code resolves that type name not only as the GraphQL type but also as a way to determine which column in the data source to select which is named time_types. So we'd need to decouple the GraphQL Data Type value from being used to determine the SQL table column name.

seantleonard: added to pr description.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/MySqlGQLSupportedTypesTests.cs
seantleonard: updated

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
abhishekkumams: makes sense

abhishekkumams (approved):

Aniruddh25 (commented):
Comments so far

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
```diff
@@ -565,7 +595,7 @@ public async Task UpdateTypeColumnWithArgument(string type, object value)
         /// Utility function to do special comparisons for some of the extended types
         /// if json compare doesn't suffice
         /// </summary>
-        private static void PerformTestEqualsForExtendedTypes(string type, string expected, string actual)
+        public static void PerformTestEqualsForExtendedTypes(string type, string expected, string actual)
```
Aniruddh25, line 598: Do we need `public` or is `protected` sufficient?

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/PostgreSqlGQLSupportedTypesTests.cs
```diff
@@ -24,13 +24,38 @@ public static async Task SetupAsync(TestContext context)
             await InitializeTestFixture();
         }
 
-        protected override string MakeQueryOnTypeTable(List<string> columnsToQuery, int id)
+        /// <summary>
+        /// Postgres requires conversion of 0.33 to 'real' type otherwise precision is lost.
```
Aniruddh25, line 28: This seems like a `Note` to me as opposed to a summary of this function. Why are we talking specifically about 0.33?

Aniruddh25 (approved):
Left few questions. But approved otherwise.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/GraphQLSupportedTypesTestsBase.cs
abhishekkumams: yes, protected can be used.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/PostgreSqlGQLSupportedTypesTests.cs
abhishekkumams: good point, I think 0.33 is just an example, but not clear why the conversion was not needed for the last test

abhishekkumams: I added the conversion for the last case as well.

PR #2283, opened 2024-06-27
[OpenAPI] Add "type" property with value "object" response schema to fix Infragistics AppBuilder tooling

### LINKED ISSUE #2212: Support for OpenAPI description in a Infragistics AppBuilder integration use case
### What happened?

It seems that DAB does not generate enough information when an endpoint is called by Infragistics AppBuilder that requires some details to be able to use the endpoint correctly, see the description [here](https://www.appbuilder.dev/help/using-data-in-your-app#openapi-swagger-endpoints-are-grayed-out). I searched for possibilities to configure the expressivness of DABs openapi interface but did not find any information about such a possibility. In a Infragistics forum post, see [here](https://www.infragistics.com/community/forums/f/indigo-design/125080/appbuilder-when-adding-rest-api-the-endpoints-are-disabled-and-not-selectable/549126#549126), the problem is described and solved with access to the underlying endpoint source code, i don't feel confident to go to the source code of DAB to check this. Thanks for your time in helping me out with this topic.

### Version

latest

Azure SQL

Local (including CLI)

REST

### Relevant log output

_No response_

### Code of Conduct

seantleonard:
## Why make this change?

- Closes #2212 which describes how the missing `"type":"object"` key/value pair on the response child object schema breaks certain client tooling. in this case: Infragistics AppBuilder. 

### Background

I found a relevant thread that discusses whether type is a required property. Consensus is that type isn't required:

https://github.com/OAI/OpenAPI-Specification/issues/1657
OpenAPI-Specification discussion
https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/issues/430
Example of how different tooling handles type or missing type differently.
Ultimately, different tooling handles the presence of the type property differently. Some may try to guess the type when not present:

The fact that the type isn't required means that  https://github.com/microsoft/OpenAPI.NET didn't complain about missing type. An error if type were required would have helped prevent this becoming an issue in the first place.

## What is this change?

- Adds `"type": "object"` to the openapi document for describing the response schema:
```json
                "responses": {
                    "200": {
                        "description": "OK",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "object", // <--- This property/value
                                    "properties": {
                                        "value": {
                                            "type": "array",
                                            "items": {
                                                "$ref": "#/components/schemas/Book"
                                            }
                                        },
                                        "nextLink": {
                                            "type": "string"
                                        }
                                    }
                                }
                            }
                        }
                    }
```

## How was this tested?

- [x] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

View generated schema at
```https
GET localhost:5001/api/openapi
```


src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -937,6 +937,7 @@ private static OpenApiMediaType CreateResponseContainer(string responseObjectSch
             {
                 Schema = new()
                 {
+                    Type = SCHEMA_OBJECT_TYPE,
```
Aniruddh25, line 940: will anything else might help the Type property?

seantleonard: We'd need an instance of Infragistics AppBuilder to test. But that is also whack-a-mole because different openapi client tooling sets different expectations on vague areas of the openapi spec. See referenced issues in PR description for examples of that point.

Aniruddh25 (approved):
LGTM

PR #2285, opened 2024-06-28
Update the retry policy for tsql query execution

vadeveka:
## Changes
1. Updates retry policy for QueryExecutor to attempt 3 times instead of 5 times as per guidance
2. Remove sql error number 18456 from the list of transient exceptions which would result in retries. 18456 is a fatal error.

## Why make this change?
Sql Error 18456 occurs when using invalid authentication methods such as invalid audience or expired token or credentials. There is a permanent error hence no benefit from retrying. 
Ref: https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/mssqlserver-18456-database-engine-error?view=sql-server-ver16
    
Additionally, the number of retries of 5 means that it would take considerable for user to see the error returned. This does not bode well for GraphQL queries which are an interactive scenario.     
Ref: https://learn.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#retry-usage-guidance-3

## How was this tested?
- Manual testing to ensure retry attempts down to 2 down from 5 via breakpoints
- Manual testing to reproduce error 18456 and ensure no retries via breakpoints
- Unit tests

### DIFF
--- a/src/Core/Resolvers/MsSqlDbExceptionParser.cs
+++ b/src/Core/Resolvers/MsSqlDbExceptionParser.cs
@@ -48,11 +48,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 
                 // These errors mainly occur when the SQL Server client can't connect to the server.
                 // This may happen when the client cannot resolve the name of the server or the name of the server is incorrect.
-                "53", "11001",
-
-                // Transient error codes compiled from:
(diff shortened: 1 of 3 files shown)


rusamant (approved):

abhishekkumams (approved):

abhishekkumams:
@vadeveka , There is a test which checks on the maxRetries, please update that. `TestRetryPolicyExhaustingMaxAttempts`

PR #2288, opened 2024-07-08
CosmosDB : GQL Schema Generation with Sampling

sourabh1007:
Refer this document for details 
[Schema Inference Design.docx](https://github.com/user-attachments/files/16318665/Schema.Inference.Design.docx)

## What is this change?
Add a utility to generate schema for No SQL database. It will be based on best efforts.

### Main Code
**Samplers:**
- [x] src/Core/Generator/Sampler/PartitionBasedSampler.cs
- [x] src/Core/Generator/Sampler/TimeBasedSampler.cs
- [x] src/Core/Generator/Sampler/TopNSampler.cs

**Schema Generator**
- [x] src/Core/Generator/SchemaGenerator.cs : Generates GQL out of given set of JSON array
  a) Add Alias name
  b) tag container entity or an entity with alias as `@model`
  c) Mark an attribute is nullable or not i.e. `!`

**Utility Classes**
- [x] src/Core/Generator/Sampler/CosmosExecutor.cs : It is responsible to run a CosmosDB Query
- [x] src/Core/Generator/SchemaGeneratorFactory.cs : 
       a) Create connection with Cosmos DB 
       b) Runs Required Sampler
       c) Generates Schema

**Export Command**
- [x] src/Cli/Commands/ExportOptions.cs
- [x] src/Cli/Exporter.cs

### Test Coverage
- [x] Samplers: src/Service.Tests/CosmosTests/SamplerTests.cs
   a) Might notice that, I am creating item with the gap of 1 sec, as Sampler queries are using `_ts` and it is auto generated by cosmosDB (and there is no way to control it). So, waiting time is added to get the different values for this column
- [x] SchemaGenerator: src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs

### Other Changes
- [x] src/Cli/CustomLoggerProvider.cs : Minimal Logger level to `Debug` as I added few debug logs to get more insights (if required) on, what is happening in sampler.

## How can you test it?
1. Go to this location, here you will find the data builder executable: 

image

2. `--help` provides all the available options

image

3.  Below is the minimal command required to run this feature

**TopNSampler**
image

**PartitionBasedSampler**
image

**TimeBasedSampler**
image

## User Journey
1. If customer doesn't have schema file and wants to use dab to generate it.
  a) `dab init` : Initialize DAB without schema file
       image
  b)  `dab add`: Add an entity
       image
  c)  `dab export --generate`
       image
  d) `dab start`
       image

2. If customer have schema file, then dab can be initialized with schema file, as before.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests


abhishekkumams:
Please update the PR description

src/Core/Generator/GraphQlSchemaGenerator.cs
```diff
@@ -0,0 +1,21 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Azure.DataApiBuilder.Core.Generator
+{
+    internal class GraphQLSchemaGenerator
+    {
+        public static async Task<string> Generate(ISchemaGeneratorSampler sampler, Container container)
+        {
+            // Get Sample Data
+            JArray dataArray = await sampler.GetSampleAsync(container);
```
seantleonard, line 15: Why not System.Text.Json?

seantleonard:
just curious since i haven't fully looked through: make sure this feature is feature flagged and doesn't just execute by default

sourabh1007:
> just curious since i haven't fully looked through: make sure this feature is feature flagged and doesn't just execute by default

this feature is not part of DAB flow. Customer has to run it explicitly and generate the gql.

src/Core/Generator/GraphQlSchemaGenerator.cs
sourabh1007: As we discussed offline, removed newtonsoft dependency

seantleonard (commented):
Still need more time to review. The schema generation feature could have been introduced gradually to limit the size of the PRs. 
- Introduce CLI options changes (they may remain a no-op)
- Introduce 1 sampler per PR with associated tests. (CLI schema generation function can still be no-op).
- Introduce schema generator/factory and tests.

src/Cli/Commands/ExportOptions.cs
```diff
@@ -21,7 +30,26 @@ public ExportOptions(bool graphql, string outputDirectory, string? config, strin
...
         public string OutputDirectory { get; }
 
-        [Option('g', "graphql-schema-file", HelpText = "The GraphQL schema file name (default schema.graphql)")]
+        [Option('g', "graphql-schema-file", HelpText = "The GraphQL schema file name (default schema.graphQL)")]
         public string GraphQLSchemaFile { get; }
+
+        [Option("generate", HelpText = "To generate schema file from CosmosDB database")]
+        public bool Generate { get; }
+
+        [Option("sampling-mode", HelpText = "Sampling Modes: TopNSampler, PartitionBasedSampler, TimeBasedSampler")]
```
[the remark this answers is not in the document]
sourabh1007: I didn't give any shortcut intentionally, but I don't mind providing one.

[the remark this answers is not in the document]
sourabh1007: made the changes. provided `m` as shortcut.

src/Cli/CustomLoggerProvider.cs
```diff
@@ -19,7 +19,7 @@ public ILogger CreateLogger(string categoryName)
     public class CustomConsoleLogger : ILogger
     {
         // Minimum LogLevel. LogLevel below this would be disabled.
-        private readonly LogLevel _minimumLogLevel = LogLevel.Information;
+        private readonly LogLevel _minimumLogLevel = LogLevel.Debug;
```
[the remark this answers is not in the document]
sourabh1007: In this feature, the minimum log level is set to debug due to its complexity. While this feature requires debug level logging for thorough monitoring and diagnostics, customers can configure the log level as needed if they encounter issues with the volume of debug logs.
I don't think, it is a breaking change functionally, it is one time run command and customer would see extra logs (if we are logging with debug flag which we are not I believe)

src/Core/Generator/Sampler/CosmosExecutor.cs
```diff
@@ -0,0 +1,98 @@
...
+        private ILogger _logger;
+
+        public CosmosExecutor(Container container, ILogger logger)
+        {
+            this._container = container;
+            this._logger = logger;
+        }
+
+        /// <summary>
+        /// This function execute the passed query and returns the result in the given type.
```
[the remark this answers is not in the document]
sourabh1007: Rewrite the documentation.

src/Cli/Commands/ExportOptions.cs
```diff
@@ -21,7 +30,26 @@ public ExportOptions(bool graphql, string outputDirectory, string? config, strin
...
+        public string SamplingMode { get; } = SamplingModes.TopNSampler.ToString();
+
+        [Option('n', "sampling-count", HelpText = "Sampling Count, For TopNSampler: Total number of Samples, PartitionBasedSampler: Total number of Samples from each partition, TimeBasedSampler : Total number of sample in each time range group.")]
+        public int? NumberOfRecords { get; }
+
+        [Option("sampling-partitionKeyPath", HelpText = "Applicable only when 'PartitionBasedSampler' is selected")]
+        public string? PartitionKeyPath { get; }
+
+        [Option('d', "sampling-days", HelpText = "Data should be fetched for number of days, TopNSampling: filter on number of days, PartitionBasedSampler: Data fetched for a number of days from each partition, TimeBasedSampler: Decide the data range which will be divided into subranges.")]
```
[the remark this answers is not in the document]
sourabh1007: Rewrite it. Let me know if makes sense.

src/Cli/Exporter.cs
```diff
@@ -31,26 +32,34 @@ public static int Export(ExportOptions options, ILogger logger, FileSystemRuntim
                     out RuntimeConfig? runtimeConfig,
                     replaceEnvVar: true) || runtimeConfig is null)
             {
-                logger.LogError("Failed to read the config file: {runtimeConfigFile}.", runtimeConfigFile);
+                logger.LogError($"Failed to read the config file: {runtimeConfigFile}.");
```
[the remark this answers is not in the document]
sourabh1007: Just curious, what is the benefit of it over other?

src/Cli/Exporter.cs
```diff
@@ -31,26 +32,34 @@ public static int Export(ExportOptions options, ILogger logger, FileSystemRuntim
...
             {
-                int retryCount = 5;
                 int tries = 0;
 
                 while (tries < retryCount)
                 {
                     try
                     {
-                        ExportGraphQL(options, runtimeConfig, fileSystem);
+                        ExportGraphQL(options, runtimeConfig, fileSystem, logger).Wait();
```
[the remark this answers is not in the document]
sourabh1007: There are only 2 conditions:
1. ExportGraphQL() => it clearly says which logic to run, to generate/fetch the schema
2. Export constructor => it has a condition, where we are starting graphql serve if generation is not requested.

I don't think, it is going to be more complex in future. If it does, we can definitely refactor it. what do you think?

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
+            PartitionKeyPath = partitionKeyPath;
+            MaxDays = maxDays;
+            GroupCount = groupCount;
         }
 
         [Option("graphql", HelpText = "Export GraphQL schema")]
         public bool GraphQL { get; }
 
-        [Option('o', "output", HelpText = "Directory to save to", Required = true)]
+        [Option('o', "output", HelpText = "Specify the directory where the schema file will be saved. This option is required.", Required = true)]
```
sajeetharan, line 30: change it as "specify the directory where the schema file will be saved. This option is required"

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
         public string OutputDirectory { get; }
 
-        [Option('g', "graphql-schema-file", HelpText = "The GraphQL schema file name (default schema.graphql)")]
+        [Option('g', "graphql-schema-file", HelpText = "Specify the filename for the exported GraphQL schema. Default is 'schema.gql'.")]
```
sajeetharan, line 33: Specify the filename for the exported GraphQL schema (default is:).

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
         public string GraphQLSchemaFile { get; }
+
+        [Option("generate", HelpText = "Generates a schema file from the specified CosmosDB database.")]
+        public bool Generate { get; }
+
+        [Option('m', "sampling-mode", HelpText = "Specifies the sampling mode to use. Available modes include: " +
+                                                 "- TopNSampler: It retrieves a specified number of recent records from a Cosmos DB container, optionally filtering by a maximum number of days." +
```
sajeetharan, line 40: change Cosmos DB as Azure Cosmos DB everywhere in the messaging, its a guidance from our marketing

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
+                                                 "- PartitionBasedSampler: It retrieves a specified number of records from a Cosmos DB container by fetching records from each partition based on a specified partition key.The sampling is configurable by the number of records per partition and the time range considered." +
```
sajeetharan, line 41: It retrieves a specified number of records from an Azure Cosmos DB container by fetching records from each partition using a given partition key. The number of records per partition and the time range are configurable.

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
+                                                 "- TimeBasedSampler:.It retrieves a specified number of records by dividing the container data, time range into subranges and retrieving the top N records from each subrange based on a specified configuration")]
```
sajeetharan, line 42: It retrieves a specified number of records by dividing the container data and time range into subranges, then selecting the top N records from each subrange based on a given configuration.

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
+        public string SamplingMode { get; } = SamplingModes.TopNSampler.ToString();
+
+        [Option('n', "sampling-count", HelpText = "Specify the total number of samples to be retrieved. For different sampling modes:\n" +
```
sajeetharan, line 45: Specify the total number of samples to retrieve for each sampling mode.

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
+                                                  "- TopNSampler: Total number of records to select.\n" +
+                                                  "- PartitionBasedSampler: Number of records to retrieve from each partition.\n" +
+                                                  "- TimeBasedSampler: Number of records to retrieve per time range group.")]
```
sajeetharan, line 48: I am not sure customers would be exposed to this , because we don't expose physical partition to customers . can't we abstract and have a default value?

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
+        public int? NumberOfRecords { get; }
+
+        [Option("sampling-partition-key-path", HelpText = "Specify the partition key path. This option is applicable only when the 'PartitionBasedSampler' mode is selected.")]
+        public string? PartitionKeyPath { get; }
+
+        [Option('d', "sampling-days", HelpText = "Specify the number of days for which data should be fetched. Different sampling modes include: \n" +
```
sajeetharan, line 54: Specify the number of days to fetch data. Sampling modes include:

src/Cli/Commands/ExportOptions.cs
sourabh1007: Here, we are talking about logical partition.

src/Cli/Exporter.cs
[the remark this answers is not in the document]
seantleonard: To clarify my intent with the comment, you currently have
- `Export(...)` Top Level CLI verb
  - `ExportGraphQL(options, ...)` Helper function which can do both export from CosmosDB AND export from DAB service

It would be cleaner to have separate helper to handle only CosmosDB export:
- `Export(...)` Top Level CLI verb
  - `private ExportGraphQLFromCosmosDB(options, ...)`
  - `private ExportGraphQLFromDabService(options, ...)`

src/Core/Generator/Sampler/CosmosExecutor.cs
```diff
@@ -0,0 +1,114 @@
...
+            while (queryIterator.HasMoreResults)
+            {
+                ResponseMessage item = await queryIterator.ReadNextAsync();
+
+                if (item.IsSuccessStatusCode)
+                {
+                    using StreamReader sr = new(item.Content);
+                    string content = await sr.ReadToEndAsync();
+
+                    JsonDocument jsonDocument = JsonDocument.Parse(content);
```
seantleonard, line 55: JsonDocument requires `using` statement

src/Core/Generator/Sampler/CosmosExecutor.cs
```diff
@@ -0,0 +1,114 @@
...
+                        }
+                    }
+                    else
+                    {
+                        Process(callback, dataArray, root);
+                    }
+                }
+                else
+                {
+                    throw new Exception(string.Format("Failed to execute query: {0} with status code {1}, Error Message : {2}", query, item.StatusCode, item.ErrorMessage));
```
seantleonard, line 72: where is this caught? The code which references this function doesn't use try/catch. Is it helpful to print the whole exception to the console for the user? I suppose just the error message you're creating here is sufficient.

src/Core/Generator/Sampler/PartitionBasedSampler.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        /// <summary>
+        /// Retrieves unique partition key values from the Azure Cosmos DB container.
+        /// </summary>
+        /// <param name="partitionKeyPaths">A list of partition key paths used to query for unique values.</param>
+        /// <returns>A list of JsonDocument objects representing unique partition key values.</returns>
+        internal async Task<List<JsonDocument>> GetUniquePartitionKeyValues(List<string> partitionKeyPaths)
+        {
+            List<JsonDocument> uniquePartitionKeyValues = new();
```
seantleonard, line 109: is this line needed? it's unused.

src/Core/Generator/Sampler/TimeBasedSampler.cs
```diff
@@ -0,0 +1,134 @@
...
+            List<long> maxTimestamp = await this._cosmosExecutor.ExecuteQueryAsync<long>(MAX_TIMESTAMP_QUERY);
+            List<long> minTimestamp = new(capacity: 1);
+            if (_maxDays > 0)
+            {
+                // Calculate the timestamp threshold for the timespan
+                minTimestamp.Add(GetTimeStampThreshold());
+            }
+            else
+            {
+                minTimestamp = await this._cosmosExecutor.ExecuteQueryAsync<long>(MIN_TIMESTAMP_QUERY);
```
seantleonard, line 88: Sorry , i might be missing something, how do you guarantee that this query execution will only ever have one result to fit into : `List<long> minTimestamp = new(capacity: 1);` while you don't have the same capacity restrictions for `maxTimestamp`

src/Core/Generator/Sampler/TimeBasedSampler.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            return (minTimestamp[0], maxTimestamp[0]);
```
seantleonard, line 91: why `maxTimestamp[0]` and not any other entry?

src/Core/Generator/Sampler/TimeBasedSampler.cs
```diff
@@ -0,0 +1,134 @@
...
+            return await GetDataFromSubranges(minTimestamp, maxTimestamp, _groupCount, _numberOfRecordsPerGroup);
+        }
+
+        /// <summary>
+        /// Fetches the minimum and maximum timestamps from the data within the specified time range.
+        /// </summary>
+        /// <returns>A tuple containing the minimum and maximum timestamps.</returns>
+        private async Task<(long minTimestamp, long maxTimestamp)> GetHighestAndLowestTimestampsAsync()
+        {
+            List<long> maxTimestamp = await this._cosmosExecutor.ExecuteQueryAsync<long>(MAX_TIMESTAMP_QUERY);
```
seantleonard, line 79: why would this query return a list of values instead of just one value?

src/Core/Generator/SchemaGeneratorFactory.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            logger.LogInformation("Sampling Started using {0}", schemaGeneratorSampler.GetType().Name);
+
+            // Get sample data from the selected sampler.
+            List<JsonDocument> dataArray = await schemaGeneratorSampler.GetSampleAsync();
+
+            logger.LogInformation("{0} records collected as Sample", dataArray.Count);
+            if (dataArray.Count == 0)
+            {
+                logger.LogError("No data got sampled out. Please try different sampling Mode or Sampling configuration");
```
seantleonard, line 91:
```suggestion
                logger.LogError("No data was sampled. Please try a different sampling mode or provide different sampling options.");
```

src/Core/Generator/SchemaGeneratorFactory.cs
```diff
@@ -0,0 +1,114 @@
...
+                throw new ArgumentException("No data got sampled out. Please try different sampling Mode or Sampling configuration");
+            }
+
+            logger.LogInformation("Generating Schema Started");
```
seantleonard, line 95:
```suggestion
            logger.LogInformation("GraphQL schema generation started.");
```

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -0,0 +1,336 @@
...
+        /// <param name="containerName">The name of the Azure Cosmos DB container.</param>
+        /// <param name="config">Optional configuration that maps GraphQL entity names to their singular forms.</param>
+        /// <returns>A string representing the generated GraphQL schema.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if the JSON data is empty or the container name is null or empty.</exception>
+        public static string Generate(List<JsonDocument> jsonData, string containerName, RuntimeConfig? config = null)
+        {
+            // Validate input parameters.
+            if (jsonData == null || jsonData.Count == 0 || string.IsNullOrEmpty(containerName))
+            {
+                throw new InvalidOperationException("JArray must contain at least one JSON object and Container Name can not be blank");
```
seantleonard, line 66: not using newtonsoft

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -0,0 +1,336 @@
...
+                sb.Append($"type {entity.Key} ");
+
+                // Append model directive if applicable.
+                if (_entityAndSingularNameMapping.ContainsKey(entity.Key) && _entityAndSingularNameMapping[entity.Key] != entity.Key)
+                {
+                    sb.Append($"@model(name:\"{_entityAndSingularNameMapping[entity.Key]}\") ");
+                }
+                else if (isRoot)
+                {
+                    sb.Append("@model ");
```
seantleonard, line 121: I though we no longer allow `@model(name=)` without a name value. Is that not accurate?

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -0,0 +1,336 @@
...
+                        }
+
+                        break;
+                    case JsonValueKind.True:
+                    case JsonValueKind.False:
+                        gqlFieldType = "Boolean";
+                        break;
+                    case JsonValueKind.Null or JsonValueKind.String:
+                        if (DateTime.TryParse(token.GetString(), DateTimeFormatInfo.InvariantInfo, DateTimeStyles.None, out DateTime _))
+                        {
```
seantleonard, line 240: wouldn't this error out if this case is `JsonValueKind.Null` and then you try token.GetString() ?

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -0,0 +1,213 @@
...
+            CosmosClientProvider cosmosClientProvider = _application.Services.GetService<CosmosClientProvider>();
+            CosmosClient cosmosClient = cosmosClientProvider.Clients[cosmosClientProvider.RuntimeConfigProvider.GetConfig().DefaultDataSourceName];
+            _database = await cosmosClient.CreateDatabaseIfNotExistsAsync(DATABASE_NAME);
+
+            _containerWithIdPk = await _database.CreateContainerIfNotExistsAsync(CONTAINER_NAME_ID_PK, "/id");
+            _containerWithNamePk = await _database.CreateContainerIfNotExistsAsync(CONTAINER_NAME_NAME_PK, "/name");
+
+            // Insert sample items with a delay to ensure unique _ts values for each item.
+            // This delay allows for accurate timestamp-based sampling in the test cases.
+            CreateItems(DATABASE_NAME, CONTAINER_NAME_ID_PK, 10, waitInMs: 1000);
```
seantleonard, line 59: For test consistency, should you also clear out records from container when the container already exists? 
repro:
1. run test, executes setup, then tests
2. stop process, cleanup doesn't run.

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -0,0 +1,213 @@
...
+            Assert.AreEqual("id", result[0]);
+
+            // Check if partition key path is getting fetched correctly if customer has multiple partitions
+            Container _containerWithHPk = await _database.CreateContainerIfNotExistsAsync("newcontainerwithhpk", "/anotherPojo/anotherProp");
+            partitionBasedSampler = new(_containerWithHPk, null, 1, 1, _mockLogger.Object);
+            result = await partitionBasedSampler.Object.GetPartitionKeyPaths();
+            Assert.AreEqual("anotherPojo", result[0]);
+            Assert.AreEqual("anotherProp", result[1]);
+
+            await _containerWithHPk.DeleteContainerAsync();
```
seantleonard, line 165: seems like three different tests testing different things. (yes, same overall goal of testing partitionbasedsampler) HOwever, these are unique tests for the partition based sampler. I'm thinking of it like this: if one of the tests falis, what does that indicate is wrong with the sampler? Most likely something wrong that is different than what the other few tests are checking.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -0,0 +1,288 @@
...
+        /// Tests the <see cref="SchemaGenerator.Generate"/> method using a single JSON file to generate a GraphQL schema.
+        /// Verifies that the generated schema matches the expected schema output.
+        /// </summary>
+        /// <param name="jsonFilePath">Path to the directory containing the input JSON file.</param>
+        /// <param name="gqlFilePath">Path to the directory containing the expected GraphQL schema file.</param>
+        [TestMethod]
+        [DataRow("CosmosTests/TestData/CosmosData", "CosmosTests/TestData/GeneratedGqlSchema")]
+        public void TestSchemaGenerator(string jsonFilePath, string gqlFilePath)
+        {
+            string json = Regex.Replace(File.ReadAllText($"{jsonFilePath}/EmulatorData.json", Encoding.UTF8), @"\s+", string.Empty);
```
seantleonard, line 36: what's this doing/why is it needed?

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -0,0 +1,288 @@
...
+            List<JsonDocument> jsonArray = new() { JsonSerializer.Deserialize<JsonDocument>(json) };
+
+            string actualSchema = SchemaGenerator.Generate(jsonArray, "containerName");
+            string expectedSchema = File.ReadAllText($"{gqlFilePath}/EmulatorData.gql");
+
+            AreEqualAfterCleanup(expectedSchema, actualSchema);
+        }
+
+        /// <summary>
+        /// Tests the <see cref="SchemaGenerator.Generate"/> method using multiple JSON files to generate a GraphQL schema.
```
seantleonard, line 46: what do you mean by multiple json files? multiple DAB runtime configs? multiple json structures in cosmosdb?

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -0,0 +1,288 @@
...
+                    23456,
+                    34567,
+                    45678
+                  ]
+                }
+                ")};
+
+            string gqlSchema = SchemaGenerator.Generate(jsonArray, "containerName");
+
+            string expectedSchema = @"type ContainerName @model {
```
seantleonard, line 116: `@model(name=)` without name?

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        /// <summary>
+        /// Tests the <see cref="SchemaGenerator.Generate"/> method with a JSON array containing mixed data types.
+        /// Verifies that the generated GraphQL schema includes all possible fields and types present in the input data.
+        /// </summary>
+        [TestMethod]
+        public void TestMixedJsonArray()
+        {
+            List<JsonDocument> jsonArray = new() {
+                JsonDocument.Parse(@"{ ""name"": ""John"", ""age"": 30, ""isStudent"": false, ""birthDate"": ""1980-01-01T00:00:00Z"" }"),
```
seantleonard, line 208: what happens when birthday is null? i recognized in other code that there is a case for attempting to parse a null value as a date.

src/Cli/CustomLoggerProvider.cs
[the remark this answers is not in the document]
abhishekkumams: This will start printing more information to the CLi and won't be very user friendly.

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
+                                                 "- PartitionBasedSampler: It retrieves a specified number of records from an Azure Cosmos DB container by fetching records from each partition using a given partition key.The number of records per partition and the time range are configurable." +
+                                                 "- TimeBasedSampler:.It retrieves a specified number of records by dividing the container data and time range into subranges, then selecting the top N records from each subrange based on a given configuration.")]
+        public string SamplingMode { get; } = SamplingModes.TopNSampler.ToString();
+
+        [Option('n', "sampling-count", HelpText = "Specify the total number of samples to retrieve for each sampling modes:\n" +
+                                                  "- TopNSampler: Total number of records to select.\n" +
+                                                  "- PartitionBasedSampler: Number of records to retrieve from each partition.\n" +
+                                                  "- TimeBasedSampler: Number of records to retrieve per time range group.")]
+        public int? NumberOfRecords { get; }
+
```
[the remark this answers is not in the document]
sourabh1007: It tells significance of this config in that sampler. It is not about Sampler.

sajeetharan:
@sourabh1007 can you add samples for all 3 modes, I have few feedbacks on the cli command variables and also on the PR, will add once you have the samples in PR

src/Core/Generator/Sampler/TimeBasedSampler.cs
sourabh1007: There must be only 1 max time and 1 min time, Hence, it is guaranteed that query would return, only single result. it should fail otherwise.
I didn't put any capacity restriction on max timestamp because result is coming from the cosmosdb query only. So just returned the same.

src/Core/Generator/Sampler/TimeBasedSampler.cs
sourabh1007: can you please elaborate your question? Here, I am trying to find out the highest range of data.

src/Core/Generator/Sampler/TimeBasedSampler.cs
sourabh1007: `ExecuteQueryAsync` is generic function which returns list of data.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
sourabh1007: removed...i added it during testing.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
sourabh1007: imitating multiple records in cosmos DB. In Cosmos DB, each record can have different schema.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
sourabh1007: it won't fail, it will go to else and consider it as `string`.

sajeetharan:
@sourabh1007 What is the sequence of commands here? I see that all the sampling methods require config file as the input, however dab init which is the initial command to generate the config, which cannot be executed without a schema file.

sourabh1007:
> @sourabh1007 What is the sequence of commands here? I see that all the sampling methods require config file as the input, however dab init which is the initial command to generate the config, which cannot be executed without a schema file.

Generation of schema file is very first command (if required) then, after that it is normal flow.

src/Cli.Tests/InitTests.cs
```diff
@@ -164,8 +164,8 @@ bool expectSuccess
         /// </summary>
         [DataRow(null, "testcontainer", "", false, DisplayName = "Both database and schema are either null or empty.")]
         [DataRow("", "testcontainer", "testschema", false, DisplayName = "database is empty.")]
-        [DataRow("testDatabase", "testcontainer", "", false, DisplayName = "database is provided, Schema is null.")]
-        [DataRow("testDatabase", null, "", false, DisplayName = "database is provided, container and Schema is null/empty.")]
+        [DataRow("testDatabase", "testcontainer", "", true, DisplayName = "database is provided, Schema is null.")]
```
abhishekkumams, line 167: Previously, we used to error out if schema file and database name both are not provided for cosmosDb_NoSQL.  I checked the logic in the ConfigGenerator.cs file in your PR. Shouldn't it still be false?

src/Cli/CustomLoggerProvider.cs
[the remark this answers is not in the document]
abhishekkumams: I See that in the PR description the screenshot, i see SQL statements coming up. This is too much information for user. Can this not be done without changing the minimal value? If it's only required for debugging and monitoring, better to not do it in this PR, we are working on a change to introduce `minimal-log-level` through CLI, but for now, I don't think we should change this value here. Can this PR not be made without this change?

sourabh1007:
> Also, can you please create separate PRs for the changes of CLI and the DAB engine?

I’ve already provided an explanation of the files included in the PR. Could you please let me know if there's anything specific that’s unclear or if there's an issue you'd like me to address? I'm happy to help.

sajeetharan:
> > Also, can you please create separate PRs for the changes of CLI and the DAB engine?
> 
> I’ve already provided an explanation of the files included in the PR. Could you please let me know if there's anything specific that’s unclear or if there's an issue you'd like me to address? I'm happy to help.

Please separate the CLI behavior changes and the DAB engine changes for Cosmos DB into different PRs. This will make them easier to manage. I’ll leave the decision up to you and @seantleonard  . Also, could you address the comment on the first question?

seantleonard:
pr fine as is. Just need time to go in and review latest changes. In future, big changes need to be broken up, even with change descriptions.

sajeetharan (approved):
LGTM

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -145,24 +152,27 @@ Mock<PartitionBasedSampler> partitionBasedSampler
...
-            Assert.AreEqual("anotherPojo", result[0]);
-            Assert.AreEqual("anotherProp", result[1]);
+            if (partitionKeyPath == "anotherPojo/anotherProp")
+            {
+                Assert.AreEqual("anotherPojo", result[0]);
+                Assert.AreEqual("anotherProp", result[1]);
+            }
+            else
+            {
+                Assert.AreEqual(partitionKeyPath, result[0]);
```
seantleonard, line 172: do you need to assert that result.Count == 1? 

when partitionkeyapth == "anotherPojo/anotherProp" then at least 2 elements in the list exist.

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -180,23 +190,26 @@ public async Task TestGetPartitionInfoInPartitionBasedSampler()
...
             Mock<TimeBasedSampler> timeBasedSampler
                 = new(_containerWithNamePk, groupCount, numberOfRecordsPerGroup, maxDays, _mockLogger.Object);
-            if (maxDays is not null)
+
+            if (maxDays is null || maxDays == 0)
             {
-                timeBasedSampler
-                    .Setup<long>(x => x.GetTimeStampThreshold())
-                    .Returns((long)(_sortedTimespansNamePk[0] - maxDays));
+                maxDays = 10;
```
seantleonard, line 203: same as previous comments: elaborate on why magic number 10

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -78,9 +78,9 @@ public void TestSchemaGeneratorUsingMultipleJson(string jsonFilePath, string gql
             }
 
             string actualSchema = SchemaGenerator.Generate(jArray, "planet", baseConfig);
-            string expectedSchema = File.ReadAllText($"{gqlFilePath}/{gqlFileName}");
+            string expectedSchema = File.ReadAllText($"{gqlFilePath}/{gqlFileName}").Trim();
```
seantleonard, line 81: were some operations adding extra whitespace? Does needing to add `Trim()` here indicate some issue where output isn't consistent?

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -206,8 +219,7 @@ Mock<TimeBasedSampler> timeBasedSampler
         [TestCleanup]
         public async Task CleanUp()
         {
-            await _containerWithIdPk.DeleteContainerAsync();
-            await _containerWithNamePk.DeleteContainerAsync();
+            await _database.DeleteAsync();
```
seantleonard, line 222: just to confirm, this doesn't run after each datarow, right?

sajeetharan:
@sourabh1007 can we rename the param names as below for schema extraction?

1. TopNExtractor
2. TimePartitionedSampler
3. EligibleDataSampler

src/Service.Tests/CosmosTests/SamplerTests.cs
sourabh1007: no..it run only once.

src/Service.Tests/CosmosTests/SamplerTests.cs
sourabh1007: resetting to default.

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -125,13 +129,16 @@ public async Task TestPartitionBasedSampler(string partitionKeyPath, int? number
...
             Mock<PartitionBasedSampler> partitionBasedSampler
                 = new(_containerWithNamePk, partitionKeyPath, numberOfRecordsPerPartition, maxDaysPerPartition, _mockLogger.Object);
-            if (maxDaysPerPartition is not null)
+
+            if (maxDaysPerPartition is null || maxDaysPerPartition == 0)
             {
-                partitionBasedSampler
-                    .Setup<long>(x => x.GetTimeStampThreshold())
-                    .Returns((long)(_sortedTimespansNamePk[0] - maxDaysPerPartition));
+                maxDaysPerPartition = 30;
```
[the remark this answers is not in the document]
sourabh1007: yes, null and 0 is passed, default value will take over.

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -92,13 +93,16 @@ await executor
...
         {
             Mock<TopNSampler> topNSampler = new(_containerWithIdPk, count, maxDays, _mockLogger.Object);
-            if (maxDays is not null)
+
+            if (maxDays is null || maxDays == 0)
             {
-                topNSampler
-                    .Setup<long>(x => x.GetTimeStampThreshold())
-                    .Returns((long)(_sortedTimespansIdPk[0] - maxDays));
+                maxDays = 10;
```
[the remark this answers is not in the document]
sourabh1007: yes, null and 0 is passed, default value will take over.

src/Cli/Exporter.cs
```diff
@@ -100,71 +90,96 @@ public static int Export(ExportOptions options, ILogger logger, FileSystemRuntim
...
             if (!fileSystem.Directory.Exists(options.OutputDirectory))
             {
                 fileSystem.Directory.CreateDirectory(options.OutputDirectory);
             }
 
+            if (string.IsNullOrEmpty(content))
+            {
+                logger.LogError("There is nothing to write");
+                return;
+            }
```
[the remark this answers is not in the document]
sourabh1007: makes sense

src/Cli/ConfigGenerator.cs
```diff
@@ -143,13 +143,19 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
...
                     {
-                        _logger.LogError("Missing mandatory configuration option for CosmosDB_NoSql: --cosmosdb_nosql-database, and --graphql-schema");
+                        _logger.LogError("Missing mandatory configuration options for CosmosDB_NoSql: --cosmosdb_nosql-database, and --graphql-schema");
                         return false;
                     }
 
-                    if (!fileSystem.File.Exists(graphQLSchemaPath))
+                    if (string.IsNullOrEmpty(graphQLSchemaPath))
+                    {
+                        graphQLSchemaPath = "schema.gql"; // Default to schema.gql
```
abhishekkumams, line 154: what if the default file itself is not found in the directory or is inaccessible?

src/Cli/Commands/ExportOptions.cs
```diff
@@ -1,27 +1,63 @@
...
     public class ExportOptions : Options
     {
-        public ExportOptions(bool graphql, string outputDirectory, string? config, string? graphqlSchemaFile) : base(config)
+        public ExportOptions(bool graphql, string outputDirectory, string? graphqlSchemaFile,
+                bool? generate, string? samplingMode, int? numberOfRecords, string? partitionKeyPath, int? maxDays, int? groupCount, string? config) : base(config)
         {
             GraphQL = graphql;
             OutputDirectory = outputDirectory;
-            GraphQLSchemaFile = graphqlSchemaFile ?? "schema.graphql";
+            GraphQLSchemaFile = graphqlSchemaFile ?? "schema.gql";
```
abhishekkumams, line 17: I know we don't have any existing tests for export command. But would be possible to add one since there are changes related to it ?

src/Cli/ConfigGenerator.cs
sourabh1007: it is not looking for the file, it will create a file with the name.

src/Cli/Commands/ExportOptions.cs
sourabh1007: lets have a separate PR to do test coverage for Cli project. This PR already have feature changes. lets not make it bigger :)

src/Cli/ConfigGenerator.cs
abhishekkumams: And if the file already exists? will it use the one already present or will it overwrite the existing file?

abhishekkumams (approved):

src/Cli.Tests/InitTests.cs
```diff
@@ -164,8 +164,8 @@ bool expectSuccess
...
+        [DataRow("testDatabase", null, "", true, DisplayName = "database is provided, container and Schema is null/empty.")]
```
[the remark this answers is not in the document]
seantleonard: What was the resolution to this thread?

src/Core/Generator/Sampler/TopNExtractor.cs
```diff
@@ -11,11 +11,11 @@ namespace Azure.DataApiBuilder.Core.Generator.Sampler
     /// The TopNSampler class is responsible for retrieving a specified number of recent records 
     /// from an Azure Cosmos DB container, optionally filtering by a maximum number of days.
     /// </summary>
-    public class TopNSampler : ISchemaGeneratorSampler
+    public class TopNExtractor : ISchemaGeneratorSampler
```
seantleonard, line 14: comments now out of date -> `TopNSampler`

src/Core/Generator/SamplingModes.cs
```diff
@@ -12,18 +12,18 @@ public enum SamplingModes
         /// Represents the mode where the top N records are sampled.
         /// This mode selects a specified number of records in descending order based on a timestamp or other sorting criteria.
         /// </summary>
-        TopNSampler,
+        TopNExtractor,
```
seantleonard, line 15: why the name change? pr description and design doc now out of date.

Not too big of an issue but the name changes now result in inconsistent sampler names. Two of the three have the suffix "sampler" and this one does not.

src/Core/Generator/SamplingModes.cs
```diff
@@ -12,18 +12,18 @@ public enum SamplingModes
...
 
         /// <summary>
         /// Represents the mode where data is sampled based on partitions.
         /// In this mode, data is fetched from each partition with a specified limit per partition, which helps in distributing the sampling across different data segments.
         /// </summary>
-        PartitionBasedSampler,
+        EligibleDataSampler,
```
seantleonard, line 21: Where does the name "eligibledata" come from? all the descriptions and design talk about "partition" but i don't see the relation to "eligibledata"

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -76,72 +76,74 @@ await executor
...
-            topNSampler
+            topNExtractor
                 .Setup<long>(x => x.GetTimeStampThreshold())
                 .Returns((long)(_sortedTimespansIdPk[0] - maxDays));
 
-            List<JsonDocument> result = await topNSampler.Object.GetSampleAsync();
-            Assert.AreEqual(expectedCount, result.Count);
+            List<JsonDocument> result = await topNExtractor.Object.GetSampleAsync();
+
+            // The actual count can vary by 1 record because of time based record creation
```
seantleonard, line 108: what does this mean? example? are records created with same time stamp?

src/Service.Tests/CosmosTests/SamplerTests.cs
seantleonard, line 1: All the test method summaries and data row display names in this file read as:
- "this tests <sampler>"
- "this test has X Y Z as input"

which, as you've alluded to in the past, is self-explanatory through code. However, what is important for future developers contributing to DAB and to us when revisiting in the future AND isn't obvious is:
- Why each test's `expectedCount` value is set the way it is. For example, if you say: "test input is x and y, and output is z", a brief explanation on why Z is correct or expected would be helpful. When a test doesn't return the expected value for Z, we can more quickly reason and identify where an error or regression was introduced.


Example:
> [DataRow(5, 2, 3, DisplayName = "TopNExtractor: Retrieve 3 records with max days configured as 2.")]
> `TestTopNExtractor(int count, int? maxDays, int expectedCount)`

Expects 3 records with maxRecordsCount==5 and maxDays==2. -> perhaps in test method summary, point to which testschema file is used and that the test database only has 3 records created in the last 2 days. Additionally, you should also note that the "expected Count" checked in the test actually varies. when you have 3, you actually mean 3 or 4.

seantleonard (approved):
few more questions about name changes and testing documentation to help future engineers.

src/Cli.Tests/InitTests.cs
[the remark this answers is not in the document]
sourabh1007: Actually test is correct, because schema is optional now. Previously it was mandatory information to pass during dab init.
I have updated the comment accordingly.

src/Core/Generator/SamplingModes.cs
sourabh1007: Suggested by @sajeetharan https://github.com/Azure/data-api-builder/pull/2288#issuecomment-2316769732

src/Core/Generator/SamplingModes.cs
sourabh1007: Behind the scene, we are using partition information to collect the sample data but for customer, we are not asking for any partition related input parameter, so name might confuse them, Hence, renamed to this. @sajeetharan  do you want to add here?

src/Service.Tests/CosmosTests/SamplerTests.cs
sourabh1007: If we create records without a delay, they will all have the same timestamp, which would prevent us from testing time-based filtering in test cases. To address this, I've added a delay to ensure that each record has a unique timestamp.

sourabh1007: Because we're relying on a delay to create records with different timestamps, I noticed that, in some cases, the actual result intermittently varies by one record, particularly in pipelines. To ensure these tests aren't flaky, I had to adjust the assertion.

src/Service.Tests/CosmosTests/SamplerTests.cs
sourabh1007: The mentioned expectation is correct based on the configuration, but since we're creating records dynamically, the results sometimes vary. Therefore, this adjustment was necessary. I've added more information to the test cases, which should help clarify the situation.

sourabh1007:
\azp run

src/Core/Generator/SamplingModes.cs
seantleonard: gotcha. Closing out since that's what he recommended.

src/Core/Generator/SamplingModes.cs
seantleonard: gotcha, thanks for clarifying

src/Service.Tests/CosmosTests/SamplerTests.cs
```diff
@@ -65,7 +65,8 @@ await executor
                     .ExecuteQueryAsync<JsonDocument>("SELECT DISTINCT c._ts FROM c ORDER BY c._ts desc",
                         callback: (item) => _sortedTimespansIdPk.Add(item.RootElement.GetProperty("_ts").GetInt32()));
 
-            // Insert additional items into the second container with a delay for unique timestamps.
+            // Insert additional items into the second container with a delay for unique timestamps and partitioned over name i.e planets name.
+            // Number of partitions would 9 as we have 9 unique names.
```
seantleonard, line 69:
```suggestion
            // Number of partitions would be 9 as we have 9 unique names.
```

seantleonard (approved):
Thanks for your patience and understanding. Appreciate this new feature and all your work invested in making this available to DAB users!

PR #2291, opened 2024-07-09
Bumping Version to 1.2 (main)

abhishekkumams:
Updating version to 1.2 in Directory.Build.Props

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <BaseOutputPath>..\out</BaseOutputPath>
-    <Version>1.1</Version>
+    <Version>1.2</Version>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">


seantleonard (approved):

src/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <BaseOutputPath>..\out</BaseOutputPath>
-    <Version>1.1</Version>
+    <Version>1.2</Version>
```
Aniruddh25, line 5:
```suggestion
    <Version>1.3</Version>
```

Shouldn't this be 1.3 in main? And 1.2 in the release branch for 1.2rc?

abhishekkumams: Since, it was previously at `1.1`, it will ensure consistency. Also, we haven't released the stable version for 1.2 yet. so i think it's better to make it `1.2` and then update it to `1.3`.

Aniruddh25: I agree to make 1.2 but shouldn't that happen in the release branch you created for 1.2rc as well? didn't we do that already?

Aniruddh25 (approved):

src/Directory.Build.props
seantleonard: Already set as 1.2 in release branch: https://github.com/Azure/data-api-builder/blob/release/1.2/src/Directory.Build.props

PR #2292, opened 2024-07-10
Fix unpredictable behaviour with pagination resolver

### LINKED ISSUE #2281: [Bug]: GraphQL pagination attempts to return hasNextPage = null and fails with "Cannot return null for non-nullable field"
### What happened?

We attempt to query data from an Azure SQL database table with the GraphQL interface page-by-page.
For the repro we simplified the query to only fetch the primary key `ThingID (PK, int, not null)` and the `Created (datetime2(7), null)` fields and removed all filters.
```
query GetAllThings {
  things(
    first: 100,
    after: null,
  ) {
    items {
      ThingID
      Created
    }
    hasNextPage
    endCursor
  }
}
```
We get the following error response with HTTP status 200:
```
{
    "errors": [
        {
            "message": "Cannot return null for non-nullable field.",
            "locations": [
                {
                    "line": 10,
                    "column": 5
                }
            ],
            "path": [
                "things",
                "hasNextPage"
            ],
            "extensions": {
                "code": "HC0018"
            }
        }
    ],
    "data": {}
}
```

This error result is reproducible as long as we don't change the query above.

However, the result is seemingly unpredictable if we change the number of items requested:
- with $first: 17 => OK, hasNextPage = true, endCursor has a valid value
- with $first: 18 => error 
- with $first: 50 => error
- with $first: 100 => error
- with $first: 130 => OK, hasNextPage = true, endCursor has a valid value

Additional observations/cases:
- if we move the `hasNextPage` and `endCursor` fields before the `items {...}` => suddenly OK with all $first values above => why does this depend on the order of the fields in the query?
- if we remove the `Created` field from the `items {...}` (query only the ID) => suddenly OK with all $first values above => why only reproducible with the DateTime2 field in the query?
- we remove the `ID` field from the `items {...}` (query only the Created) => the same result as before, depending on the $first value => why only reproducible with the DateTime2 field in the query?

What might be the root cause of this behaviour?
Is it a reliable workaround/fix if we move the paging fields "up" before the items in our queries? 

### Version

1.1.7

Azure SQL

Local (including CLI)

GraphQL

### Relevant log output

```Text
      e2efd26f-c7ba-40d0-a626-d293fd3366a7 Executing query: SELECT TOP 101 [table0].[Created] AS [Created], [table0].[ThingID] AS [ThingID] FROM [dbo].[Thing] AS [table0] WHERE 1 = 1 ORDER BY [table0].[ThingID] ASC FOR JSON PATH, INCLUDE_NULL_VALUES
      Request finished HTTP/1.1 POST http://localhost:5000/graphql - 200 - application/json;+charset=utf-8 217.6527ms
```

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #2281
  - Unpredictable behavior was noticed with pagination fields. 

## What's causing this issue?

The error was caused by the timing of the disposal of the `JsonDocument`. `JsonDocument` was disposed of prematurely, before we had finished using its value. This premature disposal was leading to accessing invalid or disposed resources, which causes the error.

To understand this better, let's break it down:

1. **JsonDocument Lifecycle**:
   - A `JsonDocument` is used to parse and hold JSON data.
   - When a `JsonDocument` is disposed, all its resources are released, and any attempt to access its data afterward will result in an error.

2. **Expected Behavior**:
   - `JsonDocument` is disposed correctly only after we are done using it.
   - Proper disposal is crucial to free up resources and avoid memory leaks.

3. **Actual Issue**:
   - The `JsonDocument` was disposed of before we finished using its value.
   - This premature disposal means that any subsequent access to the `JsonDocument`'s data will fail because the data is no longer available.

## What is this change?

- To prevent this issue, we need to ensure that we copy the root element of the `JsonDocument` before it is disposed.
- By copying the root element, we can continue to use the data even after the `JsonDocument` has been disposed.

## How was this tested?

- [x] Integration Tests

### Before
image
image

### After
image
image


ayush3797 (approved):

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -272,17 +271,17 @@ public object ResolveList(JsonElement array, IObjectField fieldSchema, ref IMeta
...
-                int written = Encoding.UTF8.GetBytes(text, buffer.GetSpan(neededCapacity));
-                buffer.Advance(written);
-
-                Utf8JsonReader reader = new(buffer.GetWrittenSpan());
-                foreach (JsonElement element in JsonElement.ParseValue(ref reader).EnumerateArray())
+                string jsonString = array.GetString()!;
+                if (!string.IsNullOrEmpty(jsonString))
                 {
-                    resolvedList.Add(element);
+                    JsonDocument jsonDocument = JsonDocument.Parse(jsonString);
```
Aniruddh25, line 277: When is this JsonDocument disposed off? The primary reason of not using JsonDocument was leaked memory..  Is that problem no longer applicable here?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -272,17 +271,17 @@ public object ResolveList(JsonElement array, IObjectField fieldSchema, ref IMeta
             }
             else if (array.ValueKind is JsonValueKind.String)
             {
-                using ArrayPoolWriter buffer = new();
```
Aniruddh25, line 275: Instead of removing usage of ArrayPoolWriter altogether, can we fix the memory advancement bugs in its implementation?
ArrayPoolWriter was introduced specifically to avoid usage of JsonDocument.Parse.. Reintroducing JsonDocument, seems like we are going back in time.

If we indeed end up removing ArrayPoolWriter usage, we should get rid of its implementation as well from our project.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -932,6 +932,68 @@ public async Task TestColumnTieBreak()
...
+
+            // Assert
+            Assert.IsNotNull(graphQLResponse);
+
+            // Check for no errors
+            Assert.IsFalse(graphQLResponse.TryGetProperty("errors", out _), "Response contains errors");
+
+            // Check the number of items
+            JsonElement items = graphQLResponse.GetProperty("items");
+            Assert.AreEqual(items.GetArrayLength() == pageSize, pageSize <= 200);
```
Aniruddh25, line 970: why does it matter pageSize to be more <=200 when items returned are = pageSize?

Aniruddh25 (commented):
Posting few questions on the approach..

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
abhishekkumams: because items returned won't be equal to pageSize, when the total number of rows in the table is less than the pageSize.
for example: if pageSize==1000, but no. of rows in our table is only 20. we will only get 20 rows in response.

src/Core/Resolvers/SqlQueryEngine.cs
abhishekkumams: This was what i was trying before, used different approaches like increasing neededCapacity, but this was again uncertain as to how much would be correct as it was still giving uncertain results. And it was harder to debug as well due to manual buffer management and encoding conversion as to what exactly is causing this issue.

That's why i resorted to JsonDocument to reduce the complexity of this process.

abhishekkumams: can do a clean up in a separate PR

src/Core/Resolvers/SqlQueryEngine.cs
abhishekkumams: thanks, updated

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -272,17 +271,19 @@ public object ResolveList(JsonElement array, IObjectField fieldSchema, ref IMeta
...
-                int written = Encoding.UTF8.GetBytes(text, buffer.GetSpan(neededCapacity));
-                buffer.Advance(written);
-
-                Utf8JsonReader reader = new(buffer.GetWrittenSpan());
-                foreach (JsonElement element in JsonElement.ParseValue(ref reader).EnumerateArray())
+                string jsonString = array.GetString()!;
+                if (!string.IsNullOrEmpty(jsonString))
                 {
-                    resolvedList.Add(element);
+                    using (JsonDocument jsonDocument = JsonDocument.Parse(jsonString))
```
aaronburtle, line 277: Will this JsonDocument be properly disposed of?

src/Service.Tests/SqlTests/GraphQLPaginationTests/MsSqlGraphQLPaginationTests.cs
```diff
@@ -112,5 +112,95 @@ public static async Task SetupAsync(TestContext context)
                 endCursorIdValue,
                 isLastPage);
         }
+
+        /// <inheritdoc />
+        [DataTestMethod]
+        [DataRow(1, DisplayName = "1 item per page")]
```
aaronburtle, line 118: Will this set of cases be enough to reveal this bug? Should we add some odd sizes or other possible edge cases?

src/Core/Resolvers/SqlQueryEngine.cs
abhishekkumams: yeah, wondering if have you noticed any issues regarding this in the past?

src/Service.Tests/SqlTests/GraphQLPaginationTests/MsSqlGraphQLPaginationTests.cs
abhishekkumams: i will add few more.

src/Core/Resolvers/SqlQueryEngine.cs
aaronburtle: we had some issues with json documents leading to memory leaks, but the docs say this is how to use it

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-dom#use-jsondocument


> Creates the JsonDocument instance in a [using statement](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using) because JsonDocument implements IDisposable. After a JsonDocument instance is disposed, you lose access to all of its JsonElement instances also. To retain access to a JsonElement instance, make a copy of it before the parent JsonDocument instance is disposed. To make a copy, call [JsonElement.Clone](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.clone). For more information, see [JsonDocument is IDisposable](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft#jsondocument-is-idisposable).

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
Aniruddh25: can you please add a comment here as to what this is asserting?

Effectively its saying that 
When PageSize > 200, returned items length CANNOT BE = whatever the pagesize is. 
When pageSize <=200, returned items length = the pagesize

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -272,17 +271,19 @@ public object ResolveList(JsonElement array, IObjectField fieldSchema, ref IMeta
...
+                    {
+                        if (jsonDocument.RootElement.ValueKind == JsonValueKind.Array)
```
Aniruddh25, line 279: What do we do when its not an ArrayKind?

Aniruddh25 (commented):
Still some questions on why the number 200?

seantleonard:
Reviewing will be easier if you can also provide details about:
> ArrayPoolWriter uses pooled memory, which can lead to issues if not managed correctly. If the buffer is not properly sized or if there is a memory leak, it can cause errors when handling larger datasets as we are seeing in our case when we try to get more number of rows in a single query after a certain value, it starts to fail due to loss of "hasNext" property.

Can you provide examples of the issues are you experiencing in your PR description? By reverting back to JsonDocuments for all the results, you're allocating memory for each of those even when using "using." Michael's intent here was to avoid all those allocations. 
- Are there uncaught/unhandled silent errors occurring in DAB's `ArrayPoolWriter.cs` class? 
- Perhaps breakpoints there can show us where there might be capacity issues when renting a buffer from the array pool? e.g. `ArrayPool<byte>.Shared.Rent(size);`

seantleonard:
In your test repro, stepping through debugger, 
`ExecutionHelper::ExecuteLeafField(IPureResolverContext context)` L140
- place conditional breakpoint : `((NamePathSegment)context.Path).Name.Value == "hasNextPage"`
  - Issue is that hasNextPage resolves to null. However, stepping through another iteration, "endCursor" resolves fine. 
I'm not sure why the arraypoolwriter would result in just "hasNextPage" being null. Perhaps the ExecutionHelper/ResolverTypeInterceptor has a bug somewhere. 

Here:
image
For whatever reason, all the JSON from the result set is captured without issue... however, hasNextPage is not persisted to the json object. Perhaps there is a missing function call or a function is operating on the wrong json document.

abhishekkumams:
> In your test repro, stepping through debugger, `ExecutionHelper::ExecuteLeafField(IPureResolverContext context)` L140
> 
> * place conditional breakpoint : `((NamePathSegment)context.Path).Name.Value == "hasNextPage"`
>   
>   * Issue is that hasNextPage resolves to null. However, stepping through another iteration, "endCursor" resolves fine.
>     I'm not sure why the arraypoolwriter would result in just "hasNextPage" being null. Perhaps the ExecutionHelper/ResolverTypeInterceptor has a bug somewhere.
> 
> Here: image For whatever reason, all the JSON from the result set is captured without issue... however, hasNextPage is not persisted to the json object. Perhaps there is a missing function call or a function is operating on the wrong json document.

the problem is the bug is inconsistent, i.e, it happens for certain page size only which is also not fixed, because of which it's hard to debug. The cause that i mentioned is the best guess i could make as ArrayPoolWriter uses low level handling, as it introduces more complexity. while JsonDocument is taking care of whatever that is missing from the ArrayPoolWriter Implementation.

seantleonard:
More notes from investigating. Seems there might be issue with how resolver is written.


Given the following request:
```graphql
query myQuery {
  supportedTypes(first: 4) {
    items {
      typeid
      boolean_types
      byte_types
      bytearray_types
      date_types
      datetime2_types
      datetime_types
      datetimeoffset_types
      decimal_types
    }
    hasNextPage

  }
}
```

Pure resolvers are executed for:
- `items` -> **ResolverTypeInterceptor::_listFieldResolver => ExecutionHelper.ExecuteListField(ctx)**
	- fields for each item ResolverTypeInterceptor::_leafFieldResolver => ExecutionHelper.ExecuteLeafField(ctx); 
- `hasNextPage` -> **ResolverTypeInterceptor::_leafFieldResolver => ExecutionHelper.ExecuteLeafField(ctx);**

Depending on order of fields written in query:
1. `items` before `hasNextPage` -> items result overwrites pureresolvercontext 
2. `items` after `hasNextPage` -> result as expected.

HotChocolate.Execution.Processing.MiddlewareContext.IsResultModified = true
HotChocolate.Execution.Processing.MiddlewareContext.Result -> only contains resolved value of "items" field and "endCursor", if included in query.

Consequentially, for order bullet point 1, the PureResolverContext passed to the leafFieldResolver no longer contains the key/value pair for "`hasNextPage:true/false`"
thus, the value is set to null in the ResultMap built by `HC::ResolverTaskFactory::EnqueueOrInlineResolverTasks(...)`

I acknowledge in HC docs:
"Because of this it is important that resolvers, with the exception of top level mutation field resolvers, do not contain side-effects, since their execution order may vary."
However, our pure resolvers shouldn't have any side effects as they are just parsing the json result from the database.

seantleonard:
after more investigation: I think I found another fix.. which only involves ensuring we "JsonElement.Clone()" in at least 1 more place. 

Here where we write the context result, this is what I was observing to be overwritten in the cases where we see "repro"
image

At least a few exercises of trying to repro result in the error no longer showing. IF this is the solution, then the error was due to timing of when "jsondocument" disposal occurred, which ironically is what line 257 is supposed to be doing for us. Perhaps the disposal occurred before we were finished using the value from the jsondocument, thus needing to ensure copying the root element

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
abhishekkumams: it has nothing to do with default size. Actually based on the test data, i put 200.

When PageSize > 200, returned items length CANNOT BE = whatever the pagesize is. YES, because based on the test data number of rows in the DB < 200.

So, 200 is just an arbitrary number put there to test different fields in the result based on the number of rows in the DB.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -932,6 +932,72 @@ public async Task TestColumnTieBreak()
             SqlTestHelper.PerformTestEqualJsonStrings(expected2, actual2.ToString());
         }
 
+        /// <summary>
+        /// Tests the GraphQL query for retrieving supported types with a specified number of rows.
```
seantleonard, line 936: nit: need to update this description to note how this test is a 'best effort' attempt to reproduce an issue that occurs due to jsondocument disposal occurring before we are finished using it's values. As a result, this test attempts to catch circumstances where we don't call `JsonDocument.RootElement.Clone()`

seantleonard (approved):
thanks for fixing and adding regression tests! Also, thanks for your patience while we identified the true root cause of the issue.

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -966,6 +970,8 @@ protected async Task TestPaginantionForGivenPageSize(int pageSize, string fields
             Assert.IsFalse(graphQLResponse.TryGetProperty("errors", out _), "Response contains errors");
 
             // Check the number of items
+            // Total number of rows in DB < 200, so when the pagesize exeeds 200, the number of rows returned in the result
```
Aniruddh25, line 973:
```suggestion
            // Total number of rows in DB < 200, so when the pagesize exceeds 200, the number of rows returned in the result
```

Aniruddh25 (approved):
LGTM! Thanks for checking in the right fix and not giving up :)

seantleonard:
Test pipelines failed for at least 2 iterations with the following error messages:
>at Azure.DataApiBuilder.Service.Tests.Configuration.ConfigurationTests.ValidateMutationSucceededAtDbLayer(TestServer server, HttpClient client, String query, String queryName, String auth***, String clientRoleHeader) in D:\a\1\s\src\Service.Tests\Configuration\ConfigurationTests.cs:line 2750


> Unhandled exception. System.InvalidOperationException: The requested operation requires an element of type 'Object', but the target element has type 'Null'.
>    at System.Text.Json.ThrowHelper.ThrowJsonElementWrongTypeException(Json***Type expectedType, Json***Type actualType)
>    at System.Text.Json.JsonDocument.TryGetNamedPropertyValue(Int32 index, ReadOnlySpan`1 propertyName, JsonElement& value)
>    at System.Text.Json.JsonElement.TryGetProperty(String propertyName, JsonElement& value)
>    at Azure.DataApiBuilder.Service.Tests.Configuration.ConfigurationTests.ValidateMutationSucceededAtDbLayer(TestServer server, HttpClient client, String query, String queryName, String auth***, String clientRoleHeader) in D:\a\1\s\src\Service.Tests\Configuration\ConfigurationTests.cs:line 2750
>    at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
>    at System.Threading.QueueUserWorkItemCallback.Execute()
>    at System.Threading.ThreadPoolWorkQueue.Dispatch()
>    at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
> The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.InvalidOperationException: The requested operation requires an element of type 'Object', but the target element has type 'Null'.
>    at System.Text.Json.ThrowHelper.ThrowJsonElementWrongTypeException(Json***Type expectedType, Json***Type actualType)
>    at System.Text.Json.JsonDocument.TryGetNamedPropertyValue(Int32 index, ReadOnlySpan`1 propertyName, JsonElement& value)
>    at System.Text.Json.JsonElement.TryGetProperty(String propertyName, JsonElement& value)
>    at Azure.DataApiBuilder.Service.Tests.Configuration.ConfigurationTests.ValidateMutationSucceededAtDbLayer(TestServer server, HttpClient client, String query, String queryName, String auth***, String clientRoleHeader) in D:\a\1\s\src\Service.Tests\Configuration\ConfigurationTests.cs:line 2750
>    at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
>    at System.Threading.QueueUserWorkItemCallback.Execute()
>    at System.Threading.ThreadPoolWorkQueue.Dispatch()
>    at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()

PR #2302, opened 2024-07-17
Updating the version to 1.3

abhishekkumams:
minor version is bumped to 3 for the next upcoming release version 1.3

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <BaseOutputPath>..\out</BaseOutputPath>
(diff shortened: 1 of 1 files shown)


src/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <BaseOutputPath>..\out</BaseOutputPath>
-    <Version>1.2</Version>
+    <Version>1.3</Version>
```
Aniruddh25, line 5: shouldn't we do this after we release the stable version of 1.2?

Aniruddh25 (approved):

seantleonard (approved):

Issue #2284, opened 2024-06-28
[Bug]: Float fields get saved based on local Windows Regional format setting


evacotelin:
### What happened?

I have the following GraphQL schema definition:

```
type AppReport @model {
  id: ID
  description: String
  latitude: Float
  longitude: Float
  created: DateTime
}
```

When executing the following mutation, the Float fields (i.e. latitude & longitude) get saved as integers (the decimal separator is removed, not really converted to integer). 
![image](https://github.com/Azure/data-api-builder/assets/145988561/2418c210-ef79-4327-ad35-02afc684eba3)

This is due to my Windows _Regional format_ setting, because the comma (,) is used as the default decimal separator: 
![image](https://github.com/Azure/data-api-builder/assets/145988561/09e840e8-781b-4e1c-88ef-db944a4319d1)

When changing my Windows Regional format setting to _English (United States)_ the GraphQL mutation works as expected: saves latitude & longitude fields as Floats (because the point (.) is used as the default decimal separator).

### Version

1.1.7+74ea6c5f37f8629fd7f8b13fc56027bf0bf0a93a

### What database are you using?

CosmosDB NoSQL

### What hosting model are you using?

Static Web Apps (SWA)

### Which API approach are you accessing DAB through?

GraphQL

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

sourabh1007:
Fixed.

PR #2307, opened 2024-07-25
Cultural Invariant Conversion of datatypes during mutation

sourabh1007:
## Why make this change?

It addresses https://github.com/Azure/data-api-builder/issues/2284 bug raised by the customer.

**Note:** Made change in`src/Core/Resolvers/SqlMutationEngine.cs` just for consistency.

## What is this change?
Made sure datatype conversion is `CultureInfo.InvariantCulture`

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

### DIFF
--- a/src/Core/Resolvers/SqlMutationEngine.cs
+++ b/src/Core/Resolvers/SqlMutationEngine.cs
@@ -3,7 +3,6 @@
 
 using System.Data;
 using System.Data.Common;
-using System.Globalization;
 using System.Net;
 using System.Text.Json;
 using System.Text.Json.Nodes;
@@ -161,7 +160,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                         // If the number of records affected by DELETE were zero,
                         if (resultProperties is not null
                             && resultProperties.TryGetValue(nameof(DbDataReader.RecordsAffected), out object? value)
-                            && Convert.ToInt32(value, CultureInfo.InvariantCulture) == 0)
+                            && Convert.ToInt32(value) == 0)
                         {
                             // the result was not null previously, it indicates this DELETE lost
                             // a concurrent request race. Hence, empty the non-null result.
@@ -506,7 +505,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                     // DbDataReader.RecordsAffected contains the number of rows changed deleted. 0 if no records were deleted.
                     // When the flow reaches this code block and the number of records affected is 0, then it means that no failure occurred at the database layer
                     // and that the item identified by the specified PK was not found.
-                    if (Convert.ToInt32(value, CultureInfo.InvariantCulture) == 0)
+                    if (Convert.ToInt32(value) == 0)
                     {
                         string prettyPrintPk = "<" + string.Join(", ", context.PrimaryKeyValuePairs.Select(kv_pair => $"{kv_pair.Key}: {kv_pair.Value}")) + ">";
 
--- a/src/Core/Resolvers/SqlMutationEngine.cs
+++ b/src/Core/Resolvers/SqlMutationEngine.cs
@@ -3,6 +3,7 @@
 
 using System.Data;
 using System.Data.Common;
+using System.Globalization;
 using System.Net;
 using System.Text.Json;
 using System.Text.Json.Nodes;
@@ -160,7 +161,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                         // If the number of records affected by DELETE were zero,
                         if (resultProperties is not null
                             && resultProperties.TryGetValue(nameof(DbDataReader.RecordsAffected), out object? value)
-                            && Convert.ToInt32(value) == 0)
+                            && Convert.ToInt32(value, CultureInfo.InvariantCulture) == 0)
                         {
                             // the result was not null previously, it indicates this DELETE lost
                             // a concurrent request race. Hence, empty the non-null result.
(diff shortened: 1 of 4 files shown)


seantleonard:
please explain why this works, what's the behavior before/after.

sourabh1007:
> please explain why this works, what's the behavior before/after.

you just want me to put in description or it is a question, because changes are not clear enough? I have test cases added for this, you can refer that, that case will fail if you run it with main branch.
To know the "before" behavior, you can check the mentioned issue.

src/Core/Resolvers/SqlMutationEngine.cs
seantleonard, line 1: why the changes in this file? The linked issue and the tests in this pr only test Cosmos. Not sure why sql changes are made and why the changes were made to the "records affected" integer.. that value isn't returned to users.

src/Core/Services/TypeHelper.cs
```diff
@@ -304,12 +305,12 @@ public static bool TryGetDbTypeFromSqlDbDateTimeType(SqlDbType sqlDbType, [NotNu
             SyntaxKind valueKind = node.Kind;
             return valueKind switch
             {
-                SyntaxKind.IntValue => Convert.ToInt32(node.Value), // spec
-                SyntaxKind.FloatValue => Convert.ToDouble(node.Value), // spec
-                SyntaxKind.BooleanValue => Convert.ToBoolean(node.Value), // spec
-                SyntaxKind.StringValue => Convert.ToString(node.Value), // spec
+                SyntaxKind.IntValue => Convert.ToInt32(node.Value, CultureInfo.InvariantCulture), // spec
+                SyntaxKind.FloatValue => Convert.ToDouble(node.Value, CultureInfo.InvariantCulture), // spec
+                SyntaxKind.BooleanValue => Convert.ToBoolean(node.Value, CultureInfo.InvariantCulture), // spec
```
seantleonard, line 310: why do you need to define `CultureInfo.InvariantCulture` for Convert.ToBoolean? that parameter is ignored per documentation:https://learn.microsoft.com/dotnet/api/system.convert.toboolean?view=net-8.0#system-convert-toboolean(system-string-system-iformatprovider)

src/Core/Services/TypeHelper.cs
```diff
@@ -304,12 +305,12 @@ public static bool TryGetDbTypeFromSqlDbDateTimeType(SqlDbType sqlDbType, [NotNu
             SyntaxKind valueKind = node.Kind;
             return valueKind switch
             {
-                SyntaxKind.IntValue => Convert.ToInt32(node.Value), // spec
-                SyntaxKind.FloatValue => Convert.ToDouble(node.Value), // spec
-                SyntaxKind.BooleanValue => Convert.ToBoolean(node.Value), // spec
-                SyntaxKind.StringValue => Convert.ToString(node.Value), // spec
+                SyntaxKind.IntValue => Convert.ToInt32(node.Value, CultureInfo.InvariantCulture), // spec
+                SyntaxKind.FloatValue => Convert.ToDouble(node.Value, CultureInfo.InvariantCulture), // spec
```
seantleonard, line 309: Would be helpful to know **why** this is the correct change to make (why here?).
- Is this a breaking change? 
- does this change how values are stored in the backend database? The tests don't demonstrate this.
- examples from a chart like this would help. 

|Provided Value | client culture| culture | persisted value | returned value|
|- |- |- |- |- |
|12.32 (double)| en-DE | InvariantCulture| 12,32 | 12,32 |
|12.32 (double)| en-DE | Before change, no culture specified| 12.32| 1232??|

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -1074,6 +1075,31 @@ public async Task CanPatchMoreThan10AttributesInAnItemWithVariables()
             Assert.AreEqual(patchResponse.GetProperty("moons").ToString(), JsonSerializer.Serialize(update.moons));
         }
 
+        [TestMethod]
+        [DataRow("en-US")]
+        [DataRow("en-DE")]
+        public async Task CanCreateItemWithCultureInvariant(string cultureInfo)
+        {
+            CultureInfo ci = new(cultureInfo);
+            CultureInfo.DefaultThreadCurrentCulture = ci;
```
seantleonard, line 1084: Comment would help, this defines the culture of the DAB "host" machine or "thread." 

So in this test a double value `10.15` (en-US culture) would be persisted to database as "cultureInfo.InvariantCulture" which wouldn't change the value -> stored as `10.15`. but then when response is returned, because the DefaultThreadCulture is `en-DE`, the value is converted in the response to match the host or client's culture?

seantleonard (changes requested):

src/Core/Resolvers/SqlMutationEngine.cs
sourabh1007: I am reverting this change. 
By the way, it doesn't matter if you are returning this value to the customer or not. This code and cosmos DB code which I am fixing, both runs on the server and so issue can happen here also, So I thought to put it as safety check. 
But you are right, it needs to be tested separately. SQL team can take it up if there is any issue in the future due to this.

src/Core/Services/TypeHelper.cs
sourabh1007: In the attached issue, it is clearly mentioned that, when customer is trying to save 12.32, it is getting converted to 1232 and then same is returning back.
1) I don't think it is a breaking change. It is a fix. ideally whatever you are sending that should be saved.
2) Yes, in the issue it is already mentioned
image
3) it not about `en-DE` only, it about general application behavior, application should not take the decision of converting values. it should save whatever is coming from customer.

src/Service.Tests/CosmosTests/MutationTests.cs
sourabh1007: No, In response, value won't be converted back to host/client culture, thats what I am asserting here by putting "expected value" as cultural invariant.

neeraj-sharma2592 (approved):

seantleonard (approved):

PR #2316, opened 2024-08-05
[Cherrypick][Port] Cultural Invariant Conversion of datatypes during mutation

seantleonard:
## Port Metadata
- porting from `main` to `release/1.2`
  - Issue: #2284 
  - PR: #2307

## Why make this change?

It addresses https://github.com/Azure/data-api-builder/issues/2284 bug raised by the customer.

~~**Note:** Made change in`src/Core/Resolvers/SqlMutationEngine.cs` just for consistency.~~ [SL: this was in original PR, but is not accurate, since no changes to SqlMutationEngine.cs are included

## What is this change?
Made sure datatype conversion is `CultureInfo.InvariantCulture`

### DIFF
--- a/src/Core/Services/TypeHelper.cs
+++ b/src/Core/Services/TypeHelper.cs
@@ -3,6 +3,7 @@
 
 using System.Data;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Net;
 using Azure.DataApiBuilder.Core.Services.OpenAPI;
 using Azure.DataApiBuilder.Service.Exceptions;
@@ -304,12 +305,12 @@ namespace Azure.DataApiBuilder.Core.Services
             SyntaxKind valueKind = node.Kind;
             return valueKind switch
             {
-                SyntaxKind.IntValue => Convert.ToInt32(node.Value), // spec
-                SyntaxKind.FloatValue => Convert.ToDouble(node.Value), // spec
+                SyntaxKind.IntValue => Convert.ToInt32(node.Value, CultureInfo.InvariantCulture), // spec
+                SyntaxKind.FloatValue => Convert.ToDouble(node.Value, CultureInfo.InvariantCulture), // spec
                 SyntaxKind.BooleanValue => Convert.ToBoolean(node.Value), // spec
-                SyntaxKind.StringValue => Convert.ToString(node.Value), // spec
+                SyntaxKind.StringValue => Convert.ToString(node.Value, CultureInfo.InvariantCulture), // spec
                 SyntaxKind.NullValue => null, // spec
-                _ => Convert.ToString(node.Value)
+                _ => Convert.ToString(node.Value, CultureInfo.InvariantCulture)
             };
         }
 
--- a/src/Service.Tests/CosmosTests/TestBase.cs
+++ b/src/Service.Tests/CosmosTests/TestBase.cs
@@ -46,7 +46,7 @@ type Planet @model(name:""PlanetAlias"") {
     id : ID!,
     name : String,
     character: Character,
-    age : Int,
+    age : Float,
     dimension : String,
     earth: Earth,
     tags: [String!],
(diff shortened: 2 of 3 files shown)


src/Core/Services/TypeHelper.cs
```diff
@@ -304,12 +305,12 @@ public static bool TryGetDbTypeFromSqlDbDateTimeType(SqlDbType sqlDbType, [NotNu
...
             return valueKind switch
             {
-                SyntaxKind.IntValue => Convert.ToInt32(node.Value), // spec
-                SyntaxKind.FloatValue => Convert.ToDouble(node.Value), // spec
+                SyntaxKind.IntValue => Convert.ToInt32(node.Value, CultureInfo.InvariantCulture), // spec
+                SyntaxKind.FloatValue => Convert.ToDouble(node.Value, CultureInfo.InvariantCulture), // spec
                 SyntaxKind.BooleanValue => Convert.ToBoolean(node.Value), // spec
-                SyntaxKind.StringValue => Convert.ToString(node.Value), // spec
+                SyntaxKind.StringValue => Convert.ToString(node.Value, CultureInfo.InvariantCulture), // spec
                 SyntaxKind.NullValue => null, // spec
```
Aniruddh25, line 312: not for this PR, but do you know what the comment `// spec` refers to here?

src/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <BaseOutputPath>..\out</BaseOutputPath>
-    <Version>1.1</Version>
```
Aniruddh25, line 5: Is this part of this PR?

Aniruddh25 (approved):
Looks good.

src/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
...
+    <Version>1.2</Version>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
Aniruddh25, line 8: The PR description still says: there was a change made in SqlMutationEngine.cs but looking at the original PR, those changed were removed. Looks like the original PR didnt update the description, lets at least update in this cherry-pick PR. 
Also, do we have a similar issue for SQL? If yes, we should fix that too in this version 1.2 (in a separate PR)

src/Directory.Build.props
seantleonard: Investigating. Looks like i added 2 commits.

src/Service.Tests/CosmosTests/TestBase.cs
```diff
@@ -46,7 +46,7 @@ public class TestBase
     id : ID!,
     name : String,
     character: Character,
-    age : Int,
+    age : Float,
```
abhishekkumams, line 49: why age is not Int?

abhishekkumams (approved):

src/Directory.Build.props
seantleonard: I only added one commit, i just need to mark the "merge to" branch as `release/1.2` in GitHub. Fixed.

src/Directory.Build.props
seantleonard: 1. Updated this PR's description to highlight the fact no changes to sqlmutationengine are included.
2. The fix from the original PR affects "typehelper.cs" which is universal. I added a work item to add mirroring tests for SQL: #2319

src/Service.Tests/CosmosTests/TestBase.cs
seantleonard: That was Sourabh's choice to make in #2307 to test this behavior. When we make similar test for SQL via #2319 we don't have to go that route.

src/Core/Services/TypeHelper.cs
seantleonard: I don't recall. That was something Michael added in his perf improvement PR. 'spec' probably aligns with hotchocolate internals is my best guess.

PR #2327, opened 2024-08-09
Adding logs to monitor db query execution time in dab

rohkhann:
## Why make this change?
We want to understand how much time db execution takes and using that determine the GQL overhead. Total time = Graphql Overhead + Dbexecutiontime.

## What is this change?
this query adds the time it took for the query to execute on the db. Total execution time - time query took on db should give us the graphql overhead. 

## How was this tested?
1. added tests for logging 
2. added test for testing locking behavior on the http context object.

### DIFF
--- a/src/Core/Resolvers/IQueryExecutor.cs
+++ b/src/Core/Resolvers/IQueryExecutor.cs
@@ -4,7 +4,6 @@
 using System.Data.Common;
 using System.Text.Json.Nodes;
 using Azure.DataApiBuilder.Core.Models;
-using HotChocolate.Resolvers;
 using Microsoft.AspNetCore.Http;
 
 namespace Azure.DataApiBuilder.Core.Resolvers
@@ -26,7 +25,6 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         /// <param name="httpContext">Current request httpContext.</param>
         /// <param name="args">List of string arguments to the DbDataReader handler.</param>
         /// <param name="dataSourceName">dataSourceName against which to run query. Can specify null or empty to run against default db.</param>
-        /// <param name="middlewareContext">middleware context for gql operations.</param>
         /// <returns>An object formed using the results of the query as returned by the given handler.</returns>
         public Task<TResult?> ExecuteQueryAsync<TResult>(
             string sqltext,
@@ -34,8 +32,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             Func<DbDataReader, List<string>?, Task<TResult>>? dataReaderHandler,
             string dataSourceName,
             HttpContext? httpContext = null,
-            List<string>? args = null,
-            IMiddlewareContext? middlewareContext = null);
+            List<string>? args = null);
 
         /// <summary>
         /// Executes sql text with the given parameters and
--- a/src/Core/Resolvers/IQueryExecutor.cs
+++ b/src/Core/Resolvers/IQueryExecutor.cs
@@ -4,6 +4,7 @@
 using System.Data.Common;
 using System.Text.Json.Nodes;
 using Azure.DataApiBuilder.Core.Models;
+using HotChocolate.Resolvers;
 using Microsoft.AspNetCore.Http;
 
 namespace Azure.DataApiBuilder.Core.Resolvers
@@ -25,6 +26,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         /// <param name="httpContext">Current request httpContext.</param>
         /// <param name="args">List of string arguments to the DbDataReader handler.</param>
         /// <param name="dataSourceName">dataSourceName against which to run query. Can specify null or empty to run against default db.</param>
+        /// <param name="middlewareContext">middleware context for gql operations.</param>
         /// <returns>An object formed using the results of the query as returned by the given handler.</returns>
         public Task<TResult?> ExecuteQueryAsync<TResult>(
             string sqltext,
@@ -32,7 +34,8 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             Func<DbDataReader, List<string>?, Task<TResult>>? dataReaderHandler,
             string dataSourceName,
             HttpContext? httpContext = null,
-            List<string>? args = null);
+            List<string>? args = null,
+            IMiddlewareContext? middlewareContext =  null);
 
         /// <summary>
         /// Executes sql text with the given parameters and
--- a/src/Core/Resolvers/IQueryExecutor.cs
+++ b/src/Core/Resolvers/IQueryExecutor.cs
@@ -35,7 +35,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             string dataSourceName,
             HttpContext? httpContext = null,
             List<string>? args = null,
-            IMiddlewareContext? middlewareContext =  null);
+            IMiddlewareContext? middlewareContext = null);
 
         /// <summary>
         /// Executes sql text with the given parameters and
--- a/src/Core/Resolvers/QueryExecutor.cs
+++ b/src/Core/Resolvers/QueryExecutor.cs
@@ -3,6 +3,7 @@
 
 using System.Data;
 using System.Data.Common;
+using System.Diagnostics;
 using System.Net;
 using System.Text;
 using System.Text.Json;
@@ -10,6 +11,7 @@ using System.Text.Json.Nodes;
 using Azure.DataApiBuilder.Core.Configurations;
 using Azure.DataApiBuilder.Core.Models;
 using Azure.DataApiBuilder.Service.Exceptions;
+using HotChocolate.Resolvers;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Polly;
@@ -23,6 +25,8 @@ namespace Azure.DataApiBuilder.Core.Resolvers
     public class QueryExecutor<TConnection> : IQueryExecutor
         where TConnection : DbConnection, new()
     {
+        private const string TOTALDBXEXECUTIONTIME = "TotalDbExecutionTime";
+
         protected DbExceptionParser DbExceptionParser { get; }
         protected ILogger<IQueryExecutor> QueryExecutorLogger { get; }
         protected RuntimeConfigProvider ConfigProvider { get; }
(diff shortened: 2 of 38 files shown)


src/Core/Resolvers/IQueryExecutor.cs
seantleonard, line 1: I don't follow how measuring the database query execution time correlates to "GraphQL overhead." Wouldn't measuring query execution duration tell you what the "database overhead" is?

From my perspective, GraphQL overhead would consist of the following:
1. Time between Processing user input and providing sql query metadata to QueryExecutor
2. <query execution, does not count towards GraphQL overhead>
3. Time between receiving response to database and HotChocolate middleware providing formatted JSON as a response to the client.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -158,7 +162,8 @@ public class QueryExecutor<TConnection> : IQueryExecutor
             Func<DbDataReader, List<string>?, Task<TResult>>? dataReaderHandler,
             string dataSourceName,
             HttpContext? httpContext = null,
-            List<string>? args = null)
+            List<string>? args = null,
+            IMiddlewareContext? middlewareContext = null)
```
seantleonard, line 166: You already have access to "HttpContext" can't you just add the required metadata to that? This pr adds another function argument in many places just to keep track of a timer seems excessive.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -23,6 +25,8 @@ namespace Azure.DataApiBuilder.Core.Resolvers
     public class QueryExecutor<TConnection> : IQueryExecutor
         where TConnection : DbConnection, new()
     {
+        private const string TOTALDBXEXECUTIONTIME = "TotalDbExecutionTime";
```
pratsingms, line 28: Nit: Did you mean to name this variable `TOTALDBEXECUTIONTIME`? Adding underscore between each word may help.

pratsingms (approved):

seantleonard (changes requested):

src/Core/Resolvers/IQueryExecutor.cs
rohkhann: I went about it this way -> Total execution time = graphql overhead + time taken to execute the query (databaseoverhead). If we just calculate database overhead, then total-database overhead gives us the GQL overhead. This seemed easier than determining specific execution times in different places.I went about it this way -> Total execution time = graphql overhead + time taken to execute the query (databaseoverhead). If we just calculate database overhead, then total-database overhead gives us the GQL overhead. This seemed easier than determining specific execution times in different places.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: yes let me fix.

seantleonard:
per pr description on testing: 
> Since this is only a timer addition and no logical change. existing unit tests should cover code.

You're adding key/value pair to the httpcontext. Wouldn't it be good to test to ensure that this value exists and is added as expected? there is no existing test coverage to check that the contents were updated with this new content.

src/Core/Resolvers/IQueryExecutor.cs
seantleonard: would probably be worth updating PR description to align with title of PR "Db query execution time" because these changes don't specifically test GraphQL overhead because you wouldn't be able to tell whether a slow down was due to GraphQL query processing DB perf changes.

rohkhann: updated.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -805,5 +832,21 @@ private void ValidateSize(long availableSizeBytes, long sizeToBeReadBytes)
...
+            if (httpContext != null)
+            {
+
+                if (!httpContext.Items.ContainsKey(TOTALDBEXECUTIONTIME))
+                {
+                    httpContext.Items[TOTALDBEXECUTIONTIME] = (long)0;
+                }
+
+                long currentTime = (long)httpContext.Items[TOTALDBEXECUTIONTIME]!;
+                httpContext.Items[TOTALDBEXECUTIONTIME] = currentTime + time;
```
vadeveka, line 848: How does this work when the execution involves multiple source calls?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -805,5 +832,21 @@ private void ValidateSize(long availableSizeBytes, long sizeToBeReadBytes)
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorProcessingData);
             }
         }
+
+        private void AddDbExecutionTimeToMiddlewareContext(long time)
```
seantleonard, line 836: per pr description on testing: 
> Since this is only a timer addition and no logical change. existing unit tests should cover code.

You're adding key/value pair to the httpcontext. Wouldn't it be good to test to ensure that this value exists and is added as expected? there is no existing test coverage to check that the contents were updated with this new content.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -221,6 +235,11 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
                     }
                 }
             });
+
+            queryExecutionTimer.Stop();
+            AddDbExecutionTimeToMiddlewareContext(queryExecutionTimer.ElapsedMilliseconds);
```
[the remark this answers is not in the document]
rohkhann: Yes this makes sense. Ill wrap a try catch around it.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -805,5 +853,23 @@ private void ValidateSize(long availableSizeBytes, long sizeToBeReadBytes)
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorProcessingData);
             }
         }
+
+        internal virtual void AddDbExecutionTimeToMiddlewareContext(long time)
+        {
+            HttpContext? httpContext = HttpContextAccessor.HttpContext;
+            if (httpContext != null)
+            {
+                lock (_httpContextLock)
```
vadeveka, line 862: NIT: we should probably add some code comments to explain need of lock for future reference

vadeveka (approved):

src/Core/Resolvers/QueryExecutor.cs
rohkhann: added

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -178,49 +200,67 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
...
+                });
+            }
+            catch
+            {
+                queryExecutionTimer.Stop();
+                AddDbExecutionTimeToMiddlewareContext(queryExecutionTimer.ElapsedMilliseconds);
+                throw;
+            }
+
+            queryExecutionTimer.Stop();
```
seantleonard, line 260: Any way to de-dupe this code?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -243,13 +283,17 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
...
+            queryExecutionTimer.Start();
             await conn.OpenAsync();
             DbCommand cmd = PrepareDbCommand(conn, sqltext, parameters, httpContext, dataSourceName);
 
             try
             {
                 using DbDataReader dbDataReader = ConfigProvider.GetConfig().MaxResponseSizeLogicEnabled() ?
                     await cmd.ExecuteReaderAsync(CommandBehavior.SequentialAccess) : await cmd.ExecuteReaderAsync(CommandBehavior.CloseConnection);
+                queryExecutionTimer.Stop();
+                AddDbExecutionTimeToMiddlewareContext(queryExecutionTimer.ElapsedMilliseconds);
```
seantleonard, line 296: Your code block above had `Timer.Stop()` in the catch block. Is this `.Stop()` missed if an exception is thrown when executing query?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -805,5 +853,24 @@ private void ValidateSize(long availableSizeBytes, long sizeToBeReadBytes)
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorProcessingData);
             }
         }
+
+        internal virtual void AddDbExecutionTimeToMiddlewareContext(long time)
+        {
+            HttpContext? httpContext = HttpContextAccessor?.HttpContext;
+            if (httpContext != null)
+            {
+                // locking is because we could have multiple queries in a single http request and each query will pr processed in parallel leading to concurrent access of the httpContext.Items.
```
seantleonard, line 862: "each query will pr processed" -> pr?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -805,5 +853,24 @@ private void ValidateSize(long availableSizeBytes, long sizeToBeReadBytes)
...
+                lock (_httpContextLock)
+                {
+                    if (!httpContext.Items.ContainsKey(TOTALDBEXECUTIONTIME))
+                    {
+                        httpContext.Items[TOTALDBEXECUTIONTIME] = (long)0;
+                    }
+
+                    long currentTime = (long)httpContext.Items[TOTALDBEXECUTIONTIME]!;
+                    httpContext.Items[TOTALDBEXECUTIONTIME] = currentTime + time;
```
seantleonard, line 871: Can this be simplified so that there is only 1 read and 1 write? (e.g. TryGetValue() https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2.trygetvalue?view=net-8.0)
-> if key doesn't exist (First instance of reading httpContext.Items), then you can just write to `Items` a single time. Avoiding the second read/write.
-> if key exists -> no need to use null forgiving operator.

seantleonard (approved):

src/Core/Resolvers/QueryExecutor.cs
rohkhann: Fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: typo sorry, fixed.

src/Core/Resolvers/QueryExecutor.cs
rohkhann: used a finally statement to de-dupe.

seantleonard (approved):

PR #2330, opened 2024-08-11
Add new cli configure options

### LINKED ISSUE #2295: [Enhancement] `dab configure --data-source` in CLI
Support the following options for `dab configure` per #2251

| Configuration File Property Path                                        | CLI Flag                                                           | Data Type | Nullable |
|-------------------------------------------------------------------------|--------------------------------------------------------------------|-----------|----------|
| data-source.<br/>database-type                                          | <nobr>--data-source.database-type</nobr>                           | String: `MSSQL`, `PostgreSQL`, `CosmosDB`, `MySQL` | ❌ |
| data-source.<br/>connection-string                                      | <nobr>--data-source.connection-string</nobr>                       | String    | ❌       |
| data-source.<br/>options.database                                       | <nobr>--data-source.options.database</nobr>                        | String    | ✅       |
| data-source.<br/>options.container                                      | <nobr>--data-source.options.container</nobr>                       | String    | ✅       |
| data-source.<br/>options.schema                                         | <nobr>--data-source.options.schema</nobr>                          | String    | ✅       |
| data-source.<br/>options.set-session-context                            | <nobr>--data-source.options.set-session-context</nobr>             | Boolean: `true`, `false` (default: `false`) | ✅ |

abhishekkumams:
## Why make this change?

- Closes #2295 
  - This change will allow user to update data-source options.
  - Also, it makes CLI more powerful and developer friendly.

## What is this change?

- Introducing new options for configure command to update data-source options.

| Configuration File Property Path                                        | CLI Flag                                                           | Data Type | Nullable |
|-------------------------------------------------------------------------|--------------------------------------------------------------------|-----------|----------|
| data-source.<br/>database-type                                          | <nobr>--data-source.database-type</nobr>                           | String: `MSSQL`, `PostgreSQL`, `CosmosDB_NoSQL`, `MySQL` | ✅  |
| data-source.<br/>connection-string                                      | <nobr>--data-source.connection-string</nobr>                       | String    | ✅        |
| data-source.<br/>options.database                                       | <nobr>--data-source.options.database</nobr>                        | String    | ✅       |
| data-source.<br/>options.container                                      | <nobr>--data-source.options.container</nobr>                       | String    | ✅       |
| data-source.<br/>options.schema                                         | <nobr>--data-source.options.schema</nobr>                          | String    | ✅       |
| data-source.<br/>options.set-session-context                            | <nobr>--data-source.options.set-session-context</nobr>             | Boolean: `true`, `false` (default: `false`) | ✅   |

## NOTES:
Certain properties that are specific to database can only be updated with the correct databaseType.
For Example: If the database type initially was MSSQL, and we give cosmosDBoptions without specifying the new dbType as cosmosDB_NoSQL, the command will fail.
For better usability, we will also allow users to first update the dbType to cosmosDB_NoSQL and then in a new command provide the cosmosDB specifig options, since using all the options in one command can be very long and errornous. 

## How was this tested?

- [x] End-End Tests
- [x] Unit Tests

## Sample Request(s)

- To update to mssql
`dab configure --data-source.database-type mssql --data-source.options.set-session-context true`

- To update to cosmosDB_NoSQL
   #### Single command
  `dab configure --data-source.database-type cosmosdb_nosql --data-source.options.database testdbname --data-source.options.schema testschema.gql`

  #### Break the command
  1. `dab configure --data-source.database-type cosmosdb_nosql`
  2. `dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql`

### DIFF
--- a/src/Cli.Tests/ConfigureOptionsTests.cs
+++ b/src/Cli.Tests/ConfigureOptionsTests.cs
@@ -153,6 +153,7 @@ namespace Cli.Tests
 
         /// <summary>
         /// Tests the update of the database type in the runtime config.
+        /// dab configure `--data-source.database-type {dbType}`
         /// This method verifies that the database type can be updated to various valid values, including different cases,
         /// and ensures that the config file is correctly modified and parsed after the update.
         /// </summary>
@@ -173,14 +174,16 @@ namespace Cli.Tests
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
             Assert.IsNotNull(config.Runtime);
 
-            // Act
             ConfigureOptions options = new(
                 dataSourceDatabaseType: dbType,
                 config: TEST_RUNTIME_CONFIG_FILE
             );
-            Assert.IsTrue(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+
+            // Act
+            bool isSuccess = TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!);
 
             // Assert
+            Assert.IsTrue(isSuccess);
             string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out config));
             Assert.IsNotNull(config.Runtime);
@@ -191,7 +194,8 @@ namespace Cli.Tests
         /// Tests the update of the database type from CosmosDB_NoSQL to MSSQL in the runtime config.
         /// This method verifies that the database type can be changed from CosmosDB_NoSQL to MSSQL and that the 
         /// specific MSSQL option 'set-session-context' is correctly added to the configuration and the specific
-        /// cosmosDB options are removed. 
+        /// cosmosDB options are removed.
+        /// Command: dab configure --data-source.database-type mssql --data-source.options.set-session-context true 
         /// </summary>
         [TestMethod]
         public void TestDatabaseTypeUpdateCosmosDB_NoSQLToMSSQL()
@@ -204,15 +208,17 @@ namespace Cli.Tests
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_COSMOSDB_NOSQL_CONFIG, out RuntimeConfig? config));
             Assert.IsNotNull(config.Runtime);
 
-            // Act
             ConfigureOptions options = new(
                 dataSourceDatabaseType: "mssql",
                 dataSourceOptionsSetSessionContext: true,
                 config: TEST_RUNTIME_CONFIG_FILE
             );
-            Assert.IsTrue(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+
+            // Act
+            bool isSuccess = TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!);
 
             // Assert
+            Assert.IsTrue(isSuccess);
             string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out config));
             Assert.IsNotNull(config.Runtime);
@@ -227,6 +233,8 @@ namespace Cli.Tests
         /// Tests the update of the database type from MSSQL to CosmosDB_NoSQL in the runtime config.
         /// This method verifies that the database type can be changed from MSSQL to CosmosDB_NoSQL and that the 
         /// specific CosmosDB_NoSQL options such as database, container, and schema are correctly added to the config.
+        /// Command: dab configure --data-source.database-type cosmosdb_nosql
+        /// --data-source.options.database testdb --data-source.options.container testcontainer --data-source.options.schema testschema.gql
         /// </summary>
         [TestMethod]
         public void TestDatabaseTypeUpdateMSSQLToCosmosDB_NoSQL()
@@ -239,7 +247,6 @@ namespace Cli.Tests
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
             Assert.IsNotNull(config.Runtime);
 
-            // Act
             ConfigureOptions options = new(
                 dataSourceDatabaseType: "cosmosdb_nosql",
                 dataSourceOptionsDatabase: "testdb",
@@ -247,9 +254,12 @@ namespace Cli.Tests
                 dataSourceOptionsSchema: "testschema.gql",
                 config: TEST_RUNTIME_CONFIG_FILE
             );
-            Assert.IsTrue(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+
+            // Act
+            bool isSuccess = TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!);
 
             // Assert
+            Assert.IsTrue(isSuccess);
             string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out config));
             Assert.IsNotNull(config.Runtime);
@@ -276,14 +286,16 @@ namespace Cli.Tests
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
             Assert.IsNotNull(config.Runtime);
 
-            // Act
             ConfigureOptions options = new(
                 dataSourceDatabaseType: "invalid",
                 config: TEST_RUNTIME_CONFIG_FILE
             );
 
+            // Act
+            bool isSuccess = TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!);
+
             // Assert
-            Assert.IsFalse(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+            Assert.IsTrue(isSuccess);
         }
 
         /// <summary>
@@ -302,7 +314,6 @@ namespace Cli.Tests
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
             Assert.IsNotNull(config.Runtime);
 
-            // Act
             ConfigureOptions options = new(
                 dataSourceOptionsDatabase: "testdb",
                 dataSourceOptionsContainer: "testcontainer",
@@ -310,8 +321,11 @@ namespace Cli.Tests
                 config: TEST_RUNTIME_CONFIG_FILE
             );
 
+            // Act
+            bool isSuccess = TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!);
+
             // Assert
-            Assert.IsFalse(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+            Assert.IsTrue(isSuccess);
         }
 
         /// <summary>
@@ -330,15 +344,17 @@ namespace Cli.Tests
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
             Assert.IsNotNull(config.Runtime);
 
-            // Act
             ConfigureOptions options = new(
                 dataSourceDatabaseType: "mysql",
                 dataSourceOptionsSetSessionContext: true,
                 config: TEST_RUNTIME_CONFIG_FILE
             );
 
+            // Act
+            bool isSuccess = TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!);
+
             // Assert
-            Assert.IsFalse(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+            Assert.IsTrue(isSuccess);
         }
     }
 }
--- a/src/Cli.Tests/ConfigureOptionsTests.cs
+++ b/src/Cli.Tests/ConfigureOptionsTests.cs
@@ -150,5 +150,195 @@ namespace Cli.Tests
 
             Assert.AreEqual(newDepthLimit, config.Runtime?.GraphQL?.DepthLimit);
         }
+
+        /// <summary>
+        /// Tests the update of the database type in the runtime config.
+        /// This method verifies that the database type can be updated to various valid values, including different cases,
+        /// and ensures that the config file is correctly modified and parsed after the update.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("mssql", DisplayName = "Update the database type to MSSQL")]
+        [DataRow("MSSql", DisplayName = "Update the database type to MSSQL with different case")]
+        [DataRow("postgresql", DisplayName = "Update the database type to PostgreSQL")]
+        [DataRow("cosmosdb_nosql", DisplayName = "Update the database type to CosmosDB_NoSQL")]
+        [DataRow("cosmosdb_postgresql", DisplayName = "Update the database type to CosmosDB_PGSQL")]
+        [DataRow("mysql", DisplayName = "Update the database type to MySQL")]
+        public void TestDatabaseTypeUpdate(string dbType)
+        {
+            // Arrange
+            _fileSystem!.AddFile(TEST_RUNTIME_CONFIG_FILE, new MockFileData(INITIAL_CONFIG));
+
+            Assert.IsTrue(_fileSystem!.File.Exists(TEST_RUNTIME_CONFIG_FILE));
+
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
+            Assert.IsNotNull(config.Runtime);
+
+            // Act
+            ConfigureOptions options = new(
+                dataSourceDatabaseType: dbType,
+                config: TEST_RUNTIME_CONFIG_FILE
+            );
+            Assert.IsTrue(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+
+            // Assert
+            string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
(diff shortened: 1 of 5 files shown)


src/Cli.Tests/ConfigureOptionsTests.cs
seantleonard, line 1: in pr description you have:
> For better usability, we will also allow users to first update the dbType to cosmosDB_NoSQL and then in a new command provide the cosmosDB specifig options, since using all the options in one command can be very long and errornous.

and then an example to break up the command: 
> 1. `dab configure --data-source.database-type cosmosdb_nosql`
> 2. `dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql`

So, user must remember to run the second command, otherwise their config is in a bad state? I'm on the fence about this because this implies that execution of a single DAB cli command *could* result in your config being invalid.

src/Cli/Commands/ConfigureOptions.cs
```diff
@@ -19,13 +19,43 @@ namespace Cli.Commands
...
+        public string? DataSourceConnectionString { get; }
+
+        [Option("data-source.options.database", Required = false, HelpText = "Database name for Cosmos DB for NoSql.")]
+        public string? DataSourceOptionsDatabase { get; }
+
+        [Option("data-source.options.container", Required = false, HelpText = "Container name for Cosmos DB for NoSql.")]
+        public string? DataSourceOptionsContainer { get; }
+
+        [Option("data-source.options.schema", Required = false, HelpText = "Schema path for Cosmos DB for NoSql.")]
+        public string? DataSourceOptionsSchema { get; }
```
seantleonard, line 54: @sajeetharan, can you provide feedback on this help text for Cosmos? if i recall, we need to specify Azure Cosmos DB instead of Cosmos DB for NoSql?

src/Cli/Commands/ConfigureOptions.cs
```diff
@@ -19,13 +19,43 @@ namespace Cli.Commands
...
+
+        [Option("data-source.options.set-session-context", Required = false, HelpText = "Set session context. Allowed values: true, false (default: false).")]
```
seantleonard, line 56:
```suggestion
        [Option("data-source.options.set-session-context", Required = false, HelpText = "Enable session context. Allowed values: true (default), false.")]
```

default is true per datasource.cs line 87

src/Cli/ConfigGenerator.cs
```diff
@@ -535,6 +541,98 @@ public static bool TryConfigureSettings(ConfigureOptions options, FileSystemRunt
             return WriteRuntimeConfigToFile(runtimeConfigFile, runtimeConfig, fileSystem);
         }
 
+        /// <summary>
+        /// Configures the data source options for the runtimeconfig based on the provided options.
+        /// This method updates the database type, connection string, and other database-specific options in the config file.
+        /// It validates the provided database type and ensures that options specific to certain database types are correctly applied.
+        /// If any validation fails, it logs an error and returns false.
```
seantleonard, line 548:
```suggestion
        /// When validation fails, this function logs the validation errors and returns false.
```

src/Cli/ConfigGenerator.cs
```diff
@@ -535,6 +541,98 @@ public static bool TryConfigureSettings(ConfigureOptions options, FileSystemRunt
...
+            {
+                AddCosmosDbOptions(dbOptions, options, namingPolicy);
+            }
+            else if (!string.IsNullOrWhiteSpace(options.DataSourceOptionsDatabase)
+                    || !string.IsNullOrWhiteSpace(options.DataSourceOptionsContainer)
+                    || !string.IsNullOrWhiteSpace(options.DataSourceOptionsSchema))
+            {
+                _logger.LogError("Database, Container, and Schema options are only applicable for CosmosDB_NoSQL database type.");
+                return false;
+            }
```
seantleonard, line 589: when changing dbtype from sql to cosmos, users will need to change and provid cosmos options in a single command, right? Maybe i haven't read far enough but it seems like the 2 step command approach is prevented by this condition.

src/Cli/ConfigGenerator.cs
```diff
@@ -535,6 +541,98 @@ public static bool TryConfigureSettings(ConfigureOptions options, FileSystemRunt
...
+
+                dbOptions.Add(namingPolicy.ConvertName(nameof(MsSqlOptions.SetSessionContext)), options.DataSourceOptionsSetSessionContext.Value);
+            }
+
+            dbOptions = dbOptions.IsNullOrEmpty() ? null : dbOptions;
+            DataSource dataSource = new(dbType, dataSourceConnectionString, dbOptions);
+            runtimeConfig = runtimeConfig with { DataSource = dataSource };
+
+            return true;
+        }
```
seantleonard, line 607: Should returning true be conditioned on runtimeconfig != null to honor the promise you make in the function signature and description?

src/Cli/ConfigGenerator.cs
```diff
@@ -12,6 +12,7 @@
 using Azure.DataApiBuilder.Service;
 using Cli.Commands;
 using Microsoft.Extensions.Logging;
+using Microsoft.IdentityModel.Tokens;
```
seantleonard, line 15: why do you need this? Did i miss code which needs this?

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,195 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
 
             Assert.AreEqual(newDepthLimit, config.Runtime?.GraphQL?.DepthLimit);
         }
+
+        /// <summary>
+        /// Tests the update of the database type in the runtime config.
```
seantleonard, line 155:
```suggestion
        /// Tests running "dab config -update"
```
something like the above to demonstrate which cli command/usage is tested.

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,195 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
...
+
+            // Assert
+            string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out config));
+            Assert.IsNotNull(config.Runtime);
+            Assert.AreEqual(config.DataSource.DatabaseType, Enum.Parse<DatabaseType>(dbType, ignoreCase: true));
+        }
+
+        /// <summary>
+        /// Tests the update of the database type from CosmosDB_NoSQL to MSSQL in the runtime config.
```
seantleonard, line 191: can you include which command(s) is used? one or two commands?

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,195 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
...
+        /// This method verifies that the database type can be changed from CosmosDB_NoSQL to MSSQL and that the 
+        /// specific MSSQL option 'set-session-context' is correctly added to the configuration and the specific
+        /// cosmosDB options are removed. 
+        /// </summary>
+        [TestMethod]
+        public void TestDatabaseTypeUpdateCosmosDB_NoSQLToMSSQL()
+        {
+            // Arrange
+            _fileSystem!.AddFile(TEST_RUNTIME_CONFIG_FILE, new MockFileData(INITIAL_COSMOSDB_NOSQL_CONFIG));
```
seantleonard, line 200: does this overwrite a pre-existing file? i'd expect so to mitigate tests stepping over eachother.

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,195 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
...
+                config: TEST_RUNTIME_CONFIG_FILE
+            );
+            Assert.IsTrue(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
+
+            // Assert
+            string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out config));
+            Assert.IsNotNull(config.Runtime);
+            Assert.AreEqual(config.DataSource.DatabaseType, DatabaseType.MSSQL);
+            Assert.AreEqual(config.DataSource.Options!.GetValueOrDefault("set-session-context", false), true);
```
seantleonard, line 220: what does the `false` value do here: `Options!.GetValueOrDefault("set-session-context", false), true);`

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,195 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
...
+            Assert.IsFalse(config.DataSource.Options!.ContainsKey("database"));
+            Assert.IsFalse(config.DataSource.Options!.ContainsKey("container"));
+            Assert.IsFalse(config.DataSource.Options!.ContainsKey("schema"));
+        }
+
+        /// <summary>
+        /// Tests the update of the database type from MSSQL to CosmosDB_NoSQL in the runtime config.
+        /// This method verifies that the database type can be changed from MSSQL to CosmosDB_NoSQL and that the 
+        /// specific CosmosDB_NoSQL options such as database, container, and schema are correctly added to the config.
```
seantleonard, line 229: using which commands, and how many? 1, 2?

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,195 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
...
+            Assert.IsNotNull(config.Runtime);
+
+            // Act
+            ConfigureOptions options = new(
+                dataSourceDatabaseType: "invalid",
+                config: TEST_RUNTIME_CONFIG_FILE
+            );
+
+            // Assert
+            Assert.IsFalse(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
```
seantleonard, line 286: The "act" the last line where `TryConfigureSettings` is used

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,195 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
...
+            // Act
+            ConfigureOptions options = new(
+                dataSourceOptionsDatabase: "testdb",
+                dataSourceOptionsContainer: "testcontainer",
+                dataSourceOptionsSchema: "testschema.gql",
+                config: TEST_RUNTIME_CONFIG_FILE
+            );
+
+            // Assert
+            Assert.IsFalse(TryConfigureSettings(options, _runtimeConfigLoader!, _fileSystem!));
```
seantleonard, line 314: `TryConfigureSettings` is the "act"

src/Cli.Tests/ConfigureOptionsTests.cs
seantleonard: this is the case for most of these tests. please re-check all of them

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -1107,4 +1107,32 @@ public void TestRestRequestBodyStrictMode(bool includeRestRequestBodyStrictFlag,
         Assert.IsNotNull(runtimeConfig.Runtime.Rest);
         Assert.AreEqual(isRequestBodyStrict, runtimeConfig.Runtime.Rest.RequestBodyStrict);
     }
+
+    /// <summary>
+    /// Tests that the CLI can update the database type in the configuration file and is case insensitive.
```
seantleonard, line 1112: sample command? dab configure ...

seantleonard (changes requested):
Added some suggestions and clarification on tests. also some clarifying questions.

src/Cli.Tests/ConfigureOptionsTests.cs
abhishekkumams: it's the default value, it will give false if the key is not found.

src/Cli.Tests/ConfigureOptionsTests.cs
abhishekkumams: they can do it one command as well. They're not mandated to do it in 2 commands. It's just we give them the flexibility that if they don't want to put everything in one command and choose to break it, we support it.

src/Cli/ConfigGenerator.cs
abhishekkumams: No, it won't be. Because if the user first changes the database, that change is persisted. so the dbType will be cosmosDB when the second command will be run.

src/Cli/ConfigGenerator.cs
abhishekkumams: sure.

src/Cli/ConfigGenerator.cs
abhishekkumams: it's for using methods from CollectionUtils.
alt text

src/Cli.Tests/ConfigureOptionsTests.cs
abhishekkumams: yes, it overwrites the pre-existing one.

src/Cli/Commands/ConfigureOptions.cs
sajeetharan: Should be Azure Cosmos DB for NOSQL (full caps)

abhishekkumams: why full caps for NoSQL? for example:
image
image

Also i feel fullcaps is not that readable.

src/Cli.Tests/ConfigureOptionsTests.cs
seantleonard: > since using all the options in one command can be very long and errornous.
> It's just we give them the flexibility that if they don't want to put everything in one command and choose to break it, we support it.

would this reasoning be motivation for us to include Options short names? that way, developers can do something like:
`dab configure -dbname "myCosmosDbName" -dbType "CosmosDb" -container "cosmosContainer" -graphQLSchema "schema.gql"`

That way options are memorable and mitigate super verbose cmd line usage.

seantleonard (approved):

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -150,5 +150,211 @@ public void TestUpdateDepthLimitForGraphQL(int? newDepthLimit)
...
+        /// This method verifies that the database type can be changed from CosmosDB_NoSQL to MSSQL and that the 
+        /// specific MSSQL option 'set-session-context' is correctly added to the configuration and the specific
+        /// cosmosDB options are removed.
+        /// Command: dab configure --data-source.database-type mssql --data-source.options.set-session-context true 
+        /// </summary>
+        [TestMethod]
+        public void TestDatabaseTypeUpdateCosmosDB_NoSQLToMSSQL()
+        {
+            // Arrange
+            _fileSystem!.AddFile(TEST_RUNTIME_CONFIG_FILE, new MockFileData(INITIAL_COSMOSDB_NOSQL_CONFIG));
```
aaronburtle, line 204: Looks like shared code between the //arrange steps in these tests, could refactor this out to eliminate the duplication.

aaronburtle:
Question with respect to breaking the command up

`1. dab configure --data-source.database-type cosmosdb_nosql`
`2. dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql`

what happens if we put another command in between these 2, will it still accept the final command? If so, what happens if the command in between them changes the database-type to something other than cosmos?

so something like

`1. dab configure --data-source.database-type cosmosdb_nosql`
`2. dab configure --data-source.database-type mssql`
`3. dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql`



And then another case. If the 2nd command also included the database-type command

`1. dab configure --data-source.database-type cosmosdb_nosql`
`2.dab configure --data-source.database-type mssql --data-source.options.database testdbname --data-source.options.schema testschema.gql`

I presume that all of the above would fail but just want to double check.

aaronburtle:
> And then another case. If the 2nd command also included the database-type command
> 
> `1. dab configure --data-source.database-type cosmosdb_nosql` 
> `2.dab configure --data-source.database-type mssql --data-source.options.database testdbname --data-source.options.schema testschema.gql`
> 
> I presume that all of the above would fail but just want to double check.


Follow up question, for the case that I quoted. Does order of the options within the command itself here matter? So for example what if we send these 2 commands, and note that the changing of the dbtype to mssql happens at the end of the 2nd command


`1. dab configure --data-source.database-type cosmosdb_nosql` 
`2.dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql` --data-source.database-type mssql 


Does this fail?

aaronburtle (approved):
LGTM! Just a comment about some code duplication in the tests, and then a question about some edge cases to make sure they get handled.

abhishekkumams:
> Question with respect to breaking the command up
> 
> `1. dab configure --data-source.database-type cosmosdb_nosql` `2. dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql`
> 
> what happens if we put another command in between these 2, will it still accept the final command? If so, what happens if the command in between them changes the database-type to something other than cosmos?
> 
> so something like
> 
> `1. dab configure --data-source.database-type cosmosdb_nosql` `2. dab configure --data-source.database-type mssql` `3. dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql`
> 
> And then another case. If the 2nd command also included the database-type command
> 
> `1. dab configure --data-source.database-type cosmosdb_nosql` `2.dab configure --data-source.database-type mssql --data-source.options.database testdbname --data-source.options.schema testschema.gql`
> 
> I presume that all of the above would fail but just want to double check.

Yes, both the commands will fail. Every command gets checked for correct database-type. if that's changed in midway. it will error it out.

abhishekkumams:
> > And then another case. If the 2nd command also included the database-type command
> > `1. dab configure --data-source.database-type cosmosdb_nosql`
> > `2.dab configure --data-source.database-type mssql --data-source.options.database testdbname --data-source.options.schema testschema.gql`
> > I presume that all of the above would fail but just want to double check.
> 
> Follow up question, for the case that I quoted. Does order of the options within the command itself here matter? So for example what if we send these 2 commands, and note that the changing of the dbtype to mssql happens at the end of the 2nd command
> 
> `1. dab configure --data-source.database-type cosmosdb_nosql` `2.dab configure --data-source.options.database testdbname --data-source.options.schema testschema.gql --data-source.database-type mssql`
> 
> Does this fail?

yes, because the options provided do not belong to mssql type.

Issue #2321, opened 2024-08-09
Replace copies of RuntimeConfig with RuntimeConfigProvider


aaronburtle:
There are classes we currently use, such as the SqlMetaDataProvider, OpenApiDocumentor, and CostmosSqlMetaDataProvider, that save internal copies of the RuntimeConfig. These classes instead should have a copy of the RuntimeConfigProvider, and when a runtimeconfig is needed they should instead get that runtimeconfig from the provider as needed. This will cause the runtimeconfig to always be the most up to date copy.

seantleonard:
@aaronburtle the design doc is not publicly accessible. Doesn't need to be referenced here unless you end up created a markdown file to include in this repo

aaronburtle:
> @aaronburtle the design doc is not publicly accessible. Doesn't need to be referenced here unless you end up created a markdown file to include in this repo

thanks for pointing that out Sean, ill remove from these issues.

PR #2337, opened 2024-08-15
Replace internal references to RuntimeConfig with immutable copies to maintain consistent states

aaronburtle:
## Why make this change?

closes https://github.com/Azure/data-api-builder/issues/2321

## What is this change?

Hot-Reload will mean refreshing the `RuntimeConfig` that the `RuntimeConfigProvider` retrieves. To make sure that we have consistency throughout DAB, we refactor how certain classes store the `RuntimeConfig` to guarantee consistency within that class and with DAB. For the `CosmosSqlMetadataProvider` this means that we make a copy of the Entities at the time of construction and store that copy for use within the class. We also save the state of the `isDevelopmentMode` function from the runtimeConfig to keep this class consistent to the `RuntimeConfig` at time of construction. This guarantees that no other class can modify those Entities and we will have a consistent state throughout the life of the `CosmosSqlMetadataProvider`. For the `OpenApiDocumentor` we change the class to get a fresh `RuntimeConfig` when it's public facing functions are called so that the output of those functions matches the current state of DAB.

## How was this tested?

Against our current test suite. This will be tested from a hot-reload perspective with the hot-reload integration tests.

## Sample Request(s)

Any request will work.

### DIFF
--- a/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
@@ -59,7 +59,9 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
             RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
             _runtimeConfigProvider = runtimeConfigProvider;
             _fileSystem = fileSystem;
-            // Make a copy of the runtime entities to guarantee consistency
+            // Many classes have references to the RuntimeConfig, therefore to guarantee
+            // that the Runtime Entities are not mutated by another class we make a copy of them
+            // to store internally.
             _runtimeConfigEntities = new RuntimeEntities(runtimeConfig.Entities.Entities);
             _databaseType = runtimeConfig.DataSource.DatabaseType;
 
--- a/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
@@ -83,7 +83,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
             }
 
             ParseSchemaGraphQLFieldsForGraphQLType();
-            ParseSchemaGraphQLFieldsForJoins();
+            ParseSchemaGraphQLFieldsForJoins(runtimeConfig);
 
             InitODataParser();
         }
@@ -141,7 +141,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
         /// EntityWithJoins dictionary indicates the paths for each entity. There "Planet" has one path i.e. "c" on the other hand Star has 3 paths.with one join statement.
         /// This information is getting used to resolve DB Policy and generate cosmos DB sql query conditions for them.
         /// </summary>
-        private void ParseSchemaGraphQLFieldsForJoins()
+        private void ParseSchemaGraphQLFieldsForJoins(RuntimeConfig runtimeConfig)
         {
             IncrementingInteger tableCounter = new();
 
@@ -162,7 +162,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
                 {
                     string modelName = GraphQLNaming.ObjectTypeToEntityName(node);
 
-                    AssertIfEntityIsAvailableInConfig(modelName);
+                    AssertIfEntityIsAvailableInConfig(runtimeConfig, modelName);
 
                     if (EntityWithJoins.TryGetValue(modelName, out List<EntityDbPolicyCosmosModel>? entityWithJoins))
                     {
@@ -178,7 +178,12 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
                            });
                     }
 
-                    ProcessSchema(node.Fields, schemaDefinitions, CosmosQueryStructure.COSMOSDB_CONTAINER_DEFAULT_ALIAS, tableCounter);
+                    ProcessSchema(
+                        fields: node.Fields,
+                        schemaDocument: schemaDefinitions,
+                        currentPath: CosmosQueryStructure.COSMOSDB_CONTAINER_DEFAULT_ALIAS,
+                        tableCounter: tableCounter,
+                        runtimeConfig: runtimeConfig);
                 }
             }
         }
@@ -205,6 +210,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
             Dictionary<string, ObjectTypeDefinitionNode> schemaDocument,
             string currentPath,
             IncrementingInteger tableCounter,
+            RuntimeConfig runtimeConfig,
             EntityDbPolicyCosmosModel? parentEntity = null,
             HashSet<string>? visitedEntities = null)
         {
@@ -226,7 +232,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
 
                 string entityType = field.Type.NamedType().Name.Value;
 
-                AssertIfEntityIsAvailableInConfig(entityType);
+                AssertIfEntityIsAvailableInConfig(runtimeConfig, entityType);
 
                 // If the entity is already visited, then it is a circular reference
                 if (!trackerForFields.Add(entityType))
@@ -284,15 +290,16 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
                     schemaDocument: schemaDocument,
                     currentPath: isArrayType ? $"{alias}" : $"{currentPath}.{field.Name.Value}",
                     tableCounter: tableCounter,
+                    runtimeConfig: runtimeConfig,
                     parentEntity: isArrayType ? currentEntity : null,
                     visitedEntities: trackerForFields);
             }
         }
 
-        private void AssertIfEntityIsAvailableInConfig(string entityName)
+        private static void AssertIfEntityIsAvailableInConfig(RuntimeConfig runtimeConfig, string entityName)
         {
             // If the entity is not present in the runtime config, throw an exception as we are expecting all the entities to be present in the runtime config.
-            if (!_runtimeConfigProvider.GetConfig().Entities.ContainsKey(entityName))
+            if (!runtimeConfig.Entities.ContainsKey(entityName))
             {
                 throw new DataApiBuilderException(
                     message: $"The entity '{entityName}' was not found in the runtime config.",
--- a/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
@@ -24,8 +24,8 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
         private readonly IFileSystem _fileSystem;
         private readonly DatabaseType _databaseType;
         private CosmosDbNoSQLDataSourceOptions _cosmosDb;
-        private readonly RuntimeConfigProvider _runtimeConfigProvider;
         private readonly RuntimeEntities _runtimeConfigEntities;
+        private readonly bool _isDevelopmentMode;
         private Dictionary<string, string> _partitionKeyPaths = new();
 
         /// <summary>
@@ -63,6 +63,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
             // that the Runtime Entities are not mutated by another class we make a copy of them
             // to store internally.
             _runtimeConfigEntities = new RuntimeEntities(runtimeConfig.Entities.Entities);
+            _isDevelopmentMode = runtimeConfig.IsDevelopmentMode();
             _databaseType = runtimeConfig.DataSource.DatabaseType;
 
             CosmosDbNoSQLDataSourceOptions? cosmosDb = runtimeConfig.DataSource.GetTypedOptions<CosmosDbNoSQLDataSourceOptions>();
@@ -603,7 +604,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
         // state on development mode in case a hot-reload
         public bool IsDevelopmentMode()
         {
-            return _runtimeConfigProvider.GetConfig().IsDevelopmentMode();
+            return _isDevelopmentMode;
         }
 
         public bool TryGetExposedFieldToBackingFieldMap(string entityName, [NotNullWhen(true)] out IReadOnlyDictionary<string, string>? mappings)
--- a/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
@@ -24,7 +24,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
         private readonly IFileSystem _fileSystem;
         private readonly DatabaseType _databaseType;
         private CosmosDbNoSQLDataSourceOptions _cosmosDb;
-        private readonly RuntimeConfig _runtimeConfig;
+        private readonly RuntimeConfigProvider _runtimeConfigProvider;
         private Dictionary<string, string> _partitionKeyPaths = new();
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -291,7 +292,7 @@ private void ProcessSchema(
         private void AssertIfEntityIsAvailableInConfig(string entityName)
         {
             // If the entity is not present in the runtime config, throw an exception as we are expecting all the entities to be present in the runtime config.
-            if (!_runtimeConfig.Entities.ContainsKey(entityName))
+            if (!_runtimeConfigProvider.GetConfig().Entities.ContainsKey(entityName))
```
seantleonard, line 295: Take a look at the callstack of where this function is called:
- Constructor [Contains `RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();`]
  - `ParseSchemaGraphQLFieldsForJoins(...)`
    - `AssertIfEntityIsAvailableInConfig(...)`

It doesn't seem correct to again call `runtimeConfigProvider.GetConfig()` in the middle of an operation when the top-level calling function has already acquired the current instance of the runtime config. 

If the runtime config changes in the middle of the operation, we should not change which runtimeconfig we are using in the current operation. The operation will need to start over again because state will be inconsistent.

seantleonard (changes requested):
Need to address your fetching of a runtimeconfig object multiple times in a single operation. as-is, this approach will result in OpenApiDocumentor and CosmosSqlMetadataProvider entering inconsistent states (corrupt) if runtime config changes in the middle of Constructor operations or CreateDocument operations.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -611,7 +612,7 @@ private Dictionary<OperationType, bool> GetConfiguredRestOperations(string entit
 
             if (dbObject.SourceType == EntitySourceType.StoredProcedure)
             {
-                Entity entity = _runtimeConfig.Entities[entityName];
+                Entity entity = _runtimeConfigProvider.GetConfig().Entities[entityName];
```
seantleonard, line 615: This is another location where you are mid-procedure and pulling a new version of the runtimeconfig. This presents opportunity for different sections of OpenAPI doc to contain contents of two different runtimeconfig objects.

NOTE: It may not be appropriate to pass the entire runtimeconfig object. Looking at this code, can you see what object/element from the runtimeconfig is needed here?

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -842,7 +843,7 @@ private static bool IsRequestBodyRequired(SourceDefinition sourceDef, bool consi
         private string GetEntityRestPath(string entityName)
         {
             string entityRestPath = entityName;
-            EntityRestOptions entityRestSettings = _runtimeConfig.Entities[entityName].Rest;
+            EntityRestOptions entityRestSettings = _runtimeConfigProvider.GetConfig().Entities[entityName].Rest;
```
seantleonard, line 846: This is another location where you are mid-procedure and pulling a new version of the runtimeconfig. This presents opportunity for different sections of OpenAPI doc to contain contents of two different runtimeconfig objects.

NOTE: It may not be appropriate to pass the entire runtimeconfig object. Looking at this code, consider whether RuntimeConfig object even makes sense to use as argument:
- you may only need to pass in runtimeConfig.Entities[entityName].Rest -> in which case the function param should have input of `EntityRestOptions`

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -169,16 +170,16 @@ public void CreateDocument()
         /// "/EntityName"
         /// </example>
         /// <returns>All possible paths in the DAB engine's REST API endpoint.</returns>
-        private OpenApiPaths BuildPaths()
+        private OpenApiPaths BuildPaths(RuntimeConfig runtimeConfig)
```
seantleonard, line 173: does it make sense to pass in the entire runtimeconfig object? from the usage in this function, seems like there are only two properties of `runtimeConfig` that are used here and those seem to fit better in function signature.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -570,7 +578,7 @@ public string GetEntityName(string graphQLType)
             }
 
             // Fallback to looking at the singular name of the entity.
-            foreach ((string _, Entity entity) in _runtimeConfig.Entities)
+            foreach ((string _, Entity entity) in _runtimeConfigProvider.GetConfig().Entities)
```
seantleonard, line 581: Do you need to fetch the `runtimeConfig` object a second time? you already fetch at the beginning of this function.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -55,12 +55,13 @@ public class CosmosSqlMetadataProvider : ISqlMetadataProvider
...
-            _runtimeConfig = runtimeConfigProvider.GetConfig();
+            RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
 
-            _databaseType = _runtimeConfig.DataSource.DatabaseType;
+            _fileSystem = fileSystem;
+            _runtimeConfigProvider = runtimeConfigProvider;
+            _databaseType = runtimeConfig.DataSource.DatabaseType;
 
-            CosmosDbNoSQLDataSourceOptions? cosmosDb = _runtimeConfig.DataSource.GetTypedOptions<CosmosDbNoSQLDataSourceOptions>();
+            CosmosDbNoSQLDataSourceOptions? cosmosDb = runtimeConfig.DataSource.GetTypedOptions<CosmosDbNoSQLDataSourceOptions>();
```
seantleonard, line 64: I'm not really following what the benefit is of changing this constructor's code from:
```csharp
_runtimeConfig = runtimeConfigProvider.GetConfig();
ParseSchemaGraphQLFieldsForJoins();
```
to
```csharp
_runtimeConfigProvider = runtimeConfigProvider;
ParseSchemaGraphQLFieldsForJoins(runtimeConfig);
```

This constructor is executed a single time during startup, so if you hot reload the config, wouldn't you need to create a new instance of `CosmosSqlMetadataProvider`? I don't see other references to code which calls `ParseSchemaGraphQLFieldsForJoins(...)`

This class is a good use case for understanding how your hot reload event notification system is going to work. When hot-reload occurs, this class just needs to update its stateful `_runtimeConfig` and run any helper functions that setup the remaining state of the metadata provider.

With that in mind, you wouldn't need to make the other changes you've added to this file. e.g. instead of pulling a new runtimeconfig object for each of this class's public functions, each function would, by-design, be using the stateful `_runtimeConfig` member to determine what data to return, which is guaranteed to be the latest considering an implemented mechanism to handle hot reload events.

seantleonard (changes requested):

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
aaronburtle: We spoke offline but just to add some context here; the idea was to be defensive for the possible case where we have a hot-reload happen without the cosmosMetaDataProvider being updated or replaced. However, since the fields that store states which are based on the runtimeconfig are all instantiated and set during construction, we should keep that state consistent through the life of the cosmosMetaDataProvider until a hot-reload happens. When the hot-reload happens we then completely update those states to match the new runtimeConfig, likely by creating an entirely new cosmosMetaDataProvider.

Now, to ensure that we have consistency, we will make a private, read only copy of the fields that we use from the runtimeconfig during construction. We also save the runtimeConfigProvider so that we can get up to date information for the public calls that use information from the runtimeConfig that does not set the state of the cosmosMetaDataProvider at time of construction, eg isDevelopmentMode()

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -957,9 +955,9 @@ private static OpenApiMediaType CreateResponseContainer(string responseObjectSch
         private Dictionary<string, OpenApiSchema> CreateComponentSchemas()
         {
             Dictionary<string, OpenApiSchema> schemas = new();
-
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
```
seantleonard, line 958: This function should mirror input arguments change that you made for `BuildPaths(...)` no need to get runtime config again in the middle of the operation

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -598,7 +598,7 @@ private static OpenApiParameter GetOpenApiQueryParameter(string name, string des
         /// <param name="entityName">Name of the entity.</param>
         /// <param name="dbObject">Database object metadata, indicating entity SourceType</param>
         /// <returns>Collection of OpenAPI OperationTypes and whether they should be created.</returns>
-        private Dictionary<OperationType, bool> GetConfiguredRestOperations(string entityName, DatabaseObject dbObject)
+        private static Dictionary<OperationType, bool> GetConfiguredRestOperations(string entityName, DatabaseObject dbObject, RuntimeEntities entities)
```
seantleonard, line 601: make sure to update comments above. 
Also, do the input arguments `entityName` and `runtimeEntities` make sense in the context of this function? What object is actually needed here?

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -55,12 +56,14 @@ public class CosmosSqlMetadataProvider : ISqlMetadataProvider
 
         public CosmosSqlMetadataProvider(RuntimeConfigProvider runtimeConfigProvider, IFileSystem fileSystem)
         {
+            RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
+            _runtimeConfigProvider = runtimeConfigProvider;
             _fileSystem = fileSystem;
-            _runtimeConfig = runtimeConfigProvider.GetConfig();
+            // Make a copy of the runtime entities to guarantee consistency
```
seantleonard, line 62: "to guarantee consistency"
guarantee consistency of what? I know we may have spoken separately about this but no one else has this context and we may forget as time goes on.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -590,9 +597,11 @@ public string GetDefaultSchemaName()
             return string.Empty;
         }
 
+        // Use the runtimeConfigProvider to ensure we have current up to date
+        // state on development mode in case a hot
         public bool IsDevelopmentMode()
         {
-            return _runtimeConfig.IsDevelopmentMode();
+            return _runtimeConfigProvider.GetConfig().IsDevelopmentMode();
```
seantleonard, line 604: if after hot-reload, you're intending to create a new CosmosSqlMetadataProvider, is this change needed? Same with associated function comment.

seantleonard (changes requested):

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
aaronburtle: We might not always be creating a new CosmosSqlMetadataProvider, and this just makes sure that no matter what happens outside of this class that we are always returning an answer that is consistent with the current mode

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
aaronburtle: ya we can remove those and just pass in the entity we want.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
seantleonard: That means CosmosSqlMetadataProvider will have inconsistent state: this class could have internal state with old config, and may have *some* helpers return newer config? That doesn't seem right.

You'll need to clarify when new `CosmosSqlMetadataProvider` objects are created and when they are not -> "We might not always be creating a new CosmosSqlMetadataProvider." A better solution here would be to save `_runtimeConfigProvider.GetConfig().IsDevelopmentMode();` as private state to the metadata provider, and update that value when a hotreload event is detected. 

Further analysis shows that having `MetadataProvider` classes expose an `IsDevelopment()` mode helper is... odd design. Please open a refactoring GitHub issue which instructs us to reconsider (and possibly remove) `IsDevelopmentMode()` from the metadataprovider classes. The inclusion of that function seems like a hacky workaround for getting access to "isDevelopmentMode" from code that doesn't have access to the RuntimeConfig -> e.g. SqlQueryStructure classes. Development mode is not set per metadata provider. It should still be fine to have private internal state tracking development mode, so metadata provider logic know whether to emit dev mode error messages or not. But exposing a public getter for a property of the runtime config doesn't make sense.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
seantleonard: Unresolving since this is still an issue.

seantleonard (changes requested):

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
aaronburtle: thanks had this committed but not pushed.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
seantleonard: Context from Aaron:
> It is a weird function because all it is doing is replicating the work that the provider is doing. The reason I favored asking the provider instead of saving that state internally is because at some point we might want to avoid having to refresh every single dependency on the runtimeconfig if those portions that we hot-reload are not related to the particular dependency in question. But in that case, this helper function still needs to return a consistent value with the runtimeconfig. Because this helper function doesn't depend on any internal state, but just asks the provider. It should be refactored out imo, as you suggested

This depends on your event notification design: 

- What and how many events will you emit?
- Dependencies can listen for events related to the runtime config properties they care about.

> Because this helper function doesn't depend on any internal state, but just asks the provider. It should be refactored out imo, as you suggested

Please open a refactoring GitHub issue which instructs us to reconsider (and possibly remove) IsDevelopmentMode() from the metadataprovider classes. Add that issue number here before resolving.

aaronburtle: My reasoning for leaving this was so that we don't add a dependency on the mode to this class with respect to hot-reloading the config when the class itself doesn't depend on that state. This class doesn't use the mode internally and this function is simply passing through what the `RuntimeConfigProvider`'s config says about mode.

So ultimately, if we optimize the hot-reload process to only refresh those dependencies that have changed in the config, it would be beneficial to not have things like this stored internally because the class itself doesn't depend on it, and so now we have added a dependency that we should not have to add in the case of a hot-reload. For now, this shouldn't matter because we will refresh all dependencies to start with.

I will save this state then on construction and we can just return that state, so that the `CosmosMetadataProvider`, but this does mean that this function is not returning the current runtimeconfig's development mode state, but whatever that state was when the `CosmosMetadataProvider` was constructed. And that is a bit weird because the `CosmosMetadataProvider` itself doesn't depend on this state.

will add a task to refactor this out.

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -55,12 +56,16 @@ public class CosmosSqlMetadataProvider : ISqlMetadataProvider
 
         public CosmosSqlMetadataProvider(RuntimeConfigProvider runtimeConfigProvider, IFileSystem fileSystem)
         {
+            RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
             _fileSystem = fileSystem;
-            _runtimeConfig = runtimeConfigProvider.GetConfig();
+            // Many classes have references to the RuntimeConfig, therefore to guarantee
+            // that the Runtime Entities are not mutated by another class we make a copy of them
+            // to store internally.
+            _runtimeConfigEntities = new RuntimeEntities(runtimeConfig.Entities.Entities);
```
Aniruddh25, line 64: are we doing this because we dont want the entities or isDevelopmentMode to be refreshed for CosmosSqlMetadataProvider?

Aniruddh25 (approved):
LGTM, few questions

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -192,7 +192,8 @@ private OpenApiPaths BuildPaths(RuntimeEntities entities, string defaultDataSour
 
                 // Entities which disable their REST endpoint must not be included in
                 // the OpenAPI description document.
-                if (entities.TryGetValue(entityName, out Entity? entity) && entity is not null)
+                Entity? entity;
+                if (entities.TryGetValue(entityName, out entity) && entity is not null)
```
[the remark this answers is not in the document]
aaronburtle: we can continue if the entity is null, but that has to be scoped outside of the if for it to be recognized in the call the `GetConfigureRestOperaitons()`

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -192,15 +192,19 @@ private OpenApiPaths BuildPaths(RuntimeEntities entities, string defaultDataSour
...
                     {
                         continue;
                     }
                 }
 
+                if (entity is null)
+                {
+                    continue;
+                }
+
```
seantleonard, line 207: this could just be in an "else" block

seantleonard (approved):

PR #2343, opened 2024-08-20
Fix context sql

### LINKED ISSUE #2341: [Bug]: MSSQL (and related) - EXEC sp_set_session_context is executed with @read_only = 1
### What happened?

Running on Analytics SQL Endpoint on Fabric

I was trying a row-level-security scenario, based on the one in the documentation.

So I did:
`"data-source": {
    "database-type": "DWSQL",
    "connection-string": "Data Source=XXXXXXXl",
    "options": {
      "set-session-context": true
    }
  },`

And then configured JWT AzureAD authentication with a custom role. This all works fine.

When I do a query now, just prior to executing the query, in the \`src\Core\Resolvers\MsSqlQueryExecutor.cs:222` we see the following:
`string statementToSetReadOnlyParam = "EXEC sp_set_session_context " + $"'{claimType}', " + paramName + ", @read_only = 1;";`

This is executed on `every` query execution from the frontend:

`string sessionParamsQuery = GetSessionParamsQuery(httpContext, parameters, dataSourceName);
cmd.CommandText = sessionParamsQuery + sqltext;`

I see two issues:
- This second request fails with `Cannot set key 'roles' in the session context. The key has been set as read_only for this session.` from the database, as the context is created with read_only=1
- If a second request comes in, from a different user it will re-use the existing connection, leading to possible race conditions if multiple users are querying since the connection is shared between all users. 

The second issue could be my misunderstading of how the backend SQL engine handles these cases.

### Version

1.2.10

Azure SQL

Container Apps

GraphQL

### Relevant log output

_No response_

### Code of Conduct

M4Al:
## Why make this change?

- Fixes #2341 (partly)
  - do not set the session context in SQL server with read_only = 1 as this prevents us from doing a second request on the same connection
  - keep a copy of the original 'roles' from the JWT token so we can use it on the SQL Filter Predicate to accurately implement row level security

## What is this change?

The changes itself are quite simple and should break nothing, see the diff :-)

## How was this tested?

I have not created/adapted any tests yet. Just looking for comments before I go there.

## Sample Request(s)

Sample of a JWT token (only the relevant part)
`{
  "aud": "api://ddcf6b31-5d01-407d-97cf-8efefc455d32",
  "iss": "https://sts.windows.net/9215c785-95c3-49b0-bdba-2062df5aedb5/",
  "roles": [
    "user",
    "Allow_Customer_OPS025235",
    "Allow_Customer_OPS004095"
  ],
  "ver": "1.0"
}`

X-MS-API-ROLE: user

before my change the extra 'roles' that do not match the X-MS-API-ROLE header would never reach the database context.
With my change you can do things like this in SQL Predicates to filter out only subsets of the data:

`CREATE FUNCTION dbo.ops_fact_order_Predicate(@CustomerNo varchar(max))
RETURNS TABLE
WITH SCHEMABINDING
AS RETURN SELECT 1 AS fn_securitypredicate_result
WHERE @CustomerNo in (
		select trim(replace(replace(replace([value], '"', ''), ']', ''), 'Allow_Customer_', '')) 
		from STRING_SPLIT ( 
			CAST(SESSION_CONTEXT(N'original_roles') as varchar(max)) 
			, ',' 
			, 0) 
			where trim(replace(replace([value], '"', ''), ']', '')) like 'Allow_Customer%'
		)

CREATE SECURITY POLICY dbo.ops_fact_order_Policy
ADD FILTER PREDICATE dbo.ops_fact_order_Predicate(CustomerNo)
ON [gold_ops].[ops_fact_order];`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
 # Version values referenced from https://hub.docker.com/_/microsoft-dotnet-aspnet
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0-cbl-mariner2.0. AS build
+FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build
 
 
 WORKDIR /src
 COPY [".", "./"]
-RUN dotnet build "./src/Service/Azure.DataApiBuilder.Service.csproj" -c Docker -o /out -r linux-x64
+RUN dotnet build "./src/Service/Azure.DataApiBuilder.Service.csproj" -f net8.0 -o /out -r linux-x64 --self-contained
 
-FROM mcr.microsoft.com/dotnet/aspnet:6.0-cbl-mariner2.0 AS runtime
+FROM mcr.microsoft.com/dotnet/aspnet:8.0 AS runtime
 
 COPY --from=build /out /App
 WORKDIR /App
--- a/src/Auth/Azure.DataApiBuilder.Auth.csproj
+++ b/src/Auth/Azure.DataApiBuilder.Auth.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
     <ImplicitUsings>enable</ImplicitUsings>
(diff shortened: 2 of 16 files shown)


M4Al:
@microsoft-github-policy-service agree company="CNext"

PR #2344, opened 2024-08-20
Fix Session Context Key set as Read_Only & maintain original roles from the JWT token

### LINKED ISSUE #2341: [Bug]: MSSQL (and related) - EXEC sp_set_session_context is executed with @read_only = 1

M4Al:
## Why make this change?
- Fixes #2341
  - The session context in SQL server is `read_only = 1` which prevents users from doing multiple requests on the same connection.
  - The row level security is not accurately implemented when using a JWT token.

## What is this change?
Changes the session context from `read_only = 1` to `read_only = 0` to allow multiple requests to be done in the same connection,
Creates a copy of the original 'roles' from the JWT token to use it on the SQL Filter Predicate to accurately implement row level security.

## How was this tested?
- [ ] Integration Tests
- [x] Unit Tests

Updated `AuthorizationResolver` tests to ensure the original roles copy is working properly.

## Sample Request(s)

Sample of a JWT token (only the relevant part)
```
{
  "aud": "api://ddcf6b31-5d01-407d-97cf-8efefc455d32",
  "iss": "https://sts.windows.net/9215c785-95c3-49b0-bdba-2062df5aedb5/",
  "roles": [
    "user",
    "Allow_Customer_OPS025235",
    "Allow_Customer_OPS004095"
  ],
  "ver": "1.0"
}
```

X-MS-API-ROLE: user

before my change the extra 'roles' that do not match the X-MS-API-ROLE header would never reach the database context.
With my change you can do things like this in SQL Predicates to filter out only subsets of the data:

```
CREATE FUNCTION dbo.ops_fact_order_Predicate(@CustomerNo varchar(max))
RETURNS TABLE
WITH SCHEMABINDING
AS RETURN SELECT 1 AS fn_securitypredicate_result
WHERE @CustomerNo in (
		select trim(replace(replace(replace([value], '"', ''), ']', ''), 'Allow_Customer_', '')) 
		from STRING_SPLIT ( 
			CAST(SESSION_CONTEXT(N'original_roles') as varchar(max)) 
			, ',' 
			, 0) 
			where trim(replace(replace([value], '"', ''), ']', '')) like 'Allow_Customer%'
		)

CREATE SECURITY POLICY dbo.ops_fact_order_Policy
ADD FILTER PREDICATE dbo.ops_fact_order_Predicate(CustomerNo)
ON [gold_ops].[ops_fact_order];
```


src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -217,9 +217,9 @@ public override string GetSessionParamsQuery(HttpContext? httpContext, IDictiona
             foreach ((string claimType, string claimValue) in sessionParams)
             {
                 string paramName = $"{SESSION_PARAM_NAME}{counter.Next()}";
-                parameters.Add(paramName, new(claimValue));
+                parameters.Add(paramName, new(claimValue));
```
Aniruddh25, line 220:
```suggestion
                parameters.Add(paramName, new(claimValue));
```
nit: removing trailing spaces

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -217,9 +217,9 @@ public override string GetSessionParamsQuery(HttpContext? httpContext, IDictiona
             foreach ((string claimType, string claimValue) in sessionParams)
             {
                 string paramName = $"{SESSION_PARAM_NAME}{counter.Next()}";
-                parameters.Add(paramName, new(claimValue));
+                parameters.Add(paramName, new(claimValue));          
                 // Append statement to set read only param value - can be set only once for a connection.
-                string statementToSetReadOnlyParam = "EXEC sp_set_session_context " + $"'{claimType}', " + paramName + ", @read_only = 1;";
+                string statementToSetReadOnlyParam = "EXEC sp_set_session_context " + $"'{claimType}', " + paramName + ", @read_only = 0;";
```
Aniruddh25, line 222: curious, did you experience  - your mutation operations being blocked because of this setting?

Aniruddh25 (approved):
Thanks for this smart change to provide the original roles for additional filtering. Left a few nit comments and some questions, looks good to merge otherwise!

src/Core/Resolvers/MsSqlQueryExecutor.cs
M4Al: Yes, the SQL Endpoint did reject the second query as described in https://github.com/Azure/data-api-builder/issues/2341

src/Core/Parsers/RequestParser.cs
```diff
@@ -30,7 +30,7 @@ public class RequestParser
         /// <summary>
         /// Prefix used for specifying limit in the query string of the URL.
         /// </summary>
-        public const string FIRST_URL = "$first";
+        public const string FIRST_URL = "$top";
```
Aniruddh25, line 33: We will provide a synonym $top soon. Lets not make this change in this PR. Thank you for your patience.

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -51,6 +51,13 @@ public static OkObjectResult FormatFindResult(
                                                   ? DetermineExtraFieldsInResponse(findOperationResponse, context.FieldsToBeReturned)
                                                   : DetermineExtraFieldsInResponse(findOperationResponse.EnumerateArray().First(), context.FieldsToBeReturned);
 
+            //Remove RecordCOunt from extraFieldsInResponse if present
+            /*
```
Aniruddh25, line 55: If this code is not used, please remove it out completely.

src/Directory.Packages.props
```diff
@@ -20,7 +20,7 @@
     <PackageVersion Include="Microsoft.AspNetCore.Http" Version="2.2.2" />
     <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.38.1" />
     <!--When updating Microsoft.Data.SqlClient, update license URL in scripts/notice-generation.ps1-->
-    <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.2.0" />
+    <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.1.4" />
```
Aniruddh25, line 23: Please retain the updated 5.2.0 version

Aniruddh25 (changes requested):
The original PR contained fewer changes, please restrict the changes to the purpose of the PR and lets not add additional changes in the same PR. Waiting for removing the unnecessary changes

Dockerfile
```diff
@@ -1,13 +1,13 @@
 # Version values referenced from https://hub.docker.com/_/microsoft-dotnet-aspnet
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0-cbl-mariner2.0. AS build
+FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build
```
Aniruddh25, line 3: we will soon update this to 8.0. Please refrain from doing so right now.

abhishekkumams:
@M4Al , let us know if you would be able to address the comments...

JerryNixon:
@Aniruddh25 is this something we could wrap?

Aniruddh25:
> @Aniruddh25 is this something we could wrap?

The original PR was useful. It now contains additional changes which are not really necessary like $top, DwSQL builder changes. We need to clean those up before merging this in.
@RubenCerna2079, could you take this task up?

Aniruddh25:
@M4Al, we will take this PR forward and merge from here.

RubenCerna2079 (approved):
LGTM

souvikghosh04 (approved):
LGTM

PR #2356, opened 2024-08-30
Refactor Config Loader and Provider for Hot-Reload

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/1966

## What is this change?

Move the `RuntimeConfig` to the `RuntimeConfigLoader`, which is now responsible for maintaining that class in the proper state. We also move the Hot-Reload function to this class, since it now has the responsibility of maintaining a correct `RuntimeConfig`.

The provider therefore needs to have a `RuntimeConfigLoader` so that it can access the `RuntimeConfig` through that class. 

Initially, we were planning on splitting the `RuntimeConfigProvider` into 2 classes that share a common interface: one for hosted scenarios, and one for local. However, it became clear that this would create additional problems with identifying the correct version of the provider needed in the registered services. Please see the above related issue for more information.

The `RuntimeConfigProvider` will maintain private access to the `RuntimeConfigLoader`, however, the `RuntimeConfigLoader` will have public access to its `RuntimeConfig`. This allows the current functions to work with minimal changes, since we utilize out parameters in the parsing and loading of the config, which then leads to certain functions that modify the `RuntimeConfig` during initialization such as in the Hosted scenario when we call the Provider's initialize function.

This will eventually be refactored so that the only class responsible for modifying the `RuntimeConfig` is the loader, and we have an issue to track this work here: https://github.com/Azure/data-api-builder/issues/2362

## How was this tested?

Against current test suite.

## Sample Request(s)

Any request is valid.


src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,45 +90,76 @@ public string GetConfigFileName()
...
+        if (_configFileWatcher is not null)
+        {
+            return false;
+        }
+
+        if (RuntimeConfig is not null && RuntimeConfig.IsDevelopmentMode())
+        {
+            try
+            {
+                _configFileWatcher = new(this, GetConfigDirectoryName(), GetConfigFileName());
```
abhishekkumams, line 109: It would be good to have checkpoint logs indicating reload.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -182,19 +133,19 @@ public async Task<bool> Initialize(
...
-            _runtimeConfig = runtimeConfig;
+            ConfigLoader.RuntimeConfig = runtimeConfig;
 
             if (string.IsNullOrEmpty(runtimeConfig.DataSource.ConnectionString))
             {
                 throw new ArgumentException($"'{nameof(runtimeConfig.DataSource.ConnectionString)}' cannot be null or empty.", nameof(runtimeConfig.DataSource.ConnectionString));
             }
 
-            if (_runtimeConfig.DataSource.DatabaseType == DatabaseType.CosmosDB_NoSQL)
+            if (ConfigLoader.RuntimeConfig.DataSource.DatabaseType == DatabaseType.CosmosDB_NoSQL)
```
abhishekkumams, line 143: Based on line number 136, we could also have used `runtimeConfig` instead of `ConfigLoader.RuntimeConfig` in L:143 or other places, shorter the better.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -24,6 +24,8 @@ public abstract class RuntimeConfigLoader
 {
     protected readonly string? _connectionString;
 
+    public RuntimeConfig? RuntimeConfig;
```
seantleonard, line 27: can you add comment here and details in pr description describing the `RuntimeConfigLoader` API changes: 
1. Why is `RuntimeConfig` public, could it be protected?
2. Could the function `TryParseConfig` which has `out` paramter supplying runtime config also save state?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,45 +90,76 @@ public string GetConfigFileName()
...
-    /// <param name="dataSourceName">If provided and not empty, this is the data source name that will be used in the loaded config.</param>
+    /// <param name="logger">ILogger for logging errors.</param>
     /// <returns>True if the config was loaded, otherwise false.</returns>
     public bool TryLoadConfig(
         string path,
         [NotNullWhen(true)] out RuntimeConfig? config,
         bool replaceEnvVar = false,
-        string dataSourceName = "")
+        ILogger? logger = null,
+        string? defaultDataSourceName = null)
```
seantleonard, line 138: This seems like a breaking change to `TryLoadConfig(...)`
1. why null instead of empty string? I vaguely recall there being a reason we set empty string.
2. Callers' assumption that excluding the argument `defaultDataSourceName` means "use default datasource" is now broken.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -141,18 +102,8 @@ public bool TryGetConfig([NotNullWhen(true)] out RuntimeConfig? runtimeConfig)
...
-
-    /// <summary>
-    /// Hot Reloads the runtime config when the file watcher
-    /// is active and detects a change to the underlying config file.
-    /// </summary>
-    public void HotReloadConfig()
-    {
-        // _runtimeconfig can not be null in a hot reload scenario
-        ConfigLoader.TryLoadKnownConfig(out _runtimeConfig, replaceEnvVar: true, _runtimeConfig!.DefaultDataSourceName);
+        runtimeConfig = ConfigLoader.RuntimeConfig;
```
seantleonard, line 105: I don't follow why both `RuntimeConfigProvider` and `RuntimeConfigLoader`(and by definition, any derived classes) maintain the state of `RuntimeConfig`. The responsibilities of these two classes overlap such that it seems incorrect that the `ConfigLoader` is now acting as the 'provider.'

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -182,19 +133,19 @@ public async Task<bool> Initialize(
                 replaceEnvVar: false,
                 replacementFailureMode: EnvironmentVariableReplacementFailureMode.Ignore))
         {
-            _runtimeConfig = runtimeConfig;
+            ConfigLoader.RuntimeConfig = runtimeConfig;
```
seantleonard, line 136: This seems circular where 
1. The `RuntimeConfigProvider` is maintaining the `ConfigLoader`'s `RuntimeConfig` state (this line)
2. This `RuntimeConfigProvider`'s other function utilize `ConfigLoader.RuntimeConfig` state to update the state of this provider. (line 105)

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,45 +90,76 @@ public string GetConfigFileName()
...
+        if (logger is null)
+        {
+            Console.Error.WriteLine(errorMessage);
         }
         else
         {
-            // Unable to use ILogger because this code is invoked before LoggerFactory
-            // is instantiated.
-            Console.WriteLine($"Unable to find config file: {path} does not exist.");
+            logger.LogError(message: errorMessage);
```
seantleonard, line 162: need to use structured logging when the `ilogger` is present.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -138,11 +172,19 @@ public bool TryLoadConfig(
...
+    }
+
+    /// <summary>
+    /// Hot Reloads the runtime config when the file watcher
+    /// is active and detects a change to the underlying config file.
+    /// </summary>
+    public void HotReloadConfig(string defaultDataSourceName)
     {
-        return TryLoadConfig(ConfigFilePath, out config, replaceEnvVar, dataSourceName);
+        TryLoadKnownConfig(out _, replaceEnvVar: true, defaultDataSourceName);
```
seantleonard, line 187: It seems weird that the hot reload function calls `tryloadknownconfig`? shouldn't a new config be loaded? is this function needed? Wouldn't the detection of a filechanged event just need to call "TryLoadConfig"?

seantleonard (changes requested):
Need more clarifcation on design in the pr description. The ConfigLoader and ConfigProvider classes seem to overlap in responsibilities and now establish a circular dependency where these classes both maintain runtimeconfig state and update the other's runtimeconfig state.

src/Config/ConfigFileWatcher.cs
```diff
@@ -31,15 +36,19 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
     {
         try
         {
-            if (_configProvider is null)
+            if (_configLoader is null)
```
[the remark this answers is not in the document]
aaronburtle: with the current code it can't be will remove.

src/Config/ConfigFileWatcher.cs
```diff
@@ -31,15 +36,19 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
...
 
-            if (!_configProvider.IsLateConfigured && _configProvider.GetConfig().IsDevelopmentMode())
+            if (_configLoader.RuntimeConfig is null)
             {
-                // pass along the original default data source name for consistency within runtime config's data structures
-                _configProvider.HotReloadConfig();
+                throw new ArgumentNullException("RuntimeConfig can not be null.");
+            }
+
+            if (_configLoader.RuntimeConfig.IsDevelopmentMode())
```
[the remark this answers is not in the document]
aaronburtle: Yes in the case that DAB was started in development mode, in that case you can swap back and forth. If DAB is started in production mode then we will not allow hot-reload (with the exception of loglevel).

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: do you mea log that we are in a hot-reload scenario here because we are instantiating the filewatcher?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -132,7 +132,7 @@ Runtime.GraphQL is null ||
     }
 
     [JsonIgnore]
-    public string DefaultDataSourceName { get; private set; }
+    public string DefaultDataSourceName { get; set; }
```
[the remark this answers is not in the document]
aaronburtle: so that we can set it with the Loader, which we need to pass along the same name from one hot-reaload to another. This is because we have some validation that happens at request time and if this changes, that validation fails.

src/Core/Configurations/RuntimeConfigProvider.cs
aaronburtle: I don't follow, maybe we can sync offline and go over this one.

src/Config/ConfigFileWatcher.cs
```diff
@@ -36,17 +36,8 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
...
-                throw new ArgumentNullException("ConfigLoader can not be null.");
-            }
-
-            if (_configLoader.RuntimeConfig is null)
-            {
-                throw new ArgumentNullException("RuntimeConfig can not be null.");
-            }
-
-            if (_configLoader.RuntimeConfig.IsDevelopmentMode())
+            // if RuntimeConfig is null we hould abort any hot-reload.
```
seantleonard, line 39:
```suggestion
            // When RuntimeConfig is null, we abort hot-reload.
```

src/Core/Configurations/RuntimeConfigProvider.cs
seantleonard, line 1: Does RuntimeConfigLoader.ConfigLoader need to be public? Isn't RuntimeConfigProvider acting as the public api to get config state?

src/Config/RuntimeConfigLoader.cs
aaronburtle: added a comment. This is something we should consider for a future refactor. But for the moment would like to avoid this can of worms if possible.

src/Config/ConfigFileWatcher.cs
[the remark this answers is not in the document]
aaronburtle: I still need to dig into the multiple file changed notification issue.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -41,17 +41,33 @@ public class RuntimeConfigProvider
...
+        }
+    }
 
     public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
     {
         ConfigLoader = runtimeConfigLoader;
     }
 
+   
+
```
seantleonard, line 69:
```suggestion
```

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -61,56 +77,23 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
         // While loading the config file, replace all the environment variables with their values.
-        if (ConfigLoader.TryLoadKnownConfig(out RuntimeConfig? config, replaceEnvVar: true))
-        {
-            _runtimeConfig = config;
-            TrySetupConfigFileWatcher();
-        }
+        ConfigLoader.TryLoadKnownConfig(out _, replaceEnvVar: true);
 
-        if (_runtimeConfig is null)
+        if (ConfigLoader.RuntimeConfig is null)
```
seantleonard, line 88: Can these two lines be combined? why can't you use the result of `TryLoadKnownConfig` ?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -61,56 +77,23 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
-                // Need to remove the dependencies in startup on the RuntimeConfigProvider
-                // before we can have an ILogger here.
-                Console.WriteLine($"Attempt to configure config file watcher for hot reload failed due to: {ex.Message}.");
-            }
-
-            return _configFileWatcher is not null;
-        }
-
-        return false;
+        return ConfigLoader.RuntimeConfig;
```
seantleonard, line 96: Can't you return the result (out param) of `TryLoadKnownConfig`?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -120,17 +103,13 @@ private bool TrySetupConfigFileWatcher()
...
-                _runtimeConfig = config;
-                TrySetupConfigFileWatcher();
-            }
+            ConfigLoader.TryLoadKnownConfig(out _, replaceEnvVar: true);
         }
 
-        runtimeConfig = _runtimeConfig;
-        return _runtimeConfig is not null;
+        runtimeConfig = ConfigLoader.RuntimeConfig;
+        return ConfigLoader.RuntimeConfig is not null;
```
seantleonard, line 112: why not use the value of the `TryLoadKnownConfig` out param?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -143,12 +147,13 @@ public bool TryLoadConfig(
             string json = _fileSystem.File.ReadAllText(path);
             bool configParsed = TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
             TrySetupConfigFileWatcher();
-            config = RuntimeConfig;
-            if (RuntimeConfig is not null)
+
+            if (RuntimeConfig is not null && !string.IsNullOrEmpty(defaultDataSourceName))
```
seantleonard, line 151: why "!string.IsNullOrEmpty(defaultDataSourceName)" ? aren't we intentionally setting default datasource name to empty string ?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -83,17 +80,17 @@ public RuntimeConfig GetConfig()
         }
 
         // While loading the config file, replace all the environment variables with their values.
-        ConfigLoader.TryLoadKnownConfig(out _, replaceEnvVar: true);
+        ConfigLoader.TryLoadKnownConfig(out RuntimeConfig? runtimeConfig, replaceEnvVar: true);
 
-        if (ConfigLoader.RuntimeConfig is null)
+        if (runtimeConfig is null)
```
seantleonard, line 85: why not use the output of `TryLoadKnownConfig`? that result will implicitly tell you whether out param `runtimeConfig` is null or not.

seantleonard (approved):

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: This is for overwriting the default data source name, which we don't want to do in the case of an empty string, but only when a specific defaultDataSourceName is provided. So for example in a hot-reload scenario we can pass along the previous name and it will overwrite here, but in the case that we have a name already, we don't want to overwrite with the empty string.

src/Core/Configurations/RuntimeConfigProvider.cs
aaronburtle: saving 1 bit of memory on the stack of course ;)

sure Ill swap that around :D

aaronburtle: The actual reason was to get the compiler to be quiet about returning a possible null value, it doesnt recognize that if the return value is true here we have a non-null, but I can add the ! operator to quiet it up.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -61,56 +74,23 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
+            return ConfigLoader.RuntimeConfig;
         }
 
         // While loading the config file, replace all the environment variables with their values.
-        if (ConfigLoader.TryLoadKnownConfig(out RuntimeConfig? config, replaceEnvVar: true))
-        {
-            _runtimeConfig = config;
-            TrySetupConfigFileWatcher();
-        }
+        bool isLoaded = ConfigLoader.TryLoadKnownConfig(out RuntimeConfig? runtimeConfig, replaceEnvVar: true);
```
seantleonard, line 83: These last few lines can be condensed and can avoid the `!` null forgiving operator with the following:
```csharp
        if (!ConfigLoader.TryLoadKnownConfig(out RuntimeConfig? runtimeConfig, replaceEnvVar: true))
        {
                message: "Runtime config isn't setup.",
        }

        return runtimeConfig;
```

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -264,13 +229,13 @@ public async Task<bool> Initialize(
 
         if (RuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig? runtimeConfig, replaceEnvVar: replaceEnvVar, replacementFailureMode: replacementFailureMode))
         {
-            _runtimeConfig = runtimeConfig.DataSource.DatabaseType switch
+            ConfigLoader.RuntimeConfig = runtimeConfig.DataSource.DatabaseType switch
```
Aniruddh25, line 232: are we ok setting this directly in a different class? should we instead have a private setter?

aaronburtle: I created an issue for more refactor to eliminate this being public, for now we need to be able to change the RuntimeConfig from the provider since we have parse functions that are not able to save the state of the RuntimeConfig directly inside of the loader. This is mainly because the hosted scenario uses static functions and modifies the runtimeconfig outside the loader through the out param.

src/Config/FileSystemRuntimeConfigLoader.cs
seantleonard: this would be a good comment to include in the code.

abhishekkumams (approved):
Looks good to me, Would be good to add logs to indicate when the config is reloaded.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -41,11 +41,24 @@ public class RuntimeConfigProvider
     /// </summary>
     public Dictionary<string, string?> ManagedIdentityAccessToken { get; private set; } = new Dictionary<string, string?>();
 
-    public RuntimeConfigLoader ConfigLoader { get; private set; }
+    private RuntimeConfigLoader ConfigLoader { get; set; }
```
Aniruddh25, line 44: if this is private, should rename it to _configLoader as per the conventions.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,45 +94,77 @@ public string GetConfigFileName()
...
-        string dataSourceName = "")
+        ILogger? logger = null,
+        string defaultDataSourceName = "")
     {
         if (_fileSystem.File.Exists(path))
         {
             Console.WriteLine($"Loading config file from {path}.");
             string json = _fileSystem.File.ReadAllText(path);
+            bool configParsed = TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar);
+            TrySetupConfigFileWatcher();
```
Aniruddh25, line 149: if the Parsing failed, configParsed will be false. Why aren't we checking if the confiParsed successfully or not before setting up the File watching? Do we really want file watcher even if parsing fails?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,45 +94,77 @@ public string GetConfigFileName()
...
 
-            if (TryParseConfig(
-                json: json,
-                config: out config,
-                connectionString: _connectionString,
-                replaceEnvVar: replaceEnvVar))
+            if (RuntimeConfig is not null && !string.IsNullOrEmpty(defaultDataSourceName))
```
seantleonard, line 151: similar to Ani's comment above, you 'manually' check `RuntimeConfig` here, however you could make this more concise by wrapping `TryParseConfig` in the `if` statement. Anything that requires RuntimeConfig to be non-null can occur in the if block when parsing is successful.

Aniruddh25 (approved):
Looks good once all the remaining conversations are resolved.

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: will add to the hot-reload function

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -145,16 +145,24 @@ public bool TryLoadConfig(
...
+                    logger?.LogInformation("Monitoring config: {ConfigFilePath} for hot-reloading.", ConfigFilePath);
+                }
+
+                if (!string.IsNullOrEmpty(defaultDataSourceName))
+                {
+                    RuntimeConfig.DefaultDataSourceName = defaultDataSourceName;
+                }
+
+                config = RuntimeConfig;
+                return true;
```
seantleonard, line 161: It makes sense that `TrySetupConfigFileWatcher()` is best effort. Within that function, are there any logging mechanisms to emit information explaining why it may have failed in a failure case?

seantleonard (approved):

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: we pass along the exception message if the instantiation of the file watcher fails

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -87,45 +94,85 @@ public string GetConfigFileName()
...
-                if (!string.IsNullOrEmpty(dataSourceName))
+                if (TrySetupConfigFileWatcher())
+                {
+                    logger?.LogInformation("Monitoring config: {ConfigFilePath} for hot-reloading.", ConfigFilePath);
+                }
+
+                if (!string.IsNullOrEmpty(defaultDataSourceName))
                 {
-                    config.UpdateDefaultDataSourceName(dataSourceName);
+                    RuntimeConfig.DefaultDataSourceName = defaultDataSourceName;
```
abhishekkumams, line 157: Why this change? this is causing inconsistencies.

Issue #1645, opened 2023-08-23
Add `log-level` property to runtime config schema


abhishekkumams:
This Issue is to propose a new property called `log-level` under the runtime section of dab config file to specify the minimum logging level in our Application. for example if it is set to `Information`, trace and debug logs won't be displayed while warning and errors along with Information Logs will show up.
Having LogLevel specified in the config file enable users to specify the logLevel when CLI is not envolved(where we can use --LogLevel).

```
{
  "runtime": {
    ...
    "log-level": "Information",
    ...
}
```

This logLevel will be used for logging in the console as well as sending logs to Applicaiton Insights.

seantleonard:
#1829

seantleonard:
Moving this + hot reload related tasks to post GA -> tracking via 1.2rc milestone.

seantleonard:
Lower priority because you can set log level via argument to dab.exe process 
```sh
$ ./Microsoft.DataApiBuilder --LogLevel Information
```

PR #2359, opened 2024-09-04
Add log-level property to config schema

RubenCerna2079:
## Why make this change?

Adds log-level property to dab config file under the runtime section, closing issue #1645. The log-level is necessary to receive updates from the program, and with this property, it gives the users the ability to change what type of information they receive from the program. Lastly, it allows the users to have more options on how to change the log-level property, be it through the config file or the CLI. 

## What is this change?

First, the log-level property was added to the config file schema, in order for the user to have the ability to add the property into their config file. In order to save the information that is inside the property, a new object model named `LogLevelOptions` was created, inside of it is a new object model named `Level` with the purpose of showing all of the possible values in an enum type. A `LogLevelOptions` was then added to `RuntimeOptions` so the program knows in which section it is supposed to parse the information. A new converter `LogLevelOptionsConverterFactory` was created in order to allow the property in the config file to be deserialized. The logic in `Startup` file was modified to first tries to set the loggers based on the log-level property, in the case that log-level is null, it will fall back on the logic that was set up for `Host Mode`. New tests were created in `ConfigurationTests` to test the validity of the code.

## How was this tested?

- [x] Integration Tests
- [ ] Unit Tests


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3450,6 +3450,69 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+        {
+            RuntimeConfig configWithCustomLogLevel = InitRuntimeWithLogLevel(expectedLevel);
+
+            //Try should fail and go to catch exception
+            try
+            {
+                string configWithCustomLogLevelJson = configWithCustomLogLevel.ToJson();
+                Assert.Fail();
+            }
+            //Catch verifies that the exception is due to LogLevel having a value that does not exist
```
aaronburtle, line 3495:
```suggestion
            // Catch verifies that the exception is due to LogLevel having a value that does not exist
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3450,6 +3450,69 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+        private static RuntimeConfig InitRuntimeWithLogLevel(Level expectedLevel)
+        {
+            TestHelper.SetupDatabaseEnvironment(MSSQL_ENVIRONMENT);
+            RuntimeConfig config = SqlTestHelper.SetupRuntimeConfig();
+
+            return config with
+            {
+                Runtime = config.Runtime with
+                {
+                    LogLevel = config.Runtime.LogLevel with { Value = expectedLevel }
```
seantleonard, line 3511: TestConfig files probably don't know about and aren't populated with `config.Runtime.LogLevel`. You'd want to identify how to write this section of code such that:
- preserve pre-existing `config.Runtime` value
- append/add/insert `LogLevel` property and value. 

This is a good test of backwards compatibility: your LogLevelOptionsConverter should not fail if the `LogLevel` property is not present in the `dab-config.json` file.

src/Service/Startup.cs
```diff
@@ -429,6 +429,22 @@ public static LogLevel GetLogLevelBasedOnMode(RuntimeConfig runtimeConfig)
             return LogLevel.Error;
         }
 
+        /// <summary>
+        /// Takes in the RuntimeConfig object and checks the LogLevel level.
+        /// If LogLevel is not null, it will return the current value as a LogLevel,
+        /// else it will take the default option by going to GetLogLevelBasedOnMode
+        /// </summary>
+        public static LogLevel GetLogLevelBasedOnLevel(RuntimeConfig runtimeConfig)
```
seantleonard, line 437:
```suggestion
        public static LogLevel GetLogLevel(RuntimeConfig runtimeConfig)
```
nit: redundant name.

I added my suggestion. Another option, if you like better, is `GetConfiguredLogLevel()`

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -536,4 +536,12 @@ public uint GetPaginationLimit(int? first)
             return defaultPageSize;
         }
     }
+
+    /// <summary>
+    /// Retrieve the value of LogLevel from runtime
+    /// </summary>
+    public Level? LogLevelValue()
+    {
+        return Runtime?.LogLevel?.Value;
```
seantleonard, line 545: In the startup class I see 2 the following two helpers which both return the type `LogLevel`:
- `public static LogLevel GetLogLevelBasedOnMode(...)`
- `public static LogLevel GetLogLevelBasedOnLevel(...)`

What do you think about moving log level determination logic to `RuntimeConfig`?
e.g. `LogLevel GetLogLevel()` which, internally, will check whether dab is in development mode -> returns correct log level and handles casting of `Level` to `LogLevel`.

On that same note, this function returns `Level`. Are there components of DAB which use `Level` directly or should the `RuntimeConfig` object's public API only return `LogLevel` types?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -536,4 +536,12 @@ public uint GetPaginationLimit(int? first)
             return defaultPageSize;
         }
     }
+
+    /// <summary>
+    /// Retrieve the value of LogLevel from runtime
+    /// </summary>
+    public Level? LogLevelValue()
```
seantleonard, line 543:
```suggestion
    public Level? LogLevel()
```
nit: "value" is implied.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3450,6 +3450,81 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+        [DataRow(12, DisplayName = "Validates that a bigger positive log level value that does not exist, fails to build")]
+        public void TestNonExistingLogLevels(Level expectedLevel)
+        {
+            RuntimeConfig configWithCustomLogLevel = InitRuntimeWithLogLevel(expectedLevel);
+
+            // Try should fail and go to catch exception
+            try
+            {
+                string configWithCustomLogLevelJson = configWithCustomLogLevel.ToJson();
+                Assert.Fail();
```
seantleonard, line 3493: How does your `LogLevelOptionsConverterFactory.Write(...)` function behave? Another good test would be similar to this one here, but testing valid values for log level, then seeing how they are written to a json file. 
Additonally, you'd want to test this scenario:
- user creates dab-config.json, but doesn't define a value for, nor include the `log-level` property. When the converter.Write() function is called, does it always write out the `log-level` property/default value to the json file? We don't necessarily want to write properties to files when users didn't explicitly include them.

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: The only components that uses `Level` is `LogLevelOptions` so `RuntimeConfig` objects should only return `LogLevel` types.

src/Service/Startup.cs
```diff
@@ -445,7 +461,7 @@ public static ILoggerFactory CreateLoggerFactoryForHostedAndNonHostedScenario(IS
                 RuntimeConfigProvider configProvider = serviceProvider.GetRequiredService<RuntimeConfigProvider>();
                 if (configProvider.TryGetConfig(out RuntimeConfig? runtimeConfig))
                 {
-                    MinimumLogLevel = GetLogLevelBasedOnMode(runtimeConfig);
+                    MinimumLogLevel = GetLogLevelBasedOnLevel(runtimeConfig);
```
aaronburtle, line 464: We have a couple helper functions in startup now who are trying to get the log level based on values in the config, and the helper function here `GetLogLevelBasedOnLevel` depends on the other helper function in the case where the log level is null.

I think as Sean was pointing out in another comment that we should combine these into a GetLogLevel() function that sorts out all of the values and based on the precedence will return the correct log level. This function should probably live in the RuntimeConfig as Sean suggested, and in fact that previous helper function that we already had in startup probably should have been living in the RuntimeConfig to begin with.

aaronburtle (approved):
Looks good, just need to cleanup the GetLogLevel() function and find a better place for it to live.

src/Config/ObjectModel/RuntimeConfig.cs
seantleonard: There was more to this comment than identifying which components use `Level` directly. 

Any thoughts on this:
In the startup class I see 2 the following two helpers which both return the type LogLevel:

> - `public static LogLevel GetLogLevelBasedOnMode(...)`
> - `public static LogLevel GetLogLevelBasedOnLevel(...)`
> What do you think about moving log level determination logic to `RuntimeConfig`?
> e.g. `LogLevel GetLogLevel()` which, internally, will check whether dab is in development mode -> returns correct log level and handles casting of `Level` to `LogLevel`.

src/Service/Startup.cs
seantleonard: How was this resolved? if you decide not to make any changes, please add follow up comment indicating which direction you chose to follow and your deciding factors.

src/Config/ObjectModel/RuntimeConfig.cs
seantleonard: How was this resolved? if you decide not to make any changes, please add follow up comment indicating which direction you chose to follow and your deciding factors.

src/Config/ObjectModel/Level.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
+
+public enum Level
```
abhishekkumams, line 6: Better to use `LogLevel` than just `Level`

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3450,6 +3450,81 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+            catch (Exception ex)
+            {
+                Assert.AreEqual(typeof(KeyNotFoundException), ex.GetType());
+            }
+        }
+
+        /// <summary>
+        /// Helper method to create RuntimeConfig with specificed LogLevel value
+        /// </summary>
+        private static RuntimeConfig InitRuntimeWithLogLevel(Level expectedLevel)
```
abhishekkumams, line 3505: `InitializeRuntimeConfigWithLogLevel` this properly reflects its functionality of initializing a RuntimeConfig .

src/Service/Startup.cs
```diff
@@ -429,6 +429,22 @@ public static LogLevel GetLogLevelBasedOnMode(RuntimeConfig runtimeConfig)
             return LogLevel.Error;
         }
 
+        /// <summary>
+        /// Takes in the RuntimeConfig object and checks the LogLevel level.
```
abhishekkumams, line 433:
```suggestion
        /// Takes in the RuntimeConfig object and checks the LogLevel.
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3450,6 +3450,81 @@ public async Task OpenApi_InteractiveSwaggerUI(
             }
         }
 
+        /// <summary>
+        /// Test different loglevel values that are avaliable by deserializing RuntimeConfig with LogLevel value
```
abhishekkumams, line 3454:
```suggestion
        /// Test different loglevel values that are avaliable by deserializing RuntimeConfig with specified LogLevel 
```

src/Config/ObjectModel/Level.cs
abhishekkumams: Also i don't see any extra options that are already there in LogLevel

schemas/dab.draft.schema.json
```diff
@@ -261,6 +261,28 @@
             }
           }
         },
+        "log-level": {
+          "type": "object",
+          "description": "Global configuration of log level",
+          "additionalProperties": false,
+          "properties": {
+            "level": {
+              "description": "Set between the different level options for log level",
```
abhishekkumams, line 270:
```suggestion
              "description": "Defines logging severity levels.",
```
also specify, what happens in default value.

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: Changed to IsLogLevel

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Created new tests called `LogLevelSerialization` and `LogLevelSerializationForNonExistentProperty`

src/Config/ObjectModel/ExtendedLogLevel.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
-public enum Level
+public enum ExtendedLogLevel
```
seantleonard, line 6: how is this different from: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=net-8.0

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -540,8 +541,31 @@ public uint GetPaginationLimit(int? first)
     /// <summary>
     /// Retrieve the value of LogLevel from runtime
     /// </summary>
-    public Level? LogLevelValue()
+    public ExtendedLogLevel? IsLogLevel()
```
seantleonard, line 544: Is log level what?
we typically use "Is" "Has" etc for booleans. e.g. "IsDeveloperMode() -> true/false". You're just returning the value of `Runtime?.LogLevel?.Value;` here.

Is this function really necessary? could this be baked into your function `GetConfiguredLogLevel()`? This was brought up in this [discussion ](https://github.com/Azure/data-api-builder/pull/2359/files/c15667ba34010df8932732584512de4d8ffead13..a8212a6bccd278d07985f0ddf70b944369763209#r1747557812) but I didn't see any further responses there.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -645,6 +652,7 @@ private static bool TryParseAndAssertOnDefaults(string json, out RuntimeConfig p
             Assert.IsFalse(parsedConfig.IsDevelopmentMode());
             Assert.IsTrue(parsedConfig.IsStaticWebAppsIdentityProvider);
             Assert.IsTrue(parsedConfig.IsRequestBodyStrict);
+            Assert.IsTrue(parsedConfig.IsLogLevel() is null);
```
seantleonard, line 655: here you can see other examples of naming using prefix "is". They are all boolean values.

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -261,8 +261,9 @@ public void TestNullableOptionalProps()
             // Test with empty sub properties of runtime
             minJson.Append(@"{ ""rest"": { }, ""graphql"": { },
                             ""base-route"" : """",");
-            StringBuilder minJsonWithHostSubProps = new(minJson + @"""telemetry"" : { }, ""host"" : ");
-            StringBuilder minJsonWithTelemetrySubProps = new(minJson + @"""host"" : { }, ""telemetry"" : ");
+            StringBuilder minJsonWithHostSubProps = new(minJson + @"""telemetry"" : { }, ""log-level"" : { }, ""host"" : ");
+            StringBuilder minJsonWithTelemetrySubProps = new(minJson + @"""host"" : { }, ""log-level"" : { }, ""telemetry"" : ");
+            StringBuilder minJsonWithLogLevelSubProps = new(minJson + @"""host"" : { }, ""telemetry"" : { }, ""log - level"" : ");
```
seantleonard, line 266: did you mean for this to be "log - level" instead of "log-level"?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3499,10 +3499,68 @@ public void TestNonExistingLogLevels(Level expectedLevel)
...
+        /// Tests serialization for the case that the loglevel property is not included,
+        /// to ensure that it is not written to the Json config.
+        /// </summary>
+        [TestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        public void LogLevelSerializationForNonExistentProperty()
+        {
+            RuntimeConfig runtimeConfig = CreateBasicRuntimeConfigWithNoEntity();
+            string configJson = runtimeConfig.ToJson();
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(configJson, out RuntimeConfig deserializedRuntimeConfig));
```
seantleonard, line 3545: what are you trying to confirm here? that `CreateBasicRuntimeConfigWithNoEntity();` returned a valid runtimeconfig?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -540,8 +541,31 @@ public uint GetPaginationLimit(int? first)
...
+    /// else it will take the default option by checking host mode.
+    /// If host mode is Development, return `LogLevel.Debug`, else
+    /// for production returns `LogLevel.Error`.
+    /// </summary>
+    public static LogLevel GetConfiguredLogLevel(RuntimeConfig runtimeConfig)
+    {
+        ExtendedLogLevel? value = runtimeConfig.IsLogLevel();
+        if (value is not null)
+        {
+            return (LogLevel)value;
```
seantleonard, line 561: pending answers to previous feedback threads: if there is no difference between Microsoft.Extensions.Logging.LogLevel enum and ExtendedLog level... we should just be saving Microsoft.Extensions.Logging.LogLevel  to the runtimeconfig and that would negate the need for the casting here. 

If the need for the new enum is just to support "null" , see the example for `EntityCacheOptions` where properties are nullable and there is also a property which tracks whether value is user-defined: https://github.com/Azure/data-api-builder/blob/main/src/Config/ObjectModel/EntityCacheOptions.cs

Aniruddh25 (commented):
Posting comments so far

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: Changing IsLogLevel to IsLogLevelNull, so it will only return true/false depending on whether `Runtime?.LogLevel?.Value` is null. So that it can be used in the `RuntimeConfigLoaderJsonDeserializerTests`

src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
RubenCerna2079: Changing IsLogLevel to IsLogLevelNull

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Deleted test as it was manually tested that there is no accidental writing made to the config file

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: Deleted `ExtendedLogLevel`, although the casting is still needed as it is going from `LogLevel?` to `LogLevel`

src/Service/Startup.cs
```diff
@@ -429,6 +429,22 @@ public static LogLevel GetLogLevelBasedOnMode(RuntimeConfig runtimeConfig)
...
+        /// If LogLevel is not null, it will return the current value as a LogLevel,
+        /// else it will take the default option by going to GetLogLevelBasedOnMode
+        /// </summary>
+        public static LogLevel GetLogLevelBasedOnLevel(RuntimeConfig runtimeConfig)
+        {
+            Level? value = runtimeConfig.LogLevelValue();
+            if (value is not null)
+            {
+                return (LogLevel)value;
```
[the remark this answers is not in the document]
RubenCerna2079: LogLevel is an enum that already exists, it will be used instead of Level

src/Config/ObjectModel/RuntimeOptions.cs
```diff
@@ -15,6 +15,7 @@ public record RuntimeOptions
     public TelemetryOptions? Telemetry { get; init; }
     public EntityCacheOptions? Cache { get; init; }
     public PaginationOptions? Pagination { get; init; }
+    public LogLevelOptions? LogLevel { get; init; }
```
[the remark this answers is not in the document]
RubenCerna2079: Changed to `LoggerLevel`

schemas/dab.draft.schema.json
```diff
@@ -261,6 +261,28 @@
             }
           }
         },
+        "logger-level": {
```
seantleonard, line 264: We should stick with the originally proposed name `log-level`. 

To address Ani's comment about naming not colliding with built in types: you can have the publicly exposed JSON Property name be `log-level` and the actual record member name be `LoggerLevel`

src/Config/ObjectModel/RuntimeOptions.cs
```diff
@@ -15,6 +15,7 @@ public record RuntimeOptions
     public TelemetryOptions? Telemetry { get; init; }
     public EntityCacheOptions? Cache { get; init; }
     public PaginationOptions? Pagination { get; init; }
+    public LogLevelOptions? LoggerLevel { get; init; }
```
seantleonard, line 18: I think it would be appropriate so set prefix this line with:
```
[JsonPropertyName("log-level")]
```

src/Config/Converters/LogLevelOptionsConverterFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.ObjectModel;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+
+/// <summary>
+/// Defines how DAB reads and writes log level options
+/// </summary>
+internal class LogLevelOptionsConverterFactory : JsonConverterFactory
```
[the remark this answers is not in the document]
RubenCerna2079: We use a converter factory because it was made with the idea in mind that in the future, we can allow the user to change each of the different providers individually, and they would reside inside the `LogLevelOptions`. It also helps keep the pattern that was set by the other properties inside of runtime.

seantleonard (approved):

Issue #2351, opened 2024-08-26
Update `export` command to be flexible with HTTP/HTTPS ports


abhishekkumams:
Finalize how the export command will be more secure and flexible with HTTP/HTTPS ports

PR #2375, opened 2024-09-10
Fix export command

### LINKED ISSUE #1542: [Bug]: Export command fails to export the graphql schema
### What happened?

Even when the engine is started, `dab export` fails to export the graphql schema file.

# How to reproduce?
1. Use the `dab-config.Mssql.json` file update the connection string accordingly.
2. run the command 
```
dab export --graphql -o ./schema --config dab-config.Mssql.json
```

3. Error message is not helpful, it just mentions it failed to export schema file.

### Version

running from main

Azure SQL

Local (including CLI)

_No response_

### Relevant log output

_No response_

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #1542 
  - With .Net deciding to remove HTTPS endpoint support by default,  `export` command started to fail as it uses HTTPS endpoint.

## What is this change?

- Added a fallback HTTP url, which will be used when HTTPS endpoint fails to fetch the graphql schema.

## How was this tested?

- [X] Unit Tests
- [X] Manual Tests

## Sample Request(s)
command:
`dab export --graphql -o ./testgqlschemaexport`

image

### DIFF
--- a/src/Cli/Exporter.cs
+++ b/src/Cli/Exporter.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.IO.Abstractions;
+using System.Runtime.CompilerServices;
 using Azure.DataApiBuilder.Config;
 using Azure.DataApiBuilder.Config.ObjectModel;
 using Azure.DataApiBuilder.Core.Generator;
@@ -10,6 +11,7 @@ using HotChocolate.Utilities.Introspection;
 using Microsoft.Extensions.Logging;
 using static Cli.Utils;
 
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
 namespace Cli
 {
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/Cli/Exporter.cs
```diff
@@ -10,12 +11,13 @@
 using Microsoft.Extensions.Logging;
 using static Cli.Utils;
 
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
```
seantleonard, line 14: add comment that indicates this is needed so that Mock can utilize your internal functions. Was it not enough to configure internals visible to just the test project?

src/Cli/Exporter.cs
```diff
@@ -115,27 +118,66 @@ private static async Task ExportGraphQL(ExportOptions options, RuntimeConfig run
...
-            HttpClient client = new( // CodeQL[SM02185] Loading internal server connection
+            try
+            {
+                logger.LogInformation("Trying to fetch schema from DAB Service using HTTPS endpoint.");
+                schemaText = GetGraphQLSchema(runtimeConfig, useFallbackURL: false);
+            }
+            catch
+            {
+                logger.LogInformation("Failed to fetch schema from DAB Service using HTTPS endpoint. Trying with HTTP endpoint.");
+                schemaText = GetGraphQLSchema(runtimeConfig, useFallbackURL: true);
```
seantleonard, line 143: Is there any way we could make `useFallbackURL` configurable via CLI option? HTTP is insecure so while in development, this may be okay, a production scenario should require that http downgrade be explicit:
`dab export graphqlschema --allow-http-fallback` something like that. 
So if user has dab behind reverse proxy in production (dab is only http), then they will need to explicitly set that fallback option.

if user doesn't want insecure http fallback option, they'd need to setup PKI (certs)

src/Cli/Exporter.cs
```diff
@@ -115,27 +118,66 @@ private static async Task ExportGraphQL(ExportOptions options, RuntimeConfig run
...
+            }
+            else
             {
-                BaseAddress = new Uri($"https://localhost:5001{runtimeConfig.GraphQLPath}")
-            };
+                client = new()
+                {
+                    BaseAddress = new Uri($"http://localhost:5000{runtimeConfig.GraphQLPath}")
+                };
+            }
```
seantleonard, line 172: We need to rethink this function now that we're visiting it in your PR. Few thoughts:
1. we hardcode localhost:5001/5000 -> real use cases may override these ports and endpoints.  I don't recall whether we restrict GraphQL schema export for dev mode only. (i don't think we have restrictions here).
2. we use `ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` which is insecure for production. I think this delegate should only be set when developer does not use my other discussion's proposed `--http-fallback` or `--allow-insecure-fallback`.
3. order of operations:( let's discuss this)
    1. HTTPS (valid certificate, not self-signed). -> `dab export graphql`
    2. HTTPS (self-signed dev scenario, only works with fallback flag, set the `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` delegate
    3. If step 2 fails, then fall back again to http

src/Cli/Exporter.cs
```diff
@@ -115,27 +118,66 @@ private static async Task ExportGraphQL(ExportOptions options, RuntimeConfig run
...
 
-            HttpClient client = new( // CodeQL[SM02185] Loading internal server connection
+            try
+            {
+                logger.LogInformation("Trying to fetch schema from DAB Service using HTTPS endpoint.");
+                schemaText = GetGraphQLSchema(runtimeConfig, useFallbackURL: false);
+            }
+            catch
+            {
+                logger.LogInformation("Failed to fetch schema from DAB Service using HTTPS endpoint. Trying with HTTP endpoint.");
```
Aniruddh25, line 142: What happens if the schema fetch fails even from the fallback option i.e. http

Aniruddh25 (commented):
Waiting to see what we decide on whether to ask user to allow fallback or not

src/Cli/Exporter.cs
Aniruddh25: the previous code says usage of `DangerousAcceptAnyServerCertificateValidator` was exempt here because of internal server usage.

I dont think we have an API that exposes the export functionality. The only usage seems to be via the CLI from the export command for a short duration so unlikely to have production mode usage. Hence, wonder if we need to add additional flags for such a simple functionality?

src/Cli/Exporter.cs
abhishekkumams: it will give failure with log indicating we tried HTTPS and then HTTP and then failed to export grpahQL

src/Cli/Exporter.cs
abhishekkumams: yep, I have this issue created for the same: https://github.com/Azure/data-api-builder/issues/2351

src/Cli/Exporter.cs
abhishekkumams: yes, export functionality is only for CLI.

src/Cli/Exporter.cs
abhishekkumams: No, It was giving access error.

src/Cli/Exporter.cs
seantleonard: I think 1 is still important here even if you decide to continue usage of `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator`. We hard code the port used to host dab and don't consider a developer overriding those ports which could be possible for development mode.

seantleonard: please open an issue to track customizing the port used for export. port -> ports 5000/5001 are hardcoded and should allow customization through option to CLI. This thread can be closed using the issue number created.

seantleonard (approved):

aaronburtle (approved):
LGTM!

PR #2384, opened 2024-09-14
Create new OpenApi Document in the case of a Hot Reload

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/1869

## What is this change?

Create a function in `OpenApiDocumentor` that invokes a new call to the `CreateDocument()` function to refresh the `OpenApiDocument`. Add an event handler, and subscribe the function for calling `CreateDocument()` to that handler when constructing the `OpenApiDocumentor`

## How was this tested?

Current test suite, manually ran.

## Sample Request(s)

N/A


src/Config/HotReloadEventHandler.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config
+{
+    /// <summary>
+    /// HotReloadEventHandler manages the events that are needed to signal refreshing
+    /// classes that must be updated during a hot reload.
```
seantleonard, line 8: Need a comment here describing why events are defined in this class instead of the RuntimeConfigLoader class. 

From .NET Event Programming Guide: https://learn.microsoft.com/en-us/dotnet/standard/events/#events
> The event is typically a member of the event sender; 

In this case, we can't do that because of this other behavioral issue that would arise:https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-raise-base-class-events-in-derived-classes

> When you create a class that can be used as a base class for other classes, you should consider the fact that events are a special type of delegate that can only be invoked from within the class that declared them.

-> Adding this event handler class as the class to be injected via DI to all dependent services instead of each of the dependent services loaded "RuntimeConfigLoader" class.

seantleonard (changes requested):
Some questions and feedback for touching this up

src/Config/HotReloadEventHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    /// <typeparam name="TEventArgs">Args used for hot reload events.</typeparam>
+    public class HotReloadEventHandler<TEventArgs> where TEventArgs : CustomEventArgs
+    {
+        public event EventHandler<TEventArgs>? MetadataProvider_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? Documentor_ConfigChangeEventOccurred;
```
abhishekkumams, line 14: we can use `OnConfigChanged` as it's clear and concise.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -22,15 +22,30 @@ namespace Azure.DataApiBuilder.Config;
...
+    protected virtual void Documentor_ConfigChangeEventOccurred(CustomEventArgs args)
     {
+        Handler?.Documentor_OnConfigChangeEventOccurred(this, args);
+    }
+
+    // Sends all of the notifications when a hot reload occurs.
+    public void SendEventNotification(string message = "")
+    {
+        CustomEventArgs args = new(message);
+        Documentor_ConfigChangeEventOccurred(args);
```
abhishekkumams, line 43: we can use `Documentor_OnConfigChanged(..)`

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -56,13 +57,22 @@ public class OpenApiDocumentor : IOpenApiDocumentor
         public const string DOCUMENT_CREATION_UNSUPPORTED_ERROR = "OpenAPI description document can't be created when the REST endpoint is disabled globally.";
         public const string DOCUMENT_CREATION_FAILED_ERROR = "OpenAPI description document creation failed";
 
+        public void Documentor_ConfigChangeEventReceived(object? sender, CustomEventArgs args)
```
Aniruddh25, line 60: nit: why the `_` in the name? As per our naming convention, wouldnt this be `DocumentorConfigChangeEventReceived`? Same for DocumentorSubscribe

src/Config/HotReloadEventHandler.cs
Aniruddh25: Good observation! @aaronburtle it might be helpful to add this in the code comments/PR description..

Aniruddh25 (approved):

Aniruddh25 (approved):
LGTM, as long as all the open comments are addressed.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -22,15 +22,30 @@ namespace Azure.DataApiBuilder.Config;
...
     // Public to allow the RuntimeProvider and other users of class to set via out param.
     // May be candidate to refactor by changing all of the Parse/Load functions to save
     // state in place of using out params.
     public RuntimeConfig? RuntimeConfig;
 
-    public RuntimeConfigLoader(string? connectionString = null)
+    // Signals a hot reload event for OpenApiDocumentor due to config change.
+    protected virtual void Documentor_ConfigChangeEventOccurred(CustomEventArgs args)
     {
+        Handler?.Documentor_OnConfigChangeEventOccurred(this, args);
```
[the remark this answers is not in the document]
aaronburtle: it needs to be nullable actually I have updated.

src/Config/RuntimeConfigLoader.cs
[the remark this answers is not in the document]
aaronburtle: The args are needed to match the signature of the method we are sing, it dervies from the base event handler class.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -22,15 +22,30 @@ namespace Azure.DataApiBuilder.Config;
...
     protected readonly string? _connectionString;
 
     // Public to allow the RuntimeProvider and other users of class to set via out param.
     // May be candidate to refactor by changing all of the Parse/Load functions to save
     // state in place of using out params.
     public RuntimeConfig? RuntimeConfig;
 
-    public RuntimeConfigLoader(string? connectionString = null)
+    // Signals a hot reload event for OpenApiDocumentor due to config change.
+    protected virtual void Documentor_ConfigChangeEventOccurred(CustomEventArgs args)
```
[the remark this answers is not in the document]
aaronburtle: Tried to make the naming more consistent.

src/Core/Services/OpenAPI/IOpenApiDocumentor.cs
```diff
@@ -26,6 +26,6 @@ public interface IOpenApiDocumentor
         /// <exception cref="DataApiBuilderException">Raised when document is already generated
         /// or a failure occurs during generation.</exception>
         /// <seealso cref="https://github.com/microsoft/OpenAPI.NET/blob/1.6.3/src/Microsoft.OpenApi/OpenApiSpecVersion.cs"/>
-        public void CreateDocument();
+        public void CreateDocument(bool isHotReloadScenario);
```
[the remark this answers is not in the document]
aaronburtle: some local changes were missing

src/Config/HotReloadEventHandler.cs
aaronburtle: added comment to try and clarify this.

src/Config/HotReloadEventHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config;
+
+/// <summary>
+/// HotReloadEventHandler manages the events that are needed to signal refreshing
+/// classes that must be updated during a hot reload. Events are defied in this class
+/// rather than in the RuntimeConfigLoader where they are raised because of the interaction
+/// between events and base and derived classes. For more information please see:
+/// https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-raise-base-class-events-in-derived-classes
```
seantleonard, line 11:
```suggestion
/// HotReloadEventHandler defines event invocation and subscription functions that are
/// used to facilitate updating DAB components' state due to a hot reload.
/// The events defined in this class are invoked in this class (versus being invoked in RuntimeConfigLoader)
/// because events are a special type of delegate that can only be invoked from within the class that declared them.
/// For more information about where events should be invoked, please see:
/// https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-raise-base-class-events-in-derived-classes
```

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -100,10 +110,10 @@ public bool TryGetDocument([NotNullWhen(true)] out string? document)
         /// <exception cref="DataApiBuilderException">Raised when document is already generated
         /// or a failure occurs during generation.</exception>
         /// <seealso cref="https://github.com/microsoft/OpenAPI.NET/blob/1.6.3/src/Microsoft.OpenApi/OpenApiSpecVersion.cs"/>
-        public void CreateDocument()
+        public void CreateDocument(bool isHotReloadScenario = false)
```
seantleonard, line 113: nit: after looking at this with fresh eyes, I think a param name like `doOverrideExistingDocument` may be more appropriate because it self-describes what your intent is. HotReload doesn't technically mean anything inside this `CreateDocument` function. Your added `Documentor_ConfigChangeEventReceived` function is the "hot-reload aware" function and that function just needs to have openApiDocumentor create a new doc even if one may already exist.

seantleonard (approved):

Issue #2323, opened 2024-08-09
Hot-reload and refresh dependencies on the DataSource section


aaronburtle:
We can currently deserialize and store the new values for the DataSource during a hot-reload scenario, however the dependencies that rely on the DataSource section during startup are not being refreshed. This task is to add that refreshing logic for the many components that are configured on startup and depend on the DataSource

![image](https://github.com/user-attachments/assets/4fafe421-9ca2-48ab-baa5-fa2970f3bfb7)

PR #2391, opened 2024-09-21
Hot-Reload dependencies with an enforced order of operations

aaronburtle:
## Why make this change?

closes https://github.com/Azure/data-api-builder/issues/2415
Relates to https://github.com/Azure/data-api-builder/issues/67
Relates to https://github.com/Azure/data-api-builder/issues/2323
Relates to https://github.com/Azure/data-api-builder/issues/1869

This change adds event handling for the dependencies which are set during startup, and refreshing their internal state to match the states that are updated during a hot-reload. In order to help facilitate the final portions of the configuration file which need to be handled in hot-reload scenarios, we need these dependencies to be updated without any stale values from before the hot-reload. And we enforce an order of this update to make sure that if B depends on A, that the refresh happens first with A, then with B.

## What is this change?

There are a number of classes which are instantiated and registered with the Service Provider on startup using values that depend on the 'RuntimeConfig'. These are registered as singletons. Therefore, during a hot-reload scenario, we need to update those states because the 'RuntimeConfig' has changed, and the values that those objects were instantiated with are now potentially stale. To handle this updating, when one of these singleton services is constructed, we subscribe an update function with an Event Handler that is provided to the constructor. We also save internally any arguments provided to the constructor that were needed during construction so that they will be available and usable during the updating.

The RuntimeConfig Loader then sequentially sends notifications for all of these subscribed events to inform them of the hot-reload process. This happens in a single function with a single thread so that the order of operation is enforced, with the thread executing each update one at a time.

Note that we save the handler internally in some of the classes that we will be calling back to do configurations during hot-reload. Generally speaking, this happens in the factory classes when their re-configuration during hot-reload requires them to create other classes that will need to be re-configured during a hot-reload. Those classes then need to have a handler provided to their constructor so that they can properly subscribe to hot-reload events.

## How was this tested?

Manually verified the updating process by modifying the config while DAB is running and then verifying that the dependent objects successfully refresh to match the new values in the config, and do so in the desired order. The following issue tracks specifically testing this behavior: https://github.com/Azure/data-api-builder/issues/2416

## Sample Request(s)

Any change to the config used on startup will demonstrate the hot-reload functionality added in this PR.


src/Core/Resolvers/Factories/QueryManagerFactory.cs~RF32b1084c.TMP
seantleonard, line 1: why this file

src/Core/Resolvers/Factories/MutationEngineFactory.cs
```diff
@@ -18,7 +19,15 @@ namespace Azure.DataApiBuilder.Core.Resolvers.Factories
...
-        private readonly Dictionary<DatabaseType, IMutationEngine> _mutationEngines;
+        private Dictionary<DatabaseType, IMutationEngine> _mutationEngines;
+        private readonly RuntimeConfigProvider _runtimeConfigProvider;
+        private readonly IAbstractQueryManagerFactory _queryManagerFactory;
+        private readonly IMetadataProviderFactory _metadataProviderFactory;
+        private readonly CosmosClientProvider _cosmosClientProvider;
+        private readonly IQueryEngineFactory _queryEngineFactory;
+        private readonly IHttpContextAccessor _httpContextAccessor;
+        private readonly IAuthorizationResolver _authorizationResolver;
+        private readonly GQLFilterParser _gQLFilterParser;
```
seantleonard, line 30: we have to add all these dependencies to mutationenginefactory? and all for just subscribing to the event? is there another way to accomplish what we need to do without adding a large number of dependencies in our factory classes?

seantleonard (changes requested):

src/Core/Resolvers/Factories/MutationEngineFactory.cs
aaronburtle: Another option that I considered was to have all of these dependencies on the Loader and then send them as arguments to the event handler.

src/Config/HotReloadEventHandler.cs
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config
+{
+    public class HotReloadEventHandler<TEventArgs> where TEventArgs : CustomEventArgs
+    {
+        public event EventHandler<TEventArgs>? QueryManagerFactory_ConfigChangeEventOccurred;
```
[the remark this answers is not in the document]
Aniruddh25: Good suggestion, need to ensure the members of this dictionary are populated accurately in respective constructors

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -60,8 +60,8 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// </summary>
     public string ConfigFilePath { get; internal set; }
 
-    public FileSystemRuntimeConfigLoader(IFileSystem fileSystem, string baseConfigFilePath = DEFAULT_CONFIG_FILE_NAME, string? connectionString = null)
-        : base(connectionString)
+    public FileSystemRuntimeConfigLoader(IFileSystem fileSystem, HotReloadEventHandler<CustomEventArgs>? handler, string baseConfigFilePath = DEFAULT_CONFIG_FILE_NAME, string? connectionString = null)
```
Aniruddh25, line 63: nit: move each parameter on a single line for easier readability

src/Config/RuntimeConfigLoader.cs
```diff
@@ -22,15 +22,66 @@ namespace Azure.DataApiBuilder.Config;
...
+    public void SendEventNotification(string message = "")
+    {
+        CustomEventArgs args = new(message);
+        QueryManagerFactory_ConfigChangeEventOccurred(args);
+        MetadataProviderFactory_ConfigChangeEventOccurred(args);
+        QueryEngineFactory_ConfigChangeEventOccurred(args);
+        MutationEngineFactory_ConfigChangeEventOccurred(args);
+        QueryExecutor_ConfigChangeEventOccurred(args);
+        MsSqlQueryExecutor_ConfigChangeEventOccurred(args);
+        MySqlQueryExecutor_ConfigChangeEventOccurred(args);
```
Aniruddh25, line 79: PostgreSqlExecutor is missing

src/Config/HotReloadEventHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+        public event EventHandler<TEventArgs>? MetadataProviderFactory_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? QueryEngineFactory_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? MutationEngineFactory_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? QueryExecutor_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? MsSqlQueryExecutor_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? MySqlQueryExecutor_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? PostgreSqlQueryExecutor_ConfigChangeEventOccurred;
```
Aniruddh25, line 15: Dont we have anything specific for CosmosDb?

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -61,7 +62,7 @@ public class OpenApiDocumentor : IOpenApiDocumentor
         /// </summary>
         /// <param name="sqlMetadataProvider">Provides database object metadata.</param>
         /// <param name="runtimeConfigProvider">Provides entity/REST path metadata.</param>
-        public OpenApiDocumentor(IMetadataProviderFactory metadataProviderFactory, RuntimeConfigProvider runtimeConfigProvider)
+        public OpenApiDocumentor(IMetadataProviderFactory metadataProviderFactory, RuntimeConfigProvider runtimeConfigProvider, HotReloadEventHandler<CustomEventArgs> handler)
```
Aniruddh25, line 65: how is the handler used here?

src/Config/HotReloadEventHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+namespace Azure.DataApiBuilder.Config
+{
+    public class HotReloadEventHandler<TEventArgs> where TEventArgs : CustomEventArgs
+    {
+        public event EventHandler<TEventArgs>? QueryManagerFactory_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? MetadataProviderFactory_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? QueryEngineFactory_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? MutationEngineFactory_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? QueryExecutor_ConfigChangeEventOccurred;
+        public event EventHandler<TEventArgs>? MsSqlQueryExecutor_ConfigChangeEventOccurred;
```
Aniruddh25, line 13: When do you instantiate all these event handlers? i.e. new EventHandler<TEventArgs> ?

Aniruddh25:
Some questions on CosmosDB/missing dependencies

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -94,6 +109,19 @@ public MsSqlQueryExecutor(
             }
         }
 
+        /// <summary>
+        /// Function registered for callback during a hot-reload scenario.
+        /// </summary>
+        /// <param name="sender">The calling object.</param>
+        /// <param name="args">Event arguments.</param>
+        public void MsSqlQueryExecutor_ConfigChangeEventReceived(object? sender, CustomEventArgs args)
```
seantleonard, line 117: Adding the prefix `MsSqlQueryExecutor_` to this function name add unnecessary redundancy. `MsSqlQueryExecutor` is already implied due to the class name.

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
```diff
@@ -16,31 +17,55 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
...
                 _metadataProviders.Add(dataSourceName, metadataProvider);
             }
         }
 
+        public void MetadataProviderFactory_ConfigChangeEventReceived(object? sender, CustomEventArgs args)
+        {
+            ConfigureMetadataProviders();
+        }
+
+
```
seantleonard, line 67:
```suggestion
```

src/Config/HotReloadEventHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+            MySqlQueryExecutor_ConfigChangeEventOccurred?.Invoke(sender, args);
+        }
+
+        public void MySqlQueryExecutor_Subscribe(EventHandler<TEventArgs> handler)
+        {
+            MySqlQueryExecutor_ConfigChangeEventOccurred += handler;
+        }
+
+        public void PostgreSqlQueryExecutor_OnConfigChangeEventOccurred(object sender, TEventArgs args)
+        {
```
seantleonard, line 88: there are no references to this

src/Config/RuntimeConfigLoader.cs
```diff
@@ -22,15 +22,66 @@ namespace Azure.DataApiBuilder.Config;
...
+    }
+    protected virtual void MySqlQueryExecutor_ConfigChangeEventOccurred(CustomEventArgs args)
+    {
+        Handler?.MySqlQueryExecutor_OnConfigChangeEventOccurred(this, args);
+    }
+
+    /// <summary>
+    /// Sends the notification to the event handler to trigger the hot-reload events
+    /// that have subscribed for configuration changes.
+    /// </summary>
```
seantleonard, line 68: Since order is important here, you should indicate that in the comment.

src/Config/HotReloadEventHandler.cs
aaronburtle: This are instantiated through the Subscribe function. In the subscribing class's constructor we take the HotReloadEventHandler instance that we created on startup, and we invoke its subscribe function, then we pass along the callback function as the eventhandler that is then assigned. When the event is later triggered it is this callback function that is then triggered.

src/Core/Resolvers/MsSqlQueryExecutor.cs
aaronburtle: That's true, but we need a way to make the callbacks disambiguous between the many different QueryExecutors.

src/Core/Resolvers/Factories/MutationEngineFactory.cs
```diff
@@ -56,11 +86,17 @@ public MutationEngineFactory(RuntimeConfigProvider runtimeConfigProvider,
 
             if (config.CosmosDataSourceUsed)
             {
-                IMutationEngine mutationEngine = new CosmosMutationEngine(cosmosClientProvider, metadataProviderFactory, authorizationResolver);
+                IMutationEngine mutationEngine = new CosmosMutationEngine(_cosmosClientProvider, _metadataProviderFactory, _authorizationResolver);
                 _mutationEngines.Add(DatabaseType.CosmosDB_NoSQL, mutationEngine);
             }
         }
 
+        public void MutationEngineFactory_ConfigChangeEventReceived(object? sender, CustomEventArgs args)
```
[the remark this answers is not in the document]
aaronburtle: Like with the QueryExecutor prefixes, if we don't include this then we would struggle to distinguish between the different events. We could have this function maintain a different name than the event in the handler, but the issue is that you literally pass this function along as the event, so keeping the names the same would be most clear.

`handler.Subscribe(nameof(MutationEngineFactoryOnConfigChanged), MutationEngineFactoryOnConfigChanged);`

src/Core/Resolvers/Factories/MutationEngineFactory.cs
```diff
@@ -18,7 +19,15 @@ namespace Azure.DataApiBuilder.Core.Resolvers.Factories
     /// </summary>
     public class MutationEngineFactory : IMutationEngineFactory
     {
-        private readonly Dictionary<DatabaseType, IMutationEngine> _mutationEngines;
+        private Dictionary<DatabaseType, IMutationEngine> _mutationEngines;
+        private readonly RuntimeConfigProvider _runtimeConfigProvider;
```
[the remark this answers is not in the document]
aaronburtle: before calling the ConfigureMutationEngine function we create a new dictionary for _mutationEngines. That creation of a new dictionary isn't done inside of the configure function because it needs to be set in the constructor since it is non null.

src/Core/Resolvers/Factories/QueryEngineFactory.cs
```diff
@@ -49,10 +83,15 @@ public QueryEngineFactory(RuntimeConfigProvider runtimeConfigProvider,
 
             if (config.CosmosDataSourceUsed)
             {
-                IQueryEngine queryEngine = new CosmosQueryEngine(cosmosClientProvider, metadataProviderFactory, authorizationResolver, gQLFilterParser, runtimeConfigProvider, cache);
+                IQueryEngine queryEngine = new CosmosQueryEngine(_cosmosClientProvider, _metadataProviderFactory, _authorizationResolver, _gQLFilterParser, _runtimeConfigProvider, _cache);
                 _queryEngines.Add(DatabaseType.CosmosDB_NoSQL, queryEngine);
             }
+        }
 
+        public void QueryEngineFactory_ConfigChangeEventReceived(object? sender, CustomEventArgs args)
```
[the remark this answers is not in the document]
aaronburtle: as in the other comments, just use to remove ambiguity.

src/Core/Resolvers/Factories/QueryManagerFactory.cs
```diff
@@ -77,6 +95,14 @@ public QueryManagerFactory(RuntimeConfigProvider runtimeConfigProvider, ILogger<
             }
         }
 
+        public void QueryManagerFactory_ConfigChangeEventReceived(object? sender, CustomEventArgs args)
```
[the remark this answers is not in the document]
aaronburtle: as in the other comments, just use to remove ambiguity.

src/Core/Resolvers/Factories/QueryManagerFactory.cs
```diff
@@ -49,23 +67,23 @@ public QueryManagerFactory(RuntimeConfigProvider runtimeConfigProvider, ILogger<
                         break;
                     case DatabaseType.MSSQL:
                         queryBuilder = new MsSqlQueryBuilder();
-                        exceptionParser = new MsSqlDbExceptionParser(runtimeConfigProvider);
-                        queryExecutor = new MsSqlQueryExecutor(runtimeConfigProvider, exceptionParser, logger, contextAccessor);
+                        exceptionParser = new MsSqlDbExceptionParser(_runtimeConfigProvider);
+                        queryExecutor = new MsSqlQueryExecutor(_runtimeConfigProvider, exceptionParser, _logger, _contextAccessor, _handler);
```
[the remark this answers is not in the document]
aaronburtle: Added the comment to the PR description. It is there in some classes because when they configure there state (such as during a hot-reload) they have classes they create which will need to be registering hot-reload events. The ones that don't save a handler dont need worry about this case.

src/Core/Resolvers/MySqlQueryExecutor.cs
```diff
@@ -84,6 +99,18 @@ public MySqlQueryExecutor(
             }
         }
 
+        /// <summary>
+        /// Function registered for callback during a hot-reload scenario.
+        /// </summary>
+        /// <param name="sender">The calling object.</param>
+        /// <param name="args">Event arguments.</param>
+        public void MySqlQueryExecutor_ConfigChangeEventReceived(object? sender, CustomEventArgs args)
```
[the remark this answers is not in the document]
aaronburtle: as in the other comments, just use to remove ambiguity.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
aaronburtle: was leftover from splitting branches, should be fixed now

src/Config/HotReloadEventHandler.cs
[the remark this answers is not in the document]
aaronburtle: I added this, very nice form, though I couldn't get nameof to work inside of this class because there are no references to those names so I just used hard coded strings.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -30,25 +30,77 @@ public abstract class RuntimeConfigLoader
...
+    {
+        _handler = handler;
+        _connectionString = connectionString;
+    }
+
+    protected virtual void QueryManagerFactoryOnConfigChangedEvent(HotReloadEventArgs args)
+    {
+        _handler?.OnConfigChangedEvent(nameof(QueryManagerFactoryOnConfigChangedEvent), this, args);
+    }
+
```
seantleonard, line 43: RE: the `QueryManagerFactoryOnConfigChangedEvent` prefix `QueryManagerFactory`.

`nameof(QueryManagerFactoryOnConfigChangedEvent)` is resolving the symbol name for `RuntimeConfigLoader.QueryManagerFactoryOnConfigChangedEvent` so you'd need to keep naming in sync in two locations (I'm assuming this is due to the ciruclar dependency issue between the Config and Core projects?):
1. `RuntimeConfigLoader.QueryManagerFactoryOnConfigChangedEvent`
2. `QueryManagerFactory.QueryManagerFactoryOnConfigChangedEvent`

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -60,7 +60,11 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// </summary>
     public string ConfigFilePath { get; internal set; }
 
-    public FileSystemRuntimeConfigLoader(IFileSystem fileSystem, HotReloadEventHandler<HotReloadEventArgs>? handler = null, string baseConfigFilePath = DEFAULT_CONFIG_FILE_NAME, string? connectionString = null)
+    public FileSystemRuntimeConfigLoader(
+        IFileSystem fileSystem,
+        HotReloadEventHandler<HotReloadEventArgs>? handler = null,
+        string baseConfigFilePath = DEFAULT_CONFIG_FILE_NAME, string?
```
seantleonard, line 66: `string?` needs to be on next line.

src/Config/RuntimeConfigLoader.cs
aaronburtle: Right, if we add a reference to the config proj to pickup the names from the classes in core we get that same cycle issue. I couldnt find a more graceful way to get the names to match aside from just maintaining two names that are the same in the different classes in sync.

seantleonard: if you had a class in the config project such as "EventNames", "Events", or "DabConfigEvents" with constants for the event name, that would be usable from the Core project because the config project is already imported. 

Then in the Core project code that subscribes to events, you could use "eventName = DabConfigEvents.MY_EVENT_NAME_1" and then drop the redundant prefix in MyFactory::MyFactoryEventOnChange.
Both the config and core projects would reference one event name constant so any updates in the constant would propagate to both projects the event name is used and nothing goes out of sync.

aaronburtle: Ya that makes sense, will add that in, thanks.

src/Config/HotReloadEventHandler.cs
aaronburtle: I am going to add Cosmos in its own PR because the reconfiguration could be more complicated and I don't want to delay all the other events if I run into a rabbit hole for cosmos.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -30,25 +31,77 @@ public abstract class RuntimeConfigLoader
...
+        _handler?.OnConfigChangedEvent(QUERY_EXECUTOR_ON_CONFIG_CHANGED, this, args);
+    }
+    protected virtual void MsSqlQueryExecutorOnConfigChangedEvent(HotReloadEventArgs args)
+    {
+        _handler?.OnConfigChangedEvent(MSSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, this, args);
+    }
+    protected virtual void MySqlQueryExecutorOnConfigChangedEvent(HotReloadEventArgs args)
+    {
+        _handler?.OnConfigChangedEvent(MYSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, this, args);
+    }
```
seantleonard, line 70: nit: new lines between functions.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -30,25 +31,77 @@ public abstract class RuntimeConfigLoader
...
+
+    protected virtual void QueryEngineFactoryOnConfigChangedEvent(HotReloadEventArgs args)
+    {
+        _handler?.OnConfigChangedEvent(QUERY_ENGINE_FACTORY_ON_CONFIG_CHANGED, this, args);
+    }
+
+    protected virtual void MutationEngineFactoryOnConfigChangedEvent(HotReloadEventArgs args)
+    {
+        _handler?.OnConfigChangedEvent(MUTATION_ENGINE_FACTORY_ON_CONFIG_CHANGED, this, args);
+    }
```
seantleonard, line 58: Seems like there's more opportunity for deduplicating code. Would this work:

Supplementing `HotReloadEventArgs` with the event name may enable you to consolidate the "*OnConfigChangedEvent" functions into one generic function `DabOnConfigChangedEvent(args)`.

And in that generic function, you'd do something specific based on the event name
```csharp
public void SendEventNotification(string message = "")
{
    HotReloadEventArgs args = new(message, eventName);
    GenericOnConfigChanged(new HotReloadEventArgs(eventName: DAB_EVENT_CONST_1, message: ""));
    GenericOnConfigChanged(new HotReloadEventArgs(eventName: DAB_EVENT_CONST_2, message: ""));
    GenericOnConfigChanged(new HotReloadEventArgs(eventName: DAB_EVENT_CONST_3, message: ""));
    /// etc
}

 protected virtual void GenericOnConfigChangedEvent(HotReloadEventArgs args)
{
    _handler?.OnConfigChangedEvent(this, args);
}
```

and in the HotReloadEventHandler class, pull the eventName out of the args object passed in:
```csharp

    public void OnConfigChangedEvent(object sender, TEventArgs args)
    {
        // use the arg.EventName property (I'm proposing that new property in the args class.)
        if (_eventHandlers.TryGetValue(args.eventName, out EventHandler<TEventArgs>? handler))
        {
            handler?.Invoke(sender, args);
        }
    }
```

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
seantleonard, line 1: In pr description you have:
> How was this tested?
> Manually verified the updating process.

Can you add more detail in pr description on how/when/what issue tracks adding tests for the order of operations? Can you also add detail for what you did to manually test/validate this change?

seantleonard (commented):
Questions about testing and feedback about possibility to further simplify changes

src/Config/RuntimeConfigLoader.cs
```diff
@@ -30,23 +31,36 @@ public abstract class RuntimeConfigLoader
...
-    public RuntimeConfigLoader(HotReloadEventHandler<HotReloadEventArgs>? handler = null, string? connectionString = null)
+    /// <summary>
+    /// Sends the notification to the event handler to trigger the hot-reload events
+    /// that have subscribed for configuration changes. The order here matters since
+    /// there are dependencies that we must refresh one after another. If adding to this
+    /// method make sure that you add the event trigger after any required dependencies have
+    /// been refreshed by previously called event triggers.
+    /// </summary>
+    /// <param name="message"></param>
+    public void SendEventNotification(string message = "")
```
seantleonard, line 53: Can this be protected because it is only called from `FileSystemRuntimeConfigLoader::HotReloadConfig(...)` and I don't think we intend for other components to be able to call `SendNotification(...)`

src/Core/Resolvers/Factories/QueryEngineFactory.cs
```diff
@@ -49,10 +84,15 @@ public QueryEngineFactory(RuntimeConfigProvider runtimeConfigProvider,
 
             if (config.CosmosDataSourceUsed)
             {
-                IQueryEngine queryEngine = new CosmosQueryEngine(cosmosClientProvider, metadataProviderFactory, authorizationResolver, gQLFilterParser, runtimeConfigProvider, cache);
+                IQueryEngine queryEngine = new CosmosQueryEngine(_cosmosClientProvider, _metadataProviderFactory, _authorizationResolver, _gQLFilterParser, _runtimeConfigProvider, _cache);
                 _queryEngines.Add(DatabaseType.CosmosDB_NoSQL, queryEngine);
             }
+        }
 
+        public void QueryEngineFactoryOnConfigChanged(object? sender, HotReloadEventArgs args)
```
seantleonard, line 92: because you have the separate `DabConfigEvents` constants, do these service method names need to be prefixed with the enclosing class name still? 
e.g. `QueryEngineFactory::QueryEngineFactoryOnConfigChanged` -> `QueryEngineFactory::OnConfigChanged`

seantleonard (approved):
Looks good, thanks for cleaning this up! Two last things as a result of all the changes made so far before merging.

src/Core/Resolvers/Factories/QueryEngineFactory.cs
aaronburtle: Can simplify the names for most, a few still need them to distinguish between base and child classes. I updated to reflect this.

seantleonard (approved):

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -57,27 +58,24 @@ public class OpenApiDocumentor : IOpenApiDocumentor
...
-
         /// <summary>
         /// Constructor denotes required services whose metadata is used to generate the OpenAPI description document.
         /// </summary>
         /// <param name="sqlMetadataProvider">Provides database object metadata.</param>
         /// <param name="runtimeConfigProvider">Provides entity/REST path metadata.</param>
         public OpenApiDocumentor(IMetadataProviderFactory metadataProviderFactory, RuntimeConfigProvider runtimeConfigProvider, HotReloadEventHandler<HotReloadEventArgs> handler)
         {
-            handler.DocumentorSubscribe(DocumentorOnConfigChanged);
+            handler.Subscribe(DOCUMENTOR_ON_CONFIG_CHANGED, OnConfigChanged);
```
Aniruddh25, line 68: Looks like for OpenApiDocumentor constructor, will we always need and have this handler? So, is no need to make it nullable like the others?

Aniruddh25 (approved):
LGTM. Thanks for pushing through this. Left a few minor comments.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
aaronburtle: As of now we will always have the handler when we make a new documentor, but it might be the case that we want to construct this class without a handler, in particular if we add tests that want to create a documentor, it would be nice if we don't necessarily have to mock this, so I will make it nullable for convenience.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -270,7 +270,7 @@ public async Task TestRetryPolicySuccessfullyExecutingQueryAfterNAttempts()
             Mock<IHttpContextAccessor> httpContextAccessor = new();
             DbExceptionParser dbExceptionParser = new MsSqlDbExceptionParser(provider);
             Mock<MsSqlQueryExecutor> queryExecutor
-                = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object);
+                = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object, null);
```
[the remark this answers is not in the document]
aaronburtle: the mock constructor here just takes an array of objects, but I made a null handler here instead now for clarity.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
aaronburtle: Added more detail to the PR description including an issue to track testing this specifically.

PR #2401, opened 2024-10-01
Fix DataSourceName update for HotReload

abhishekkumams:
## Why make this change?

the correct way of updating the dataSourceName was already there in the initial PR here: https://github.com/Azure/data-api-builder/blob/741865371911ef0b8dc6a27d9bc9ef82470cfc16/src/Config/ObjectModel/RuntimeConfig.cs#L303
 
then in another refactor PR, it was incorrectly updated, ref: https://github.com/Azure/data-api-builder/pull/2356/files#r1781334859

The current code directly assigns a new value to `DefaultDataSourceName`. However, this approach does not account for updating the different mappings related to DataSourceName when the application undergoes a hot reload. During a hot reload, it is crucial to maintain the same default data source name as before the hot reload to all the mappings for being consistent. This is because references to the data source depend on this name for lookups.

The new method `UpdateDefaultDataSourceName` ensures that the default data source name is updated correctly, maintaining consistency across hot reloads.

## What is this change?

- Using the correct initial method to update the DataSourceName for HotReload Scenario.

## How was this tested?

- Manual Testing
- More tests should be a part of other HotReload PRs


aaronburtle (approved):

RubenCerna2079:
In the tests section you wrote that "more tests should be part of other HotReload PRs". Do you mean tests related to the function `UpdateDefaultDataSourceName` or what did you mean by it?

abhishekkumams:
> In the tests section you wrote that "more tests should be part of other HotReload PRs". Do you mean tests related to the function `UpdateDefaultDataSourceName` or what did you mean by it?

correct integrations tests would be verifying this update logic as well.

seantleonard:
It would be helpful to have in PR description "Why" this change is correct and why the previous one is wrong

seantleonard (approved):
This enables my local environment to work with hot reload. However please add more to pr description for WHY this change works.

Issue #2396, opened 2024-09-25
Hot-Reload Validate New Config File


RubenCerna2079:
Once we get the new config file, we need to make sure that it has no errors before Hot-Reloading. Therefore, we need to find a way to see if there is a way to validate all the information that is in the new config file, so that if it is not validated then we will abort the Hot-Reload and stay with the parameters of the old config file.

PR #2405, opened 2024-10-02
Hot Reload Validation

RubenCerna2079:
## Why make this change?

This change fixes issue #2396 

## What is this change?

In order to validate the hot-reload, first, new logic was added to the `RuntimeConfigLoader` and `FileSystemRuntimeConfigLoader`, so when hot reload occurs, it updates the new states added to the loader which are isNewConfigDetected, isNewConfigValidated, and lastValidRuntimeConfig.
Once the states are updated, it will try to parse and validate the new RuntimeConfig, in the case that it fails at any step of the way, the hot-reloading process will be canceled and the RuntimeConfig file will be changed so that it uses the information from lastValidRuntimeConfig so that the user can still use their last configuration without any problems.

## How was this tested?

- [X] Integration Tests
- [ ] Unit Tests


src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -60,8 +60,8 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// </summary>
```
RubenCerna2079, line 60: My changes made from 161-169, 207-212

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -3,12 +3,14 @@
```
RubenCerna2079, line 3: I made all the changes in this file

src/Core/Services/OpenAPI/IOpenApiDocumentor.cs
```diff
@@ -26,6 +26,6 @@ public interface IOpenApiDocumentor
         /// <exception cref="DataApiBuilderException">Raised when document is already generated
```
RubenCerna2079, line 26: Made all the changes on this file

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -100,10 +110,10 @@ public bool TryGetDocument([NotNullWhen(true)] out string? document)
         /// <exception cref="DataApiBuilderException">Raised when document is already generated
         /// or a failure occurs during generation.</exception>
         /// <seealso cref="https://github.com/microsoft/OpenAPI.NET/blob/1.6.3/src/Microsoft.OpenApi/OpenApiSpecVersion.cs"/>
-        public void CreateDocument()
+        public void CreateDocument(bool isHotReloadScenario = false, bool isNewConfigDetected = false, bool isNewConfigValidated = true)
```
RubenCerna2079, line 113: My changes are line 113 & 115

src/Core/Services/OpenAPI/IOpenApiDocumentor.cs
RubenCerna2079, line 26: Made change on this file

src/Config/RuntimeConfigLoader.cs
```diff
@@ -22,15 +22,32 @@ namespace Azure.DataApiBuilder.Config;
...
+    public HotReloadEventHandler<CustomEventArgs>? Handler;
     protected readonly string? _connectionString;
 
     // Public to allow the RuntimeProvider and other users of class to set via out param.
     // May be candidate to refactor by changing all of the Parse/Load functions to save
     // state in place of using out params.
     public RuntimeConfig? RuntimeConfig;
 
-    public RuntimeConfigLoader(string? connectionString = null)
+    public RuntimeConfig? usedRuntimeConfig;
```
RubenCerna2079, line 33: Added line 33

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -72,8 +74,25 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
-    public RuntimeConfig GetConfig()
+    public RuntimeConfig GetConfig(bool isNewConfigDetected = false, bool isNewConfigValidated = true)
     {
+        //Only used in hot reload to validate the configuration file
+        if (isNewConfigDetected && !isNewConfigValidated)
+        {
+            IFileSystem fileSystem = new FileSystem();
+            ILoggerFactory loggerFactory = new LoggerFactory();
+            ILogger<RuntimeConfigValidator> logger = loggerFactory.CreateLogger<RuntimeConfigValidator>();
+            RuntimeConfigValidator runtimeConfigValidator = new(this, fileSystem, logger, true);
```
RubenCerna2079, line 85: Not sure if this is fine, since there is no logger inside the provider

src/Config/RuntimeConfigLoader.cs
RubenCerna2079: "usedRuntimeConfig" is the old runtime config that is already parsed and validated, the purpose is to use it in the case that the new config is not able to be parsed. So the user will receive a message that the new config is invalid, but they would be able to still use dab with the old config configuration.
I will change the name to "oldRuntimeConfig" or "validatedRuntimeConfig" so it doesn't cause confusion

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -198,7 +204,12 @@ public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? c
     public void HotReloadConfig(string defaultDataSourceName, ILogger? logger = null)
     {
         logger?.LogInformation(message: "Starting hot-reload process for config: {ConfigFilePath}", ConfigFilePath);
-        TryLoadConfig(ConfigFilePath, out _, replaceEnvVar: true, defaultDataSourceName: defaultDataSourceName);
+        if (!TryLoadConfig(ConfigFilePath, out _, replaceEnvVar: true, defaultDataSourceName: defaultDataSourceName))
+        {
+            logger?.LogWarning(message: "Unable to hot-reload configuration file due to failure to parse the file");
+        }
+
+        SendEventNotification("Sending Hot-Reload event notification.");
```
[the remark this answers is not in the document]
RubenCerna2079: I made it intentionally, since I thought it would be a good idea to also see if we get any exception from the validation and the parsing. Although, thinking about it once more, I will change it so the TryLoadConfig stops if it fails.

src/Config/FileSystemRuntimeConfigLoader.cs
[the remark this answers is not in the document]
RubenCerna2079: 1. Updated the description
2. I changed it so that `FileSystemRuntimeConfig::HotReloadConfig()` is the one that changes the state of isNewConfigValidated and isNewConfigDetected, so I believe that the internal logic to validate the config will only happen when there is a hot reload.

src/Core/Configurations/RuntimeConfigProvider.cs
seantleonard: For now, this is the only straightforward way of achieving validation during a hot reload.

My original response was the following, however we need to re-evaluate how we configure services in startup.cs such that we don't need to have runtimeconfigprovider resolved. That will need to be tracked via a separate task.
--------------------Obsolete-----------------------
There looks to be a better way of accomplishing instantiation of the `RuntimeConfigValidator`.

In Startup.cs:
```csharp
            FileSystemRuntimeConfigLoader configLoader = new(fileSystem, hotReloadEventHandler, configFileName, connectionString);
            RuntimeConfigValidator configValidator = new(
                runtimeConfigProvider: configProvider, 
                fileSystem: fileSystem, 
                logger: , // this is an issue, we can't resolve an ilogger in Startup::ConfigureServices
                isValidateOnly: true);
            services.AddSingleton(fileSystem);
            services.AddSingleton(configLoader);
            services.AddSingleton<configValidator>(); // instead of "services.AddSingleton<RuntimeConfigValidator>();"

File: src/Core/Configurations/RuntimeConfigProvider.cs [hunk 8]
```diff
@@ -74,6 +76,34 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
     /// <exception cref="DataApiBuilderException">Thrown when the loader is unable to load an instance of the config from its known location.</exception>
     public RuntimeConfig GetConfig()
     {
+        // Only used in hot reload to validate the configuration file
+        if (_configLoader.isNewConfigDetected && !_configLoader.isNewConfigValidated)
```
seantleonard, line 80: your pr description checks the box for "integration tests." how do existing integration tests check hot reload and this new functionality? I'd suggest outline in your PR description how you tested this, even if manually tested.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -74,6 +76,34 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
+            }
+            else
+            {
+                _configLoader.isNewConfigValidated = true;
+                _configLoader.RuntimeConfig = _configLoader.lastValidRuntimeConfig;
+
+                throw new DataApiBuilderException(
+                    message: "Failed validation of configuration file.",
+                    statusCode: HttpStatusCode.ServiceUnavailable,
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
```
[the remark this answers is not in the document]
RubenCerna2079: Maybe we do not need to raise an exception, but at the very least there should be a warning that would allow the user to know that hot-reload failed.

src/Core/Configurations/RuntimeConfigProvider.cs
RubenCerna2079: I was not able to find the code you put from Startup.cs and that way of instantiating `RuntimeConfigValidator` did not work for me.

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -74,6 +76,34 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
+            }
```
Aniruddh25, line 104: how have we tested this? Is there a way to add an automated integration test in ConfigurationTests.cs?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -31,6 +31,12 @@ public abstract class RuntimeConfigLoader
     // state in place of using out params.
     public RuntimeConfig? RuntimeConfig;
 
+    public RuntimeConfig? lastValidRuntimeConfig;
```
Aniruddh25, line 34: All these properties are public properties. So, naming convention wise they should be capitalized. LastValidRuntimeConfig, IsNewConfigDetected, IsNewConfigValidated.

Aniruddh25 (changes requested):
Waiting for testing.

src/Config/FileSystemRuntimeConfigLoader.cs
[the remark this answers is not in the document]
RubenCerna2079: 3. The changes I made, make it so that it only makes the validation on the first `OnConfigChangeEvent` while the others use the result.

src/Core/Configurations/RuntimeConfigProvider.cs
seantleonard: Right, that's what I meant by "obsolete." The code sample i provided isn't possible, but in an ideal world, it would be. To be doable, other refactors are necessary. That's the background behind my comment of "For now, this (your current code change) is the only straightforward way of achieving validation during a hot reload."

src/Config/RuntimeConfigLoader.cs
```diff
@@ -31,6 +31,12 @@ public abstract class RuntimeConfigLoader
...
+
+    public bool isNewConfigDetected;
+
+    public bool isNewConfigValidated = true;
```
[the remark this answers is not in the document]
seantleonard: sorry not sure why both vars were highlighted, I meant `public bool isNewConfigValidated = true;` should `isNewConfigValidated ` be defaulted to false? e.g. a config is invalid until proven otherwise.

aaronburtle:
This is great and the way that you update with the Last Known Good will close this bug here: https://github.com/Azure/data-api-builder/issues/2418

So you can added the closes tag for that issue to the PR description.

One you address the comments should be good to go.

Aniruddh25 (approved):

Aniruddh25 (changes requested):
Still waiting to know how this was tested, and resolution of some comments

src/Core/Configurations/RuntimeConfigProvider.cs
RubenCerna2079: Tests were done manually, by using the placeholder schema url as an invalid configuration schema. I will see if there is a way to add automated integration tests.

src/Core/Resolvers/Factories/QueryManagerFactory.cs
```diff
@@ -102,6 +102,11 @@ private void ConfigureQueryManagerFactory()
 
         public void OnConfigChanged(object? sender, HotReloadEventArgs args)
         {
+            //First use of GetConfig during hot reload, in order to do validation of
+            //config file before any changes are made for hot reload.
+            //In case validation fails, an exception will be thrown and hot reload will be canceled.
+            _runtimeConfigProvider.GetConfig();
```
Aniruddh25, line 108: Is this really the first use of GetConfig? why is this inside an OnConfigChanged of one particular class? Can we move it before OnConfigChanged function of this class is called?

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +198,164 @@ public void HotReloadConfigRestRuntimeOptions()
...
+                DataSource: dataSource,
+                Runtime: new(
+                    Rest: new(initialRestEnabled),
+                    GraphQL: new(initialGQLEnabled),
+                    Host: new(null, null, HostMode.Development)
+                ),
+                Entities: new(new Dictionary<string, Entity>())
+            );
+
+            string updatedConfig = GenerateWrongRuntimeSection();
```
Aniruddh25, line 334: The previous test and this differ only in the fact that they generate a different wrong section. Can we somehow have a common function to minimize code duplication?

Aniruddh25 (changes requested):
The first caller of GetConfig need not be QuerymanagerFactory. Need to find a better placement for the call.

src/Core/Resolvers/Factories/QueryManagerFactory.cs
RubenCerna2079: It is the first use after the creating the new config, I put it in that class specifically because it is the first one to be loaded and in order to not allow the program to start saving new information that is still not validated.
OnConfigChanged is called from `RuntimeConfigLoader` which calls the previous function from `FileSystemRuntimeConfigLoader` and neither of them has an object `RuntimeConfigProvider` that would allow us to call GetConfig.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -61,6 +65,37 @@ HostMode mode
             return runtimeString;
         }
 
+        private static string GenerateWrongRuntimeSection()
```
seantleonard, line 68: add comment to clarify what you mean by "wrong"

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -61,6 +65,37 @@ HostMode mode
...
+    }
+  },
+  ""runtime"": {
+    ""rest"": {
+      ""enabled"":
+}";
+            return runtimeString;
+        }
+
+        private static string GenerateWrongSchema()
```
seantleonard, line 87: Comment to describe what "wrong" means

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +198,164 @@ public void HotReloadConfigRestRuntimeOptions()
                 File.Delete(configName);
             }
         }
+
+        /// <summary>
+        /// Creates a hot reload scenario in which the schema is wrong which causes
+        /// hot reload to fail, then we check that the program is still able to work
+        /// properly by showing us that it is still using the same configuration file
+        /// from before the hot reload.
```
seantleonard, line 206: - "hot reload scenario in which the schema is wrong" clarify what you mean by "wrong"

/// properly by validating that the DAB engine is still using the same configuration file
/// from before the hot reload.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +198,164 @@ public void HotReloadConfigRestRuntimeOptions()
...
+        /// </summary>
+        [TestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        public void HotReloadValidationFail()
```
seantleonard, line 210: These two new tests don't really belong in ConfigFileWatcherUnitTests since they aren't related to ConfigFileWatcher

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +198,164 @@ public void HotReloadConfigRestRuntimeOptions()
...
+            System.Threading.Thread.Sleep(1000);
+
+            try
+            {
+                configProvider.GetConfig();
+            }
+            catch (DataApiBuilderException dbException)
+            {
+                Assert.AreEqual(expected: "Failed validation of configuration file.", actual: dbException.Message);
+            }
```
seantleonard, line 282: Can you add a comment for what you're validating here?

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +198,164 @@ public void HotReloadConfigRestRuntimeOptions()
...
+            {
+                File.Delete(schemaName);
+            }
+        }
+
+        /// <summary>
+        /// Creates a hot reload scenario in which the updated configuration file is wrong so
+        /// hot reload to fail, then we check that the program is still able to work
+        /// properly by showing us that it is still using the same configuration file
+        /// from before the hot reload.
```
seantleonard, line 302: So the test here is demonstrating that runtimeconfigLoader and runtimeconfigProvider properly coordinate validating the new config and setting LKG config and restoring LKG config if a validation error occurs? Just want to clarify this isn't testing hot reload's effect on all downstream dependencies.

RubenCerna2079: Yes, the test is only demonstrating what you said, it isn't testing any effect on the downstream dependencies.

src/Core/Resolvers/Factories/QueryManagerFactory.cs
seantleonard: This code shows the event handler event raising and then lastly has the change token notification:
https://github.com/Azure/data-api-builder/blob/c9aa87b0257507b1ea223953cfc8b1c76928610b/src/Config/RuntimeConfigLoader.cs#L81-L91

In the last sentence of last response, What I wasn't sure of was whether `RaiseChanged(); ` for change token notification was a blocking call. Aaron explained that the event handler event raising was blocking. That's an important distinction because you'd want to move `RaiseChanged()` to come before ` OnConfigChangedEvent()` but you'd also want the operations trigger by `RaiseChanged()` (Config Validation that you add in this PR for hot reload) to complete before ` OnConfigChangedEvent()` code is executed

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -112,6 +114,33 @@ public string ConfigFilePath
...
+                _configLoader.CreateNewLkgConfig();
+            }
+            else
+            {
+                _configLoader.RestoreLkgConfig();
+
+                throw new DataApiBuilderException(
+                    message: "Failed validation of configuration file.",
+                    statusCode: HttpStatusCode.ServiceUnavailable,
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
```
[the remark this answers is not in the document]
RubenCerna2079: image

This is how it looks right now for the user, it doesn't show on dab itself and it just shows in the console, and in the lower right corner you can see there are multiple responses to queries that were done before and after trying the hot reload validation and they all worked.
We could add a log besides the exception, but I would still like to keep the exception as it stops the whole hot reload process completely, without having to do any additional changes

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -112,6 +114,33 @@ public string ConfigFilePath
...
+
+            // Saves the lastValidRuntimeConfig as the new RuntimeConfig if it is validated for hot reload
+            if (_configLoader.IsNewConfigValidated)
+            {
+                _configLoader.CreateNewLkgConfig();
+            }
+            else
+            {
+                _configLoader.RestoreLkgConfig();
+
```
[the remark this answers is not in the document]
RubenCerna2079: Put the `_configLoader.IsNewConfigDetected = false` inside of `_configLoader.DoesConfigNeedValidation()` because it needs to go before `TryValidateConfig` as if it were to go after it would cause an infinite loop.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
RubenCerna2079: Changed to invalid and added comments

src/Core/Resolvers/Factories/QueryManagerFactory.cs
RubenCerna2079: From my testing it looks like RaiseChanged completes its execution before OnConfigChangedEvent starts

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -65,6 +65,11 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
     /// </seealso>
     private void RaiseChanged()
     {
+        //First use of GetConfig during hot reload, in order to do validation of
```
Aniruddh25, line 68:
```suggestion
        // First use of GetConfig during hot reload, in order to do validation of
```

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -65,6 +65,11 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
+        //config file before any changes are made for hot reload.
```
Aniruddh25, line 69:
```suggestion
        // config file before any changes are made for hot reload.
```

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -65,6 +65,11 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
...
+        //In case validation fails, an exception will be thrown and hot reload will be canceled.
```
Aniruddh25, line 70:
```suggestion
        // In case validation fails, an exception will be thrown and hot reload will be canceled.
```

Aniruddh25 (approved):
LGTM!

seantleonard (approved):

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -112,6 +119,32 @@ public string ConfigFilePath
     /// <exception cref="DataApiBuilderException">Thrown when the loader is unable to load an instance of the config from its known location.</exception>
     public RuntimeConfig GetConfig()
     {
+        // Only used in hot reload to validate the configuration file
+        if (_configLoader.DoesConfigNeedValidation())
```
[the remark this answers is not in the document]
Aniruddh25: good callout, agree!

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -65,10 +65,17 @@ public RuntimeConfigProvider(RuntimeConfigLoader runtimeConfigLoader)
     /// </seealso>
     private void RaiseChanged()
     {
+<<<<<<< HEAD
```
seantleonard, line 68: make sure to get rid of the git merge conflict text

PR #2414, opened 2024-10-10
AuthN Config V2 -> HotReload Aware Authentication settings in dev mode.

### LINKED ISSUE #1969: Hot Reload Authentication settings
The authentication settings are configured during startup. This means that simply refreshing the `RuntimeConfig` to reflect saved changes is not enough to update the authentication settings that the service will be using. Instead, those changes that need to be made must be done so in a way that refreshes the settings that were configured during startup.

seantleonard:
## Why make this change?

- Closes #1969 which tracks making DAB's authentication configuration hot-reload aware.

## What is this change?
The following changes are only honored in development mode when testing locally using "Hot Reload"
- Introduces `ConfigureAuthenticationV2()` in startup.cs which wires up all DAB supported Authentication providers: 
  - EasyAuth AppService/StaticWebApps
  - JWT (AzureAD/EntraID, and Custom OAuth2 providers)
- Utilizes change tokens within RuntimeConfigProvider and RuntimeConfigLoader to signal IOptionsMonitor<JwtBearerHandler> that authentication settings have changed.
- Update ClientRoleHeaderAuthenticationMiddleware to emit condensed and more helpful log event for each request:
> AuthN state: Anonymous. Role: Anonymous. Scheme: Bearer

  - The Scheme metadata in the log event will indicate which scheme the authentication handler is using. This will help users observe that hot reload has succeeded.

## How was this tested?

- [x] Manual tests with real identity provider that we don't have in our pipelines. This is a development mode only feature.

### Scenario 1 - Swap JWT Provider Details

1. Open the dab-config.json file used by DAB.
2. Ensure the config sets `Mode: Development`
3. Modify the Authentication section to include JWT authN with your AzureAD provider (issuer/audience).
4. Send a request which requires validation: see HTTP 200
5. Modify the Authentication section with an "invalid" audience
6. Select "SAVE" -> hot reload will occur
7. Send a new request using the previously used token -> HTTP 401 invalid issuer.

### Scenario 2 - Swap from JWT to Simulator

1. Open the dab-config.json file used by DAB.
2. Ensure the config sets `Mode: Development`
3. Modify the Authentication section to include JWT authN with your AzureAD provider (issuer/audience).
4. Send a request which requires validation: see HTTP 200
5. Modify the Authentication section by removing the jwt section with audience/issuer and leave only the "Provider" property
6. Modify the authentication provider property to be Simulator
7. Select "SAVE" -> hot reload will occur
8. Send a new request using the previously used token -> authenticated 

## Usage of ChangeTokens versus EventHandler in this specific workstream:

The objective was to trigger the `IOptionsMonitor` within the `JwtBearerHandler` to detect changes to authentication programmatically (as opposed to the out-of-box change detection for a bound configuration file such as `appsettings.json`, which we don't use). In order to alert the `IOptionsMonitor` that a change has occurred, I needed to register:
```csharp
services.AddSingleton<IOptionsChangeTokenSource<JwtBearerOptions>>(new JwtBearerOptionsChangeTokenSource(runtimeConfigProvider));
```
which is this resolved via DI in the `OptionsMonitor` class:
https://github.com/dotnet/runtime/blob/88397049d6aa2c8505bd11309ddd314169e3f73f/src/libraries/Microsoft.Extensions.Options/src/OptionsMonitor.cs#L31-L53

```csharp
        public OptionsMonitor(IOptionsFactory<TOptions> factory, IEnumerable<IOptionsChangeTokenSource<TOptions>> sources, IOptionsMonitorCache<TOptions> cache)
        {
            _factory = factory;
            _cache = cache;

            void RegisterSource(IOptionsChangeTokenSource<TOptions> source)
            {
                IDisposable registration = ChangeToken.OnChange(
                          source.GetChangeToken,
                          InvokeChanged,
                          source.Name);

                _registrations.Add(registration);
            }

            // The default DI container uses arrays under the covers. Take advantage of this knowledge
            // by checking for an array and enumerate over that, so we don't need to allocate an enumerator.
            if (sources is IOptionsChangeTokenSource<TOptions>[] sourcesArray)
            {
                foreach (IOptionsChangeTokenSource<TOptions> source in sourcesArray)
                {
                    RegisterSource(source);
                }
```

once our `runtimeconfigloader` detects a file change, it signals the provider that new `runtimeconfig` is available. Then the `RuntimeConfigProvider` (which is the service resolved via DI in most of our classes already) can signal the change token whose listener is the `OptionsMonitor` via registering the `IOptionsChangeTokenSource`. 

This mechanism was not as intrusive to our services' constructors:
1. No need to resolve `IOptionsMonitorCache<jwtBearerOptions>` in our `RuntimeConfigProvider` or `RuntimeConfigLoader` class constructors. Updating the `optionsmonitorcache` would be( i tested this) the manual way of changing `IOptionsMonitor<JwtBearerOptions>` used by the `JwtBearerHandler`, but this didn't work because of the next bullet point: 
    2. I made a judgement call that injecting `IoptionsmonitorCache<jwtBearerOptions>` into either of the two referenced classes seemed to leak implementation details of the provider/loader and wouldn't be straightforward without more time consuming and costly refactors:
        -  Because both the provider/loader are manually instantiated in `Startup::ConfigureServices` and don't have the opportunity to resolve services via DI.
        - Because we'd then need to make even more cascading changes in the test projects to accommodate setting up the `ioptionsmonitorcache` object to then be added to our mock runtimeconfigprovider/loader objects.

The event-handlers we have recently added don't fit this specific use case of updating the underlying authentication configuration. In addition to the above design decisions, I didn't identify a solution where we could subscribe the `JwtBearerHandler`'s `OptionsMonitor`  to an eventhandler signal to notify that a change occurred. I pursued using the built-in mechanism of change detection for optionsmonitor provided to us via `IOptionsChangeTokenSource`.


src/Config/RuntimeConfigLoader.cs
```diff
@@ -30,6 +31,39 @@ public abstract class RuntimeConfigLoader
...
+    /// signals that a change has occurred.
+    /// </summary>
+    /// <seealso cref="https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration/src/ConfigurationProvider.cs">
+    /// Example usage of Interlocked.Exchange(...) to refresh change token.</seealso>
+    /// <seealso cref="https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.exchange?view=net-8.0">
+    /// Sets a variable to a specified value as an atomic operation.
+    /// </seealso>
+    private void RaiseChanged()
+    {
+        DabChangeToken previousToken = Interlocked.Exchange(ref _changeToken, new DabChangeToken());
```
Aniruddh25, line 63: Why is a change token mechanism required as opposed to the event handlers that have already been added? What is different in signaling a change has occurred v/s an event has happened? Both seem to have subscribers...

seantleonard: The objective was to trigger the `IOptionsMonitor` within the `JwtBearerHandler` to detect changes to authentication programmatically (as opposed to the out-of-box change detection for a bound configuration file such as `appsettings.json`, which we don't use). In order to alert the `IOptionsMonitor` that a change has occurred, I needed to register:
```csharp
services.AddSingleton<IOptionsChangeTokenSource<JwtBearerOptions>>(new JwtBearerOptionsChangeTokenSource(runtimeConfigProvider));
```
which is this resolved via DI in the `OptionsMonitor` class:
https://github.com/dotnet/runtime/blob/88397049d6aa2c8505bd11309ddd314169e3f73f/src/libraries/Microsoft.Extensions.Options/src/OptionsMonitor.cs#L31-L53

```csharp
        public OptionsMonitor(IOptionsFactory<TOptions> factory, IEnumerable<IOptionsChangeTokenSource<TOptions>> sources, IOptionsMonitorCache<TOptions> cache)
        {
            _factory = factory;
            _cache = cache;

            void RegisterSource(IOptionsChangeTokenSource<TOptions> source)
            {
                IDisposable registration = ChangeToken.OnChange(
                          source.GetChangeToken,
                          InvokeChanged,
                          source.Name);

                _registrations.Add(registration);
            }

            // The default DI container uses arrays under the covers. Take advantage of this knowledge
            // by checking for an array and enumerate over that, so we don't need to allocate an enumerator.
            if (sources is IOptionsChangeTokenSource<TOptions>[] sourcesArray)
            {
                foreach (IOptionsChangeTokenSource<TOptions> source in sourcesArray)
                {
                    RegisterSource(source);
                }
```

once our `runtimeconfigloader` detects a file change, it signals the provider that new `runtimeconfig` is available. Then the `RuntimeConfigProvider` (which is the service resolved via DI in most of our classes already) can signal the change token whose listener is the `OptionsMonitor` via registering the `IOptionsChangeTokenSource`. 

This mechanism was not as intrusive to our services' constructors:
1. No need to resolve `IOptionsMonitorCache<jwtBearerOptions>` in our `RuntimeConfigProvider` or `RuntimeConfigLoader` class constructors. Updating the `optionsmonitorcache` would be( i tested this) the manual way of changing `IOptionsMonitor<JwtBearerOptions>` used by the `JwtBearerHandler`, but this didn't work because of the next bullet point: 
    2. I made a judgement call that injecting `IoptionsmonitorCache<jwtBearerOptions>` into either of the two referenced classes seemed to leak implementation details of the provider/loader and wouldn't be straightforward without more time consuming and costly refactors:
        -  Because both the provider/loader are manually instantiated in `Startup::ConfigureServices` and don't have the opportunity to resolve services via DI.
        - Because we'd then need to make even more cascading changes in the test projects to accommodate setting up the `ioptionsmonitorcache` object to then be added to our mock runtimeconfigprovider/loader objects.

The event-handlers we have recently added don't fit this specific use case of updating the underlying authentication configuration. In addition to the above design decisions, I didn't identify a solution where we could subscribe the `JwtBearerHandler`'s `OptionsMonitor`  to an eventhandler signal to notify that a change occurred. I pursued using the built-in mechanism of change detection for optionsmonitor provided to us via `IOptionsChangeTokenSource`.

seantleonard: adding the above response to the PR description

src/Core/AuthenticationHelpers/AuthenticationSimulator/SimulatorAuthenticationHandler.cs
```diff
@@ -85,7 +85,7 @@ protected override Task<AuthenticateResult> HandleAuthenticateAsync()
 
         // AuthenticationTicket is Asp.Net Core Abstraction of Authentication information
         // of the authenticated user.
-        AuthenticationTicket ticket = new(claimsPrincipal, EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME);
+        AuthenticationTicket ticket = new(claimsPrincipal, SimulatorAuthenticationDefaults.AUTHENTICATIONSCHEME);
```
abhishekkumams, line 88: why this change?

src/Core/AuthenticationHelpers/EasyAuthAuthenticationBuilderExtensions.cs
```diff
@@ -27,20 +27,65 @@ public static AuthenticationBuilder AddEasyAuthAuthentication(
...
 
-        builder.AddScheme<EasyAuthAuthenticationOptions, EasyAuthAuthenticationHandler>(
-            authenticationScheme: EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME,
-            displayName: EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME,
-            options =>
-            {
-                if (easyAuthAuthenticationProvider is EasyAuthType.StaticWebApps)
+        if (easyAuthAuthenticationProvider is EasyAuthType.StaticWebApps)
+        {
+            //builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<EasyAuthAuthenticationOptions>, EasyAuthConfigureOptions>());
```
abhishekkumams, line 32: is this comment required?

src/Core/AuthenticationHelpers/EasyAuthAuthenticationBuilderExtensions.cs
```diff
@@ -27,20 +27,65 @@ public static AuthenticationBuilder AddEasyAuthAuthentication(
...
+
+        builder.AddScheme<EasyAuthAuthenticationOptions, EasyAuthAuthenticationHandler>(
+           authenticationScheme: EasyAuthAuthenticationDefaults.SWAAUTHSCHEME,
+           displayName: EasyAuthAuthenticationDefaults.SWAAUTHSCHEME,
+           options =>
+           {
+               options.EasyAuthProvider = EasyAuthType.StaticWebApps;
+           });
+
+        bool appServiceEnvironmentDetected = AppServiceAuthenticationInfo.AreExpectedAppServiceEnvVarsPresent();
```
abhishekkumams, line 76: should we log if appServiceEnvironment is detected or not? could be useful for debugging.

src/Service/Startup.cs
```diff
@@ -174,7 +175,16 @@ public void ConfigureServices(IServiceCollection services)
             //Enable accessing HttpContext in RestService to get ClaimsPrincipal.
             services.AddHttpContextAccessor();
 
-            ConfigureAuthentication(services, configProvider);
+            // Use HotReload aware authentication configuration when in development mode
```
abhishekkumams, line 178: What is `HotReload aware authentication configuration`?

src/Service/Startup.cs
```diff
@@ -524,6 +535,23 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
             }
         }
 
+        /// <summary>
+        /// Wires up all DAB supported authentication providers. DAB uses the user configured
+        /// authentcation provider to determine which provider to use for authentication at request time.
```
abhishekkumams, line 540:
```suggestion
        /// authentication provider to determine which provider to use for authentication at request time.
```

src/Service/Startup.cs
```diff
@@ -524,6 +535,23 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+        /// JwtBearerOptions are hydrated when ConfigureJwtBearerOptions is called by OptionsFactory internally by .NET.
+        /// </summary>
+        /// <seealso cref="https://github.com/dotnet/aspnetcore/issues/49586#issuecomment-1671838595">Guidance for registering IOptionsChangeTokenSource</seealso>
+        /// <seealso cref="https://github.com/dotnet/aspnetcore/issues/21491#issuecomment-624240160">Guidance for registering named options.</seealso>
+        private static void ConfigureAuthenticationV2(IServiceCollection services, RuntimeConfigProvider runtimeConfigProvider)
+        {
+            services.AddSingleton<IOptionsChangeTokenSource<JwtBearerOptions>>(new JwtBearerOptionsChangeTokenSource(runtimeConfigProvider));
+            services.AddSingleton<IConfigureOptions<JwtBearerOptions>, ConfigureJwtBearerOptions>();
```
abhishekkumams, line 548: would be good to add a comment stating something like
```
// Register the JWT Bearer options change token source for dynamic config updates
```

src/Service/Startup.cs
seantleonard: I used that phrasing to describe the purpose of the `ConfigureAuthenticationV2()` function. That function wires up all authentication methods unconditionally. Today, we conditionally wire up authentication methods. 

Authentication method (scheme) configuration must occur at startup. This PR adds the newly referenced "V2" function which wires up all the schemes together. Then, DAB will reference the configured authN provider at request time to determine which authentication scheme to use to authenticate.

src/Service/Startup.cs
seantleonard: yes, good callout.

src/Core/AuthenticationHelpers/EasyAuthAuthenticationBuilderExtensions.cs
seantleonard: nope, thanks for catching.

src/Core/AuthenticationHelpers/AuthenticationSimulator/SimulatorAuthenticationHandler.cs
seantleonard: It didn't make sense that `EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME` was used to denote the scheme used to authenticate a Simulator request. 

Looking at source: https://github.com/dotnet/aspnetcore/blob/3f1acb59718cadf111a0a796681e3d3509bb3381/src/Http/Authentication.Abstractions/src/AuthenticationTicket.cs

I don't observe the authenticationScheme serving functional purpose besides acting as metadata.

src/Core/AuthenticationHelpers/EasyAuthAuthenticationBuilderExtensions.cs
seantleonard: I'm adding logging for case when appservice IS detected, so that it is obvious we are registering the scheme. Doesn't make sense to log when it isn't detected since this would log even when SWA is used.

src/Service/Startup.cs
seantleonard: Updated function comment with details.

src/Service/Startup.cs
```diff
@@ -524,6 +536,27 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+        /// - IOptionsChangeTokenSource<JwtBearerOptions> : Registers a change token source for dynamic config updates which
+        /// is used internally by JwtBearerHandler's OptionsMonitor to listen for changes in JwtBearerOptions.
+        /// - IConfigureOptions<JwtBearerOptions> : Registers named JwtBearerOptions whose "Configure(...)" function is
+        /// called by OptionsFactory internally by .NET to fetch the latest configuration from the RuntimeConfigProvider.
+        /// </summary>
+        /// <seealso cref="https://github.com/dotnet/aspnetcore/issues/49586#issuecomment-1671838595">Guidance for registering IOptionsChangeTokenSource</seealso>
+        /// <seealso cref="https://github.com/dotnet/aspnetcore/issues/21491#issuecomment-624240160">Guidance for registering named options.</seealso>
+        private static void ConfigureAuthenticationV2(IServiceCollection services, RuntimeConfigProvider runtimeConfigProvider)
+        {
+            services.AddSingleton<IOptionsChangeTokenSource<JwtBearerOptions>>(new JwtBearerOptionsChangeTokenSource(runtimeConfigProvider));
```
Aniruddh25, line 552: In V1, we added authentication conditionally. But in V2, we are adding all schemes unconditionally. Is there benefit in keeping V1 around? If not, to keep the code base simple, is it better to do V2 irrespective of whether its a hot reload scenario or not?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -61,6 +89,9 @@ protected void SendEventNotification(string message = "")
         OnConfigChangedEvent(new HotReloadEventArgs(MYSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, message));
         OnConfigChangedEvent(new HotReloadEventArgs(POSTGRESQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, message));
         OnConfigChangedEvent(new HotReloadEventArgs(DOCUMENTOR_ON_CONFIG_CHANGED, message));
+
```
Aniruddh25, line 92: nit: rename to indicate this function also handles signaling the change token. Currently, it says send event notification only.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -57,23 +59,41 @@ public ClientRoleHeaderAuthenticationMiddleware(RequestDelegate next,
...
             return;
         }
 
-        string clientDefinedRole = AuthorizationType.Anonymous.ToString();
+        // Manually set the httpContext.User to the Principal from the AuthenticateResult
+        // when we exclude setting a default authentication scheme in Startup.cs AddAuthentication().
+        // https://learn.microsoft.com/aspnet/core/security/authorization/limitingidentitybyscheme?view=aspnetcore-8.0
+        if (authNResult.Succeeded)
+        {
+            httpContext.User = authNResult.Principal;
```
Aniruddh25, line 93: This setup of httpContext.User is the only required change for hot reload in this file,right? Is my understanding correct that the rest of the changes in this file are good to have but not really relevant to hot reload?

I dont mind the changes but trying to understand how the value of scheme affects hot reload..

src/Service/JwtBearerOptionsChangeTokenSource.cs
```diff
@@ -0,0 +1,44 @@
...
+/// (via RuntimeConfigLoader -> RuntimeConfigProvider -> JwtBearerOptionsChangeTokenSource)
+/// that a change has occurred and IOptionsMonitor will reload the JwtBearerOptions.
+/// </summary>
+/// <seealso cref="https://github.com/dotnet/aspnetcore/issues/49586#issuecomment-1671838595"/>
+public class JwtBearerOptionsChangeTokenSource : IOptionsChangeTokenSource<JwtBearerOptions>
+{
+    private readonly RuntimeConfigProvider _configProvider;
+
+    /// <summary>
+    /// Get RuntimeConfigProvider to use as the change event source.
```
Aniruddh25, line 23:
```suggestion
    /// Get RuntimeConfigProvider to use as the change token source.
```

src/Core/AuthenticationHelpers/EasyAuthAuthenticationDefaults.cs
```diff
@@ -4,14 +4,27 @@
 namespace Azure.DataApiBuilder.Core.AuthenticationHelpers;
 
 /// <summary>
-/// Default values related to EasyAuthAuthentication handler.
+/// EasyAuth authentication scheme names granularized by provider
+/// to enable compatility with HotReloading authentication settings.
```
Aniruddh25, line 8:
```suggestion
/// to enable compatibility with HotReloading authentication settings.
```

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
seantleonard: We need all the code in this file and on lines that precede this code. The following line is important because we must pass scheme to authenticateAsync to ensure the proper authentication handler is used. Scheme is determined using the runtime config. If the provider changes in runtime config, DAB must resolve that change and ensure the request is authenticated using the hotreloaded scheme.
```csharp
await httpContext.AuthenticateAsync(scheme);
```

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
```diff
@@ -41,17 +41,24 @@ public static class WebHostBuilderHelper
         /// - DAB's Simulator/ EasyAuth authentication middleware and ClientRoleHeader middleware
         /// - dotnet's authorization middleware.
         /// </summary>
-        /// <param name="provider">Runtime configured identity provider name.</param>
+        /// <param name="provider">Runtime configured identity provider name. This is different that authentication scheme name because
```
Aniruddh25, line 44:
```suggestion
        /// <param name="provider">Runtime configured identity provider name. This is different than authentication scheme name because
```

Aniruddh25 (approved):
LGTM, left few nits and questions. Thanks for this change.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -151,6 +170,34 @@ public static bool IsSystemRole(string roleName)
...
+        switch (configuredProviderName)
+        {
+            case AuthenticationOptions.SIMULATOR_AUTHENTICATION:
+                return SimulatorAuthenticationDefaults.AUTHENTICATIONSCHEME;
+            case nameof(EasyAuthType.AppService):
+                return EasyAuthAuthenticationDefaults.APPSERVICEAUTHSCHEME;
+            case nameof(EasyAuthType.StaticWebApps):
+                return EasyAuthAuthenticationDefaults.SWAAUTHSCHEME;
+            case "AzureAD":
+            case "EntraID":
```
sezal98, line 191: should we make these constants and use case-insensitive matching?

abhishekkumams (approved):
LGTM, thanks for the quick update and clarifying doubts.

src/Config/RuntimeConfigLoader.cs
seantleonard: updating to `SignalConfigChanged` from `SendEventNotification`

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
seantleonard: yes, good call and this is absolutely necessary. added.

src/Service/Startup.cs
seantleonard: Once we get full hot reload testing in, we can remove older code. I want the validated auth functions to remain for now in Prod mode.

PR #2424, opened 2024-10-14
Hot Reload File IO -> React to 1 file change notification per file change

### LINKED ISSUE #2423: Hot Reload - File IO issues -> DAB reacts to multiple file change notifications for single file change
When saving a file, I've observed that multiple file changed events are triggered. This is something that the dotnet team acknowledges and provides sample code in the documentation to hash the file contents and check new hash versus old hash before proceeded with reacting to a file change notification. DAB should integrate those suggestions: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-8.0#simple-startup-change-token

seantleonard:
## Why make this change?

- Closes #2423
  - Hot reloading reacts to >1 file change notification representing a single "Save File" action. DAB should only react to the first file change event which represents processing the changed file. DAB should only react to subsequent file change events when the file has actually changed.

## What is this change?

- Modifies `ConfigFileWatcher` class such that it no longer needs to have `FileSystemRuntimeConfigLoader` as a dependency/constructor parameter.  `ConfigFileWatcher` just called the passed in loader's methods which was a confusing layer of abstraction. 
- ConfigFileWatcher now (due to this PR's changes) has a hash persistence mechanism of runtimeconfig file contents and compares the currently tracked hash to a new hash calculated on the file after a "file change" event is emitted from the file watcher. One file change (e.g. clicking save in a text editor) causes multiple file change events to be detected by the file watcher (Added code comments to clarify). Dotnet docs describe the hashing mechanism included in this PR as a way to "ignore" subsequent file change events and only act on one event per file change.
    - Hash mechanism is SHA256 (32 bytes) as recommended by this [.NET documentation guidance](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha1?view=net-8.0) That's why my implementation of the hash mechanism differs from the [ChangeToken documentation referenced implementation ](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-8.0#:~:text=exponential%20back%2Doff.-,Utilities/Utilities.cs%3A,-C%23) which uses SHA1. We are **not** using the hash for encryption.
    > Due to collision problems with SHA1, Microsoft recommends a security model based on SHA256 or better.
- ConfigFileWatcher now raises an event (**NewFileContentsDetected**) when the hash of the old versus new runtime config file is detected.  `FileSystemRuntimeConfigLoader` subscribes to `NewFileContentsDetected` and 
reacts by calling `HotReloadConfig(...)`
- Fixes how `FileSystemRuntimeConfigLoader` reads the config file to avoid issues where the disposable doesn't immediately close the file handle and we get occurrences of "IOException file being used by another process". Now using File.ReadAllText so that the file handle is closed immediately.
- Adds field `_isCliLoader` to FIleSystemRuntimeConfigLoader so DAB knows not to init filewatcher for CLI usage. This fixed a pipeline file IO bug due to CLI usage of filesystemruntimeconfigloader initializing the file watcher. 

## How was this tested?

The following two integration tests are added to validate the functionality of the modified `ConfigFileWatcher` class. These tests immitate a user change the runtime config file's contents (or not changing contents) and the test check that the `NewFileContentsDetected` is raised the expected number times. 

- [x] Integration Tests 
    1. `ConfigFileWatcher_NotifiedOfOneNetNewChanges`
    2. `ConfigFileWatcher_NotifiedOfZeroNetNewChange`

### DIFF
--- a/src/Config/ConfigFileWatcher.cs
+++ b/src/Config/ConfigFileWatcher.cs
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.IO.Abstractions;
+using Azure.DataApiBuilder.Config.Utilities;
+
 namespace Azure.DataApiBuilder.Config;
 
 /// <summary>
@@ -12,7 +15,7 @@ public class ConfigFileWatcher
 {
     private FileSystemWatcher? _fileWatcher;
     private FileSystemRuntimeConfigLoader _configLoader;
-
+    private byte[] _runtimeConfigHash;
     public ConfigFileWatcher(FileSystemRuntimeConfigLoader configLoader, string directoryName, string configFileName)
     {
         _fileWatcher = new FileSystemWatcher(directoryName)
@@ -20,6 +23,7 @@ public class ConfigFileWatcher
             Filter = configFileName,
             EnableRaisingEvents = true
         };
+        _runtimeConfigHash = FileUtilities.ComputeHash(filePath: directoryName+"/"+configFileName);
 
         _configLoader = configLoader;
         _fileWatcher.Changed += OnConfigFileChange;
@@ -36,9 +40,15 @@ public class ConfigFileWatcher
     {
         try
         {
-            // When RuntimeConfig is null, we abort hot-reload.
-            if (_configLoader.RuntimeConfig is not null && _configLoader.RuntimeConfig.IsDevelopmentMode())
+            if (_fileWatcher is null || _configLoader.RuntimeConfig is null || !_configLoader.RuntimeConfig.IsDevelopmentMode())
             {
+                return;
+            }
+
+            byte[] updatedRuntimeConfigFileHash = FileUtilities.ComputeHash(filePath: _fileWatcher.Path + "/" + _fileWatcher.Filter);
+            if (!_runtimeConfigHash.SequenceEqual(updatedRuntimeConfigFileHash))
+            {
+                _runtimeConfigHash = updatedRuntimeConfigFileHash;
                 _configLoader.HotReloadConfig(_configLoader.RuntimeConfig.DefaultDataSourceName);
             }
         }
--- a/src/Config/ConfigFileWatcher.cs
+++ b/src/Config/ConfigFileWatcher.cs
@@ -57,7 +57,7 @@ public class ConfigFileWatcher
         _fileWatcher = fileWatcher;
         _fileWatcher.Path = WatchedDirectory;
         _fileWatcher.Changed += OnConfigFileChange;
-        _runtimeConfigHash = FileUtilities.ComputeHash(_fileWatcher.FileSystem, filePath: WatchedDirectory + "/" + WatchedFile);
+        _runtimeConfigHash = FileUtilities.ComputeHash(_fileWatcher.FileSystem, filePath: Path.Combine(WatchedDirectory, WatchedFile));
     }
 
     /// <summary>
--- a/src/Config/ConfigFileWatcher.cs
+++ b/src/Config/ConfigFileWatcher.cs
@@ -1,38 +1,71 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System.IO.Abstractions;
 using Azure.DataApiBuilder.Config.Utilities;
 
 namespace Azure.DataApiBuilder.Config;
 
 /// <summary>
 /// This class is responsible for monitoring the config file from the
-/// local file system, and if changes are detected, triggering the
-/// required logic to begin a hot reload scenario.
+/// local file system. This watcher maintains a file hash to only emit
+/// one event for each file change occurrence. Because .NET may raise >1
+/// event for 1 file change instance, this class swallows extraneous events.
+/// - "OnChanged is called when changes are made to the size, system attributes,
+/// last write time, last access time, or security permissions of a file or directory
+/// in the directory being monitored."
+/// - The file hashing mechanism is a solution suggested by .NET documentation
+/// to handle duplicate events evented for the same file change event.
 /// </summary>
+/// <seealso cref="https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.onchanged?view=net-8.0#remarks"/>
+/// <seealso cref="https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.notifyfilter?view=net-8.0"/>
+/// <seealso cref="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-8.0#:~:text=exponential%20back%2Doff.-,Utilities/Utilities.cs%3A,-C%23"/>
 public class ConfigFileWatcher
 {
+    /// <summary>
+    /// Watches a specific file for modifications and alerts
+    /// this class when a change is detected.
+    /// </summary>
     private FileSystemWatcher? _fileWatcher;
-    private FileSystemRuntimeConfigLoader _configLoader;
+
+    /// <summary>
+    /// Hash of runtime config file.
+    /// Used to determine whether hot-reload should proceed due
+    /// to detecting new runtime config file content.
+    /// </summary>
     private byte[] _runtimeConfigHash;
-    public ConfigFileWatcher(FileSystemRuntimeConfigLoader configLoader, string directoryName, string configFileName)
+
+    /// <summary>
+    /// Orchestrates sending NewFileContentsDetected events to all subscribers.
+    /// </summary>
+    public event EventHandler? NewFileContentsDetected;
+
+    public ConfigFileWatcher(string directoryName, string configFileName)
     {
         _fileWatcher = new FileSystemWatcher(directoryName)
         {
             Filter = configFileName,
             EnableRaisingEvents = true
         };
-        _runtimeConfigHash = FileUtilities.ComputeHash(filePath: directoryName+"/"+configFileName);
 
-        _configLoader = configLoader;
+        _runtimeConfigHash = FileUtilities.ComputeHash(filePath: directoryName + "/" + configFileName);
         _fileWatcher.Changed += OnConfigFileChange;
     }
 
     /// <summary>
+    /// Raises the NewFileContentsDetected event which signals to listeners
+    /// that a new configuration has been detected. This will only be raised
+    /// for the first file-change event detected for a single file change.
+    /// </summary>
+    protected virtual void OnNewFileContentsDetected()
+    {
+        NewFileContentsDetected?.Invoke(this, EventArgs.Empty);
+    }
+
+    /// <summary>
     /// When a change is detected in the Config file being watched this trigger
-    /// function is called and handles the hot reload logic when appropriate,
-    /// ie: in a local development scenario.
+    /// function is called. It will compare the hash of the file to the cached hash
+    /// value in order to discard duplicate notifications for a file which has
+    /// only changed once.
     /// </summary>
     /// <param name="sender"></param>
     /// <param name="e"></param>
@@ -40,19 +73,17 @@ public class ConfigFileWatcher
     {
         try
         {
-            if (_fileWatcher is null || _configLoader.RuntimeConfig is null || !_configLoader.RuntimeConfig.IsDevelopmentMode())
+            if (_fileWatcher is not null)
             {
-                return;
-            }
-
-            // Multiple file change notifications may be raised for a single file change.
-            // Use file hashes to ensure that HotReload operation is only executed when a net-new
-            // runtime config is detected.
-            byte[] updatedRuntimeConfigFileHash = FileUtilities.ComputeHash(filePath: _fileWatcher.Path + "/" + _fileWatcher.Filter);
-            if (!_runtimeConfigHash.SequenceEqual(updatedRuntimeConfigFileHash))
-            {
-                _runtimeConfigHash = updatedRuntimeConfigFileHash;
-                _configLoader.HotReloadConfig(_configLoader.RuntimeConfig.DefaultDataSourceName);
+                // Multiple file change notifications may be raised for a single file change.
+                // Use file hashes to ensure that HotReload operation is only executed when a net-new
+                // runtime config is detected.
+                byte[] updatedRuntimeConfigFileHash = FileUtilities.ComputeHash(filePath: _fileWatcher.Path + "/" + _fileWatcher.Filter);
+                if (!_runtimeConfigHash.SequenceEqual(updatedRuntimeConfigFileHash))
+                {
+                    _runtimeConfigHash = updatedRuntimeConfigFileHash;
+                    OnNewFileContentsDetected();
+                }
             }
         }
         catch (Exception ex)
--- a/src/Config/ConfigFileWatcher.cs
+++ b/src/Config/ConfigFileWatcher.cs
@@ -87,7 +87,7 @@ public class ConfigFileWatcher
                 // Multiple file change notifications may be raised for a single file change.
                 // Use file hashes to ensure that HotReload operation is only executed when a net-new
                 // runtime config is detected.
-                byte[] updatedRuntimeConfigFileHash = FileUtilities.ComputeHash(_fileWatcher.FileSystem, filePath: WatchedDirectory + "/" + WatchedFile);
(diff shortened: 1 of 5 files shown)


src/Config/ConfigFileWatcher.cs
```diff
@@ -36,9 +40,15 @@ private void OnConfigFileChange(object sender, FileSystemEventArgs e)
...
         try
         {
-            // When RuntimeConfig is null, we abort hot-reload.
-            if (_configLoader.RuntimeConfig is not null && _configLoader.RuntimeConfig.IsDevelopmentMode())
+            if (_fileWatcher is null || _configLoader.RuntimeConfig is null || !_configLoader.RuntimeConfig.IsDevelopmentMode())
+            {
+                return;
+            }
+
+            byte[] updatedRuntimeConfigFileHash = FileUtilities.ComputeHash(filePath: _fileWatcher.Path + "/" + _fileWatcher.Filter);
```
abhishekkumams, line 48: it will be good to add in the comments why we needed to computeHash

src/Config/Utilities/FileUtilities.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.DataApiBuilder.Config.Utilities
```
abhishekkumams, line 10: can be added some class/method summary

src/Config/Utilities/FileUtilities.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    internal class FileUtilities
+    {
+        public static byte[] ComputeHash(string filePath)
+        {
+            int runCount = 1;
```
abhishekkumams, line 16: what is runCount here?

src/Config/Utilities/FileUtilities.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            while (runCount < 4)
```
abhishekkumams, line 18: why 4?

src/Config/Utilities/FileUtilities.cs
```diff
@@ -0,0 +1,51 @@
...
+                    {
+                        throw;
+                    }
+
+                    Thread.Sleep(TimeSpan.FromSeconds(Math.Pow(2, runCount)));
+                    runCount++;
+                }
+            }
+
+            return new byte[20];
```
abhishekkumams, line 48: why 20? can we something else?

seantleonard: 20 is SHA1.HashSizeInBytes https://en.wikipedia.org/wiki/SHA-1

src/Config/Utilities/FileUtilities.cs
seantleonard: arbitrary. Adding comment.

src/Config/Utilities/FileUtilities.cs
seantleonard: Used for exponential backoff mechanism for when file access results in an IO exception. This code is taken from example provided by dotnet team in the Change Token docs. Added comments to reference this. https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-8.0#:~:text=exponential%20back%2Doff.-,Utilities/Utilities.cs%3A,-C%23

src/Config/Utilities/FileUtilities.cs
seantleonard: updated to sha256

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +161,109 @@ public void HotReloadConfigRestRuntimeOptions()
...
         }
+
+        /// <summary>
+        /// This test validates that overwriting a file with the different contents
+        /// results in ConfigFileWatcher's NewFileContentsDetected event being raised.
+        /// Internally, ConfigFileWatcher.NewFileContentsDetected is raised when the hash
+        /// of the file contents differ from what was previously detected. For this test,
+        /// the file contents differ MyValue -> MyChangedValue.
+        /// The config file name is specific to this test in order to avoid concurrently
+        /// running tests from stepping over eachother due to acquiring a handle(s) to the file.
```
Aniruddh25, line 172:
```suggestion
        /// running tests from stepping over each other due to acquiring a handle(s) to the file.
```

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +161,109 @@ public void HotReloadConfigRestRuntimeOptions()
...
+                catch (IOException ex)
+                {
+                    Console.WriteLine($"IO Exception, retrying due to {ex.Message}");
+                    if (runCount == 3)
+                    {
+                        throw;
+                    }
+
+                    // Constant backoff because we don't want this to hold up the CI/CD pipelines.
+                    // Wait time is arbitrary.
```
Aniruddh25, line 262: Arbitrary wait times make this test non-deterministic.. Is there any way to make this more deterministic? Perhaps, write again only after we are sure of not getting an open handle exception?

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +161,109 @@ public void HotReloadConfigRestRuntimeOptions()
...
+                // For testing, modification of fileChangeNotificationsRecieved
+                // should be atomic. 
+                Interlocked.Increment(ref fileChangeNotificationsReceived);
+            };
+
+            // Act - Write to the file multiple times with the same value to implicitly trigger
+            // multiple file change events for the ConfigFileWatcher's filesystem watcher.
+            ModifyConfigFile(configFileName: configName, configContent: "MyValue");
+
+            // Assert -> allow time for the file change events to be processed.
```
Aniruddh25, line 230: can we assert on more than 1 file change events detected instead of waiting for arbitrary amount of time? I wary of test getting flaky..

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
seantleonard: Found a way to utilize `TestableIO.System.IO.Abstractions` like we do elsewhere to mock the filesystem and filewatcher objects. Then I was able to manually induce the file-changed event instead of waiting for the event to be raised by the filesystem directly (which as you correctly point out, leads to possible flakiness due to depending on filesystem 'eventually' raising the file changed event). 
In a real hot-reload scenario, I am keeping the 'exponential' backoff in the utility which calculates file content hash because there may actually be IOExceptions when files are inaccessible due to other processes. However for testing, we want quick and determinate.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
seantleonard: Yep, good call. Added 2 invocations of the file-changed event to simulate how a real file system would notify our `ConfigFileWatcher` component. That way I'm directly demonstrating and asserting that `ConfigFileWatcher` only raises its `NewFileContentsDetected` event when a new file contents hash is determined, while validating that >1 filechange event was detected.

src/Config/Utilities/FileUtilities.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <seealso cref="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-8.0#:~:text=exponential%20back%2Doff.-,Utilities/Utilities.cs%3A,-C%23"/>
+    /// <exception cref="FileNotFoundException">https://learn.microsoft.com/en-us/dotnet/api/system.io.file.openread?view=net-8.0#exceptions</exception>
+    public static byte[] ComputeHash(IFileSystem fileSystem, string filePath)
+    {
+        // Exponential back-off retry mechanism.
+        int runCount = 1;
+
+        // Arbitrary limit of 3 attempts to load the file.
+        // Maximum 8 seconds (2^3) of wait time due to retries.
+        while (runCount < 4)
```
aaronburtle, line 34: nit: prefer to use a named variable like "runLimit" in place of magic number '4'

seantleonard: updated with:
```csharp
    /// <summary>
    /// Limit the number of retries when failures occur.
    /// </summary>
    private static readonly int _runLimit = 3;

    /// <summary>
    /// Base of the exponential retry back-off.
    /// -> base ^ runCount
    /// e.g. 2^1 , 2^2, 2^3, etc.
    /// </summary>
    private static readonly int _exponentialRetryBase = 2;
```

aaronburtle (approved):
Looks great!

src/Config/ConfigFileWatcher.cs
```diff
@@ -39,16 +40,24 @@ public class ConfigFileWatcher
...
+    public string WatchedFile { get; private set; }
+
+    public ConfigFileWatcher(IFileSystemWatcher fileWatcher, string directoryName, string configFileName)
+    {
+        WatchedDirectory = directoryName;
+        WatchedFile = configFileName;
+        _fileWatcher = fileWatcher;
+        _fileWatcher.Path = WatchedDirectory;
         _fileWatcher.Changed += OnConfigFileChange;
+        _runtimeConfigHash = FileUtilities.ComputeHash(_fileWatcher.FileSystem, filePath: WatchedDirectory + "/" + WatchedFile);
```
Aniruddh25, line 60: should we utilize any cross platform file utility helper (if dotnet provides one) to specify the path separator? For example, on linux it is / but on windows it would be `\` ?
Perhaps use Path.IO.combine to get the full path name

seantleonard: updated.

abhishekkumams (approved):

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +162,159 @@ public void HotReloadConfigRestRuntimeOptions()
...
+            Mock.Get(fileSystem).Setup(fs => fs.File.Exists(It.IsAny<string>())).Returns(true);
+
+            // Mock file system to return different byte arrays for each read which occurs in
+            // FileUtilities.ComputeHash() called by ConfigFileWatcher.
+            // The first read occurs during ConfigFileWatcher's initialization
+            // The second read occurs during the manual invocation of a "file changed" event.
+            Mock.Get(fileSystem).SetupSequence(fs => fs.File.ReadAllBytes(It.IsAny<string>()))
+                .Returns(Encoding.UTF8.GetBytes("FirstValue"))
+                .Returns(Encoding.UTF8.GetBytes("SecondValue"))
+                .Returns(Encoding.UTF8.GetBytes("SecondValue"));
```
Aniruddh25, line 201: IF we care only about 2 reads, why mock a 3rd read as well?

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +162,159 @@ public void HotReloadConfigRestRuntimeOptions()
...
+
+        private const string UNEXPECTED_INVOCATION_COUNT_ERR = "Unexpected number of invocations of the NewFileContentsDetected event.";
+        private const string FILECHANGE_EVENT_NOT_RAISED_ERR = "The file system's file-changed event was not raised.";
+
+        /// <summary>
+        /// This test validates that overwriting a file with the different contents
+        /// results in ConfigFileWatcher's NewFileContentsDetected event being raised.
+        /// Internally, ConfigFileWatcher.NewFileContentsDetected is raised when the hash
+        /// of the file contents differ from what was previously detected. For this test,
+        /// the file contents differ MyValue -> MyChangedValue.
```
Aniruddh25, line 176: is MyValue now FirstValue
and MyChangedValue = SecondValue as per the modified code? If yes, fix the comment to avoid confusion

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +162,159 @@ public void HotReloadConfigRestRuntimeOptions()
...
+        }
+
+        /// <summary>
+        /// This test validates that overwriting a file with the same contents does not
+        /// result in ConfigFileWatcher's NewFileContentsDetected event being raised.
+        /// Internally, ConfigFileWatcher.NewFileContentsDetected is raised when the hash
+        /// of the file contents differ from what was previously detected.
+        /// In this test, the file contents are the same: MyValue -> MyValue.
+        /// The config file name is specific to this test in order to avoid concurrently
+        /// running tests from stepping over eachother due to acquiring a handle(s) to the file.
```
Aniruddh25, line 255:
```suggestion
        /// running tests from stepping over each other due to acquiring a handle(s) to the file.
```

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -159,5 +162,159 @@ public void HotReloadConfigRestRuntimeOptions()
...
+                // should be atomic. 
+                Interlocked.Increment(ref fileChangeNotificationsReceived);
+            };
+
+            // Track the number of invocations so far so we can prove that
+            // fileSystemWatcherMock.Raise() actually triggers a "file changed" event.
+            int numberFileSystemWatcherMockInvocations = fileSystemWatcherMock.Invocations.Count;
+
+            // Act
+            // Manually induce two "file changed" events to simulate a file change
```
Aniruddh25, line 295:
```suggestion
            // Manually induce a "file changed" event to simulate a file change
```

Aniruddh25 (approved):
Looks good, left some nits and questions.

src/Config/Utilities/FileUtilities.cs
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    throw;
+                }
+
+                Thread.Sleep(TimeSpan.FromSeconds(Math.Pow(_exponentialRetryBase, runCount)));
+                runCount++;
+            }
+        }
+
+#if NET8_0_OR_GREATER
```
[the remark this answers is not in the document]
seantleonard: I'll add this as part of 'removing .net6' PR as this will go in first

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
seantleonard: We care about two reads as a result of the hot reload.

First Read: original configuration loading.
Second Read: First simulation of file system FileChangedEvent
Third Read: Result of 2nd simulation of file system FileChangedEvent because a real filesystem returns >1 filechanged event. This test validates that the second return of "SecondValue" is a no-op because we've already reacted to seeing "SecondValue"

src/Config/ConfigFileWatcher.cs
```diff
@@ -1,45 +1,98 @@
...
         {
-            // When RuntimeConfig is null, we abort hot-reload.
-            if (_configLoader.RuntimeConfig is not null && _configLoader.RuntimeConfig.IsDevelopmentMode())
+            if (_fileWatcher is not null)
             {
-                _configLoader.HotReloadConfig(_configLoader.RuntimeConfig.DefaultDataSourceName);
+                // Multiple file change notifications may be raised for a single file change.
+                // Use file hashes to ensure that HotReload operation is only executed when a net-new
+                // runtime config is detected.
+                byte[] updatedRuntimeConfigFileHash = FileUtilities.ComputeHash(_fileWatcher.FileSystem, filePath: WatchedDirectory + "/" + WatchedFile);
```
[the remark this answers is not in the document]
seantleonard: thanks! updated

src/Config/ConfigFileWatcher.cs
```diff
@@ -1,45 +1,98 @@
...
+    /// this class when a change is detected.
+    /// </summary>
+    private IFileSystemWatcher? _fileWatcher;
 
-    public ConfigFileWatcher(FileSystemRuntimeConfigLoader configLoader, string directoryName, string configFileName)
-    {
-        _fileWatcher = new FileSystemWatcher(directoryName)
-        {
-            Filter = configFileName,
-            EnableRaisingEvents = true
```
seantleonard, line 21: I should not have removed this.

seantleonard: fixing in #2425

PR #2417, opened 2024-10-11
Add Support for Hot Reload of DataSource Section of Config file.

abhishekkumams:
## Why make this change?

- Closes #2323 

  - **Improved Developer Experience:** Seamless updates allow developers to update the database type or connection string without restarting the application.
  - **Enhanced Flexibility:** The application can dynamically switch between different databases based on current needs or environment such as database migration.
  - **Increased Productivity:** Reduced downtime by eliminating the need to restart the application for data-source changes.

## What is this change?
Supporting hot-reload for the data-source section involved two major parts:

### Part 1: Refreshing Dependencies (PR:#2391)
The first part focused on supporting the refreshing of various dependencies that rely on the DataSource section during startup. However, these dependencies were not being refreshed upon changes to the data-source configuration. This task involved adding the logic to refresh those components that are configured on startup that acts as a pre-requirement for Data-Source reload.

This critical part of the full support for hot-reloading was completed in PR: #2391 by Aaron. This PR ensured that all necessary components are refreshed when the data-source configuration changes.

### Part 2: Running Initialization with Updated Components/Dependencies
The second part involved running the initialization process once all dependencies are updated. This initialization is crucial as it uses the new data-source configuration to set up the connection and entity refresh.

Key Changes:

Ensured that the initialization process is triggered after all dependencies are refreshed and the same DataSourceName is maintained.
Added the initialization logic to refresh the new data-source configuration.

## How was this tested?

- [x] Manual Tests since pipelines do not support switching between databases.

#### Scenario A: Handling Invalid and Valid Passwords
Objective: To verify the system's behavior when using invalid and valid passwords in the connection string.

Steps:
1. Start with MSSQL with a valid string.
2. update it to an incorrect credential in the connection string.
3. Show the logs to capture error messages.
4. Run the query and observe the failure.
5. Update the connection string with a valid credential.
6. Run the query again and observe the success.

Results:
Logs clearly indicated an invalid password error when using the incorrect password.
The query failed with an invalid password for DB.
The query succeeded with a valid password for DB. 

#### Scenario B: Switching Between Databases
Objective: To ensure seamless transitions and query execution across different databases.

Steps:

1. Start with MSSQL.
2. Run a query and observe the results.
3. Update the configuration to use MySQL.
4. Run the same query and observe the results.
5. Update the configuration to use PostgreSQL.
6. Run the same query and observe the results.

Observed Results:
The query executed successfully in MSSQL.
The query executed successfully in MySQL after updating the configuration.
[...]

PR #2425, opened 2024-10-14
Update AuthorizationResolver to be hot-reload aware

### LINKED ISSUE #2324: Hot-Reload Authorization rules
While we currently deserialize the new entities into our object mapping that makes up the runtimeconfig, we do not update the GraphQL Schema. For entities to properly hot-reload, we need the GraphQL Schema to update as new entities are added or old entities are removed.

We also do not currently refresh the AuthorizationResolver. This too needs to be refreshed to reflect the new authorization configuration as it is in the newly hot-reloaded entity section.

seantleonard:
## Why make this change?

- Closes #2324 which tracks making DAB's AuthorizationResolver hot-reload aware. Hot-reload aware just means that the service should update its internal state to align with the latest runtimeconfig changes.

## What is this change?

- Adds Event for DAB's HotReload event handler to signal changes to the runtime config. The AuthorizationResolver can only refresh once the SqlMetadataProvider has fetched updated database object metadata.
- GraphQL is not yet supported (separate work item for schema eviction). 
- REST should work when #2417 is merged which fixes bug with updating metadata providers after hot reload.

## How was this tested?
- [x] Integration Tests -> setup test fixture with initial config specifying a Book entity with certain permissions. The test then hot-reloads a new config with a different entity and different permissions (different include/exclude column names) and validates that only the new entity is accessible and the expected fields are accessible. Validates that the old entity is not accessible and has "been forgotten" by DAB.


src/Config/DabConfigEvents.cs
```diff
@@ -14,4 +14,5 @@ public class DabConfigEvents
     public const string MYSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED = "MYSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED";
     public const string POSTGRESQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED = "POSTGRESQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED";
     public const string DOCUMENTOR_ON_CONFIG_CHANGED = "DOCUMENTOR_ON_CONFIG_CHANGED";
+    public const string AUTHZ_RESOLVER_ON_CONFIG_CHANGED = "AUTHZ_RESOLVER_ON_CONFIG_CHANGED";
 }
```
Aniruddh25, line 18: Does this comment in the description still apply? 

`"However, there may be issue with metadataprovider hot-reload not updating all internal state properly."`

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -739,23 +752,6 @@ public IEnumerable<string> GetRolesForEntity(string entityName)
         return EntityPermissionsMap[entityName].RoleToOperationMap.Keys;
     }
 
-    /// <summary>
```
Aniruddh25, line 742: why remove this piece of code?

src/Core/Authorization/AuthorizationResolver.cs
```diff
@@ -39,10 +40,13 @@ public class AuthorizationResolver : IAuthorizationResolver
 
     public AuthorizationResolver(
         RuntimeConfigProvider runtimeConfigProvider,
-        IMetadataProviderFactory metadataProviderFactory
-        )
+        IMetadataProviderFactory metadataProviderFactory,
+        HotReloadEventHandler<HotReloadEventArgs>? handler = null)
```
Aniruddh25, line 44: remind me is this handler filled in by Dependency injection?

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: unused with Visual Studio denoting "0 references"

src/Config/DabConfigEvents.cs
seantleonard: no, the pr that was pending merge when that comment was written was merged. Updated PR description.

Aniruddh25 (approved):
Looks good, specially the TestFixture. Left a few questions. Thanks for pushing out these changes so quickly.

src/Core/Authorization/AuthorizationResolver.cs
seantleonard: Yes. We register that in startup: 
https://github.com/Azure/data-api-builder/blob/c9aa87b0257507b1ea223953cfc8b1c76928610b/src/Service/Startup.cs#L81-L82

`HotReloadEventHandler<HotReloadEventArgs>?` is nullable to avoid cascading changes within the 7 other test files that don't mock the hotreloadeventhandler. When nullable, `hander?.Subscribe()` is a no-op. The tests included in this PR add coverage for supplying a hotreloadeventhandler for hot-reloading.

src/Service.Tests/Configuration/HotReload/AuthorizationResolverHotReloadTests.cs
```diff
@@ -0,0 +1,215 @@
...
+        Assert.AreEqual(
+            expected: HttpStatusCode.OK,
+            actual: initialResponse.StatusCode,
+            message: "Initial configuration bootstrap failed.");
+    }
+
+    [ClassCleanup]
+    public static void ClassCleanup()
+    {
+        _testServer.Dispose();
```
aaronburtle, line 212: Can delete the config file that was only created for these tests in cleanup.

abhishekkumams (approved):
LGTM

src/Service.Tests/Configuration/HotReload/AuthorizationResolverHotReloadTests.cs
seantleonard: updated

aaronburtle (approved):

Issue #2427, opened 2024-10-15
[Bug]: Duplicated Query Exector Initialization during hot reload


aaronburtle:
### What happened?

The Query Manager Factory reload is already settings up the query executors, and so by signaling a refresh on those executors as part of the hot reload duplicates that process and creates errors.1.3

### Version

1.3

### What database are you using?

Azure SQL

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #2426, opened 2024-10-15
[Bug]: SqlMetadataProviders not being properly initialized during hot-reload


aaronburtle:
### What happened?

When the SqlMetadataProviderFactories are initialized on startup, they also call into an initialization function which properly configures the providers that they hold. During a hot reload, we also need to do this initialization.

### Version

1.3

### What database are you using?

Azure SQL

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2428, opened 2024-10-15
Fix MetadataProvider initialization in Hot Reload and remove duplicate Query Executor reloading

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2426
Closes https://github.com/Azure/data-api-builder/issues/2427

## What is this change?

Properly create and initialize a new set of `SqlMetadataProviders` as part of the hot-reload process of the `SqlMetadataProviderFactory`, by utilizing the `initiaizeAsync()` function. This guarantees a correctly configured set of meta data providers for the factory.

Remove the event handling to hot-reload the `QueryExecutors` classes specifically, since this is already handled during the hot-reload process of the `QueryManagerFactory`. That process creates new and up to date `QueryExecutors` as part of a hot-reload.

With the correct initialization of the meta data providers, we can remove the code that saves the `DefaultDataSourceName` to be passed along the hot-reloaded configs. With all of the data structures we initialized on new meta data providers, we do not need to save the data source name and hot-reloading will be more resilient. This is because we will build all of the data structures which depend on this DefaultDataSourceName fresh from newly created or cleared objects, and we do not have to worry about maintaining previous configuration's DefaultDataSourceName, everything will simply use the new name generated during a hot-reload.

### Testing Scenario
Scenario A: Handling a change in entities in the config 
Objective: Verify that hot-reload correctly refreshes the MetadataProviders and QueryExecutors when we hot-reload the entities by making a REST request.

Steps:

Start with MSSQL with a valid config file that contains some valid entities.
Make a REST request that targets one of the entities and observe the results to compare against later.
Change the entities section by removing an entity and any entities that have a relationship with that entity.
Make a REST request for that entity and observe that the entity is not found.
Add the removed entities back to the config file.
Make a REST request for that same entity and observe the results match the original REST request.

## Sample Request(s)

For the "Book" entity with a rest path of "rest"

https://localhost:5001/rest/Book

### DIFF
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -142,8 +142,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
         string path,
         [NotNullWhen(true)] out RuntimeConfig? config,
         bool replaceEnvVar = false,
-        ILogger? logger = null,
-        string defaultDataSourceName = "")
+        ILogger? logger = null)
     {
         if (_fileSystem.File.Exists(path))
         {
@@ -185,9 +184,9 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     /// <param name="replaceEnvVar">Whether to replace environment variable with its
     /// value or not while deserializing.</param>
     /// <returns>True if the config was loaded, otherwise false.</returns>
-    public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false, string defaultDataSourceName = "")
+    public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false)
     {
-        return TryLoadConfig(ConfigFilePath, out config, replaceEnvVar, defaultDataSourceName: defaultDataSourceName);
+        return TryLoadConfig(ConfigFilePath, out config, replaceEnvVar);
     }
 
     /// <summary>
@@ -197,7 +196,7 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
     public void HotReloadConfig(string defaultDataSourceName, ILogger? logger = null)
     {
         logger?.LogInformation(message: "Starting hot-reload process for config: {ConfigFilePath}", ConfigFilePath);
-        TryLoadConfig(ConfigFilePath, out _, replaceEnvVar: true, defaultDataSourceName: defaultDataSourceName);
+        TryLoadConfig(ConfigFilePath, out _, replaceEnvVar: true);
         SendEventNotification();
     }
 
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -156,11 +156,6 @@ public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
                     logger?.LogInformation("Monitoring config: {ConfigFilePath} for hot-reloading.", ConfigFilePath);
                 }
 
-                if (!string.IsNullOrEmpty(defaultDataSourceName))
-                {
-                    RuntimeConfig.DefaultDataSourceName = defaultDataSourceName;
-                }
-
                 config = RuntimeConfig;
                 return true;
             }
--- a/src/Config/HotReloadEventHandler.cs
+++ b/src/Config/HotReloadEventHandler.cs
@@ -22,6 +22,9 @@ public class HotReloadEventHandler<TEventArgs> where TEventArgs : HotReloadEvent
     {
         _eventHandlers = new Dictionary<string, EventHandler<TEventArgs>?>
         {
+            // QueryManagerFactory will create a new set of QueryExecutors
+            // during Hot-Reload so it is not necessary to specifically reconfigure
+            // QueryExecutors as part of the Hot-Reload process.
             { QUERY_MANAGER_FACTORY_ON_CONFIG_CHANGED, null },
             { METADATA_PROVIDER_FACTORY_ON_CONFIG_CHANGED, null },
             { QUERY_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
--- a/src/Config/HotReloadEventHandler.cs
+++ b/src/Config/HotReloadEventHandler.cs
@@ -27,8 +27,8 @@ public class HotReloadEventHandler<TEventArgs> where TEventArgs : HotReloadEvent
             // QueryExecutors as part of the Hot-Reload process.
             { QUERY_MANAGER_FACTORY_ON_CONFIG_CHANGED, null },
             { METADATA_PROVIDER_FACTORY_ON_CONFIG_CHANGED, null },
-            { QUERY_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
-            { MUTATION_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
+            { QUERY_ENGINE_FACTORY_ON_CONFIG_CHANGED, null },
+            { MUTATION_ENGINE_FACTORY_ON_CONFIG_CHANGED, null },
             { DOCUMENTOR_ON_CONFIG_CHANGED, null }
         };
     }
--- a/src/Config/HotReloadEventHandler.cs
+++ b/src/Config/HotReloadEventHandler.cs
@@ -26,10 +26,6 @@ public class HotReloadEventHandler<TEventArgs> where TEventArgs : HotReloadEvent
             { METADATA_PROVIDER_FACTORY_ON_CONFIG_CHANGED, null },
             { QUERY_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
             { MUTATION_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
-            { QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
-            { MSSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
-            { MYSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
-            { POSTGRESQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
             { DOCUMENTOR_ON_CONFIG_CHANGED, null }
         };
     }
--- a/src/Config/RuntimeConfigLoader.cs
+++ b/src/Config/RuntimeConfigLoader.cs
@@ -93,7 +93,6 @@ public abstract class RuntimeConfigLoader
         ILogger? logger = null,
         string? connectionString = null,
         bool replaceEnvVar = false,
-        string dataSourceName = "",
         Dictionary<string, string>? datasourceNameToConnectionString = null,
         EnvironmentVariableReplacementFailureMode replacementFailureMode = EnvironmentVariableReplacementFailureMode.Throw)
     {
@@ -109,7 +108,7 @@ public abstract class RuntimeConfigLoader
             }
 
             // retreive current connection string from config
-            string updatedConnectionString = config.DataSource.ConnectionString;
+            string updatedConnectionString = config.DataSource.ConnectionString;            
 
             if (!string.IsNullOrEmpty(connectionString))
             {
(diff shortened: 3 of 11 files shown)


src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
```diff
@@ -62,7 +62,10 @@ private void ConfigureMetadataProviders()
 
         public void OnConfigChanged(object? sender, HotReloadEventArgs args)
         {
+            _metadataProviders = new Dictionary<string, ISqlMetadataProvider>();
             ConfigureMetadataProviders();
+            // Blocks the current thread until initialization is finished.
+            this.InitializeAsync().GetAwaiter().GetResult();
```
seantleonard, line 68: #2417 handles this in a different way. Can you and @abhishekkumams identify which method you want to ultimately use?

src/Config/FileSystemRuntimeConfigLoader.cs
seantleonard, line 1: Please add more detail in pr description for how you manually tested these changes. Use
#2417 and #2414 as examples via the "test scenarios" sections

src/Config/RuntimeConfigLoader.cs
seantleonard, line 1: can you add more details to the PR description for why we need to change how `dataSourceName` is used: using the passed in argument versus using the "default" data source name.

src/Config/HotReloadEventHandler.cs
```diff
@@ -26,10 +26,6 @@ public HotReloadEventHandler()
             { METADATA_PROVIDER_FACTORY_ON_CONFIG_CHANGED, null },
             { QUERY_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
             { MUTATION_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
-            { QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
-            { MSSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
-            { MYSQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
-            { POSTGRESQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED, null },
```
seantleonard, line 32: Can you add comment surrounding `QUERY_MANAGER_FACTORY_ON_CONFIG_CHANGED` here that describes how the "query executors since those are already being correctly refreshed already by the query manager factory reload process."

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
aaronburtle: they are the same, but this way will give us the internal exceptions that come from initializeAsync, instead of wrapping them in an aggregate exception.

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
```diff
@@ -17,7 +17,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders
     /// <inheritdoc />
     public class MetadataProviderFactory : IMetadataProviderFactory
     {
-        private readonly IDictionary<string, ISqlMetadataProvider> _metadataProviders;
+        private IDictionary<string, ISqlMetadataProvider> _metadataProviders;
```
abhishekkumams, line 20: just for new allocation, we shouldn't change its access modifier

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
```diff
@@ -62,7 +62,10 @@ private void ConfigureMetadataProviders()
 
         public void OnConfigChanged(object? sender, HotReloadEventArgs args)
         {
+            _metadataProviders = new Dictionary<string, ISqlMetadataProvider>();
```
abhishekkumams, line 65: better to clear() because it avoids re-allocating the _metadataProviders dictionary every time the OnConfigChanged method is called. it is more efficient in terms of memory and performance.

I'll address it in my PR. so you can remove revert this

src/Config/RuntimeConfigLoader.cs
```diff
@@ -97,7 +93,6 @@ public static bool TryParseConfig(string json,
         ILogger? logger = null,
         string? connectionString = null,
         bool replaceEnvVar = false,
-        string dataSourceName = "",
```
sezal98, line 100: `string dataSourceName = "",`
If this has been removed, we should modify the doc string

src/Config/RuntimeConfigLoader.cs
```diff
@@ -133,7 +122,7 @@ public static bool TryParseConfig(string json,
             }
```
sezal98, line 122: We should put `Dictionary<string, string>? datasourceNameToConnectionString =  new Dictionary<string, string>();` in the function defination itself instead of putting null there and then checking and replacing it with new object here.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -156,11 +156,6 @@ public bool TryLoadConfig(
                     logger?.LogInformation("Monitoring config: {ConfigFilePath} for hot-reloading.", ConfigFilePath);
                 }
 
-                if (!string.IsNullOrEmpty(defaultDataSourceName))
```
Aniruddh25, line 159: if we can remove assigning of defaultDataSourceName, the function definition should remove that parameter since its no longer used. 

And all callers of this function should remove passing this argument along.

Aniruddh25 (changes requested):
Need to fix remove references to defaultDataSourceName

aaronburtle (commented):
.

src/Config/HotReloadEventHandler.cs
```diff
@@ -22,14 +22,13 @@ public HotReloadEventHandler()
     {
         _eventHandlers = new Dictionary<string, EventHandler<TEventArgs>?>
         {
+            // QueryManagerFactory will create a new set of QueryExecutors
+            // during Hot-Reload so it is not necessary to specifically reconfigure
+            // QueryExecutors as part of the Hot-Reload process.
```
seantleonard, line 27: below this comment, ensure that there is consistent spacing between string constants and null. e.g. the following are inconsistent with the rest:
```csharp
{ QUERY_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
{ MUTATION_ENGINE_FACTORY_ON_CONFIG_CHANGED,null },
```

src/Config/RuntimeConfigLoader.cs
```diff
@@ -120,16 +115,12 @@ protected void SignalConfigChanged(string message = "")
...
-    /// <param name="dataSourceName"> datasource name for which to add connection string</param>
-    /// <param name="datasourceNameToConnectionString"> dictionary of datasource name to connection string</param>
     /// <param name="replacementFailureMode">Determines failure mode for env variable replacement.</param>
     public static bool TryParseConfig(string json,
         [NotNullWhen(true)] out RuntimeConfig? config,
         ILogger? logger = null,
         string? connectionString = null,
         bool replaceEnvVar = false,
-        string dataSourceName = "",
-        Dictionary<string, string>? datasourceNameToConnectionString = null,
```
seantleonard, line 132: I wonder if this is being used by GraphQL feature that @rohkhann is building.

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
```diff
@@ -64,7 +64,8 @@ public void OnConfigChanged(object? sender, HotReloadEventArgs args)
         {
             _metadataProviders.Clear();
             ConfigureMetadataProviders();
-            InitializeAsync().Wait();
+            // Blocks the current thread until initialization is finished.
+            this.InitializeAsync().GetAwaiter().GetResult();
```
seantleonard, line 68: Now that Abhishek's PR is in, can we use what he merged? 
`GetAwaiter().GetResult();` should only be used by the compiler. See: https://www.nikouusitalo.com/blog/why-is-getawaiter-getresult-bad-in-c/

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
seantleonard, line 1: looks like there are build errors in this file.

seantleonard: /workspaces/data-api-builder/src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs(95,37): error CS1501: No overload for method 'TryLoadKnownConfig' takes 3 arguments [/workspaces/data-api-builder/src/Service.Tests/Azure.DataApiBuilder.Service.Tests.csproj::TargetFramework=net6.0]
/workspaces/data-api-builder/src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs(164,37): error CS1501: No overload for method 'TryLoadKnownConfig' takes 3 arguments [/workspaces/data-api-builder/src/Service.Tests/Azure.DataApiBuilder.Service.Tests.csproj::TargetFramework=net6.0]
/workspaces/data-api-builder/src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs(263,37): error CS1501: No overload for method 'TryLoadKnownConfig' takes 3 arguments [/workspaces/data-api-builder/src/Service.Tests/Azure.DataApiBuilder.Service.Tests.csproj::TargetFramework=net6.0]

seantleonard (changes requested):
Worried that stripping out tracking of datasourcename -> connection string and other tracking of datasourcename will break @rohkhann 's GraphQL usage.

src/Core/Services/MetadataProviders/MetadataProviderFactory.cs
aaronburtle: Wait() has the same problem according to that article except it also wraps any exceptions with its own exception handling.

src/Service.Tests/Unittests/MultiSourceQueryExecutionUnitTests.cs
aaronburtle: The most recent changes should have fixed these. They were mocks not matching the new method signature.

src/Config/RuntimeConfigLoader.cs
aaronburtle: I removed because right now we dont have any calls going into this function with a dictionary, but if he utilizes this, there's no problem with it being added back. But for now this is totally unused in the code.

src/Config/HotReloadEventHandler.cs
aaronburtle: i swear sometimes VS just randomly changes my formatting when I copy paste things. Fixed.

src/Config/RuntimeConfigLoader.cs
seantleonard: @aaronburtle, please reach out to @rohkhann to determine whether to add this code back.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -120,16 +115,12 @@ protected void SignalConfigChanged(string message = "")
...
     /// value or not while deserializing. By default, no replacement happens.</param>
-    /// <param name="dataSourceName"> datasource name for which to add connection string</param>
-    /// <param name="datasourceNameToConnectionString"> dictionary of datasource name to connection string</param>
     /// <param name="replacementFailureMode">Determines failure mode for env variable replacement.</param>
     public static bool TryParseConfig(string json,
         [NotNullWhen(true)] out RuntimeConfig? config,
         ILogger? logger = null,
         string? connectionString = null,
         bool replaceEnvVar = false,
-        string dataSourceName = "",
```
rohkhann, line 131: This field was earlier added to be used as a fallback in case stuff broke with multi-source. should be fine to remove.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -100,9 +96,8 @@ protected void SignalConfigChanged(string message = "")
     /// <param name="config">The loaded <c>RuntimeConfig</c>, or null if none was loaded.</param>
     /// <param name="replaceEnvVar">Whether to replace environment variable with its
     /// value or not while deserializing.</param>
-    /// <param name="dataSourceName">The data source name to be used in the loaded config.</param>
     /// <returns>True if the config was loaded, otherwise false.</returns>
-    public abstract bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false, string dataSourceName = "");
+    public abstract bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false);
```
rohkhann, line 100: This would cause a breakage on our end when we do an update as we override this function. but I believe it should be a minor fix that we should be able to make if this is now unused.

src/Config/RuntimeConfigLoader.cs
rohkhann: Should be good. these fields were earlier added as fallback mechanism's when multi-source db was being introduced. Should be good to remove now.

aaronburtle: @rohkhann confirmed they are not using this, should be fine to remove.

Aniruddh25 (approved):
LGTM. Thanks for these changes, looking forward to end-2-end tests

seantleonard (approved):

PR #2432, opened 2024-10-19
Drop .NET 6 Support

seantleonard:
## Why make this change?

- .NET 6 is going out of support in a few weeks. DAB doesn't need to maintain support as it's published binaries are self-contained, meaning that dotnet runtime is included.

## What is this change?

- Removes .NET6 references from .csproj files for target framework.
- Removes .NET6 specific code in authN handlers
- Removes .NET6 sdk install in pipeline config
- Removes .NET6 dependency section from `src/Directory.Packages.props`
- CI/CD pipelines now complete quicker, as removing .NET6 shaves off about ~12-15minutes per test run. (Config file creation + integration tests)
- updated links as well.
- Removed the build-pipeline file from here, as it is not used.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests


sezal98:
Should we make changes in the scripts as well? FileName : scripts.publish.ps1
`$dotnetTargetFrameworks = "net6.0", "net8.0"`

src/Service.Tests/Unittests/SerializationDeserializationTests.cs
```diff
@@ -281,7 +281,6 @@ private void InitializeObjects()
         {
             _options = new()
             {
-#if NET8_0_OR_GREATER
                 // ObjectConverter behavior different in .NET8 most likely due to
                 // .NET7 breaking change:
                 // - https://learn.microsoft.com/dotnet/core/compatibility/serialization/7.0/polymorphic-serialization#affected-apis
```
sezal98, line 286: We should modify the comments according to the new changes

src/Core/AuthenticationHelpers/AuthenticationSimulator/SimulatorAuthenticationHandler.cs
```diff
@@ -42,25 +41,6 @@ public SimulatorAuthenticationHandler(
         : base(options, logger, encoder)
```
sezal98, line 41: Can be removed ?
```
// ISystemClock is obsolete in .NET 8.0 and later
// https://learn.microsoft.com/dotnet/core/compatibility/aspnet-core/8.0/isystemclock-obsolete
```

sezal98 (approved):

Aniruddh25 (commented):
Good initiative, we should check with @neeraj-sharma2592 with respect to dependencies on hosting before merging

abhishekkumams:
> Good initiative, we should check with @neeraj-sharma2592 with respect to dependencies on hosting before merging

I'm in touch with [Neeraj](https://github.com/neeraj-sharma2592) for the same. Will add update accordingly.

sezal98:
image This query will show files which still contain dotnet 6.0 reference. 
There are some files with comments pointing to 6.0 related classes. Please check if these can be updated to 8.0

abhishekkumams:
> image This query will show files which still contain dotnet 6.0 reference. There are some files with comments pointing to 6.0 related classes. Please check if these can be updated to 8.0

Thanks, it was pointing to urls, updated and verified.

src/Service.Tests/SqlTests/SqlTestHelper.cs
```diff
@@ -354,7 +354,7 @@ public static SqlException CreateSqlException(int number, string message = "")
 
             // Create SqlError object.
             // For details on what the parameters stand for please refer:
-            // https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlerror.number?view=dotnet-plat-ext-6.0#examples
+            // https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlerror.number?view=dotnet-plat-ext-8.0#examples
```
Aniruddh25, line 357: for these doc links, is there a non-version specific link, so that next time we need to update - this remains a no-op?

Aniruddh25 (approved):
Overall looks good, few comments to see if we can reduce the load next time we update.

src/Service.Tests/SqlTests/SqlTestHelper.cs
aaronburtle: if we remove the `?view=dotnet-plat-ext-8.0` it will use the default page which ought to be the latest version

`https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlerror.number#examples`

ill add this in.

aaronburtle (approved):

Issue #2438, opened 2024-10-26
Add End to End testing for runtime section in Hot-Reload scenario


aaronburtle:
This task covers the need for end to end testing to validate the hot-reloading of the runtime section of the config.

PR #2439, opened 2024-10-26
Add End to End test for runtime section in a Hot-Reload scenario

### LINKED ISSUE #2109: Update test in ConfigFileWatcherUnitTests.cs for Hot Reload
As a part of this pull request https://github.com/Azure/data-api-builder/pull/2108 we are turning off the unit tests for the hot reload feature while it is disabled. Before we can have the hot reload feature available we need to turn this test back on.

The test in question is within the `ConfigFileWatcherUnitTests.cs` file on line 73. This test needs to have its ignore removed so that it will be able to be picked up by our test suite.

image

aaronburtle:
## Why make this change?

Adds support for End to End tests within the ConfigurationTests class, along with tests for hot-reloading the runtime section of the config.

Closes https://github.com/Azure/data-api-builder/issues/2438
Closes #2109

## What is this change?

Add a function for generating a file to be used for starting the engine and then for hot-reloading. In other words, this function creates a valid string using arguments to replace portions of the string that it generates to represent the config file, it then writes that string as a file to the current working directory. If this file is what we use to start the engine, calling this function again will overwrite that file and initiate a hot-reload. We use this function then to overwrite the file with new runtime values and hot-reload, demonstrating that hot-reload will work for the runtime section.

Once the hot-reload has happened we can validate that the engine is able to correctly process requests.

## How was this tested?

Tests create a valid config file, start the engine with that file, assert that requests function correctly, modify that file by saving a new one with the same path triggering a hot-reload, run new requests and assert that they function correctly.

Specifically we test that new paths still result in valid results. And that changing the value of enabled from true to false results in the response changing from status OK to status NotFound.

## Sample Request(s)


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+                        id
+                        title
+                        publisher_id
+                    }
+                }
+            }";
+
+            GenerateConfigFile(connectionString: $"{GetConnectionStringFromEnvironmentConfig(TestCategory.MSSQL)}");
+
+            TestServer server = new(Program.CreateWebHostBuilder(new string[] { "--ConfigFileName", HOT_RELOAD_CONFIG_FILE_NAME }));
```
seantleonard, line 651: When we continue adding more and more hot reload tests, creating a new testserver to execute all dab startup code may be taxing in terms of total execution time spent setting up a test and checking one thing (especially in the pipelines).

Are there other ways to setup the hot reload test fixture such that we startup DAB with original config a single time and then each tests validates the hot reloaded values?

For example:
1. Test Class Setup (runs once for entire hotReloadTestClass): 
    - Load in original config with one entity and spot check that this original config is loaded (perhaps a single get request to check endpoint is wired up.)
    - Modify original config with your hot reloaded contents.
    - Now your test fixture is ready for each individual test.
2. Each test case validates that your new config had the intended effects:
    - Original REST/GraphQL endpoint -> Not Found
    - New REST/GraphQL endpoint -> found
    - (allows you to add more tests: Validate original AuthN/AuthZ config isn't honored and new config is)

Other benefits here are that your individual tests don't need to be involved in:
- new config creation with custom values. 
- hot reloading the config (this would be done in class setup, and tests just check expected hot reloaded state)

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+            string initialRestContents = await result.Content.ReadAsStringAsync();
+            JsonElement initialGQLContents = await GraphQLRequestExecutor.PostGraphQLRequestAsync(
+                httpClient,
+                configProvider,
+                graphQLQueryName,
+                graphQLQuery);
+
+            // Assert
+            Assert.IsTrue(SqlTestHelper.JsonStringsDeepEqual($"{restBookContents}", initialRestContents));
+            SqlTestHelper.PerformTestEqualJsonStrings(BOOK_DBO_CONTENTS, initialGQLContents.GetProperty("items").ToString());
```
seantleonard, line 666: why do we need to validate that all the book contents are returned for a test that's just checking the REST path? Would check HTTP200 response be sufficient here?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+        /// requests for the book database object returns the expected results.
+        /// Then hot-reload the config runtime paths both for rest and GQL, and
+        /// again assert that the requests return the expected results against
+        /// the book database object.
+        /// </summary>
+        /// <returns></returns>
+        [TestCategory(MSSQL_ENVIRONMENT)]
+        [TestMethod("Hot-reload runtime paths.")]
+        public async Task HotReloadConfigRuntimePathsEndToEndTest()
+        {
```
seantleonard, line 632: This summary differs from the test method name which describes what you're validating here. The test is checking that a hot reloaded config changes the GraphQL and REST paths, yes? Validating these changes would ensure that old path -> 404 and new path -> 200

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+                Content = JsonContent.Create(payload)
+            };
+
+            // Act
+            HttpResponseMessage restResult = await httpClient.GetAsync("/rest/Book");
+            HttpResponseMessage gQLResult = await httpClient.SendAsync(request);
+
+            // Assert
+            Assert.AreEqual(HttpStatusCode.OK, restResult.StatusCode);
+            Assert.AreEqual(HttpStatusCode.OK, gQLResult.StatusCode);
```
seantleonard, line 722: nit: Are these really act and assert on the changes you want to validate? These are technically the "Arrange" section setting up the test so that you can validate the side effects of the hot reload.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+            };
+            GenerateConfigFile(
+                connectionString: $"{GetConnectionStringFromEnvironmentConfig(TestCategory.MSSQL)}",
+                restEnabled: restEnabled,
+                gQLEnabled: gQLEnabled);
+            System.Threading.Thread.Sleep(1000);
+
+            // Act
+            restResult = await httpClient.GetAsync($"rest/Book");
+            gQLResult = await httpClient.SendAsync(request);
```
seantleonard, line 737: nit: The hot reload operation is the "Act" as it triggers the side effects you want to validate. The "Assert" section would include sending your requests to DAB's rest/graphql endpoints to ensure status is 404

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+                        publisher_id
+                    }
+                }
+            }";
+
+            GenerateConfigFile(connectionString: $"{GetConnectionStringFromEnvironmentConfig(TestCategory.MSSQL)}");
+
+            TestServer server = new(Program.CreateWebHostBuilder(new string[] { "--ConfigFileName", HOT_RELOAD_CONFIG_FILE_NAME }));    
+            HttpClient httpClient = server.CreateClient();
+            RuntimeConfigProvider configProvider = server.Services.GetService<RuntimeConfigProvider>();
```
seantleonard, line 653: while this was done in the past, it does seem odd that we need to acquire the runtimeconfigprovider from the testserver.Perhaps a modification to `GraphQLRequestExecutor.PostGraphQLRequestAsync(` would make these new tests cleaner:
1. Move most logic of `GraphQLRequestExecutor.PostGraphQLRequestAsync(` to another public static function except for using "runtimeconfigprovider" to get the graphql path, the new public static function would have the same method signature with exception to runtimeconfigprovider, instead of that, you'd just have `string graphQLPath`. then that would enable you to avoid dealing with test server internals. 
1. we wouldn't need to refactor other tests.. yet. They could move to improved api later.

seantleonard (changes requested):
Need to look at whether there's a more efficient way to set up the hot reload test fixture so that tests are succinct and test exactly what they say they're testing, and avoid needing to create  a new test server for each individual test.

src/Service.Tests/Configuration/ConfigurationTests.cs
seantleonard: having one test server created during TestClassSetup may not be appropriate for this particular set of tests. There will be a situation where that may help make testing more efficient. Keep this in mind for future tests and this feedback need not block this bit of coverage being added.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+                        ""mappings"": {
+                          """ + entityBackingColumn + @""": """ + entityExposedName + @"""
+                        }
+                      }
+                  }
+              }");
+
+        }
+
+            [TestCleanup]
```
Aniruddh25, line 611:
```suggestion
        [TestCleanup]
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+        }
+
+        /// <summary>
+        /// Start DAB with the pre-formed config file, assert that rest and GQL
+        /// requests for the book database object returns the expected results.
+        /// Then hot-reload the config runtime paths both for rest and GQL, and
+        /// again assert that the requests return the expected results against
+        /// the book database object.
+        /// </summary>
+        /// <returns></returns>
```
Aniruddh25, line 628:
```suggestion
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+        }
+
+        /// <summary>
+        /// Start DAB with pre-formed config file, assert that requests sent to
+        /// the rest and gql endpoints result in result status OK. Hot-reload
+        /// the config and set rest and graphql enabled to false, and assert
+        /// that requests sent to the rest and gql endpoints result in result
+        /// status NotFound.
+        /// </summary>
+        /// <returns></returns>
```
Aniruddh25, line 696:
```suggestion
```

Aniruddh25 (approved):
As noted in other comments, this test can be refactored for scaling to add other hot reload scenarios. Approving it for now with the assumption that this refactoring will happen.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
...
+                    items {
+                        id
+                        title
+                        publisher_id
+                    }
+                }
+            }";
 
-        [TestCleanup]
+        internal const string BOOK_DBO_CONTENTS =
```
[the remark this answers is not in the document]
aaronburtle: Would this not be the same thing except in its own file?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -466,11 +467,278 @@ type Moon {
     character: Character
 }
 ";
+        internal const string HOT_RELOAD_CONFIG_FILE_NAME = "hot-reload.dab-config.json";
```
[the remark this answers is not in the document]
aaronburtle: I made the config file generator take in the name as argument so you can overload, but I suppose it doesn't hurt to be a little more defensive and make the names unique to begin with in each test.

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: I've added a class initialization, which I think is best unless we end up parallelizing things in the future.

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: ya that should be fine.

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: Since there's only 1 path property validating that the new path works ought to also cover the old path not working, but I can add in an assert that the old path isnt valid.

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: I will mention this and some other improvements we can do in the task we have tracking test cleanup and improvements.

seantleonard (approved):

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
```diff
@@ -72,7 +72,6 @@ HostMode mode
     /// on hot reload.
     /// </summary>
     [TestMethod]
-    [Ignore]
```
seantleonard, line 75: did you sync with @RubenCerna2079 about this as he was working on [#2109](https://github.com/Azure/data-api-builder/issues/2109) already.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -0,0 +1,302 @@
...
+    /// <summary>
+    /// Hot reload the configuration by saving a new file with different rest and graphQL paths.
+    /// Validate that the response is correct when making a request with the newly hot-reloaded paths.
+    /// </summary>
+    [TestCategory(MSSQL_ENVIRONMENT)]
+    [TestMethod("Hot-reload runtime paths.")]
+    public async Task HotReloadConfigRuntimePathsEndToEndTest()
+    {
+        // Arrange
+        string restBookContents = $"{{\"value\":{_bookDBOConents}}}";
```
seantleonard, line 209: does this compile locally? Codespaces shows there are a few errors in this file.

aaronburtle: it did before I fixed only 1 of the typos. fixing the others now.

src/Service.Tests/Unittests/ConfigFileWatcherUnitTests.cs
aaronburtle: Not yet, he was oof today so I just added the fix here, which was small.

aaronburtle: But will sync tomorrow when he is back :)

Issue #1866, opened 2023-11-07
Add integration and end to end testing for hot reload of config


aaronburtle:
We have unit testing that covers the hot reload scenario in testing, but not integration testing or end to end testing. This issue seeks to address that missing test coverage.

One thing to consider is that with this kind of testing there is a race condition created between the modification of the file, the file watcher detecting that modification, the updating of the config, and the follow up request that utilizes the hot reloaded config. Effective testing needs to solve this race condition.

seantleonard:
Moving this + hot reload related tasks to post GA -> tracking via 1.2rc milestone.

PR #2453, opened 2024-11-07
Added End-To-End Test for Datasource Hot-Reload

RubenCerna2079:
## Why make this change?

This change solves the issue #1866

## What is this change?

Added an end-to-end test to ensure that DAB hot-reloads appropriately when there is a connection change from one database to another database (e.g. MSSQL to PostgreSQL).
In order to make tests related to hot reload less flaky, we stopped using the sleep() function and added an event in `FileSystemRuntimeConfigLoader`, which triggers when hot reload is complete. This event is used by the tests in order to wait until hot reload finishes.

## How was this tested?

- [X] Integration Tests
- [ ] Unit Tests

### DIFF
--- a/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
+++ b/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
@@ -304,13 +304,15 @@ public class ConfigurationHotReloadTests
     }
 
     /// <summary>
-    /// Hot reload the configuration file by saving a new database type and connection string.
-    /// Validate that the response from the server is correct when making a new request after
-    /// the change in database type.
+    /// Hot reload the configuration file by saving a new session-context and connection string.
+    /// Validate that the response from the server is correct, by ensuring that the session-context
+    /// inside the DataSource parameter is different from the session-context before hot reload.
+    /// By asserting that hot reload worked properly for the session-context it also implies that
+    /// the new connection string with additional parameters is also valid.
     /// </summary>
     [TestCategory(MSSQL_ENVIRONMENT)]
     [TestMethod]
-    public async Task HotReloadConfigDataSourceEndToEndTest()
+    public async Task HotReloadConfigDataSource()
     {
         // Arrange
         RuntimeConfig previousRuntimeConfig = _configProvider.GetConfig();
@@ -346,7 +348,7 @@ public class ConfigurationHotReloadTests
     /// </summary>
     [TestCategory(MSSQL_ENVIRONMENT)]
     [TestMethod]
-    public async Task HotReloadConfigConnectionStringEndToEndTest()
+    public async Task HotReloadConfigConnectionString()
     {
         // Arrange
         _writer = new StringWriter();
@@ -390,7 +392,7 @@ public class ConfigurationHotReloadTests
     /// </summary>
     [TestCategory(MSSQL_ENVIRONMENT)]
     [TestMethod]
-    public async Task HotReloadConfigDatabaseTypeEndToEndTest()
+    public async Task HotReloadConfigDatabaseType()
     {
         // Arrange
         _writer = new StringWriter();
--- a/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
+++ b/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
@@ -1,12 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.IO;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Json;
 using System.Text.Json;
 using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config;
 using Azure.DataApiBuilder.Config.ObjectModel;
 using Azure.DataApiBuilder.Core.Configurations;
 using Azure.DataApiBuilder.Service.Tests.SqlTests;
@@ -23,8 +25,11 @@ public class ConfigurationHotReloadTests
     private static TestServer _testServer;
     private static HttpClient _testClient;
     private static RuntimeConfigProvider _configProvider;
+    private static FileSystemRuntimeConfigLoader _configLoader;
     internal const string CONFIG_FILE_NAME = "hot-reload.dab-config.json";
     internal const string GQL_QUERY_NAME = "books";
+    private static TaskCompletionSource<bool> _tsc;
+    private static Task<bool> _task;
 
     internal const string GQL_QUERY = @"{
                 books(first: 100) {
@@ -162,6 +167,9 @@ public class ConfigurationHotReloadTests
         _testServer = new(Program.CreateWebHostBuilder(new string[] { "--ConfigFileName", CONFIG_FILE_NAME }));
         _testClient = _testServer.CreateClient();
         _configProvider = _testServer.Services.GetService<RuntimeConfigProvider>();
+        _configLoader = _testServer.Services.GetService<FileSystemRuntimeConfigLoader>();
+
+        _configLoader.HotReloadCompleted += OnHotReloadCompleted;
 
         string query = GQL_QUERY;
         object payload =
@@ -311,6 +319,7 @@ public class ConfigurationHotReloadTests
     public async Task HotReloadConfigDatasourceDatabaseTypesEndToEndTest(DatabaseType expectedDatabaseType, string expectedTestCategory)
     {
         // Arrange
+        StartTaskCompletionOperation();
         RuntimeConfig previousRuntimeConfig = _configProvider.GetConfig();
         DatabaseType previousDatabaseType = previousRuntimeConfig.DataSource.DatabaseType;
         string expectedConnectionString = $"{ConfigurationTests.GetConnectionStringFromEnvironmentConfig(expectedTestCategory).Replace("\\", "\\\\")}";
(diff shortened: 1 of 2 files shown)


src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +299,36 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+    [DataRow(DatabaseType.PostgreSQL, TestCategory.POSTGRESQL, DisplayName = "Hot-reload change to PostgreSQL Datasource")]
+    public async Task HotReloadConfigDatasourceDatabaseTypesEndToEndTest(DatabaseType expectedDatabaseType, string expectedTestCategory)
+    {
+        // Arrange
+        string expectedConnectionString = $"{ConfigurationTests.GetConnectionStringFromEnvironmentConfig(expectedTestCategory).Replace("\\", "\\\\")}";
+
+        GenerateConfigFile(
+            databaseType: expectedDatabaseType,
+            connectionString: expectedConnectionString);
+        System.Threading.Thread.Sleep(3000);
```
Aniruddh25, line 319: Instead of adding a timer here, why dont we check for the existence of the new file, and wait until the new file is generated? Adding timer to the test makes it flaky.

RubenCerna2079: The file is already created beforehand and we are only updating the file.
We use a timer in order to ensure that the server has enough time to hot-reload before moving on with the rest of the testing.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +299,36 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+
+        // Act
+        RuntimeConfig updatedRuntimeConfig = _configProvider.GetConfig();
```
Aniruddh25, line 322: it might be useful to retrieve what the database type was before the hot reload, then assert it was not the same as the expected one - because the idea would be to test changing the db type..

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +299,36 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+    /// </summary>
+    [DataTestMethod]
+    [DataRow(DatabaseType.MySQL, TestCategory.MYSQL, DisplayName = "Hot-reload change to MySQL Datasource")]
+    [DataRow(DatabaseType.PostgreSQL, TestCategory.POSTGRESQL, DisplayName = "Hot-reload change to PostgreSQL Datasource")]
+    public async Task HotReloadConfigDatasourceDatabaseTypesEndToEndTest(DatabaseType expectedDatabaseType, string expectedTestCategory)
+    {
+        // Arrange
+        string expectedConnectionString = $"{ConfigurationTests.GetConnectionStringFromEnvironmentConfig(expectedTestCategory).Replace("\\", "\\\\")}";
+
+        GenerateConfigFile(
```
Aniruddh25, line 316: Does this function replace the generated config file - I mean overwrite it?

RubenCerna2079: It generates the config file if it doesn't exist, and it overwrites it if it exists.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
Aniruddh25: I understand why you are using timer, but what if the update doesnt happen within 3 sec? can we instead check for the condition you expect to be true after the expiry of the timer instead of using a timer?

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +299,36 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+            databaseType: expectedDatabaseType,
+            connectionString: expectedConnectionString);
+        System.Threading.Thread.Sleep(3000);
+
+        // Act
```
Aniruddh25, line 321: The //Act part of the test is actually `GenerateConfigFile`

Here you are simply retrieving the latest values from the server to verify they are accurate.

Aniruddh25 (changes requested):
Need to avoid usage of timer

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
RubenCerna2079: If I remember correctly, DAB is not able to work on any requests until hot-reload is finished. If that is the case, by having a request with await then the test should wait until the request and hot-reload are completed.
(Please correct me if I am wrong on that)

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +299,41 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+            _configProvider,
+            GQL_QUERY_NAME,
+            GQL_QUERY);
+        RuntimeConfig updatedRuntimeConfig = _configProvider.GetConfig();
+        DatabaseType actualDatabaseType = updatedRuntimeConfig.DataSource.DatabaseType;
+
+        // Assert
+        Assert.AreNotEqual(previousDatabaseType, actualDatabaseType);
+        Assert.AreEqual(expectedDatabaseType, actualDatabaseType);
+        SqlTestHelper.PerformTestEqualJsonStrings(_bookDBOContents, reloadGQLContents.GetProperty("items").ToString());
```
abhishekkumams, line 337: _bookDBOContents has values from MSSQL, currently all the DBs have the same values, so we don't know if it actually picked the value from MSSQL or the updated DB. Would you be able to add one row to each db with different values which we can use for more robust testing?

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
abhishekkumams: @RubenCerna2079 , what happens if we don't add a sleep, does it fail?

abhishekkumams:
Can you also add one more test where you update the connection string with invalid value and then change it to a valid value, with 200 response?

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
abhishekkumams: Also Based on how our pipelines work, we do not support cross database access in single pipeline. But you can add another test which i mentioned in this issue itself.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
RubenCerna2079: @abhishekkumams, yes, since it does not give enough time for the hot-reload to take place

RubenCerna2079:
> Can you also add one more test where you update the connection string with invalid value and then change it to a valid value, with 200 response?

I don't think we need this test, since the hot-reload has a last known good config file to go to when there is an error, so it will not change anything by trying to change the connection string to an invalid value. And a successful connection string change is already tested with the current test that I am working on.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
RubenCerna2079: Created an event that triggers when hot reload is completed.
This will be used in the tests to make them wait until hot reload finishes, before continuing with the tests.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +302,127 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
     }
+
+    /// <summary>
+    /// Hot reload the configuration file by saving a new database type and connection string.
+    /// Validate that the response from the server is correct when making a new request after
+    /// the change in database type.
+    /// </summary>
+    [TestCategory(MSSQL_ENVIRONMENT)]
+    [TestMethod]
+    public async Task HotReloadConfigDataSourceEndToEndTest()
```
abhishekkumams, line 313: No need to add EndToEnd on the test names.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +302,127 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+
+        string failedKeyWord = "Unable to hot reload configuration file due to";
+        string succeedKeyWord = "Validated hot-reloaded configuration file";
+
+        // Act
+        // Hot Reload should fail here
+        GenerateConfigFile(
+            databaseType: DatabaseType.PostgreSQL,
+            connectionString: $"{ConfigurationTests.GetConnectionStringFromEnvironmentConfig(TestCategory.POSTGRESQL).Replace("\\", "\\\\")}");
+        System.Threading.Thread.Sleep(5000);
```
abhishekkumams, line 407: why someplace sleep time is 5000, someplace its 8000? was the test failing?

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -299,4 +302,127 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
         // Assert
         Assert.AreEqual(HttpStatusCode.NotFound, gQLResult.StatusCode);
     }
+
+    /// <summary>
+    /// Hot reload the configuration file by saving a new database type and connection string.
```
abhishekkumams, line 307: is this test changing database Type?

abhishekkumams (approved):
Thanks for adding the tests. Just few nits.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
RubenCerna2079: Yes, the test was failing and it was due to a lack of enough time for it to complete

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
RubenCerna2079: Forgot to update the summary

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
RubenCerna2079: Changed the order of some operations and changed sleep from 8000 to 5000 to make test less flaky

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -301,34 +304,129 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+
+        // Act
+        // Hot Reload should fail here
+        GenerateConfigFile(
+            databaseType: DatabaseType.PostgreSQL,
+            connectionString: $"{ConfigurationTests.GetConnectionStringFromEnvironmentConfig(TestCategory.POSTGRESQL).Replace("\\", "\\\\")}");
+        System.Threading.Thread.Sleep(5000);
+
+        // Log that shows that hot-reload was not able to validate properly
+        string failedConfigLog = $"{_writer.ToString()}";
```
Aniruddh25, line 414: once the failedConfigLog is gathered, can you please clear up the writer's contents so that we dont check for old contents as well?

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -301,34 +304,129 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+
+        // Act
+        // Hot Reload should fail here
+        GenerateConfigFile(
+            databaseType: DatabaseType.MSSQL,
+            connectionString: "");
+        System.Threading.Thread.Sleep(8000);
+
+        // Log that shows that hot-reload was not able to validate properly
+        string failedConfigLog = $"{_writer.ToString()}";
```
Aniruddh25, line 368: clear up writer's contents

Aniruddh25 (approved):
Looks good, left a few suggestions

PR #2466, opened 2024-11-19
Added ValidateAspNetCoreUrls

### LINKED ISSUE #2465: [Bug]: Incomprehensible error message
In my ENV file, I set the HTTP and HTTPS ports like this:

`ASPNETCORE_URLS="http://localhost:5000;https://localhost:5001"`

If I remove the quotes and write it like this:

`ASPNETCORE_URLS=http://localhost:5000;https://localhost:5001`

It fails, as expected. However, the error message doesn’t explain how to fix the issue.

### Error message

```
fail: Microsoft.Extensions.Hosting.Internal.Host[11]
      Hosting failed to start
      System.InvalidOperationException: A path base can only be configured using IApplicationBuilder.UsePathBase().
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.ParseAddress(String address, Boolean& https)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.AddressesStrategy.BindAsync(AddressBindContext context, CancellationToken cancellationToken)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.BindAsync(ListenOptions[] listenOptions, AddressBindContext context, Func`2 useHttps, CancellationToken cancellationToken)
         at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerImpl.BindAsync(CancellationToken cancellationToken)
         at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerImpl.StartAsync[TContext](IHttpApplication`1 application, CancellationToken cancellationToken)
         at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
         at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__15_1(IHostedService service, CancellationToken token)
         at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__15_1(IHostedService service, CancellationToken token)
         at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__15_1(IHostedService service, CancellationToken token)
         at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](IEnumerable`1 services, CancellationToken token, Boolean concurrent, Boolean abortOnFirstException, List`1 exceptions, Func`3 operation)
Unable to launch the runtime due to: System.InvalidOperationException: A path base can only be configured using IApplicationBuilder.UsePathBase().
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.ParseAddress(String address, Boolean& https)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.AddressesStrategy.BindAsync(AddressBindContext context, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.AddressBinder.BindAsync(ListenOptions[] listenOptions, AddressBindContext context, Func`2 useHttps, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerImpl.BindAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerImpl.StartAsync[TContext](IHttpApplication`1 application, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__15_1(IHostedService service, CancellationToken token)
   at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](IEnumerable`1 services, CancellationToken token, Boolean concurrent, Boolean abortOnFirstException, List`1 exceptions, Func`3 operation)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.Run(IHost host)
   at Azure.DataApiBuilder.Service.Program.StartEngine(String[] args) in /_/src/Service/Program.cs:line 39
Error: Failed to start the engine.
```

JerryNixon:
This pull request introduces a validation check for the `ASPNETCORE_URLS` environment variable to ensure it is in the correct format before starting the engine in the `Program` class. This change helps in preventing runtime errors due to misconfigured URLs.

Key changes:

* Added a new method `ValidateAspNetCoreUrls` to check the format of the `ASPNETCORE_URLS` environment variable. This method splits the environment variable by commas or semicolons and ensures each part is a valid absolute URI.
* Integrated the `ValidateAspNetCoreUrls` method into the `Main` method to validate the URLs before proceeding with the engine start. If the validation fails, an error message is printed, and the application exits with an error code.

## Why make this change?

The primary motivation is to return a more clear error message to the user. 

Closes #2465

## What is this change?

Startup validation.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

Valid value in END file: `ASPNETCORE_URLS="http://localhost:5000;https://localhost:5001"`

Invalid value in END file: `ASPNETCORE_URLS=http://localhost:5000;https://localhost:5001`


abhishekkumams:
@JerryNixon , you can run `dotnet format` command to fix formatting issues

src/Service/Program.cs
```diff
@@ -23,10 +23,23 @@ public class Program
...
+                Console.Error.WriteLine("Invalid ASPNETCORE_URLS format. See: https://github.com/Azure/data-api-builder/issues/2237");
+                Environment.Exit(-1);
+            }
+
             if (!StartEngine(args))
             {
                 Environment.ExitCode = -1;
             }
+
+            static bool ValidateAspNetCoreUrls()
```
abhishekkumams, line 37: This should be moved out of `Main` method and a small summary should be added.

src/Service/Program.cs
```diff
@@ -23,10 +23,23 @@ public class Program
 
         public static void Main(string[] args)
         {
+            if (!ValidateAspNetCoreUrls())
+            {
+                Console.Error.WriteLine("Invalid ASPNETCORE_URLS format. See: https://github.com/Azure/data-api-builder/issues/2237");
```
abhishekkumams, line 28: Also instead of issue, can we just put examples in the summary of the method:`ValidateAspNetCoreUrls`?

src/Service/Program.cs
JerryNixon: Why?

JerryNixon: A local method communicates that its scope is limited to this method. Moving it to the class level implies it might be intended for multiple methods, but it is not. I can move it if "just because we do it that way" is the answer. Otherwise, are you sure?

JerryNixon: I moved it, but it feels wrong.

src/Service/Program.cs
```diff
@@ -205,5 +213,17 @@ private static void AddConfigurationProviders(
...
+        private static bool ValidateAspNetCoreUrls()
+        {
+            if (Environment.GetEnvironmentVariable("ASPNETCORE_URLS") is not string value)
+            {
+                return true; // If the environment variable is missing, then it cannot be invalid.
+            }
+
+            return value
+                .Split(new[] { ',', ';' }, StringSplitOptions.RemoveEmptyEntries)
+                .All(x => Uri.TryCreate(x.Trim(), UriKind.Absolute, out _));
```
RubenCerna2079, line 226: Could someone explain what is the purpose of the different types of UriKind?

src/Service/Program.cs
```diff
@@ -23,6 +24,13 @@ public class Program
 
         public static void Main(string[] args)
         {
+            if (!ValidateAspNetCoreUrls())
+            {
+                Console.Error.WriteLine("Invalid ASPNETCORE_URLS format. e.g.: ASPNETCORE_URLS=\"http://localhost:5000;https://localhost:5001\"");
+                Environment.ExitCode = -1;
```
aaronburtle, line 30: Can we use an enumeration instead of simply an integer, just to communicate what the code means more clearly?

src/Service.Tests/Unittests/EnvironmentTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        // Arrange
+        Environment.SetEnvironmentVariable(ASPNETCORE_URLS_NAME, ASPNETCORE_URLS_INVALID_VALUE);
+        using StringWriter consoleOutput = new();
+        Console.SetError(consoleOutput);
+
+        // Act
+        Program.Main(Array.Empty<string>());
+
+        // Assert
+        Assert.AreEqual(-1, Environment.ExitCode);
```
aaronburtle, line 39: if we do use an enum for the exit code than would make sense to use that here too.

src/Service/Program.cs
aaronburtle: > Could someone explain what is the purpose of the different types of `UriKind`?

It is to indicate if you are using an "absolute" URL, which means you are including the entire URL, or if you are using a "relative" URL, which would then be concatenated onto the end of the "base" URL.

ie: "http://foo.com" vs "foo.com" where we have a base URL of "http://"

src/Service/Program.cs
aaronburtle: actually I see now that we were aclready just using -1 to mean failure generally, so probably don't need an enum here, since this isn't specific to the kind of failure you've added.

src/Service.Tests/Unittests/EnvironmentTests.cs
aaronburtle: we arent adding a new exit code here so an enum isnt really needed.

aaronburtle:
The failing unit test appears to be caused by stale environment variables from one test causing another to fail due to the new validation, when that test expects the program main function to work.

I was able to get the pipeline to pass on another branch (https://github.com/Azure/data-api-builder/tree/dev/aaronburtle/ASP_URL_DEBUGG) by adding in the Environment variable that you are validating to the Environment variable clearing helper function in testhelper.

https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=7640&view=results

This is the change the seemed to fix it, which also requires that the environment variable constant be moved to the class as a public field so we can borrow its use for testhelper.

image

src/Service.Tests/Unittests/EnvironmentTests.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// </summary>
+    /// <remarks>
+    /// This test sets the <c>ASPNETCORE_URLS</c> environment variable to an invalid value, invokes the <c>Main</c> method,
+    /// and verifies that the application exits with an error code of -1. Additionally, it checks if the error message
+    /// contains the name of the invalid environment variable.
+    /// </remarks>
+    [TestMethod]
+    public void Main_WhenAspNetCoreUrlsInvalid_ShouldExitWithError()
+    {
+        const string ASPNETCORE_URLS_NAME = "ASPNETCORE_URLS";
```
aaronburtle, line 27: if you make this public and put it at the class level, can re-use it for the testhelper function that unsets all dab environment variables.

aaronburtle (approved):
Looks good, just needs a small update so all the tests will pass.

src/Service/Program.cs
abhishekkumams: my reasoning was that i think this could be used by CLI in future, and it usage could be more than just local.

abhishekkumams (approved):

Issue #2119, opened 2024-03-19
Handle transient error messages while editing config file during hot reload


aaronburtle:
In order to hot reload we must monitor the config file for changes. This results in constant reads by the file watcher that we instantiate to do this task. However, when editing the config file, that editor will take a write-lock on the config file, resulting in the file watcher throwing errors. These errors are expected and will continue until the editor has relinquished its write lock on the config file.

Before we introduce the hot-reload feature, we must determine how we will handle these error messages.

For example, some editors will only take a write lock while saving changes made to the file. In this case, the error messaging will be limited, perhaps only a single message, and is likely to cause confusion. It might be best to ignore these limited transient errors by not recording the first small number of errors.

But it could also be the case that the editor has taken a write-lock the entire time the file is open, not just during saves. In this case a hot reload will not be possible until that editor is closed, and the error messaging will be valuable. We should maintain these error messages, perhaps by making sure we do in fact log errors beyond the first few.

Whatever the design decision, we must gracefully handle both cases so that the error messaging is consistent and clear.

Related to https://github.com/Azure/data-api-builder/issues/2064

Related to https://github.com/Azure/data-api-builder/pull/2108

PR #2467, opened 2024-11-19
Add exponential backoff to reading config file

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2119

## What is this change?

We add a try catch around the file reading in the `FileSystemRuntimeConfigLoader` class, and implement an exponential backoff for reading said file.

## How was this tested?

Manually tested and ran against the test suite. A manual test can be run by having an editor with the file in question open and then closing the editor during the reading process.

## Sample Request(s)

N/A

### DIFF
--- a/src/Config/Utilities/FileUtilities.cs
+++ b/src/Config/Utilities/FileUtilities.cs
@@ -11,14 +11,14 @@ internal class FileUtilities
     /// <summary>
     /// Limit the number of retries when failures occur.
     /// </summary>
-    private static readonly int _runLimit = 3;
+    public static readonly int RunLimit = 3;
 
     /// <summary>
     /// Base of the exponential retry back-off.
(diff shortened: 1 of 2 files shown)


src/Config/Utilities/FileUtilities.cs
```diff
@@ -42,7 +42,7 @@ public static byte[] ComputeHash(IFileSystem fileSystem, string filePath)
         int runCount = 1;
 
         // Maximum 2^_runLimit seconds of wait time due to retries.
```
RubenCerna2079, line 44: Change the variable name in the comment
```suggestion
        // Maximum 2^RunLimit seconds of wait time due to retries.
```

RubenCerna2079:
LGTM

abhishekkumams (approved):

PR #2471, opened 2024-11-20
Fix namespace collision for string and CollectionUtils method `IsNullOrEmpty`

### LINKED ISSUE #2469: [Bug]: Conflicting namespace for method isNullOrEmpty causing update mutation to break
### What happened?

Attempt by method 'Azure.DataApiBuilder.Core.Resolvers.MsSqlQueryBuilder.Build(Azure.DataApiBuilder.Core.Models.LabelledColumn, System.String)' to access method 'Microsoft.IdentityModel.Tokens.CollectionUtilities.IsNullOrEmpty<System.Char>(System.Collections.Generic.IEnumerable1<Char>)' failed is thrown when running an update mutation command. This is happening due to https://github.com/alexhiggins732/IdentityServer8/issues/40 and namespace conflict where string.isnullorempty should be used but it is being resolved to IdentityModel.Token.CollectionUtilities

Query:
The mutation is:
mutation{
  updateProduct(ProductID: 680 ,item: {Color: "Blue"})  {
     ProductID
     Color
  }
}
### Version

1.2

Azure SQL

_No response_

_No response_

### Relevant log output

```Text

```

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #2469 
  - There is a namespace collision for the `IsNullOrEmpty` method in the `MsSqlQueryBuilder` class between String and CollectionUtils class.

## What is this change?

- `MsSqlQueryBuilder` class had a check in the `Build` method where it was using `IsNullOrEmpty` for Strings but it was refrencing CollectionUtils Class instead
- Removed `Collection.Utilities` class from Cli and renamed it in the Core project to `EnumerableUtilities` to avoid any future namespace collision.
- Also updated all the references of `IsNullOrEmpty`

## How was this tested?

- [x] manual tests
- [x] current existing tests

Notes:
Work Around shared here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1722


src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -421,7 +421,7 @@ private string MakeOutputColumns(List<LabelledColumn> columns, string columnPref
         /// </summary>
         private string Build(LabelledColumn column, string columnPrefix)
         {
-            if (columnPrefix.IsNullOrEmpty())
+            if (string.IsNullOrEmpty(columnPrefix))
```
seantleonard, line 424: this now matches correct usage of : https://learn.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=net-8.0

String doesn't have a built-in extension method `IsNullOrEmpty()`

might be worth doing search for `IsNullOrEmpty()` throughout codebase to ensure this isn't elsewhere and identify why intellisense/compilation process doesn't complain about this.

aaronburtle (approved):
Thanks for fixing this one, I will follow up with another PR that fixes any additional usages.

aaronburtle:
Actually, after searching the codebase there was only 1 more instance where string was being null checked wrong, so just added it in here, hope that's OK @abhishekkumams

src/Core/Resolvers/MsSqlQueryBuilder.cs
aaronburtle: looks like we had already fixed nearly all of these, just 1 more case which I added to this PR.

RubenCerna2079: @aaronburtle I am not sure if it is the case but I think I found one in `Core\Resolvers\SqlMutationEngine.cs` method `BuildAndExecuteInsertDbQueries():exposedColumnNames.IsNullOrEmpty()`

abhishekkumams: @RubenCerna2079 , exposedColumnNames is **not** of String type.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -421,7 +421,7 @@ private string MakeOutputColumns(List<LabelledColumn> columns, string columnPref
...
             {
```
Aniruddh25, line 425: If we now use the string.IsNullOrEmpty() method, do we still needthe CollectionUtils class method added in this commit: [AuthN Config V2 -> HotReload Aware Authentication settings in dev mod… · Azure/data-api-builder@e777330](https://github.com/Azure/data-api-builder/commit/e77733068d6947aadde540938f47160ce6a486fe)
If we dont need it, lets remove that class in this PR.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -421,7 +421,7 @@ private string MakeOutputColumns(List<LabelledColumn> columns, string columnPref
...
                 return $"{QuoteIdentifier(column.ColumnName)} AS {QuoteIdentifier(column.Label)}";
             }
```
Aniruddh25, line 427: The description needs to be fixed to remove the leftover PR template text.

Aniruddh25 (dismissed):
Need to remove unnecessary CollectionUtils class

src/Core/Resolvers/MsSqlQueryBuilder.cs
seantleonard: there were instances in our codebase where `.IsNullorEmpty()` was appended to collections which don't natively provide such an extension method. At one point, an update started using IdentityModel.Tokens extenion method to compare collections and that wasn't supposed to be a public method. They updated that method to be internal so we needed a replacement not to make too many extensive changes to existing code. That's where collectionUtils came in: just copying into our code the extension method to check whether a collection is null or empty. 

So `string.IsNullOrEmpty()` can be used for string, but we still have collections attempting to use that extension.

sezal98 (approved):
lgtm

Issue #2457, opened 2024-11-14
[Bug]: Hot Reload Not Re-enabled When Switching from Production to Development Mode


abhishekkumams:
### What happened?

When updating the environment mode from development to production, hot reload is disabled, which is expected behavior. However, when the environment is subsequently changed back from production to development, hot reload is not re-enabled. This behavior is not expected; ideally, if the initial mode was development, hot reload should automatically reactivate when the mode is reverted to development.

**Steps to Reproduce:**

1. Start the application in development mode with hot reload active.
2. Change the environment mode to production.
3. Change the environment mode back to development.

**Expected Behavior:**
Hot reload should be re-enabled automatically when the environment is set back to development mode.

**Actual Behavior:**
Hot reload remains inactive after switching from production to development mode.

### Version

1.3.13-rc

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

seantleonard:
While yes we did consider "Production" -> "Development" mode as part of the scope, this change becomes difficult because there are checks in DAB that evaluate "IsHotReloadEnabled()" which considers dev/prod mode and forbids hot reload when in prod mode. You'd need to signal to hot reload functions that the "Mode" is explicitly changing.

If building in support for this, I'd recommend ensuring that the change includes logging that indicates hot reload is occurring because of a dev/prod mode switch.

Also would be beneficial to check dev mode to prod mode hot reload scenario. There are some instances in startup that are conditioned on Dev mode/prod mode only and those may be skipped when switching mode via hot reload as startup has already completed.

abhishekkumams:
I gave more thoughts on this and believe its best to keep the existing functionality:

1. Maintain the current behavior where hot reload does not automatically resume when switching back to development mode.
2. This forces a conscious decision to re-enable hot reload and reduces risks associated with inadvertent switches or malicious toggles.

aaronburtle:
[...]

PR #2478, opened 2024-11-26
Do not allow Mode to Hot-Reload

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2457

## What is this change?

We add in an optional parameter to the `TryLoadConfig` function so that when we Hot-Reload, we can pass along the `Mode`. When that parameter is included (and therefore non null), we overwrite the Mode of the `RuntimeConfig` to match the value of the previous `RuntimeConfig`. This means that this Mode needs to have a public setter however, which we now utilize to overwrite.

## How was this tested?

Modified our Unit test to provide coverage.

## Sample Request(s)

Any hot-reload that tries to change the mode will demonstrate the functionality. So start the engine in dev mode, so that the filewatcher will pick up any changes as we dont currently support hot-reload in production, then change the Host Mode from development to production, and note that the RuntimeConfig in-memory object is still in dev mode.


abhishekkumams:
Thanks @aaronburtle , we would also need to log error in console if user tries to hot-reload hostmode.

abhishekkumams (approved):

src/Config/ObjectModel/HostOptions.cs
```diff
@@ -32,7 +32,7 @@ public record HostOptions
     public AuthenticationOptions? Authentication { get; init; }
 
     [JsonPropertyName("mode")]
-    public HostMode Mode { get; init; }
+    public HostMode Mode { get; set; }
```
RubenCerna2079, line 35: Why do we need to change it to `set`?

RubenCerna2079 (approved):

src/Config/ObjectModel/HostOptions.cs
aaronburtle: This is because `init` will only allow the value to be changed in the object initializer. In order to overwrite this value from another class, such as when we hot-reload, we need to have a mutator that is public, which is what `set` provides.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -234,6 +235,26 @@ public bool TryLoadConfig(
 
                 config = RuntimeConfig;
 
+                // When isDevMode is not null it means we are in a hot-reload scenario, and need to save the previous
+                // mode in the new RuntimeConfig since we do not support hot-reload of the mode.
+                if (isDevMode is not null && config.Runtime is not null && config.Runtime.Host is not null)
```
sezal98, line 240: should we rename this var to `incomingRuntimeConfig` since its a little confusing if its the updated one or current one

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -234,6 +235,26 @@ public bool TryLoadConfig(
...
+                {
+                    // Log error when the mode is changed during hot-reload. 
+                    if (isDevMode.Value && HostMode.Production.Equals(config.Runtime.Host.Mode))
```
sezal98, line 243: What would happen if the mode is removed from RuntimeConfig during hot reload?

abhishekkumams: if mode is removed from the config, it will take the default value, i.e, Production.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -164,9 +164,9 @@ private void OnNewFileContentsDetected(object? sender, EventArgs e)
     {
         try
         {
-            if (RuntimeConfig is not null && RuntimeConfig.IsDevelopmentMode())
+            if (RuntimeConfig is not null)
```
Aniruddh25, line 167: by removing this condition, isnt it effectively enabling hot reload in production mode too?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -234,6 +235,26 @@ public bool TryLoadConfig(
 
                 config = RuntimeConfig;
 
+                // When isDevMode is not null it means we are in a hot-reload scenario, and need to save the previous
+                // mode in the new RuntimeConfig since we do not support hot-reload of the mode.
+                if (isDevMode is not null && config.Runtime is not null && config.Runtime.Host is not null)
+                {
+                    // Log error when the mode is changed during hot-reload. 
+                    if (isDevMode.Value && HostMode.Production.Equals(config.Runtime.Host.Mode))
+                    {
```
Aniruddh25, line 244: what if mode is attempted to be changed from PRoduction to Development? No warning logged then?

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: No, because we only ever file monitor if we start the engine in devmode.

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: That shouldnt be possible for now, but since we will have production file monitoring in the future, can make this check cover that case as well for when we add in the functionality.

Issue #2477, opened 2024-11-26
[Bug]: Improve schema inference experience for azure cosmosdb


sajeetharan:
### What happened?

If the config file lacks a container name, running the `export` command throws an error, which shouldn't happen. The error message should also be more relevant. Since schema inference is now optional, it should work seamlessly with all CLI commands. Please update the `dab init` command to ensure all required parameters for schema generation are properly set.

### Version

1.3.13-rc

### What database are you using?

CosmosDB NoSQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

GraphQL

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

sourabh1007:
Addressed in attached PR

Issue #2476, opened 2024-11-26
[Bug]: Schema inference for Azure Cosmos DB does not work


sajeetharan:
### What happened?

Shema inference should work with dab export --generate command, currently it throws the below error and also shows successful message as schema generated successfully

![Image](https://github.com/user-attachments/assets/47664ebe-9e78-4ea5-bc2f-1b39fc774f9e)

### Version

1.3.13-rc

### What database are you using?

CosmosDB NoSQL

### What hosting model are you using?

Local (including CLI)

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

sajeetharan:
Please fix the error message also show relevant message (In this case all connection string, account name, database name, container name was part of the config)

sourabh1007:
There are two points to note:
1) The log message is incorrect.
2) From a functionality perspective, we had previously decided (you might recall this discussion) to support only a single/global container configuration for now, with plans to expand support based on customer requests in the future. This is why the error occurs when container name information is missing.
It seems we now need to add that support. Therefore, I am including these changes in my new PR.

sourabh1007:
Addressed in attached PR

PR #2479, opened 2024-11-27
Cosmos DB: Adds multi container support in schema generation

sourabh1007:
## Why make this change?

ref. https://github.com/Azure/data-api-builder/issues/2477 and https://github.com/Azure/data-api-builder/issues/2476

## What is this change?
1. Added functionality to traverse all configured containers in the configuration file and generate schemas based on the entities defined in the file.
2. Bug Fix: Resolved an issue where schema generation failed if an empty array was present in the JSON.
3. Bug Fix: Resolved an issue where schema generation failed if singular/plurals are defined in PascalCase.
4. Improved logging: Adjusted the success logger to trigger only when a non-empty schema is successfully generated.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

## Sample Request(s)


abhishekkumams:
this is a breaking change. since it was not mandatory before.

sourabh1007:
> this is a breaking change. since it was not mandatory before.

Thanks for your comment...I have changed the approach..

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -44,7 +44,7 @@ private SchemaGenerator(List<JsonDocument> data, string containerName, RuntimeCo
                 // Populate entity and singular name mapping if configuration is provided.
                 foreach (KeyValuePair<string, Entity> item in config.Entities)
                 {
-                    _entityAndSingularNameMapping.Add(item.Value.GraphQL.Singular, item.Key);
+                    _entityAndSingularNameMapping.Add(item.Value.GraphQL.Singular.Pascalize(), item.Key);
```
abhishekkumams, line 47: why it needs to be pascalized?

src/Service.Tests/CosmosTests/SchemaGeneratorFactoryTests.cs
```diff
@@ -0,0 +1,121 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace Azure.DataApiBuilder.Service.Tests.CosmosTests
+{
+    [TestClass, TestCategory(TestCategory.COSMOSDBNOSQL)]
+    public class SchemaGeneratorFactoryTests
+    {
+        [TestMethod]
+        public async Task ExportGraphQLFromCosmosDB_GeneratesSchemaSuccessfully()
```
abhishekkumams, line 24: please add summary.

src/Core/Generator/SchemaGeneratorFactory.cs
```diff
@@ -56,46 +58,78 @@ public static async Task<string> Create(RuntimeConfig config, string mode, int?
...
-            {
-                logger.LogError("No data was sampled. Please try a different sampling mode or provide different sampling options.");
-                throw new ArgumentException("No data was sampled. Please try a different sampling mode or provide different sampling options.");
-            }
+                // Connect to the Azure Cosmos DB container.
+                container ??= ConnectToCosmosDB(connectionString, dbContainer[0], dbContainer[1]);
+                SamplingModes samplingMode = (SamplingModes)Enum.Parse(typeof(SamplingModes), mode);
+
+                // Determine the appropriate sampler based on the sampling mode.
+                ISchemaGeneratorSampler schemaGeneratorSampler = samplingMode switch
```
abhishekkumams, line 107: we can move some logic out of this like getSampler

src/Core/Generator/SchemaGenerator.cs
sourabh1007: found a bug where if singular/ plural are defined in pascalized form(which is valid entry IMO), then it was failing while generating schema.

src/Core/Generator/SchemaGeneratorFactory.cs
sourabh1007: I believe its fine, its not too many lines of code for get the sampler

abhishekkumams (approved):
LGTM

src/Core/Generator/Sampler/CosmosExecutor.cs
```diff
@@ -53,7 +53,15 @@ public async Task<List<T>> ExecuteQueryAsync<T>(string query, Action<T?>? callba
...
+                    JsonElement root = jsonDocument.RootElement;
+                    if (jsonDocument.RootElement.ValueKind != JsonValueKind.Array)
+                    {
+                        if (!jsonDocument.RootElement.TryGetProperty("Documents", out root))
+                        {
+                            root = jsonDocument.RootElement;
+                        }
+                    }
 
                     if (root.ValueKind == JsonValueKind.Array)
```
Aniruddh25, line 66: Please consider renaming the root variable. 

At Line 66 it has a confusing double usage of the variable `root` which could either mean 

A. the jsonDocument's root element in 2 cases: 
1. when valuekind is array or
2. when we fail to get the Documents property if the valueKind is not array.

OR the `root` could also mean 
value of the Documents property when the jsonDocument's value kind is NOT Array.

src/Core/Generator/Sampler/CosmosExecutor.cs
```diff
@@ -53,7 +53,15 @@ public async Task<List<T>> ExecuteQueryAsync<T>(string query, Action<T?>? callba
                     string content = await sr.ReadToEndAsync();
 
                     using JsonDocument jsonDocument = JsonDocument.Parse(content);
-                    JsonElement root = jsonDocument.RootElement.GetProperty("Documents");
+
+                    JsonElement root = jsonDocument.RootElement;
+                    if (jsonDocument.RootElement.ValueKind != JsonValueKind.Array)
+                    {
```
Aniruddh25, line 59: The PR description says that it allows schema generation if "an empty array was present in the JSON." however you are checking for value kind != Array here, doesnt explain how this fixes empty array situation.. can you please explain?

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -44,7 +44,7 @@ private SchemaGenerator(List<JsonDocument> data, string containerName, RuntimeCo
...
                 }
```
Aniruddh25, line 48: if you only store the key in pascal format, would non-pascal singular name still work?

src/Core/Generator/Sampler/CosmosExecutor.cs
Aniruddh25: Or is this code fixing something else? If yes, what is it trying to accomplish? Can you provide some valid examples of supported jsonDocuments here?

src/Core/Generator/SchemaGeneratorFactory.cs
```diff
@@ -56,46 +58,78 @@ public static async Task<string> Create(RuntimeConfig config, string mode, int?
...
             {
-                logger.LogError("Connection String, Database and container name must be provided in the config file");
-                throw new ArgumentException("Connection String, Database and container name must be provided in the config file");
+                logger.LogError("Connection String and Database name must be provided in the config file");
+                throw new ArgumentException("Connection String and Database name must be provided in the config file");
             }
 
-            logger.LogInformation("Connecting to Cosmos DB Database: {0}, Container: {1}", databaseName, containerName);
+            HashSet<string> dbAndContainerToProcess = new();
+            if (globalDatabaseName != null && globalContainerName != null)
```
Aniruddh25, line 75: if globalDatabaseName is provided but globalContainerName is not, what is your course of action?

src/Core/Generator/SchemaGeneratorFactory.cs
```diff
@@ -56,46 +58,78 @@ public static async Task<string> Create(RuntimeConfig config, string mode, int?
...
+                if (string.IsNullOrEmpty(dbAndContainer)) // Safety check
+                {
+                    continue;
+                }
 
-            logger.LogInformation("Sampling Started using {0}", schemaGeneratorSampler.GetType().Name);
+                string[] dbContainer = dbAndContainer.Split('.');
+                if (dbContainer.Length != 2)
+                {
+                    logger.LogError("Invalid Database and Container Name provided in the config file");
```
Aniruddh25, line 96: this is mandating entity source name to always be in the format of database.container. What if I dont want to provide database name, and would like to rely on globaldatabase name? In that situation, I should be able to simply use the source name of entity as the container name and rely on global databasename.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -236,6 +236,26 @@ public void TestEmptyJsonArray()
             Assert.ThrowsException<InvalidOperationException>(() => SchemaGenerator.Generate(jsonArray, "containerName"));
         }
 
+        /// <summary>
+        /// Tests the <see cref="SchemaGenerator.Generate"/> method with an empty JSON array.
+        /// Ensures that the method correctly handles an empty input by throwing an appropriate exception.
```
Aniruddh25, line 241: how do you test that an exception was thrown? To me it seems, that the method doesnt throw any exception at all and ignore the an empty array input field

src/Service.Tests/CosmosTests/SchemaGeneratorFactoryTests.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// </summary>
+        [TestMethod]
+        public async Task ExportGraphQLFromCosmosDB_GeneratesSchemaSuccessfully()
+        {
+            // Arrange: Set up test configuration, mocks, and test data.
+
+            // Runtime configuration for the schema generation.
+            RuntimeConfig runtimeConfig = new(
+               Schema: "schema",
+               DataSource: new DataSource(DatabaseType.CosmosDB_NoSQL, "dummy-connection-string", new()
```
Aniruddh25, line 37: consider saying noop-connection-string instead of dummy - policheck will complain

src/Service.Tests/CosmosTests/SchemaGeneratorFactoryTests.cs
```diff
@@ -0,0 +1,133 @@
...
+               {
+                   {"database", "dummy-database" },
+                   {"container", "Container0" }
```
Aniruddh25, line 40: consider adding a mock test where global container is absent, and source of one of the entities has no database name but only a container name. Does the existing code pick up the global database as the source for this entity or does it error out? What is the desired behavior?

Aniruddh25 (changes requested):
Waiting for a few fixes.

src/Core/Generator/Sampler/CosmosExecutor.cs
sourabh1007: It is there to find out the root, in case of simple JSON, root would be actual root but if root has _Documents_ then, documents tag would become the root. It is more of a safety check. In cosmos db, we get documents under "Documents" tag.
Once root is finalized then, it checks if whatever is under that, is an array or not, then process the data accordingly.

src/Core/Generator/Sampler/CosmosExecutor.cs
sourabh1007: this is the fix for that https://github.com/Azure/data-api-builder/blob/dc872a612c084cf1bcd30bd5612f6fb0a117e465/src/Core/Generator/SchemaGenerator.cs#L279, previously if array is empty, it was messing up the logic. After this fix, it won't process empty array at all.

src/Core/Generator/SchemaGenerator.cs
sourabh1007: yes..in existing test, it is in lowercase and still works.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
sourabh1007: typo..fixed.

src/Core/Generator/SchemaGeneratorFactory.cs
sourabh1007: it will check at entity level, what all containers are mentioned, and it will scan only those containers for schema generation.

src/Core/Generator/SchemaGeneratorFactory.cs
sourabh1007: fixed.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -236,6 +236,26 @@ public void TestEmptyJsonArray()
...
+        /// </summary>
+        [TestMethod]
+        public void TestEmptyJsonArrayInPayload()
+        {
+            List<JsonDocument> jsonArray = new() {
+                JsonDocument.Parse(@"{ ""name"": ""John"", ""product"": [], ""isStudent"": false, ""birthDate"": ""1980-01-01T00:00:00Z"" }")};
+            string actualSchema = SchemaGenerator.Generate(jsonArray, "containerName");
+
+            string expectedSchema = @"type ContainerName @model(name: ""ContainerName"") {
+  name: String!,
```
Aniruddh25, line 251: so, the expectation is to skip the empty JSON array as a potential field altogether? here `product`?

src/Core/Generator/SchemaGeneratorFactory.cs
```diff
@@ -56,46 +58,89 @@ public static async Task<string> Create(RuntimeConfig config, string mode, int?
...
-                logger.LogError("No data was sampled. Please try a different sampling mode or provide different sampling options.");
-                throw new ArgumentException("No data was sampled. Please try a different sampling mode or provide different sampling options.");
-            }
+                if (string.IsNullOrEmpty(dbAndContainer)) // Safety check
+                {
+                    continue;
+                }
+
+                string[] dbContainer = dbAndContainer.Split('.');
+                if (dbContainer.Length != 2)
```
Aniruddh25, line 105: With the fix in the latest iteration using the switch statement, Length will always be == 2, right? otherwise you would have thrown an exception already on line 90.. Is this check needed here?

src/Service.Tests/CosmosTests/SchemaGeneratorFactoryTests.cs
```diff
@@ -0,0 +1,155 @@
...
+    [TestClass, TestCategory(TestCategory.COSMOSDBNOSQL)]
+    public class SchemaGeneratorFactoryTests
+    {
+        /// <summary>
+        /// Tests the ExportGraphQLFromCosmosDB method to ensure it generates a GraphQL schema successfully from Cosmos DB.
+        /// </summary>
+        [TestMethod]
+        [DataRow("noop-database", "Container0", "mydb1", "container1", null, "Container0,Container1,Container2", DisplayName = "Test with global and entity level database and container names")]
+        [DataRow(null, "Container0", "mydb1", "container1", "Connection String and Database name must be provided in the config file", null, DisplayName = "Test with missing global database name")]
+        [DataRow("noop-database", null, "mydb1", "container1", "Unexpected Source format", null, DisplayName = "Test with missing global container name")]
```
Aniruddh25, line 33: if global container is missing, wouldnt you rely on entity's container name to be sufficient? So, here it should be `container1`..

src/Service.Tests/CosmosTests/SchemaGeneratorFactoryTests.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// Contains unit tests for SchemaGeneratorFactory related to Cosmos DB NoSQL.
+    /// </summary>
+    [TestClass, TestCategory(TestCategory.COSMOSDBNOSQL)]
+    public class SchemaGeneratorFactoryTests
+    {
+        /// <summary>
+        /// Tests the ExportGraphQLFromCosmosDB method to ensure it generates a GraphQL schema successfully from Cosmos DB.
+        /// </summary>
+        [TestMethod]
+        [DataRow("noop-database", "Container0", "mydb1", "container1", null, "Container0,Container1,Container2", DisplayName = "Test with global and entity level database and container names")]
```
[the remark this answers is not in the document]
Aniruddh25: I didnt realize you had Container2 as a constant value in the test function itself. Perhaps, its better to mention in the test function header that you are using a const entity value as well.

Aniruddh25 (approved):
Looks good overall, but leaving few suggestions to provide clearer description to the Test.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
sourabh1007: yes, because, we don't know the type of this array, it will come up if any other records have this array as non-empty.

src/Service.Tests/CosmosTests/SchemaGeneratorFactoryTests.cs
sourabh1007: added doc

Issue #2059, opened 2024-02-24
[Bug]: `numeric` data type not supported as a parameter for a stored procedure


Aniruddh25:
### What happened?

System.Data.Common.SqlDbType
https://learn.microsoft.com/en-us/dotnet/api/system.data.sqldbtype?view=net-8.0
doesn't have a
Numeric
type, thats why we unfortunately see this error. The autodetection is done by querying
sys.dm_exec_describe_first_result_set
which gives back
numeric
from SQL. We will need to handle this discrepancy between these dependencies in DAB explicitly for this and other possible types which may be absent in
System.Data.Common.SqlDataType

<img width="550" alt="image" src="https://github.com/Azure/data-api-builder/assets/3513779/8e04ba40-9158-45bf-811a-6885087db8b1">

### Version

main

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

REST

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chrispharmac:
Here is a simple minded patch which got us around this issue. It's weird that Microsoft don't include numeric in their enum.

` 
.../Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs  | 1 +
 data-api-builder-1.2.10/src/Core/Services/TypeHelper.cs           | 8 ++++++++
 .../src/Service.Tests/DatabaseSchema-MsSql.sql                    | 5 +++--
 3 files changed, 12 insertions(+), 2 deletions(-)

diff --git a/data-api-builder-1.2.10/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs b/data-api-builder-1.2.10/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
index d7d900d..5c978a1 100644
--- a/data-api-builder-1.2.10/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs	
+++ b/data-api-builder-1.2.10/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs	
@@ -438,6 +438,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                 "Single" => float.Parse(param),
                 "Double" => double.Parse(param),
                 "Decimal" => decimal.Parse(param),
+                "Numeric" => decimal.Parse(param),
                 "Boolean" => bool.Parse(param),
[...]

PR #2497, opened 2024-12-18
Adding numeric type support for data api builder

rohkhann:
## Why make this change?
Addresses bug #2059 

## What is this change?
adds support for the numeric data type.

## How was this tested?
Unit test added.

### DIFF
--- a/src/Core/Services/TypeHelper.cs
+++ b/src/Core/Services/TypeHelper.cs
@@ -111,7 +111,7 @@ namespace Azure.DataApiBuilder.Core.Services
             [SqlDbType.TinyInt] = typeof(byte),
             [SqlDbType.UniqueIdentifier] = typeof(Guid),
             [SqlDbType.VarBinary] = typeof(byte[]),
-            [SqlDbType.VarChar] = typeof(string)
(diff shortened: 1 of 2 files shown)


src/Core/Services/TypeHelper.cs
```diff
@@ -111,7 +111,7 @@ public static class TypeHelper
             [SqlDbType.TinyInt] = typeof(byte),
             [SqlDbType.UniqueIdentifier] = typeof(Guid),
             [SqlDbType.VarBinary] = typeof(byte[]),
-            [SqlDbType.VarChar] = typeof(string)
+            [SqlDbType.VarChar] = typeof(string),
```
abhishekkumams, line 114: why this change?

rohkhann: Reverted.

rusamant (approved):

sezal98:
Can we add a test case with numeric data type to validate this scenario

abhishekkumams:
should we not update `ParseParamAsSystemType` in class `BaseSqlQueryStructure.cs`?

PR #2498, opened 2024-12-18
Fix mssql filter parser for wildcard characters for `LIKE` query

### LINKED ISSUE #2434: [Bug]: filter endsWith: "_" is incorrectly resolved
### What happened?

I created a query like

```graphql
query Customers{
    customer(
        filter: { id: {endsWith: "_CON" } }
    ) {
        endCursor
        hasNextPage
        items {
            id
        }
    }
}
```

I received an empty json response, although the database contains customers with ids ending like "_CON".

I investigated the SQL queries the database receives and found this.

`WHERE Id LIKE '%\_CON'`

### Version

1.2.11+c7ca8db8558a63919c530e454c8f18b45d5b931c

Azure SQL

Container Apps, Local (including CLI)

GraphQL

### Relevant log output

_No response_

### Code of Conduct

abhishekkumams:
## Why make this change?

- Closes #2434 
  - when wildcard characters are used as literal character in search, filter query fails to resolve correctly.
  -  we never use the wild characters of SQL for their intended use as wild characters when queried from GraphQL. They are always used as literal search characters.

## What is this change?
- we are adding `\` whenever a special character (wildcard) is used in the filter query.
- while mysql,pgsql are able to understand and treat `\` as escape charachters, for MsSQL, we have to use `ESCAPE` clause.
- The `ESCAPE` clause in a `LIKE` query specifies a character to treat as an escape character. This is especially useful when your search pattern contains special characters like %, _, or the actual escape character, which otherwise have special meanings in LIKE patterns.
- I have updated the query builder for MsSql, which checks if it is a `LIKE` query and only then it adds an `ESCAPE` clause.

There is a special case for mysql for wildcard (`\`), which needs to be handled separately. I have created an issue to tackle it. #2499

## How was this tested?

- [X] Integration Tests
- [x] manual tests

## Sample Request(s)
image
image
image


src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -60,9 +61,17 @@ public string Build(SqlQueryStructure structure)
                                     Build(structure.PaginationMetadata.PaginationPredicate));
             }
 
+            // we add '\' carachter to escape the special characters in the string, but if special characters are needed to be searched
```
Aniruddh25, line 64:
```suggestion
            // we add '\' character to escape the special characters in the string, but if special characters are needed to be searched
```

Aniruddh25 (commented):
Can you please mention in the PR description that we never use the wild characters of SQL for their intended use as wild characters when queried from GraphQL? They are always used as literal search characters and hence your PR to add the ESCAPE clause makes sense.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
```diff
@@ -407,28 +407,6 @@ public async Task TestStringFiltersNotContains()
...
-        [TestMethod]
-        public async Task TestStringFiltersContainsWithSpecialChars()
-        {
-            string gqlQuery = @"{
-                planets(first: 10, " + QueryBuilder.FILTER_FIELD_NAME + @" : {name: {contains: ""%""}})
-                {
-                    items {
-                        name
-                    }
-                }
```
Aniruddh25, line 424: Why remove the test altogether? We could keep the test and remove the Ignore clause if its working now.

src/Service.Tests/SqlTests/GraphQLFilterTests/MsSqlGQLFilterTests.cs
```diff
@@ -58,6 +58,62 @@ ORDER BY [__column1] asc
...
+            DisplayName = "endsWith: ']'"
+        )]
+        [DataRow(
+            "{ title: { contains: \"\\\\\" } }",
+            "%\\\\%",
+            DisplayName = "Contains single backslash: '\\' "
+        )]
+        [DataRow(
+            "{ title: { contains: \"\\\\\\\\\" } }",
+            "%\\\\\\\\%",
```
Aniruddh25, line 102: Also test for `[^a]` since that is one of the wild characters.. See here: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver16#pattern

src/Service.Tests/CosmosTests/QueryFilterTests.cs
Aniruddh25: I dont see any equivalent test added here for CosmosDB similar to how its added for the other Sql types.

Aniruddh25 (approved):
Looks good overall, left some comments to address.

src/Service.Tests/CosmosTests/QueryFilterTests.cs
abhishekkumams: it was redundant i added this test as a data row in the tests i added in the current PR

src/Service.Tests/SqlTests/GraphQLFilterTests/MsSqlGQLFilterTests.cs
abhishekkumams: that is not a character per- se. Its for giving the range. it could be [^a-x] or [^d], there could be so many combinations.

abhishekkumams: and just for `^`, escape is not required.

RubenCerna2079 (approved):
LGTM, I just left a comment

src/Service.Tests/SqlTests/RestApiTests/Find/DwSqlFindApiTests.cs
```diff
@@ -328,7 +328,7 @@ public class DwSqlFindApiTests : FindApiTestBase
             {
                 "FindTestWithFirstSingleKeyPaginationAndOrderBy",
                 $"SELECT TOP 1 * FROM { _integrationTableName } " +
-                $"ORDER BY title asc, id asc " +
+                $"ORDER BY id asc, title asc " +
```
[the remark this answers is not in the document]
abhishekkumams: yeah, since we added few more rows, the result was changed as title was used as primary sort key, which doesn't give consistent result, updating it to have id as primary sort key, This test is now consistent and independent of new additions.

Issue #2463, opened 2024-11-17
[Bug]: ‼️ `request-body-strict` does not work properly.


JerryNixon:
### What happened?

Setting `request-body-strict` does not work properly.

### Desired (Correct) behavior:

REST methods Put, Post, Patch entity payloads can include primary key properties. They are ignored.

### Actual (incorrect) behavior:

Exception. Entity payloads including primary key properties result in exception.

## Status

It should be noted that the desired behavior is the only reason for the setting. 

## Workaround

Developers using .NET (at least) must unceremoniously remove the primary key properties from their models before calls.

```csharp
public async Task<Player> UpdateAsync(Player player)
{
    ArgumentNullException.ThrowIfNull(player);

    // remove primary key
    var node = JsonSerializer.SerializeToNode(player);
    node?.AsObject().Remove("id");

    var url = $"{baseUrl}/Id/{player.Id}";
    var response = await http.PutAsJsonAsync(url, node);
    var json = await response.Content.ReadAsStringAsync();
    var root = JsonSerializer.Deserialize<PlayerRoot>(json)
        ?? throw new Exception("Response Json is invalid.");
    return root.Players.Single();
}
```

This is what it should look like:

```csharp
public async Task<Player> UpdateAsync(Player player)
{
    ArgumentNullException.ThrowIfNull(player);

    var url = $"{baseUrl}/Id/{player.Id}";
    var response = await http.PutAsJsonAsync(url, player);
    var json = await response.Content.ReadAsStringAsync();
    var root = JsonSerializer.Deserialize<PlayerRoot>(json)
        ?? throw new Exception("Response Json is invalid.");
    return root.Players.Single();
}
```

aaronburtle:
After some investigation it seems that the issue is really with auto-generated primary keys. These operations work with `request-body-strict` but no if the primary key is autogenerated. Fix is incoming.

PR #2507, opened 2025-01-07
Ignore Auto-Generated primary keys in request body when `request-body-strict` is false

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2463

## What is this change?

When we validate columns in the `RequestValidator` through the `ValidateColumn` function we now ignore columns in the request body that are auto-generated primary keys if `request-body-strict` is false.

## How was this tested?

Added an integration test to validate we can do an insert when including the primary keys in the body of the request when request-body-strict is false and those primary keys are auto-generated.

## Sample Request(s)

https://localhost:5001/restApi/Book/

with a body such as

{
    "id": 9999,
    "title": "foo",
    "publisher_id": 1234
}


abhishekkumams (approved):

src/Core/Services/RequestValidator.cs
```diff
@@ -433,12 +435,18 @@ public void ValidateUpsertRequestContext(UpsertRequestContext upsertRequestCtx)
         private static bool ValidateColumn(ColumnDefinition column,
```
abhishekkumams, line 435: need to update the summary with the new params

src/Core/Services/RequestValidator.cs
```diff
@@ -433,12 +435,18 @@ public void ValidateUpsertRequestContext(UpsertRequestContext upsertRequestCtx)
...
                                            IEnumerable<string> fieldsInRequestBody,
-                                           bool isReplacementUpdate)
+                                           IEnumerable<string> primaryKey,
+                                           bool isReplacementUpdate,
+                                           bool isRequestBodyStrict)
         {
             string message;
-            // Autogenerated values should never be present in a request body.
+            // Autogenerated values should never be present in a request body, unless they are both a primary key
+            // and request-body-strict is false.
```
RubenCerna2079, line 444: Can someone explain what is the purpose of `request-body-strict`? Also why is it necessary that request-body-strict is false for the primary keys that are autogenerated to be present?

aaronburtle: `request-body-strict` is used to indicate whether or not we are being "strict" with what is allowed to be included in the request body when we make a POST PUT or PATCH. Jerry gives an example how of this looks to the user in this issue: https://github.com/Azure/data-api-builder/issues/2463?reload=1?reload=1

But the basic idea is that if you have some object that you want to do one of the above operations with to your DB, the request-body-strict being set to false allows you to do that operation without first removing the primary keys from that object. From the issue linked above:

image

RubenCerna2079: So just to make sure, if `request-body-strict` is `true` then it will not accept any primary key properties, and if it is `false` it should accept them, but it ignores them?

RubenCerna2079 (approved):

src/Core/Services/RequestValidator.cs
aaronburtle: Right, in the case of true youd get an error, and for false it would be ignored and youd get the primary key set by the auto generated incrementation, or if you instead use a primary key path, by using the values in that path.

src/Service.Tests/SqlTests/RestApiTests/RestBodyNonStrictModeTests/MsSqlRestBodyNonStrictModeTests.cs
```diff
@@ -169,6 +174,32 @@ await SetupAndRunRestApiTest(
...
+        public async Task InsertOneWithAutoGeneratedPrimaryKeyInRequestBody()
+        {
+            string requestBody = @"
+            {
+                ""id"": 0,
+                ""title"": ""Foo"",
+                ""publisher_id"": 1234
+            }";
+
+            await SetupAndRunRestApiTest(
```
Aniruddh25, line 191: is the default value of request-body-strict header false?

src/Core/Services/RequestValidator.cs
```diff
@@ -428,17 +430,27 @@ public void ValidateUpsertRequestContext(UpsertRequestContext upsertRequestCtx)
...
             // TODO: Add a similar check for read-only fields.
-            if (column.IsAutoGenerated && fieldsInRequestBody.Contains(exposedName))
+            if (column.IsAutoGenerated &&
+                fieldsInRequestBody.Contains(exposedName) &&
```
Aniruddh25, line 451: why dont we ignore the fields when they are NOT primary and request body is NOT strict? I.e. why dont we apply the same logic of ignoring the fields to all columns irrespective of whether they are a primary key or not?

src/Core/Services/RequestValidator.cs
```diff
@@ -428,17 +430,27 @@ public void ValidateUpsertRequestContext(UpsertRequestContext upsertRequestCtx)
...
+                (!primaryKey.Contains(exposedName) ||
```
Aniruddh25, line 452: comparing primarykey with exposed name doesnt tell you whether the column is primary key or not.. exposed name could be anything...

src/Core/Services/RequestValidator.cs
aaronburtle: I will check with Jerry, the bug that was opened was specific to primary keys, but Ill follow up on that one.

Issue #2452, opened 2024-11-06
Move Global Log-Level


RubenCerna2079:
As of this moment, the log-level property is globally found under `/runtime`, this property should be moved to `/runtime/telemetry`. Once the property is moved and it is shown to be working properly, log-level should be added to the JSON schema.

JerryNixon:
Is there a target milestone?

JerryNixon:
original https://github.com/Azure/data-api-builder/issues/2371

PR #2508, opened 2025-01-07
Move Global Log-Level

RubenCerna2079:
## Why make this change?

Solves #2452

## What is this change?

Moves the log-level property from `/runtime` to `/runtime/telemetry` where it was decided it should live. It also changes the files so that the log-level property is under `TelemetryOptions` instead of `RuntimeOptions` and fix any tests related to the changes.

## How was this tested?

- [X] Integration Tests
- [ ] Unit Tests


src/Service.Tests/Unittests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -273,7 +273,7 @@ public void TestNullableOptionalProps()
             TryParseAndAssertOnDefaults("{" + emptyHostSubProps, out _);
 
             // Test with empty telemetry sub-properties
-            minJsonWithTelemetrySubProps.Append(@"{ ""application-insights"": { } } }");
+            minJsonWithTelemetrySubProps.Append(@"{ ""application-insights"": { }, ""log-level"": { } } }");
```
Aniruddh25, line 276: do we need to modify any other tests that check the log-level value?

Aniruddh25: nvm saw the configuration tests.

Aniruddh25 (approved):
LGTM, lets make sure our documentation is up to date as well.

abhishekkumams (approved):
LGTM, thanks

PR #2510, opened 2025-01-08
Add Support for Hot Reload of GraphQL Schemas

### LINKED ISSUE #2431: Hot Reload - GraphQL Schema Eviction and Regeneration
GraphQL:
- https://github.com/ChilliCream/hotchocolate-examples/tree/master/misc/TypeModules
- https://chillicream.com/blog/2021/09/27/hot-chocolate-12#dynamic-schemas

Example usage to trigger eviction:
https://github.com/ChilliCream/graphql-platform/issues/6113#issuecomment-1609919259

abhishekkumams:
## Why make this change?

- Closes #2431 
- This change is necessary to enable dynamic updates to GraphQL schemas without requiring a full application restart.
- It enhances the developer experience by allowing schema changes to be reflected immediately during runtime, improving efficiency in development and deployment processes.

## What is this change?

- Addition of new constants in `DabConfigEvents.cs` for GraphQL schema events.
- Updates to `HotReloadEventHandler.cs` to include handlers for GraphQL schema-related events.
- Modifications in `RuntimeConfigLoader.cs` to ensure proper firing order of events for schema eviction and creation.
- Updates to `GraphQLSchemaCreator.cs` to handle configuration changes and update runtime entities accordingly.
- Modifications in `Startup.cs` to register hot reload event handlers and methods for refreshing and evicting the GraphQL schema.

## How was this tested?

- [X] Integration Tests
- [X] Manual Tests

## Sample Request(s)

https://github.com/user-attachments/assets/0a2e6353-d466-453e-b1d3-d2f64b3f1d75


src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -57,8 +58,10 @@ public GraphQLSchemaCreator(
             IQueryEngineFactory queryEngineFactory,
             IMutationEngineFactory mutationEngineFactory,
             IMetadataProviderFactory metadataProviderFactory,
-            IAuthorizationResolver authorizationResolver)
+            IAuthorizationResolver authorizationResolver,
+            HotReloadEventHandler<HotReloadEventArgs>? handler = null)
```
RubenCerna2079, line 62: Shouldn't the summary param be updated to include the description of the HotReloadEventHandler?

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -70,6 +73,18 @@ public GraphQLSchemaCreator(
             _runtimeConfigProvider = runtimeConfigProvider;
         }
 
+        /// <summary>
+        /// Executed when a hot-reload event occurs. Pulls the latest
+        /// runtimeconfig object from the provider and updates the flag indicating
+        /// whether multiple create operations are enabled, and the entities based on the new config.
```
RubenCerna2079, line 79: What is the purpose of `multiple create operations`? Since you are checking making the changes before the `entities` is the `multiple create operations` it directly related to it?

src/Service/Startup.cs
```diff
@@ -471,6 +492,15 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, RuntimeC
             });
         }
 
+        /// <summary>
+        /// Evicts the GraphQL schema from the request executor resolver.
```
RubenCerna2079, line 496: Why do we need to evict the graphql schema?

abhishekkumams: it removes the old schema that we built.

src/Core/Services/GraphQLSchemaCreator.cs
abhishekkumams: multiple create operation allows user to insert rows to multiple entities. we are just checking the flag so order of operation not so required.

src/Config/DabConfigEvents.cs
```diff
@@ -15,4 +15,8 @@ public static class DabConfigEvents
     public const string POSTGRESQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED = "POSTGRESQL_QUERY_EXECUTOR_ON_CONFIG_CHANGED";
     public const string DOCUMENTOR_ON_CONFIG_CHANGED = "DOCUMENTOR_ON_CONFIG_CHANGED";
     public const string AUTHZ_RESOLVER_ON_CONFIG_CHANGED = "AUTHZ_RESOLVER_ON_CONFIG_CHANGED";
+    public const string GRAPHQL_SCHEMA_ON_CONFIG_CHANGED = "GRAPHQL_SCHEMA_ON_CONFIG_CHANGED";
+    public const string GRAPHQL_SCHEMA_EVICTION_ON_CONFIG_CHANGED = "GRAPHQL_SCHEMA_EVICTION_ON_CONFIG_CHANGED";
```
Aniruddh25, line 19: good idea to record the video of how its working! Was there an audio component to it as well? if not, may be next time add the audio too.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -100,6 +100,11 @@ protected void SignalConfigChanged(string message = "")
         // RuntimeConfig must already be updated and is implied to have been updated by the time
         // this function is called.
         OnConfigChangedEvent(new HotReloadEventArgs(AUTHZ_RESOLVER_ON_CONFIG_CHANGED, message));
+
+        // Order of event firing matters: Eviction must be done before creating a new schema and then updating the schema.
+        OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_EVICTION_ON_CONFIG_CHANGED, message));
+        OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_CREATOR_ON_CONFIG_CHANGED, message));
+        OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_ON_CONFIG_CHANGED, message));
```
Aniruddh25, line 107: What is the purpose of `GRAPHQL_SCHEMA_ON_CONFIG_CHANGED`?

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -304,6 +320,166 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+        GenerateConfigFile(
+            connectionString: $"{ConfigurationTests.GetConnectionStringFromEnvironmentConfig(TestCategory.MSSQL).Replace("\\", "\\\\")}",
+            gQLEntityEnabled: gQLEntityEnabled);
+        System.Threading.Thread.Sleep(2000);
+
+        // Act
+        HttpResponseMessage gQLResult = await _testClient.SendAsync(request);
+
+        // Assert
+        Assert.AreEqual(HttpStatusCode.InternalServerError, gQLResult.StatusCode);
```
Aniruddh25, line 351: Why would this be internal server error? With the change you made, you would at least have the Publisher entity available right? And the gQLEntityEnabled flag seems to be per entity.. so this shouldnt disable graphql altogether...

Even if it were disabled for all entities, we should give out service not available error instead of internal server error since the latter implies, the service is there but some error happened whereas the former communicated the fact that the graphql service is actually disabled.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -304,6 +320,166 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+        }";
+        payload = new { query = queryWithNewEntity };
+        request = new(HttpMethod.Post, "/graphQL")
+        {
+            Content = JsonContent.Create(payload)
+        };
+        HttpResponseMessage gQLResultWithNewEntity = await _testClient.SendAsync(request);
+
+        // Assert
+        Assert.AreEqual(HttpStatusCode.InternalServerError, gQLResultWithOldEntity.StatusCode);
```
Aniruddh25, line 403: Can you please double check if this is the right error code we should be giving out in case of a missing entity?

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -304,6 +320,166 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
...
+        Assert.AreEqual(HttpStatusCode.OK, gQLResultWithNewEntity.StatusCode);
+        string responseContent = await gQLResultWithNewEntity.Content.ReadAsStringAsync();
+        JsonDocument jsonResponse = JsonDocument.Parse(responseContent);
+        JsonElement items = jsonResponse.RootElement.GetProperty("data").GetProperty("authors").GetProperty("items");
+        string expectedResponse = @"[
+            {
+                ""name"": ""Jelte""
+            }
+        ]";
```
Aniruddh25, line 412:
```suggestion
        ]";
        
```

src/Service/Startup.cs
```diff
@@ -471,6 +492,15 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, RuntimeC
...
+        /// </summary>
+        private static void EvictGraphQLSchema(IRequestExecutorResolver requestExecutorResolver)
+        {
+            Console.WriteLine("Evicting old GraphQL schema.");
+            requestExecutorResolver.EvictRequestExecutor();
```
Aniruddh25, line 501: Great find! To evict, then refresh the schema. This was a difficult issue to resolve!!

Aniruddh25 (approved):
Thank you for this wonderful change! Excellent find, left some comments to address the status codes that we return, looks good otherwise.

src/Config/DabConfigEvents.cs
abhishekkumams: no audio for this.

src/Config/RuntimeConfigLoader.cs
abhishekkumams: makes sense.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
abhishekkumams: updated with asserting error message as well.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
abhishekkumams: good catch. updated with asserting error message as well.

aaronburtle (approved):
Looks good once the questions about internal server error being returned are resolved.

Issue #2526, opened 2025-01-15
[Bug]: Request-Body-Strict not working for calculated and auto-generated fields


aaronburtle:
### What happened?

When request-body strict is set to `false` it should be possible to include calculated or auto-generated fields in the request body of an insert operation. Currently, this leads to an error. This is a logical bug in the `RequestValidator.ValidateColumn` function that needs to be fixed. 

### Version

1.3

### What database are you using?

Azure SQL

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2528, opened 2025-01-15
Update column validation logic for insert operations when considering request-body-strict

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2526

## What is this change?

Update the logic within `RequestValidator.ValidateColumn` to correctly handle the `request-body-strict` setting for calculated and auto-generated fields.

## How was this tested?

Added an additional test to cover auto-generated fields in the request body when `request-body-strict` is set to `false`. This code path is for all read only fields which also covers computed columns, and other columns that will return as read only during startup.

## Sample Request(s)

Can make a post request for any table that has auto-generated or computed columns. For example;

https://localhost:5001/restApi/Book/

with a request body:

{
    "id": 9999,
    "publisher_id": 1234,
    "title": "Foo"
}


abhishekkumams (approved):

sezal98 (commented):
looks good

sezal98 (approved):

src/Service.Tests/SqlTests/RestApiTests/RestBodyNonStrictModeTests/MsSqlRestBodyNonStrictModeTests.cs
```diff
@@ -200,6 +206,33 @@ await SetupAndRunRestApiTest(
...
+
+            await SetupAndRunRestApiTest(
+                    primaryKeyRoute: null,
+                    queryString: null,
+                    entityNameOrPath: _integration_AutoGenNonPK_EntityName,
+                    sqlQuery: GetQuery("InsertOneWithAutoGeneratedFieldInRequestBody"),
+                    operationType: EntityActionOperation.Insert,
+                    requestBody: requestBody,
+                    expectedStatusCode: HttpStatusCode.Created,
+                    expectedLocationHeader: string.Empty
```
Aniruddh25, line 232: Wouldnt the expected location header be the location of the newly inserted record?

Aniruddh25:
> This code path is for all read only files

What do you mean by files here?

Aniruddh25 (approved):
Approving it assuming test will pass and LocationHeader is really not expected.

aaronburtle:
> > This code path is for all read only files
> 
> What do you mean by files here?

typo, meant to be fields, fixed.

src/Service.Tests/SqlTests/RestApiTests/RestBodyNonStrictModeTests/MsSqlRestBodyNonStrictModeTests.cs
aaronburtle: Yes it would be, however, we leave this blank for all the tests in this test class because we are just testing the validation process for the request body when we have request-body-strict set to false.

Issue #2527, opened 2025-01-15
[Bug]: Missing fields in body of request can lead to SQL syntax error


aaronburtle:
### What happened?

When a field is missing from the body for a request that maps to insert operations this should either result in a working request where DAB is able to determine the field in question and assign insert something like a default value, or it should result in error messaging that is clear to the user. Currently we are instead running into a SQL Syntax exception.

![Image](https://github.com/user-attachments/assets/31a2a998-870d-4906-b4fe-0ec3b0dfca56)

### Version

1.3

### What database are you using?

Azure SQL

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2530, opened 2025-01-23
Fix SQL Syntax when doing a valid insert with request body empty

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2527

## What is this change?

We introduce new SQL text for PostgreSQL and MSSQL when the collection of insertion columns is empty, when doing an insert. MSSQL and PostgreSQL provide key words DEFAULT VALUES in order to handle this case, and MySQL uses an empty VALUES () text that we already generate.

## How was this tested?

We add a new test to MSSQL and PostreSQL called InsertOneWithDefaultValuesAndEmptyRequestBody which makes a simple insertion without a request body, we then validate that the response is as expected.

## Sample Request(s)

Any valid POST request to a table that is made up of columns that all have some kind of default value, with an empty request body.

ie:

{POST}https://localhost:5001/restApi/Book/

Request Body:

{

}


src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -387,6 +388,11 @@ CREATE TABLE user_profiles (
 	userid INT
 );
 
+CREATE TABLE default_books(
```
abhishekkumams, line 391: Are we not updating MySql?

aaronburtle: MySQL for default values just uses an empty values () statement, which is what our code is already doing.

abhishekkumams: That's great. we should still add the tests for mysql as well to make sure they're working fine.

src/Service.Tests/SqlTests/RestApiTests/Insert/MsSqlInsertApiTests.cs
```diff
@@ -367,6 +372,26 @@ await SetupAndRunRestApiTest(
                 );
         }
 
+        [TestMethod]
```
abhishekkumams, line 375: need test summary

abhishekkumams (approved):

rohkhann (approved):

PR #2536, opened 2025-01-23
Fixed the timers to track proper DB execution duration. 

lingxiao-microsoft:
## Why make this change?
Previously we created a PR to add timers to track the duration of DB execution. However, there are some issues within that results in incorrect time duration logged. 

Please exclude the whitespaces diff to have a better reviewing experience
image

1. Previously we added timer in [ExecuteQueryAsync](https://github.com/Azure/data-api-builder/blob/b95b95f3b38f959310338202a9de67a7cdf06761/src/Core/Resolvers/QueryExecutor.cs#L175) and [ExecuteQueryAgainstDbAsync](https://github.com/Azure/data-api-builder/blob/b95b95f3b38f959310338202a9de67a7cdf06761/src/Core/Resolvers/QueryExecutor.cs#L310C4-L312C1)
2.  However, within function `ExecuteQueryAsync`, it will invoke `ExecuteQueryAgainstDbAsync` .
3. Which will result in logging the time duration twice. 
4.  See repro below 
image

We can see the stored value is always twice as the stop watch
image
image

## What is this change?

- Fixing the timer to log properly
  - Removed the timer from upper function `ExecuteQueryAsync` and `ExecuteQuery`
  - Put the timer inside `ExecuteQueryAgainstDbAsync` and `ExecuteQueryAgainstDb`
  - Added `try-finally` block to wrap the functions. 

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

## Sample Request(s)

image
  - Putting a breakpoint at the `ExecuteQueryAsync` level
  - Time durations are matched now
image


JerryNixon:
@sezal98

rohkhann (approved):
Thanks for the fix!

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -112,61 +112,51 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
...
+            result = _retryPolicy.Execute(() =>
             {
-                result = _retryPolicy.Execute(() =>
+                retryAttempt++;
+                try
                 {
-                    retryAttempt++;
-                    try
+                    // When IsLateConfigured is true we are in a hosted scenario and do not reveal query information.
+                    if (!ConfigProvider.IsLateConfigured)
```
JerryNixon, line 123: If `IsLateConfigured` is how we determine we are in a hosted scenario, you should create a var called `IsHostedScenario` and determine its boolean value from whatever logic makes sense. Then use that var as the predicate. This is far better than using a comment to convey this information. It also allows you mod the logic in the future without letting comments slip out of date - as we see in other places in the code base. 

Is `IsHostedScenario` logic useful in other parts of the code? Raise the scope of the var to a higher area where it can be reused. This is important to help ensure the logic determining the hosting scenario (is scenario the right word?) is not out of sync in different locations. A one-off determination of this can create a dangerous maintenance issue later.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -112,61 +112,51 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
...
-                            string correlationId = HttpContextExtensions.GetLoggerCorrelationId(httpContext);
-                            int maxRetries = _maxRetryCount + 1;
-                            // This implies that the request got successfully executed during one of retry attempts.
-                            QueryExecutorLogger.LogInformation("{correlationId} Request executed successfully in {retryAttempt} attempt of {maxRetries} available attempts.", correlationId, retryAttempt, maxRetries);
-                        }
+                    if (retryAttempt > 1)
+                    {
+                        string correlationId = HttpContextExtensions.GetLoggerCorrelationId(httpContext);
+                        int maxRetries = _maxRetryCount + 1;
+                        // This implies that the request got successfully executed during one of retry attempts.
```
JerryNixon, line 135: This comment seems to address the logging line below it, but I believe it was intended to describe the overall if-then statement. While moving the comment to the relevant code is the minimum fix, I recommend creating a variable called `successAfterRetry` and assigning it a boolean value based on the appropriate logic. Then, use this variable as the predicate for the if-then clause. 

This change eliminates the need for the comment, ensures the logic is self-explanatory, and prevents the comment from being misplaced or becoming outdated, as seen elsewhere in the codebase. This simple refactoring would greatly enhance the maintainability of the code.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -279,39 +259,42 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
...
-                    result = await dataReaderHandler(dbDataReader, args);
+                    using DbDataReader dbDataReader = ConfigProvider.GetConfig().MaxResponseSizeLogicEnabled() ?
+                        await cmd.ExecuteReaderAsync(CommandBehavior.SequentialAccess) : await cmd.ExecuteReaderAsync(CommandBehavior.CloseConnection);
+                    if (dataReaderHandler is not null && dbDataReader is not null)
+                    {
+                        result = await dataReaderHandler(dbDataReader, args);
+                    }
+                    else
+                    {
+                        result = default(TResult);
```
JerryNixon, line 277: This `else` statement sets `result` to its initialization value, which is redundant since `result` already retains that value. The `else` block is unnecessary in this situation and can be safely removed without affecting the logic.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -112,61 +112,51 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
...
+                        // This implies that the request got successfully executed during one of retry attempts.
+                        QueryExecutorLogger.LogInformation("{correlationId} Request executed successfully in {retryAttempt} attempt of {maxRetries} available attempts.", correlationId, retryAttempt, maxRetries);
+                    }
 
-                        return result;
+                    return result;
+                }
+                catch (DbException e)
+                {
+                    if (DbExceptionParser.IsTransientException((DbException)e) && retryAttempt < _maxRetryCount + 1)
```
JerryNixon, line 143: Consider a comment for this if statement if you are not going to create a boolean variable to self-document it. "If the exception is transient and retry attempts are less than the maximum allowed retries."

src/Core/Resolvers/QueryExecutor.cs
lingxiao-microsoft: Hello @JerryNixon , thank you for time on this PR. 

For the comments you made, these changes are marked as "diff" by Git was due to the whitespaces. None of these logics are changed by this PR. 

Please kindly use this setting while reviewing the PR which could be more helpful! 

image

src/Core/Resolvers/QueryExecutor.cs
lingxiao-microsoft: @JerryNixon please kindly checkout my reply here, the highlighted areas are caused by whitespaces. https://github.com/Azure/data-api-builder/pull/2536#discussion_r1931297016

src/Core/Resolvers/QueryExecutor.cs
lingxiao-microsoft: ^^ Same as above

src/Core/Resolvers/QueryExecutor.cs
lingxiao-microsoft: ^^ Same as above

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -112,61 +112,51 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
 
             SetManagedIdentityAccessTokenIfAny(conn, dataSourceName);
 
-            Stopwatch queryExecutionTimer = new();
-            queryExecutionTimer.Start();
             TResult? result = default(TResult?);
 
-            try
```
Aniruddh25, line 119: Would removing the `try` clause here end up having sideeffects of not being able to catch some exceptions that we were catching earlier?

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -355,22 +360,36 @@ Mock<MsSqlQueryExecutor> queryExecutor
...
+            }
+            catch (Exception)
+            {
+                // as the SqlConnection object is a sealed class and can't be mocked, ignore any exceptions caused to bypass
+            }
+
+            stopwatch.Stop();
+
+            Assert.IsTrue(context.Items.ContainsKey(TOTAL_DB_EXECUTION_TIME), "HttpContext object must contain the total db execution time after execution of a query");
+            Assert.IsTrue(stopwatch.ElapsedMilliseconds >= (long)context.Items[TOTAL_DB_EXECUTION_TIME], "The execution time stored in http context must be valid.");
```
Aniruddh25, line 392: How can we assert that its not counted twice? Perhaps add a condition:

ElapsedMilliseconds < 2*context.Items[Total_DB_Execution_Time] ?

Aniruddh25 (approved):
Looks good overall! Left a few comments.

src/Core/Resolvers/QueryExecutor.cs
lingxiao-microsoft: Great question! However the `try` on line 119 does not have a `catch` and only has a `finally` instead on line 165. 

Hence removing it would not have effect on the original `try-catch` pair on line 124 & 145.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
lingxiao-microsoft: Here the condition `stopwatch.ElapsedMilliseconds >= (long)context.Items[TOTAL_DB_EXECUTION_TIME` aims to be more generic that indicates the inner function timers would not exceed the upper function timer. 

And the counted twice scenarios are covered by it as well

abhishekkumams (approved):

PR #2541, opened 2025-01-25
GroupBy type creation and updating connection object with groupby type.

rohkhann:
## Why make this change?
This pr makes the following changes:
1. Converts the input type objects for aggregations input to enum type to enable those enums to be used as output as well.
2. Adds in groupby type.

These changes set up the schema appropriately to move to resolver phase of the execution.

## What is this change?
- `GenerateReturnType` method in `QueryBuilder` updated to include aggregation fields if `_isAggregationEnabled` is true. This creates the Connections object and only includes the  groupby field if aggregation is present.
- Added temporary default `_groupByFieldResolver` to `ResolverTypeInterceptor` to handle `groupBy` operations, which will throw `NotSupportedException`. The resolver functionality will be added in following prs.
- `ResolverTypeInterceptor` updated to block `groupBy` operations at the root query level and set `_groupByFieldResolver` for `GroupBy` type fields.

## How was this tested?
Unit tests added for all code changes
Query creation experience:
1. Regular select * query:
image
2. Group by/aggregation experience:
image
3.Groupby with aggregation added

image

### DIFF
--- a/src/Core/Services/GraphQLSchemaCreator.cs
+++ b/src/Core/Services/GraphQLSchemaCreator.cs
@@ -189,6 +189,8 @@ namespace Azure.DataApiBuilder.Core.Services
             // (relationship/column) and non-relationship fields from the linking table.
             Dictionary<string, ObjectTypeDefinitionNode> objectTypes = new();
 
+            Dictionary<string, EnumTypeDefinitionNode> enumTypes = new();
+
             // 1. Build up the object and input types for all the exposed entities in the config.
             foreach ((string entityName, Entity entity) in entities)
             {
@@ -241,12 +243,15 @@ namespace Azure.DataApiBuilder.Core.Services
 
                             if (_runtimeConfigProvider.GetConfig().EnableAggregation)
                             {
-                                InputTypeBuilder.GenerateAggregationNumericInputForObjectType(node, inputObjects);
+                                EnumTypeBuilder.GenerateAggregationNumericEnumForObjectType(node, enumTypes);
+                                EnumTypeBuilder.GenerateScalarFieldsEnumForObjectType(node, enumTypes);
                                 // Generate aggregation type for the entity
-                                ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(entityName, node);
+                                ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(node.Name.Value, node);
                                 if (aggregationType.Fields.Any())
                                 {
+                                    Console.WriteLine($"Aggregation type generated for {node.Name.Value}");
                                     objectTypes.Add(SchemaConverter.GenerateObjectAggregationNodeName(entityName), aggregationType);
+                                    objectTypes.Add(SchemaConverter.GenerateGroupByTypeName(entityName), SchemaConverter.GenerateGroupByTypeForEntity(node.Name.Value, node));
                                 }
                             }
                         }
@@ -304,6 +309,7 @@ namespace Azure.DataApiBuilder.Core.Services
                 fields.Values.ToImmutableList()));
 
             List<IDefinitionNode> nodes = new(objectTypes.Values);
+            nodes.AddRange(enumTypes.Values);
             return new DocumentNode(nodes);
         }
 
--- a/src/Core/Services/GraphQLSchemaCreator.cs
+++ b/src/Core/Services/GraphQLSchemaCreator.cs
@@ -44,6 +44,7 @@ namespace Azure.DataApiBuilder.Core.Services
         private readonly IAuthorizationResolver _authorizationResolver;
         private readonly RuntimeConfigProvider _runtimeConfigProvider;
         private bool _isMultipleCreateOperationEnabled;
+        private bool _isAggregationEnabled;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="GraphQLSchemaCreator"/> class.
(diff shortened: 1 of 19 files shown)


src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -239,14 +244,17 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
...
+                                EnumTypeBuilder.GenerateAggregationNumericEnumForObjectType(node, enumTypes);
+                                EnumTypeBuilder.GenerateScalarFieldsEnumForObjectType(node, enumTypes);
                                 // Generate aggregation type for the entity
-                                ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(entityName, node);
+                                ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(node.Name.Value, node);
                                 if (aggregationType.Fields.Any())
                                 {
+                                    Console.WriteLine($"Aggregation type generated for {node.Name.Value}");
                                     objectTypes.Add(SchemaConverter.GenerateObjectAggregationNodeName(entityName), aggregationType);
+                                    objectTypes.Add(SchemaConverter.GenerateGroupByTypeName(entityName), SchemaConverter.GenerateGroupByTypeForEntity(node.Name.Value, node));
```
lingxiao-microsoft, line 257: Are we going to have two Node type - `Aggregation` and `GroupBy`? Are they both sub-part of Query node?

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -239,14 +244,17 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
                         {
                             InputTypeBuilder.GenerateInputTypesForObjectType(node, inputObjects);
 
-                            if (_runtimeConfigProvider.GetConfig().EnableAggregation)
+                            if (_isAggregationEnabled)
                             {
-                                InputTypeBuilder.GenerateAggregationNumericInputForObjectType(node, inputObjects);
+                                EnumTypeBuilder.GenerateAggregationNumericEnumForObjectType(node, enumTypes);
+                                EnumTypeBuilder.GenerateScalarFieldsEnumForObjectType(node, enumTypes);
                                 // Generate aggregation type for the entity
```
lingxiao-microsoft, line 251: Nit: Still having some troubles understanding the usage of EnumType here, are they used to 
1. Populate the numeric aggregation or scalar groupBy type name based on the schema
2. Then used to generate the node `Aggregation` under the query node?

Would appreciate it if we can elaborate bit here to help me understand

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: It would be in screenshot like above. A groupby type added to the return object of a query node today. This one would then have aggregations within it.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: This is how the sample structure would look like:
 type BookAggregations {
    max(field: BookNumericFields, having: HavingInput, distinct: Boolean): Float
    min(field: BookNumericFields, having: HavingInput, distinct: Boolean): Float
    avg(field: BookNumericFields, having: HavingInput, distinct: Boolean): Float
    sum(field: BookNumericFields, having: HavingInput, distinct: Boolean): Float
    count(field: BookNumericFields, having: HavingInput, distinct: Boolean): Int
    mode(field: BookNumericFields, having: HavingInput, distinct: Boolean): Float
    mean(field: BookNumericFields, having: HavingInput, distinct: Boolean): Float
}

type BooksConnection {
    items: [Book]
    aggregations: BookAggregations
    groupBy(fields: [BookFields], orderBy: [OrderByInput]): [BookGroupBy]
}

enum BookFields {
    ID
    TITLE
    AUTHOR
    AGE
    RATING
}

enum BookNumericFields {
    ID
    AGE
    RATING
}
Enum is used to define the input to the aggregations object. We need two enums, one for all the scalar fields in books so that they can be used in the groupby call and another enum with numeric fields that has id age and rating that would be the numeric fields that go as input to the aggregations.

lingxiao-microsoft (approved):

src/Service.GraphQLBuilder/Queries/QueryBuilder.cs
```diff
@@ -239,15 +242,11 @@ public static bool IsPaginationType(NamedTypeNode objectType)
             return objectType.Name.Value.EndsWith(PAGINATION_OBJECT_TYPE_SUFFIX);
         }
 
-        public static ObjectTypeDefinitionNode GenerateReturnType(NameNode name)
+        public static ObjectTypeDefinitionNode GenerateReturnType(NameNode name, bool aggregationEnabled = false)
```
abhishekkumams, line 245:
```suggestion
        public static ObjectTypeDefinitionNode GenerateReturnType(NameNode name, bool isAggregationEnabled = false)
```

abhishekkumams (approved):
LGTM

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -239,14 +244,17 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
...
-                                ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(entityName, node);
+                                ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(node.Name.Value, node);
```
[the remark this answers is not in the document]
rohkhann: No this should not be null as it would be the name of the node in the graphql schema.

PR #2548, opened 2025-01-31
Cosmos DB: Improve Error message During Schema Generation

sourabh1007:
## Why make this change?

This PR enhances user experience by providing more informative error and warning messages in case of misconfigurations.

**Changes:**
1. **Improved Missing Flag Handling-** Added an ERROR log when `--graphql` is not passed, preventing a generic failure message and clearly informing users that `--graphql` is required.
2. **Better Schema Validation-** If the generated schema is empty, an ERROR log now states: _"Generated GraphQL schema is empty. Please check if you have data to generate the schema."_ This provides clearer guidance instead of a vague failure message.
3. **Handling Missing Entities in Config-** An exception is now thrown when no entities are configured in the config file, ensuring this is logged for better debugging.
4. **Logging for Empty Data-** Added an ERROR log if data is missing and an empty array is received for schema generation.
5. **Warnings for Skipped Entities-** Introduced a WARNING log if a specific entity is missing from the config while processing. This helps users understand why an entity was skipped in schema generation.
6. **Test Fixes-** Updated and fixed tests to align with the new error-handling improvements.

## What is this change?

Added more error message.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

### DIFF
--- a/src/Cli/Exporter.cs
+++ b/src/Cli/Exporter.cs
@@ -119,7 +119,7 @@ namespace Cli
 
             if (string.IsNullOrEmpty(schemaText))
             {
-                logger.LogError("Generated GraphQL schema is empty. Please check if you have data to generate the schema out of it.");
+                logger.LogError("Generated GraphQL schema is empty. Please ensure data is available to generate the schema.");
                 return;
             }
 
--- a/src/Cli/Exporter.cs
+++ b/src/Cli/Exporter.cs
@@ -80,6 +80,10 @@ namespace Cli
                     logger.LogError("Failed to export GraphQL schema.");
                 }
             }
+            else
+            {
+                logger.LogError("Exporting GraphQL schema is not enabled. You need to pass --graphql.");
+            }
 
             _cancellationTokenSource.Cancel();
             return isSuccess;
@@ -115,7 +119,7 @@ namespace Cli
 
             if (string.IsNullOrEmpty(schemaText))
             {
-                logger.LogError("Failed to export the GraphQL schema.");
+                logger.LogError("Generated GraphQL schema is empty. Please check if you have data to generate the schema out of it.");
                 return;
             }
 
--- a/src/Core/Generator/SchemaGenerator.cs
+++ b/src/Core/Generator/SchemaGenerator.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Text.Json;
 using Azure.DataApiBuilder.Config.ObjectModel;
 using Humanizer;
+using Microsoft.Extensions.Logging;
 using static System.Text.Json.JsonElement;
 
 namespace Azure.DataApiBuilder.Core.Generator
@@ -18,6 +19,8 @@ namespace Azure.DataApiBuilder.Core.Generator
     {
         // Azure Cosmos DB reserved properties, these properties will be ignored in the schema generation as they are not user-defined properties.
         private readonly List<string> _cosmosDbReservedProperties = new() { "_ts", "_etag", "_rid", "_self", "_attachments" };
+
+        private readonly ILogger? _logger;
 
         // Maps GraphQL entities to their corresponding attributes.
         private Dictionary<string, HashSet<AttributeObject>> _attrMapping = new();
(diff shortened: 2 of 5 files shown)


src/Cli/Exporter.cs
```diff
@@ -80,6 +80,10 @@ public static bool Export(ExportOptions options, ILogger logger, FileSystemRunti
                     logger.LogError("Failed to export GraphQL schema.");
                 }
             }
+            else
+            {
+                logger.LogError("Exporting GraphQL schema is not enabled. You need to pass --graphql.");
```
Aniruddh25, line 85: why do we need to pass the `--graphql` option if export is by default always to export the GraphQL schema as per the description? Shouldnt remove the `if` check altogether?

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -35,19 +38,25 @@ internal class SchemaGenerator
...
         {
+            this._logger = logger;
+
             this._data = data;
             this._containerName = containerName;
             if (config != null)
             {
+                if (config.Entities == null || config.Entities.Count() == 0)
+                {
+                    throw new Exception("Define one or more entities to generate the schema in the config file. ");
```
Aniruddh25, line 52:
```suggestion
                    throw new Exception("Define one or more entities in the config file to generate the GraphQL schema.");
```

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -165,6 +189,7 @@ private void TraverseJsonObject(JsonDocument jsonObject, string parentType)
                 // Check if the parent type is not in the entity mapping.
                 if (_entityAndSingularNameMapping.Count != 0 && !_entityAndSingularNameMapping.ContainsKey(parentType.Pascalize()))
                 {
+                    _logger?.LogWarning($"{parentType.Pascalize()} is not available, If it is unexpected, add an entry of it, in the config file.");
```
RubenCerna2079, line 192:
```suggestion
                    _logger?.LogWarning($"{parentType.Pascalize()} is not available. If it is unexpected, add its entry in the config file.");
```

Aniruddh25, line 192:
```suggestion
                    _logger?.LogWarning($"{parentType.Pascalize()} is not available, If it is unexpected, add an entity of it, in the config file.");
```

src/Cli/Exporter.cs
```diff
@@ -115,7 +119,7 @@ private static async Task ExportGraphQL(ExportOptions options, RuntimeConfig run
 
             if (string.IsNullOrEmpty(schemaText))
             {
-                logger.LogError("Failed to export the GraphQL schema.");
+                logger.LogError("Generated GraphQL schema is empty. Please check if you have data to generate the schema out of it.");
```
RubenCerna2079, line 122:
```suggestion
                logger.LogError("Generated GraphQL schema is empty. Please ensure data is available to generate the schema.");
```

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
```diff
@@ -227,13 +227,12 @@ public void TestMixedJsonArray()
...
         /// Tests the <see cref="SchemaGenerator.Generate"/> method with an empty JSON array.
-        /// Ensures that the method correctly handles an empty input by throwing an appropriate exception.
         /// </summary>
         [TestMethod]
         public void TestEmptyJsonArray()
         {
             List<JsonDocument> jsonArray = new();
-            Assert.ThrowsException<InvalidOperationException>(() => SchemaGenerator.Generate(jsonArray, "containerName"));
+            Assert.AreEqual(string.Empty, SchemaGenerator.Generate(jsonArray, "containerName"));
         }
```
RubenCerna2079, line 236: What is the purpose of this test if it doesn't need to check if there are any appropriate exceptions to throw?

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -144,6 +163,11 @@ private string GenerateGQLSchema()
                 sb.AppendLine("}");
             }
 
+            if (sb.Length == 0)
+            {
+                _logger?.LogWarning("Generated Schema is empty.");
+            }
```
RubenCerna2079, line 169:
```suggestion
            {
                _logger?.LogWarning("Generated GraphQL schema is empty.");
            }
```

src/Core/Generator/SchemaGeneratorFactory.cs
```diff
@@ -122,15 +122,13 @@ public static async Task<string> Create(RuntimeConfig config, string mode, int?
                 List<JsonDocument> dataArray = await schemaGeneratorSampler.GetSampleAsync();
 
                 logger.LogInformation("{0} records collected as Sample", dataArray.Count);
-                if (dataArray.Count == 0)
```
Aniruddh25, line 125: Do we no longer error out when no data is present?

Aniruddh25 (changes requested):
Need to still error out when data is empty..

src/Core/Generator/SchemaGenerator.cs
sourabh1007: addressed..in annirudh comment.

src/Service.Tests/CosmosTests/SchemaGeneratorTest.cs
sourabh1007: it will tell, what is the expectation with generator in case data is not there.

src/Core/Generator/SchemaGeneratorFactory.cs
sourabh1007: we will error in later stage, when processing has been done for all the containers. ref. https://github.com/Azure/data-api-builder/blob/582a1156dae066ae98ca71319c6915f307b9689a/src/Cli/Exporter.cs#L120

src/Cli/Exporter.cs
sourabh1007: This is existing code, so I didn't want to modify it; I only added the error message. Passing this flag has its own benefits, as it ensures that customers have the right expectations when running this command, given that we support both REST and GraphQL APIs.

Aniruddh25 (approved):
Thanks for addressing the comments!

abhishekkumams (approved):

PR #2553, opened 2025-02-07
Adds Query-Identifying IDs to HTTPContext for SQL Queries 

gilemos:
## Why make this change?

- We want to be able to associate the statement ID from the SQL query execution call to the DAB response
- We are now retrieving the statement ID whenever we make a SQL query execution call and adding it to the httpcontext of our DAB response.
- This change will help DAB users better track DAB performance - by having access to the statement ID, users can find performance-related telemetry on the SQL database side, and therefore understand how much time was spent on the database side vs DAB code. 

## What is this change?

- We are creating a new virtual method in QueryExecutor.cs that contains the logic to create a connection to the data source. This method is overridden in MsSqlQueryExecutor.cs to include an info message handler that adds the query identifying IDs to httpcontext.

## Scope of this PR
This change should add the statement ID to the httpcontext for the following Fabric data sources: SQLDB, Lakehouse and Data Warehouse.

We are still investigating how to enable that for Fabric Datamart, Azure SQLDB, and MySql and PgSql databases. Those changes - if needed - will come in a separate PR

## How was this tested?

- [X] Integration Tests
- [X] Unit Tests
- [X] Local testing for Fabric data sources - connected DAB to a Fabric data source and verified that the query succeeds and that the statement id was added: works for SQLDB, Lakehouse, Data Warehouse. For Datamart, the query succeeds as expected, but no statement ID is returned -- we believe that is a limitation from the Datamart side.
- [X] Local testing - connecting DAB to Azure SQLDB and testing if query succeeds.

## Sample Request(s)

Screenshot of the httpcontext item added under the header "QueryIdentifyingIds"
image


src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -198,6 +200,48 @@ private bool IsDefaultAccessTokenValid()
             return _defaultAccessToken?.Token;
         }
 
+        /// <inheritdoc/>
+        public override TResult ExecuteQuery<TResult>(
```
lingxiao-microsoft, line 204: One concern I have is we're making changes to these fundamental functions that we might have less context on, which seems bit risky. 

Is it possible that, we can have a new virtual function in the `QueryExecutor` to build the connection. 

For `MsSqlQueryExecutor` we can override it and use your CreateDbConnection to add the statement id, but for rest of the Executor we can fall back to an original default one. 

In that case, you probably would not need to make these adjustments. Thinking out loud here

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -365,5 +409,73 @@ public static void PopulateDbTypeForParameter(KeyValuePair<string, DbConnectionP
...
+                // This helps in correlating with SQL engine telemetry.
+
+                // If the info message has an error code that matches the well-known codes used for returning statement ID,
+                // then we can be certain that the message contains no PII.
+                IEnumerable<SqlError> errorsReceived = e.Errors.Cast<SqlError>();
+
+                IEnumerable<SqlInformationalCodes> allInfoCodesKnown = Enum.GetValues(typeof(SqlInformationalCodes)).Cast<SqlInformationalCodes>();
+
+                IEnumerable<string> infoErrorMessagesReceived = errorsReceived.Join(allInfoCodesKnown, error => error.Number, code => (int)code, (error, code) => error.Message);
+
```
lingxiao-microsoft, line 450: Thanks for making this finally happen! 

Can we please get some screenshots on how these errors/information would look like together with the statement id?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -365,5 +409,73 @@ public static void PopulateDbTypeForParameter(KeyValuePair<string, DbConnectionP
...
+                {
+                    // Add statement ID to request
+                    AddStatementIDToMiddlewareContext(infoErrorMessageReceived);
+                }
+            };
+
+            return conn;
+        }
+
+        private void AddStatementIDToMiddlewareContext(string statementId)
```
lingxiao-microsoft, line 461: Nit: Not related to the PR but more of future work. 

How are we going to use this statement id for analyzing the performance? Do we just need it to engage data source teams

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: That is a great idea! I'll work on that :D

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: Yes, it will be very useful to engage the data source teams. We can also use it to look up data source query performance info from telemetry ourselves

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: Done!

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: Added the screenshot to the PR description :)
I am adding the statement ID, distributed request ID, and query hash - I am not sure if they could show up in a different format/sequence depending on the data source, so I decided to keep all the values. Lemme know what you think!

Aniruddh25:
Please update the PR title to reflect what your change plans to do. Also, fix the PR description with sample request etc or remove text thats not needed from the template.

amitabhiiita (approved):

rohkhann (approved):
LGTM

src/Core/Models/SqlInformationalCodes.cs
```diff
@@ -0,0 +1,26 @@
...
+    public enum SqlInformationalCodes : int
+    {
+
+        /// <summary>
+        /// Prints statement ID, query hash and distributed request ID for development purposes.
+        /// </summary>
+        /// <remarks>
+        /// Text in summary is copied verbatim from the engine definition.
+        /// </remarks>
+        POLARIS_QUERY_IDENTIFIER = 15806,
```
Aniruddh25, line 15: nit: Rename Polaris to use a public facing name

src/Core/Models/SqlInformationalCodes.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Prints statement ID for supportability purposes. If a non-distributed query hits an issue, it'd be helpful if the customer could share the statement id.
+        /// </summary>
+        /// <remarks>
+        /// Text in summary is copied verbatim from the engine definition.
+        /// </remarks>
+        DW_TRIDENT_QUERY_IDENTIFIER = 24528,
```
Aniruddh25, line 23: nit: use Fabric, not Trident

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -237,6 +222,27 @@ public QueryExecutor(DbExceptionParser dbExceptionParser,
...
+        /// Creates and setups a TConnection to the data source of given name
+        /// </summary>
+        /// <param name="dataSourceName">The data source name</param>
+        /// <returns>A connection to the data source</returns>
+        /// <exception cref="DataApiBuilderException">Exeption thrown if the data source could not be found</exception>
+        public virtual TConnection CreateConnection(string dataSourceName)
+        {
+            if (!ConnectionStringBuilders.ContainsKey(dataSourceName))
+            {
+                throw new DataApiBuilderException("Query execution failed. Could not find datasource to execute query against", HttpStatusCode.BadRequest, DataApiBuilderException.SubStatusCodes.DataSourceNotFound);
```
Aniruddh25, line 235: nit: keep 100 char limit, split the line in multiple and honor the indentation

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -365,5 +410,25 @@ public static void PopulateDbTypeForParameter(KeyValuePair<string, DbConnectionP
...
+        {
+            HttpContext? httpContext = HttpContextAccessor?.HttpContext;
+            if (httpContext != null)
+            {
+                // locking is because we could have multiple queries in a single http request and each query will be processed in parallel leading to concurrent access of the httpContext.Items.
+                lock (_httpContextLock)
+                {
+                    if (httpContext.Items.TryGetValue(QUERYIDHEADER, out object? currentValue) && currentValue is not null)
+                    {
+                        httpContext.Items[QUERYIDHEADER] = (string)currentValue + ";" + statementId;
```
Aniruddh25, line 424: what if the currentValue object cannot be casted to a string?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -83,6 +85,49 @@ public MsSqlQueryExecutor(
...
+            {
+                throw new DataApiBuilderException("Query execution failed. Could not find datasource to execute query against", HttpStatusCode.BadRequest, DataApiBuilderException.SubStatusCodes.DataSourceNotFound);
+            }
+
+            SqlConnection conn = new()
+            {
+                ConnectionString = ConnectionStringBuilders[dataSourceName].ConnectionString,
+            };
+
+            // If connection is a SQLConnection and it is not null, we can extract the info message
```
Aniruddh25, line 107: where do we check if conn is not null?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -83,6 +85,49 @@ public MsSqlQueryExecutor(
...
+            conn.InfoMessage += (object sender, SqlInfoMessageEventArgs e) =>
+            {
+                // Log the statement ids returned by the SQL engine when we executed the batch.
+                // This helps in correlating with SQL engine telemetry.
+
+                // If the info message has an error code that matches the well-known codes used for returning statement ID,
+                // then we can be certain that the message contains no PII.
+                IEnumerable<SqlError> errorsReceived = e.Errors.Cast<SqlError>();
```
Aniruddh25, line 115: The same executor is used for SQL Server on premises, can we please look make sure at least all flavors of mssql have this change. 

As a followup change, I'd expect PgSql, Mysql QueryExecutors implement this function as well to complete the feature - otherwise its a broken functionality

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
```diff
@@ -279,6 +282,10 @@ Mock<MsSqlQueryExecutor> queryExecutor
                 = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object, handler);
 
             queryExecutor.Setup(x => x.ConnectionStringBuilders).Returns(new Dictionary<string, DbConnectionStringBuilder>());
+
```
Aniruddh25, line 285: Can you please add a test here similar to TestHttpContextIsPopulatedWithDbExecutionTime? 

TestHttpContextIsPopulatedWithStatementId for e.g.

Aniruddh25 (changes requested):
Waiting on making sure MsSql Information Codes are handled as well..

src/Core/Resolvers/QueryExecutor.cs
gilemos: Done!

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: Good point! That was an outdated comment. We create the connection right before this line, so the connection should not be null

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: Added a catch block! If there is an error in casting, we won't add the value to httpcontext.

src/Core/Models/SqlInformationalCodes.cs
gilemos: Done!

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: Discussed on teams. 
Created an action item to follow up on the implementations for PgSql and Mysql QueryExecutors in a separate PR.

Regarding the first part of this comment, please let me know if I missed any other mssql query executor file.

src/Service.Tests/Unittests/SqlQueryExecutorUnitTests.cs
gilemos: Created a test that checks whether an info message handler was added to the connection.

I think it would be a bit tricky to test if we are actually getting the statement ID in a unit test, since we don't actually make a request to the data source - I am not sure on how to mock a database response that would trigger the event handler. 

However, I have tested this behavior when running DAB locally.

Let me know if you think the testing is sufficient and whether you have suggestions on this!

src/Core/Resolvers/MsSqlQueryExecutor.cs
Aniruddh25: What if the connection creation fails? Do we have enough safeguards against that to handle gracefully?

Aniruddh25 (approved):
Looks good, just left some questions to double check.

gilemos:
Testing the changes with an Azure SQLDB!

src/Core/Resolvers/MsSqlQueryExecutor.cs
gilemos: Added some more checks to:
- Throw an exception (with a helpful error message) in case the connection is null
- Log an error but continue the execution in case there was an exception thrown in the logic that adds the statement ID to httpcontext

Aniruddh25 (approved):

PR #2567, opened 2025-02-13
Adding version check to startup

JerryNixon:
Closes #3292

## Why make this change?

It is easy to continue running an outdated version of Data API builder after a newer version has been released to NuGet. This PR updates the startup process, checking the current version against the latest version available on NuGet.

## What is this change?

1. Introducing `Azure.DataApiBuilder.Product.VersionChecker`.
2. Validating the versions during `Program.Main()`.
3. Write out a log message only; this does NOT exit the engine. 

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

### DIFF
--- a/src/Product/VersionChecker.cs
+++ b/src/Product/VersionChecker.cs
@@ -1,0 +1,51 @@
+using System;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Reflection;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Product;
+
+public static class VersionChecker
+{
+    private const string PackageName = "Microsoft.DataApiBuilder";
+    private const string NuGetApiUrl = "https://api.nuget.org/v3-flatcontainer/{0}/index.json";
+
+    public static (string? LatestVersion, string? CurrentVersion) GetVersions()
+    {
+        var latestVersion = FetchLatestNuGetVersion();
+        var currentVersion = GetCurrentVersionFromAssembly(Assembly.GetExecutingAssembly());
+        return (latestVersion, currentVersion);
+    }
+
+    private static string? FetchLatestNuGetVersion()
+    {
+        try
+        {
+            using var httpClient = new HttpClient { Timeout = TimeSpan.FromSeconds(2) }; 
+            var url = string.Format(NuGetApiUrl, PackageName.ToLower());
+            var versionData = httpClient.GetFromJsonAsync<NuGetVersionResponse>(url).GetAwaiter().GetResult();
+
+            return versionData?.Versions
+                ?.Where(v => !v.Contains("-rc"))
+                .Max(); // Get the latest stable version
+        }
+        catch
+        {
+            return null; // Assume no update available on failure
+        }
+    }
+
+    private static string? GetCurrentVersionFromAssembly(Assembly assembly)
+    {
+        var version = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
+        return !string.IsNullOrEmpty(version) ? version.Split('+')[0] : assembly.GetName().Version?.ToString();
+    }
+
+    private class NuGetVersionResponse
+    {
+        [JsonPropertyName("versions")]
+        public string[]? Versions { get; set; }
+    }
+}
--- a/src/Service/Program.cs
+++ b/src/Service/Program.cs
@@ -27,6 +27,13 @@ namespace Azure.DataApiBuilder.Service
 
         public static void Main(string[] args)
         {
+            // Compare current version of DAB with latest (non-rc) version in NuGet. 
+            var (latestVersion, currentVersion) = await Azure.DataApiBuilder.Product.VersionChecker.GetVersionsAsync();
+            if (!string.IsNullOrEmpty(latestVersion) && latestVersion != currentVersion)
+            {
+                Console.Error.WriteLine("A newer version of Data API builder is available. {currentVersion} -> {latestVersion}");
+            }
+
             if (!ValidateAspNetCoreUrls())
             {
                 Console.Error.WriteLine("Invalid ASPNETCORE_URLS format. e.g.: ASPNETCORE_URLS=\"http://localhost:5000;https://localhost:5001\"");


src/Product/VersionChecker.cs
```diff
@@ -0,0 +1,48 @@
...
+    private static string? FetchLatestNuGetVersion()
+    {
+        try
+        {
+            using HttpClient httpClient = new HttpClient { Timeout = TimeSpan.FromSeconds(2) };
+            NuGetVersionResponse? versionData = httpClient.GetFromJsonAsync<NuGetVersionResponse>(NuGetApiUrl)
+                .GetAwaiter().GetResult();
+
+            return versionData?.Versions
+                ?.Where(version => !version.Contains("-rc"))
```
gilemos, line 28: Does rc mean that it is not a stable release?

src/Product/VersionChecker.cs
```diff
@@ -0,0 +1,48 @@
...
+                .Max(); // Get the latest stable version
```
gilemos, line 29: Since Versions is an array of strings, would Max() return the version that has the higher number (instead of the one that comes last when ordered alphabetically)?

Here's what I mean: https://stackoverflow.com/questions/45582461/c-sharp-max-from-string-array-not-always-biggest-number

src/Product/VersionChecker.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using System.Linq;
+using System.Net.Http;
+using System.Reflection;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Product;
+
+public static class VersionChecker
```
gilemos, line 9: nit: is it possible to add some unit tests for the methods here? :)

src/Product/VersionChecker.cs
JerryNixon: Yes, that is correct. RC stands for Release Candidate and those "versions" aren't stable, so we don't want to recommend them to users for upgrading.

src/Product/VersionChecker.cs
JerryNixon: Great catch, I used `Version` to resolve it.

src/Product/VersionChecker.cs
JerryNixon: Let's see if anyone likes it first.

src/Service/Program.cs
```diff
@@ -27,6 +27,13 @@ public class Program
 
         public static void Main(string[] args)
         {
+            // Compare current version of DAB with latest (non-rc) version in NuGet. 
+            VersionChecker.GetVersions(out string? latestVersion, out string? currentVersion);
+            if (!string.IsNullOrEmpty(latestVersion) && latestVersion != currentVersion)
+            {
+                Console.Error.WriteLine($"A newer version of Data API builder is available. {currentVersion} -> {latestVersion}");
```
Aniruddh25, line 34: Should we log a warning instead of an error? Would this affect the docker scenario too?

src/Service/Program.cs
```diff
@@ -27,6 +27,13 @@ public class Program
 
         public static void Main(string[] args)
         {
+            // Compare current version of DAB with latest (non-rc) version in NuGet.
```
Aniruddh25, line 30: Would this change work for the docker scenario too?

src/Service/Program.cs
```diff
@@ -27,6 +27,13 @@ public class Program
 
         public static void Main(string[] args)
         {
+            // Compare current version of DAB with latest (non-rc) version in NuGet. 
+            VersionChecker.GetVersions(out string? latestVersion, out string? currentVersion);
```
Aniruddh25, line 31: Should this check instead go to the CLI dab `start` command? instead of inside the engine?  That way docker isnt affected if this wasnt meant for that scenario..

Aniruddh25: Doing this version check everytime at the start of the engine seems time consuming, moving to dab init ensure the developers get this message when they initialize the configuration.

src/Core/VersionChecker.cs
```diff
@@ -0,0 +1,49 @@
...
+        currentVersion = GetCurrentVersionFromAssembly(Assembly.GetCallingAssembly());
+    }
+
+    private static string? FetchLatestNuGetVersion()
+    {
+        try
+        {
+            using HttpClient httpClient = new() { Timeout = TimeSpan.FromSeconds(2) };
+            NuGetVersionResponse? versionData = httpClient
+                .GetFromJsonAsync<NuGetVersionResponse>("https://api.nuget.org/v3-flatcontainer/microsoft.dataapibuilder/index.json")
```
Aniruddh25, line 21: save this url as a constant string property o the class

src/Core/VersionChecker.cs
```diff
@@ -0,0 +1,49 @@
+using System.Net.Http.Json;
+using System.Reflection;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Core;
+
+public static class VersionChecker
+{
+    public static void GetVersions(out string? latestVersion, out string? currentVersion)
```
Aniruddh25, line 9: better to return the versions as a value instead of the out references

src/Service/Program.cs
```diff
@@ -27,6 +28,13 @@ public class Program
...
+            if (!string.IsNullOrEmpty(latestVersion) && latestVersion != currentVersion)
```
Aniruddh25, line 33: what if people clone the repo, their currentVersion might be > latestVersion. Saying newer version is available doesn't make sense in that scenario, we should check for latestVersion > currentVersion

src/Core/VersionChecker.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            return null; // Assume no update available on failure
+        }
+    }
+
+    private static string? GetCurrentVersionFromAssembly(Assembly assembly)
+    {
+        string? version = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
+
+        return version is { Length: > 0 } && version.Contains('+')
```
Aniruddh25, line 39: I think there is an existing function to slice out the actual version, can we please reuse that?

Aniruddh25 (changes requested):
Great idea, but left some questions and suggestions to consider.

src/Core/VersionChecker.cs
JerryNixon: Since it has multiple outputs, aren't `out` parameters a better option than a single return type from the method which would require a complex type or a tuple? Let me know how imperative this request is, and I can change it.

src/Service/Program.cs
JerryNixon: I moved it to only run in `validate` which seems like a slow introduction.

src/Service/Program.cs
JerryNixon: If it fails to reach NuGet it does not emit a warning.

src/Service/Program.cs
JerryNixon: Since moving it to `dab validate` I have the logger and updated it to Warn.

src/Core/VersionChecker.cs
JerryNixon: I updated the method signature to return bool.

src/Core/VersionChecker.cs
JerryNixon: Yes. Used it. Good idea.

src/Core/VersionChecker.cs
```diff
@@ -1,15 +1,38 @@
...
+    //  If the internet is unavailable or NuGet is down or the HTTP request fails for any reason
+    /// (there is a 2 second Timeout on the request), then the NuGet version will be <c>null</c> and
+    //  this method will return <c>true</c>. This is mostly because this check is a user courtesy.
+    // </remarks>
+    public static bool IsCurrentVersion(out string? nugetVersion, out string? localVersion)
     {
-        latestVersion = FetchLatestNuGetVersion();
-        currentVersion = GetCurrentVersionFromAssembly(Assembly.GetCallingAssembly());
+        nugetVersion = FetchLatestNuGetVersion();
+        localVersion = ProductInfo.GetProductVersion(false);
```
Aniruddh25, line 34: nit: name the argument that is passed to GetProductVersion with value `false` for better readability..

src/Core/VersionChecker.cs
```diff
@@ -1,15 +1,38 @@
...
+        return string.IsNullOrEmpty(nugetVersion) || nugetVersion == localVersion;
```
Aniruddh25, line 35: What if the localVersion is > than the nugetVersion? Could happen when people clone our repo and we are in the process of developing 1.5, But nuget version will be 1.4 in that timeframe.

src/Cli.Tests/VersionCheckTests.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Azure.DataApiBuilder.Core;
+
+namespace Cli.Tests;
+
+[TestClass]
+public class VersionCheckTests
+{
+    [TestMethod]
+    public void GetVersions_LatestVersionNotNull()
```
Aniruddh25, line 12:
```suggestion
    public void GetVersions_NugetVersionNotNull()
```

Aniruddh25 (changes requested):
Good to move to dab validate for starters. PR description needs to be updated accordingly.

Still waiting for checking what happens if currentVersion > nugetVersion - could happen when people clone the repo and use the cli from the cloned build.

JerryNixon:
We've decided to start by adding this only to `dab validate`. And release in 1.6.

JerryNixon:
@Aniruddh25 can we assign this PR/branch to someone?

JerryNixon:
We wont do this

Issue #2280, opened 2024-06-25
[Bug]: caseInsensitive not working on GraphQL query


tim-tmp:
### What happened?

We are developing an Azure Static Web App with a CosmosDB accessed via GraphQL queries using the feature outlined here:
https://learn.microsoft.com/en-us/azure/static-web-apps/database-overview

When including **caseInsensitive** on a GraphQL query we get the following error:

_**"Operation caseInsensitive on int type not supported."**_

The caseInsensitive input field exists in the schema which Postman validates also.

`
StringFilterInput
 
          {
            "name": "caseInsensitive",
            "description": "Case Insensitive",
            "type": {
              "kind": "SCALAR",
              "name": "Boolean",
              "ofType": null
            },
            "defaultValue": "false"
          },
`

The query we are using is:
`{
        records: assets(
            filter: { Title: { contains: "hello world", caseInsensitive: true } },
            first: 10
        ) {
                items {
                    id, Title
                }
                endCursor
                hasNextPage
                __typename
            }
    }`

### Version

0.10

### What database are you using?

CosmosDB NoSQL

### What hosting model are you using?

Static Web Apps (SWA)

### Which API approach are you accessing DAB through?

GraphQL

### Relevant log output

```Text
Operation caseInsensitive on int type not supported.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Baker-Denn:
I'm having the same issue, using SWA on Cosmos NoSQL.

tim-tmp:
We were hoping you would implement the feature rather than just remove it from the generated schema but thanks for the update.

vadeveka:
Reopening issue as caseInsensitive operator was reintroduced in schema due to bad merge during HotChocolate 15 upgrade
https://github.com/Azure/data-api-builder/pull/2348/files#diff-f74a036f2a72f2b3f2642c66613b149c5dca7a9696859a26d4ce95863c124d9f

PR #2607, opened 2025-02-26
Remove caseInsensitive filter operator from schema generated

vadeveka:
## Why make this change?
caseInsensitive filter operator is not implemented in the resolvers leading to runtime error when used. Including it in schema seems to be an oversight. Removing it for schema generated until we have an approach defined for case insensitive comparisons with different source types

Relevant issue #2280 
image5

## What is this change?
Remove caseInsensitive filter operator from schema generated

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

### DIFF
--- a/src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
+++ b/src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
@@ -169,7 +169,6 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder.Queries
                     new(null, new NameNode("startsWith"), new StringValueNode("Starts With"), new StringType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("endsWith"), new StringValueNode("Ends With"), new StringType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("neq"), new StringValueNode("Not Equals"), new StringType().ToTypeNode(), null, new List<DirectiveNode>()),
-                    new(null, new NameNode("caseInsensitive"), new StringValueNode("Case Insensitive"), new BooleanType().ToTypeNode(), new BooleanValueNode(false), new List<DirectiveNode>()),
                     new(null, new NameNode("isNull"), new StringValueNode("Is null test"), new BooleanType().ToTypeNode(), null, new List<DirectiveNode>())
                 }
             );
@@ -232,7 +231,6 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder.Queries
                     new(null, new NameNode("startsWith"), new StringValueNode("Starts With"), new UuidType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("endsWith"), new StringValueNode("Ends With"), new UuidType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("neq"), new StringValueNode("Not Equals"), new UuidType().ToTypeNode(), null, new List<DirectiveNode>()),
-                    new(null, new NameNode("caseInsensitive"), new StringValueNode("Case Insensitive"), new BooleanType().ToTypeNode(), new BooleanValueNode(false), new List<DirectiveNode>()),
                     new(null, new NameNode("isNull"), new StringValueNode("Is null test"), new BooleanType().ToTypeNode(), null, new List<DirectiveNode>())
                 }
             );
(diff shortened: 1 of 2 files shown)


src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
```diff
@@ -232,7 +231,6 @@ public static InputObjectTypeDefinitionNode UuidInputType() =>
                     new(null, new NameNode("startsWith"), new StringValueNode("Starts With"), new UuidType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("endsWith"), new StringValueNode("Ends With"), new UuidType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("neq"), new StringValueNode("Not Equals"), new UuidType().ToTypeNode(), null, new List<DirectiveNode>()),
-                    new(null, new NameNode("caseInsensitive"), new StringValueNode("Case Insensitive"), new BooleanType().ToTypeNode(), new BooleanValueNode(false), new List<DirectiveNode>()),
```
Aniruddh25, line 235: were you able to see which PR merged this change in and perhaps find why the support was not added?

src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
```diff
@@ -169,7 +169,6 @@ public static InputObjectTypeDefinitionNode StringInputType() =>
                     new(null, new NameNode("startsWith"), new StringValueNode("Starts With"), new StringType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("endsWith"), new StringValueNode("Ends With"), new StringType().ToTypeNode(), null, new List<DirectiveNode>()),
                     new(null, new NameNode("neq"), new StringValueNode("Not Equals"), new StringType().ToTypeNode(), null, new List<DirectiveNode>()),
-                    new(null, new NameNode("caseInsensitive"), new StringValueNode("Case Insensitive"), new BooleanType().ToTypeNode(), new BooleanValueNode(false), new List<DirectiveNode>()),
```
Aniruddh25, line 172: nit: fix title typo, should be:  `caseInsensitive`

src/Service.Tests/MultiSourceTestSchema.gql
```diff
@@ -77,8 +77,6 @@ input StringFilterInput {
   endsWith: String
   "Not Equals"
   neq: String
-  "Case Insensitive"
```
Aniruddh25, line 80: is the same in any other gql files?

src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
vadeveka: Fixed

src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
vadeveka: This file originally came with https://github.com/Azure/data-api-builder/pull/250 but it isnt clear where Aaron borrowed all of the operators from looking at other modified files.

src/Service.Tests/MultiSourceTestSchema.gql
vadeveka: Didnt find this in any other files.

Aniruddh25 (approved):

Aniruddh25:
I wonder why is the error saying that the caseInsensitive operator is not supported for Int type since clearly it is a StringType that we are using it against.

vadeveka:
> I wonder why is the error saying that the caseInsensitive operator is not supported for Int type since clearly it is a StringType that we are using it against.

That would be because the exception is hardcoded here
https://github.com/Azure/data-api-builder/blob/f2b4a2452906697d798d7f4bfc0a654388db5065/src/Core/Models/GraphQLFilterParsers.cs#L697

lingxiao-microsoft (approved):

src/Service.GraphQLBuilder/Queries/StandardQueryInputs.cs
vadeveka: Leveraging git search, it seems like above commit is the one that introduced the string caseInsensitive and it didnt exist before that.
```
Q:\Repos\data-api-builder> git log -S caseInsensitive --oneline
a9d0a622 (HEAD -> caseInsensitive, origin/caseInsensitive) Remove casInsenitive filter operator from schema generated
3176820a Data Api Builder support for multiple databases. (#1709)
8b05b5ba Fix for guid filter in GraphQL (#1644)
a0e9881f Generating a full GraphQL schema from data models (#250)
```

PR #2609, opened 2025-02-28
[Bugfix] Ensure aggregation and groupby nodes are properly handled in case tables have no fields to aggregate or groupby on

rohkhann:
## Why make this change?
When the table has no numeric types we were erroring out because of the way groupby and aggregation nodes are generated. This code handles logic where if there are no applicable groupby or aggregation nodes available for an entity we do not generate those nodes for that entity

## What is this change?

Refactored enum generation logic in GraphQL schema creation to ensure enums are created only when necessary. Updated methods in EnumTypeBuilder to return success status and renamed `GenerateFieldsEnum` to `TryGenerateEnum`. Added a new parameter in SchemaConverter for conditional aggregation node generation.

Added `date_only_table` to SQL schemas with sample data. Updated configuration files to include the new entity and permissions. Added tests to validate non-generation of aggregation nodes without numeric fields.

## How was this tested?

tests added.


JerryNixon:
Can you add a bug-specific test to ensure we don't have a regression?

rohkhann:
> Can you add a bug-specific test to ensure we don't have a regression?

Yes have added.

src/Service.GraphQLBuilder/Queries/EnumTypeBuilder.cs
```diff
@@ -39,7 +39,7 @@ public static string GenerateScalarFieldsEnumName(string entityName)
         /// <param name="fieldSelector">Function to select which fields to include in the enum.</param>
         /// <param name="enumNameSuffix">Suffix to append to the enum type name.</param>
         /// <param name="description">Description for the generated enum type.</param>
-        public static void GenerateFieldsEnum(
+        public static bool TryGenerateEnum(
```
vadeveka, line 42: NIT; add return description in summary

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -246,14 +246,18 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
...
+                                bool aggregationEnumCreated = EnumTypeBuilder.GenerateAggregationNumericEnumForObjectType(node, enumTypes);
+                                if (aggregationEnumCreated)
                                 {
+                                    // Generate aggregation type for the entity
+                                    ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(node.Name.Value, node);
                                     objectTypes.Add(SchemaConverter.GenerateObjectAggregationNodeName(entityName), aggregationType);
-                                    objectTypes.Add(SchemaConverter.GenerateGroupByTypeName(entityName), SchemaConverter.GenerateGroupByTypeForEntity(node.Name.Value, node));
+                                }
+
+                                bool groupByColumnsEnumCreated = EnumTypeBuilder.GenerateScalarFieldsEnumForObjectType(node, enumTypes);
```
vadeveka, line 257: NIT: a code comment/schema snippet explaining what this is generating would be helpful

vadeveka (approved):

lingxiao-microsoft (approved):

src/Service.GraphQLBuilder/Queries/EnumTypeBuilder.cs
rohkhann: fixed.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: fixed.

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -246,14 +246,19 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
 
                             if (_isAggregationEnabled)
                             {
-                                EnumTypeBuilder.GenerateAggregationNumericEnumForObjectType(node, enumTypes);
-                                EnumTypeBuilder.GenerateScalarFieldsEnumForObjectType(node, enumTypes);
-                                // Generate aggregation type for the entity
-                                ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(node.Name.Value, node);
-                                if (aggregationType.Fields.Any())
+                                bool aggregationEnumCreated = EnumTypeBuilder.GenerateAggregationNumericEnumForObjectType(node, enumTypes);
```
aaronburtle, line 249: nit: try to use a prefix for bool names of "is", "has", etc

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -246,14 +246,19 @@ private DocumentNode GenerateSqlGraphQLObjects(RuntimeEntities entities, Diction
...
+                                if (aggregationEnumCreated)
                                 {
+                                    // Generate aggregation type for the entity
+                                    ObjectTypeDefinitionNode aggregationType = SchemaConverter.GenerateAggregationTypeForEntity(node.Name.Value, node);
                                     objectTypes.Add(SchemaConverter.GenerateObjectAggregationNodeName(entityName), aggregationType);
-                                    objectTypes.Add(SchemaConverter.GenerateGroupByTypeName(entityName), SchemaConverter.GenerateGroupByTypeForEntity(node.Name.Value, node));
+                                }
+
+                                // this generates a scalar fields enum that has the list of fields you can have in groupby arguments.
+                                bool groupByColumnsEnumCreated = EnumTypeBuilder.GenerateScalarFieldsEnumForObjectType(node, enumTypes);
```
aaronburtle, line 258: nit: bool name prefix

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -336,11 +336,11 @@ private static FieldDefinitionNode CreateNumericAggregationField(string operatio
         /// </summary>
         /// <param name="entityName">Name of the entity</param>
         /// <param name="entityNode">The entity's ObjectTypeDefinitionNode</param>
+        /// <param name="shouldGenerateAggregationNode">Should aggregation node be generated.</param>
         /// <returns>ObjectTypeDefinitionNode for the GroupBy type</returns>
-        public static ObjectTypeDefinitionNode GenerateGroupByTypeForEntity(string entityName, ObjectTypeDefinitionNode entityNode)
+        public static ObjectTypeDefinitionNode GenerateGroupByTypeForEntity(string entityName, ObjectTypeDefinitionNode entityNode, bool shouldGenerateAggregationNode)
```
aaronburtle, line 341: nit: bool parameter tends to be a code smell. Is it possible to split the conditional out of this method and either into a separate method that does the additional behavior, or have 2 methods, one to generate the group by type for an entity with aggregation node and one without?

It is your call if think a bool param here helps or not, but generally we try to avoid.

aaronburtle (approved):
Look great, just a couple of nits.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: fixed.

src/Core/Services/GraphQLSchemaCreator.cs
rohkhann: fixed.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
rohkhann: fixed.

PR #2610, opened 2025-03-03
Add more tests for EntraId/AzureAd providers

### LINKED ISSUE #2407: ⭐ [Enhancement]: Add authentication.provider synonyms.
Update `runtime.host.authentication.provider` with synonyms for future planning.

> These are already accounted for in the `application_name` specification.

## Values

| Current      | New Synonyms
|-|-|
| StaticWebApp | None
| AppService   |
| AzureAd      | EntraId, OAuth
| Simulator    |

### Tasks

- [ ] Create synonyms
- [ ] Ensure `application_name` compatibility
- [ ] Update documentation

## None
Some Data API implementations don't require any security at all. There is no option for `None`, although `StaticWebApp` handles this in most cases. The truth is, this is slightly more like `Simulator` except it behaves the opposite—none is treated as `authenticated`.

## EntraId
Azure Active Directory was renamed to Microsoft Entra Identity. We will retain the previous `AzureAd` moniker for backward compatibility, but it will be marked as deprecated in the documentation once this synonym is available.

## OAuth
Currently, `AzureAd` is just a simple OAuth implementation, and any custom JWT implementation can use it to handle claims. However, it’s far from ideal to make customers choose `AzureAd` or `EntraId` when their underlying authentication technology is not from Microsoft. This also opens up a future direction should we need to handle special cases with `EntraId`.

sander1095:
## Why make this change?

Closes #2407 

This issue adds more tests for the `EntraId` authentication provider.

`EntraId` already seemed to be supported, so no business logic was needed for this change.

## What is this change?

Several tests have been added.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)

Use `EntraId` and `AzureAd` providers like before.


src/Core/AuthenticationHelpers/ConfigureJwtBearerOptions.cs
```diff
@@ -59,6 +59,10 @@ public void Configure(string? name, JwtBearerOptions options)
             RoleClaimType = AuthenticationOptions.ROLE_CLAIM_TYPE
         };
 
+        // TODO: The PR says "This (OAuth) also opens up a future direction should we need to handle special cases with EntraId."
```
sander1095, line 62: This needs a discussion :)

sander1095:
This PR still needs some discussions with the team before it can be removed from the draft status, as the issue's requirements don't really match DAB's behaviour.

- [ ] There is a TODO in the code where adding a synonym would cause behavioral differences, which isn't desired according to the issue. Please take a look at the comment and let's discuss!
- [ ] The issue mentions adding properties to the JSON Schema. However, the JSON schema currently [doesn't validate](https://github.com/Azure/data-api-builder/blob/1eada8430c45630476b1f3fce3d93a78a64fa84b/schemas/dab.draft.schema.json#L275) any `authentication.provider` entries. I can add this, but is this desired in this PR?
  - [ ]  Nor does it validate (in the json schema) that `jwt` should be null when the provider isn't `AzureAd`/`EntraId`/`OAuth`. This is [validated in code](https://github.com/azure/data-api-builder/blob/bc0f8f075985e97a5c0aee7761f2dd90043a8f84/src/Cli/Utils.cs#L509). Is there a reason this is only validated at runtime and not in config? Are any changes desired in this PR?
- [ ] DAB's documentation isn't updated in this PR because the docs aren't part of this repo, it seems. If desired, I can write some docs, or is this usually done by people at Microsoft?
- [ ] Perhaps some more tests are desired. If so, please point out some locations!

src/Core/AuthenticationHelpers/EasyAuthAuthenticationHandler.cs
```diff
@@ -68,6 +68,8 @@ protected override Task<AuthenticateResult> HandleAuthenticateAsync()
             {
                 EasyAuthType.StaticWebApps => StaticWebAppsAuthentication.Parse(Context, Logger),
                 EasyAuthType.AppService => AppServiceAuthentication.Parse(Context, Logger),
+                EasyAuthType.EasyAuth => StaticWebAppsAuthentication.Parse(Context, Logger), // TODO: Is this correct?
```
sander1095, line 71: @DAB team: Please advice

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -179,8 +179,13 @@ public static bool IsSystemRole(string roleName)
     /// <returns>Authentication Scheme</returns>
     private static string ResolveConfiguredAuthNScheme(string? configuredProviderName)
     {
+        // TODO: Jerry Nixon says that EASY_AUTH is also a synonym for STATIC_WEB_APPS/APP_SERVICE/NONE in DAB.
```
sander1095, line 182: @dab team: Please advise

Aniruddh25:
- [ ] There is a TODO in the code where adding a synonym would cause behavioral differences, which isn't desired according to the issue. Please take a look at the comment and let's discuss!
 
Got it, lets discuss in line with respect to each TODO
 
- [ ]  The issue mentions adding properties to the JSON Schema. However, the JSON schema currently [doesn't validate](https://github.com/Azure/data-api-builder/blob/1eada8430c45630476b1f3fce3d93a78a64fa84b/schemas/dab.draft.schema.json#L275) any authentication.provider entries. I can add this, but is this desired in this PR?

Yes please add the properties in the schema in this PR or in an immediate subsequent one even though they are not being validated. 

- [ ]  Nor does it validate (in the json schema) that jwt should be null when the provider isn't AzureAd/EntraId/OAuth. This is [validated in code](https://github.com/azure/data-api-builder/blob/bc0f8f075985e97a5c0aee7761f2dd90043a8f84/src/Cli/Utils.cs#L509). Is there a reason this is only validated at runtime and not in config? Are any changes desired in this PR?

Lets create a separate PR to validate the config using the schema. It was simply a miss by previous contributors. 

- [ ]  DAB's documentation isn't updated in this PR because the docs aren't part of this repo, it seems. If desired, I can write some docs, or is this usually done by people at Microsoft?

Could you please provide edits to the docs here: 

https://learn.microsoft.com/en-us/azure/data-api-builder/authentication-local#use-the-simulator-provider 

- [ ]  Perhaps some more tests are desired. If so, please point out some locations!
- [ ]  I'm running on Windows with the git pre commit hook installed. When I run dotnet format and try to commit, the hook fails. Perhaps because of CRLF changes? In any case, I disabled the hook and ran dotnet format manually but this might still need some looking into.

Will look into the PR this week. Thanks for your patience.

src/Core/AuthenticationHelpers/ConfigureJwtBearerOptions.cs
RubenCerna2079: I don't understand how that would be a behavioral change? From my understanding (and correct me if I am wrong), EntraId will be the one that would need the behavioral changes in the future as it is EntraId the one that will handle the special cases.

src/Core/AuthenticationHelpers/EasyAuthAuthenticationHandler.cs
RubenCerna2079: My only preference would be to group together all the `EasyAuthType`'s that parse the `StaticWebAppsAuthentication`

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -145,6 +145,79 @@ public void TestValidateConfigFailsWithInvalidGraphQLDepthLimit(object? depthLim
         }
     }
 
+    /// <summary>
+    /// This Test is used to verify that the None authentication scheme is valid
```
RubenCerna2079, line 149:
```suggestion
    /// This Test is used to verify that the None authentication schema is valid.
```

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -145,6 +145,79 @@ public void TestValidateConfigFailsWithInvalidGraphQLDepthLimit(object? depthLim
...
+            Assert.IsTrue(ConfigGenerator.IsConfigValid(validateOptions, _runtimeConfigLoader!, _fileSystem!));
+        }
+        catch (Exception ex)
+        {
+            Assert.Fail($"Unexpected Exception thrown: {ex.Message}");
+        }
+    }
+
+    /// <summary>
+    /// This Test is used to verify that the None authentication scheme is valid
```
RubenCerna2079, line 172:
```suggestion
    /// This Test is used to verify that the None authentication schema is valid.
```

src/Core/AuthenticationHelpers/ConfigureJwtBearerOptions.cs
sander1095: Hopefully my reply will make my other comments clear, so please revisit them and the ones you already commented on now that you have extra context.

This will be 1 big comment to address several comments. This saves me some time :)

--------------

Let me start [by repeating myself](https://github.com/Azure/data-api-builder/pull/2610#issuecomment-2703135286):

> If my team agree that my TODO's require this PR to become more complex, it might be worth putting the generic EasyAuth auth provider in a seperate issue to keep this PR mergeable.

This PR used to simply add `None` and `OAuth` as synonyms for specific auth schemes and was basically 100% mergeable; there were [a couple small TODO's I still needed to the team to look at](https://github.com/Azure/data-api-builder/pull/2610#issuecomment-2694776786).

I believe that the TODO's/problems I mention in comments in this PR add extra complexity to this PR. This will require me to put more time into this PR right now, which I don't have because I am busy with moving into a new house.

Therefore, I'd still like to propose to merge this and put [the extra work Jerry and Aniruddh mention](https://github.com/Azure/data-api-builder/pull/2610#issuecomment-2746848756) in a seperate issue and PR.

If that's not desired, I'll have to ask the DAB team to finish this PR and the work from the comment above, as I won't have time for the next month to do this, sadly.

--------------

## ConfigureJwtBearerOptions.cs
The core of the problem is that `EasyAuth` and `OAuth` aren't just synonyms, unlike `None` and `OAuth`.

`EasyAuth` and `OAuth` are the "protocols", for which `StaticWebApps` and `AzureAD` and `EntraId` are implementations.

- `OAuth` is NOT specific to Microsoft Entra Id, you can also use other identity servers. However, if I were to add `OAuth` to this if statement, it would enable `EnableAadSigningKeyIssuerValidation`, which sounds like an Azure AD specific call.
  - So, if someone would use something like keycloak or some other OAuth server, it wouldn't make sense to include AadSigningKeyIssuerValidation as that has nothing to do with their identity server.

## EasyAuthAuthenticationBuilderExtensions.cs
Another example in this PR: Take a look at `src/Core/AuthenticationHelpers/EasyAuthAuthenticationBuilderExtensions.cs`. Please reply in that comment with what is desired.

I can add `EasyAuth` to that if statement and treat it as a synonym, but that would make it use `authenticationScheme: EasyAuthAuthenticationDefaults.SWAAUTHSCHEME`. That doesn't make sense, because EasyAuth can mean both `StaticWebApps` and `AppService`, so we are making an incorrect assumption. Even if we don't directly support App Service now, if in the future other EasyAuth providers are added, this can cause bugs.

**In the same file**, there is another issue:

```csharp
EasyAuthType.StaticWebApps => StaticWebAppsAuthentication.Parse(Context, Logger),
EasyAuthType.AppService => AppServiceAuthentication.Parse(Context, Logger),
EasyAuthType.EasyAuth => StaticWebAppsAuthentication.Parse(Context, Logger), // TODO: Is this correct?
```

If the user specified `EasyAuth`, we're using the Parse method from `StaticWebAppsAuthentication`. Again, we're making an assumption that EasyAuth's implementation will always be StaticWebApps, but in that code block you can also see that AppService is an option. What if someone wanted to use App Service and specifies EasyAuth? This will be weird behavior and can cause bugs.

## ClientRoleHeaderAuthenticationMiddleware

In `ClientRoleHeaderAuthenticationMiddleware.cs` I have another comment with a similar issue where `EASY_AUTH` defaults to StaticWebApps, skipping over the fact that AppService (which is also used in this codebase) also does EASY_AUTH stuff.

RubenCerna2079: Thanks for explaining, based on this comment I think it would be better to put `EasyAuth` on a different PR. But I would like confirmation of @Aniruddh25 or @JerryNixon before we decide on this

src/Cli.Tests/InitTests.cs
```diff
@@ -299,9 +299,13 @@ public void EnsureFailureOnReInitializingExistingConfig()
         /// </summary>
         [DataTestMethod]
         [DataRow("StaticWebApps", null, null, DisplayName = "StaticWebApps with no audience and no issuer specified.")]
+        [DataRow("None", null, null, DisplayName = "None with no audience and no issuer specified.")]
+        [DataRow("EasyAuth", null, null, DisplayName = "EasyAuth with no audience and no issuer specified.")]
         [DataRow("AppService", null, null, DisplayName = "AppService with no audience and no issuer specified.")]
         [DataRow("Simulator", null, null, DisplayName = "Simulator with no audience and no issuer specified.")]
         [DataRow("AzureAD", "aud-xxx", "issuer-xxx", DisplayName = "AzureAD with both audience and issuer specified.")]
+        [DataRow("EntraId", "aud -xxx", "issuer-xxx", DisplayName = "EntraId with both audience and issuer specified.")]
```
RubenCerna2079, line 307:
```suggestion
        [DataRow("EntraId", "aud-xxx", "issuer-xxx", DisplayName = "EntraId with both audience and issuer specified.")]
```

src/Cli.Tests/InitTests.cs
```diff
@@ -299,9 +299,13 @@ public void EnsureFailureOnReInitializingExistingConfig()
...
+        [DataRow("OAuth", "aud -xxx", "issuer-xxx", DisplayName = "OAuth with both audience and issuer specified.")]
```
RubenCerna2079, line 308:
```suggestion
        [DataRow("OAuth", "aud-xxx", "issuer-xxx", DisplayName = "OAuth with both audience and issuer specified.")]
```

src/Core/AuthenticationHelpers/ConfigureJwtBearerOptions.cs
sander1095: @RubenCerna2079 Alright! Keep in mind that the `OAuth` synonym also still needs a discussion, as the use here adds the AadSigningKeyValidation.

I think this is fixable by simply not adding it, but this would be a behavioral change and cause it to no longer be a simple synonym.

Aniruddh25: We need to check the differences between AppService and StaticWebApps. 

If both are essentially following the same approach, we could make EasyAuth = AppService = StaticWebApps all synonymous. Looking purely at the code, the fact that AppService was created as a separate Auth Scheme means that there had been some difference, but we need to validate that and make a decision accordingly.

That said, as of this standing, just replacing EasyAuth with StaticWebApps would create confusion and be incorrect. I agree to keep that in a separate PR. 

@sander1095, I know you could be busy, but if you can, please remove the `EasyAuth` synonym changes. 
Otherwise, @RubenCerna2079 can take over as well. Let us know either way :)

src/Cli/ConfigGenerator.cs
```diff
@@ -222,9 +222,9 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
 
             if (runtimeBaseRoute is not null)
             {
-                if (!Enum.TryParse(options.AuthenticationProvider, ignoreCase: true, out EasyAuthType authMode) || authMode is not EasyAuthType.StaticWebApps)
+                if (!Enum.TryParse(options.AuthenticationProvider, ignoreCase: true, out EasyAuthType authMode) || authMode is not EasyAuthType.StaticWebApps or EasyAuthType.EasyAuth or EasyAuthType.None)
```
Aniruddh25, line 225: Lets remove the change about EasyAuth.

src/Config/ObjectModel/EasyAuthType.cs
```diff
@@ -6,5 +6,15 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 public enum EasyAuthType
 {
     StaticWebApps,
-    AppService
+    AppService,
+
+    /// <summary>
```
Aniruddh25, line 11: Remove the `EasyAuth` change from this PR.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -179,8 +179,13 @@ public static bool IsSystemRole(string roleName)
...
+        // But as far as I know, EASY_AUTH is an abstraction on the auth schemes that SWA/AS use.
+        // So the question is: Is this IF statement still correct now that I added EASY_AUTH to it? It would now default to SWA auth.
         if (string.IsNullOrWhiteSpace(configuredProviderName)
-            || string.Equals(configuredProviderName, SupportedAuthNProviders.STATIC_WEB_APPS, StringComparison.OrdinalIgnoreCase))
+            || string.Equals(configuredProviderName, SupportedAuthNProviders.STATIC_WEB_APPS, StringComparison.OrdinalIgnoreCase)
+            || string.Equals(configuredProviderName, SupportedAuthNProviders.EASY_AUTH, StringComparison.OrdinalIgnoreCase)
```
Aniruddh25, line 187: Lets remove Easy_AUTH from this PR.

src/Core/AuthenticationHelpers/ConfigureJwtBearerOptions.cs
Aniruddh25: @sander1095, you are right about OAuth synonym as well. We cannot simply take it as AzureAD synonym. Please remove that synonym as well. Lets only use `EntraId` as the synonym for `AzureAd`

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -193,7 +198,8 @@ private static string ResolveConfiguredAuthNScheme(string? configuredProviderNam
             return SimulatorAuthenticationDefaults.AUTHENTICATIONSCHEME;
         }
         else if (string.Equals(configuredProviderName, SupportedAuthNProviders.AZURE_AD, StringComparison.OrdinalIgnoreCase) ||
-            string.Equals(configuredProviderName, SupportedAuthNProviders.ENTRA_ID, StringComparison.OrdinalIgnoreCase))
+            string.Equals(configuredProviderName, SupportedAuthNProviders.ENTRA_ID, StringComparison.OrdinalIgnoreCase) ||
+            string.Equals(configuredProviderName, SupportedAuthNProviders.OAUTH, StringComparison.OrdinalIgnoreCase))
```
Aniruddh25, line 202: Lets remove OAUTH as the synonym for AzureAD too. Only EntraID should be the synonym for AzureAD.

src/Cli/ConfigGenerator.cs
```diff
@@ -222,9 +222,9 @@ public static bool TryCreateRuntimeConfig(InitOptions options, FileSystemRuntime
...
                 {
-                    _logger.LogError("Runtime base-route can only be specified when the authentication provider is Static Web Apps.");
+                    _logger.LogError("Runtime base-route can only be specified when the authentication provider is Static Web Apps, EasyAuth or None.");
```
Aniruddh25, line 227: Lets also remove `None` as synonym for `StaticWebApp`. I know the issue said to add None, but we really need to verify what if AppService == StaticWebApp and make this decision.

src/Config/ObjectModel/EasyAuthType.cs
```diff
@@ -6,5 +6,15 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
...
+    /// A synonym for <see cref="StaticWebApps"/>
+    /// </summary>
+    EasyAuth,
+
+    /// <summary>
+    /// Another synonym for <see cref="StaticWebApps"/>, like <see cref="EasyAuth"/>
+    /// </summary>
+    None
```
Aniruddh25, line 19: Let us remove the `None` synonym as well.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
sander1095: Not relevant anymore since latest changes

src/Cli/ConfigGenerator.cs
sander1095: I've implemented requested changes. I'm resolving this comment chain.

sander1095:
@Aniruddh25 @JerryNixon  @RubenCerna2079 

This PR has been a wild ride ;)!

This PR now contains all the latest requested changes. Now that `None`, OAuth` and `EasyAuth` have been removed from this PR's changeset, the resulting PR only contains some missing tests for `EntraId`; `EntraId` already existed as a synonym for `AzureAd`, so it seems no changes were required there. The test results will tell.

**Some actionable items for @Aniruddh25 / @JerryNixon :**
I suggest to create separate issues for adding `None`, `OAuth` and `EasyAuth` in the future, where each of these can get the attention they need. The comments in this PR can be used as discussion points, as I had already picked out several spots of code that would cause behavioral differences.

------

In an earlier comment, I discussed the issue of the JSON schema not having a list of auth providers to choose from, which also meant that weren't validated. I also talked about this at the MVP Summit. @Aniruddh25 said they could be added in this PR, but I disagree with this because this PR is currently nice, small and (hopefully) mergeable. I've created #2643 to further discuss this.

Let me know if any small changes are still needed.

src/Cli.Tests/InitTests.cs
```diff
@@ -302,6 +302,7 @@ public void EnsureFailureOnReInitializingExistingConfig()
         [DataRow("AppService", null, null, DisplayName = "AppService with no audience and no issuer specified.")]
         [DataRow("Simulator", null, null, DisplayName = "Simulator with no audience and no issuer specified.")]
         [DataRow("AzureAD", "aud-xxx", "issuer-xxx", DisplayName = "AzureAD with both audience and issuer specified.")]
+        [DataRow("EntraId", "aud-xxx", "issuer-xxx", DisplayName = "EntraId with both audience and issuer specified.")]
```
RubenCerna2079, line 305: I see that this test is failing and I think it is because of a misspelling in `EntraId`, if you go to `SupportedAuthNProviders.cs` you can see that it is spelled as `EntraID`.

```suggestion
        [DataRow("EntraID", "aud-xxx", "issuer-xxx", DisplayName = "EntraID with both audience and issuer specified.")]
```

src/Service.Tests/Authentication/EasyAuthAuthenticationUnitTests.cs
```diff
@@ -156,21 +156,23 @@ public async Task TestMissingEasyAuthHeader(EasyAuthType easyAuthType)
...
         [DataTestMethod]
-        [DataRow(false, true, DisplayName = "Valid StaticWebApps EasyAuth header only")]
-        [DataRow(true, true, DisplayName = "Valid StaticWebApps EasyAuth header and authorization header")]
+        [DataRow(EasyAuthType.StaticWebApps, false, true, DisplayName = "Valid EasyAuth (Static Web Apps) header only")]
+        [DataRow(EasyAuthType.StaticWebApps, true, true, DisplayName = "Valid EasyAuth (Static Web Apps) header and authorization header")]
+        [DataRow(EasyAuthType.AppService, false, true, DisplayName = "Valid EasyAuth (App Service) header only")]
+        [DataRow(EasyAuthType.AppService, true, true, DisplayName = "Valid EasyAuth (App Service) header and authorization header")]
         [TestMethod]
-        public async Task TestValidStaticWebAppsEasyAuthToken(bool sendAuthorizationHeader, bool addAuthenticated)
+        public async Task TestValidEasyAuthToken(EasyAuthType easyAuthType, bool sendAuthorizationHeader, bool addAuthenticated)
```
RubenCerna2079, line 170: I see this test is failing with the `AppService`, the way it was created was supposed to only work for `StaticWebApps` so I think it is a good idea to revert the changes in this test.

sander1095: I have reset it. Thanks

src/Cli.Tests/InitTests.cs
```diff
@@ -302,6 +302,7 @@ public void EnsureFailureOnReInitializingExistingConfig()
         [DataRow("AppService", null, null, DisplayName = "AppService with no audience and no issuer specified.")]
         [DataRow("Simulator", null, null, DisplayName = "Simulator with no audience and no issuer specified.")]
         [DataRow("AzureAD", "aud-xxx", "issuer-xxx", DisplayName = "AzureAD with both audience and issuer specified.")]
+        [DataRow("EntraID", "aud-xxx", "issuer-xxx", DisplayName = "EntraID with both audience and issuer specified.")]
```
RubenCerna2079, line 305: I believe this is the final fix. The problem is that this test doesn't have a snapshot counterpart to compare its result to. If this is too much work for you at this point, I can take care of it :)
- Run the test locally, this is expected to fail
- Look for a file that was created called `InitTests.EnsureCorrectConfigGenerationWithDifferentAuthenticationProviders_(some string of random words).verified.txt`
- This file should be under `Cli.Tests->Snapshots` there are multiple files with this name, the one you are looking for should be completely empty.
- Paste the following config in the empty `verified.txt` 
```
{
  DataSource: {
    DatabaseType: MSSQL,
    Options: {
      set-session-context: false
    }
  },
  Runtime: {
    Rest: {
      Enabled: true,
      Path: /api,
      RequestBodyStrict: true
    },
    GraphQL: {
      Enabled: true,
      Path: /graphql,
      AllowIntrospection: true
    },
    Host: {
      Cors: {
        AllowCredentials: false
      },
      Authentication: {
        Provider: EntraID,
        Jwt: {
          Audience: aud-xxx,
          Issuer: issuer-xxx
        }
      },
      Mode: Production
    }
  },
  Entities: []
}
```
- Run it again to ensure that the test works properly

sander1095:
Hi @RubenCerna2079 ! Thanks for the assist on the snapshot test. I've fixed it.

I performed [an extra commit](https://github.com/Azure/data-api-builder/pull/2610/commits/7fd24a17b2b4a36bf2698ee36c2fe1ed112fd087) with extra tests; I saw there were more tests in the codebase that tested `AzureAD` without testing for `EntraID`. These tests have been updated to test `EntraID`, too.

I had some difficulties getting tests to run on my machine, for some reason. I expect the pipeline to pass. If not, I can have a look at it tomorrow. Perhaps you can perform another pipeline run :)?

RubenCerna2079 (approved):
LGTM! Thank you so much @sander1095, hope everything works out while you are moving to your new house.

Aniruddh25 (approved):
LGTM, thank you @sander1095 for this change. 

You asked REALLY good questions that made us rethink the supposedly similar synonymous EasyAuth, StaticWebApp,  AppService. We will followup with another issue to clarify those if they really are or not.

PR #2473, opened 2024-11-24
Support quoted table names

thomasfroehle:
## Why make this change?
- Fixes #2385 

## What is this change?
- In [SqlMetadataProvider.PopulateColumnDefinitionsWithReadOnlyFlag](https://github.com/Azure/data-api-builder/blob/c9e836b551eb6bc729e88e71b3aa4d1c1b235345/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs#L1355) the QuoteIdentifier-method of the SqlQueryBuilder was not used to add quotes to the table name

## How was this tested?

- [ ] Integration Tests -> I'm happy to extend the integration tests, if this fix seems like a good idea
- [ ] Unit Tests
- [x] manually

## Sample Request(s)
Sample configuraiton: 
```
"MyTable": {
  "source": {
    "object": "schema.\"MyTable\"",
    "type": "table"
  }
}
```

PR #2616, opened 2025-03-13
Support quoted table names

### LINKED ISSUE #2385: [Bug]: Quoted Table name are not supported
### What happened?

I tried to use DAB on my Postgresql DB. My table name are case sensitive.
When i try to start i get a failure.

Here is a sample of my dab-config.json
"AssetHealthState": {
      "source": {
        "object": "asset_health.\"AssetHealthState\"",
        "type": "table"
      },

### Version

Microsoft.DataApiBuilder 1.2.10+c7ca8db8558a63919c530e454c8f18b45d5b931c

PostgreSQL

Local (including CLI)

REST

### Relevant log output

```Text
failinfo: Azure.DataApiBuilder.Service.Startup[0]
      Unable to complete runtime initialization. Refer to exception for error details.
      Azure.DataApiBuilder.Service.Exceptions.DataApiBuilderException: Cannot obtain Schema for entity AssetHealthState with underlying database object source: asset_health."AssetHealthState" due to: 42P01: relation "asset_health."AssetHealthState"" does not exist
```

### Code of Conduct

tomfroehle:
Duplicates #2473 because I cannot contribute with my enterprise account anymore. 

Original PR description:

## Why make this change?
- Fixes #2385 

## What is this change?
- In [SqlMetadataProvider.PopulateColumnDefinitionsWithReadOnlyFlag](https://github.com/Azure/data-api-builder/blob/c9e836b551eb6bc729e88e71b3aa4d1c1b235345/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs#L1355) the QuoteIdentifier-method of the SqlQueryBuilder was not used to add quotes to the table name

## How was this tested?

- [ ] Integration Tests -> I'm happy to extend the integration tests, if this fix seems like a good idea
- [ ] Unit Tests
- [x] manually

## Sample Request(s)
Sample configuraiton: 
```
"MyTable": {
  "source": {
    "object": "schema.\"MyTable\"",
    "type": "table"
  }
}
```


aaronburtle:
Thanks for following up!

aaronburtle (approved):

JerryNixon:
What happens to existing customers who have indicated their table name but have used the incorrect casing for those tables? Will this break those customers?

src/Core/Resolvers/MySqlQueryBuilder.cs
```diff
@@ -367,5 +367,11 @@ public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName)
         {
             throw new NotImplementedException();
         }
+
+        public string QuoteTableNameAsDBConnectionParam(string param)
+        {
+            // Table names in MSSQL should not be quoted when used as DB Connection Params.
```
Aniruddh25, line 373:
```suggestion
            // Table names in MySQL should not be quoted when used as DB Connection Params.
```

Aniruddh25:
> What happens to existing customers who have indicated their table name but have used the incorrect casing for those tables? Will this break those customers?

Good question. Lets consider the 2 cases if customers have used incorrect casing, 

1) and their table names are case sensitive - it must not have been working for them anyway right now without the quotes. After adding the quotes, it will still not work since they are doing incorrect casing.

2) and their table names are case insensitive - adding quotes shouldn't be a problem, since the names are case insensitive. Should continue to work. @aaronburtle, could you please confirm this scenario?

Aniruddh25 (approved):
Approving existing changes. Any leftover testing should follow up in the subsequent issue tracked by @aaronburtle

PR #2617, opened 2025-03-14
2554 Enhance REST OTEL instrumentation with custom metrics and traces

### LINKED ISSUE #2554: 🥕⭐ [Enhancement]: Custom DAB OTEL Traces & Counters
## Now that we have OTEL, add DAB info

The current OTEL implementation only has generic/basic ASPNET information. 

## Traces:

https://github.com/Azure/data-api-builder/blob/jerry-otel-design/docs/design/otel-trace.md

```
Trace: DAB Request
│   ├── Tags:
│   │   ├── trace_id: <unique_trace_id>
│   │   ├── service.name: "Data API Builder"
│   │   ├── deployment.environment: "production" (or "development")
│
├── Activity: Request Setup
│   ├── Tags:
│   │   ├── http.method: "GET" | "POST" | "PUT" | "DELETE"
│   │   ├── action.type: "CREATE" | "READ" | "UPDATE" | "DELETE" | "EXECUTE"
│   │   ├── http.status_code: "200" | "400" | "500"
│   │   ├── http.url: "/api/entity"
│   │   ├── http.querystring: "?xyz" {only when mode = development}
│   │   ├── user.role: "Anonymous"
│   │   ├── api.type: "REST" | "GRAPHQL"
│   │   ├── api.entity: "User"
│   │   │
│   ├── Span: Validate Request Context
│   │   ├── Tags:
│   │   │   ├── validation.result: "Success" | "Failure"
│   │   ├── Event: Validation Error (if any)
│   │       ├── Tags: { error.message: "<error_details>", error.type: "<exception_type>" }
│   │
│   ├── Span: Process Request
│   │   │
│   │   ├── Event: Error (if any)
│   │       ├── Tags: { error.message: "<error_details>", error.type: "<exception_type>" }
│   │
│   ├── Span: Format Response
│   │   │
│   │   ├── Event: Response Sent
│   │       ├── Tags: { event.name: "Response Sent", timestamp: <time> }
│
└── Activity: Database Query
    ├── Tags:
    │   ├── data-source.name: {data-source.telemetry.name}
    │   ├── data-source.type: "mssql" | "mysql" | "postgresql" | "cosmos" | "sqldw"
    │   ├── db.operation: "SELECT" | "INSERT" | "UPDATE" | "DELETE" | "EXECUTE"
    │   ├── db.statement: "<SQL_query_text>" {only when mode = development}
    │
    ├── Span: Data Operation
    │   │   ├── query.rows_returned: <num_rows>
    │   │   ├── query.mb_returned: <double>
    │   │   ├── query.used_cache: true | false
    │   │
    │   ├── Event: Cache Returned
    │   │
    │   ├── Event: Connection Opened
    │   │
    │   ├── Event: Query Executed
    │   │
    │   ├── Event: Connection Opened
    │   │
    │   ├── Event: Error (if any)
    │       ├── Tags: { error.message: "<error_details>", error.type: "<exception_type>" }
```

#### TelemetryHelper.cs

```csharp
using System;
using System.Diagnostics.Metrics;

public static class TelemetryHelper
{
    private static readonly Meter Meter = new("Dab.Metrics");
    private static readonly Counter<long> RequestsTotal = Meter.CreateCounter<long>("requests_total");
    private static readonly UpDownCounter<long> RequestsActive = Meter.CreateUpDownCounter<long>("requests_active");
    private static readonly Counter<long> RequestErrors = Meter.CreateCounter<long>("request_errors");

    public static void TrackRequest(string method, string statusCode, string endpoint, string apiType)
    {
        RequestsTotal.Add(1, 
            new("method", method), 
            new("status_code", statusCode), 
            new("endpoint", endpoint), 
            new("api_type", apiType));
    }

    public static void TrackError(string method, string statusCode, string endpoint, string apiType, Exception ex)
    {
        RequestErrors.Add(1, 
            new("method", method), 
            new("status_code", statusCode), 
            new("endpoint", endpoint), 
            new("api_type", apiType), 
            new("error_type", ex.GetType().Name));
    }

    public static void IncrementActiveRequests() => RequestsActive.Add(1);

    public static void DecrementActiveRequests() => RequestsActive.Add(-1);
}
```

Then use it like.

```csharp
[HttpGet]
public IActionResult GetAsync(...)
{
    TelemetryHelper.IncrementActiveRequests(); 

   try
    {
        // Simulating success
        TelemetryHelper.TrackRequest("GET", "200", "User", "REST");
        return Ok(new { Message = "Hello, World!" });
    }
    catch (Exception ex)
    {
        TelemetryHelper.TrackError("GET", "500", "User", "REST", ex);
        return StatusCode(500, new { Error = "An error occurred" });
    }
    finally
    {
        TelemetryHelper.DecrementActiveRequests();
    }
}
```

tommasodotNET:
## Why make this change?

- Closes #2554
  - Enhances OTEL instrumentation with custom traces and metrics for the REST APIs

## What is this change?

This PR enhances the OTEL instrumentation for the REST APIs by adding custom traces and metrics.

I have removed ASP NET Core standard instrumentation since it does not provide great value given the custom nature of the webservice. I have written two main Helper classes: `TelemetryMetricsHelper` and `TelemetryTracesHelper` to provide a single point of management for custom traces and metrics.

Metrics can be filtered for `status_code`, `api_type`, `endpoint` and `method`.

I have also fixed the loggings which are now sent to the configured OTEL endpoint.

### Logs
image

### Metrics
Screenshot 2025-03-14 190730

### Traces
image (2)
image (3)
image (4)

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

To test everything locally I recommend using [this repo](https://github.com/tommasodotNET/dab-workbench) that allows to run the local build of the dab cli and send metrics to the .NET Aspire OTEL endoint.


JerryNixon:
@Aniruddh25 is March realistic?

src/Service/Controllers/RestController.cs
```diff
@@ -231,6 +277,12 @@ private async Task<IActionResult> HandleOperation(
                     HttpContextExtensions.GetLoggerCorrelationId(HttpContext));
 
                 Response.StatusCode = (int)ex.StatusCode;
+                if (activity is not null)
+                {
+                    activity.TrackRestControllerActivityFinishedWithWithException(ex, Response.StatusCode);
+                }
```
aaronpowell, line 283: Can't this be simplified as:

```suggestion
                activity?.TrackRestControllerActivityFinishedWithWithException(ex, Response.StatusCode);
```

src/Service/Controllers/RestController.cs
```diff
@@ -241,11 +293,28 @@ private async Task<IActionResult> HandleOperation(
                     HttpContextExtensions.GetLoggerCorrelationId(HttpContext));
 
                 Response.StatusCode = (int)HttpStatusCode.InternalServerError;
+                if(activity is not null)
+                {
+                    activity.TrackRestControllerActivityFinishedWithWithException(ex, Response.StatusCode);
+                }
```
aaronpowell, line 299: Can't this be simplified as:

```suggestion
                activity?.TrackRestControllerActivityFinishedWithWithException(ex, Response.StatusCode);
```

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+            if(activity.IsAllDataRequested)
+            {
+                activity.SetTag("data-source.type", databaseType);
+                activity.SetTag("data-source.name", dataSourceName);
+            }
+
+        }
+
+        public static void TrackRestControllerActivityFinished(this Activity activity,
+            int statusCode)
```
RubenCerna2079, line 52:
```suggestion
        public static void TrackRestControllerActivityFinished(
            this Activity activity,
            int statusCode)
```

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            if (activity.IsAllDataRequested)
+            {
+                activity.SetTag("status.code", statusCode);
+            }
+        }
+
+        public static void TrackRestControllerActivityFinishedWithWithException(this Activity activity,
+            Exception ex,
```
RubenCerna2079, line 61:
```suggestion
        public static void TrackRestControllerActivityFinishedWithWithException(
            this Activity activity,
            Exception ex,
```

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Diagnostics;
+using OpenTelemetry.Trace;
+
+namespace Azure.DataApiBuilder.Service.Telemetry
+{
+    public static class TelemetryTracesHelper
+    {
+        public static readonly ActivitySource DABActivitySource = new("DataApiBuilder");
+
+        public static void TrackRestControllerActivityStarted(this Activity activity,
```
RubenCerna2079, line 14: Please add comments that describe what this functions are doing in order to have more context

src/Service/Controllers/RestController.cs
```diff
@@ -241,11 +289,25 @@ private async Task<IActionResult> HandleOperation(
...
                     HttpStatusCode.InternalServerError);
             }
+            finally
+            {
+                stopwatch.Stop();
+                TelemetryMetricsHelper.TrackRequestDuration(HttpContext.Request.Method, Response.StatusCode, route, "REST", stopwatch.Elapsed.TotalMilliseconds);
+                if (activity is not null && activity.IsAllDataRequested)
+                {
+                    activity.Dispose();
+                }
```
RubenCerna2079, line 307: Since this pattern is repeated you might want to add it as a function in `TelemetryTracesHelper`

src/Service/Startup.cs
```diff
@@ -116,31 +117,46 @@ public void ConfigureServices(IServiceCollection services)
...
                 && runtimeConfig.Runtime.Telemetry.OpenTelemetry.Enabled)
             {
+                services.Configure<OpenTelemetryLoggerOptions>(options =>
+                {
+                    options.IncludeScopes = true;
+                    options.ParseStateValues = true;
+                    options.IncludeFormattedMessage = true;
+                });
                 services.AddOpenTelemetry()
+                .WithLogging(logging =>
```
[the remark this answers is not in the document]
tommasodotNET: Because the current logic for the logger factory seems to not be working. When we reach the logger factory code, the runtime hasn't been read yet and logging was not being enabled. In the previous version I've used the same logic as for AppInsights and via debug found out that the check on OTEL logging enable was false.

src/Service/Startup.cs
```diff
@@ -116,31 +117,46 @@ public void ConfigureServices(IServiceCollection services)
...
-                        .AddAspNetCoreInstrumentation()
-                        .AddHttpClientInstrumentation()
                         .AddOtlpExporter(configure =>
                         {
                             configure.Endpoint = new Uri(runtimeConfig.Runtime.Telemetry.OpenTelemetry.Endpoint!);
                             configure.Headers = runtimeConfig.Runtime.Telemetry.OpenTelemetry.Headers;
                             configure.Protocol = OtlpExportProtocol.Grpc;
                         })
-                        .AddRuntimeInstrumentation();
+                        .AddMeter(TelemetryMetricsHelper.MeterName);
```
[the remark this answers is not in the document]
tommasodotNET: We need to remove standard ASP NET Core instrumentation because it relies on standard metrics for ASP NET Core apps that don't really match the logic we use in DAB (e.g. no filters available, no clarity on path, etc)

src/Service/Controllers/RestController.cs
```diff
@@ -211,8 +232,25 @@ private async Task<IActionResult> HandleOperation(
```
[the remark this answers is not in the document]
tommasodotNET: I've updated the title. The description already mentioned that. Do you think we could improve the description further?

src/Service/Controllers/RestController.cs
tommasodotNET: To do that, I would need to use the activity object, which will result in a warning because activity might be null. That warning could be solved by checking that activity is not null. Therefore I think it's easier to leave it as it is and avoid a double check.

src/Service/Program.cs
```diff
@@ -161,22 +162,6 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, Tele
                         .AddFilter<ApplicationInsightsLoggerProvider>(category: string.Empty, logLevel);
                     }
```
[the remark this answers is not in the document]
tommasodotNET: This is linked to the question in the `Startup.cs`.

src/Service/Controllers/RestController.cs
```diff
@@ -179,14 +185,30 @@ public async Task<IActionResult> UpsertIncremental(
...
+        /// <param name="method">The method.</param>
         /// <param name="route">The entire route.</param>
         /// <param name="operationType">The kind of operation to handle.</param>
         private async Task<IActionResult> HandleOperation(
             string route,
             EntityActionOperation operationType)
         {
-            try
+            var stopwatch = Stopwatch.StartNew();
+            using Activity? activity = TelemetryTracesHelper.DABActivitySource.StartActivity($"{HttpContext.Request.Method} {route.Split('/')[1]}");
```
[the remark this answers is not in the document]
tommasodotNET: You are right, @RubenCerna2079. No need to do that. If there is no additional path, it won't create the activity, but won't cause any issue.

src/Service/Controllers/RestController.cs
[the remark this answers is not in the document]
RubenCerna2079: I think just restating in the `What is this change?` section should be more than enough

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -36,7 +51,14 @@ public static void TrackRestControllerActivityStarted(this Activity activity,
...
         }
 
-        public static void TrackQueryActivityStarted(this Activity activity,
+        /// <summary>
+        /// Tracks the start of a query activity.
+        /// </summary>
+        /// <param name="activity">The activity instance.</param>
+        /// <param name="databaseType">The type of database being queried.</param>
+        /// <param name="dataSourceName">The name of the data source being queried.</param>
+        public static void TrackQueryActivityStarted(
```
RubenCerna2079, line 60: Out of curiosity, why does the `QueryActivity` doesn't need a finished function in comparison to the `RestControllerActivity`?

src/Service/Startup.cs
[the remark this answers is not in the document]
RubenCerna2079: I think this is worth discussing more, if there are any more errors that you found in the logic for the logger factory, please let us know so that we can solve them.

src/Service/Telemetry/TelemetryTracesHelper.cs
tommasodotNET: That is because we don't need to append any tag to this activity when it finishes, so I didn't create a specific function for that. If and when we will have some tags to add when the activity is finished, we can create a `TrackQueryActivityFinished` method.

src/Service/Controllers/RestController.cs
```diff
@@ -185,8 +191,25 @@ private async Task<IActionResult> HandleOperation(
             string route,
             EntityActionOperation operationType)
         {
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            // This activity tracks the entire REST request.
+            using Activity? activity = TelemetryTracesHelper.DABActivitySource.StartActivity($"{HttpContext.Request.Method} {route.Split('/')[1]}");
+            if(activity is not null)
+            {
```
RubenCerna2079, line 198:
```suggestion
            if (activity is not null)
            {
```

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+            this Activity activity,
+            string databaseType,
+            string dataSourceName)
+        {
+            if(activity.IsAllDataRequested)
```
RubenCerna2079, line 65:
```suggestion
            if (activity.IsAllDataRequested)
```

RubenCerna2079:
@tommasodotNET just wanted to let you know that I updated the branch with the newest changes done to `main`

tommasodotNET:
> @tommasodotNET just wanted to let you know that I updated the branch with the newest changes done to `main`

Thanks. I see MS_SQL Tests are failing. Can't really figure out why, though...

src/Service/Controllers/RestController.cs
```diff
@@ -185,8 +191,25 @@ private async Task<IActionResult> HandleOperation(
             string route,
             EntityActionOperation operationType)
         {
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            // This activity tracks the entire REST request.
+            using Activity? activity = TelemetryTracesHelper.DABActivitySource.StartActivity($"{HttpContext.Request.Method} {route.Split('/')[1]}");
```
RubenCerna2079, line 196: This is the section that is causing the failure and it is related to what @sezal98 was mentioning. Currently the `route.Split('/')[1]` is expected to retrieve the second value of an array that is split by `/`, but in the test cases that it is failing we have strings that don't have any `/` for example, `favicon.io`.
This is causing the failure as the current logic tries to retrieve the second element of an array that only has an element at `[0]`.

RubenCerna2079:
@tommasodotNET I figured out why the tests are failing, I left a comment in the section that is causing the failure and the reason for it.

src/Service/Controllers/RestController.cs
tommasodotNET: I see. I added a check on the length of the split result

src/Service/Controllers/RestController.cs
```diff
@@ -185,8 +191,25 @@ private async Task<IActionResult> HandleOperation(
             string route,
             EntityActionOperation operationType)
         {
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            // This activity tracks the entire REST request.
+            using Activity? activity = TelemetryTracesHelper.DABActivitySource.StartActivity($"{HttpContext.Request.Method} {(route.Split('/').Length > 1 ? route.Split('/')[1] : "")}");
```
RubenCerna2079, line 196:
```suggestion
            using Activity? activity = TelemetryTracesHelper.DABActivitySource.StartActivity($"{HttpContext.Request.Method} {(route.Split('/').Length > 1 ? route.Split('/')[1] : string.Empty)}");
```

RubenCerna2079 (approved):
LGTM! I just left one more comment. Thank you for working on this feature, it is really cool.

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+            string? userRole,
+            string apiType)
+        {
+            if (activity.IsAllDataRequested)
+            {
+                activity.SetTag("http.method", httpMethod);
+                activity.SetTag("user-agent", userAgent);
+                activity.SetTag("action.type", actionType);
+                activity.SetTag("http.url", httpURL);
+                if (queryString is not null && queryString != string.Empty)
```
Aniruddh25, line 44: could use `!string.IsNullOrEmpty(queryString)`

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    activity.SetTag("http.querystring", queryString);
+                }
+
+                activity.SetTag("user.role", userRole);
```
Aniruddh25, line 49: Do we need a similar check for `!IsNullOrEmpty` for `user.role`?

src/Service/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Tracks the completion of a REST controller activity with an exception.
+        /// </summary>
+        /// <param name="activity">The activity instance.</param>
+        /// <param name="ex">The exception that occurred.</param>
+        /// <param name="statusCode">The HTTP status code of the response.</param>
+        public static void TrackRestControllerActivityFinishedWithWithException(
```
Aniruddh25, line 94:
```suggestion
        public static void TrackRestControllerActivityFinishedWithException(
```

src/Service/Startup.cs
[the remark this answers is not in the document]
Aniruddh25: +1 to @RubenCerna2079, please let us know if you find more errors in our current logic of logger factory for AppInsights or anything else.

src/Service/Controllers/RestController.cs
```diff
@@ -185,8 +191,25 @@ private async Task<IActionResult> HandleOperation(
...
+            if (activity is not null)
+            {
+                activity.TrackRestControllerActivityStarted(
+                    HttpContext.Request.Method,
+                    HttpContext.Request.Headers["User-Agent"].ToString(),
+                    operationType.ToString(),
+                    route,
+                    HttpContext.Request.QueryString.ToString(),
+                    HttpContext.User.FindFirst("role")?.Value,
```
Aniruddh25, line 205: We also check for the value of the request header `X-MS-API-ROLE` (if set) to determine what the role the request should be executed as.. instead of finding the first using `FindFirst`. 
Can we use the value of this header here and default to `FindFirst` if that doesnt exist?

Aniruddh25 (approved):
LGTM, Thank you for adding this support!

Aniruddh25:
> Copilot reviewed 7 out of 8 changed files in this pull request and generated no comments.
> 
> Files not reviewed (1)
> * **src/Service/Azure.DataApiBuilder.Service.csproj**: Language not supported
> 
> Comments suppressed due to low confidence (2)
> **src/Service/Telemetry/TelemetryTracesHelper.cs:94**
> 
> * The method name 'TrackRestControllerActivityFinishedWithWithException' contains a duplicate 'With'. Consider renaming it to 'TrackRestControllerActivityFinishedWithException' for clarity.
> 
> ```
> public static void TrackRestControllerActivityFinishedWithWithException(
> ```
> 
> **src/Service/Controllers/RestController.cs:249**
> 
> * Manual disposal of 'queryActivity' is redundant since it is declared within a 'using' block, which will automatically dispose it. Removing the explicit call can prevent potential unexpected behavior from double disposal.
> 
> ```
> if (queryActivity is not null && queryActivity.IsAllDataRequested)
>                     {
>                         queryActivity.Dispose();
>                     }
> ```

@tommasodotNET Seems like queryActivity.Dispose() could be removed? Can it?

tommasodotNET:
Hey @Aniruddh25, I've addressed copilot review

src/Service/Telemetry/TelemetryTracesHelper.cs
tommasodotNET: Is this a default style for the repo?

src/Service/Startup.cs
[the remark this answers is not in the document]
tommasodotNET: We realized app insights was fine

src/Service/Startup.cs
```diff
@@ -125,17 +126,17 @@ public void ConfigureServices(IServiceCollection services)
...
-                    logging.SetResourceBuilder(ResourceBuilder.CreateDefault().AddService(runtimeConfig.Runtime.Telemetry.OpenTelemetry.ServiceName!))
-                    .AddOtlpExporter(configure =>
-                    {
-                        configure.Endpoint = new Uri(runtimeConfig.Runtime.Telemetry.OpenTelemetry.Endpoint!);
-                        configure.Headers = runtimeConfig.Runtime.Telemetry.OpenTelemetry.Headers;
-                        configure.Protocol = OtlpExportProtocol.Grpc;
-                    });
-
-                })
+                //.WithLogging(logging =>
```
Aniruddh25, line 129: if we dont need this logging, please remove the commented out code

Aniruddh25:
@tommasodotNET, waiting on all tests to pass..

RubenCerna2079:
@tommasodotNET I made some changes to your PR, it was mostly changing methods to use more specific classes instead of `string` where I found it possible. I also fixed small formatting issues that were causing the unit tests to fail.

RubenCerna2079:
@tommasodotNET I also found out a new issue with the tests related to a format issue in the `dab-config.CosmosDb_NoSql.json`, so I fixed it as it seems that was the only test failing.

tommasodotNET:
Thanks @RubenCerna2079 for your support

src/Service/Controllers/RestController.cs
```diff
@@ -271,9 +274,11 @@ private async Task<IActionResult> HandleOperation(
                     HttpContextExtensions.GetLoggerCorrelationId(HttpContext));
 
                 Response.StatusCode = (int)ex.StatusCode;
-                activity?.TrackRestControllerActivityFinishedWithException(ex, Response.StatusCode);
+                activity?.TrackRestControllerActivityFinishedWithException(ex, Enum.Parse<HttpStatusCode>(Response.StatusCode.ToString(), ignoreCase: true));
```
Aniruddh25, line 277: could reuse the `httpStatusCode` local variable here if we define it before this call

Aniruddh25 (approved):
LGTM, given all tests are passing. This has come a long way! 

ship it!!

src/Service/Controllers/RestController.cs
RubenCerna2079: I think we can simplify it, but not reuse the local variable since there are two different `catch` for different exceptions

Aniruddh25 (approved):

RubenCerna2079 (approved):

Issue #2585, opened 2025-02-14
Add filters to schema


RubenCerna2079:
Figure how to add multiple filters without the need to add them all to the base schema

Issue #2563, opened 2025-02-13
Create Log Level Filters


RubenCerna2079:
Allow the user to add filters to have more fine tuning on which classes have a specific level of logging. 
This allows the user to have more control over what they want to know from specific sections of DAB.

e.g.
```
"log-level": {
      "default": "information",
      "microsoft": "warning",
      "microsoft.hosting.lifetime": "error"
}
```

PR #2620, opened 2025-03-14
Log Level Filters

RubenCerna2079:
## Why make this change?

This change solves tasks, #2584, #2585, #2586, #2587, #2588, #2569. 
All of them are part of the bigger task of #2563 which adds filters for logging.

## What is this change?
This change allows the users to add more specific logging filters to their config file, this means that log-levels meant for a specific class will take priority over the default log-level.
In the following example, even if the default is in `debug`, the logging for `IQueryExecutor` will be in `information`.
```
"Azure.DataApiBuilder.Core.Resolvers.IQueryExecutor": "information",
"default": "debug"
```
Specific Changes:
- Changed the schema file so that it allows for different keywords that represent the classes, allowing each of them to have a different log-level. It also allows the user to add a general log-level through the use of the keyword `default` 
- Updated DAB Validate so that it checks if the keywords used in the config file are valid classes that we use to give logging information to the users.
- Implement logic that deserializes the new properties in the config file and stores them inside of the `TelemetryOptions` class and it uses the saved log-levels to dictate if a specific logger uses a different log-level from the default. 

## How was this tested?
New tests where created, in order to ensure that the new filters are deserialized properly from the config file, and that adding filters that do not fit the requirements throw an error.

- [ ] Integration Tests
- [X] Unit Tests


src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -129,6 +130,28 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
         }
     }
 
+    /// <summary>
+    /// Only certain classes can have different log levels, this function ensures that only those classes are used in the config file.
+    /// </summary>
+    public void ValidateLoggerFilters(RuntimeConfig runtimeConfig)
```
RubenCerna2079, line 136: Still not entirely sure if we only want to ensure that the validation of the logging filters is only done here or if we want to change the config schema so that it enforces those rules as well.
Open to suggestions :)

src/Config/ObjectModel/LoggerFilters.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
```
RubenCerna2079, line 1: Move to DataApiBuilder.Config file?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1323,4 +1346,24 @@ private bool IsValidPermissionAction(EntityActionOperation action, Entity entity
...
         }
     }
+
+    /// <summary>
+    /// Returns whether the log-level keyword is valid or not.
+    /// </summary>
+    /// <param name="loggerFilter">String keyword that comes from log-level in config file</param>
+    private static bool IsLoggerFilterValid(string loggerFilter)
+    {
+        return loggerFilter == LoggerFilters.RUNTIMECONFIGVALIDATORFILTER ||
```
abhishekkumams, line 1356: its very hard to read, can you add `_` like `IQUERY_EXECUTER_FILTER`

src/Config/ObjectModel/TelemetryOptions.cs
```diff
@@ -2,14 +2,15 @@
...
 using System.Text.Json.Serialization;
+using Microsoft.Extensions.Logging;
 
 namespace Azure.DataApiBuilder.Config.ObjectModel;
 
 /// <summary>
 /// Represents the options for telemetry.
 /// </summary>
-public record TelemetryOptions(ApplicationInsightsOptions? ApplicationInsights = null, OpenTelemetryOptions? OpenTelemetry = null, LogLevelOptions? LoggerLevel = null)
+public record TelemetryOptions(ApplicationInsightsOptions? ApplicationInsights = null, OpenTelemetryOptions? OpenTelemetry = null, SortedList<string, LogLevel?>? LoggerLevel = null)
```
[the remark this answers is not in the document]
RubenCerna2079: A sorted list has a key and a value, while a list only has a set of values and we check if the key that it is using is valid. The other option I would be open to using instead is a dictionary.

schemas/dab.draft.schema.json
```diff
@@ -376,9 +376,9 @@
               "type": "object",
               "description": "Global configuration of log level",
               "additionalProperties": false,
-              "properties": {
-                "level": {
-                  "description": "Defines logging severity levels, when in default value it will set logging level based on 'host: mode' property",
+              "patternProperties": {
+                "^[a-zA-Z_]\\w{0,}(\\.\\w{1,}){0,}$": {
+                  "description": "Defines logging severity levels for specific classes, when in null value it will set logging level based on 'host: mode' property",
```
Aniruddh25, line 381:
```suggestion
                  "description": "Defines logging severity levels for specific classes, when `null` it will set logging level based on 'host: mode' property",
```

src/Config/ObjectModel/LoggerFilters.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel
+{
+    public class LoggerFilters
+    {
+        public const string RUNTIME_CONFIG_VALIDATOR_FILTER = "Azure.DataApiBuilder.Core.Configurations.RuntimeConfigValidator";
```
Aniruddh25, line 8: Instead of hard coding names of all possible filters - which are fully classified class names here, cant we use for e.g. `nameof(RuntimeConfigValidator)` to get the string name of the class where these constants are used?

src/Config/ObjectModel/TelemetryOptions.cs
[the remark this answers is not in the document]
Aniruddh25: Lets use the dictionary - marking this as unresolved since the changes addressing this are not pushed yet.

src/Service/Startup.cs
```diff
@@ -240,12 +240,12 @@ public void ConfigureServices(IServiceCollection services)
...
             services.AddSingleton<ILogger<IAuthorizationHandler>>(implementationFactory: (serviceProvider) =>
             {
-                ILoggerFactory? loggerFactory = CreateLoggerFactoryForHostedAndNonHostedScenario(serviceProvider);
+                ILoggerFactory? loggerFactory = CreateLoggerFactoryForHostedAndNonHostedScenario(serviceProvider, LoggerFilters.IAUTHORIZATION_HANDLER_FILTER);
                 return loggerFactory.CreateLogger<IAuthorizationHandler>();
             });
             services.AddSingleton<ILogger<IAuthorizationResolver>>(implementationFactory: (serviceProvider) =>
             {
-                ILoggerFactory? loggerFactory = CreateLoggerFactoryForHostedAndNonHostedScenario(serviceProvider);
+                ILoggerFactory? loggerFactory = CreateLoggerFactoryForHostedAndNonHostedScenario(serviceProvider, LoggerFilters.IAUTHORIZATION_RESOLVER_FILTER);
```
Aniruddh25, line 248: this could be `nameof(IAuthorizationResolver)` or something like that to get the full namespace name

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -572,11 +572,22 @@ public uint GetPaginationLimit(int? first)
...
-        Runtime is null ||
-        Runtime.Telemetry is null ||
-        Runtime.Telemetry.LoggerLevel is null ||
-        Runtime.Telemetry.LoggerLevel.Value is null;
+    public bool IsLogLevelNull()
+    {
+        for (int i = 0; i < Runtime?.Telemetry?.LoggerLevel?.Count; i++)
+        {
+            string? keyVal = Runtime?.Telemetry?.LoggerLevel?.GetKeyAtIndex(i);
+            if (keyVal == null)
```
Aniruddh25, line 580: Are we retrieving the key or the value here? why is this string names `keyVal`?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -585,9 +596,16 @@ Runtime.Telemetry.LoggerLevel is null ||
     /// If host mode is Development, return `LogLevel.Debug`, else
     /// for production returns `LogLevel.Error`.
     /// </summary>
-    public static LogLevel GetConfiguredLogLevel(RuntimeConfig runtimeConfig)
+    public static LogLevel GetConfiguredLogLevel(RuntimeConfig runtimeConfig, string loggerFilter = "")
     {
-        LogLevel? value = runtimeConfig.Runtime?.Telemetry?.LoggerLevel?.Value;
+        LogLevel? value = null;
+        runtimeConfig.Runtime?.Telemetry?.LoggerLevel?.TryGetValue(loggerFilter, out value);
```
Aniruddh25, line 602: What is the time complexity of retrieving the value of the said LoggerFilter from the SortedList LoggerLevel? How can we do better?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -129,6 +130,28 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
         }
     }
 
+    /// <summary>
+    /// Only certain classes can have different log levels, this function ensures that only those classes are used in the config file.
+    /// </summary>
+    public void ValidateLoggerFilters(RuntimeConfig runtimeConfig)
+    {
+        if (runtimeConfig.Runtime!.Telemetry is not null && runtimeConfig.Runtime.Telemetry.LoggerLevel is not null)
```
Aniruddh25, line 138: how can we be sure Runtime wont be null here?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -129,6 +130,28 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
...
+        {
+            SortedList<string, LogLevel?> loggerLevelOptions = runtimeConfig.Runtime.Telemetry.LoggerLevel;
+
+            for (int i = 0; i < loggerLevelOptions.Count; i++)
+            {
+                if (!IsLoggerFilterValid(loggerLevelOptions.GetKeyAtIndex(i)))
+                {
+                    HandleOrRecordException(new DataApiBuilderException(
+                    message: $"Log level {loggerLevelOptions.GetKeyAtIndex(i)} needs to be of a valid log class.",
+                    statusCode: HttpStatusCode.BadRequest,
```
Aniruddh25, line 148: How is this a BadRequest? This is not a http request. Isnt this part of the validation workflow? 

Shouldnt we use the current logger in scope to log an Error like the other validation mechanisms are doing?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3588,24 +3588,93 @@ public void LogLevelSerialization(LogLevel expectedLevel)
...
+        [DataRow(LoggerFilters.DEFAULT_FILTER)]
+        public void ValidLogLevelFilters(string loggingFilter)
+        {
+            RuntimeConfig configWithCustomLogLevel = InitializeRuntimeWithLogLevel(LogLevel.Debug, loggingFilter);
+
+            string configWithCustomLogLevelJson = configWithCustomLogLevel.ToJson();
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(configWithCustomLogLevelJson, out RuntimeConfig deserializedRuntimeConfig));
+
+            SortedList<string, LogLevel?> actualLoggerLevel = deserializedRuntimeConfig.Runtime.Telemetry.LoggerLevel;
+            Assert.IsTrue(actualLoggerLevel.ContainsKey(loggingFilter) && actualLoggerLevel.Count == 1);
```
Aniruddh25, line 3623: why do we assume actualLoggerLevel is not null? Lets assert for that first then check for the contains key

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3588,24 +3588,93 @@ public void LogLevelSerialization(LogLevel expectedLevel)
...
+        }
```
Aniruddh25, line 3624: Why are we not asserting the actual value of the log-level with the specified logger filter?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3588,24 +3588,93 @@ public void LogLevelSerialization(LogLevel expectedLevel)
...
+        [DataRow(LoggerFilters.CONFIGURATION_CONTROLLER_FILTER)]
+        [DataRow(LoggerFilters.IAUTHORIZATION_HANDLER_FILTER)]
+        [DataRow(LoggerFilters.IAUTHORIZATION_RESOLVER_FILTER)]
+        [DataRow(LoggerFilters.DEFAULT_FILTER)]
+        public void ValidLogLevelFilters(string loggingFilter)
+        {
+            RuntimeConfig configWithCustomLogLevel = InitializeRuntimeWithLogLevel(LogLevel.Debug, loggingFilter);
+
+            string configWithCustomLogLevelJson = configWithCustomLogLevel.ToJson();
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(configWithCustomLogLevelJson, out RuntimeConfig deserializedRuntimeConfig));
```
Aniruddh25, line 3620: Does the validation check run even when we parse the configuration?

Aniruddh25 (changes requested):
Waiting on using a `Dictionary` instead of `SortedList`

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: We are retrieving the key, will change name to key

src/Core/Configurations/RuntimeConfigValidator.cs
RubenCerna2079: Runtime is a required object, so it should already be confirmed that it is not null beforehand

src/Core/Configurations/RuntimeConfigValidator.cs
RubenCerna2079: I think it is better to send out an exception, although it is true that this is not an http exception

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: Will change it to dictionary in order to have a smaller time complexity

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: You mean that the variable or the value inside actualLoggerLevel?

src/Config/ObjectModel/LoggerFilters.cs
RubenCerna2079: Cannot do this due to dependency issues, if I find a way to solve this I will implement it

src/Config/LoggerFilters.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config
+{
+    public class LoggerFilters
+    {
+        public static List<string> validFilters = new();
+
+        public static void AddFilter(string? loggerFilter)
```
Aniruddh25, line 10: why is the parameter nullable loggerFilter? are there any callers that could invoke this with a null value?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1323,4 +1344,39 @@ private bool IsValidPermissionAction(EntityActionOperation action, Entity entity
...
+    /// E.g. Azure.DataApiBuilder is valid. While Azure.DataA is invalid.
+    /// </summary>
+    /// <param name="loggerFilter">String keyword that comes from log-level in config file</param>
+    private static bool IsLoggerFilterValid(string loggerFilter)
+    {
+        string[] loggerSub = loggerFilter.Split('.');
+        for (int i = 0; i < LoggerFilters.validFilters.Count; i++)
+        {
+            bool isValid = true;
+            string[] validFiltersSub = LoggerFilters.validFilters[i].Split('.');
```
Aniruddh25, line 1360: if the `LoggerFilters.validFilters` are used in `Core`  why can't we define it inside Core, why does it need to live in the Config project?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -129,6 +131,25 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
...
+    {
+        if (runtimeConfig.Runtime?.Telemetry is not null && runtimeConfig.Runtime.Telemetry.LoggerLevel is not null)
+        {
+            Dictionary<string, LogLevel?> loggerLevelOptions = runtimeConfig.Runtime.Telemetry.LoggerLevel;
+
+            foreach (KeyValuePair<string, LogLevel?> logger in loggerLevelOptions)
+            {
+                if (!IsLoggerFilterValid(logger.Key))
+                {
+                    throw new NotSupportedException($"Log level {logger.Key} needs to be of a valid log class.");
```
Aniruddh25, line 147:
```suggestion
                    throw new NotSupportedException($"Log level filter {logger.Key} needs to be of a valid log class.");
```

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -572,11 +572,26 @@ public uint GetPaginationLimit(int? first)
     /// <summary>
     /// Checks if the property log-level or its value are null
     /// </summary>
-    public bool IsLogLevelNull() =>
-        Runtime is null ||
-        Runtime.Telemetry is null ||
-        Runtime.Telemetry.LoggerLevel is null ||
-        Runtime.Telemetry.LoggerLevel.Value is null;
+    public static bool IsLogLevelNull(RuntimeConfig runtimeConfig)
```
Aniruddh25, line 575: Why is this made static?

src/Service/Startup.cs
```diff
@@ -822,5 +828,24 @@ private static bool IsUIEnabled(RuntimeConfig? runtimeConfig, IWebHostEnvironmen
         {
             return (runtimeConfig is not null && runtimeConfig.IsDevelopmentMode()) || env.IsDevelopment();
         }
+
+        /// <summary>
+        /// Adds all of the class namespaces that have loggers that the user is able to change
+        /// </summary>
+        private static void AddValidFilters()
```
Aniruddh25, line 835: why do we keep this private? We can make it a public function that can be called in `Cli` so that we dont need to duplicate the definition in Cli project.

src/Config/LoggerFilters.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config
+{
+    public class LoggerFilters
```
Aniruddh25, line 6: If we make this a static class, the ValidFilters property might get loaded at initialization time of any project that depends on `Core` (if we move it to the `Core` project), without us needing to call the `AddFilter` function. Can you please try that?

src/Config/ObjectModel/RuntimeConfig.cs
Aniruddh25: I see that GetConfiguredLogLevel is static thats why. But cant we make that function non-static instead of making this one static? Especially since both take runtimeConfig as an object parameter input.

src/Service.Tests/TestHelper.cs
```diff
@@ -356,5 +363,25 @@ public static string AddPropertiesToJson(string configuration, string entityProp
             });
             return configurationJson.ToString();
         }
+
+        /// <summary>
+        /// Adds all of the class namespaces that have loggers that the user is able to change
+        /// This function is used when testing for validation
+        /// </summary>
+        public static void AddValidFilters()
```
Aniruddh25, line 371: we should not add this function in the test project. When testing, test the function defined in the startup project.

src/Config/LoggerFilters.cs
RubenCerna2079: The typeof(class).FullName has the possibility of invoking a null value, which is the reason why put that parameter as nullable.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3709,24 +3711,130 @@ public void LogLevelSerialization(LogLevel expectedLevel)
...
+        [DataRow(LogLevel.Critical, typeof(ComprehensiveHealthReportResponseWriter))]
+        [DataRow(LogLevel.None, typeof(RestController))]
+        [DataRow(LogLevel.Trace, typeof(ClientRoleHeaderAuthenticationMiddleware))]
+        [DataRow(LogLevel.Debug, typeof(ConfigurationController))]
+        [DataRow(LogLevel.Information, typeof(IAuthorizationHandler))]
+        [DataRow(LogLevel.Warning, typeof(IAuthorizationResolver))]
+        public void ValidLogLevelFilters(LogLevel logLevel, Type loggingType)
+        {
+            string loggingFilter = loggingType.FullName;
+            TestHelper.AddValidFilters();
```
Aniruddh25, line 3740: We should NOT call AddValidFilters from the test. We should ensure that the RuntimeConfigValidator can validate it on its own - defining the LoggerFilters class `static`  might help here.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3709,24 +3711,130 @@ public void LogLevelSerialization(LogLevel expectedLevel)
...
+        [TestCategory(TestCategory.MSSQL)]
+        [DataRow(LogLevel.Trace, "default")]
+        [DataRow(LogLevel.Debug, "Azure")]
+        [DataRow(LogLevel.Information, "Azure.DataApiBuilder")]
+        [DataRow(LogLevel.Warning, "Azure.DataApiBuilder.Core")]
+        [DataRow(LogLevel.Error, "Azure.DataApiBuilder.Core.Configurations")]
+        [DataRow(LogLevel.Critical, "Azure.DataApiBuilder.Core.Resolvers")]
+        [DataRow(LogLevel.None, "Azure.DataApiBuilder.Core.Services")]
+        [DataRow(LogLevel.Trace, "Azure.DataApiBuilder.Service")]
+        [DataRow(LogLevel.Debug, "Azure.DataApiBuilder.Service.HealthCheck")]
```
Aniruddh25, line 3774: Why do we have a duplicate test? Lets add these as string input to the above test itself. You can modify the above test with typeof().FullName as the value of the DataRow.

RubenCerna2079: I tried doing that beforehand, and it throws an error, which is the reason why I separated them. The other option I can think of is to not use typeof().FullName and just write the entire name as a string for the test.

src/Service/Startup.cs
```diff
@@ -519,8 +520,13 @@ private static void EvictGraphQLSchema(IRequestExecutorResolver requestExecutorR
         /// minimum log level based on host.mode in the runtime config if available.
         /// Creates a logger factory with the minimum log level.
         /// </summary>
-        public static ILoggerFactory CreateLoggerFactoryForHostedAndNonHostedScenario(IServiceProvider serviceProvider)
+        public static ILoggerFactory CreateLoggerFactoryForHostedAndNonHostedScenario(IServiceProvider serviceProvider, string? loggerFilter)
```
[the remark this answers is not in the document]
RubenCerna2079: The typeof(class).FullName has the possibility of invoking a null value, which is the reason why put that parameter as nullable.

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Changed the function so that it accesses it from Startup, which is not the most ideal solution, but it is the best one we have so far.

src/Service.Tests/TestHelper.cs
RubenCerna2079: Deleted function from TestHelpee

src/Service/Startup.cs
RubenCerna2079: Changed function to public so that it is also used in the `Cli`

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: Changed the functions to non-static as well as all the instances in which those functions are used

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -585,15 +600,37 @@ Runtime.Telemetry.LoggerLevel is null ||
...
+
+        if (!IsLogLevelNull())
         {
-            return (LogLevel)value;
+            int max = 0;
+            string currentFilter = string.Empty;
+            foreach (KeyValuePair<string, LogLevel?> logger in Runtime!.Telemetry!.LoggerLevel!)
+            {
+                // Checks if the new key that is valid has more priority than the current key
+                if (logger.Key.Length > max && loggerFilter.StartsWith(logger.Key))
```
sezal98, line 613: Why do we need to check length here?

src/Config/ObjectModel/TelemetryOptions.cs
```diff
@@ -2,14 +2,15 @@
...
 
 /// <summary>
 /// Represents the options for telemetry.
 /// </summary>
-public record TelemetryOptions(ApplicationInsightsOptions? ApplicationInsights = null, OpenTelemetryOptions? OpenTelemetry = null, LogLevelOptions? LoggerLevel = null)
+public record TelemetryOptions(ApplicationInsightsOptions? ApplicationInsights = null, OpenTelemetryOptions? OpenTelemetry = null, Dictionary<string, LogLevel?>? LoggerLevel = null)
 {
     [JsonPropertyName("log-level")]
-    public LogLevelOptions? LoggerLevel { get; init; } = LoggerLevel;
+    public Dictionary<string, LogLevel?>? LoggerLevel { get; init; } = LoggerLevel;
```
sezal98, line 15: Can you attach an example of how the config file would look with this change? what does the key here define

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1323,4 +1343,39 @@ private bool IsValidPermissionAction(EntityActionOperation action, Entity entity
...
         }
     }
+
+    /// <summary>
+    /// Returns whether the log-level keyword is valid or not.
+    /// It does this by checking each section of the name of the class
+    /// E.g. Azure.DataApiBuilder is valid. While Azure.DataA is invalid.
+    /// </summary>
+    /// <param name="loggerFilter">String keyword that comes from log-level in config file</param>
+    private static bool IsLoggerFilterValid(string loggerFilter)
```
sezal98, line 1353: Can you elaborate a little more about this function. I am unable to understand what it does.

From what i understand, I feel that simple comaprision of the string should work instead of splitting and comparing each word

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3709,24 +3717,162 @@ public void LogLevelSerialization(LogLevel expectedLevel)
...
+        }
+
+        /// <summary>
+        /// General method that is used to test the valid log level filters.
+        /// </summary>
+        /// <param name="logLevel"></param>
+        /// <param name="loggingFilter"></param>
+        private static void ValidateLogLevelFilters(LogLevel logLevel, string loggingFilter)
+        {
+            Startup.AddValidFilters();
```
sezal98, line 3779: should this be part of test case? Shouldnt we test what is already implemented in the statup file?

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: We check length in order to ensure that the more specific key is used. For example, if we have `Azure: debug` and `Azure.DataApiBuilder: warning` in the config file, then it is expected that for all classes that start with `Azure.DataApiBuilder` they will use the log level `warning`.

src/Core/Configurations/RuntimeConfigValidator.cs
RubenCerna2079: I compare by splitting the string because we want to make sure that the final word is complete, if we make a simple comparison then there exists the possibility that the final word is not complete. For example, if we `Azu` then a simple comparison would think that is correct when the correct format should be `Azure`.

RubenCerna2079: Will add a bit more description to the comment section to make it more clear

src/Config/ObjectModel/TelemetryOptions.cs
RubenCerna2079: I added an example in the description of the PR

src/Config/Converters/LogLevelOptionsConverterFactory.cs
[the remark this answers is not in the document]
RubenCerna2079: What I meant with this part was that there is a change in logic, since before we made a factory for the log-level which is unnecessary.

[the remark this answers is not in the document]
RubenCerna2079: Changed description to make it easier to understand the change that was implemented

src/Service/Startup.cs
```diff
@@ -519,8 +520,13 @@ private static void EvictGraphQLSchema(IRequestExecutorResolver requestExecutorR
...
         {
+            if (loggerFilter == null)
+            {
+                loggerFilter = string.Empty;
```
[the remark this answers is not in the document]
RubenCerna2079: This is a case that should not really be possible, however the `typeof().FullName` function has the possibility of returning `null` so in order to avoid having to pass that value onwards I changed it so that it gives a string.empty().

Aniruddh25 (dismissed):
LGTM, suggest to change PR title to explain a bit more about what this change is enabling

e.g. "Enable custom log level for classes via `log-level` filters"

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -572,11 +572,26 @@ public uint GetPaginationLimit(int? first)
...
+            Runtime.Telemetry is null ||
+            Runtime.Telemetry.LoggerLevel is null ||
+            Runtime.Telemetry.LoggerLevel.Count == 0)
+        {
+            return true;
+        }
+
+        foreach (KeyValuePair<string, LogLevel?> logger in Runtime!.Telemetry.LoggerLevel)
+        {
+            if (logger.Key == null)
```
sezal98, line 587: Dont we need to check logger.value here?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -585,15 +600,37 @@ Runtime.Telemetry.LoggerLevel is null ||
     /// If host mode is Development, return `LogLevel.Debug`, else
     /// for production returns `LogLevel.Error`.
     /// </summary>
-    public static LogLevel GetConfiguredLogLevel(RuntimeConfig runtimeConfig)
+    public LogLevel GetConfiguredLogLevel(string loggerFilter = "")
```
sezal98, line 603: should use string.empty here

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -585,15 +600,37 @@ Runtime.Telemetry.LoggerLevel is null ||
...
+                }
+            }
+
+            Runtime!.Telemetry!.LoggerLevel!.TryGetValue(currentFilter, out LogLevel? value);
+            if (value is not null)
+            {
+                return (LogLevel)value;
+            }
+
+            Runtime!.Telemetry!.LoggerLevel!.TryGetValue("default", out value);
```
sezal98, line 626: instead of this, can we simply return LogLevel.Default

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -129,6 +130,25 @@ public void ValidateAppInsightsTelemetryConnectionString(RuntimeConfig runtimeCo
...
+    /// </summary>
+    public static void ValidateLoggerFilters(RuntimeConfig runtimeConfig)
+    {
+        if (runtimeConfig.Runtime?.Telemetry is not null && runtimeConfig.Runtime.Telemetry.LoggerLevel is not null)
+        {
+            Dictionary<string, LogLevel?> loggerLevelOptions = runtimeConfig.Runtime.Telemetry.LoggerLevel;
+
+            foreach (KeyValuePair<string, LogLevel?> logger in loggerLevelOptions)
+            {
+                if (!IsLoggerFilterValid(logger.Key))
```
sezal98, line 144: while validating logger filters, should we check if the value of the filter is valid or not?

sezal98 (commented):
Thanks for addressing all the comments so far. Looks good to merge, just have a couple of pointers

schemas/dab.draft.schema.json
```diff
@@ -376,9 +376,9 @@
               "type": "object",
               "description": "Global configuration of log level",
               "additionalProperties": false,
-              "properties": {
-                "level": {
-                  "description": "Defines logging severity levels, when in default value it will set logging level based on 'host: mode' property",
+              "patternProperties": {
```
aaronburtle, line 379: A comment here describing in plain english what the regular expression does would be great

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: I think for default param values you need a compile time constant, so string.empty probably get complaints from the compiler

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: What happens if youve got a namespace that is similar to another namespace? For example `Foo: debug` and `FooBar: warning` do we end up evicting because they both start with `Foo`?

aaronburtle: Also, please add a comment explaining the logic here.

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: What @aaronburtle said is correct

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: The value is verified at deserialization, so we do not need to check logger.value at this point.

schemas/dab.draft.schema.json
RubenCerna2079: Changed the description so that it shows what the regex is doing

sezal98 (approved):

schemas/dab.draft.schema.json
```diff
@@ -367,11 +367,11 @@
             },
             "log-level": {
               "type": "object",
-              "description": "Global configuration of log level",
+              "description": "Global configuration of log level, defines logging severity levels for specific classes, when 'null' it will set logging level based on 'host: mode' property",
               "additionalProperties": false,
               "patternProperties": {
                 "^[a-zA-Z_]\\w{0,}(\\.\\w{1,}){0,}$": {
-                  "description": "Defines logging severity levels for specific classes, when 'null' it will set logging level based on 'host: mode' property",
+                  "description": "Regex pattern checks that it only allows for namespace values. So it only accepts a letter as its first value, from there it accepts any alphanumeric character, each section is divided by '.' and there cannot be more than 1 '.' in a row",
```
Aniruddh25, line 374: Even though the Regex explains checking for namespaces, we should mention that only specific namespaces (classes) are valid here. Since we don't support adding any generic namespace name.

Aniruddh25:
```suggestion
                  "description": "Regex pattern checks that it only allows for namespace values. So it only accepts a letter as its first value, from there it accepts any alphanumeric character, each section is divided by '.' and there cannot be more than 1 '.' in a row. Additionally, only valid DataApiBuilder namespaces are allowed.",
```

Aniruddh25 (approved):
LGTM

sezal98 (approved):

PR #2621, opened 2025-03-18
Enhance error handling with innerException in DataApiBuilderException

rohkhann:
## Why make this change?
Having the inner exception makes root causing and debugging easier for downstream consumers of data-api-builder.

## What is this change?

Added an innerException parameter to the DataApiBuilderException constructor, passing the original exception (ex). This provides better context and aids in debugging by preserving the root cause of the error.


aaronburtle (approved):
Thanks for fixing this!

vadeveka (approved):

Issue #2286, opened 2024-07-02
[Bug]: orderBy clause has support for OR/AND operator and relationship fields which do not work 


vadeveka:
### What happened?

OrderBy clause for the queries has support for OR and AND operators per schema. 

```
input GeographyOrderByInput {
  GeographyID: OrderBy
  ...
  State: OrderBy
  Country: OrderBy
  ZipCode: OrderBy
  and: [GeographyOrderByInput]
  or: [GeographyOrderByInput]
}
```

However, using it in a query fails with an exception.
Sample query
```
query {
  geographies(orderBy: { or: { GeographyID: ASC, State: ASC } }) {
    items {
      GeographyID
    }
  }
}
```

Response in development mode:
```
{
  "errors": [
    {
      "message": "Mapped fieldname could not be found.",
      "extensions": {
        "message": "Mapped fieldname could not be found.",
[...]

PR #2625, opened 2025-03-20
Remove "And", "Or", and Relationships as valid OrderByInput types.

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2286

## What is this change?

The `InputTypeBuilder` class holds a function called `GenerateOrderByInputTypeForObjectType`, which is responsible for populating the GraphQL Schema with the valid OrderByInput types. It was mistakenly adding in "and", "or", and columns that are in entities associated through relationships as valid OrderByInput types. "and" and "or" will never be valid input types, and we do not yet support order by using columns from related entities.

## How was this tested?

Manually verified that the correct error was returned when trying to use the incorrect orderBy queries, and manually verified that we no longer have them in the schema.

Added regression test to verify we get an error when attempting these queries. 

## Sample Request(s)

These are some example graph QL queries that are invalid but were showing as valid in our schema without this fix.

>publishers (first: 5 orderBy:  {books:  {  title: DESC }}){
>  items {
>    id
>     books {
>       items {
>        title
>      }
>   }
> }
> }

>books(orderBy: { or: { id: ASC, title: ASC } }) {
    items {
      id
    }
  }

### DIFF
--- a/src/Service.GraphQLBuilder/Queries/InputTypeBuilder.cs
+++ b/src/Service.GraphQLBuilder/Queries/InputTypeBuilder.cs
@@ -24,16 +24,24 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder.Queries
         {
             List<InputValueDefinitionNode> inputFields = GenerateFilterInputFieldsForBuiltInFields(node, inputTypes);
             string filterInputName = GenerateObjectInputFilterName(node);
-
-            GenerateInputTypeFromInputFields(inputTypes, inputFields, filterInputName, $"Filter input for {node.Name} GraphQL type");
+            GenerateFilterInputTypeFromInputFields(inputTypes, inputFields, filterInputName, $"Filter input for {node.Name} GraphQL type");
         }
 
         internal static void GenerateOrderByInputTypeForObjectType(ObjectTypeDefinitionNode node, IDictionary<string, InputObjectTypeDefinitionNode> inputTypes)
         {
             List<InputValueDefinitionNode> inputFields = GenerateOrderByInputFieldsForBuiltInFields(node);
             string orderByInputName = GenerateObjectInputOrderByName(node);
-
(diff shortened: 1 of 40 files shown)


src/Service.GraphQLBuilder/Queries/InputTypeBuilder.cs
```diff
@@ -53,27 +62,12 @@ private static List<InputValueDefinitionNode> GenerateOrderByInputFieldsForBuilt
                             new List<DirectiveNode>())
                         );
                 }
-                else
```
Aniruddh25, line 56: as per https://github.com/Azure/project-hawaii/blob/main/rfcs/graphql.md#sorting, this should be supported. We should add a discussion idea, to check how many upvotes do we get that are interested in this feature

Aniruddh25 (approved):
LGTM

src/Service.GraphQLBuilder/Queries/InputTypeBuilder.cs
```diff
@@ -53,27 +62,12 @@ private static List<InputValueDefinitionNode> GenerateOrderByInputFieldsForBuilt
...
-                {
```
RubenCerna2079, line 57: What is the reason for this to be deleted?

RubenCerna2079 (approved):
lgtm

src/Service.GraphQLBuilder/Queries/InputTypeBuilder.cs
aaronburtle: This is where the relationship bsaed InputType gets added to the OrerBy inputFields. If you stick a break point inside of this else statement and then run the engine, youll see on startup where it gets added, the first one should be the input types for Publisher, and you will notice that the 3rd one which is being added inside of this else branch is the book input type that is from the relationship between publisher and book.

PR #2627, opened 2025-03-20
Make parameter length based on underlying column length except for complex ops

simonsabin:
## Why make this change?

Fixes issue #2619 

## What is this change?

Where a parameter is linked to an underlying column in the data model the length is stored
When the query is executed the length is used to ensure parameters have consistent lengths

For certain operations the value is padded with escape characters or %s, these need to be allowed for.  In those cases the length is  set to -1 (max) in order that the extra characters are allowed for AND consistent lengths are maintained

## How was this tested?

- [x] Integration Tests

Checking the resultant query is using the right length was done using trace. Not sure with the architecture whether it would be possible to mock at that level.

## Sample Request(s)

See new tests TestFilterParamForStringFilterWorkWithComplexOp and TestFilterParamForStringFilterWorkWithNotContains

### DIFF
--- a/src/Config/DatabasePrimitives/DatabaseObject.cs
+++ b/src/Config/DatabasePrimitives/DatabaseObject.cs
@@ -229,6 +229,17 @@ public class SourceDefinition
 
         return null;
     }
+
+    public virtual int? GetLengthForParam(string paramName)
+    {
+        if (Columns.TryGetValue(paramName, out ColumnDefinition? columnDefinition))
+        {
+            return columnDefinition.Length
+            ;
+        }
+
+        return null;
+    }
 }
 
 /// <summary>
@@ -264,6 +275,7 @@ public class ColumnDefinition
     public bool IsNullable { get; set; }
     public bool IsReadOnly { get; set; }
     public object? DefaultValue { get; set; }
+    public int? Length { get; set; }
 
     public ColumnDefinition() { }
 
--- a/src/Core/Models/GraphQLFilterParsers.cs
+++ b/src/Core/Models/GraphQLFilterParsers.cs
@@ -486,7 +486,7 @@ public class GQLFilterParser
         string schemaName,
         string tableName,
         string tableAlias,
-        Func<object, string?, string> processLiterals,
+        Func<object, string?,bool, string> processLiterals,
         bool isListType = false)
     {
         Column column = new(schemaName, tableName, columnName: fieldName, tableAlias);
@@ -614,7 +614,7 @@ public static class FieldFilterParser
         IInputField argumentSchema,
         Column column,
         List<ObjectFieldNode> fields,
-        Func<object, string?, string> processLiterals,
+        Func<object, string?, bool, string> processLiterals,
         bool isListType = false)
     {
         List<PredicateOperand> predicates = new();
@@ -634,6 +634,8 @@ public static class FieldFilterParser
             {
                 continue;
             }
+
+            bool lengthOverride = false;
 
             PredicateOperation op;
             switch (name)
@@ -665,6 +667,7 @@ public static class FieldFilterParser
                     {
                         op = PredicateOperation.LIKE;
                         value = $"%{EscapeLikeString((string)value)}%";
+                        lengthOverride = true;
                     }
 
                     break;
@@ -677,16 +680,19 @@ public static class FieldFilterParser
                     {
                         op = PredicateOperation.NOT_LIKE;
                         value = $"%{EscapeLikeString((string)value)}%";
+                        lengthOverride = true;
                     }
 
                     break;
                 case "startsWith":
                     op = PredicateOperation.LIKE;
                     value = $"{EscapeLikeString((string)value)}%";
+                    lengthOverride = true;
                     break;
                 case "endsWith":
                     op = PredicateOperation.LIKE;
                     value = $"%{EscapeLikeString((string)value)}";
+                    lengthOverride = true;
                     break;
                 case "isNull":
                     processLiteral = false;
@@ -699,10 +705,10 @@ public static class FieldFilterParser
             }
 
             predicates.Push(new PredicateOperand(new Predicate(
-                new PredicateOperand(column),
+                new(column),
                 op,
-                new PredicateOperand(processLiteral ? $"{processLiterals(value, column.ColumnName)}" : value.ToString()))
-                ));
+                new(processLiteral ? $"{processLiterals(value, column.ColumnName, lengthOverride)}" : value.ToString())
+                )));
         }
 
         return GQLFilterParser.MakeChainPredicate(predicates, PredicateOperation.AND);
--- a/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
@@ -1321,7 +1321,8 @@ namespace Azure.DataApiBuilder.Core.Services
                     SystemType = (Type)columnInfoFromAdapter["DataType"],
                     // An auto-increment column is also considered as a read-only column. For other types of read-only columns,
                     // the flag is populated later via PopulateColumnDefinitionsWithReadOnlyFlag() method.
-                    IsReadOnly = (bool)columnInfoFromAdapter["IsAutoIncrement"]
+                    IsReadOnly = (bool)columnInfoFromAdapter["IsAutoIncrement"],
+                    Length = (int)columnInfoFromAdapter["ColumnSize"]
                 };
 
                 // Tests may try to add the same column simultaneously
--- a/src/Service.Tests/DatabaseSchema-MsSql.sql
+++ b/src/Service.Tests/DatabaseSchema-MsSql.sql
@@ -82,7 +82,7 @@ CREATE TABLE publishers_mm(
 
 CREATE TABLE books(
     id int IDENTITY(5001, 1) PRIMARY KEY,
-    title varchar(max) NOT NULL,
+    title varchar(30) NOT NULL,
     publisher_id int NOT NULL
 );
 
@@ -514,7 +514,7 @@ SET IDENTITY_INSERT books ON
 INSERT INTO books(id, title, publisher_id)
 VALUES (1, 'Awesome book', 1234),
 (2, 'Also Awesome book', 1234),
-(3, 'Great wall of china explained', 2345),
+(3, 'Great wall of china explained]', 2345),
 (4, 'US history in a nutshell', 2345),
 (5, 'Chernobyl Diaries', 2323),
 (6, 'The Palace Door', 2324),
--- a/src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
+++ b/src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
@@ -165,6 +165,34 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.GraphQLQueryTests
             await QueryWithMultipleColumnPrimaryKey(msSqlQuery);
         }
 
+        /// <sumary>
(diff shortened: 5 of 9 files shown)


src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -229,6 +229,17 @@ public bool IsAnyColumnNullable(List<string> columnsToCheck)
 
         return null;
     }
+
+    public virtual int? GetLengthForParam(string paramName)
+    {
+        if (Columns.TryGetValue(paramName, out ColumnDefinition? columnDefinition))
+        {
+            return columnDefinition.Length
```
aaronburtle, line 237:
```suggestion
            return columnDefinition.Length;
```

src/Config/DatabasePrimitives/DatabaseObject.cs
```diff
@@ -229,6 +229,17 @@ public bool IsAnyColumnNullable(List<string> columnsToCheck)
...
+            ;
```
aaronburtle, line 238:
```suggestion
```

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -486,7 +486,7 @@ private static Predicate ParseScalarType(
         string schemaName,
         string tableName,
         string tableAlias,
-        Func<object, string?, string> processLiterals,
+        Func<object, string?,bool, string> processLiterals,
```
aaronburtle, line 489:
```suggestion
        Func<object, string?, bool, string> processLiterals,
```

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -514,7 +514,7 @@ SET IDENTITY_INSERT books ON
 INSERT INTO books(id, title, publisher_id)
 VALUES (1, 'Awesome book', 1234),
 (2, 'Also Awesome book', 1234),
-(3, 'Great wall of china explained', 2345),
+(3, 'Great wall of china explained]', 2345),
```
aaronburtle, line 517: I am curious, why the extra "]"?

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -165,6 +165,93 @@ SELECT TOP 1 content FROM reviews
...
+  }
+]";
+
+            JsonElement actual = await ExecuteGraphQLRequestAsync(graphQLQuery, graphQLQueryName, isAuthenticated: false);
+
+            SqlTestHelper.PerformTestEqualJsonStrings(expected, actual.GetProperty("items").ToString());
+        }
+
+        /// <sumary>
+        /// Test if filter param successfully filters when string filter results in a value longer than the column
```
aaronburtle, line 198: I think I confused myself, if it is varchar(30) and the value is 30 characters long, what is the value that is loner than the column?

aaronburtle (commented):
This looks great, thank you for raising and then solving this issue. Just had a few nits, and a question.

simonsabin:
It’s to cover the edge case of matching with the LiKE that you get with contains and starts with, and also the escaping of values that occurs in the predicates. If you don’t you can end up with false positives and false negatives

Simon Sabin

________________________________

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
simonsabin: If you do a startswith or ends with or contains using that as value, it gets escaped to be longer AND a % gets added at the front, end or both. So ends up with a value that is longer than the original 30 character value

simonsabin:
applied suggested changes

aaronburtle:
Pipeline failures are just whitespace format.

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -82,7 +82,7 @@ CREATE TABLE publishers_mm(
 
 CREATE TABLE books(
     id int IDENTITY(5001, 1) PRIMARY KEY,
-    title varchar(max) NOT NULL,
+    title varchar(30) NOT NULL,
```
Aniruddh25, line 85: Should we test something similar for PostgreSQL/MySQL?

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
Aniruddh25: Had a similar question, @simonsabin could we add the reasoning as a comment in the code itself?
As Im reading it, is this testing lengthOverride = true codepath?

src/Service.Tests/DatabaseSchema-MsSql.sql
simonsabin: That would be a good idea, but I don't have those setup to validate.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1321,7 +1321,8 @@ private async Task PopulateSourceDefinitionAsync(
                     SystemType = (Type)columnInfoFromAdapter["DataType"],
                     // An auto-increment column is also considered as a read-only column. For other types of read-only columns,
                     // the flag is populated later via PopulateColumnDefinitionsWithReadOnlyFlag() method.
-                    IsReadOnly = (bool)columnInfoFromAdapter["IsAutoIncrement"]
+                    IsReadOnly = (bool)columnInfoFromAdapter["IsAutoIncrement"],
+                    Length = (int)columnInfoFromAdapter["ColumnSize"]
```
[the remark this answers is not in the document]
Aniruddh25: @simonsabin  if you aren't able to validate, PostgreSql/MySQL could you please move this code to the derived class `MsSqlMetadaProvider` and create a backlog item for the other databases so that this doesn't regress those dbs?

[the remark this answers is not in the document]
simonsabin: Seems a big thing to copy all fo this in the MSSQL provider, there is already logic checking the provider in this method. 

We could check either the column exists in the columnInfoFromAdapter, or check the provider and only set the Length if its MSSQL.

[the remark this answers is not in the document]
simonsabin: Just read that ColumnSize is implemented in the underlying class https://learn.microsoft.com/en-us/dotnet/api/system.data.datatablereader.getschematable?view=net-9.0 so pretty sure its safe to use.

src/Service.Tests/DatabaseSchema-MsSql.sql
simonsabin: Yes and for it to include characters that need escaping.

src/Config/DatabasePrimitives/DatabaseObject.cs
simonsabin: should now be resolved

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
simonsabin: added description to test

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -197,13 +197,22 @@ public virtual async Task TestFilterParamForStringFilter()
         /// <sumary>
         /// Test if filter param successfully filters when string filter results in a value longer than the column
         /// </summary>
+        /// <remarks>
+        /// When using complex operators i.e. NotContains due to wildcards being added or special characters being escaped
+        /// the string being passed as a parameter maybe longer than the length of the column. The parameter data type
+        /// can't be fixed to the length of the underlying column, otherwise the parameter value would be truncated and
+        /// we'd get incorret results
```
Aniruddh25, line 204:
```suggestion
        /// we'd get incorrect results
```

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -197,13 +197,22 @@ public virtual async Task TestFilterParamForStringFilter()
...
+        /// Thus checking the parameter length is overridden to cater for teh extra length i.e. lengthOverride = true codepath?
```
Aniruddh25, line 205:
```suggestion
        /// Thus checking the parameter length is overridden to cater for the extra length i.e. lengthOverride = true codepath?
```

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -229,10 +238,18 @@ public virtual async Task TestFilterParamForStringFilterWorkWithComplexOp(string
         /// <sumary>
         /// Test if filter param successfully filters when string filter results in a value longer than the column
         /// </summary>
+        /// <remarks>
+        /// When using complex operators i.e. NotContains due to wildcards being added or special characters being escaped
+        /// the string being passed as a parameter maybe longer than the length of the column. The parameter data type
+        /// can't be fixed to the length of the underlying column, otherwise the parameter value would be truncated and
+        /// we'd get incorret results.
```
Aniruddh25, line 245:
```suggestion
        /// we'd get incorrect results.
```

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -229,10 +238,18 @@ public virtual async Task TestFilterParamForStringFilterWorkWithComplexOp(string
...
+        /// Thus checking the parameter length is overridden to cater for teh extra length i.e. lengthOverride = true codepath?
```
Aniruddh25, line 246:
```suggestion
        /// Thus checking the parameter length is overridden to cater for the extra length i.e. lengthOverride = true codepath.
```

Aniruddh25 (dismissed):
LGTM, thank you again for finding this issue and fixing it as well!!

aaronburtle (approved):
Thanks again for making these changes, looks great.

simonsabin:
whats the state of this?

simonsabin:
@Aniruddh25 @aaronburtle whats needed to get this merged in?

Aniruddh25:
Running the tests to ensure they pass. Will merge this soon.

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -778,13 +786,13 @@ private static PredicateOperand GenerateRightOperand(
...
-                .Select(inValue => processLiterals(inValue!, null))
+                .Select(inValue => processLiterals(inValue!, null, false))
                 .ToList();
 
             return new PredicateOperand("(" + string.Join(", ", encodedParams) + ")");
         }
 
-        return new PredicateOperand(processLiteral ? processLiterals(value, null) : value.ToString());
+        return new PredicateOperand(processLiteral ? processLiterals(value, null, lengthOverride) : value.ToString());
     }
```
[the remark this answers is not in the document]
aaronburtle: @copilot add this change to this PR based on [this feedback](https://github.com/Azure/data-api-builder/pull/2627#discussion_r2897376228)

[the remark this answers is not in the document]
aaronburtle: copilot seemed to ignore this, addressed manually

simon-sabin-BE:
thanks for perservering with this

PR #2631, opened 2025-03-21
Improved N-1 join query performance for DW SQL

lingxiao-microsoft:
## Why make this change?
- Back 1 year ago, as DW SQL does not support `JSON PATH` for converting the execution to json format. Hence we had to use `STRING_AGG` as the workaround. 

- Recently, we've noticed `JSON PATH` is now supported for outer query in DW, and we can use `JSON_OBJECT` + `JSON_PATH` to address the json conversion for N-to-1 relations, which can optimize the performance. 

- For N-N relations, we're still looking into any resolutions as `JSON_ARRAYAGG` does not provide much performance improvements. 
- 
- For other scenarios when joins are not needed for a simple SELECT, we will `JSON PATH` instead of `STRING_AGG` for better performance.

## What is this change?

This PR covers
1. Introduced a feature flag to safeguard the changes, the feature flag is default as False when not provided to avoid any regressions. It will be removed once the changes are validated in production with scoped audiences. 
2. For DW query builder, use `JSON_OBJECT` to generate the columns for sub-queries and applied `JSON PATH` to handle outer query, which fully replace the need of `STRING_AGG`.
3. Also, for non-join queries (in which we don't need to handle the relations), used `JSON PATH` to replace the need of `STRING_AGG` for better performance as well. This will have impact on aggregations, non-join queries and pagination. 
4. Added some helper functions into the unit tests module, which aims to compare the results from GraphQL & DB engine easily for deeply nested queries. 

## How was this tested?

- [x] Unit Tests
  - As this change does not introduce new scenarios, so mostly added some new test cases to get more coverage when M-M / M-1 join queries are needed.

- [x] Integration Tests

#### Query 1-1 relation - As expected, optimization applied
image
image

#### Query N-1 relation - As expected, optimization applied
image
image

#### Query 1-N Relation - As expected, optimization not applied
image
image

#### Query N-N Relation - As expected, optimization not applied
image
image

### Other Scenarios
We've applied the `JSON PATH` when there is no join in the query to replace the `STRING_AGG` for better performance. 

#### Aggregation
image

#### Non-Join Query
image

#### Pagination
- N to 1, total items: 3
image
image

- N to N
image
image
image


src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
```diff
@@ -97,6 +98,341 @@ ORDER BY [table0].[id] ASC
             await OneToOneJoinQuery(dwSqlQuery);
         }
 
+        /// <summary>
```
vadeveka, line 101: Can we add a test case where JSON PATH gets applied?

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
```diff
@@ -97,6 +98,341 @@ ORDER BY [table0].[id] ASC
...
+        /// DwNTo1JoinOpt is enabled by default for testing
+        /// Below test case will ensure the results are identical with using STRING_AGG
+        /// </summary>
+        /// <returns></returns>
+        [TestMethod]
+        public async override Task DeeplyNestedManyToOneJoinQuery()
```
vadeveka, line 107: Should we be having JSON PATH applied in this case for the outer query?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +44,155 @@ public override string QuoteIdentifier(string ident)
...
+            }
+        }
+
+        /// <summary>
+        /// Recursively checks the structure to see if
+        /// 1. It only has to-1 relations
+        /// 2. It does not have any relations, which means it is a simple query against one table
+        /// We should apply the json funcs instead of string_agg for both cases
+        /// </summary>
+        /// <param name="structure"></param>
```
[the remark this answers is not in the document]
lingxiao-microsoft: Will do

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +44,155 @@ public override string QuoteIdentifier(string ident)
...
+        /// <returns></returns>
+        private static bool HasToOneOrNoRelation(SqlQueryStructure structure, bool isSubQuery)
+        {
+            if (structure?.JoinQueries?.Values == null)
+            {
+                // if there is no sub-queries, use JSON PATH for performance improvements as well
+                return true;
+            }
+
+            if (structure.IsListQuery && isSubQuery)
```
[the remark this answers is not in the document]
lingxiao-microsoft: if it is a list query which means it is a to-N relation, it equals to the `items` keywords in the graphql query.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +44,155 @@ public override string QuoteIdentifier(string ident)
...
+                    query += "," + WITHOUT_ARRAY_WRAPPER_SUFFIX;
+                }
+            }
+
+            return query;
+        }
+
+        /// <summary>
+        /// Helper function for BuildWithJsonFunc that generates "FROM" portion of the query
+        /// </summary>
```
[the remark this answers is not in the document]
lingxiao-microsoft: Willl update

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
lingxiao-microsoft: Yes. The idea here is to do one step further to make sure the results we generated using JSON PATH is same as the results generated previously using STRING AGG to ensure backward compatibility.

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
lingxiao-microsoft: There are [test cases](https://github.com/Azure/data-api-builder/blob/aee9690761018b2995642ca925764eeec0ebdaa4/src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs#L143) against queries using `JSON PATH` for non-join queries. 

We should already be covered by it.

src/Cli.Tests/ModuleInitializer.cs
```diff
@@ -53,6 +53,10 @@ public static void Init()
         VerifierSettings.IgnoreMember<RuntimeConfig>(options => options.EnableAggregation);
         // Ignore the EnableAggregation as that's unimportant from a test standpoint.
         VerifierSettings.IgnoreMember<GraphQLRuntimeOptions>(options => options.EnableAggregation);
+        // Ignore the EnableDwNto1JoinOpt as that's unimportant from a test standpoint.
+        VerifierSettings.IgnoreMember<RuntimeConfig>(options => options.EnableDwNto1JoinOpt);
```
rohkhann, line 57: Adding to runtime config makes this publicly facing and something users should be aware of @JerryNixon on his inputs here.

src/Config/ObjectModel/GraphQLRuntimeOptions.cs
```diff
@@ -11,7 +11,8 @@ public record GraphQLRuntimeOptions(bool Enabled = true,
                                     bool AllowIntrospection = true,
                                     int? DepthLimit = null,
                                     MultipleMutationOptions? MultipleMutationOptions = null,
-                                    bool EnableAggregation = true)
+                                    bool EnableAggregation = true,
+                                    bool EnableDwNto1JoinOpt = false)
```
rohkhann, line 15: Dab users would not see the benefits of this perf improvement and it is a very unintuitive thing to go to config file and enable. @aaronburtle any other way of having a feature flag apart from adding it in the runtime config?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -15,8 +15,17 @@ namespace Azure.DataApiBuilder.Core.Resolvers
     public class DwSqlQueryBuilder : BaseSqlQueryBuilder, IQueryBuilder
     {
         private static DbCommandBuilder _builder = new SqlCommandBuilder();
+        private readonly bool _enableNto1JoinOpt;
+        private const string FOR_JSON_SUFFIX = " FOR JSON PATH, INCLUDE_NULL_VALUES";
```
rohkhann, line 19: nit: for other constants we dont prefix it with space (see the WITHOUT_ARRAY_WRAPPER) should probably stick to ensure we follow same format and there are not ghost spaces added.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +44,155 @@ public override string QuoteIdentifier(string ident)
...
+
+            return query;
+        }
+
+        /// <summary>
+        /// Helper function for BuildWithJsonFunc that generates "FROM" portion of the query
+        /// </summary>
+        /// <param name="structure"></param>
+        /// <returns></returns>
+        private string BuildWithJsonFunc(SqlQueryStructure structure)
```
rohkhann, line 129: My understanding is this should exactly mimic the logic for mssql file, any chance of code reuse to prevent having to maintain two version os essentially the same TSQL code?

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -129,6 +277,27 @@ private string BuildSqlQuery(SqlQueryStructure structure)
             return query;
         }
 
+        /// <summary>
+        /// Generate the columns selected and wrap them with JSON_OBJECT
+        /// </summary>
+        /// <param name="structure"></param>
+        /// <returns></returns>
+        private static string GenerateColumnsAsJsonObject(SqlQueryStructure structure)
```
rohkhann, line 285: Same as above, lets see what we can reuse from the mssql and the dwsql classes. we can even have an intermediate TSQL queryBuilder class with common functions where mssql and dwsql inherit and overrride off of it if necessary.

src/Config/ObjectModel/GraphQLRuntimeOptions.cs
lingxiao-microsoft: Maybe we can flip this flag to always true once we've verified with customers. So that everyone else can benefit from it without flip it in config file

@aaronburtle , pls let me know if there is a better way to achieve the purpose

vadeveka: DAB users will benefit if they use dwsql with Synapse or fabric warehouses/SAE. It doesnt need to be a public configuration though. We should enable this optimization by a default after building confidence in GraphQL workload.

src/Core/Resolvers/DWSqlQueryBuilder.cs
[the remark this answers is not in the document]
lingxiao-microsoft: Done

src/Core/Resolvers/BaseSqlQueryBuilder.cs
```diff
@@ -353,6 +356,123 @@ protected virtual string Build(Predicate? predicate)
             }
         }
 
+        /// <summary>
+        /// Build the Json Path query needed to append to the main query
+        /// </summary>
+        /// <param name="structure">Sql query structure to build query on</param>
+        /// <returns></returns>
+        protected virtual string BuildJsonPath(SqlQueryStructure structure)
```
lingxiao-microsoft, line 364: Centralizing the functions so it can be used by both dwsql and mssql query builder.

src/Config/ObjectModel/GraphQLRuntimeOptions.cs
JerryNixon:
```json
{
    "feature-flags": {
        "enable-sqldw-join-rewrite": true
    }
    "runtime": {
        ...
    }
}
```
Great. Let's not document these or add them to `dab validate` or `dab config`. I think feature flags are temporary and might be removed later. So, docs would confuse the matter. I think updating JSON schema for `feature-flags` still makes sense so schema validation can still pass.

src/Config/Converters/GraphQLRuntimeOptionsConverterFactory.cs
```diff
@@ -113,6 +113,17 @@ internal GraphQLRuntimeOptionsConverter(bool replaceEnvVar)
                                 throw new JsonException($"Unexpected type of value entered for enable-aggregation: {reader.TokenType}");
                             }
 
+                            break;
+                        case "enable-dw-nto1joinopt":
```
Aniruddh25, line 117: based on our discussion, this may not be needed.

src/Config/ObjectModel/GraphQLRuntimeOptions.cs
Aniruddh25: latest update: looks like we dont necessarily need this option altogether in the config/schema. So, @lingxiao-microsoft is looking at if he can get away without adding the `feature-flags` section altogether

src/Config/Converters/GraphQLRuntimeOptionsConverterFactory.cs
lingxiao-microsoft: Removed the support of read/write from the dab config

src/Config/ObjectModel/GraphQLRuntimeOptions.cs
lingxiao-microsoft: Thanks for the feedback! Done. Updated the PR.

To summarize the changes
1. Removed the support of applying this flag from dab config.
2. Used `FeatureFlags` class as the container for ephemeral flags that are passed into the GraphQLRuntimeOptions. In this way, every time when a new flag is needed for the same purpose, we don't need to change the signature every time. 
3. When we're confident about the change, we can flip the flag in the `FeatureFlags` container as always True so that the rest of DAB users can benefit from it. 

@Aniruddh25 @vadeveka @rohkhann please review it again when you have some time!

rohkhann (approved):

src/Core/Resolvers/DWSqlQueryBuilder.cs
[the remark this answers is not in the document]
Aniruddh25: Name is added but the description of the parameter is still not added.
```suggestion
        /// <param name="structure">Sql query structure to build query on</param>
```

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +42,131 @@ public override string QuoteIdentifier(string ident)
...
+        ///    <Sub-query-tables>
+        /// WHERE
+        ///    <Conditions>
+        /// ORDER BY
+        ///    <Conditions>
+        /// FOR JSON PATH, INCLUDE_NULL_VALUES;
+        /// </summary>
+        /// <param name="structure">Sql query structure to build query on</param>
+        /// <param name="isSubQuery">Used for recursive call purpose to generated different queries for sub-queries</param>
+        /// <returns></returns>
```
Aniruddh25, line 108: nit: mention what is the return value.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +42,131 @@ public override string QuoteIdentifier(string ident)
...
+        private string BuildWithJsonFunc(SqlQueryStructure structure, bool isSubQuery)
+        {
+            string query;
```
Aniruddh25, line 111: Please use `StringBuilder` for better performance while doing concatenation +=

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +42,131 @@ public override string QuoteIdentifier(string ident)
...
+
+            string predicates = BuildPredicates(structure);
+
+            string aggregations = BuildAggregationColumns(structure);
+
+            string query = $"SELECT TOP {structure.Limit()} {WrappedColumns(structure)} {aggregations}"
+                + $" FROM {fromSql}"
+                + $" WHERE {predicates}";
+
+            query += BuildGroupBy(structure);
```
Aniruddh25, line 154: Please use `StringBuilder` for `query`

Aniruddh25 (changes requested):
Overall looks good, but looking for better testing. And making sure we still continue to test when the feature flag is off because thats the default behavior today. 
Turning it on always by default in testing keeps us at risk of regression in the false scenario.

src/Service.Tests/SqlTests/SqlTestBase.cs
```diff
@@ -341,7 +344,8 @@ protected static void SetUpSQLMetadataProvider(RuntimeConfigProvider runtimeConf
                     break;
                 case TestCategory.DWSQL:
                     Mock<ILogger<MsSqlQueryExecutor>> DwSqlQueryExecutorLogger = new();
-                    _queryBuilder = new DwSqlQueryBuilder();
+                    // enable the feature by default for testing purpose. 
+                    _queryBuilder = new DwSqlQueryBuilder(enableNto1JoinOpt: true);
```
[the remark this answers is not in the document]
lingxiao-microsoft: Thanks for bring it up! Actually this line should be removed.  Even if we specify it here as `true`, the tests will still run as disabled.

The reasons are due to constraints below
1. When initializing the test base, the test will generate a runtime config in [memory](https://github.com/Azure/data-api-builder/blob/97c25dbfd9abd8507d9222b65a96c52f4fb27446/src/Service.Tests/SqlTests/SqlTestBase.cs#L130C1-L131C1)
image

2. Within it, it will build a mock file system and convert the defined runtime config[ to JSON ](https://github.com/Azure/data-api-builder/blob/97c25dbfd9abd8507d9222b65a96c52f4fb27446/src/Service.Tests/TestHelper.cs#L306C88-L306C110)and load it to memory. 
image

3. As it is invoking the `ToJSON` function, since we're completely removing the function of read/write from the dab-config file for the feature flag `enableNto1JoinOpt`, the property defined in the [RuntimeConfig ](https://github.com/Azure/data-api-builder/blob/eaf9501c010467ba5c682bd77bb9d32b0c4e9156/src/Config/ObjectModel/GraphQLRuntimeOptions.cs#L32)will be ignored after the json serialization is invoked. 

4. As a result of that, even if we define a `True` or `False` value in the RuntimeConfig for testing, it will be overridden as the serialization excluded it.  

I've done two things to ensure the chance of regressions are avoided
1. Currently in `FeatureFlag` class, the default value is disabled, which means when running the integration tests in pipeline, it is against when it is disabled. All tests are passing
2. I've forced the value as `enabled` and pushed the changes to let the integration tests ran against it, there were 5 [failed ](https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=9657&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=f2696cc9-a35b-5158-31dc-9689759fe573&l=335)tests which I found it was due to when running the DW SQL tests, a MSSQL engine is used instead and in SQL 2019, the `JSON_OBJECT` function is not supported. 
3. I've ran the DWQueryBuilderTest locally and all of them are passing
image

5. This can be fixed once we've flipped the values as `True` in DAB and use the right engine for integration testing for DW. 

This is so far the best workaround I could think of having an ephemeral feature flag which serves the purpose and at the meantime not exposing it in dab config files. 

Please let me know if that does not make sense

src/Core/Resolvers/DWSqlQueryBuilder.cs
lingxiao-microsoft: Done!

src/Core/Resolvers/DWSqlQueryBuilder.cs
lingxiao-microsoft: Updated the dwsql and mssql query builder for consistency.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -35,16 +42,131 @@ public override string QuoteIdentifier(string ident)
...
+        /// <returns></returns>
+        private string BuildWithJsonFunc(SqlQueryStructure structure)
+        {
+            string dataIdent = QuoteIdentifier(SqlQueryStructure.DATA_IDENT);
+            string fromSql = $"{QuoteIdentifier(structure.DatabaseObject.SchemaName)}.{QuoteIdentifier(structure.DatabaseObject.Name)} " +
+                             $"AS {QuoteIdentifier($"{structure.SourceAlias}")}{Build(structure.Joins)}";
+
+            fromSql += string.Join(
+                    "",
+                    structure.JoinQueries.Select(
```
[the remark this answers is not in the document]
lingxiao-microsoft: Yes - it is initialized as an empty object https://github.com/Azure/data-api-builder/blob/97c25dbfd9abd8507d9222b65a96c52f4fb27446/src/Core/Resolvers/Sql%20Query%20Structures/SqlQueryStructure.cs#L556C1-L557C1

src/Core/Resolvers/DWSqlQueryBuilder.cs
lingxiao-microsoft: Updated the PR.

src/Core/Resolvers/DWSqlQueryBuilder.cs
```diff
@@ -129,6 +251,41 @@ private string BuildSqlQuery(SqlQueryStructure structure)
...
+        private static string GenerateColumnsAsJsonObject(SqlQueryStructure structure)
+        {
+            List<string> columns = new();
+            foreach (LabelledColumn column in structure.Columns)
+            {
+                string col_value = $"\'{column.Label}\': [{column.TableAlias}].[{column.Label}]";
+                columns.Add(col_value);
+            }
+
+            string? joinedColumns = columns.Count > 1 ?
```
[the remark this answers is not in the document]
lingxiao-microsoft: To my best understanding, it should at least have one column, as in
1. It only applies when N-1 join happens
2. Clients needs to specify at least one column for the leaf 
3. If it is a nested query (>=3 table joins), there will always be a column in 2nd layer created for aggregating the results from the 3rd table. 

I've updated the PR to avoid making the variable as nullable

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
```diff
@@ -97,6 +99,341 @@ ORDER BY [table0].[id] ASC
...
+                    }
+                  }
+                }";
+
+            JsonElement actual = await ExecuteGraphQLRequestAsync(graphQLQuery, graphQLQueryName, isAuthenticated: false);
+            JsonNode cleanedActual = SqlTestHelper.RemoveItemsKeyFromJson(actual.GetProperty("items"));
+
+            string expected = await GetDatabaseResultAsync(dwSqlQuery);
+
+            SqlTestHelper.PerformTestEqualJsonStringsForNestedQueries(expected, cleanedActual.ToString());
```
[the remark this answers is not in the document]
lingxiao-microsoft: There are reasons that blocking us to test when the feature is disabled / enabled, I've replied with some details in https://github.com/Azure/data-api-builder/pull/2631#discussion_r2023652394

We can add more tests once this feature is enabled by default in DAB

src/Service.Tests/SqlTests/SqlTestBase.cs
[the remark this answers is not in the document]
Aniruddh25: Makes sense, thanks for testing this. Could you please follow up with better testing of DW?

Aniruddh25 (approved):
LGTM, thanks for addressing all comments and answering the questions! We should look at better DwSql testing (not using SQL2019 instance for e.g.) as a follow up to this PR.

rohkhann (approved):

src/Service.Tests/SqlTests/SqlTestBase.cs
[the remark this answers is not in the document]
lingxiao-microsoft: Per discussion offline, we should use the right SQL version when the flag is enabled in DAB

Issue #2531, opened 2025-01-23
[HealthEndpoint] Roles Support 


sezal98:
Add Roles runtime.health.roles
- Run basic health check if roles dont match
- Run the comprehensive health check report when roles match
Include the string array, runtime.health.roles (roles which, when match, checks run and, when don't match, return only basic health)

      * There is no default to this value, they must include "anonymous".

JerryNixon:
`runtime.health.roles` (string array)

Default value: none.

| Condition                          | User is in Group?            | Behavior                                                                 |
|------------------------------------|------------------------------|-------------------------------------------------------------------------|
| Empty (default)                    | -                          | No checks, just high-level health information (also omit `configuration`).     |
| `anonymous`            | -                          | All checks are performed all the time.                                  |
| `authenticated`         | Yes                          | All checks for any role other than anonymous.                           |
| `authenticated`         | No (anonymous user)          | Treated as empty when the user is anonymous.                            |
| `{custom-role}`         | Yes (matches at least one custom role)    | All checks only when the user's role matches the specified custom role. |
| `{custom-role}`         | No (does not match at least one role)     | Treated as empty when the user's role does not match.                   |

PR #2632, opened 2025-03-24
Allow access to health endpoint as per defined roles

sezal98:
## Why make this change?
Closes [#2531](https://github.com/Azure/data-api-builder/issues/2531)

## What is this change?
This PR involves creating another parameter in the runtime config which defines the **allowed roles** for the health endpoint.
+ It created `src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs` which takes care of serialization and deserialization of runtime config custom way.
+ The incoming role and token are taken up to firct check if the role in the header is matching with the allowed roles in the runtime config.
+ Further this role and token are passed to execute the REST and GraphQL queries for DAB.

Scenarios
1. Allowed Roles are not configured.
+ Mode is Development - we allow all requests
+ Mode is Production - we dont allow any requests. We show 403
3. Allowed Roles are configured
+ Mode is Development or Production - we performt the role check in allowed roles.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

## Sample Request(s)
Roles: ["authenticated"]
Mode: Development or Production
image
image

Roles: Not Configured
Mode: Development
image

Mode: Production
image

Roles: ["admin"]
Mode: Development or Production
image


src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -58,6 +62,48 @@ public ComprehensiveHealthCheckReport GetHealthCheckResponse(HttpContext context
...
+
+            if (clientRoleHeader.Count == 1)
+            {
+                _incomingRoleHeader = clientRoleHeader.ToString().ToLowerInvariant();
+            }
+
+            if (clientRoleHeader.Count == 1)
+            {
+                _incomingRoleToken = clientTokenHeader.ToString();
+            }
```
RubenCerna2079, line 78: Why not have both under the same if statement?

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -0,0 +1,135 @@
...
+                                    // Read the array elements one by one
+                                    while (reader.Read() && reader.TokenType != JsonTokenType.EndArray)
+                                    {
+                                        if (reader.TokenType == JsonTokenType.String)
+                                        {
+                                            string? currentRole = reader.GetString();
+                                            if (!string.IsNullOrEmpty(currentRole))
+                                            {
+                                                stringList.Add(currentRole);
+
```
RubenCerna2079, line 79: nit: delete this line

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -0,0 +1,135 @@
...
+                                    roles = stringList;
+                                }
+                                else
+                                {
+                                    // Handle case where the token is not an array (e.g., throw an exception or handle differently)
+                                    throw new JsonException("Expected an array of strings, but the token type is not an array.");
+                                }
+                            }
+
+                            break;
```
RubenCerna2079, line 106: I think it would be a good idea to add a `default` case that throws an exception if the string is an unwanted value?

src/Service.Tests/Configuration/HealthEndpointRolesTests.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Helper function to write custom configuration file. with minimal REST/GraphQL global settings
+        /// using the supplied entities.
+        /// </summary>
+        /// <param name="entityMap">Collection of entityName -> Entity object.</param>
+        /// <param name="role">Allowed Roles for comprehensive health endpoint.</param>
+        private static void CreateCustomConfigFile(Dictionary<string, Entity> entityMap, string? role, HostMode hostMode = HostMode.Production)
```
RubenCerna2079, line 104: I'm confused, wouldn't this need to be `HostMode? hostMode` in order to accept the null parameter you pass from the test?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -58,6 +62,48 @@ public ComprehensiveHealthCheckReport GetHealthCheckResponse(HttpContext context
             return ComprehensiveHealthCheckReport;
         }
 
+        public void UpdateIncomingRoleHeader(HttpContext httpContext)
```
RubenCerna2079, line 65: Could you add a comment to describe this function?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -58,6 +62,48 @@ public ComprehensiveHealthCheckReport GetHealthCheckResponse(HttpContext context
...
+            {
+                // When allowedRoles is null or empty, all roles are allowed if Mode = Development.
+                return hostMode == HostMode.Development;
+            }
+
+            switch (hostMode)
+            {
+                case HostMode.Development:
+                case HostMode.Production:
+                    return allowedRoles.Contains(_incomingRoleHeader);
```
RubenCerna2079, line 101: What is the point of this switch case if both lead to the same return? Is it for when it is null?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -189,7 +235,7 @@ private void PopulateEntityHealth(ComprehensiveHealthCheckReport ComprehensiveHe
                             ThresholdMs = entityValue.EntityThresholdMs
                         },
                         Tags = [HealthCheckConstants.REST, HealthCheckConstants.ENDPOINT],
-                        Exception = !thresholdCheck ? TIME_EXCEEDED_ERROR_MESSAGE : response.Item2,
+                        Exception = response.Item2 ?? (!thresholdCheck ? TIME_EXCEEDED_ERROR_MESSAGE : null),
```
RubenCerna2079, line 238: What is the reason for this change?

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -86,7 +87,7 @@ public void ConfigureApiRoute(HttpContext httpContext)
         }
 
         // Executes the REST query by sending a GET request to the API.
-        public string? ExecuteRestQuery(string restUriSuffix, string entityName, int first)
+        public string? ExecuteRestQuery(string restUriSuffix, string entityName, int first, string _incomingRoleHeader, string _incomingRoleToken)
```
RubenCerna2079, line 90:
```suggestion
        public string? ExecuteRestQuery(string restUriSuffix, string entityName, int first, string incomingRoleHeader, string incomingRoleToken)
```

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -104,11 +105,22 @@ public void ConfigureApiRoute(HttpContext httpContext)
...
+                    }
+
+                    HttpResponseMessage response = client.SendAsync(message).Result;
                     if (response.IsSuccessStatusCode)
                     {
                         LogTrace($"The REST HealthEndpoint query executed successfully with code {response.IsSuccessStatusCode}.");
                     }
+                    else
+                    {
+                        errorMessage = $"The HealthEndpoint REST query failed with code: {response.StatusCode}.";
```
RubenCerna2079, line 122:
```suggestion
                        errorMessage = $"The REST HealthEndpoint query failed with code: {response.StatusCode}.";
```

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -122,7 +134,7 @@ public void ConfigureApiRoute(HttpContext httpContext)
 
         // Executes the GraphQL query by sending a POST request to the API.
         // Internally calls the metadata provider to fetch the column names to create the graphql payload.
-        public string? ExecuteGraphQLQuery(string graphqlUriSuffix, string entityName, Entity entity)
+        public string? ExecuteGraphQLQuery(string graphqlUriSuffix, string entityName, Entity entity, string _incomingRoleHeader, string _incomingRoleToken)
```
RubenCerna2079, line 137:
```suggestion
        public string? ExecuteGraphQLQuery(string graphqlUriSuffix, string entityName, Entity entity, string incomingRoleHeader, string incomingRoleToken)
```

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -144,17 +156,34 @@ public void ConfigureApiRoute(HttpContext httpContext)
...
+
+                        HttpResponseMessage response = client.SendAsync(message).Result;
+
                         if (response.IsSuccessStatusCode)
                         {
                             LogTrace("The GraphQL HealthEndpoint query executed successfully.");
                         }
+                        else
+                        {
+                            errorMessage = $"The HealthEndpoint GraphQL query failed with code: {response.StatusCode}.";
```
RubenCerna2079, line 185:
```suggestion
                            errorMessage = $"The GraphQL HealthEndpoint query failed with code: {response.StatusCode}.";
```

RubenCerna2079 (changes requested):

src/Service/HealthCheck/HealthCheckHelper.cs
sezal98: The threshold check can failr due to two conditions. Either it is -1 or the limit has exceeded the threshold. In case it is -1, we dont want to send the time exceeded msg even when we have the error message coming from the underlined function.

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -153,8 +142,8 @@ private void UpdateDataSourceHealthCheckResults(ref ComprehensiveHealthCheckRepo
             if (ComprehensiveHealthCheckReport.Checks != null && runtimeConfig.DataSource.IsDatasourceHealthEnabled)
             {
                 string query = Utilities.GetDatSourceQuery(runtimeConfig.DataSource.DatabaseType);
-                (int, string?) response = ExecuteDBQuery(query, runtimeConfig.DataSource.ConnectionString);
-                bool thresholdCheck = response.Item1 >= 0 && response.Item1 < runtimeConfig.DataSource.DatasourceThresholdMs;
+                (int, string?) response = ExecuteDatasourceQueryCheck(query, runtimeConfig.DataSource.ConnectionString);
+                bool IsResponseTimeWithinThreshold = response.Item1 >= 0 && response.Item1 < runtimeConfig.DataSource.DatasourceThresholdMs;
```
RubenCerna2079, line 146: Change variable name to `isResponseTimeWithinThreshold`

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -223,7 +212,7 @@ private void PopulateEntityHealth(ComprehensiveHealthCheckReport ComprehensiveHe
                     ComprehensiveHealthCheckReport.Checks ??= new List<HealthCheckResultEntry>();
                     string entityPath = entityValue.Rest.Path != null ? entityValue.Rest.Path.TrimStart('/') : entityKeyName;
                     (int, string?) response = ExecuteRestEntityQuery(runtimeConfig.RestPath, entityPath, entityValue.EntityFirst);
-                    bool thresholdCheck = response.Item1 >= 0 && response.Item1 < entityValue.EntityThresholdMs;
+                    bool IsResponseTimeWithinThreshold = response.Item1 >= 0 && response.Item1 < entityValue.EntityThresholdMs;
```
RubenCerna2079, line 215: Change to `isResponseTimeWithinThreshold`

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -245,7 +234,7 @@ private void PopulateEntityHealth(ComprehensiveHealthCheckReport ComprehensiveHe
                     ComprehensiveHealthCheckReport.Checks ??= new List<HealthCheckResultEntry>();
 
                     (int, string?) response = ExecuteGraphQLEntityQuery(runtimeConfig.GraphQLPath, entityValue, entityKeyName);
-                    bool thresholdCheck = response.Item1 >= 0 && response.Item1 < entityValue.EntityThresholdMs;
+                    bool IsResponseTimeWithinThreshold = response.Item1 >= 0 && response.Item1 < entityValue.EntityThresholdMs;
```
RubenCerna2079, line 237: Change to `isResponseTimeWithinThreshold`

src/Config/Converters/DataSourceConverterFactory.cs
```diff
@@ -151,7 +146,7 @@ public DataSourceConverter(bool replaceEnvVar)
                 }
             }
 
-            throw new JsonException("Datasource is a mandatory field and cannot be null.");
+            throw new JsonException("data-source property has a missing }.");
```
RubenCerna2079, line 149: Change comment to `"The 'data-source' property has a missing '}' at the end of the line."`

src/Config/Converters/DatasourceHealthOptionsConvertorFactory.cs
```diff
@@ -87,7 +109,7 @@ private class HealthCheckOptionsConverter : JsonConverter<DatasourceHealthCheckC
                 }
             }
 
-            throw new JsonException("Datasource HealthOptions must be a proper object.");
+            throw new JsonException("Datasource Health Options has a missing }.");
```
RubenCerna2079, line 112: Change comment to `"Datasource Health Options has a missing '}' at the end of the line."`

src/Config/Converters/EntityHealthOptionsConvertorFactory.cs
```diff
@@ -93,7 +93,7 @@ private class HealthCheckOptionsConverter : JsonConverter<EntityHealthCheckConfi
                 }
             }
 
-            throw new JsonException("EntityHealth Options must be a proper object.");
+            throw new JsonException("Entity Health Options has a missing }.");
```
RubenCerna2079, line 96: Change comment to `"Entity Health Options has a missing '}' at the end of the line."`

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -104,11 +106,14 @@ private class HealthCheckOptionsConverter : JsonConverter<RuntimeHealthCheckConf
...
                             break;
+
+                        default:
+                            throw new JsonException($"Unexpected property {property}");
                     }
                 }
             }
 
-            throw new JsonException();
+            throw new JsonException("Runtime Health Options has a missing }.");
```
RubenCerna2079, line 116: Change comment to `"Runtime Health Options has a missing '}' at the end of the line."`

RubenCerna2079 (dismissed):
LGTM! Thanks for addressing my previous comments and questions, only needs some small changes 😄

RubenCerna2079 (dismissed):
LGTM

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -69,6 +69,14 @@ public Task WriteResponse(HttpContext context)
             // Global comprehensive Health Check Enabled
             if (config.IsHealthEnabled)
             {
+                _healthCheckHelper.UpdateIncomingRoleHeader(context);
+                if (!_healthCheckHelper.IsUserAllowedToAccessHealthCheck(context, config.IsDevelopmentMode(), config.AllowedRolesForHealth))
+                {
+                    LogTrace("Comprehensive Health Check Report is not allowed: 403 Forbidden due to insufficient permissions.");
```
Aniruddh25, line 75:
```suggestion
                    LogError("Comprehensive Health Check Report is not allowed: 403 Forbidden due to insufficient permissions.");
```

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -57,6 +61,38 @@ public ComprehensiveHealthCheckReport GetHealthCheckResponse(HttpContext context
...
+        /// <returns></returns>
+        public bool IsUserAllowedToAccessHealthCheck(HttpContext httpContext, bool isDevelopmentMode, List<string> allowedRoles)
+        {
+            if (allowedRoles == null || allowedRoles.Count == 0)
+            {
+                // When allowedRoles is null or empty, all roles are allowed if Mode = Development.
+                return isDevelopmentMode;
+            }
+
+            return allowedRoles.Contains(_incomingRoleHeader);
```
Aniruddh25, line 93: If we just want to check for presence of this `roleHeader`, shouldn't we create a `HashSet` for allowedRoles instead of a List<string> so that its O(1) time complexity?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -57,6 +61,38 @@ public ComprehensiveHealthCheckReport GetHealthCheckResponse(HttpContext context
...
+        public void UpdateIncomingRoleHeader(HttpContext httpContext)
+        {
+            StringValues clientRoleHeader = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER];
+            StringValues clientTokenHeader = httpContext.Request.Headers[AuthenticationOptions.CLIENT_PRINCIPAL_HEADER];
+
+            if (clientRoleHeader.Count == 1)
+            {
+                _incomingRoleHeader = clientRoleHeader.ToString().ToLowerInvariant();
+                _incomingRoleToken = clientTokenHeader.ToString();
+            }
```
Aniruddh25, line 74: what happens if clientRoleHeader Count is > 1 or 0?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -57,6 +61,38 @@ public ComprehensiveHealthCheckReport GetHealthCheckResponse(HttpContext context
             return ComprehensiveHealthCheckReport;
         }
 
+        // Updates the incoming role header with the appropriate value from the request headers.
+        public void UpdateIncomingRoleHeader(HttpContext httpContext)
```
Aniruddh25, line 65: nit: consider renaming this function to `StoreIncomingRoleHeader` since we are storing the role header from the httpContext to a private member. `Update` sounds like we are updating the role header in the context itself since external functions shouldnt be knowing private member names.

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -104,11 +105,22 @@ public void ConfigureApiRoute(HttpContext httpContext)
...
+                    if (string.IsNullOrEmpty(incomingRoleToken) && string.IsNullOrEmpty(incomingRoleHeader))
+                    {
+                        message.Headers.Add(AuthenticationOptions.CLIENT_PRINCIPAL_HEADER, incomingRoleToken);
+                        message.Headers.Add(AuthorizationResolver.CLIENT_ROLE_HEADER, incomingRoleHeader);
+                    }
+
+                    HttpResponseMessage response = client.SendAsync(message).Result;
                     if (response.IsSuccessStatusCode)
                     {
                         LogTrace($"The REST HealthEndpoint query executed successfully with code {response.IsSuccessStatusCode}.");
```
Aniruddh25, line 118: All the exceptions should be logged as `LogError`. I am not sure if we have a followup issue created for this.

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -104,11 +105,22 @@ public void ConfigureApiRoute(HttpContext httpContext)
                 {
                     // Send a GET request to the API
                     apiRoute = $"{apiRoute}{Utilities.CreateHttpRestQuery(entityName, first)}";
-                    HttpResponseMessage response = client.GetAsync(apiRoute).Result;
+                    HttpRequestMessage message = new(method: HttpMethod.Get, requestUri: apiRoute);
+                    if (string.IsNullOrEmpty(incomingRoleToken) && string.IsNullOrEmpty(incomingRoleHeader))
```
Aniruddh25, line 109: Shouldnt this be checking for `!string.IsNullOrEmpty()` i.e. add the role header and token only if it is NOT empty?

src/Service.Tests/Configuration/HealthEndpointRolesTests.cs
```diff
@@ -0,0 +1,129 @@
...
+                message.Headers.Add(AuthenticationOptions.CLIENT_PRINCIPAL_HEADER, swaTokenPayload);
+                message.Headers.Add(AuthorizationResolver.CLIENT_ROLE_HEADER, STARTUP_CONFIG_ROLE);
+                HttpResponseMessage authorizedResponse = await client.SendAsync(message);
+
+                switch (role)
+                {
+                    case null:
+                        if (hostMode == HostMode.Development)
+                        {
+                            Assert.AreEqual(expected: HttpStatusCode.OK, actual: authorizedResponse.StatusCode);
```
Aniruddh25, line 78: I dont see any tests for checking roles in executing REST/GraphQL query for each entity. We should Assert that the respective entity checks pass as well, (in addition to the Overall Status code of ok)

Aniruddh25 (changes requested):
Need more tests/assertion on the Rest Query/GraphQL query execution.

Aniruddh25:
nit: please provide a better title to the PR that explains what this PR is doing.

src/Service/HealthCheck/HttpUtilities.cs
sezal98: yes have a tracking task for this

src/Service.Tests/Configuration/HealthEndpointRolesTests.cs
sezal98: we can discuss this once, Having trouble writing this test case as the connection while running the test case is already made to /health endpoint. Its unable to run the rest or graphql query and get a response.

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -57,6 +62,44 @@ public ComprehensiveHealthCheckReport GetHealthCheckResponse(HttpContext context
...
+            if (clientRoleHeader.Count > 1 || clientTokenHeader.Count > 1)
+            {
+                throw new ArgumentException("Multiple values for the client role or token header are not allowed.");
+            }
+
+            // Role Header is not present in the request, set it to anonymous.
+            if (clientRoleHeader.Count == 1)
+            {
+                _incomingRoleHeader = clientRoleHeader.ToString().ToLowerInvariant();
+                _incomingRoleToken = clientTokenHeader.ToString();
```
Aniruddh25, line 80: The client role header is not mandatory for the ClientTokenHeader to be defined. So, we should set the volue of the token _incomingRoleToken irrespective of the clientRoleHeader.

Aniruddh25: https://learn.microsoft.com/en-us/azure/data-api-builder/authentication-azure#azure-static-web-apps-authentication-easyauth

See here. these are independent headers. 1 can be present without the other. `X-MS-API-ROLE` header can optionally be defined by the client app to provide the value of a custom role other than `anonymous` and `authenticated`.

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -144,17 +156,34 @@ public void ConfigureApiRoute(HttpContext httpContext)
...
                         string jsonPayload = Utilities.CreateHttpGraphQLQuery(databaseObjectName, columnNames, entity.EntityFirst);
                         HttpContent content = new StringContent(jsonPayload, Encoding.UTF8, Utilities.JSON_CONTENT_TYPE);
-                        HttpResponseMessage response = client.PostAsync(apiRoute, content).Result;
+
+                        HttpRequestMessage message = new(method: HttpMethod.Post, requestUri: apiRoute)
+                        {
+                            Content = content
+                        };
+
+                        if (!(string.IsNullOrEmpty(incomingRoleToken) && string.IsNullOrEmpty(incomingRoleHeader)))
```
Aniruddh25, line 171: Can we have this if condition separately for each header - since they are unrelated?

Aniruddh25: also, can the header be added with the value = an empty string? because with the current condition, even if one of these tokens are empty and the other not empty, the headers will be added.

src/Service/HealthCheck/HealthCheckHelper.cs
sezal98: separated them out. Like they were before

RubenCerna2079 (approved):
LGTM

src/Service.Tests/Configuration/HealthEndpointRolesTests.cs
Aniruddh25: Better resolution of these tests - Pushed to RC2

Issue #2534, opened 2025-01-23
HealthEndpoint: Caching of HealthCheck queries


sezal98:
We need to introduce a caching behaviour in HealthCheck queries so that we dont need to bumbard the DB again and again and return our values from the health check cache based on an time interval or ttl of cache

JerryNixon:
Let's copy the current TTL default and set it to be 5 seconds. Hard coded right now is fine.

JerryNixon:
Remember concurrent calls for health should be disallowed (two at the same time).

PR #2633, opened 2025-03-24
Caching response from Health Endpoint

sezal98:
## Why make this change?
Closes [#2534](https://github.com/Azure/data-api-builder/issues/2534)

## What is this change?
This PR involves creating another parameter in the runtime config which defines the cache-ttl-seconds for the health endpoint. 
+ I am using the IMemoryCache to define the cache which has the CacheKey: `HealthCheckResponse` in ComprehensiveResponseWriter and saves the serialized version of the response in the cache. 

+ The changes in Convertor Factory are in respect to the serialization and deserialization of this attribute appropriately. We consider default as 5 seconds in case not provided in the config file. And in case this is 0, caching is disabled. 

+ Added timestamp to the health check report

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)
Call this CURL
curl --request GET \
  --url http://localhost:5000/health \
  --header 'User-Agent: insomnia/10.3.0'
Define the cache value as 5. Then in case of hitting the request without any delay, we get the same response back from the cache. In case hitting after the delay, we get a new response.


src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -62,32 +68,72 @@ public ComprehensiveHealthReportResponseWriter(
...
         /// <param name="context">HttpContext for writing the response.</param>
         /// <returns>Writes the http response to the http context.</returns>
-        public Task WriteResponse(HttpContext context)
+        public async Task WriteResponse(HttpContext context)
         {
             RuntimeConfig config = _runtimeConfigProvider.GetConfig();
 
             // Global comprehensive Health Check Enabled
-            if (config.IsHealthEnabled)
+            if (config != null && config.IsHealthEnabled)
```
RubenCerna2079, line 76: Why are you checking if the config is null when the `GetConfig()` function already ensures that?

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -62,32 +68,72 @@ public ComprehensiveHealthReportResponseWriter(
...
+                if (config.CacheTtlSeconds != null && config.CacheTtlSeconds > 0)
+                {
+                    if (!_cache.TryGetValue(CACHE_KEY, out response))
+                    {
+                        ComprehensiveHealthCheckReport dabHealthCheckReport = _healthCheckHelper.GetHealthCheckResponse(context, config);
+
+                        response = JsonSerializer.Serialize(dabHealthCheckReport, options: new JsonSerializerOptions { WriteIndented = true, DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull });
+                        LogTrace($"Health check response writer writing status as: {dabHealthCheckReport.Status}");
+
+                        // Cache the response for 5 minutes (or any other duration you prefer)
```
RubenCerna2079, line 98: This sounds like a comment for the user

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -62,32 +68,72 @@ public ComprehensiveHealthReportResponseWriter(
...
+                        await context.Response.WriteAsync("Failed to generate health check response.");
+                    }
+                }
+                else
+                {
+                    ComprehensiveHealthCheckReport dabHealthCheckReport = _healthCheckHelper.GetHealthCheckResponse(context, config);
+                    response = JsonSerializer.Serialize(dabHealthCheckReport, options: new JsonSerializerOptions { WriteIndented = true, DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull });
+                    LogTrace($"Health check response writer writing status as: {dabHealthCheckReport.Status}");
+                    await context.Response.WriteAsync(response);
+
```
RubenCerna2079, line 126: nit: Delete this line

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -62,32 +68,72 @@ public ComprehensiveHealthReportResponseWriter(
...
+                string? response;
+                // Check if the cache is enabled 
+                if (config.CacheTtlSeconds != null && config.CacheTtlSeconds > 0)
+                {
+                    if (!_cache.TryGetValue(CACHE_KEY, out response))
+                    {
+                        ComprehensiveHealthCheckReport dabHealthCheckReport = _healthCheckHelper.GetHealthCheckResponse(context, config);
+
+                        response = JsonSerializer.Serialize(dabHealthCheckReport, options: new JsonSerializerOptions { WriteIndented = true, DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull });
+                        LogTrace($"Health check response writer writing status as: {dabHealthCheckReport.Status}");
```
RubenCerna2079, line 96: Do you think it would look better if it was all capitalized or linked with `-`?
```suggestion
                        LogTrace($"Health Check Response Writer writing status as: {dabHealthCheckReport.Status}");
```

src/Service/HealthCheck/Model/ComprehensiveHealthCheckReport.cs
```diff
@@ -36,6 +37,12 @@ public record ComprehensiveHealthCheckReport
         [JsonPropertyName("app-name")]
         public string? AppName { get; set; }
 
+        /// <summary>
+        /// The timestamp of the response.
+        /// </summary>
+        [JsonPropertyName("timestamp")]
+        public DateTime TimeStamp { get; set; }
```
RubenCerna2079, line 44: Out of curiosity, what is the purpose of having the timestamp? Will it come out with the response?

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -58,6 +59,19 @@ private class HealthCheckOptionsConverter : JsonConverter<RuntimeHealthCheckConf
                                 enabled = reader.GetBoolean();
                             }
 
+                            break;
+                        case "cache-ttl-seconds":
+                            if (reader.TokenType is not JsonTokenType.Null)
+                            {
+                                int parseTtlSeconds = reader.GetInt32();
```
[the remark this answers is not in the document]
sezal98: The upper limit for now is mas limit for INT. We can modify in case Jerry suggests

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -58,6 +59,19 @@ private class HealthCheckOptionsConverter : JsonConverter<RuntimeHealthCheckConf
...
+                                if (parseTtlSeconds < 0)
+                                {
+                                    throw new JsonException($"Invalid value for ttl-seconds: {parseTtlSeconds}. Value must be greater than 0.");
```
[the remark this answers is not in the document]
sezal98: value can be equal to 0. This would mean that cache is disabled

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -158,6 +158,10 @@ Runtime.GraphQL is not null &&
     public HostMode HostMode =>
         Runtime?.Host?.Mode ?? HostMode.Production;
 
+    [JsonIgnore]
+    public int? CacheTtlSeconds =>
```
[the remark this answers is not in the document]
sezal98: modified

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                Assert.AreNotEqual(responseContent2, responseContent1);
+            }
+            else
+            {
+                Assert.AreEqual(responseContent2, responseContent1);
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
sezal98: no, we only have one form of caching for the health endpoint. It either caches the entire report or nothing at all

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
sezal98: i think capitalized or not it should be okay since we are not quoting a particular file.

src/Service/HealthCheck/Model/ComprehensiveHealthCheckReport.cs
sezal98: yes, suggested by jerry to show when was this health report calculated

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -1,24 +1,42 @@
...
+    public int? CacheTtlSeconds { get; set; }
 
+    [JsonPropertyName("roles")]
     public List<string>? Roles { get; set; }
 
     // TODO: Add support for parallel stream to run the health check query in upcoming PRs
     // public int MaxDop { get; set; } = 1; // Parallelized streams to run Health Check (Default: 1)
 
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
+    [MemberNotNullWhen(true, nameof(CacheTtlSeconds))]
```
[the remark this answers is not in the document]
aaronburtle: I dont think that `MemberNotNullWhen(True)` implies the member will be null when false. Simply that if it is true it is guaranteed to be non null. When it is false I don't think we have any guarantee.

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    private const string CUSTOM_CONFIG_FILENAME = "custom-config.json";
+
+    /// <summary>
+    /// Simulates two GET requests to DAB's comprehensive health check endpoint ('/health') and validates the contents of the response.
+    /// The expected behavior is that both these response should be same in case cache is enabled with no delay and different in case of delay.
+    /// </summary>
+    [TestMethod]
+    [TestCategory(TestCategory.MSSQL)]
+    [DataRow(true, null, DisplayName = "Validation of default cache TTL and delay.")]
```
aaronburtle, line 32: Would prefer to have 2 tests, one for a delay, and one for no delay, rather than including a bool parameter with only 1 test handling both cases. Bools in tests like this tend to reduce readability, and it is something we've been trying to refactor out of already existing tests.

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Helper function to write custom configuration file. with minimal REST/GraphQL global settings
+    /// using the supplied entities.
+    /// </summary>
+    /// <param name="entityMap">Collection of entityName -> Entity object.</param>
+    /// <param name="cacheTtlSeconds">flag to enable or disabled REST globally.</param>
+    private static void CreateCustomConfigFile(Dictionary<string, Entity> entityMap, int? cacheTtlSeconds)
```
aaronburtle, line 100: We have a helper function in SqlTestHelper called `InitBasicRuntimeConfigWithNoEntity` which creates a minimum config that is close to what you are adding here, I wonder if these could combine into a single helper function. Maybe your case is too different though with health and entities to make it worthwhile.

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -74,20 +80,66 @@ public Task WriteResponse(HttpContext context)
...
+                    if (response != null)
+                    {
+                        // Return the cached or newly generated response
+                        await context.Response.WriteAsync(response);
+                    }
+                    else
+                    {
+                        // Handle the case where cachedResponse is still null
+                        LogTrace("Error: The cached health check response is null.");
+                        context.Response.StatusCode = 500; // Internal Server Error
```
aaronburtle, line 123: is this an Internal Server Error or a DAB error?

src/Service/Startup.cs
```diff
@@ -253,6 +253,9 @@ public void ConfigureServices(IServiceCollection services)
                 ConfigureAuthentication(services, configProvider);
             }
 
+            // Register IMemoryCache
```
aaronburtle, line 256: I am curious if you looked at FusionCache, which is what we use for our regular query caching. If so, did IMemoryCache offer some advantage?

https://github.com/ZiggyCreatures/FusionCache

aaronburtle (dismissed):
Looks really good, just had a couple questions and comments. Should be good to go once those are addressed as you see fit.

schemas/dab.draft.schema.json
```diff
@@ -407,6 +407,11 @@
                 "type": "string"
               },
               "default": null
+            },
+            "cache-ttl-seconds": {
+              "type": "integer",
+              "description": "Time to live in seconds for the Comprehensive Health Check Report.",
```
RubenCerna2079, line 413:
```suggestion
              "description": "Time to live of cache in seconds for the Comprehensive Health Check Report.",
```

src/Service/Startup.cs
RubenCerna2079: I'm also quite curious about this :)

schemas/dab.draft.schema.json
Aniruddh25: "description": "Time to live in seconds for the Comprehensive Health Check Report cache entry."

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -73,6 +74,19 @@ internal HealthCheckOptionsConverter(bool replaceEnvVar)
...
                             }
 
+                            break;
+                        case "cache-ttl-seconds":
+                            if (reader.TokenType is not JsonTokenType.Null)
+                            {
+                                int parseTtlSeconds = reader.GetInt32();
+                                if (parseTtlSeconds < 0)
+                                {
+                                    throw new JsonException($"Invalid value for ttl-seconds: {parseTtlSeconds}. Value must be greater than or equal to 0.");
```
RubenCerna2079, line 84:
```suggestion
                                    throw new JsonException($"Invalid value for health cache-ttl-seconds: {parseTtlSeconds}. Value must be greater than or equal to 0.");
```

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -74,20 +80,66 @@ public Task WriteResponse(HttpContext context)
...
+                    return;
                 }
 
-                ComprehensiveHealthCheckReport dabHealthCheckReport = _healthCheckHelper.GetHealthCheckResponse(context, config);
-                string response = JsonSerializer.Serialize(dabHealthCheckReport, options: new JsonSerializerOptions { WriteIndented = true, DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull });
-                LogTrace($"Health check response writer writing status as: {dabHealthCheckReport.Status}");
-                return context.Response.WriteAsync(response);
+                string? response;
+                // Check if the cache is enabled 
+                if (config.CacheTtlSecondsForHealthReport != null && config.CacheTtlSecondsForHealthReport > 0)
```
RubenCerna2079, line 89:
```suggestion
                if (config.CacheTtlSecondsForHealthReport is not null && config.CacheTtlSecondsForHealthReport > 0)
```

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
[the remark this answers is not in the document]
Aniruddh25: This property is defined as non-nullable, do we still need the MemberNotNullWhen attribute? It can never be null right?

src/Service/Startup.cs
Aniruddh25: Me 3.

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -74,20 +80,66 @@ public Task WriteResponse(HttpContext context)
...
+                        LogTrace($"Health check response writer writing status as: {dabHealthCheckReport.Status}");
+
+                        try
+                        {
+                            // Cache the response for cache-ttl-seconds provided by the user in the config
+                            MemoryCacheEntryOptions cacheEntryOptions = new MemoryCacheEntryOptions()
+                                .SetAbsoluteExpiration(TimeSpan.FromSeconds((double)config.CacheTtlSecondsForHealthReport));
+
+                            _cache.Set(CACHE_KEY, response, cacheEntryOptions);
+                            LogTrace($"Health check response writer writing status as: {dabHealthCheckReport.Status}");
```
Aniruddh25, line 105: This trace will be written twice - once on line 96 and another time here on line 105 after caching. The second trace here is not required - unless we want to change the message to mean we are using this to cache.

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -74,20 +80,66 @@ public Task WriteResponse(HttpContext context)
...
+                    // Ensure cachedResponse is not null before calling WriteAsync
+                    if (response != null)
+                    {
+                        // Return the cached or newly generated response
+                        await context.Response.WriteAsync(response);
+                    }
+                    else
+                    {
+                        // Handle the case where cachedResponse is still null
+                        LogTrace("Error: The cached health check response is null.");
```
Aniruddh25, line 122: response could be uncached and still be null. is there a possibility of `JsonSerializer.Serialize(dabHealthCheckReport, options: new JsonSerializerOptions { WriteIndented = true, DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull });` evaluating to null?

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -74,20 +80,66 @@ public Task WriteResponse(HttpContext context)
...
+                        context.Response.StatusCode = 500; // Internal Server Error
+                        await context.Response.WriteAsync("Failed to generate health check response.");
+                    }
+                }
+                else
+                {
+                    ComprehensiveHealthCheckReport dabHealthCheckReport = _healthCheckHelper.GetHealthCheckResponse(context, config);
```
Aniruddh25, line 129: The code in this else block is repeated. Can be put in a common function to avoid duplication.

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            string responseContent2 = await response.Content.ReadAsStringAsync();
+
+            // Regex to remove the "timestamp" field from the JSON string
+            string pattern = @"""timestamp""\s*:\s*""[^""]*""\s*,?";
+            responseContent1 = Regex.Replace(responseContent1, pattern, string.Empty);
+            responseContent2 = Regex.Replace(responseContent2, pattern, string.Empty);
+
+            if (delay || cacheTtlSeconds == 0)
+            {
+                Assert.AreNotEqual(responseContent2, responseContent1);
```
Aniruddh25, line 85: Is there an example of the content that is expected to be different when there is a delay? If the timestamp is overwritten by string.Empty - I would assume the health check would return the same response irrespective of if caching is enabled or not.

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+
+            HttpRequestMessage healthRequest2 = new(HttpMethod.Get, "/health");
+            response = await client.SendAsync(healthRequest2);
+            string responseContent2 = await response.Content.ReadAsStringAsync();
+
+            // Regex to remove the "timestamp" field from the JSON string
+            string pattern = @"""timestamp""\s*:\s*""[^""]*""\s*,?";
+            responseContent1 = Regex.Replace(responseContent1, pattern, string.Empty);
+            responseContent2 = Regex.Replace(responseContent2, pattern, string.Empty);
```
Aniruddh25, line 81: I would not expect the requirement to replace the timeStamp from the response - because if its coming from the cache - it should return the old timestamp, no?

Aniruddh25 (changes requested):
Please update PR title to better explain what it is doing e.g. `Caching response from Health Endpoint` 

Looking for clarification on tests.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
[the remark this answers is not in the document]
sezal98: The MemberNotNullWhen attribute is used to specify that a particular member (property or field) of a class is guaranteed not null when a certain condition is true. Let's consider a scenario where you have a class with a property, and the property is null only if a method returns false. You can use MemberNotNullWhen to tell the compiler that the property will not be null when the method returns true.

Hence, when user provides a value for cache ttl, we say that this property of `UserDefinedTtl` cannot be null.

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
sezal98: yes these methods are similar but i need to add specific properties related to caching. And overloading this function with more parameters to pass and hinder all the other tests seems unnecessary. Hence created a new one in the file itself like in `configurationTests.cs`

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
sezal98: since we are comapring values, its better to use == or !=. We should use is and is not when comparing types

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
sezal98: modified

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
sezal98: The response could never be null. It will always add the basic details in the `BasicHealthReport`
image

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
sezal98: No the DB call returns responseTime in ms different

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
sezal98: when we compare the responses and the two have diff results then we remove the timestamp to compare.

src/Service/Startup.cs
sezal98: modified

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -1,24 +1,41 @@
...
     public RuntimeHealthCheckConfig() : base()
     {
     }
 
-    public RuntimeHealthCheckConfig(bool? Enabled, HashSet<string>? Roles = null) : base(Enabled)
+    public RuntimeHealthCheckConfig(bool? Enabled, HashSet<string>? Roles = null, int? CacheTtlSeconds = null) : base(Enabled)
     {
         this.Roles = Roles;
+
+        if (CacheTtlSeconds is not null)
```
Aniruddh25, line 31:
```suggestion
        if (cacheTtlSeconds is not null)
```

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -1,24 +1,41 @@
...
+        {
+            this.CacheTtlSeconds = (int)CacheTtlSeconds;
```
Aniruddh25, line 33:
```suggestion
            this.CacheTtlSeconds = (int)cacheTtlSeconds;
```

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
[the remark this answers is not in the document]
Aniruddh25: But @sezal98, the property `CacheTtlSeconds` can never be null based on how we defined its data type, right? We defined it as `int`, not `int?` - so CacheTtlSeconds is ALWAYS guaranteed to be non-null.
We need this attribute only if `CacheTtlSeconds` is defined as `int?`

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
Aniruddh25: Looks like we are removing the timestamp without checking if the results are different or not.

Aniruddh25: My bad, looks like this test expects the two results to be different. So, removing timestamp here makes sense.

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        using (TestServer server = new(Program.CreateWebHostBuilder(args)))
+        using (HttpClient client = server.CreateClient())
+        {
+            HttpRequestMessage healthRequest1 = new(HttpMethod.Get, "/health");
+            HttpResponseMessage response = await client.SendAsync(healthRequest1);
+            string responseContent1 = await response.Content.ReadAsStringAsync();
+            Assert.AreEqual(expected: HttpStatusCode.OK, actual: response.StatusCode, message: "Received unexpected HTTP code from health check endpoint.");
+
+            // No delay, force a new request to be sent to the database.
```
Aniruddh25, line 134:
```suggestion
            // No delay, so we won't force a new request to be sent to the database.
```

Shouldn't this be we won't force since there is no delay?

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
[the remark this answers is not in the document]
sezal98: yes that is correct. Good catch. This was changed later on in the suggestions. Updated now. Thanks

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
sezal98: Modified the comment to 
```
// Simulate no delay scenario to make sure that the cache is not expired (in case available)
// and force a new request to be sent to the database.
// Further check if the responses are same.
```
 Does this make more sens?

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -87,6 +87,7 @@ public async Task ComprehensiveHealthEndpointCachingValidateWithDelay(int? cache
             responseContent1 = Regex.Replace(responseContent1, pattern, string.Empty);
             responseContent2 = Regex.Replace(responseContent2, pattern, string.Empty);
 
+            // Response are not the same as a new request was made to the DB (change in responseTimeMs for DB health check)
```
Aniruddh25, line 90:
```suggestion
            // Responses are not the same as a new request was made to the DB (change in responseTimeMs for DB health check)
```

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -143,10 +144,13 @@ public async Task ComprehensiveHealthEndpointCachingValidateNoDelay(int? cacheTt
                 string pattern = @"""timestamp""\s*:\s*""[^""]*""\s*,?";
                 responseContent1 = Regex.Replace(responseContent1, pattern, string.Empty);
                 responseContent2 = Regex.Replace(responseContent2, pattern, string.Empty);
+
+                // Response are not the same as a new request was made to the DB (change in responseTimeMs for DB health check)
```
Aniruddh25, line 148:
```suggestion
                // Responses are not the same as a new request was made to the DB (change in responseTimeMs for DB health check)
```

Aniruddh25 (dismissed):
LGTM, thanks for clarifying all the outstanding questions irrespective of the challenging time zones.

Aniruddh25 (approved):

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            File.Delete(CUSTOM_CONFIG_FILENAME);
+        }
+
+        TestHelper.UnsetAllDABEnvironmentVariables();
+    }
+
+    /// <summary>
+    /// Simulates GET requests to DAB's comprehensive health check endpoint ('/health') and validates the contents of the response.
+    /// The expected behavior is that these response should be different as we supply delay here.
```
RubenCerna2079, line 39:
```suggestion
    /// The expected behavior is that these responses should be different as we supply a delay between the requests.
```

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,190 @@
...
+            responseContent2 = Regex.Replace(responseContent2, pattern, string.Empty);
+
+            // Responses are not the same as a new request was made to the DB (change in responseTimeMs for DB health check)
+            Assert.AreNotEqual(responseContent2, responseContent1);
+        }
+    }
+
+    /// <summary>
+    /// Simulates GET request to DAB's comprehensive health check endpoint ('/health') and validates the contents of the response.
+    /// The expected behavior is that both these response should be same in case cache is enabled with no delay.
```
RubenCerna2079, line 97:
```suggestion
    /// The expected behavior is that both responses should be same since the cache is enabled and there is no delay between requests.
```

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
```diff
@@ -74,20 +81,71 @@ public Task WriteResponse(HttpContext context)
...
+                        response = await _cache.GetOrSetAsync<string?>(
+                            key: CACHE_KEY,
+                            async (FusionCacheFactoryExecutionContext<string?> ctx, CancellationToken ct) =>
+                            {
+                                string? response = await ExecuteHealthCheck(context, config).ConfigureAwait(false);
+                                ctx.Options.SetDuration(TimeSpan.FromSeconds(config.CacheTtlSecondsForHealthReport));
+                                return response;
+                            });
+
+                        _logger.LogTrace($"Health check response is fetched from cache with key: {CACHE_KEY} and TTL: {config.CacheTtlSecondsForHealthReport} seconds.");
```
RubenCerna2079, line 103: Is there any reason why we want to give the cache key?

RubenCerna2079 (approved):
LGTM, I only left a few comments

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -154,6 +154,10 @@ Runtime.GraphQL is not null &&
     public HashSet<string> AllowedRolesForHealth =>
         Runtime?.Health?.Roles ?? new HashSet<string>();
 
+    [JsonIgnore]
+    public int CacheTtlSecondsForHealthReport =>
+        Runtime?.Health?.CacheTtlSeconds ?? EntityCacheOptions.DEFAULT_TTL_SECONDS;
+
     /// <summary>
     /// Retrieves the value of runtime.graphql.dwnto1joinopt.enabled property if present, default is false.
```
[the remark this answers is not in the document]
sezal98: In line 99 `if (config.CacheTtlSecondsForHealthReport > 0)` to check if the cache is enabled or not in ComprehensiveResponseWriter

src/Service/HealthCheck/ComprehensiveHealthReportResponseWriter.cs
sezal98: This is a trace log for debugging, we can give it here.

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -73,6 +74,19 @@ internal HealthCheckOptionsConverter(bool replaceEnvVar)
                                 enabled = reader.GetBoolean();
                             }
 
+                            break;
+                        case "cache-ttl-seconds":
+                            if (reader.TokenType is not JsonTokenType.Null)
+                            {
+                                int parseTtlSeconds = reader.GetInt32();
```
aaronburtle, line 81: nit:
```suggestion
                                int parsedTtlSeconds = reader.GetInt32();
```

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,190 @@
...
+    /// The expected behavior is that both these responses should be same in case cache is enabled with no delay between the two requests.
+    /// </summary>
+    [TestMethod]
+    [TestCategory(TestCategory.MSSQL)]
+    [DataRow(null, DisplayName = "Validation of default cache TTL and no delay.")]
+    [DataRow(0, DisplayName = "Validation of cache TTL set to 0 and no delay.")]
+    [DataRow(10, DisplayName = "Validation of cache TTL set to 10 and no delay.")]
+    public async Task ComprehensiveHealthEndpointCachingValidateNoDelay(int? cacheTtlSeconds)
+    {
+        Entity requiredEntity = new(
```
aaronburtle, line 106: the arrange portion of these tests seem to be doing the same thing, create an entity, create an entity map, and then create a custom config. Can we move that all to a shared helper function they both call?

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -0,0 +1,190 @@
...
+        };
+
+        CreateCustomConfigFile(entityMap, cacheTtlSeconds);
+
+        string[] args = new[]
+        {
+            $"--ConfigFileName={CUSTOM_CONFIG_FILENAME}"
+        };
+
+        using (TestServer server = new(Program.CreateWebHostBuilder(args)))
```
aaronburtle, line 127: the act portion of these tests also seem to share a bunch of code, with the difference being the delay. Can we move that to a shared helper function as well?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -140,20 +148,20 @@ private static void UpdateDabConfigurationDetails(ref ComprehensiveHealthCheckRe
         }
 
         // Main function to internally call for data source and entities health check.
-        private void UpdateHealthCheckDetails(ref ComprehensiveHealthCheckReport ComprehensiveHealthCheckReport, RuntimeConfig runtimeConfig)
+        private async Task UpdateHealthCheckDetails(ComprehensiveHealthCheckReport ComprehensiveHealthCheckReport, RuntimeConfig runtimeConfig)
```
aaronburtle, line 151: just curious how come these had to be changed from void to async?

aaronburtle (approved):
Looks great! Just had some questions, mostly about format.

src/Service/HealthCheck/HealthCheckHelper.cs
sezal98: moved from result to async call

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
```diff
@@ -82,19 +66,14 @@ public async Task ComprehensiveHealthEndpointCachingValidateWithDelay(int? cache
             response = await client.SendAsync(healthRequest2);
             string responseContent2 = await response.Content.ReadAsStringAsync();
 
-            // Regex to remove the "timestamp" field from the JSON string
-            string pattern = @"""timestamp""\s*:\s*""[^""]*""\s*,?";
```
RubenCerna2079: Just curious as to why was the regex pattern removed?

RubenCerna2079 (approved):
LGTM!

Aniruddh25 (approved):

src/Service.Tests/Caching/HealthEndpointCachingTests.cs
sezal98: This was checking timestamp after removing it. But thats not needed since caching is done of timestamp too. Returned response of timestamp would be same before and after the cache

Aniruddh25 (approved):

PR #2634, opened 2025-03-24
Hot Reload for Loggers

RubenCerna2079:
## Why make this change?
Closes issues #2580 and #2582

## What is this change?
Adds hot reload feature for loggers which makes it possible for the user to change the log level filters without restarting the application. This is done by ensuring that the logger factories are updated when there is a hot reload event.
This feature is also available in production mode, it is the only hot reloadable property in production mode.

## How was this tested?
- [ ] Integration Tests
- [x] Unit Tests
- [X] Manual Testing

Test added:
- Check if log-level value is properly updated with hot reload, by checking if it is different from previous value and if it is the expected value.

## Sample Request(s)
Before hot reload:
image
image

After hot reload:
image
image


src/Cli/ConfigGenerator.cs
```diff
@@ -1674,7 +1674,7 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
                 minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
                 HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
 
-                _logger.LogInformation("Setting default minimum LogLevel: {minimumLogLevel} for {hostMode} mode.", minimumLogLevel, hostModeType);
+                _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
```
RubenCerna2079, line 1677: Fixed an issue that I found with this logging message

Aniruddh25: thank you for noticing this and making our logs healthy

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -318,6 +318,8 @@ private void HotReloadConfig(bool isDevMode, ILogger? logger = null)
         IsNewConfigDetected = true;
         IsNewConfigValidated = false;
         SignalConfigChanged();
+
+        Console.WriteLine("Hot-reload process finished");
```
Aniruddh25, line 322: if we have the logger here, please use that.

src/Config/DabConfigEvents.cs
```diff
@@ -18,5 +18,5 @@ public static class DabConfigEvents
     public const string GRAPHQL_SCHEMA_REFRESH_ON_CONFIG_CHANGED = "GRAPHQL_SCHEMA_REFRESH_ON_CONFIG_CHANGED";
     public const string GRAPHQL_SCHEMA_EVICTION_ON_CONFIG_CHANGED = "GRAPHQL_SCHEMA_EVICTION_ON_CONFIG_CHANGED";
     public const string GRAPHQL_SCHEMA_CREATOR_ON_CONFIG_CHANGED = "GRAPHQL_SCHEMA_CREATOR_ON_CONFIG_CHANGED";
-
+    public const string LOG_LEVEL_INITIALIZER_ON_CONFIG_CHANGE = "LOG_LEVEL_FACTORY_ON_CONFIG_CHANGE";
```
Aniruddh25, line 21:
```suggestion
    public const string LOG_LEVEL_INITIALIZER_ON_CONFIG_CHANGE = "LOG_LEVEL_INITIALIZER_ON_CONFIG_CHANGE";
```

src/Config/HotReloadEventHandler.cs
```diff
@@ -49,7 +50,7 @@ public void Subscribe(string eventName, EventHandler<TEventArgs> handler)
     {
         if (_eventHandlers.ContainsKey(eventName))
         {
-            _eventHandlers[eventName] = handler;
+            _eventHandlers[eventName] += handler;
```
Aniruddh25, line 53: does this mean we had a bug previously? where only the last added handler would work?

RubenCerna2079: I wouldn't call it a bug, it just wasn't necessary before because the `+=` allows us to subscribe multiple instances of the same class, which is only really useful for the logger, since each of them has its own logger factory.

src/Service/Telemetry/LogLevelInitializer.cs
```diff
@@ -0,0 +1,51 @@
...
+    public class LogLevelInitializer
+    {
+        private RuntimeConfigProvider? _runtimeConfigProvider;
+        private string _loggerFilter;
+
+        public LogLevel MinLogLevel { get; private set; }
+
+        public LogLevelInitializer(LogLevel logLevel, string? loggerFilter, HotReloadEventHandler<HotReloadEventArgs>? handler)
+        {
+            handler?.Subscribe(LOG_LEVEL_INITIALIZER_ON_CONFIG_CHANGE, OnConfigChanged);
```
Aniruddh25, line 21: Shouldnt the runtimeConfigProvider be set from within the constructor?

src/Service/Startup.cs
```diff
@@ -194,32 +200,37 @@ public void ConfigureServices(IServiceCollection services)
...
-                ILoggerFactory? loggerFactory = CreateLoggerFactoryForHostedAndNonHostedScenario(serviceProvider, typeof(ClientRoleHeaderAuthenticationMiddleware).FullName);
+                LogLevelInitializer logLevelRuntime = new(MinimumLogLevel, typeof(ClientRoleHeaderAuthenticationMiddleware).FullName, _hotReloadEventHandler);
+                ILoggerFactory? loggerFactory = CreateLoggerFactoryForHostedAndNonHostedScenario(serviceProvider, logLevelRuntime);
                 return loggerFactory.CreateLogger<ClientRoleHeaderAuthenticationMiddleware>();
             });
 
             services.AddSingleton<ILogger<ConfigurationController>>(implementationFactory: (serviceProvider) =>
             {
-                ILoggerFactory? loggerFactory = CreateLoggerFactoryForHostedAndNonHostedScenario(serviceProvider, typeof(ConfigurationController).FullName);
+                LogLevelInitializer logLevelRuntime = new(MinimumLogLevel, typeof(ConfigurationController).FullName, _hotReloadEventHandler);
```
Aniruddh25, line 232: what does the Runtime suffix in the name of `logLevelRuntime` mean? 
Are there logLevelInitializers for each logger factory?

Aniruddh25 (approved):
Looks good overall, Great feature to be able to modify loglevel in production mode.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -318,6 +318,8 @@ private void HotReloadConfig(bool isDevMode, ILogger? logger = null)
         IsNewConfigDetected = true;
         IsNewConfigValidated = false;
         SignalConfigChanged();
+
+        logger?.LogInformation("Hot-reload process finished");
```
aaronburtle, line 322:
```suggestion
        logger?.LogInformation("Hot-reload process finished.");
```

src/Config/RuntimeConfigLoader.cs
```diff
@@ -89,22 +89,27 @@ protected void SignalConfigChanged(string message = "")
...
-        // MetadataProviderFactory has been updated with latest database object metadata.
-        // RuntimeConfig must already be updated and is implied to have been updated by the time
-        // this function is called.
-        OnConfigChangedEvent(new HotReloadEventArgs(AUTHZ_RESOLVER_ON_CONFIG_CHANGED, message));
-
-        // Order of event firing matters: Eviction must be done before creating a new schema and then updating the schema.
-        OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_EVICTION_ON_CONFIG_CHANGED, message));
-        OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_CREATOR_ON_CONFIG_CHANGED, message));
-        OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_REFRESH_ON_CONFIG_CHANGED, message));
+        if (RuntimeConfig is not null && RuntimeConfig.IsDevelopmentMode())
```
aaronburtle, line 92: Can you add a comment please that explains why we only do this in dev mode? Does this support log level hot reload in production mode?

src/Service/Program.cs
```diff
@@ -77,7 +78,7 @@ public static IHostBuilder CreateHostBuilder(string[] args)
                 .ConfigureWebHostDefaults(webBuilder =>
                 {
                     Startup.MinimumLogLevel = GetLogLevelFromCommandLineArgs(args, out Startup.IsLogLevelOverriddenByCli);
-                    ILoggerFactory? loggerFactory = GetLoggerFactoryForLogLevel(Startup.MinimumLogLevel);
+                    ILoggerFactory? loggerFactory = Program.GetLoggerFactoryForLogLevel(Startup.MinimumLogLevel);
```
aaronburtle, line 81: is this addition of Program needed?

src/Service/Program.cs
```diff
@@ -133,17 +134,26 @@ private static ParseResult GetParseResult(Command cmd, string[] args)
...
             return LoggerFactory
                 .Create(builder =>
                 {
                     // Category defines the namespace we will log from,
                     // including all sub-domains. ie: "Azure" includes
                     // "Azure.DataApiBuilder.Service"
-                    builder.AddFilter(category: "Microsoft", logLevel);
-                    builder.AddFilter(category: "Azure", logLevel);
-                    builder.AddFilter(category: "Default", logLevel);
+                    if (logLevelInitializer == null)
```
aaronburtle, line 145: nit: prefer `is` to `==`

src/Service/Program.cs
```diff
@@ -157,8 +167,15 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, Tele
                                 }
                             },
                             configureApplicationInsightsLoggerOptions: (options) => { }
-                        )
-                        .AddFilter<ApplicationInsightsLoggerProvider>(category: string.Empty, logLevel);
+                        );
+                        if (logLevelInitializer == null)
```
aaronburtle, line 171: nit: `is` vs `==`

src/Service/Telemetry/LogLevelInitializer.cs
```diff
@@ -0,0 +1,51 @@
...
+            MinLogLevel = logLevel;
+            if (loggerFilter != null)
```
aaronburtle, line 23: nit: `is not` vs `!=`

aaronburtle (dismissed):
Approved, but please address comments, and double check the production hot-reloading before merge.

src/Service/Startup.cs
RubenCerna2079: Will change name to `logLevelInit` as Runtime does not reflect what the variable is.
Yes, there is an initializer for each logger factory, since all of them need to be updated separately.

src/Config/RuntimeConfigLoader.cs
RubenCerna2079: Changed the place where it checks for production mode

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -316,6 +316,16 @@ public void ValidateConfig()
 
             _configLoader.IsNewConfigValidated = runtimeConfigValidator.TryValidateConfig(ConfigFilePath, loggerFactory).Result;
 
+            // Validates that only log-level is changed if DAB is in Production Mode.
+            if (!_configLoader.RuntimeConfig!.IsDevelopmentMode())
+            {
+                _configLoader.IsNewConfigValidated = _configLoader.IsConfigValidInProductionMode();
```
Aniruddh25, line 322: Why do we need to do this extra validation? It adds to the startup time, and that is not good for Production scenarios

Aniruddh25: Cant we ensure in code that hot reload only happens for loglevel in Production mode? We shouldnt need to compare the serialized config files, we can add a test case to do that. But shouldnt do the check in the engine.

src/Config/RuntimeConfigLoader.cs
```diff
@@ -105,6 +105,8 @@ protected void SignalConfigChanged(string message = "")
         OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_EVICTION_ON_CONFIG_CHANGED, message));
         OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_CREATOR_ON_CONFIG_CHANGED, message));
         OnConfigChangedEvent(new HotReloadEventArgs(GRAPHQL_SCHEMA_REFRESH_ON_CONFIG_CHANGED, message));
+
```
Aniruddh25, line 108: In a previous iteration, we had all the above events fired only in development mode - which was the right step to take. Why did we remove that condition? 

I noticed we are now disallowing any changes to config file - and erroring out, but that's an extra check we are doing. We can document that only changes to LogLevel will be honored for hot-reload in Production mode, and simply not fire the events for changes to other sections, even if they were changed in Production mode. 

We should not error out - that will regress existing customers as well who were not using hot reload.

src/Core/Configurations/RuntimeConfigProvider.cs
RubenCerna2079: Changed function so that it only inserts the loglevel into the last valid runtime config if it is in production mode

src/Config/RuntimeConfigLoader.cs
RubenCerna2079: Changed it back with new insertion of the loglevel property into the runtime config so that no errors occur

src/Config/RuntimeConfigLoader.cs
```diff
@@ -388,4 +395,29 @@ public void RestoreLkgConfig()
...
+        if (!RuntimeConfig!.IsDevelopmentMode())
+        {
+            // Creates copy of last valid runtime config and only adds the new logger level changes
+            RuntimeConfig runtimeConfigCopy = LastValidRuntimeConfig! with
+            {
+                Runtime = LastValidRuntimeConfig.Runtime! with
+                {
+                    Telemetry = LastValidRuntimeConfig.Runtime!.Telemetry! with
+                    {
+                        LoggerLevel = RuntimeConfig.Runtime!.Telemetry!.LoggerLevel
```
aaronburtle, line 415: Just want to double check that this flow still checks if the Log Level is valid before updating it, ie: someone changes loglevel to "foo", we don't just overwrite with the value of "foo"

aaronburtle (approved):
Looks good, just one question about setting the log level during a hot-reload in production scenario.

src/Config/RuntimeConfigLoader.cs
RubenCerna2079: This function is done before the `TryValidate` which should validate that the key is not something wrong (eg. "foo", "debug"), while the value is checked when it is deserialized which means it does it at the start of the hot reload (eg. "namespace": "foo")

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -314,17 +314,11 @@ public void ValidateConfig()
             ILogger<RuntimeConfigValidator> logger = loggerFactory.CreateLogger<RuntimeConfigValidator>();
             RuntimeConfigValidator runtimeConfigValidator = new(this, fileSystem, logger, true);
 
-            _configLoader.IsNewConfigValidated = runtimeConfigValidator.TryValidateConfig(ConfigFilePath, loggerFactory).Result;
+            // This function only works if DAB is started in Production Mode
+            _configLoader.InsertWantedChangesInProductionMode();
```
Aniruddh25, line 318: with this new approach, do we still hit the data source name mismatch issue? Because of which you needed to hot-reload it previously?

Aniruddh25 (approved):
Looks good to merge. Thanks for addressing the delicate issue around hot reloading only the log-level in Production mode.

src/Core/Configurations/RuntimeConfigProvider.cs
RubenCerna2079: Yes, I confirmed that the data source name mismatch is not a problem anymore

PR #2646, opened 2025-04-09
Add authentication validation to the JSON schema

### LINKED ISSUE #2643: Add Authentication providers to JSON schema for autocomplete and validation
At the time of writing, the JSON schema does not provide a list of possible authentication providers:

https://github.com/Azure/data-api-builder/blob/main/schemas/dab.draft.schema.json#L275-L299

- This means that the user needs to read the documentation (or the source code) to find out all possible values, which is a bad developer experience
- This means that there is no validation for the auth provider name until `dab start` is called, which badly affects the inner development loop.
- This means there is **also** no validation of the auth provider name + `jwt` object combination for `OAuth` providers, like `Custom`.
  - It would be great for a user to get an error in their editor when they use the `Custom` auth provider but haven't provided the required `jwt` object.

These issues all have unit tests and are validated at `dab start`. Adding these cases to the JSON Schema will make it easier for developers to understand the authentication options, reduce time needed to fix issues as validation will take place in the editor as well, and make it easier to use `Custom` validation as the required `jwt` option would be pointed out by the schema.

This change doesn't seem like it would be to difficult to implement. As far as I know:

- We need to add a list of auth providers to the JSON schema
- We need to add conditional logic to the JSON schema that errors out when devs forget to add `jwt` when `Custom` is used, and also errors out when it is used for non-`Custom` auth options as this isn't valid.
- We need to add a test that checks that the use of the `jwt` object with non-`Custom` auth providers throws an error.

I'd love to work on this in a while, when I'm done with moving houses and have some more headspace! :)

sander1095:
## Why make this change?

Closes #2643 

The JSON schema does not contain a list of auth providers, nor does it validate the use of the `jwt` property. This results in a suboptimal developer experience. By adding this list and validation, developers have an easier time learning about all the options and can be more productive.

## What is this change?

- This change modifies the JSON schema with a list of providers and validation of the providers and the `jwt` property. [StackOverflow helped me](https://stackoverflow.com/q/79557612/3013479) with the finer details of the validation.
  - I also added tests to validate my changes.
- Migrate from `NJsonSchema` to `Newtonsoft.Json.Schema` because `NJsonSchema` [doesn't validate `anyof + const` correctly](https://github.com/RicoSuter/NJsonSchema/issues/1596). `Newtonsoft.Json.Schema` does work in this case.
- Changed `https` to `http` in the schema file to be spec compliant.  Draft-07's original URL is http-based.
  - If you try to validate a document with the `https` url, it will fail because of the unknown schema URL.
  - This does **not** impact security. The schema should just be a placeholder for the json validator binary to know what rules to use, and it's redirected to https when you navigate to it, anyway.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)

image

image

image


schemas/dab.draft.schema.json
```diff
@@ -277,8 +277,33 @@
...
-                  "type": "string",
                   "description": "The name of authentication provider",
+                  "oneOf": [
+                    {
+                      "const": "StaticWebApps",
+                      "description": "Authentication provided by Azure Static Web Apps."
+                    },
+                    {
+                      "const": "EntraID",
+                      "description": "Authentication provided by Microsoft Entra ID (formerly Azure AD)."
```
Aniruddh25, line 288: Perhaps, mention in the description of EntraID as well that `jwt` property is required to configure similar to `Custom`?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1686,6 +1816,36 @@ public async Task TestBasicConfigSchemaWithNoOptionalFieldsIsValid(string jsonDa
...
+
+            JsonSchemaValidationResult result = await jsonSchemaValidator.ValidateJsonConfigWithSchemaAsync(jsonSchema, jsonData);
+            Assert.IsFalse(result.IsValid);
+            Assert.IsFalse(EnumerableUtilities.IsNullOrEmpty(result.ValidationErrors));
+            schemaValidatorLogger.Verify(
+                x => x.Log(
+                    LogLevel.Information,
+                    It.IsAny<EventId>(),
+                    It.Is<It.IsAnyType>((o, t) => o.ToString()!.Contains($"The config satisfies the schema requirements.")),
+                    It.IsAny<Exception>(),
```
Aniruddh25, line 1844: Is this saying that there shouldn't be any log with exception?

Aniruddh25 (approved):
LGTM! Thanks for adding this validation and improving developer experience.

schemas/dab.draft.schema.json
sander1095: Good feedback, done!

sander1095:
@Aniruddh25 I've processed your comments.

I took a look at [the failures of the CI pipeline of the last commits](https://github.com/Azure/data-api-builder/pull/2646/checks?check_run_id=40416066239). They seem to be unrelated to my changes: `The task has failed because you are using Ubuntu 24.04 or later without mono installed. See https://aka.ms/nuget-task-mono for more information.`.

Let's hope that a re-run will make them go green!

Aniruddh25 (approved):

sander1095:
I see I broke some config tests. I'll hopefully have time to look into this next week!

sander1095:
Hi! I sadly haven't been able to work on this issue this week. I'll be moving house during the next few weeks. I probably won't have time to look until this until the end of May/beginning of June.

If the team wants this merged, I'll sadly have to ask them to fix these tests. Otherwise I will fix them when I have time.

@Aniruddh25 @RubenCerna2079

Aniruddh25:
@sander1095 thank you for your update. This PR can wait when you get more time to work on it. We appreciate your contributions and look forward to when you have updates on this.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -1686,6 +1816,36 @@ public async Task TestBasicConfigSchemaWithNoOptionalFieldsIsValid(string jsonDa
...
+        /// <summary>
+        /// This test method validates that the JSON schema validates that only known auth providers can be used.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(CONFIG_FILE_WITH_UNKNOWN_AUTHENTICATION_PROVIDER, DisplayName = "Validates schema of the config file when there is an unknown authentication provider.")]
+        [DataRow(CONFIG_FILE_WITH_MISSING_JWT_PROPERTY, DisplayName = "Validates schema of the config file a missing JWT property")]
+        [DataRow(CONFIG_FILE_WITH_MISSING_JWT_CHILD_PROPERTIES, DisplayName = "Validates schema of the config file with missing JWT child properties.")]
+        [DataRow(CONFIG_FILE_WITH_AUTHENTICATION_PROVIDER_THAT_SHOULD_NOT_HAVE_JWT, DisplayName = "Validates schema of the config file when an auth provider is chosen WITH a JWT property, "
+        + "even though the JWT property should not exist.")]
+        public async Task TestBasicConfigWithAuthProviders(string jsonData)
```
RubenCerna2079, line 1828: I would change the name of the test to something like `TestConfigWithInvalidAuthProviders` since it is only testing invalid config file options.

RubenCerna2079 (approved):
LGTM! I really like the way the test is made, very clever.

sander1095:
Hi @Aniruddh25 and @RubenCerna2079 !

I've found some to look into this PR. I've updated it. The failed unit tests now run correctly on my machine!

There's 1 impactful change you should be aware of: I decided to move away from `NJsonSchema` and use `JsonSchemaNet` instead. The former doesn't validate `oneof + const` correctly, which we use for letting the user decide a specific auth provider. `JsonSchemaNet` DOES support this. If you disagree with this change, we could stay on `NJsonSchema` and use `anyof + const` instead. This works, but will have a minor performance impact 😞.

I think it's worth a try to run the pipelines again and see the results!

If the integration fail again, please post the stacktraces. I can't access them as I get a 403 error when trying to see more details in Azure Pipelines.

src/Service.Tests/Configuration/ConfigurationTests.cs
sander1095: Done!

RubenCerna2079 (commented):
@sander1095 I left a few comments, besides that it LGTM, I will approve once the comments are resolved.

src/Core/Models/JsonSchemaValidationResult.cs
```diff
@@ -31,10 +29,8 @@ public JsonSchemaValidationResult(bool isValid, ICollection<ValidationError>? er
     /// error message, line number, and line position where the error occurred.
     /// </summary>
     /// <param name="validationErrors">list of schema validation errors</param>
-    private static string FormatSchemaValidationErrorMessage(ICollection<ValidationError> validationErrors)
+    private static string FormatSchemaValidationErrorMessage(IReadOnlyDictionary<string, string> validationErrors)
     {
-        return $"> Total schema validation errors: {validationErrors.Count}\n" +
-            string.Join("", validationErrors.Select(e => $"> {e} at " +
-            $"{e.LineNumber}:{e.LinePosition}\n\n"));
+        return $"> Total schema validation errors: {validationErrors.Count}";
```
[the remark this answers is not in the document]
sander1095: The new library for parsing JSON schema does not contain line number information for parse failures. That's why I removed it. I believe it's alright as tests will most likely produce other logs to indicate the problem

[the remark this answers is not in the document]
sander1095: Done

sander1095:
Hey @RubenCerna2079 ! Good news!

I just fixed the unit tests by switching from JsonSchema.Net to Newtonsoft.Json.Schema. This package fixes the issues I had with NJsonSchema AND it actually has functioning error reporting. 

This package seems to be a much better fit as the project already has a dependency on Newtonsoft.Json. 
It also allowed me to restore the LineNumber/LinePosition message.

sander1095:
I can now see the integration test failure reasons! Some of them fail because the package migration changed the JSON Schema result messages. I can update this soon.

Aniruddh25 (approved):
LGTM, thanks for pushing through this.

RubenCerna2079:
Hey @sander1095, I saw that the tests that were failing are not due to your PR, so I decided to ignore them in order to get your PR in. I will solve this issue in a separate PR. Thanks for your work!

RubenCerna2079 (approved):
LGTM! Thank you for these changes!

Aniruddh25 (approved):
Thank you for these contributions!

PR #2648, opened 2025-04-14
Fixing Pipelines due to Ubuntu 24.04 update removing Mono

### LINKED ISSUE #2649: [Bug]: Pipelines failing due to update from Ubuntu 24.04
### What happened?

Following error is seen while running pipelines on Github.

```
##[error]The task has failed because you are using Ubuntu 24.04 or later without mono installed. See https://aka.ms/nuget-task-mono for more information.
```

### Version

1.5

Azure SQL

_No response_

_No response_

### Relevant log output

```Text

```

### Code of Conduct

sezal98:
## Why make this change?
Resolves #2649

## What is this change?
The pipelines are failing due to udpate from ubutu 24.04. 
We are using the latest version in our yamls hence, this change needs to be handled. 
image

The error message we are seeing is due to the fact that default Ubuntu 24.04+ image on Microsoft-hosted agents does not include Mono, which is required for the NuGet client that powers NuGetCommand@2 in our pipeline. Microsoft recommends migrating to using the NuGetAuthenticate@1 task and .NET CLI commands instead of the older NuGetCommand@2 task, which relies on Mono.

Steps to Fix the Issue:
- Remove NuGetCommand@2 task: Since this task requires Mono (which is no longer available by default on Ubuntu 24.04+), we should remove it and replace it with .NET CLI commands for restoring the NuGet packages.
- Add .NET CLI restore command: Instead of using the NuGetCommand@2 task, use the DotNetCoreCLI@2 task to restore NuGet packages.
- Update NuGet authentication: Ensure that you're using NuGetAuthenticate@1 before the restore task to handle authentication.

## How was this tested?
The pipelines are running as expected.

### DIFF
--- a/.pipelines/cosmos-pipelines.yml
+++ b/.pipelines/cosmos-pipelines.yml
@@ -51,12 +51,12 @@ steps:
 - task: NuGetToolInstaller@1
 
 - task: DotNetCoreCLI@2
-    displayName: Restore NuGet packages
-    inputs:
-      command: restore
-      projects: '$(solution)'
-      feedsToUse: config
-      nugetConfigPath: Nuget.config
+  displayName: Restore NuGet packages
+  inputs:
+    command: restore
+    projects: '$(solution)'
+    feedsToUse: config
+    nugetConfigPath: Nuget.config
 
 - task: DotNetCoreCLI@2
   displayName: Build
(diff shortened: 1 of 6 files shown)


Aniruddh25 (approved):
LGTM, thanks for the quick fix!!

aaronburtle (approved):
Thanks for fixing this!

Issue #2603, opened 2025-02-25
Support cache in the 'add' and 'update' verbs in the CLI


aaronburtle:
This task tracks the work to complete adding support for cache to the CLI through the `add` and `update` verbs. For more information see: https://github.com/Azure/data-api-builder/issues/2253

PR #2655, opened 2025-04-16
Include cache options in the add and update verbs of CLI

aaronburtle:
## Why make this change?

closes https://github.com/Azure/data-api-builder/issues/2603

## What is this change?

Includes caching options with the `add` and `update` verbs for the CLI. These verbs are used to add or update new entities in a config file, and will now include support for the `cache` entry in those entities.

## How was this tested?

Current test suite.

## Sample Request(s)

You can test with the CLI by using the `add` or `update` verb along with the flags `--cache.enabled` and `--cache.ttl`, default values are false and 5.


src/Cli.Tests/UpdateEntityTests.cs
```diff
@@ -1093,7 +1093,9 @@ private static UpdateOptions GenerateBaseUpdateOptions(
             IEnumerable<string>? relationshipFields = null,
             IEnumerable<string>? map = null,
             IEnumerable<string>? restMethodsForStoredProcedure = null,
-            string? graphQLOperationForStoredProcedure = null
+            string? graphQLOperationForStoredProcedure = null,
+            string? cacheEnabled = null,
```
Aniruddh25, line 1097: We should add automated tests that test if the these cache option get updated or not using the flags in the update command.

src/Cli.Tests/AddEntityTests.cs
```diff
@@ -67,6 +69,8 @@ public Task AddNewEntityWhenEntitiesNotEmpty()
                 fieldsToExclude: new string[] { },
                 policyRequest: null,
                 policyDatabase: null,
+                cacheEnabled: null,
```
Aniruddh25, line 72: We should add automated tests that test if the these cache option get added or not using the flags in the add command.

src/Cli/Commands/EntityOptions.cs
```diff
@@ -76,5 +80,11 @@ public EntityOptions(
 
         [Option("policy-database", Required = false, HelpText = "Specify an OData style filter rule that will be injected in the query sent to the database.")]
         public string? PolicyDatabase { get; }
+
+        [Option("cache.enabled", Required = false, HelpText = "Specify if caching is enabled for Entity.")]
+        public string? CacheEnabled { get; }
+
+        [Option("cache.ttl", Required = false, HelpText = "Specify time to live for cache entries for Entity.")]
```
Aniruddh25, line 87: Mention what is the default value in the HelpText

src/Cli/Commands/EntityOptions.cs
```diff
@@ -76,5 +80,11 @@ public EntityOptions(
 
         [Option("policy-database", Required = false, HelpText = "Specify an OData style filter rule that will be injected in the query sent to the database.")]
         public string? PolicyDatabase { get; }
+
+        [Option("cache.enabled", Required = false, HelpText = "Specify if caching is enabled for Entity.")]
```
Aniruddh25, line 84: Mention what is the default value in help text

src/Cli/Commands/EntityOptions.cs
Aniruddh25: Also mention the units of ttl

src/Cli/Utils.cs
```diff
@@ -828,6 +828,25 @@ public static EntityGraphQLOptions ConstructGraphQLTypeDetails(string? graphQL,
...
 
+        public static EntityCacheOptions ConstructCacheOptions(string? cacheEnabled, string? cacheTtl)
+        {
+            EntityCacheOptions cacheOptions = new();
+            bool isEnabled = false;
+            int ttl = EntityCacheOptions.DEFAULT_TTL_SECONDS;
+
+            if (cacheEnabled is not null && !bool.TryParse(cacheEnabled, out isEnabled))
+            {
+                _logger.LogError("Invalid format for --cache.enabled. Accepted values are true/false.");
```
Aniruddh25, line 839: Do we have an enum for boolean options? what do we do for other boolean properties?

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForCosmos.verified.txt
```diff
@@ -226,7 +226,11 @@
               }
             ]
           }
-        ]
+        ],
+        Cache: {
+          Enabled: false,
```
Aniruddh25, line 231: None of these changes to the Snapshots are required. Since we should not create a new object if the user didnt specify explicit values to the cache options. 

CacheOptions object should be created only when they are specified by the user

Aniruddh25 (changes requested):
Need to not instantiate CacheOptions when not specified.

src/Cli/Utils.cs
aaronburtle: we parse the same way e.g.

     ```
           if (bool.TryParse(graphQL, out bool graphQLEnabled))
                {
                    graphQLType = graphQLType with { Enabled = graphQLEnabled };
                }
```

File: src/Cli.Tests/AddEntityTests.cs [hunk 7]
```diff
@@ -40,6 +40,8 @@ public Task AddNewEntityWhenEntitiesEmpty()
...
+                cacheTtl: null,
                 config: TEST_RUNTIME_CONFIG_FILE,
                 restMethodsForStoredProcedure: null,
```
Aniruddh25, line 46: PR title should be changed from 

> Include the Add and Update verbs for Cache Options in the CLI

to 

> Include cache options in the `add` and `update` verbs of CLI

src/Cli/Utils.cs
```diff
@@ -828,6 +828,49 @@ public static EntityGraphQLOptions ConstructGraphQLTypeDetails(string? graphQL,
...
+            }
+
+            if (cacheTtl is not null && !int.TryParse(cacheTtl, out ttl))
+            {
+                _logger.LogError("Invalid format for --cache.ttl. Accepted values are any integer.");
```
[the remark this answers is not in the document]
Aniruddh25: >=0 makes more sense, since seconds cannot be -ve

src/Cli.Tests/AddEntityTests.cs
```diff
@@ -40,6 +40,8 @@ public Task AddNewEntityWhenEntitiesEmpty()
                 fieldsToExclude: new string[] { },
                 policyRequest: null,
                 policyDatabase: null,
+                cacheEnabled: null,
+                cacheTtl: null,
```
sezal98, line 44: Instead of adding this in every test. Can we take by default value as null so we wont have to add everywhere?

src/Cli/Utils.cs
```diff
@@ -828,6 +828,49 @@ public static EntityGraphQLOptions ConstructGraphQLTypeDetails(string? graphQL,
             return graphQLType;
         }
 
+        /// <summary>
+        /// Constructs the EntityCacheOption for Add/Update.
+        /// </summary>
+        /// <param name="cacheEnabled">Bool that flags if the cache is enabled.</param>
```
sezal98, line 834: `String value which defines if cache is enabled`

src/Cli/Utils.cs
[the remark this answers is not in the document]
sezal98: If cache ttl is zero does it mean that cache is disabled?

src/Cli.Tests/AddEntityTests.cs
aaronburtle: maybe we should have originally, but now all of the options arguments are like this, so I want to keep consistency.

Aniruddh25 (approved):

RubenCerna2079 (approved):
LGTM!

PR #2667, opened 2025-04-28
[Health]: Test Case for Rest and GraphQL Requests to DAB during health check

### LINKED ISSUE #2662: [HealthEndpoint] Test cases for GraphQL and REST in Health Check Report
The testing for Health Endpoint is failing for GraphQL and Rest calls with the error `One or more errors occurred. (No connection could be made because the target machine actively refused it. (localhost:5000)`. This is due to an in memory server running and not an http server for internal calls to work on http://localhost endpoint. 
We need to modify the HttpUtilities to pass the HttpClient from the Startup.cs and use the client from the factory instead of creating new. 

But if /health internally calls /api and /graphql, and you're using the same in-memory TestServer, you might hit issues like:

- Loopbacks not working: Since the same server is handling the request and the internal call, it could cause reentrancy issues or socket errors.
- Your test app is running in-memory (via TestServer or WebApplicationFactory), and not actually listening on a real network port.

Boot up your app in-memory for integration tests.
Run your app in-memory via test code, likely using WebApplicationFactory, TestServer, or similar.
That means: Your app should be bound to an external port like [http://localhost:80](http://localhost/)
Instead of in-memory (via the HttpClient returned from factory.CreateClient())
You need to inject the HttpClient from the test context into your health check logic, so it uses the in-memory test server instead of trying to call a real external address.

sezal98:
## Why make this change?
Resolves #2662 

## What is this change?
1. Created the HttpUtilities with HttpClient already instantiated and basic details filled from Startup.cs
2. Using this httpClient, call the REST and GraphQL requests with relative URI 
3. Created a function in DatabaseObject to access mapping dict as it is an abstract class which was not allowed to be mocked. 
4. Test cases for both REST and GraphQL
5. Rest Test case: Mock the client to return a 200 response when we receive a GET request with specific URI. Validate the error message from HttpUtilities should be null.
6. GraphQL Test case: Mock the client to return a 200 response when we get a POST request with /graphql URI and content payload. Mock the Dabase Object with columns in the entity map to only return the db object when a certain entity is called for. 

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

## Sample Request(s)
Requests are running as expected. Test cases added for Rest and Graphql

### DIFF
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -3725,7 +3725,6 @@ type Planet @model(name:""PlanetAlias"") {
         /// <summary>
         /// Tests different log level filters that are valid and check that they are deserialized correctly
         /// </summary>
-        [Ignore]
         [DataTestMethod]
         [TestCategory(TestCategory.MSSQL)]
         [DataRow(LogLevel.Trace, typeof(RuntimeConfigValidator))]
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -3725,6 +3725,7 @@ type Planet @model(name:""PlanetAlias"") {
         /// <summary>
         /// Tests different log level filters that are valid and check that they are deserialized correctly
         /// </summary>
+        [Ignore]
         [DataTestMethod]
         [TestCategory(TestCategory.MSSQL)]
         [DataRow(LogLevel.Trace, typeof(RuntimeConfigValidator))]
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -4054,7 +4054,7 @@ type Planet @model(name:""PlanetAlias"") {
                 { "Book", requiredEntity }
             };
 
-            CreateCustomConfigFile(entityMap, enableGlobalRest, enableGlobalGraphql, enableGlobalHealth, enableDatasourceHealth);
+            CreateCustomConfigFile(entityMap, enableGlobalRest, enableGlobalGraphql, enableGlobalHealth, enableDatasourceHealth, HostMode.Development);
 
             string[] args = new[]
             {
@@ -4667,14 +4667,14 @@ type Planet @model(name:""PlanetAlias"") {
         /// </summary>
         /// <param name="entityMap">Collection of entityName -> Entity object.</param>
         /// <param name="enableGlobalRest">flag to enable or disabled REST globally.</param>
-        private static void CreateCustomConfigFile(Dictionary<string, Entity> entityMap, bool enableGlobalRest = true, bool enableGlobalGraphql = true, bool enableGlobalHealth = true, bool enableDatasourceHealth = true)
+        private static void CreateCustomConfigFile(Dictionary<string, Entity> entityMap, bool enableGlobalRest = true, bool enableGlobalGraphql = true, bool enableGlobalHealth = true, bool enableDatasourceHealth = true, HostMode hostMode = HostMode.Production)
         {
             DataSource dataSource = new(
                 DatabaseType.MSSQL,
                 GetConnectionStringFromEnvironmentConfig(environment: TestCategory.MSSQL),
                 Options: null,
                 Health: new(enableDatasourceHealth));
-            HostOptions hostOptions = new(Cors: null, Authentication: new() { Provider = nameof(EasyAuthType.StaticWebApps) });
+            HostOptions hostOptions = new(Mode: hostMode, Cors: null, Authentication: new() { Provider = nameof(EasyAuthType.StaticWebApps) });
 
             RuntimeConfig runtimeConfig = new(
                 Schema: string.Empty,
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -20,7 +20,6 @@ using System.Threading.Tasks;
 using System.Web;
 using Azure.DataApiBuilder.Auth;
 using Azure.DataApiBuilder.Config;
-using Azure.DataApiBuilder.Config.HealthCheck;
 using Azure.DataApiBuilder.Config.ObjectModel;
 using Azure.DataApiBuilder.Core;
 using Azure.DataApiBuilder.Core.AuthenticationHelpers;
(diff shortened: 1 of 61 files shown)


src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -166,36 +139,33 @@ public void ConfigureApiRoute(HttpContext httpContext)
                 // In case any primitive column names are present, execute the query
                 if (columnNames.Any())
                 {
-                    using (HttpClient client = CreateClient(apiRoute))
+                    string jsonPayload = Utilities.CreateHttpGraphQLQuery(graphqlObjectName, columnNames, entity.EntityFirst);
+                    HttpContent content = new StringContent(jsonPayload, Encoding.UTF8, Utilities.JSON_CONTENT_TYPE);
+
+                    HttpRequestMessage message = new(method: HttpMethod.Post, requestUri: graphqlUriSuffix)
```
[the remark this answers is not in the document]
sezal98: We want to create a relative URIs since base address is already set in the startup.cs

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -208,23 +178,6 @@ public void ConfigureApiRoute(HttpContext httpContext)
...
-                DefaultRequestHeaders =
-                {
-                    Accept = { new MediaTypeWithQualityHeaderValue(Utilities.JSON_CONTENT_TYPE) }
-                },
-                Timeout = TimeSpan.FromSeconds(200),
-            };
-        }
-
         // Updates the entity key name to camel case for the health check report.
         public static string LowerFirstLetter(string input)
```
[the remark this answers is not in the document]
sezal98: `string graphqlObjectName = LowerFirstLetter(entity.GraphQL.Plural.Pascalize());` I did account for two words like Lastname which gets changed to `lastnames`. The function will also generate this which is what is required in GraphQL
image
However since we already have this function, modified to use this.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4667,22 +4454,20 @@ public async Task TestNoDepthLimitOnGrahQLInNonHostedMode(int? depthLimit)
         /// </summary>
         /// <param name="entityMap">Collection of entityName -> Entity object.</param>
         /// <param name="enableGlobalRest">flag to enable or disabled REST globally.</param>
-        private static void CreateCustomConfigFile(Dictionary<string, Entity> entityMap, bool enableGlobalRest = true, bool enableGlobalGraphql = true, bool enableGlobalHealth = true, bool enableDatasourceHealth = true, HostMode hostMode = HostMode.Production)
+        private static void CreateCustomConfigFile(Dictionary<string, Entity> entityMap, bool enableGlobalRest = true)
```
Aniruddh25, line 4457: Is `enableGlobalGraphQL` not used by any of the existing tests?

src/Service/Startup.cs
```diff
@@ -270,6 +271,27 @@ public void ConfigureServices(IServiceCollection services)
             //Enable accessing HttpContext in RestService to get ClaimsPrincipal.
             services.AddHttpContextAccessor();
 
+            services.AddHttpClient("ContextConfiguredHealthCheckClient")
+                .ConfigureHttpClient((serviceProvider, client) =>
+                {
```
Aniruddh25, line 276: does adding HttpClient to the service provider help with testing?

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -0,0 +1,437 @@
...
+            // Act
+            string errorMessageFromGraphQL = httpUtilities.ExecuteGraphQLQuery(
+                graphqlUriSuffix: "/graphql",
+                entityName: runtimeConfig.Entities.First().Key,
+                entity: runtimeConfig.Entities.First().Value,
+                incomingRoleHeader: string.Empty,
+                incomingRoleToken: string.Empty);
+
+            // Assert
+            Assert.IsNull(errorMessageFromGraphQL);
```
Aniruddh25, line 233: We should have a test for the failure case as well. Mock a failed response, and verify the httputilities returns the error

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4017,219 +4016,7 @@ public async Task HealthEndpoint_ValidateContents()
...
-                Assert.AreEqual(
-                    expected: ProductInfo.GetDataApiBuilderUserAgent(),
-                    actual: appNameValue.ToString(),
-                    message: "Unexpected or missing DAB user agent string.");
-            }
-            else
-            {
-                Assert.Fail();
-            }
+            HealthEndpointTests.ValidateBasicDetailsHealthCheckResponse(responseProperties);
```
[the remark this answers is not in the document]
sezal98: I dont see any missing tab in this line. Doesnt show up in formatting as well.

src/Service.Tests/Configuration/ConfigurationTests.cs
sezal98: GraphQL was always enabled by default. Rest was switched on and off in multiple tests.

src/Service/Startup.cs
sezal98: yes we can create a mock httpclient now and pass it to the Httputilities. Before this I was creating httpclient internally in this file and hence mocking was not possible.

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -0,0 +1,437 @@
...
+            HttpUtilities httpUtilities = new(
+                logger.Object,
+                metadataProviderFactory.Object,
+                provider,
+                mockHttpClientFactory.Object);
+
+            // Act
+            string errorMessageFromGraphQL = httpUtilities.ExecuteGraphQLQuery(
+                graphqlUriSuffix: "/graphql",
+                entityName: runtimeConfig.Entities.First().Key,
```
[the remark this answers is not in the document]
sezal98: yes it would have

src/Service.Tests/Configuration/HealthEndpointTests.cs
sezal98: Added negative test cases

Aniruddh25 (approved):
LGTM, thanks for adding these tests!

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -249,7 +248,7 @@ private void PopulateEntityHealth(ComprehensiveHealthCheckReport ComprehensiveHe
                 {
                     ComprehensiveHealthCheckReport.Checks ??= new List<HealthCheckResultEntry>();
 
-                    (int, string?) response = ExecuteGraphQLEntityQuery(runtimeConfig.GraphQLPath, entityValue, entityKeyName);
+                    (int, string?) response = await ExecuteGraphQLEntityQueryAsync(runtimeConfig.GraphQLPath, entityValue, entityKeyName);
```
RubenCerna2079, line 251: Why only change the name for this one by adding `Async` and not the rest of the functions?

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -88,50 +74,44 @@ public void ConfigureApiRoute(HttpContext httpContext)
...
-                string apiRoute = Utilities.GetServiceRoute(_apiRoute, restUriSuffix);
+                string apiRoute = $"{restUriSuffix}{Utilities.CreateHttpRestQuery(entityName, first)}";
                 if (string.IsNullOrEmpty(apiRoute))
                 {
                     LogTrace("The API route is not available, hence HealthEndpoint is not available.");
                     return errorMessage;
                 }
 
-                if (!Program.CheckSanityOfUrl(apiRoute))
+                if (!Program.CheckSanityOfUrl($"{_httpClient.BaseAddress}{restUriSuffix}"))
```
RubenCerna2079, line 90: Why was the variable `apiRoute` changed?

src/Service/HealthCheck/HttpUtilities.cs
```diff
@@ -145,18 +125,11 @@ public void ConfigureApiRoute(HttpContext httpContext)
 
         // Executes the GraphQL query by sending a POST request to the API.
         // Internally calls the metadata provider to fetch the column names to create the graphql payload.
-        public string? ExecuteGraphQLQuery(string graphqlUriSuffix, string entityName, Entity entity, string incomingRoleHeader, string incomingRoleToken)
+        public async Task<string?> ExecuteGraphQLQuery(string graphqlUriSuffix, string entityName, Entity entity, string incomingRoleHeader, string incomingRoleToken)
         {
             string? errorMessage = null;
-            // Base URL of the API that handles SQL operations
-            string apiRoute = Utilities.GetServiceRoute(_apiRoute, graphqlUriSuffix);
-            if (string.IsNullOrEmpty(apiRoute))
```
RubenCerna2079, line 153: Why was this check removed while in the RestQuery it is still there?

src/Service/Startup.cs
```diff
@@ -270,6 +271,27 @@ public void ConfigureServices(IServiceCollection services)
...
+                        string baseUri = $"{httpContext.Request.Scheme}://{httpContext.Request.Host}";
+                        client.BaseAddress = new Uri(baseUri);
+
+                        // Set default Accept header
+                        client.DefaultRequestHeaders.Accept.Clear();
+                        client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+                    }
+
+                    // Optional: Set a timeout
+                    client.Timeout = TimeSpan.FromSeconds(200);
```
RubenCerna2079, line 292: How is this optional and why would we want it optional?

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -0,0 +1,492 @@
...
+                    if (enableEntityHealth)
+                    {
+                        ValidateEntityRestAndGraphQLResponse(responseProperties, enableEntityRest, enableEntityGraphQL, enableGlobalRest, enableGlobalGraphql);
+                    }
+                }
+            }
+        }
+
+        [TestMethod]
+        public async Task TestHealthCheckRestResponseAsync()
```
RubenCerna2079, line 105: I think you should add a comment explaining what this test does

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -0,0 +1,492 @@
...
+                incomingRoleToken: string.Empty
+            );
+
+            // Assert
+            // Validate the null response from the REST API call 
+            Assert.IsNull(errorMessageFromRest);
+        }
+
+        [TestMethod]
+        public async Task TestFailureHealthCheckRestResponseAsync()
```
RubenCerna2079, line 129: I think there should also be a comment here

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -0,0 +1,492 @@
...
+                incomingRoleHeader: string.Empty,
+                incomingRoleToken: string.Empty
+            );
+
+            // Assert
+            Assert.IsNotNull(errorMessageFromRest);
+        }
+
+        [TestMethod]
+        public async Task TestHealthCheckGraphQLResponseAsync()
```
RubenCerna2079, line 152: At the very least the test methods should have a comment summarizing what they do.

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -0,0 +1,492 @@
...
+                entityName: runtimeConfig.Entities.First().Key,
+                entity: runtimeConfig.Entities.First().Value,
+                incomingRoleHeader: string.Empty,
+                incomingRoleToken: string.Empty);
+
+            // Assert
+            Assert.IsNotNull(errorMessageFromGraphQL);
+        }
+
+        private static HttpUtilities SetupRestTest(RuntimeConfig runtimeConfig, HttpStatusCode httpStatusCode = HttpStatusCode.OK)
```
RubenCerna2079, line 189: I think it would be a good idea to create a region for the helper methods, so it is easier to identify them

src/Service/HealthCheck/HttpUtilities.cs
sezal98: Not needed anymore

src/Service/HealthCheck/HttpUtilities.cs
sezal98: In rest we create the entire query in the URI itself. However, in GraphQL we take the basic query of /graphql and add parameters in payload. 
`CheckSanityOfUrl` already validates the URi of null or empty so we dont need to do again.

src/Service/Startup.cs
sezal98: If you don’t set client.Timeout manually, the HttpClient will use the default timeout, which is:
`TimeSpan.FromSeconds(100)`
This means every HTTP request made by this client will wait a maximum of 100 seconds for a response before throwing a TaskCanceledException. Hence, I am setting 200 here. We can change if we want.

Aniruddh25 (approved):

RubenCerna2079 (approved):
LGTM

PR #2673, opened 2025-05-02
Enhance GraphQL OTEL instrumentation with custom metrics and traces

### LINKED ISSUE #2642: OTEL Implementation for GraphQL
This task is related to enchancement #2554 to create the custom DAB OTEL traces & counters. 
We are missing the OTEL implementation for GraphQL, which should be similarly done to PR #2617 that does the OTEL implementation for REST.

RubenCerna2079:
## Why make this change?
This closes #2642

## What is this change?
This PR enhances the OTEL instrumentation for the GraphQL APIs by adding custom traces and metrics.
Metrics can be filtered for `status_code`, `api_type`, `endpoint` and `method`.

## How was this tested?
- [X] Local Testing
- [ ] Integration Tests
- [ ] Unit Tests

All of the tests were done locally to check if the log information that was provided was correct, for both scenarios in which the query gave the proper information or when an exception was raised. 
Another thing that was tested is that when we open GraphQL it would send a few requests called `Introspection Queries` used to ensure that GraphQL is working properly. However, we do not want the user to see these requests as part of the total count as this is done automatically, which may confuse the users.

## Sample Request(s)
image
image
image

## Instructions on how to use DAB Workbench
- Clone the following repo `https://github.com/tommasodotNET/dab-workbench.git`
- Run your DAB version in CLI so the files from the `out` folder are created, and make sure to stop it before running the DAB Workbench since both cannot be running at the same time.
- Find the path to the `Microsoft.DataApiBuilder.exe`, which should look something like `<PATH_TO_REPO>\data-api-builder\src\out\cli\net8.0\Microsoft.DataApiBuilder.exe`
- Copy the path of the `.exe` file and paste it in the file `/DABWorkbench.AppHost/Program.cs` in the variable `dabCLIPath` which is found in line 3 as follows:
`var dabCLIPath = @"<PATH_TO_REPO>\data-api-builder\src\out\cli\net8.0\Microsoft.DataApiBuilder.exe";`
- Now you should be able to run DAB Workbench with your version of DAB.


src/Core/Services/BuildRequestStateMiddleware.cs
```diff
@@ -26,14 +35,92 @@ public BuildRequestStateMiddleware(RequestDelegate next)
...
+                    else if (context.Result.ContextData.ContainsKey(WellKnownContextData.OperationNotAllowed))
+                    {
+                        statusCode = HttpStatusCode.MethodNotAllowed;
+                    }
+                    else
+                    {
+                        statusCode = HttpStatusCode.InternalServerError;
+                    }
+
+                    string errorMessage = context.Result.Errors![0].Message;
```
Aniruddh25, line 94: We should verify Errors is not null here before dereferencing it

src/Core/Services/BuildRequestStateMiddleware.cs
```diff
@@ -26,14 +35,92 @@ public BuildRequestStateMiddleware(RequestDelegate next)
...
+                    Exception ex = new(errorMessage);
+
+                    // Activity will track error
+                    activity?.TrackRestControllerActivityFinishedWithException(ex, statusCode);
```
Aniruddh25, line 98: nit: the name `TrackRestControllerActivityFinishedWithException` is misleading since this is now being used for GraphQL too.. Consider using a generic TrackController.. name

src/Core/Services/ExecutionHelper.cs
```diff
@@ -91,6 +96,8 @@ public async ValueTask ExecuteQueryAsync(IMiddlewareContext context)
         /// </param>
         public async ValueTask ExecuteMutateAsync(IMiddlewareContext context)
         {
+            using Activity? activity = StartQueryActivity(context);
```
Aniruddh25, line 99: When is the end of this activity marked?

src/Core/Services/BuildRequestStateMiddleware.cs
```diff
@@ -26,14 +35,92 @@ public BuildRequestStateMiddleware(RequestDelegate next)
...
+
+        try
         {
-            // Because Request.Headers is a NameValueCollection type, key not found will return StringValues.Empty and not an exception.
-            StringValues clientRoleHeader = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER];
-            context.ContextData.TryAdd(key: AuthorizationResolver.CLIENT_ROLE_HEADER, value: clientRoleHeader);
+            // We want to ignore introspection queries DAB uses to check access to GraphQL since they are not sent by the user.
+            if (!isIntrospectionQuery)
+            {
+                TelemetryMetricsHelper.IncrementActiveRequests(apiType);
```
Aniruddh25, line 53: How did you test this? Could you please add it to the description? 
Can we add some integration tests as well?

sourabh1007:
You should follow open telemetry semantic conventions for otel metrics and trace. 
For HTTP related information, this should be the semantic convention: [https://opentelemetry.io/docs/specs/semconv/http/http-spans/](https://opentelemetry.io/docs/specs/semconv/http/http-spans/)

There are few HTTP metrics already available, https://opentelemetry.io/docs/specs/semconv/http/http-metrics/.

Similar exercise, we did for Cosmos DB SDK. ref. https://opentelemetry.io/docs/specs/semconv/database/cosmosdb/. Right now, it is under preview.

@Aniruddh25 what do you think about it?

src/Core/Services/ExecutionHelper.cs
RubenCerna2079: Based on Tomasso's PR, I thought that we did not have to explicitly end the activity. Or am I wrong about that?

RubenCerna2079:
@sourabh1007 I will be addressing the semantic conventions for OTEL in a future PR. 
Thank you for letting us know about this!

src/Core/Services/BuildRequestStateMiddleware.cs
RubenCerna2079: It was done as part of the local testing.

RubenCerna2079: As mentioned offline, I added instructions on how to use DAB Workbench in order for others to test the changes done in this PR.

src/Core/Telemetry/TelemetryTracesHelper.cs
```diff
@@ -77,11 +77,11 @@ public static void TrackQueryActivityStarted(
         }
 
         /// <summary>
-        /// Tracks the completion of a REST controller activity.
+        /// Tracks the completion of a controller activity without any exceptions.
         /// </summary>
         /// <param name="activity">The activity instance.</param>
         /// <param name="statusCode">The HTTP status code of the response.</param>
-        public static void TrackRestControllerActivityFinished(
+        public static void TrackControllerActivityFinished(
```
Aniruddh25, line 84: Does `TrackControllerActivityFinished` mark the finish of `TrackMainControllerActivityStarted` ? If yes, consider naming both start and finish symmetrically for easier interpretation.. 

so simply this `TrackControllerActivityStarted`  or add `Main` in the finished function

Aniruddh25:
> * Copy the path of the `.exe` file and past it in the file `/DABWorkbench.AppHost/Program.cs` in the variable `dabCLIPath` which is found in line 3 as follows:

agree, the semantic conventions make sense to follow a standard. Like Ruben mentioned, we will adopt the naming conventions in a subsequent enhancement PR.

src/Core/Services/ExecutionHelper.cs
Aniruddh25: With the `using` keyword, local variable `activity` will go out of scope and probably thats why we dont need to explicitly end.

Aniruddh25 (approved):

RubenCerna2079:
> From description:
> 
> > Run your DAB version in CLI and make sure to stop it afterwards
> 
> why do we need to run DAB version in CLI and immediately stop it? Did you mean to say install DAB?

I reworded the sentence so it makes more sense, we cannot have DAB and the workbench running at the same time, which is why we first need to run DAB with the CLI so that it outputs the file that will be used in the workbench.

src/Core/Services/BuildRequestStateMiddleware.cs
```diff
@@ -26,14 +35,97 @@ public BuildRequestStateMiddleware(RequestDelegate next)
...
-            value is HttpContext httpContext)
+        bool isIntrospectionQuery = context.Request.OperationName == "IntrospectionQuery";
+        ApiType apiType = ApiType.GraphQL;
+        Kestral method = Kestral.Post;
+        string route = _runtimeConfigProvider.GetConfig().GraphQLPath.Trim('/');
+        DefaultHttpContext httpContext = (DefaultHttpContext)context.ContextData.First(x => x.Key == "HttpContext").Value!;
+        Stopwatch stopwatch = Stopwatch.StartNew();
+
+        using Activity? activity = !isIntrospectionQuery ?
+            TelemetryTracesHelper.DABActivitySource.StartActivity($"{method} /{route}") : null;
```
sezal98, line 46: There is a space in between this. Is that expected?

src/Core/Services/BuildRequestStateMiddleware.cs
```diff
@@ -26,14 +35,97 @@ public BuildRequestStateMiddleware(RequestDelegate next)
...
+                TelemetryMetricsHelper.IncrementActiveRequests(apiType);
+                if (activity is not null)
+                {
+                    activity.TrackMainControllerActivityStarted(
+                        httpMethod: method,
+                        userAgent: httpContext.Request.Headers["User-Agent"].ToString(),
+                        actionType: (context.Request.Query!.ToString().Contains("mutation") ? OperationType.Mutation : OperationType.Query).ToString(),
+                        httpURL: string.Empty, // GraphQL has no route
+                        queryString: null, // GraphQL has no query-string
+                        userRole: httpContext.Request.Headers["X-MS-API-ROLE"].FirstOrDefault() ?? httpContext.User.FindFirst("role")?.Value,
```
sezal98, line 62: Use a constant: CLIENT_ROLE_HEADER

src/Core/Services/ExecutionHelper.cs
```diff
@@ -127,6 +134,31 @@ public async ValueTask ExecuteMutateAsync(IMiddlewareContext context)
             }
         }
 
+        /// <summary>
+        /// Starts the activity for the query
+        /// </summary>
+        /// <param name="context">
```
sezal98, line 140: nit: could be in one line

sezal98:
Can we add any UTs for any scenarios where introspection query is received an telemetry is not seen?

src/Core/Services/BuildRequestStateMiddleware.cs
RubenCerna2079: I think it is just the formatting in github, check it in codespace and see there is no space between.

src/Core/Services/ExecutionHelper.cs
RubenCerna2079: The other comments follow the same format this one has, so I would like to keep it consistent

sezal98 (approved):

PR #2681, opened 2025-05-12
Handle unexpected exception due to column type change in datasource

vadeveka:
## Why make this change?
When an API owner starts up DAB, the GraphQL data types for the fields are automatically derived from the SQL column types. However, if the API owner later changes the column type without restarting DAB, the requests will fail at runtime due to failure with parsing the JSON result in the leaf node resolver.
Due to the unhandled exception, this will result in an error message `One of the identified items was in an invalid format` or `"The requested operation requires an element of type 'Number', but the target element has type 'String'.` 

## What is this change?
1. Wrap the field value parsing to catch specific exception types and throw a clearer exception
2. Ensure location and path in errors are retained for DataApiBuilderException for 4xx status codes

## How was this tested?

- [x] Manual test
- [] Integration Tests
- [] Unit Tests

## Sample Request(s)

Step 1: Create sample table in datasource
```
-- Create the table
CREATE TABLE Employees (
    EmployeeID INT PRIMARY KEY,
    Age INT NOT NULL,
    Salary INT NOT NULL
);

-- Insert sample data
INSERT INTO Employees (EmployeeID, Age, Salary)
VALUES 
    (1, 25, 50000),
    (2, 30, 60000),
    (3, 35, 70000),
    (4, 40, 80000),
    (5, 45, 90000);
```

Step 2:
Configure dab config with following entity
```
    "Employees": {
      "source": {
        "object": "dbo.Employees",
        "type": "table"
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            {
              "action": "*"
            }
          ]
        }
      ]
    }, 
```

Verify employees query runs fine with Salary field
```
{
    employees {
        items {
            EmployeeID
            Salary
        }
    }
}
```

Step 3: Update Salary column type
```
ALTER TABLE Employees
ALTER COLUMN Salary NVARCHAR(100);
```

Run employees query again, response returned
```
{
  "errors": [
    {
      "message": "The value could not be parsed for configured GraphQL data type Int",
      "locations": [
        {
          "line": 5,
          "column": 13
        }
      ],
      "path": [
        "employees",
        "items",
        0,
        "Salary"
      ],
      "extensions": {
        "code": "GraphQLMapping"
      }
    }, 
    ...
  ]
}
```


src/Config/Utilities/HttpStatusCodeExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Azure.DataApiBuilder.Config.Utilities;
+
+public static class HttpStatusCodeExtensions
+{
+    /// <summary>
+    /// Check for status code within 4xx range
+    /// </summary>
+    /// <param name="statusCode"></param>
+    /// <returns></returns>
+    public static bool IsClientError(this HttpStatusCode statusCode)
```
Alekhya-Polavarapu, line 15: Am i missing or is this not being used?

Alekhya-Polavarapu (approved):

src/Config/Utilities/HttpStatusCodeExtensions.cs
vadeveka: used in Startup

rusamant (approved):

Aniruddh25 (approved):
LGTM, thanks for this contribution.

Issue #2690, opened 2025-05-16
[Bug]: DAB connect connect to Azure SQL server


vishms:
### What happened?

We have deployed data-api-builder as a pod in Azure kubernetes service (AKS). DAB fails to start because it is unable to connect to the Azure SQL server.
SQL server's networking: Public access is blocked. apps connect to the SQL server through private endpoint. 
Connection mode is redirect. 

If I change the connection mode to 'proxy' then DAB connects to the database fine but it fails to connect when the SQL's connection mode is redirect. 
Other applications in the same AKS and same node are connecting to the same SQL server database without any issue. 

This has been raised with Azure database support. Investigation involved collaboration with SQL server, networking and AKS teams and couldn't figure out the root cause. 

I have been suggested by the Azure support team to raise it as a bug in DAB for further investigation

connection string that DAB is using: 
Server=tcp:******,1433;Initial Catalog=*****;Persist Security Info=False;User ID=****;Password=****;MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;

### Version

1.4.35, 1.2.14 and  1.5.29-rc

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

REST

### Relevant log output

```Text
info: Azure.DataApiBuilder.Core.Services.ISqlMetadataProvider[0]
      [ms_groups] REST path: /reporting-api/ms_groups
fail: Azure.DataApiBuilder.Service.Startup[0]
      Unable to complete runtime initialization. Refer to exception for error details.
      Azure.DataApiBuilder.Service.Exceptions.DataApiBuilderException: Cannot obtain Schema for entity field_definitions with underlying database object source: dbo.field_definitions due to: A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: TCP Provider, error: 26 - Error Locating Server/Instance Specified)
         at Azure.DataApiBuilder.Core.Services.SqlMetadataProvider`3.HandleOrRecordException(Exception e) in /_/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs:line 100
[...]

PR #2702, opened 2025-05-28
Upgrade Microsoft.Data.SqlClient to New Version

RubenCerna2079:
## Why make this change?

This change solves the bug #2690

## What is this change?

We update the `Microsoft.Data.SqlClient` package from version 5.2.0 to version 5.2.3 in the `Directory.Packages.props` file. As the bug is a known error from the current version we are using

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -33,7 +33,7 @@
         <PackageVersion Include="OpenTelemetry.Instrumentation.Http" Version="1.9.0" />
         <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.9.0" />
         <!--When updating Microsoft.Data.SqlClient, update license URL in scripts/notice-generation.ps1-->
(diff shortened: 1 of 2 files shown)


aaronburtle (approved):
Thanks for fixing!

src/Directory.Packages.props
```diff
@@ -33,7 +33,7 @@
         <PackageVersion Include="OpenTelemetry.Instrumentation.Http" Version="1.9.0" />
         <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.9.0" />
         <!--When updating Microsoft.Data.SqlClient, update license URL in scripts/notice-generation.ps1-->
```
Aniruddh25, line 35: This comment says to update notice-generation script as well. Can you please fix that as well in the same PR?

Aniruddh25 (changes requested):
Needs fix to notice-generation

Aniruddh25 (approved):
LGTM, Thanks for the quick fix!

Issue #2533, opened 2025-01-23
HealthEndpoint: Parallel Queries to DB


sezal98:
Paralellisation of Health check calls: runtime.health.maxDop
Based on the number in the config of Max-Dop we should run parellel threads which run the quries to DB simultaneously and reduce the time it takes to return /health endpoint

JerryNixon:
## Maximum Degree of Parallelism
Values in the configuration.
| `MaxDop`        | Value |
|------------------|-------|
| Lower limit      | 1     |
| Default value    | 4     |
| Upper limit      | 8     |

PR #2714, opened 2025-06-09
Optimize health check response using parallel queries

souvikghosh04:
## Why make this change?

- #2533 
  - Improve health check endpoint performance by running queries in parallel.
  - This allows the health check endpoint to load faster

## What is this change?

- Provide option to specify max degree of query parallelism as a config option
- The default value is 4 and max allowed value is 8. System automatically adjusts itself within this range in case a value specified is lower or higher
- .Net framework's Parallel tasks are used to run queries in parallel

### Usage

- `max-query-parallelism` can be set in the config, e.g.: `"runtime": {
	"health": {
	  "max-query-parallelism": 8
	}
}`
- User can set any value between 1 and 8, inclusive or the system automatically adjust in case of Null value or out of range value

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)

- GET http://localhost:5000/health

### DIFF
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -412,6 +412,11 @@
               "type": "integer",
               "description": "Time to live in seconds for the Comprehensive Health Check Report cache entry.",
               "default": 5
+            },
+            "max-query-parallelism": {
+              "type": "integer",
+              "description": "The max degree of parallelism for running parallel health check queries.",
+              "default": 4
             }
           }
         }
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft-07/schema",
+  "$schema": "http://json-schema.org/draft-07/schema",
   "$id": "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json",
   "title": "Data API builder",
   "description": "Schema for Data API builder engine",
@@ -277,8 +277,33 @@
               "additionalProperties": false,
               "properties": {
                 "provider": {
-                  "type": "string",
                   "description": "The name of authentication provider",
+                  "oneOf": [
+                    {
+                      "const": "StaticWebApps",
+                      "description": "Authentication provided by Azure Static Web Apps."
+                    },
+                    {
+                      "const": "EntraID",
+                      "description": "Authentication provided by Microsoft Entra ID (formerly Azure AD). Use the JWT property to configure this provider."
+                    },
+                    {
+                      "const": "Simulator",
+                      "description": "Simulated authentication for development and testing purposes."
+                    },
+                    {
+                      "const": "AppService",
+                      "description": "Authentication provided by Azure App Service."
+                    },
+                    {
+                      "const": "AzureAD",
+                      "description": "Synonymous with the EntraID value. Use the JWT property to configure this provider."
+                    },
+                    {
+                      "const": "Custom",
+                      "description": "Custom authentication provider defined by the user. Use the JWT property to configure the custom provider."
+                    }
+                  ],
                   "default": "StaticWebApps"
                 },
                 "jwt": {
@@ -291,9 +316,29 @@
                     "issuer": {
                       "type": "string"
                     }
-                  }
+                  },
+                  "required": ["audience", "issuer"]
                 }
-              }
+              },
+              "allOf": [
+                {
+                  "$comment": "We want the user to provide the JWT property when the provider requires it, and omit JWT when the provider does not require it.",
+                  "if": {
+                    "properties": {
+                      "provider": {
+                        "anyOf": [
+                          { "const": "EntraID" },
+                          { "const": "AzureAD" },
+                          { "const": "Custom" }
+                        ]
+                      }
+                    },
+                    "required": ["provider"]
+                  },
+                  "then": { "required": ["jwt"] },
+                  "else": { "properties": { "jwt": false } }
+                }
+              ]
             }
           }
         },
--- a/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
+++ b/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
@@ -160,6 +160,12 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                     JsonSerializer.Serialize(writer, value.Roles, options);
                 }
 
+                if(value?.UserProvidedMaxQueryParallelism is true)
+                {
+                    writer.WritePropertyName("max-query-parallelism");
+                    JsonSerializer.Serialize(writer, value.MaxQueryParallelism, options);
+                }
+
                 writer.WriteEndObject();
             }
             else
--- a/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
+++ b/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
@@ -160,7 +160,7 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                     JsonSerializer.Serialize(writer, value.Roles, options);
                 }
 
-                if(value?.UserProvidedMaxQueryParallelism is true)
+                if (value?.UserProvidedMaxQueryParallelism is true)
                 {
                     writer.WritePropertyName("max-query-parallelism");
                     JsonSerializer.Serialize(writer, value.MaxQueryParallelism, options);
--- a/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
+++ b/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
@@ -55,12 +55,13 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                 bool? enabled = null;
                 int? cacheTtlSeconds = null;
                 HashSet<string>? roles = null;
+                int? maxQueryParallelism = null;
 
                 while (reader.Read())
                 {
                     if (reader.TokenType is JsonTokenType.EndObject)
                     {
-                        return new RuntimeHealthCheckConfig(enabled, roles, cacheTtlSeconds);
+                        return new RuntimeHealthCheckConfig(enabled, roles, cacheTtlSeconds, maxQueryParallelism);
                     }
 
                     string? property = reader.GetString();
@@ -120,7 +121,19 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                             }
 
                             break;
+                        case "max-query-parallelism":
+                            if (reader.TokenType is not JsonTokenType.Null)
+                            {
+                                int parseMaxQueryParallelism = reader.GetInt32();
+                                if (parseMaxQueryParallelism < 4 || parseMaxQueryParallelism > 8)
+                                {
+                                    throw new JsonException($"Invalid value for health max-query-parallelism: {parseMaxQueryParallelism}. Value must be between 4 and 8.");
+                                }
 
+                                maxQueryParallelism = parseMaxQueryParallelism;
+                            }
+
+                            break;
                         default:
                             throw new JsonException($"Unexpected property {property}");
                     }
--- a/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
+++ b/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
@@ -126,7 +126,8 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                             {
                                 // Allow user to set values between 1 and 8 (inclusive). If not set, the value will be set to 4 during health check.
                                 int userValue = reader.GetInt32();
-                                int parseMaxQueryParallelism = Math.Clamp(userValue, 1, EntityCacheOptions.UPPER_LIMIT_MAX_QUERY_PARALLELISM);
+                                int parseMaxQueryParallelism = Math.Clamp(userValue, RuntimeHealthCheckConfig.LOWEST_MAX_QUERY_PARALLELISM,
+                                    RuntimeHealthCheckConfig.UPPER_LIMIT_MAX_QUERY_PARALLELISM);
                                 maxQueryParallelism = parseMaxQueryParallelism;
                             }
 
--- a/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
+++ b/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
@@ -124,12 +124,9 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                         case "max-query-parallelism":
                             if (reader.TokenType is not JsonTokenType.Null)
                             {
-                                int parseMaxQueryParallelism = reader.GetInt32();
-                                if (parseMaxQueryParallelism < 4 || parseMaxQueryParallelism > 8)
-                                {
-                                    throw new JsonException($"Invalid value for health max-query-parallelism: {parseMaxQueryParallelism}. Value must be between 4 and 8.");
-                                }
-
+                                // MaxQueryParallelism is an integer that should be between 4 and 8.
+                                int parseMaxQueryParallelism = Math.Min(Math.Max(reader.GetInt32(), EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM),
+                                    EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM_LIMIT);
                                 maxQueryParallelism = parseMaxQueryParallelism;
                             }
 
--- a/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
+++ b/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
@@ -124,9 +124,9 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                         case "max-query-parallelism":
                             if (reader.TokenType is not JsonTokenType.Null)
                             {
-                                // MaxQueryParallelism is an integer that should be between 4 and 8.
-                                int parseMaxQueryParallelism = Math.Min(Math.Max(reader.GetInt32(), EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM),
-                                    EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM_LIMIT);
+                                // Allow user to set values between 1 and 8 (inclusive). If not set, the value will be set to 4 during health check.
+                                int userValue = reader.GetInt32();
+                                int parseMaxQueryParallelism = Math.Clamp(userValue, 1, EntityCacheOptions.UPPER_LIMIT_MAX_QUERY_PARALLELISM);
                                 maxQueryParallelism = parseMaxQueryParallelism;
                             }
 
--- a/src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
+++ b/src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
@@ -30,8 +30,16 @@ public record RuntimeHealthCheckConfig : HealthCheckConfig
     [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
     public bool UserProvidedTtlOptions { get; init; } = false;
 
+    /// <summary>
+    /// Flag to indicate if the user has provided a value for MaxQueryParallelism.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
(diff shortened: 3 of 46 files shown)


schemas/dab.draft.schema.json
```diff
@@ -412,6 +412,11 @@
               "type": "integer",
               "description": "Time to live in seconds for the Comprehensive Health Check Report cache entry.",
               "default": 5
+            },
+            "max-query-parallelism": {
+              "type": "integer",
+              "description": "The max degree of parallelism for running parallel health check queries.",
+              "default": 4
```
Aniruddh25, line 464: Please update PR title to reflect what it is supposed to be implementing - a more descriptive title that is useful for developers to easily follow just by reading it - this gets added to our Release Checklist, and helps to keep a clear self-explanatory list of features.

souvikghosh04: Sorry my bad. The branch name somehow got into the title. Updated it now. Thanks!

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -120,7 +121,16 @@ internal HealthCheckOptionsConverter(bool replaceEnvVar)
                             }
 
                             break;
+                        case "max-query-parallelism":
+                            if (reader.TokenType is not JsonTokenType.Null)
+                            {
+                                // MaxQueryParallelism is an integer that should be between 4 and 8.
```
RubenCerna2079: I thought the parallelism was between 1 and 8

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -120,7 +121,16 @@ internal HealthCheckOptionsConverter(bool replaceEnvVar)
...
+                                int parseMaxQueryParallelism = Math.Min(Math.Max(reader.GetInt32(), EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM),
```
RubenCerna2079: Just to make sure I understand, this would override the choice of the user if it where more than the max limit?

src/Config/ObjectModel/EntityCacheOptions.cs
```diff
@@ -28,6 +28,16 @@ public record EntityCacheOptions
     /// </summary>
     public const string L2_CACHE_PROVIDER = "redis";
 
+    /// <summary>
+    /// Default maximum query parallelism for health check.
+    /// </summary>
+    public const int DEFAULT_MAX_QUERY_PARALLELISM = 4;
+
+    /// <summary>
+    /// Default maximum query parallelism limit for health check.
```
RubenCerna2079:
```suggestion
    /// Upper limit of maximum query parallelism for health check.
```

src/Config/ObjectModel/EntityCacheOptions.cs
```diff
@@ -28,6 +28,16 @@ public record EntityCacheOptions
...
+    /// </summary>
+    public const int DEFAULT_MAX_QUERY_PARALLELISM_LIMIT = 8;
```
RubenCerna2079:
```suggestion
    public const int UPPER_LIMIT_MAX_QUERY_PARALLELISM = 8;
```

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -196,18 +197,52 @@ private async Task UpdateDataSourceHealthCheckResultsAsync(ComprehensiveHealthCh
...
+            if (enabledEntities.Count == 0)
+            {
+                return;
+            }
+
+            ConcurrentBag<HealthCheckResultEntry> concurrentChecks = new();
+            int maxParallelism = runtimeConfig.Runtime?.Health?.MaxQueryParallelism ?? EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM;
+
+            await Parallel.ForEachAsync(enabledEntities, new ParallelOptions { MaxDegreeOfParallelism = maxParallelism }, async (entity, _) =>
+            {
```
RubenCerna2079, line 224: Could you add some comments, I had a bit of a hard time understanding how this works?

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
aaronburtle: This is just defining the max, which seems to be between 4 and 8.

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
aaronburtle: > Just to make sure I understand, this would override the choice of the user if it where more than the max limit?

ya this takes the higher of the user's choice and the default, then takes the lower of that compared to the default limit.

src/Config/ObjectModel/EntityCacheOptions.cs
aaronburtle: Similar to Ruben's suggestion, the fact this has `default` in its name implies that it can be raised, but the parsing logic seems like this is the upper limit in all cases. Can this value be increased if the user doesnt want to use the default limit?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -196,18 +197,52 @@ private async Task UpdateDataSourceHealthCheckResultsAsync(ComprehensiveHealthCh
 
         // Updates the Entity Health Check Results in the response. 
         // Goes through the entities one by one and executes the rest and graphql checks (if enabled).
-        private async Task UpdateEntityHealthCheckResultsAsync(ComprehensiveHealthCheckReport ComprehensiveHealthCheckReport, RuntimeConfig runtimeConfig)
+        private async Task UpdateEntityHealthCheckResultsAsync(ComprehensiveHealthCheckReport report, RuntimeConfig runtimeConfig)
         {
-            if (runtimeConfig?.Entities != null && runtimeConfig.Entities.Entities.Any())
+            if (runtimeConfig?.Entities?.Entities?.Any() != true)
```
aaronburtle: I think you can do `if (runtimeConfig?.Entities?.Entities is not { Count: > 0 })` here for faster lookup (by avoiding the call to .Any()) and I think a little better readability.

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -196,18 +197,52 @@ private async Task UpdateDataSourceHealthCheckResultsAsync(ComprehensiveHealthCh
...
+                        foreach (HealthCheckResultEntry check in localReport.Checks)
+                        {
+                            concurrentChecks.Add(check);
+                        }
                     }
                 }
-            }
+                catch (Exception ex)
+                {
+                    Console.Error.WriteLine($"Error processing entity '{entity.Key}': {ex.Message}");
```
aaronburtle: I noticed co-pilot was asking about using a Logger here, the class has a logger as one of its fields, will it not be available here for use?

src/Service/HealthCheck/HealthCheckHelper.cs
souvikghosh04: not necessarily faster but `Any()` is optimized and performs better in `IEnumerable` collection like this. please see the benchmarks: https://code-maze.com/any-vs-count-dotnet/ and this post: https://stackoverflow.com/questions/305092/which-method-performs-better-any-vs-count-0

src/Config/ObjectModel/EntityCacheOptions.cs
souvikghosh04: yes, the degree is configurable as mentioned in the description, under Usage. Both, `DEFAULT_MAX_QUERY_PARALLELISM` and `UPPER_LIMIT_MAX_QUERY_PARALLELISM` acts as guards. While the lower limit can still go less than 4 (default) upto 1, the higher limit is 8. In case, of a value outside these range, we internally adjust it accordingly.

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
souvikghosh04: yes correct. renamed the variables and added additional comments for better clarity

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
souvikghosh04: Yes, the parallelism is between 1 and 8. Default is 4 in case the value is not provided. Out of range values gets adjusted by the system within the limits.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -18,11 +18,14 @@ public record RuntimeHealthCheckConfig : HealthCheckConfig
     [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
     public bool UserProvidedTtlOptions { get; init; } = false;
 
+    [JsonPropertyName("max-query-parallelism")]
```
Aniruddh25, line 42: Is the above TODO still relevant? Is the `MaxDop` property similar to MaxQueryParallelism?

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -18,11 +18,14 @@ public record RuntimeHealthCheckConfig : HealthCheckConfig
...
+    public int MaxQueryParallelism { get; set; } = EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM;
+
```
Aniruddh25, line 44: Do we need a `UserProvidedMaxQueryParallelism` flag to denote the user provided values? This way configuration doesnt bloat when writing options to the file system.. Is this property going to be configurable via the CLI?

src/Service/HealthCheck/HealthCheckHelper.cs
aaronburtle: I believe those are comparing `Any()` to the extension method `Count()`, but I think you can use the property `Count` which should just always be O(1).

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
souvikghosh04: The actual/effective value is getting logged during the invocation of health check. Currently this is set via of the configuration file only. We might be able to move this to a declarative property using a new environment variable. For now, this configuration is not a compulsion and system uses the default value of 4 internally if not configured.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
souvikghosh04: yes it meant the same thing. Removed that commented code now.

src/Service/HealthCheck/HealthCheckHelper.cs
souvikghosh04: thanks for pointing that. you might already have seen that the difference is almost negligible for `Any()` vs `Count` and `Count` is slightly better. However, this check is not needed since we are already checking if `enabledEntities.Count == 0` and logging and returning from there.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
Aniruddh25: I understand the effective value is getting logged. What I am concerned about is this:
Consider this command before and after your change:

`dab init --database-type mssql` 
when it creates the default configuration file, it does not write the optional properties with their default values to the file. 
However, when you run the same command after your change, I think it will write this property to the file - `"max-query-parallelism" : 4` 
which is the undesired behavior. 

I hope I am clear on this, if not, lets set up a call to avoid the back and forth.

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -196,18 +197,56 @@ private async Task UpdateDataSourceHealthCheckResultsAsync(ComprehensiveHealthCh
...
+            if (enabledEntities.Count == 0)
+            {
+                _logger.LogInformation("No enabled entities found for health checks. Skipping entity health checks.");
+                return;
+            }
+
+            ConcurrentBag<HealthCheckResultEntry> concurrentChecks = new();
+
+            // Use MaxQueryParallelism from RuntimeConfig or default to EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM
+            int maxParallelism = runtimeConfig.Runtime?.Health?.MaxQueryParallelism ?? EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM;
```
Aniruddh25: if `MaxQueryParallelism` is set to the default value anyway in the constructor `RuntimeHealthCheckConfig`, why do we need the check here?

src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
```diff
@@ -124,9 +124,9 @@ internal HealthCheckOptionsConverter(bool replaceEnvVar)
                         case "max-query-parallelism":
                             if (reader.TokenType is not JsonTokenType.Null)
                             {
-                                // MaxQueryParallelism is an integer that should be between 4 and 8.
-                                int parseMaxQueryParallelism = Math.Min(Math.Max(reader.GetInt32(), EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM),
-                                    EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM_LIMIT);
+                                // Allow user to set values between 1 and 8 (inclusive). If not set, the value will be set to 4 during health check.
+                                int userValue = reader.GetInt32();
+                                int parseMaxQueryParallelism = Math.Clamp(userValue, 1, EntityCacheOptions.UPPER_LIMIT_MAX_QUERY_PARALLELISM);
```
RubenCerna2079: Do you think it would be a good idea to also create a constant for the lower limit?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -199,23 +199,27 @@ private async Task UpdateDataSourceHealthCheckResultsAsync(ComprehensiveHealthCh
...
+                _logger.LogInformation("No enabled entities found for health checks. Skipping entity health checks.");
                 return;
             }
 
             ConcurrentBag<HealthCheckResultEntry> concurrentChecks = new();
+
+            // Use MaxQueryParallelism from RuntimeConfig or default to EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM
             int maxParallelism = runtimeConfig.Runtime?.Health?.MaxQueryParallelism ?? EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM;
 
+            _logger.LogTrace("Executing health checks for {Count} enabled entities with parallelism of {MaxParallelism}.", enabledEntities.Count, maxParallelism);
```
RubenCerna2079: Wouldn't this be better as an `Information` log instead of a `Trace` log?

src/Service/HealthCheck/HealthCheckHelper.cs
souvikghosh04: This is an additional safeguard since these are nullable properties and set the default value when we encounter null.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -7,6 +7,20 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 
 public record RuntimeHealthCheckConfig : HealthCheckConfig
 {
+    /// <summary>
+    /// Represents the lowest maximum query parallelism for health check.
+    /// </summary>
+    public const int LOWEST_MAX_QUERY_PARALLELISM = 1;
+    /// <summary>
```
RubenCerna2079, line 15: nit: Add new line

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -7,6 +7,20 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 
 public record RuntimeHealthCheckConfig : HealthCheckConfig
 {
+    /// <summary>
+    /// Represents the lowest maximum query parallelism for health check.
+    /// </summary>
```
RubenCerna2079, line 12: Is there a reason to move these constants from the other file?

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
souvikghosh04: Thanks for the clarification, that makes sense. I have updated the code to use `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` and adjusted the constructor to set `maxQueryParallelism` to null when the default value is used. This ensures the property doesn’t get serialized into the config file unless explicitly set by the user in config. Internally, the system would convert the null/unset value to our default value of 4. I have also added a new test case `HealthEndpointConfigShouldNotSerializeDefaultMaxQueryParallelism` to validate this behavior.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
souvikghosh04: They were added in the wrong file which was unrelated to Health Check Runtime config.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
```diff
@@ -36,6 +52,8 @@ public RuntimeHealthCheckConfig(bool? enabled, HashSet<string>? roles = null, in
             this.CacheTtlSeconds = EntityCacheOptions.DEFAULT_TTL_SECONDS;
         }
 
-        this.MaxQueryParallelism = maxQueryParallelism ?? EntityCacheOptions.DEFAULT_MAX_QUERY_PARALLELISM;
+        // Allow user to set values between 1 and 8 (inclusive). If not set, the value will be set to 4 during health check.
+        this.MaxQueryParallelism = maxQueryParallelism != DEFAULT_MAX_QUERY_PARALLELISM ? maxQueryParallelism : null;
```
Aniruddh25: If the user does specify the value to be the default value itself, we should still write to the file system - because to them using the CLI to modify the config properties should result in an updated config file.

So, its incorrect to not write to the file system when its default. we should only not write it when the value is `null`

Aniruddh25: It can be simplified to simply be:

`this.MaxQueryParallelism = maxQueryParellelism`

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -206,6 +206,41 @@ public async Task TestFailureHealthCheckGraphQLResponseAsync()
...
+        public void HealthEndpointConfigShouldNotSerializeDefaultMaxQueryParallelism()
+        {
+            JsonSerializerOptions options = new() { WriteIndented = false };
+
+            // Case 1: Default value (should NOT appear in output)
+            RuntimeHealthCheckConfig configWithDefault = new(
+                enabled: true,
+                roles: null,
+                cacheTtlSeconds: null,
+                maxQueryParallelism: RuntimeHealthCheckConfig.DEFAULT_MAX_QUERY_PARALLELISM // default value
```
Aniruddh25: Case 1 from your test should still be ok to write to file system. 

Scenario 1:
So, if any of the following commands are run, 
`dab configure --runtime.health.max-query-parallelism=4` OR

`dab configure --runtime.health.max-query-parallelism=8` - we should see the property written to the file system. 

Scenario 2:
We should skip writing only when the property that is being configured is something different and this property was not already present in the config file. 
e.g. running:  `dab configure --runtime.graphql=enabled` => this should not write the property `max-query-parallelism=4` with its value. 

Scenario 3:
Consider the following sequence of actions: 
1) `dab configure --runtime.health.max-query-parallelism=4` => writes to the file
2) `dab configure --runtime.graphql=enabled` => even though the property is not being set in this command, it was already written to the file system in step 1, so step 2 should write the property --runtime.health.max-query-parallelism=4 to the file system.

src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
souvikghosh04: fixed it keeping it.

        `if (maxQueryParallelism is not null)
        {
            this.MaxQueryParallelism = maxQueryParallelism;
            UserProvidedMaxQueryParallelism = true;
        }
        else
        {
            this.MaxQueryParallelism = DEFAULT_MAX_QUERY_PARALLELISM;
        }`

this is similar to how we have `cacheTtlSeconds`

src/Service.Tests/Configuration/HealthEndpointTests.cs
souvikghosh04: made the fix in similar to how we write the `cache-ttl-seconds` in RuntimeHealthOptionsConverterFactory.cs.

```
if(value?.UserProvidedMaxQueryParallelism is true)
                {
                    writer.WritePropertyName("max-query-parallelism");
                    JsonSerializer.Serialize(writer, value.MaxQueryParallelism, options);
                }
```

Aniruddh25 (approved):
Thank you for addressing all the comments

Issue #2716, opened 2025-06-11
[Bug]: CacheTtl not correctly updating through CLI


aaronburtle:
### What happened?

When using the CLI update command to update an entity's cache options, including the ttl setting would not result in the ttl being correctly written to the configuration file.

### Version

1.5

### What database are you using?

Azure SQL

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2717, opened 2025-06-11
Write CacheTtl to entity when add/update command is used and CacheTtl is provided

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2716

## What is this change?

When we create the entity cache options, we start with a new object and then later set the values for the fields. Because of this, the `bool` `UserProvidedTtlOptions` must explicitly be set when we later set these values or it will forever remain `false`. When it is false it is not written back out into the configuration file.

## How was this tested?

Manually used the update command in the CLI to verify the field is written out.

## Sample Request(s)

### DIFF
--- a/src/Cli/Utils.cs
+++ b/src/Cli/Utils.cs
@@ -843,6 +843,7 @@ namespace Cli
 
             EntityCacheOptions cacheOptions = new();
             bool isEnabled = false;
+            bool isCacheTtlUserProvided = false;
             int ttl = EntityCacheOptions.DEFAULT_TTL_SECONDS;
 
             if (cacheEnabled is not null && !bool.TryParse(cacheEnabled, out isEnabled))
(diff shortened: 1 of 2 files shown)


src/Cli/Utils.cs
```diff
@@ -855,20 +856,25 @@ public static EntityGraphQLOptions ConstructGraphQLTypeDetails(string? graphQL,
                 _logger.LogError("Invalid format for --cache.ttl. Accepted values are any non-negative integer.");
             }
 
+            if (cacheTtl is not null)
```
RubenCerna2079, line 860: Might be good to add a comment explaining why we do this

anushakolan (approved):
From the context of this PR, this looks good to me.

However, I would wait for other reviews from the team who have better context of this change.

PR #2718, opened 2025-06-11
Cherry-Pick Cache CLI Add and Update Commands Bug Fix for 1.5 Release

RubenCerna2079:
## Why make this change?

When we create the entity cache options, we start with a new object and then later set the values for the fields. Because of this, the `bool` `UserProvidedTtlOptions` must explicitly be set when we later set these values or it will forever remain `false`. When it is false it is not written back out into the configuration file.

## What is this change?

Fixes bug found in release candidate for 1.5
Cherry-picked PR:
- #2717 

## Sample Request(s)

### DIFF
--- a/src/Cli/Utils.cs
+++ b/src/Cli/Utils.cs
@@ -843,6 +843,7 @@ namespace Cli
 
             EntityCacheOptions cacheOptions = new();
             bool isEnabled = false;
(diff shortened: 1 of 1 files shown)


aaronburtle (approved):
LGTM!

PR #2729, opened 2025-06-16
Prevent empty input node generation in mutation builder.

Alekhya-Polavarapu:
## Why make this change?
Currently today when the tables has only autogenerated fileds, then the schema generation is failing with an error saying that "Empty input for create and update mutation input".
Error trace:
`For more details look at the `Errors` property.\r\n\r\n1. InputObject `CreateNewTableInput` has no fields declared. (HotChocolate.Types.InputObjectType)\r\n2. InputObject `UpdateNewTableInput` has no fields declared.`

https://github.com/Azure/data-api-builder/issues/2739

## What is this change?
This PR addresses the issue by conditionally generating create and update mutation input types only when the table contains at least one non-auto-generated field. This ensures that the schema remains valid and avoids generating empty input objects.

## How was this tested?
- Unit tests
- Manual testing

### For create:
image

### for update:
image

### DIFF
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -457,6 +457,11 @@
               "type": "integer",
               "description": "Time to live in seconds for the Comprehensive Health Check Report cache entry.",
               "default": 5
+            },
+            "max-query-parallelism": {
+              "type": "integer",
+              "description": "The max degree of parallelism for running parallel health check queries.",
+              "default": 4
             }
           }
         }
--- a/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
+++ b/src/Config/Converters/RuntimeHealthOptionsConvertorFactory.cs
@@ -55,12 +55,13 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                 bool? enabled = null;
                 int? cacheTtlSeconds = null;
                 HashSet<string>? roles = null;
+                int? maxQueryParallelism = null;
 
                 while (reader.Read())
                 {
                     if (reader.TokenType is JsonTokenType.EndObject)
                     {
-                        return new RuntimeHealthCheckConfig(enabled, roles, cacheTtlSeconds);
+                        return new RuntimeHealthCheckConfig(enabled, roles, cacheTtlSeconds, maxQueryParallelism);
                     }
 
                     string? property = reader.GetString();
@@ -120,7 +121,17 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                             }
 
                             break;
+                        case "max-query-parallelism":
+                            if (reader.TokenType is not JsonTokenType.Null)
+                            {
+                                // Allow user to set values between 1 and 8 (inclusive). If not set, the value will be set to 4 during health check.
+                                int userValue = reader.GetInt32();
+                                int parseMaxQueryParallelism = Math.Clamp(userValue, RuntimeHealthCheckConfig.LOWEST_MAX_QUERY_PARALLELISM,
+                                    RuntimeHealthCheckConfig.UPPER_LIMIT_MAX_QUERY_PARALLELISM);
+                                maxQueryParallelism = parseMaxQueryParallelism;
+                            }
 
+                            break;
                         default:
                             throw new JsonException($"Unexpected property {property}");
                     }
@@ -147,6 +158,12 @@ internal class RuntimeHealthOptionsConvertorFactory : JsonConverterFactory
                 {
                     writer.WritePropertyName("roles");
                     JsonSerializer.Serialize(writer, value.Roles, options);
+                }
+
+                if (value?.UserProvidedMaxQueryParallelism is true)
+                {
+                    writer.WritePropertyName("max-query-parallelism");
+                    JsonSerializer.Serialize(writer, value.MaxQueryParallelism, options);
                 }
 
                 writer.WriteEndObject();
--- a/src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
+++ b/src/Config/HealthCheck/RuntimeHealthCheckConfig.cs
@@ -7,22 +7,46 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 
 public record RuntimeHealthCheckConfig : HealthCheckConfig
 {
+    /// <summary>
+    /// Represents the lowest maximum query parallelism for health check.
+    /// </summary>
+    public const int LOWEST_MAX_QUERY_PARALLELISM = 1;
+
+    /// <summary>
+    /// Default maximum query parallelism for health check.
+    /// </summary>
+    public const int DEFAULT_MAX_QUERY_PARALLELISM = 4;
+
+    /// <summary>
+    /// Upper limit of maximum query parallelism for health check.
+    /// </summary>
+    public const int UPPER_LIMIT_MAX_QUERY_PARALLELISM = 8;
+
     [JsonPropertyName("cache-ttl-seconds")]
     public int CacheTtlSeconds { get; set; }
 
     public HashSet<string>? Roles { get; set; }
 
-    // TODO: Add support for parallel stream to run the health check query in upcoming PRs
-    // public int MaxDop { get; set; } = 1; // Parallelized streams to run Health Check (Default: 1)
-
     [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
     public bool UserProvidedTtlOptions { get; init; } = false;
+
+    /// <summary>
+    /// Flag to indicate if the user has provided a value for MaxQueryParallelism.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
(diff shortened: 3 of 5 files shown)


src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -54,32 +55,35 @@ private static InputObjectTypeDefinitionNode GenerateCreateInputTypeForRelationa
...
+
+                // Create input object for this entity.
+                input =
+                    new(
+                        location: null,
+                        inputName,
+                        new StringValueNode($"Input type for creating {name}"),
+                        new List<DirectiveNode>(),
+                        inputFields!
+                    );
```
RubenCerna2079: nit:

```suggestion
                        inputFields!);
```

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
```diff
@@ -31,8 +31,8 @@ public static class CreateMutationBuilder
         /// <param name="databaseType">Database type of the relational database to generate input type for.</param>
         /// <param name="entities">Runtime config information.</param>
         /// <param name="IsMultipleCreateOperationEnabled">Indicates whether multiple create operation is enabled</param>
-        /// <returns>A GraphQL input type with all expected fields mapped as GraphQL inputs.</returns>
-        private static InputObjectTypeDefinitionNode GenerateCreateInputTypeForRelationalDb(
+        /// <returns>An optional GraphQL input type with all expected fields mapped as GraphQL inputs.</returns>
+        private static InputObjectTypeDefinitionNode? GenerateCreateInputTypeForRelationalDb(
```
RubenCerna2079, line 35: Just out of curiosity, I am not entirely sure how the mutation builder works for non-relational databases. So do you think this might also be a bug that could be found there? If so then it would also be a good idea to change the `GenerateCreateInputTypeForNonRelationalDb` function.

src/Service.GraphQLBuilder/Mutations/UpdateAndPatchMutationBuilder.cs
```diff
@@ -234,50 +248,56 @@ public static FieldDefinitionNode Build(
...
 
-            inputValues.Add(new InputValueDefinitionNode(
+                inputValues.Add(new InputValueDefinitionNode(
                     location: null,
                     new NameNode(INPUT_ARGUMENT_NAME),
                     new StringValueNode($"Input representing all the fields for updating {name}"),
                     new NonNullTypeNode(new NamedTypeNode(input.Name)),
                     defaultValue: null,
                     new List<DirectiveNode>()));
+
```
RubenCerna2079: nit: remove line

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -194,6 +194,40 @@ type Foo @model(name:""Foo"") {
...
+            {
+                { "Foo", DatabaseType.MSSQL }
+            };
+
+            DocumentNode mutationRoot = MutationBuilder.Build(
+                root,
+                entityNameToDatabasetype,
+                new(new Dictionary<string, Entity> { { "Foo", GenerateEmptyEntity() } }),
+                entityPermissionsMap: _entityPermissions
+                );
```
RubenCerna2079:
```suggestion
                entityPermissionsMap: _entityPermissions);
```

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -194,6 +194,40 @@ type Foo @model(name:""Foo"") {
...
+
+            Dictionary<string, DatabaseType> entityNameToDatabasetype = new()
+            {
+                { "Foo", DatabaseType.MSSQL }
+            };
+
+            DocumentNode mutationRoot = MutationBuilder.Build(
+                root,
+                entityNameToDatabasetype,
+                new(new Dictionary<string, Entity> { { "Foo", GenerateEmptyEntity() } }),
```
RubenCerna2079, line 220:
```suggestion
                new (new Dictionary<string, Entity> { { "Foo", GenerateEmptyEntity() } }),
```

RubenCerna2079:
@Alekhya-Polavarapu could you update the description for the PR? I see that you wanted to create some new lines but it shows as `\r\n\r\n`

src/Service.GraphQLBuilder/Mutations/UpdateAndPatchMutationBuilder.cs
```diff
@@ -89,17 +90,25 @@ private static InputObjectTypeDefinitionNode GenerateUpdateInputType(
                     return GenerateSimpleInputType(name, f, databaseType, operation);
                 });
 
-            InputObjectTypeDefinitionNode input =
+            if (inputFields.Any())
+            {
+                List<InputValueDefinitionNode> inputFieldsList = inputFields
+                    .Select(i => i!)
```
Aniruddh25, line 97: what if the inputFields IEnumerable contains some null and some non-null fields based on if GenerateSimpleType returns null or non-null? Can we still safely cast it to ignore the null by doing `i => i!`  ?

Aniruddh25 (changes requested):
Looks good overall, but waiting for clarification on one of the comments.

src/Service.GraphQLBuilder/Mutations/CreateMutationBuilder.cs
Alekhya-Polavarapu: @RubenCerna2079 , for the non-relational DBs, there are no auto generated fields and everything needs to come from a user and so it could not have an empty update or create input and so non relational dbs are not impacted.

src/Service.GraphQLBuilder/Mutations/UpdateAndPatchMutationBuilder.cs
Alekhya-Polavarapu: if i understand your comment correct, inputFields IEnumerable ends up with null node fields only in the case of complex object and the complex object has all fields auto generated inside of it and in that case it ends up with null and we can safely ignore those null's when creating the updateInput for original type.

otherthan that, i don't think Null's can end up in IEnumerable, even if in some edge case, it can be safely ignored.

example: i have a type publisher_PP which has only auto generated fields, i defined a relationship with Publisher_PP and Book for Publisher, and for the creaatePubliser, i do not see Publihser_PP showing up(which is getting filtered as it returned with null in the IEnumerable)

let me know if am missing anything here
realtionshipempty

Alekhya-Polavarapu:
> @Alekhya-Polavarapu could you update the description for the PR? I see that you wanted to create some new lines but it shows as `\r\n\r\n`

thats actually part of error trace, i updated the error trace within code block

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
Alekhya-Polavarapu: with the format(pre-commit hook), this is getting changed and i alsso noticed that at others locations in this file as well do not have space, so am assuming that it should be fine!.

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
```diff
@@ -194,6 +194,39 @@ type Foo @model(name:""Foo"") {
             Assert.AreEqual("bar", argType.Fields[0].Name.Value);
         }
 
+        [TestMethod]
+        [TestCategory("Mutation Builder - Create")]
+        public void CreateMutationExcludeIdForAllAutogeneratedFields()
```
Aniruddh25: Can you please add a similar test for UpdateMutationExcludeIdForAllAutogeneratedFields() ?

src/Service.GraphQLBuilder/Mutations/UpdateAndPatchMutationBuilder.cs
Aniruddh25: The graphic you showed makes sense for the `createPublisher` mutation. But if you test `updatePublisher` workflow, `updatePublisher_PP` will be null, as you rightly pointed out -> it will be null if complex object has all fields auto generated inside of it. 

Now isnt the following statement saying use the null `updatePublisher_PP` input type and other fields of the original type Publisher to convert it into a list?

`List<InputValueDefinitionNode> inputFieldsList = inputFields
                    .Select(i => i!)` 

I think this should instead be `.Where(i => i !=null)`

src/Service.Tests/GraphQLBuilder/MutationBuilderTests.cs
Alekhya-Polavarapu: i asserted the update and delete in the same test, i missed adding the correspodning test category, will add them. 
please let me know if am missing anything!

Aniruddh25: Oh in that case, could you just update the name of the test? not mention Create - can be simply `MutationExcludedForAll...`

Aniruddh25 (approved):
LGTM, thanks for addressing the concerns!

Issue #2735, opened 2025-06-20
[Bug]: Cache-Control Header issue while running query


rusamant:
### What happened?

Following simple instructions mentioned here :https://learn.microsoft.com/en-us/azure/data-api-builder/quickstart-sql

On Dab start and running query we are seeing error : Request Header Cache-Control is invalid

![Image](https://github.com/user-attachments/assets/89c08554-37e2-4b12-9f01-1cacc786a1ae)

code line in stack trace :https://github.com/Azure/data-api-builder/blob/d05fecb483d3dee2a990cd8f3a3fed1a1d95954c/src/Core/Resolvers/Sql%20Query%20Structures/SqlQueryStructure.cs#L591

### Version

1.5

### What database are you using?

Azure SQL

### What hosting model are you using?

Static Web Apps (SWA)

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Aniruddh25:
We are limiting the Cache-control options in the Request header to only be one of these 3: 

no-cache, no-store, only-if-cached. This is causing any other values of this header to error out. 

This is a regression in current behavior and should be fixed with a new patch for dab 1.5

PR #2737, opened 2025-06-20
Allow unsupported cache-control options without error

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2735

## What is this change?

Removes the check to make sure that only supported `cache-control` options are included in the request headers.

## How was this tested?

Manually ran GraphQL request with headers that failed in bug report. To reproduce use any valid graphQL query with cache enabled in the runtime section of the config, and for the entity being queried.

image

Then any graphQL query against that entity with `cache-control` headers should enter the code path in question.

## Sample Request(s)

```
query {
  publishers (first: 5 orderBy: {name: DESC
  })
  {
    items {
      id
    }
  }
}
```


Aniruddh25 (approved):
LGTM, once we have all tests confirmed passing.

Issue #2576, opened 2025-02-14
Add Serialization & Deserialization of Sink File Properties


RubenCerna2079:
We will add a feature that allows the user to send their logs to a file sink. In order to do that, we will require the user to include the `file` property, which will be under `runtime.telemetry`. We therefore add the following to the DAB config schema at the top level and include the ability to configure these settings use the CLI dab configure command.
```
{
  "runtime": {
    ...
    "telemetry": {
      "file": {
        "enabled": true, // (boolean, default: false)
        "path": "/logs/dab-log.txt", // (string, default: "/logs/dab-log.txt")
        "rolling-interval": "12Hourse | Day | Week", // (enum, default: "Day")
        "retained-file-count-limit": 7, // (int, default: 1)
        "file-size-limit-bytes": 1048576 // (int, default: 1048576)
      }
    }
    ...
  }
}
```

## Json Schema
Update our JSON schema. Add properties and constraints and defaults.
1. `file` requires `runtime.telemetry`

## CLI Updates
Add command line support.
1. Add `dab configure --file.enabled`
2. Add `dab configure --file.path`
3. Add `dab configure --file.rolling-interval`
4. Add `dab configure --file.retained-file-count-limit`
5. Add `dab configure --file.file-size-limit-bytes`

Reference other PRs that have included additions to the config file to see broader context of how new properties in the config need to be integrated into the broader codebase. For an example of adding to the object model and config/schema see: https://github.com/Azure/data-api-builder/pull/2606

For an example of adding CLI configure options see: https://github.com/Azure/data-api-builder/pull/2435

PR #2752, opened 2025-07-07
Add Serialization & Deserialization of Sink File Properties

RubenCerna2079:
## Why make this change?
- This change solves issue #2576.
We need to add the new `File Sink` properties that will later be used to send logs to `.txt` files locally. The properties need to have all the necessary components to be serialized and deserialized from the config file.

## What is this change?
- This change adds the `File Sink` properties to the schema file.
- Creates a new file `FileSinkConverter.cs` where the properties are serialized and deserialized.
- Creates a new file `FileSinkOptions.cs` where the deserialized properties are turned to usable objects and adds the object to the `Telemetry` options.

#### JSON Configuration Schema
The configuration now supports the following structure under `runtime.telemetry`:
```json
{
  "runtime": {
    "telemetry": {
      "file": {
        "enabled": true,
        "path": "/logs/dab-log.txt",
        "rolling-interval": "Day",
        "retained-file-count-limit": 7,
        "file-size-limit-bytes": 1048576
      }
    }
  }
}
```

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests
- [x] Manual Tests

 Tested that the newly created properties inside the config file are saved correctly as objects inside DAB and if the properties are written incorrectly, an exception is raised.


src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -819,5 +819,220 @@ private void SetupFileSystemWithInitialConfig(string jsonConfig)
...
+        [TestMethod]
+        public void TestUpdateMultipleTelemetryFileOptions()
+        {
+            // Arrange -> all the setup which includes creating options.
+            SetupFileSystemWithInitialConfig(INITIAL_CONFIG);
+
+            // Act: Attempts to update multiple file options at once
+            ConfigureOptions options = new(
+                runtimeTelemetryFileEnabled: true,
+                runtimeTelemetryFilePath: "/var/log/custom-dab.log",
```
souvikghosh04: in the draft config, the default path specified is /logs also the log file extension here has .log. is this intentional?

src/Config/ObjectModel/FileSinkOptions.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
+
+/// <summary>
+/// Represents the options for configuring file sink telemetry.
+/// </summary>
```
souvikghosh04, line 11: nit: might add description of input/output params in the summary?

src/Config/ObjectModel/FileSinkOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+public record FileSinkOptions(
+    bool Enabled = false,
+    string? Path = "/logs/dab-log.txt",
+    RollingIntervalMode? RollingInterval = RollingIntervalMode.Day,
+    int? RetainedFileCountLimit = 1,
+    int? FileSizeLimitBytes = 1048576)
```
souvikghosh04: these should be in config/Enums?

src/Config/ObjectModel/TelemetryOptions.cs
```diff
@@ -9,7 +9,7 @@ namespace Azure.DataApiBuilder.Config.ObjectModel;
 /// <summary>
 /// Represents the options for telemetry.
 /// </summary>
```
souvikghosh04, line 11: nit: please add/update the input/output params in the summary

src/Config/ObjectModel/FileSinkOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License.
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
+
+/// <summary>
+/// Represents the options for configuring file sink telemetry.
+/// </summary>
+public record FileSinkOptions(
+    bool Enabled = false,
+    string? Path = "/logs/dab-log.txt",
```
souvikghosh04: what happens if user configures an invalid path (unsupported extension, too long file name/path, invalid file name etc.)?

src/Cli/Commands/AddTelemetryOptions.cs
```diff
@@ -68,6 +78,26 @@ public AddTelemetryOptions(
         [Option("otel-service-name", Default = "dab", Required = false, HelpText = "(Default: dab) Headers for Open Telemetry for telemetry data")]
         public string? OpenTelemetryServiceName { get; }
 
+        // To specify whether File Sink telemetry should be enabled. This flag is optional and default value is false.
+        [Option("file-sink-enabled", Default = CliBool.False, Required = false, HelpText = "Enable/disable file sink telemetry logging. Default: false (boolean).")]
```
aaronburtle: For any of these properties with defaults, if the default value is being used, but was not supplied by the user, does the property end up writing back out to the config?

src/Config/ObjectModel/FileSinkOptions.cs
aaronburtle: if we try to make FileSinkOptions with values below the minimum in the schema, do we end up with an invalid config, or do we error out first?

aaronburtle (commented):
Can we add some kind of a test to validate the scenario?

src/Cli/Commands/AddTelemetryOptions.cs
RubenCerna2079: Will remove this section as it is out of scope for this PR

src/Config/ObjectModel/FileSinkOptions.cs
RubenCerna2079: Fixed the issue by having a converter that ensures that no value below the minimum is allowed.

src/Config/ObjectModel/FileSinkOptions.cs
RubenCerna2079: Should be caught by the regex pattern that is implemented

src/Config/ObjectModel/FileSinkOptions.cs
RubenCerna2079: I don't understand what you mean with this comment, could you explain a bit more?

src/Config/ObjectModel/FileSinkOptions.cs
RubenCerna2079: I think it would be too redundant since each of the variables has its own summary.

src/Cli.Tests/ConfigureOptionsTests.cs
RubenCerna2079: Deleted this test since it should be part of the CLI PR. But the file extension should not be .log, it was a mistake from my part to put it as such.

schemas/dab.draft.schema.json
```diff
@@ -483,6 +483,41 @@
...
+                },
+                "retained-file-count-limit": {
+                  "type": "integer",
+                  "description": "Maximum number of retained log files.",
+                  "default": 1,
+                  "minimum": 1
+                },
+                "file-size-limit-bytes": {
+                  "type": "integer",
+                  "description": "Maximum file size in bytes before rolling.",
```
Aniruddh25, line 514: when rolled to a new file, what will be the name of the new log file?

src/Config/ObjectModel/FileSinkOptions.cs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// <summary>
+    /// Default file size limit bytes for File Sink.
+    /// </summary>
+    public const int DEFAULT_FILE_SIZE_LIMIT_BYTES = 1048576;
+
+    /// <summary>
+    /// Whether File Sink is enabled.
+    /// </summary>
+    public bool Enabled { get; init; }
```
Aniruddh25, line 42: Shouldnt Enabled be nullable?

`bool? Enabled`

schemas/dab.draft.schema.json
RubenCerna2079: This is how it is going to look for the rolling interval follows a format of yyyyMMddHHmm, that gets more specific as the rolling interval gets smaller, so for `Day` it would be as the following.
```
/logs/dab-log-20250813.txt
/logs/dab-log-20250814.txt
```

And this is how it is going to look for the file size limit
```
/logs/dab-log-20250814_001.txt
/logs/dab-log-20250814_002.txt
```

src/Config/ObjectModel/FileSinkOptions.cs
RubenCerna2079: It is not necessary, only the parameter used in the constructor `enabled` should be nullable. If `enabled` is null then `Enabled` should be its default value, which is false.

src/Config/Converters/FileSinkConverter.cs
```diff
@@ -0,0 +1,164 @@
...
+            int? fileSizeLimitBytes = null;
+
+            while (reader.Read())
+            {
+                if (reader.TokenType == JsonTokenType.EndObject)
+                {
+                    return new FileSinkOptions(enabled, path, rollingInterval, retainedFileCountLimit, fileSizeLimitBytes);
+                }
+
+                string? propertyName = reader.GetString();
```
anushakolan, line 45: what happens if propertyName is empty? Is that possible since you used `string?`? If so, how does the code handle it?

src/Config/Converters/FileSinkConverter.cs
```diff
@@ -0,0 +1,164 @@
...
+
+                        break;
+
+                    default:
+                        throw new JsonException($"Unexpected property {propertyName}");
+                }
+            }
+        }
+
+        throw new JsonException("Failed to read the File Sink Options");
```
anushakolan, line 120: Can we also, give a reason on why did we fail to read the file sink options, so that it would be easy to debug.

src/Config/Converters/FileSinkConverter.cs
```diff
@@ -0,0 +1,164 @@
...
+using Azure.DataApiBuilder.Config.ObjectModel;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+class FileSinkConverter : JsonConverter<FileSinkOptions>
+{
+    // Determines whether to replace environment variable with its
+    // value or not while deserializing.
+    private bool _replaceEnvVar;
+
+    /// <param name="replaceEnvVar">Whether to replace environment variable with its
```
anushakolan: nit: It is cleaner to use this format, especially since the format description is exceeding a single line.

```
/// <param name="replaceEnvVar">
/// Indicates whether environment variables should be replaced with their actual values during deserialization.
/// </param>
```

schemas/dab.draft.schema.json
```diff
@@ -483,6 +483,40 @@
...
+                  "type": "string",
+                  "description": "Rolling interval for log files.",
+                  "default": "day",
+                  "enum": ["hour", "day", "week"]
+                },
+                "retained-file-count-limit": {
+                  "type": "integer",
+                  "description": "Maximum number of retained log files.",
+                  "default": 1,
+                  "minimum": 1
```
anushakolan, line 510: Does the maximum value have an upper bound?

schemas/dab.draft.schema.json
```diff
@@ -483,6 +483,40 @@
...
+                },
+                "file-size-limit-bytes": {
+                  "type": "integer",
+                  "description": "Maximum file size in bytes before rolling.",
+                  "default": 1048576,
+                  "minimum": 1
```
anushakolan, line 516: Same, do we have any limitation on the maximum size?

src/Config/Converters/FileSinkConverter.cs
```diff
@@ -0,0 +1,164 @@
...
+                        {
+                            enabled = reader.GetBoolean();
+                        }
+
+                        break;
+
+                    case "path":
+                        if (reader.TokenType is not JsonTokenType.Null)
+                        {
+                            path = reader.DeserializeString(_replaceEnvVar);
```
anushakolan, line 61: What happens when the string is not null but empty or invalid.

src/Config/Converters/FileSinkConverter.cs
```diff
@@ -0,0 +1,164 @@
...
+                        }
+
+                        break;
+
+                    case "rolling-interval":
+                        if (reader.TokenType is not JsonTokenType.Null)
+                        {
+                            rollingInterval = EnumExtensions.Deserialize<RollingIntervalMode>(reader.DeserializeString(_replaceEnvVar)!);
```
anushakolan, line 69: Are we checking anywhere else if the rollingInterval is in the range of accepted values?

src/Config/ObjectModel/FileSinkOptions.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+        else
+        {
+            FileSizeLimitBytes = DEFAULT_FILE_SIZE_LIMIT_BYTES;
+        }
+    }
+
+    /// <summary>
+    /// Flag which informs CLI and JSON serializer whether to write enabled
+    /// property and value to the runtime config file.
```
anushakolan: nit: Use `property and value` Or use `property/value` just to be consistent.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4164,12 +4165,95 @@ public void AzureLogAnalyticsSerialization(
             }
         }
 
+        /// <summary>
+        /// Tests different File Sink values to see if they are serialized and deserialized correctly to the Json config
+        /// </summary>
+        [DataTestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        [DataRow(true, "/file/path/exists.txt", RollingIntervalMode.Hour, 10, 256, true, "/file/path/exists.txt", RollingIntervalMode.Hour, 10, 256)]
+        [DataRow(false, "", RollingIntervalMode.Week, 5, 512, false, "", RollingIntervalMode.Week, 5, 512)]
```
anushakolan: As the errors below point out, there is no `RollingIntervalMode.Week`, then how did this test pass locally when you ran it?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4164,12 +4165,95 @@ public void AzureLogAnalyticsSerialization(
...
+            RollingIntervalMode? rollingInterval,
+            int? retainedFileCountLimit,
+            int? fileSizeLimitBytes,
+            bool expectedEnabled,
+            string expectedPath,
+            RollingIntervalMode expectedRollingInterval,
+            int expectedRetainedFileCountLimit,
+            int expectedFileSizeLimitBytes)
+        {
+            //Check if file values are expected to exist
```
anushakolan: nit: space after `//`

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4164,12 +4165,95 @@ public void AzureLogAnalyticsSerialization(
...
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(configWithCustomLogLevelJson, out RuntimeConfig? deserializedRuntimeConfig));
+
+            string serializedConfig = deserializedRuntimeConfig.ToJson();
+
+            using (JsonDocument parsedDocument = JsonDocument.Parse(serializedConfig))
+            {
+                JsonElement root = parsedDocument.RootElement;
+                JsonElement runtimeElement = root.GetProperty("runtime");
+
+                //Validate file property exists in runtime
```
anushakolan: nit: It is cleaner to have space after `//`

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4164,12 +4165,95 @@ public void AzureLogAnalyticsSerialization(
...
         }
 
+        /// <summary>
+        /// Tests different File Sink values to see if they are serialized and deserialized correctly to the Json config
+        /// </summary>
+        [DataTestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        [DataRow(true, "/file/path/exists.txt", RollingIntervalMode.Hour, 10, 256, true, "/file/path/exists.txt", RollingIntervalMode.Hour, 10, 256)]
+        [DataRow(false, "", RollingIntervalMode.Week, 5, 512, false, "", RollingIntervalMode.Week, 5, 512)]
+        [DataRow(null, null, null, null, null, false, "/logs/dab-log.txt", RollingIntervalMode.Day, 1, 1048576)]
```
anushakolan, line 4178: What about testing Infinite, Year, Month and Minute?

anushakolan:
Since the testing is done manually, it would be good to update the testing section on details about how the testing is done, sample configs used, sample logs files generated.

anushakolan (changes requested):

src/Config/Converters/FileSinkConverter.cs
RubenCerna2079: It throws the error found inside of the switch that there is an unexpected property.

src/Config/Converters/FileSinkConverter.cs
RubenCerna2079: This fail is for unexpected exceptions, also this exception is meant for the user to know that something is wrong with their config file. So I don't think it should be more specific.

schemas/dab.draft.schema.json
RubenCerna2079: The max value is int32, which is 2,147,483,647.

schemas/dab.draft.schema.json
RubenCerna2079: Also int32

src/Config/Converters/FileSinkConverter.cs
RubenCerna2079: The `EnumExtensions` function ensures that rollingInterval is in the range of accepted values

src/Config/Converters/FileSinkConverter.cs
RubenCerna2079: Will come in next PR, when the validation occurs it is expected that if the path that is given is not valid it will fail.

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Mistake from my part, I changed the valid values `RollingInterval`, before we had `Week` but it is unnecessary.

src/Config/ObjectModel/FileSinkOptions.cs
anushakolan: This comment is resolved but I still see both of them. Do you plan to leave it as is, or was this change missed being added.

src/Service.Tests/Configuration/ConfigurationTests.cs
anushakolan: Same, this conversation is marked resolved. Do you plan to leave it as it?

src/Config/ObjectModel/FileSinkOptions.cs
RubenCerna2079: I thought I changed all of them, I missed one

anushakolan:
There are a few minor comments that were marked as resolved without corresponding changes. If you intend to leave them as-is, that’s completely fine. However, addressing them would help maintain consistency across the codebase. Aside from that, everything else looks good to me.

src/Config/Converters/FileSinkConverter.cs
anushakolan: Curious, what happens if this is null. Is it possible that this can be null?
If `propertyName` is null, don't we get a `NullReferenceException`?

anushakolan:
LGTM!

src/Config/Converters/FileSinkConverter.cs
RubenCerna2079: The `propertyName` cannot be null, if that is the case then it will error out with a `JsonReaderException`

Aniruddh25 (approved):

PR #2757, opened 2025-07-09
Platform agnostic port resolution for health endpoint

souvikghosh04:
## Why make this change?

- Closes on #2765 

To robustly determine the internal port used by the application for self-referential HTTP calls (such as health checks), especially in containerized and reverse-proxy scenarios (e.g., Azure Container Apps), without relying on the ASP.NET Core Forwarded Headers middleware.

- Environment-agnostic: Works in Azure Container Apps, AKS, ACI, on-prem, and other containerized environments without requiring special middleware configuration.
- Resilient: Handles a variety of deployment scenarios and ingress/proxy setups.
- No security risk: Does not trust forwarded headers globally, reducing the risk of header spoofing.

## What is this change?

- Refactored ResolveInternalPort() to robustly determine the correct port for internal HTTP calls by:
  - Parsing the ASPNETCORE_URLS environment variable for the first HTTP port, including support for wildcard bindings like http://+:1234 or http://*:8080.
  - Falling back to port 5000 if no valid port is found.
  - No longer relying on X-Forwarded-Port or Host headers, which are not trustworthy for internal routing.
- Updated the health check HttpClient to always use http://localhost:<resolved-port> for internal calls, ensuring reliability across all environments.
- Updated HTTPS redirection logic to exclude /health (and /graphql if needed) from redirection, allowing internal HTTP health checks to succeed without being redirected.
- No changes are required to middleware configuration for this logic to work.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)

`GET /health`


src/Service/Startup.cs
```diff
@@ -561,10 +543,12 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, RuntimeC
...
             }
 
-            if (!Program.IsHttpsRedirectionDisabled)
-            {
-                app.UseHttpsRedirection();
-            }
+            // Use HTTPS redirection for all endpoints except /health and /graphql.
+            // This is necessary because ContextConfiguredHealthCheckClient base URI is http://localhost:{port} for internal API calls
+            app.UseWhen(
+                context => !(context.Request.Path.StartsWithSegments("/health") || context.Request.Path.StartsWithSegments("/graphql")),
```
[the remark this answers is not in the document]
souvikghosh04: /health and /graphql are always going to be fixed and no need to be configured as constants. this is a common pattern followed in health checks and similar scenarios to conditionally initializing middlewares in ASP.Net Core.

src/Service/Utilities/PortResolutionHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            // If no HTTP, fallback to HTTPS port if present
+            if (httpsPort.HasValue)
+            {
+                return httpsPort.Value;
+            }
+
+            // Configurable fallback port
+            string? defaultPortEnv = Environment.GetEnvironmentVariable("DEFAULT_PORT");
```
Aniruddh25, line 106: is DEFAULT_PORT a standard environment variable used in ACA that can be configured? or is it specific to DAB that we are adding?

Prefer this not to be DAB specific but if its specific to DAB, we should document. 

Should we use ASPNETCORE_HTTP_PORTS instead when ASPNETCORE_URL is not set? 

See this: https://learn.microsoft.com/en-us/dotnet/core/compatibility/containers/8.0/aspnet-port

src/Service/Startup.cs
```diff
@@ -283,26 +283,8 @@ public void ConfigureServices(IServiceCollection services)
...
-                        client.BaseAddress = new Uri(baseUri);
-                    }
-                    else
-                    {
-                        // Optional fallback if ever needed in non-request scenarios
-                        baseUri = $"http://localhost:{ResolveInternalPort()}";
-                        client.BaseAddress = new Uri(baseUri);
-                    }
-
-                    _logger.LogInformation($"Configured HealthCheck HttpClient BaseAddress as: {baseUri}");
```
Aniruddh25, line 306: We should retain the logger information where the HealthCheck HttpClient BaseAddress has been configured.

src/Service.Tests/UnitTests/PortResolutionHelperTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Negative tests for the <see cref="PortResolutionHelper.ResolveInternalPort"/> method.
+        /// </summary>
+        /// <param name="aspnetcoreUrls">A string representing the ASP.NET Core URLs to be tested.</param>
+        /// <param name="expectedPort">The expected port number that should be resolved.</param>
+        [DataTestMethod]
+        [DataRow("http://localhost:5000 https://localhost:443", 443)] // space delimiter: fallback to first HTTPS port
```
Aniruddh25: 443 is expected, but isnt it due to LastIndexOf(':') giving the : before 443? And that would still be from the section where you are checking for HTTP's trimmedPart. 

The reasoning to say fallback to first HTTPS port is incorrect IMO.

src/Service.Tests/UnitTests/PortResolutionHelperTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        [DataRow("http://localhost:5000|https://localhost:443", 443)] // pipe delimiter: fallback to first HTTPS port
+        [DataRow("localhost:5000", 5000)] // missing scheme: fallback to default
+        [DataRow("http://", 5000)] // incomplete URL: fallback to default
+        [DataRow("ftp://localhost:21", 5000)] // unsupported scheme: fallback to default
+        [DataRow("http://unix:/var/run/app.sock", 80)] // unix socket: defaults to 80 (no port specified)
+        [DataRow("http://unix:var/run/app.sock", 5000)] // malformed unix socket: fallback to default
+        [DataRow("http://unix:", 80)] // incomplete unix socket: defaults to 80
+        public void ResolveInternalPortResolvesCorrectPortNegativeTest(string aspnetcoreUrls, int expectedPort)
+        {
+            string originalUrls = Environment.GetEnvironmentVariable("ASPNETCORE_URLS");
```
Aniruddh25, line 117: The test case is exactly the same as the Positive test. Can we move this to a common helper function? Both the tests can call this helper

Aniruddh25 (commented):
Consider `ASPNETCORE_HTTP_PORTS` usage.

src/Service.Tests/UnitTests/PortResolutionHelperTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        /// <summary>
+        /// Negative tests for the <see cref="PortResolutionHelper.ResolveInternalPort"/> method.
+        /// </summary>
+        /// <param name="aspnetcoreUrls">A string representing the ASP.NET Core URLs to be tested.</param>
+        /// <param name="expectedPort">The expected port number that should be resolved.</param>
+        [DataTestMethod]
+        [DataRow("http://localhost:5000 https://localhost:443", 443)] // space delimiter: fallback to first HTTPS port
+        [DataRow("http://localhost:5000|https://localhost:443", 443)] // pipe delimiter: fallback to first HTTPS port
```
aaronburtle: Should these cases not simply use a default value because "http://localhost:5000|https://localhost:443" is not a valid Uri, or do we accept delimiting on space and pipe?

src/Service/Utilities/PortResolutionHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+                string[] parts = urls.Split(new[] { ';', ',' }, StringSplitOptions.RemoveEmptyEntries);
+
+                foreach (string part in parts)
+                {
+                    string trimmedPart = part.Trim();
+
+                    // Prefer HTTP
+                    if (trimmedPart.StartsWith("http://", StringComparison.OrdinalIgnoreCase))
+                    {
+                        int colonIndex = trimmedPart.LastIndexOf(':');
```
aaronburtle: This will try to get the port from URIs that are malformed. For example, "http://localhost:5000 foo bar baz :1234" will return as a port 1234. Shouldnt malformed URIs just use some default value?

Maybe we could use the URI.TryCreate first to determine if the form is valid to parse and get the port out of, with some manual handling of formats that we expect to see which URI.TryCreate can not handle, such as wildcards?

src/Service/Utilities/PortResolutionHelper.cs
souvikghosh04: `DEFAULT_PORT` is not a standard environment variable in .NET or Azure Container Apps. It is a DAB-specific fallback introduced to provide additional configurability if neither `ASPNETCORE_URLS` nor `ASPNETCORE_HTTP_PORTS` are set.
Agree with your suggestion and have updated the port resolution logic to respect that as well. Below is the precedence:

- ASPNETCORE_URLS (if set)
- ASPNETCORE_HTTP_PORTS (if set, per .NET 8 container guidance)
- DEFAULT_PORT (DAB-specific, as a last resort)
- Default to 5000

We should also document `DEFAULT_PORT` as a DAB-specific feature in our documentation.
This approach ensures compatibility with ACA and other .NET hosting environments and only uses `DEFAULT_PORT` as a fallback if the standard variables are not set.

src/Service/Startup.cs
souvikghosh04: Fixed

src/Service.Tests/UnitTests/PortResolutionHelperTests.cs
souvikghosh04: I have rephrased and updated the comment for better clarification.

src/Service.Tests/UnitTests/PortResolutionHelperTests.cs
souvikghosh04: Refactored this to a common helper

src/Service.Tests/UnitTests/PortResolutionHelperTests.cs
souvikghosh04: That's correct this is not a valid option but instead of failing it in `/health` calls, we refer the last part of the port (after last occurrence of `:`) and use that instead. also, we have Uri validation inside the `ResolveInternalPort()` to handle malformed URL.

src/Service/Utilities/PortResolutionHelper.cs
souvikghosh04: Uri validations are added now

src/Service/Utilities/PortResolutionHelper.cs
Aniruddh25: Sounds good, please open up a documentation issue for @JerryNixon to mention this precedence and `DEFAULT_PORT` fallback usage in our documentation

souvikghosh04: - #2775 has been opened

souvikghosh04:
> LGTM. Reopened 1 comment to ensure communication to update documentation happens.

- #2775

PR #2780, opened 2025-07-21
Remove caseInsensitive filter operator added back by incorrect merge

vadeveka:
## Why make this change?

caseInsensitive filter operator is not implemented in the resolvers leading to runtime error when used. Including it in schema seems to be an oversight. Removing it for schema generated until we have an approach defined for case insensitive comparisons with different source types

Relevant issue #2280

This was already addressed in PR https://github.com/Azure/data-api-builder/pull/2607
but was readded unintentionally by https://github.com/Azure/data-api-builder/pull/2348/files#diff-f74a036f2a72f2b3f2642c66613b149c5dca7a9696859a26d4ce95863c124d9f

## What is this change?

Remove caseInsensitive filter operator from schema generated

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests


rusamant (approved):

Alekhya-Polavarapu (approved):

Alekhya-Polavarapu:
just a suggestion, may be a negative test case might help here, so that no one else adds this unknowingly.

PR #2787, opened 2025-07-26
Add Logic to Connect to Azure Log Analytics 

RubenCerna2079:
## Why make this change?

Solves issue #2778

## What is this change?

This change adds logic to capture logs that DAB produces in order to send them to the users' Azure Log Analytics tables. It does this by creating a logger provider that is activated when the feature is enabled and sends the logs that are produced to a custom log collector. This custom log collector then flushes all of the saved logs periodically every certain amount of time to Azure Log Analytics.
It is also important to note, that the name of one of the properties in the feature was changed from `workspace-id` to `custom-table-name` as I found that the feature did not need the `workspace-id` to connect to the table, but it needed the name of custom log that is created as a resource by the user. The name is still not decided on, so if there are any other options that fit better, feel free to comment.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests


schemas/dab.draft.schema.json
```diff
@@ -423,9 +423,9 @@
                   "type": "object",
                   "additionalProperties": false,
```
Aniruddh25, line 424: Description still says:
>> The name is still not decided on, so if there are any other options that fit better, feel free to comment.

Is that accurate? I thought it is decided now. Please update the description if that is true.

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
```
Aniruddh25, line 1: The PR description could have been more elaborate to mention what the different classes do along with a class diagram. This helps for an easier review.

Something like: 

AzureLogAnalyticsFlusherService uploads the logs to Azure Log Analytics Workspace. 

It depends on AzureLogAnalyticsCustomLogCollector to retrieve logs from its queue.

AzureLogAnalyticsLoggerProvider creates the AzureLogAnalyticsLogger to trigger the population of logs by the CustomLogCollector.

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    public AzureLogAnalyticsLogger(string categoryName, ICustomLogCollector customLogCollector)
+    {
+        _categoryName = categoryName;
+        _customLogCollector = customLogCollector;
+    }
+
+    public IDisposable? BeginScope<TState>(TState state) where TState : notnull => default!;
+
+    public bool IsEnabled(LogLevel logLevel) => true;
```
Aniruddh25: why does this function take in the logLevel argument if its not using it?

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    public async void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
```
Aniruddh25: where is the eventId used?

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    public ILogger CreateLogger(string categoryName)
+    {
+        return new AzureLogAnalyticsLogger(categoryName, _customLogCollector);
+    }
+
+    public void Dispose() { }
+}
+
+public class AzureLogAnalyticsLogger : ILogger
```
Aniruddh25: a different class `AzureLogAnalyticsLogger` needs to go into a new file.

src/Service/Startup.cs
```diff
@@ -892,9 +905,32 @@ private void ConfigureAzureLogAnalytics(RuntimeConfig runtimeConfig)
                     return;
                 }
 
+                if (string.IsNullOrEmpty(AzureLogAnalyticsOptions.Auth?.CustomTableName))
```
Aniruddh25, line 915: Please move the logic to ConfigureAzureLogAnalyticsFlusherService into its own function

src/Service/Startup.cs
```diff
@@ -892,9 +905,32 @@ private void ConfigureAzureLogAnalytics(RuntimeConfig runtimeConfig)
...
+
+                if (string.IsNullOrEmpty(AzureLogAnalyticsOptions.Auth?.DceEndpoint))
+                {
+                    _logger.LogInformation("Logs won't be sent to Azure Log Analytics because the DCE Endpoint is not available in the config file.");
+                    return;
+                }
+
+                DefaultAzureCredential credential = new();
+                LogsIngestionClient logsIngestionClient = new(new Uri(AzureLogAnalyticsOptions.Auth.DceEndpoint), credential);
+                FlusherService = new AzureLogAnalyticsFlusherService(AzureLogAnalyticsOptions, CustomLogCollector, logsIngestionClient);
```
Aniruddh25: If a new `FlusherService` was added in the service collection on line 186, we should retrieve that singleton here and call StartAsync on it

src/Service/Startup.cs
```diff
@@ -892,9 +905,32 @@ private void ConfigureAzureLogAnalytics(RuntimeConfig runtimeConfig)
                     return;
                 }
 
+                if (string.IsNullOrEmpty(AzureLogAnalyticsOptions.Auth?.CustomTableName))
+                {
+                    _logger.LogInformation("Logs won't be sent to Azure Log Analytics because the Custom Table Name is not available in the config file.");
+                    return;
+                }
+
+                if (string.IsNullOrEmpty(AzureLogAnalyticsOptions.Auth?.DcrImmutableId))
```
Aniruddh25, line 921: The casing used in the PR title should be fixed:

`Add logic to connect to Azure Log Analytics`

only Azure Log Analytics needs to be capitalized.

src/Service/Startup.cs
```diff
@@ -892,9 +905,32 @@ private void ConfigureAzureLogAnalytics(RuntimeConfig runtimeConfig)
...
+                {
+                    _logger.LogInformation("Logs won't be sent to Azure Log Analytics because the DCR Immutable Id is not available in the config file.");
+                    return;
+                }
+
+                if (string.IsNullOrEmpty(AzureLogAnalyticsOptions.Auth?.DceEndpoint))
+                {
```
Aniruddh25, line 928: Aren't all these checks happening in dab validate?

src/Service/Startup.cs
```diff
@@ -892,9 +905,32 @@ private void ConfigureAzureLogAnalytics(RuntimeConfig runtimeConfig)
...
+                    _logger.LogInformation("Logs won't be sent to Azure Log Analytics because the DCE Endpoint is not available in the config file.");
+                    return;
```
Aniruddh25: If its only logging as an information, why are we returning pre-maturely without ?

I don't think dab should die if we Log Analytics is incorrectly configured. You can check with @JerryNixon to confirm this.

src/Service/Program.cs
```diff
@@ -195,6 +195,21 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, Tele
                         });
                     }
 
+                    if (Startup.AzureLogAnalyticsOptions.Enabled && !string.IsNullOrWhiteSpace(Startup.AzureLogAnalyticsOptions.Auth?.CustomTableName) &&
```
Aniruddh25: Can you please verify logs continue to flow to AppInsights and Open Telemetry in addition to LogAnalytics?

src/Service/Program.cs
```diff
@@ -195,6 +195,21 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, Tele
...
+                        !string.IsNullOrWhiteSpace(Startup.AzureLogAnalyticsOptions.Auth?.DcrImmutableId) && !string.IsNullOrWhiteSpace(Startup.AzureLogAnalyticsOptions.Auth?.DceEndpoint))
```
Aniruddh25: We can move all these checks to check if AzureLogAnalytics is enabled or not into a helper function and reuse it in Startup to create the singleton AzureFlusherService.

src/Service/Program.cs
```diff
@@ -195,6 +195,21 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, Tele
...
+                    {
+                        builder.AddProvider(new AzureLogAnalyticsLoggerProvider(Startup.CustomLogCollector));
+
+                        if (logLevelInitializer is null)
+                        {
+                            builder.AddFilter<AzureLogAnalyticsLoggerProvider>(category: string.Empty, logLevel);
+                        }
+                        else
+                        {
+                            builder.AddFilter<AzureLogAnalyticsLoggerProvider>(category: string.Empty, level => level >= logLevelInitializer.MinLogLevel);
```
Aniruddh25, line 208: why is category empty? If it's not needed, why add it as an argument to the LoggerProvider constructor, and hence the AzureLogAnalyticsLogger constructor?

src/Service/Startup.cs
```diff
@@ -173,6 +177,15 @@ public void ConfigureServices(IServiceCollection services)
                 });
             }
 
+            if (runtimeConfigAvailable
+                && runtimeConfig?.Runtime?.Telemetry?.AzureLogAnalytics is not null
+                && runtimeConfig.Runtime.Telemetry.AzureLogAnalytics.Enabled)
```
Aniruddh25: If you already check for Enabled or not before creating the singleton `AzureLogAnalyticsFlusherService`, why do you need to re-check if its enabled or not in its StartAsync function?

src/Service/Telemetry/AzureLogAnalyticsFlusherService.cs
```diff
@@ -0,0 +1,50 @@
...
+                List<AzureLogAnalyticsLogs> logs = _customLogCollector.DequeueAll();
+
+                if (logs.Count > 0)
+                {
+                    await _logsIngestionClient.UploadAsync<AzureLogAnalyticsLogs>(_options.Auth!.DcrImmutableId!, _options.Auth!.CustomTableName!, logs);
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Error uploading logs to Azure Log Analytics: {ex}");
```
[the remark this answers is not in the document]
Aniruddh25: @RubenCerna2079, can we use the default Startup Logger here instead of Console logging?

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.DataApiBuilder.Service.Telemetry;
+
+public interface ICustomLogCollector
```
Aniruddh25, line 17: None of the classes have any descriptions. Please add summary to better explain what the class is supposed to be accomplishing.

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    Task LogAsync(string message, LogLevel loggingLevel, string? source = null);
+    Task<List<AzureLogAnalyticsLogs>> DequeueAllAsync(string logType, int flushIntervalSeconds);
+}
+
+public class AzureLogAnalyticsCustomLogCollector : ICustomLogCollector
+{
+    private readonly Channel<AzureLogAnalyticsLogs> _logs = Channel.CreateUnbounded<AzureLogAnalyticsLogs>();
+
+    public async Task LogAsync(string message, LogLevel loggingLevel, string? source = null)
```
Aniruddh25: Is source a required property? I see that it is categoryName but we are not passing any value to the source. Why do we need this argument in that case?

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// <summary>
+    /// Tests if the logs are flushed correctly when Azure Log Analytics is enabled.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("Information Test Message", LogLevel.Information)]
+    [DataRow("Trace Test Message", LogLevel.Trace)]
+    [DataRow("Warning Test Message", LogLevel.Warning)]
+    public async void TestAzureLogAnalyticsFlushServiceSucceed(string message, LogLevel logLevel)
```
Aniruddh25: Since you also performed a manual test to ensure logs are flowing correctly into the log analytics workspace, please mention that in the PR description in the section How this was tested. Provide a screenshot of the logs that you saw.

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        DateTime dateTime = DateTime.UtcNow;
+        await _logs.Writer.WriteAsync(
+            new AzureLogAnalyticsLogs(
+                dateTime.ToString("o"),
+                loggingLevel,
```
Aniruddh25: Did you verify the format of the logs that you wrote with @JerryNixon?

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// <summary>
+    /// Custom logs ingestion to test that all the logs are being sent correctly to Azure Log Analytics
+    /// </summary>
+    private class CustomLogsIngestionClient : LogsIngestionClient
+    {
+        public List<AzureLogAnalyticsLogs> LogAnalyticsLogs { get; } = new();
+
+        public CustomLogsIngestionClient(string dceEndpoint) : base(new Uri(dceEndpoint), new DefaultAzureCredential()) { }
+
+        public override Task<Response> UploadAsync<T>(string ruleId, string streamName, IEnumerable<T> logs, LogsUploadOptions options = null, CancellationToken cancellationToken = default)
```
Aniruddh25: where is async specified here?

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
```diff
@@ -0,0 +1,139 @@
...
+        // Arrange
+        AzureLogAnalyticsOptions azureLogAnalyticsOptions = new(true, new AzureLogAnalyticsAuthOptions("custom-table-name-test", "dcr-immutable-id-test", "https://fake.dce.endpoint"));
+        CustomLogsIngestionClient customClient = new(azureLogAnalyticsOptions.Auth.DceEndpoint);
+        AzureLogAnalyticsCustomLogCollector customLogCollector = new();
+
+        AzureLogAnalyticsFlusherService flusherService = new(azureLogAnalyticsOptions, customLogCollector, customClient);
+
+        // Act
+        await customLogCollector.LogAsync(message, logLevel);
+        _ = Task.Run(() => flusherService.StartAsync());
```
Aniruddh25: Dont you need to wait for an interval before reliably testing the customClient logs since the flusherservice runs StartAsync asynchronously?

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
RubenCerna2079: Yes, this will also be included in the Enhancement.

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
RubenCerna2079: The function comes from the ILogger interface, and needs to be implemented with all its parameters even if they are not used.

src/Service/Startup.cs
RubenCerna2079: They run it two different functions that both are started at runtime, so each of them needs to check if it is enabled independently.

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
RubenCerna2079: Added it for this PR, will keep in mind for future PRs

src/Config/ObjectModel/AzureLogAnalyticsLogs.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.Logging;
+
+namespace Azure.DataApiBuilder.Config.ObjectModel;
```
souvikghosh04, line 4: nit: I think we can maintain the block-scope namespace style like we have across the code base instead of file-scoped

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.DataApiBuilder.Service.Telemetry;
```
souvikghosh04, line 12: nit: same comment as previous. stick to block-style namespace style for uniformity

src/Service/Telemetry/AzureLogAnalyticsFlusherService.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.Monitor.Ingestion;
+
+namespace Azure.DataApiBuilder.Service.Telemetry;
```
souvikghosh04, line 13: nit: same as before

src/Service/Startup.cs
```diff
@@ -29,6 +29,8 @@
 using Azure.DataApiBuilder.Service.HealthCheck;
 using Azure.DataApiBuilder.Service.Telemetry;
 using Azure.DataApiBuilder.Service.Utilities;
+using Azure.Identity;
+using Azure.Monitor.Ingestion;
```
souvikghosh04, line 33: nit: check the ordering or it might fail in build pipeline

src/Service/Startup.cs
souvikghosh04: this condition can be simplified as 
`if (runtimeConfigAvailable 
    && runtimeConfig?.Runtime?.Telemetry?.AzureLogAnalytics is { Enabled: true })`

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#property-pattern

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// <summary>
+    /// Custom logs ingestion to test that all the logs are being sent correctly to Azure Log Analytics
+    /// </summary>
+    private class CustomLogsIngestionClient : LogsIngestionClient
+    {
+        public List<AzureLogAnalyticsLogs> LogAnalyticsLogs { get; } = new();
+
+        public CustomLogsIngestionClient(string dceEndpoint) : base(new Uri(dceEndpoint), new DefaultAzureCredential()) { }
```
souvikghosh04: please note: this might come as a new S360 Code QL item. add this after a space in the same line- `// CodeQL [SM05137] DefaultAzureCredential will use Managed Identity if available or fallback to default.` you can find the same in other places in code base

src/Service/Startup.cs
RubenCerna2079: Talked with @JerryNixon and we came to the conclusion that dab should stop if Log Analytics is configured incorrectly when it is first started. I will change this so it returns an error message.

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
RubenCerna2079: Same reason as the previous comment, we do not use eventId but it is necessary to add it as a parameter in the Log function that comes from the ILog interface.

src/Service/Startup.cs
RubenCerna2079: The configuration of the telemetry for Azure Log Analytics occurs before the validation of the configuration file. Which is why I decided to put these checks in place, to return from the function before the `FlusherService` is ran.

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
RubenCerna2079: We pass the value from the `Log` function inside the `AzureLogAnalyticsLogger` class, so it is a required property that shows which of the classes inside of DAB is the one that sent the specific log.

src/Service/Program.cs
RubenCerna2079: The category parameter inside of the `AddFilter` function specifies which classes to filter, which is why we want it to be empty. It still needs to be added as it is not empty or null by default.

src/Service/Startup.cs
RubenCerna2079: The logic is already inside of a function, no-op is required.

src/Service/Startup.cs
RubenCerna2079: Changed it so that it adds all of the necessary parameters with the singleton and then create a hosted service which will be running in the background, so there is no need call StartAsync as it does it automatically.

src/Cli/Commands/ConfigureOptions.cs
```diff
@@ -184,7 +184,7 @@ public ConfigureOptions(
         [Option("azure-key-vault.retry-policy.network-timeout-seconds", Required = false, HelpText = "Configure the network timeout for requests in seconds. Default: 60.")]
         public int? AzureKeyVaultRetryPolicyNetworkTimeoutSeconds { get; }
 
-        [Option("runtime.telemetry.azure-log-analytics.enabled", Default = CliBool.False, Required = false, HelpText = "Enable/Disable Azure Log Analytics.")]
```
Aniruddh25, line 187: How do you convey the default value in the HelpText?

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
RubenCerna2079: I don't think so since the flusherService starts after the customLogCollector logs the message to see if it is saved correctly.

src/Service/Startup.cs
```diff
@@ -179,11 +178,18 @@ public void ConfigureServices(IServiceCollection services)
...
-                && runtimeConfig.Runtime.Telemetry.AzureLogAnalytics.Enabled)
+                && IsAzureLogAnalyticsAvailable(runtimeConfig.Runtime.Telemetry.AzureLogAnalytics))
             {
                 services.AddSingleton<ICustomLogCollector, AzureLogAnalyticsCustomLogCollector>();
                 services.AddSingleton<ILoggerProvider, AzureLogAnalyticsLoggerProvider>();
-                services.AddSingleton<AzureLogAnalyticsFlusherService>();
+                services.AddSingleton(sp =>
+                {
+                    AzureLogAnalyticsOptions options = runtimeConfig.Runtime.Telemetry.AzureLogAnalytics;
+                    DefaultAzureCredential credential = new();
```
Aniruddh25, line 188: Please explain in the description how you tested the feature - what setup did you have to do to ensure the authentication works?

src/Service/Telemetry/AzureLogAnalyticsFlusherService.cs
```diff
@@ -3,32 +3,40 @@
...
-    // Function that will keep periodically flushing data logs as long as Azure Log Analytics is enabled
-    public async Task StartAsync()
+    /// <summary>
+    /// Function that will keep periodically flushing data logs as long as Azure Log Analytics is enabled.
+    /// </summary>
+    /// <param name="stoppingToken">Token used to stop running service when program is shut down.</param>
+    protected async override Task ExecuteAsync(CancellationToken stoppingToken)
     {
-        while (_options.Enabled)
+        while (_options?.Enabled is true && _logsIngestionClient is not null && _customLogCollector is not null && _logger is not null)
```
Aniruddh25: Wouldnt all these conditions be always true - given that we are checking for IsAzureLogAnalyticsIsAvailable() before creating this HostedService?

Why do we need to check for them again? Can we simply do while(true)

src/Service/Program.cs
```diff
@@ -195,6 +195,20 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, Tele
...
+                        }
+                    }
```
Aniruddh25, line 210: How is the `category` argument specified here in `AddFilter` different from the `categoryName` you pass to the constructor: AzureLogAnalyticsLogger()?

src/Service/Telemetry/AzureLogAnalyticsLoggerProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+    private readonly ICustomLogCollector _customLogCollector;
+
+    public AzureLogAnalyticsLoggerProvider(ICustomLogCollector customLogCollector)
+    {
+        _customLogCollector = customLogCollector;
+    }
+
+    public ILogger CreateLogger(string categoryName)
+    {
+        return new AzureLogAnalyticsLogger(categoryName, _customLogCollector);
```
Aniruddh25: Please rename this argument from categoryName to className since it is confusing with the category argument in the AddFilter function.

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
Aniruddh25: If its not specified, the function name should change to `Upload`

src/Service/Startup.cs
Aniruddh25: The function `ExecuteAsync` of the FlusherService will start only if the service is created. You check if LogAnalytics is enabled in before creating the service. So, would you still need the checks in the ExecuteAsync function.

Aniruddh25 (changes requested):
Waiting on fixing test failures and the need to double check if LogAnalytics is enabled or not.

src/Cli/Commands/ConfigureOptions.cs
RubenCerna2079: Do you mean to state in the HelpText which is the default value?

src/Service/Telemetry/AzureLogAnalyticsFlusherService.cs
RubenCerna2079: Yes, I was testing something and forgot to remove them, will be changing to while(true)

src/Service/Startup.cs
RubenCerna2079: The check would not be needed anymore inside the ExecuteAsync function, will be changing it.

src/Service/Program.cs
RubenCerna2079: AzureLogAnalyticsLoggerProvider creates a logger with the constructor AzureLogAnalyticsLogger for each of the services that we add in startup, the `categoryName` is the name of the class for which that logger is created.
On the other hand, the `category` argument `AddFilter` would be used to if we only want to filter one specific service, while we want all services to have the same log-level filtering. Since there is already an implementation to filter specific classes.

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
```diff
@@ -68,25 +68,55 @@ public void CleanUpTelemetryConfig()
...
+        };
+        using TestServer server = new(Program.CreateWebHostBuilder(args));
+
+        // Additional assertions to check if AzureLogAnalytics is enabled correctly in services
+        IServiceProvider serviceProvider = server.Services;
+        AzureLogAnalyticsCustomLogCollector customLogCollector = serviceProvider.GetService<AzureLogAnalyticsCustomLogCollector>();
+        AzureLogAnalyticsLoggerProvider loggerProvider = serviceProvider.GetService<AzureLogAnalyticsLoggerProvider>();
+        AzureLogAnalyticsFlusherService flusherService = serviceProvider.GetService<AzureLogAnalyticsFlusherService>();
+
+        // If tracerProvider and meterProvider are not null, AzureLogAnalytics is enabled
```
Aniruddh25: what are tracerProvider, meterProvider?

souvikghosh04 (commented):
Reviewed he changes.

src/Config/ObjectModel/AzureLogAnalyticsLogs.cs
RubenCerna2079: There are many examples in the code base where we use file-scoped namespace instead of block-scope. It might be a good idea to bring this up in our meetings, since I think we should only be using one style for our code base.

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
RubenCerna2079: Comment was wrong, it was meant to be the variables that save the services

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    /// <param name="logType">Custom name to distinguish the logs sent from DAB to Azure Log Analytics.</param>
+    /// <param name="flushIntervalSeconds">Period of time between each list of logs is sent.</param>
+    /// <returns>List of logs structured to be sent to Azure Log Analytics.</returns>
+    public async Task<List<AzureLogAnalyticsLogs>> DequeueAllAsync(string logType, int flushIntervalSeconds)
+    {
+        List<AzureLogAnalyticsLogs> list = new();
+        Stopwatch time = Stopwatch.StartNew();
+
+        while (time.Elapsed < TimeSpan.FromSeconds(flushIntervalSeconds))
```
RubenCerna2079: Changed way logic work, because the function `.ReadAllAsync()` would keep waiting for new logs even if there were none, which could cause a delay from when the logs should be sent.

souvikghosh04 (approved):
LGTM

PR #2788, opened 2025-07-28
Support for relative nextLink using `next-link-relative` property for Reverse Proxy Scenarios

### LINKED ISSUE #2677: [Bug]: Pagination is broken if DAB runs behind a reverse proxy
### What happened?

If running DAB behind a reverse proxy that handled SSL termination (https -> http) the `next` link is broken as it will use `http` schema instead of `https`. 
I think the issue is in `src/Core/Resolvers/SqlResponseHelpers.cs` as the link should be constructed either:
1. as a relative link
2. as a schema-less link
3. use the `x-forwarded-host` and `x-forwarded-proto` if theyare available

### Version

1.4.26

PostgreSQL

Custom Docker host

REST

### Relevant log output

```Text

```

### Code of Conduct

souvikghosh04:
## Why make this change?

- Closes #2677  
When Data API builder (DAB) is deployed behind a reverse proxy that handles SSL termination (converting https to http), the pagination nextLink URL in REST responses was generated with the http scheme instead of https. This resulted in broken next page links for clients accessing the service securely. The reported bug highlighted that the nextLink should use the original client-facing scheme and host, or provide a relative/schema-less link, or leverage the x-forwarded-proto and x-forwarded-host headers set by the reverse proxy.

## What is this change?

  - Users can also optionally set `x-forwarded-proto` and `x-forwarded-host headers` as needed to match their deployment environment. If not set, it will pick the default path from the request context.
  - Also refer [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) and [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-Host).

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

## Sample Request(s)

- GET `http://localhost:5000/api/<your_entity>`

### DIFF
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -442,9 +442,9 @@
                     }
                   }
                 },
-                "log-type": {
+                "dab-identifier": {
                   "type": "string",
-                  "description": "Custom log table name in Log Analytics",
+                  "description": "Identifier passed on to Log Analytics",
                   "default": "DabLogs"
                 },
                 "flush-interval-seconds": {
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -4080,22 +4080,22 @@ type Planet @model(name:""PlanetAlias"") {
             string? customTableName,
             string? dcrImmutableId,
             string? dceEndpoint,
-            string? logType,
+            string? dabIdentifier,
             int? flushIntSec,
             bool expectedEnabled,
-            string expectedLogType,
+            string expectedDabIdentifier,
             int expectedFlushIntSec)
         {
             //Check if auth property and its values are expected to exist
             bool expectedExistEnabled = enabled is not null;
-            bool expectedExistLogType = logType is not null;
+            bool expectedExistDabIdentifier = dabIdentifier is not null;
             bool expectedExistFlushIntSec = flushIntSec is not null;
             bool expectedExistCustomTableName = customTableName is not null;
             bool expectedExistDcrImmutableId = dcrImmutableId is not null;
             bool expectedExistDceEndpoint = dceEndpoint is not null;
 
             AzureLogAnalyticsAuthOptions authOptions = new(customTableName, dcrImmutableId, dceEndpoint);
-            AzureLogAnalyticsOptions azureLogAnalyticsOptions = new(enabled, authOptions, logType, flushIntSec);
+            AzureLogAnalyticsOptions azureLogAnalyticsOptions = new(enabled, authOptions, dabIdentifier, flushIntSec);
             RuntimeConfig configWithCustomLogLevel = InitializeRuntimeWithAzureLogAnalytics(azureLogAnalyticsOptions);
             string configWithCustomLogLevelJson = configWithCustomLogLevel.ToJson();
             Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(configWithCustomLogLevelJson, out RuntimeConfig? deserializedRuntimeConfig));
@@ -4120,11 +4120,11 @@ type Planet @model(name:""PlanetAlias"") {
                     Assert.AreEqual(expectedEnabled, enabledElement.GetBoolean());
                 }
 
-                bool logTypeExists = azureLogAnalyticsElement.TryGetProperty("log-type", out JsonElement logTypeElement);
-                Assert.AreEqual(expected: expectedExistLogType, actual: logTypeExists);
-                if (logTypeExists)
+                bool dabIdentifierExists = azureLogAnalyticsElement.TryGetProperty("dab-identifier", out JsonElement dabIdentifierElement);
+                Assert.AreEqual(expected: expectedExistDabIdentifier, actual: dabIdentifierExists);
+                if (dabIdentifierExists)
                 {
-                    Assert.AreEqual(expectedLogType, logTypeElement.GetString());
+                    Assert.AreEqual(expectedDabIdentifier, dabIdentifierElement.GetString());
                 }
 
                 bool flushIntSecExists = azureLogAnalyticsElement.TryGetProperty("flush-interval-seconds", out JsonElement flushIntSecElement);
--- a/src/Cli.Tests/ConfigureOptionsTests.cs
+++ b/src/Cli.Tests/ConfigureOptionsTests.cs
@@ -162,7 +162,7 @@ namespace Cli.Tests
             // Act: Attempts to add Azure Log Analytics options
             ConfigureOptions options = new(
                 azureLogAnalyticsEnabled: CliBool.True,
-                azureLogAnalyticsLogType: "log-type-test",
+                azureLogAnalyticsDabIdentifier: "dab-identifier-test",
                 azureLogAnalyticsFlushIntervalSeconds: 1,
                 azureLogAnalyticsCustomTableName: "custom-table-name-test",
                 azureLogAnalyticsDcrImmutableId: "dcr-immutable-id-test",
@@ -180,7 +180,7 @@ namespace Cli.Tests
             Assert.IsNotNull(config.Runtime.Telemetry);
             Assert.IsNotNull(config.Runtime.Telemetry.AzureLogAnalytics);
             Assert.AreEqual(true, config.Runtime.Telemetry.AzureLogAnalytics.Enabled);
-            Assert.AreEqual("log-type-test", config.Runtime.Telemetry.AzureLogAnalytics.LogType);
+            Assert.AreEqual("dab-identifier-test", config.Runtime.Telemetry.AzureLogAnalytics.DabIdentifier);
             Assert.AreEqual(1, config.Runtime.Telemetry.AzureLogAnalytics.FlushIntervalSeconds);
             Assert.IsNotNull(config.Runtime.Telemetry.AzureLogAnalytics.Auth);
             Assert.AreEqual("custom-table-name-test", config.Runtime.Telemetry.AzureLogAnalytics.Auth.CustomTableName);
--- a/src/Cli.Tests/ValidateConfigTests.cs
+++ b/src/Cli.Tests/ValidateConfigTests.cs
@@ -332,7 +332,7 @@ public class ValidateConfigTests
         // Act: Attempts to add Azure Log Analytics options without Auth options
         ConfigureOptions options = new(
             azureLogAnalyticsEnabled: CliBool.True,
-            azureLogAnalyticsLogType: "log-type-test",
+            azureLogAnalyticsDabIdentifier: "dab-identifier-test",
             azureLogAnalyticsFlushIntervalSeconds: 1,
             config: TEST_RUNTIME_CONFIG_FILE
         );
--- a/src/Cli/Commands/ConfigureOptions.cs
+++ b/src/Cli/Commands/ConfigureOptions.cs
@@ -49,7 +49,7 @@ namespace Cli.Commands
             int? azureKeyVaultRetryPolicyMaxDelaySeconds = null,
             int? azureKeyVaultRetryPolicyNetworkTimeoutSeconds = null,
             CliBool? azureLogAnalyticsEnabled = null,
-            string? azureLogAnalyticsLogType = null,
+            string? azureLogAnalyticsDabIdentifier = null,
             int? azureLogAnalyticsFlushIntervalSeconds = null,
             string? azureLogAnalyticsCustomTableName = null,
             string? azureLogAnalyticsDcrImmutableId = null,
@@ -93,7 +93,7 @@ namespace Cli.Commands
             AzureKeyVaultRetryPolicyNetworkTimeoutSeconds = azureKeyVaultRetryPolicyNetworkTimeoutSeconds;
             // Azure Log Analytics
             AzureLogAnalyticsEnabled = azureLogAnalyticsEnabled;
-            AzureLogAnalyticsLogType = azureLogAnalyticsLogType;
+            AzureLogAnalyticsDabIdentifier = azureLogAnalyticsDabIdentifier;
             AzureLogAnalyticsFlushIntervalSeconds = azureLogAnalyticsFlushIntervalSeconds;
             AzureLogAnalyticsCustomTableName = azureLogAnalyticsCustomTableName;
             AzureLogAnalyticsDcrImmutableId = azureLogAnalyticsDcrImmutableId;
@@ -187,8 +187,8 @@ namespace Cli.Commands
         [Option("runtime.telemetry.azure-log-analytics.enabled", Required = false, HelpText = "Enable/Disable Azure Log Analytics. Default: False (boolean)")]
         public CliBool? AzureLogAnalyticsEnabled { get; }
 
-        [Option("runtime.telemetry.azure-log-analytics.log-type", Required = false, HelpText = "Configure Log Type for Azure Log Analytics to find table to send telemetry data. Default: DABLogs")]
-        public string? AzureLogAnalyticsLogType { get; }
+        [Option("runtime.telemetry.azure-log-analytics.dab-identifier", Required = false, HelpText = "Configure DAB Identifier to allow user to differentiate which logs come from DAB in Azure Log Analytics . Default: DABLogs")]
+        public string? AzureLogAnalyticsDabIdentifier { get; }
 
         [Option("runtime.telemetry.azure-log-analytics.flush-interval-seconds", Required = false, HelpText = "Configure Flush Interval in seconds for Azure Log Analytics to specify the time interval to send the telemetry data. Default: 5")]
         public int? AzureLogAnalyticsFlushIntervalSeconds { get; }
--- a/src/Cli/ConfigGenerator.cs
+++ b/src/Cli/ConfigGenerator.cs
@@ -780,7 +780,7 @@ namespace Cli
 
             // Telemetry: Azure Log Analytics
             if (options.AzureLogAnalyticsEnabled is not null ||
-                options.AzureLogAnalyticsLogType is not null ||
+                options.AzureLogAnalyticsDabIdentifier is not null ||
                 options.AzureLogAnalyticsFlushIntervalSeconds is not null ||
                 options.AzureLogAnalyticsCustomTableName is not null ||
                 options.AzureLogAnalyticsDcrImmutableId is not null ||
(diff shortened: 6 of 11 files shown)


src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -579,42 +579,46 @@ public static string Base64Decode(string base64EncodedData)
...
+            // Format existing query string (URL encoded)
+            string queryString = FormatQueryString(queryStringParameters);
 
-            // When a new $after payload is provided, append it to the query string with the
-            // appropriate prefix: ? if $after is the only query parameter. & if $after is one of many query parameters.
+            // Append new $after token
             if (!string.IsNullOrWhiteSpace(newAfterPayload))
             {
-                string afterPrefix = string.IsNullOrWhiteSpace(queryString) ? "?" : "&";
+                string afterPrefix = string.IsNullOrWhiteSpace(queryString) ? "" : "&";
```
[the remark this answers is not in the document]
souvikghosh04: Fixed

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -425,5 +412,129 @@ public static OkObjectResult OkMutationResponse(JsonElement jsonResult)
...
+
+            // If the base route is not empty, we need to insert it into the URI before the rest path.
+            // Path is of the form ....restPath/pathNameForEntity. We want to insert the base route before the restPath.
+            // Finally, it will be of the form: .../baseRoute/restPath/pathNameForEntity.
+            return UriHelper.BuildAbsolute(
+                scheme: scheme,
+                host: new HostString(host),
+                pathBase: string.IsNullOrWhiteSpace(baseRoute) ? PathString.Empty : new PathString(baseRoute),
+                path: req.Path,
+                query: req.QueryString);
```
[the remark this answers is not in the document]
souvikghosh04: Fixed

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -579,42 +579,46 @@ public static string Base64Decode(string base64EncodedData)
...
             {
                 queryStringParameters = new();
             }
             else
             {
-                // Purge old $after value so this function can replace it.
                 queryStringParameters.Remove("$after");
             }
 
-            // To prevent regression of current behavior, retain the call to FormatQueryString
```
RubenCerna2079, line 595: Is there a reason to simplify these comments so much?

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -425,5 +412,128 @@ public static OkObjectResult OkMutationResponse(JsonElement jsonResult)
...
+        /// </summary>
+        /// <param name="req">The HTTP request.</param>
+        /// <returns>Validated scheme string ("http" or "https").</returns>
+        /// <exception cref="DataApiBuilderException">Thrown when client explicitly sets an invalid scheme.</exception>
+        private static string GetValidatedScheme(HttpRequest req)
+        {
+            string? rawScheme = req.Headers["X-Forwarded-Proto"].FirstOrDefault();
+            string? normalized = rawScheme?.Trim().ToLowerInvariant();
+
+            bool isExplicit = !string.IsNullOrEmpty(rawScheme);
```
RubenCerna2079: nit: I think the name of the variable is a little confusing, maybe `isHeaderExplicit` would be better.

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -425,5 +412,128 @@ public static OkObjectResult OkMutationResponse(JsonElement jsonResult)
...
+        /// </summary>
+        /// <param name="req">The HTTP request.</param>
+        /// <returns>Validated host string.</returns>
+        /// <exception cref="DataApiBuilderException">Thrown when client explicitly sets an invalid host.</exception>
+        private static string GetValidatedHost(HttpRequest req)
+        {
+            string? rawHost = req.Headers["X-Forwarded-Host"].FirstOrDefault();
+            string? trimmed = rawHost?.Trim();
+
+            bool isExplicit = !string.IsNullOrEmpty(rawHost);
```
RubenCerna2079: Same as the previous comment

schemas/dab.draft.schema.json
```diff
@@ -168,6 +168,11 @@
               "description": "Sets the default number of records returned in a single response. When this limit is reached, a continuation token is provided to retrieve the next page. If set to null, the default value is 100.",
               "default": 100,
               "minimum": 1
+            },
+            "next-link-relative": {
```
RubenCerna2079, line 172: I see the descriptions says that this property can be configured through the `configure` CLI command. But I do not see any changes to the `ConfigureOptions`. Am I missing something?

souvikghosh04: planned in different PR

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: will address in next PR

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: will take in next PR

src/Core/Resolvers/SqlPaginationUtil.cs
souvikghosh04: as discussed, will take in next PR.

schemas/dab.draft.schema.json
```diff
@@ -168,6 +168,11 @@
               "description": "Sets the default number of records returned in a single response. When this limit is reached, a continuation token is provided to retrieve the next page. If set to null, the default value is 100.",
```
Aniruddh25, line 168: Description says:

`When Data API Builder (DAB) is deployed`

-> Builder should be lower cased as in -> `builder`

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -592,6 +592,11 @@ public uint MaxPageSize()
         return (uint?)Runtime?.Pagination?.MaxPageSize ?? PaginationOptions.MAX_PAGE_SIZE;
     }
 
+    public bool NextLinkRelative()
```
Aniruddh25, line 595: Please fix PR title to reflect the support added to nextLink being relative using `next-link-relative` property

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -425,5 +412,128 @@ public static OkObjectResult OkMutationResponse(JsonElement jsonResult)
...
+        {
+            HttpRequest req = httpContext.Request;
+
+            string scheme = GetValidatedScheme(req);
+            string host = GetValidatedHost(req);
+
+            // If the base route is not empty, we need to insert it into the URI before the rest path.
+            // Path is of the form ....restPath/pathNameForEntity. We want to insert the base route before the restPath.
+            // Finally, it will be of the form: .../baseRoute/restPath/pathNameForEntity.
+            return UriHelper.BuildAbsolute(
```
Aniruddh25: If we eventually just need a relative path (determined by the nextLinkRelative property), why go through the trouble to create the Absolute URI?

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -91,27 +91,14 @@ public static OkObjectResult FormatFindResult(
...
-
-                // Path is of the form ....restPath/pathNameForEntity. We want to insert the base route before the restPath.
-                // Finally, it will be of the form: .../baseRoute/restPath/pathNameForEntity.
-                path = UriHelper.BuildAbsolute(
-                    scheme: request.Scheme,
-                    host: request.Host,
-                    pathBase: baseRoute,
-                    path: request.Path);
-            }
+            string path = GetPathForNextLink(httpContext, runtimeConfig.Runtime?.BaseRoute);
```
Aniruddh25: Can we create the `path` as Absolute or relative right here? by checking for runtimeConfig.NextLinkRelative()

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -579,42 +579,46 @@ public static string Base64Decode(string base64EncodedData)
...
             {
                 string afterPrefix = string.IsNullOrWhiteSpace(queryString) ? "?" : "&";
                 queryString += $"{afterPrefix}{RequestParser.AFTER_URL}={newAfterPayload}";
             }
 
-            // ValueKind will be array so we can differentiate from other objects in the response
-            // to be returned.
-            // [{"nextLink":"[base]/api/[entity]?[queryParams_URIescaped]$after=[base64encodedPaginationToken]"}]
+            // Create UriBuilder from full URL
+            UriBuilder uriBuilder = new(path)
```
Aniruddh25: if we create a relative path even before calling this function, we may not need to build the uri here. Can we simply append the queryString to the path in that case?

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -425,5 +412,128 @@ public static OkObjectResult OkMutationResponse(JsonElement jsonResult)
...
+        /// <exception cref="DataApiBuilderException">Thrown when client explicitly sets an invalid host.</exception>
+        private static string GetValidatedHost(HttpRequest req)
+        {
+            string? rawHost = req.Headers["X-Forwarded-Host"].FirstOrDefault();
+            string? trimmed = rawHost?.Trim();
+
+            bool isExplicit = !string.IsNullOrEmpty(rawHost);
+            bool isValid = IsValidHost(trimmed);
+
+            if (isExplicit && !isValid)
```
Aniruddh25: This will now start failing clients which were sending an invalid value for `X-Forwarded-Host` header.. can there be legitimate cases where these header values are invalid?

If we dont know, should we simply log a warning and ignore these header warnings instead of throwing an exception? Do we die if we throw here?

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -2397,12 +2406,13 @@ public void ValidatePaginationOptionsInConfig(
...
                         GraphQL: new(),
                         Host: new(Cors: null, Authentication: null),
-                        Pagination: new PaginationOptions(defaultPageSize, maxPageSize)
+                        Pagination: new PaginationOptions(defaultPageSize, maxPageSize, nextLinkRelative)
                     ),
                     Entities: new(new Dictionary<string, Entity>()));
 
                 Assert.AreEqual((uint)expectedDefaultPageSize, runtimeConfig.DefaultPageSize());
                 Assert.AreEqual((uint)expectedMaxPageSize, runtimeConfig.MaxPageSize());
+                Assert.AreEqual(expected: nextLinkRelative ?? false, actual: runtimeConfig.NextLinkRelative());
```
Aniruddh25, line 2415: This only validates if the config value is accurate or not. Can we have a test which validates the effect of the config property - on a graphql pagination request? Expect a relative url in the response when the value of the property is true, false otherwise.

Aniruddh25 (changes requested):
Needs an automated test to validate nextLink is indeed relative in the response if the property is set to true.

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -2397,12 +2406,13 @@ public void ValidatePaginationOptionsInConfig(
...
             }
             catch (DataApiBuilderException dabException)
             {
```
Aniruddh25, line 2418: Should also have automated tests for the headers - X-Forwarded-Host, X-Forwarded-Port

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: yes, I think instead of failing here, we can log it. invalid values in these headers would fail to produce the nextLink. We should have this behavior documented.

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
souvikghosh04: It seems for GraphQL, the next link is generated in the form of a Base64 encoded value called `endCursor`. It does not have the domain name or API path in that. Since this PR is specifically for the REST path and `nextLink` is only generated for this scenario, GraphQL remains intact. Existing GraphQL tests should be covering scenarios for that and unrelated to this change. Please let me know if my understanding is wrong. sample GraphQL response with page limit of 2 
image

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: This is before we take out the relative path from the absolute Uri. `GetPathForNextLink()` covers for both absolute and relative but we don't get the relative path here. we do it in `CreateNextLink`. In case of relative, we get the next link as `uriBuilder.Uri.PathAndQuery`

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: I have kept the logic for absolute/relative path generic and separated it in `CreateNextLink()`

src/Core/Resolvers/SqlPaginationUtil.cs
souvikghosh04: we only create relative path out of the Uri when `runtimeConfig.NextLinkRelative()` passed as true. that's why we don't create relative path before this. this is the only place it resolves conditionally

src/Core/Resolvers/SqlResponseHelpers.cs
Aniruddh25: Keeping it generic has us do more work than we need. Validating the host and scheme is not needed if we need to provide a relative link.

I see structure this as common code needed for both relative and absolute URI. 

```
-CreateRelativePath
-GetQuery

if (!nextLinkRelative)
     - ValidateHost
     - ValidateScheme
     - GenerateFullUri
else
    return path+query
     
```

src/Core/Resolvers/SqlResponseHelpers.cs
Aniruddh25: I see that, but for relative path doing all the following is unnecessary since we already have the `path` and `query` in the caller `FormatFindResult`.

- GetValidatedScheme
- GetValidatedHost
- Build Absolute Uri

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
Aniruddh25: We should have an automated test for at least REST in that case.

souvikghosh04: yes, I have updated the test to cover this additional scenario in `ValidateNextLinkUsage()` inside `ConfigurationTests.cs`.

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: Made the suggested change by logging it instead of returning false and failing.

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
souvikghosh04: yes, I have updated the existing test- `ValidateNextLinkUsage()` to cover this scenario.

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: This is fixed as per suggesstion

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
souvikghosh04: I have added a new test- `ValidateNextLinkRespectsXForwardedHostAndProto()` to cover this scenario.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4415,16 +4417,30 @@ public async Task ValidateNextLinkUsage()
                 Mappings: null);
 
             Dictionary<string, Entity> entityMap = new()
+    {
+        { ENTITY_NAME, requiredEntity }
+    };
```
RubenCerna2079:
```suggestion
            {
                { ENTITY_NAME, requiredEntity }
            };
```

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4415,16 +4417,30 @@ public async Task ValidateNextLinkUsage()
...
 
-            CreateCustomConfigFile(entityMap, enableGlobalRest: true);
+            CreateCustomConfigFile(entityMap, enableGlobalRest: true, paginationOptions: paginationOptions);
 
             string[] args = new[]
             {
-                $"--ConfigFileName={CUSTOM_CONFIG_FILENAME}"
-            };
+        $"--ConfigFileName={CUSTOM_CONFIG_FILENAME}"
+    };
```
RubenCerna2079:
```suggestion
            $"--ConfigFileName={CUSTOM_CONFIG_FILENAME}"
        };
```

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: I have refactored this now.

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: Refactored this.

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -572,16 +574,50 @@ public static string Base64Decode(string base64EncodedData)
         }
 
         /// <summary>
-        /// Create the URL that will provide for the next page of results
-        /// using the same query options.
-        /// Return value formatted as a JSON array: [{"nextLink":"[base]/api/[entity]?[queryParams_URIescaped]$after=[base64encodedPaginationToken]"}]
+        /// Constructs
```
Aniruddh25: nit: Incomplete summary

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -699,5 +749,94 @@ public static string FormatQueryString(NameValueCollection? queryStringParameter
...
+            string? rawHost = req.Headers["X-Forwarded-Host"].FirstOrDefault();
+            string? trimmed = rawHost?.Trim();
+
+            bool isExplicit = !string.IsNullOrEmpty(rawHost);
+            bool isValid = IsValidHost(trimmed);
+
+            if (isExplicit && !isValid)
+            {
+                // Log a warning and ignore the invalid value, fallback to request's host
+                Console.WriteLine($"Warning: Invalid host '{rawHost}' in X-Forwarded-Host header. Falling back to request host: '{req.Host}'.");
```
Aniruddh25, line 794: Shouldnt this use the logger.LogWarning instead of Console.WriteLine

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4467,6 +4499,114 @@ public async Task ValidateNextLinkUsage()
             {
                 Assert.Fail(message: "$after query parameter was not a valid base64 encoded value.");
             }
+
+            // Validate nextLink is relative if nextLinkRelative is true or false otherwise.
+            // The assertion is now done directly on the original string, not on the parsed Uri object.
+            if (isNextLinkRelative)
```
Aniruddh25, line 4505: This validation should happen before we attempt to get the results using the nextLink. So, it should happen before line 4475 before we set the value of nextLink. In fact, we should do the assert then proceed to set the nextLink.

souvikghosh04: this order is similar to how our actual code path works- form Uri, form queryString (base64 after value), form the final link.
so lines-
4475 - 4487: validates the base Uri
4489 - 4501: validates the queryString (after)
4505 - 4519: validates the final link- absolute/relative

src/Core/Resolvers/SqlPaginationUtil.cs
souvikghosh04: yes, but we don't have the ILogger injected here. we need additional changes from callers till here. I will open a work item for this.

src/Service.Tests/Configuration/ConfigurationTests.cs
Aniruddh25: I understand, however, if the nextLink is invalid,  the `queryString` that we retrieve from it will be invalid as well.. So, IMO we should first check whether the nextLink is valid since queryString is a derivative of the nextLink. Anyway, not going to block the PR for this.

Aniruddh25 (approved):

Issue #2804, opened 2025-08-05
Change `log-type` property name to `dab-identifier`


RubenCerna2079:
The function of `log-type` changed from being the name of the Azure Log Analytics workspace table, to an identifier that will be sent as part of the logs to allow the user to identify which logs are sent from DAB.

PR #2805, opened 2025-08-06
Change property name of log-type to dab-identifier for Azure Log Analytics

RubenCerna2079:
## Why make this change?
- This change closes issue #2804 
The property `log-type` had a change in function, which also means that there needs to be a change in name.

## What is this change?
Only changes the name of the property from `log-type` to `dab-identifier` since the changes in logic related to the changes in function were already implemented beforehand.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests
- [X] Manual Tests
Tested if the new property is serialized and deserialized correctly and if `configure` command works as expected with new property, while leaving no traces of the old property.

## Sample Request(s)
--runtime.telemetry.azure-log-analytics.dab-identifier

### DIFF
--- a/src/Cli.Tests/ValidateConfigTests.cs
+++ b/src/Cli.Tests/ValidateConfigTests.cs
@@ -332,7 +332,7 @@ public class ValidateConfigTests
         // Act: Attempts to add Azure Log Analytics options without Auth options
         ConfigureOptions options = new(
             azureLogAnalyticsEnabled: CliBool.True,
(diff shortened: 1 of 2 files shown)


anushakolan:
LGTM! But can we run any existing tests related to the changes to see noting is broken, since you tested this manually?

RubenCerna2079:
> LGTM! But can we run any existing tests related to the changes to see noting is broken, since you tested this manually?

The tests in the pipeline that were changed should be sufficient.

Issue #2810, opened 2025-08-07
Change DefaultAzureCredential to ManagedIdentityCredential for Azure Log Analytics


RubenCerna2079:
The `ManagedIdentityCredential` class is more suitable to get the necessary credentials to connect to Azure Log Analytics than `DefaultAzureCredential`

PR #2811, opened 2025-08-07
Change use of DefaultAzureCredential to ManagedIdentityCredential for Azure Log Analytics 

RubenCerna2079:
## Why make this change?
- It resolves issue #2810 
`ManagedIdentityCredential` class only authenticates with Azure Managed Identity which makes it a better fit to use in AzureLogAnalytics than `DefaultAzureCredential`.

## What is this change?
We stop using the `DefaultAzureCredential` class and instead start using the `ManagedIdentityCredential` class. 
Important Note: As of this moment with this change, the Azure Log Analytics feature can only authenticate against system-assigned identities.

## How was this tested?
- [ ] Integration Tests
- [ ] Unit Tests
- [X] Manual Tests
This change cannot be directly tested in the pipeline so it was tested by using a VM and ensuring the logs are correctly sent to Azure Log Analytics table.
- See #2787 for more information on how this was set up.

## Sample Request(s)
image

### DIFF
--- a/src/Service/Startup.cs
+++ b/src/Service/Startup.cs
@@ -185,7 +185,7 @@ namespace Azure.DataApiBuilder.Service
                 services.AddSingleton(sp =>
                 {
                     AzureLogAnalyticsOptions options = runtimeConfig.Runtime.Telemetry.AzureLogAnalytics;
-                    DefaultAzureCredential credential = new();
+                    ManagedIdentityCredential credential = new();
                     LogsIngestionClient logsIngestionClient = new(new Uri(options.Auth!.DceEndpoint!), credential);
                     return new AzureLogAnalyticsFlusherService(options, CustomLogCollector, logsIngestionClient, _logger);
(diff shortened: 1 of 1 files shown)


src/Service/Startup.cs
```diff
@@ -185,7 +185,7 @@ public void ConfigureServices(IServiceCollection services)
                 services.AddSingleton(sp =>
                 {
                     AzureLogAnalyticsOptions options = runtimeConfig.Runtime.Telemetry.AzureLogAnalytics;
-                    DefaultAzureCredential credential = new();
+                    ManagedIdentityCredential credential = new();
```
Aniruddh25, line 188: Do we need to do a similar change where DefaultAzureCredential is used today?

RubenCerna2079: Yes we will need to do some similar changes inside the code base that are being suppressed at this moment.

Issue #2577, opened 2025-02-14
Add CLI Configure Commands for File Sink


RubenCerna2079:
Add new options for the `configure` command related to the `file` properties.
The properties are
- enabled: boolean (default: false)
- path: string (default: "/logs/dab-log.txt")
- rolling-interval: string (default: "Day")
- retained-file-count-limit: int (default: 1)
- file-size-limit-bytes: int (default: 1048576)

These new options should be:
--runtime.telemetry.file.enabled
--runtime.telemetry.file.path
--runtime.telemetry.file.rolling-interval
--runtime.telemetry.file.retained-file-count-limit
--runtime.telemetry.file.file-size-limit-bytes

Issue #2371, opened 2024-09-06
⭐ [Enhancement]: Introduce Logging Sink & Settings


JerryNixon:
## What is it

 * When running `dab start` the output logs are inconsistent and sometimes confusing. This effort would go through the output and remove unnecessary logs, while simplifying important logs. 
 * Add `log-level` to `application-insights`
 * Add a `file` sink
 * Add application-insight event: `dab-started` and `dab-start-failed`

## Log sink details

**Current syntax: [docs](https://learn.microsoft.com/en-us/azure/data-api-builder/deployment/how-to-use-application-insights?context=%2Fazure%2Fdata-api-builder%2Fcontext%2Fcontext)**

```json
{
  "runtime": {
    ...
    "telemetry": {
      "application-insights": {
        "enabled": true,
        "connection-string": "@env('app-insights-connection-string')"
      }
    },
    ...
  }
}
```

**Sample future configuration**

```json
{
  "runtime": {
    ...
    "telemetry": {
       "log-level": {
          "Default": "Warning",
          "Microsoft": "Error",
          "Microsoft.AspNetCore": "Information"
       },
      "application-insights": {
        "enabled": true,
        "connection-string": "@env('app-insights-connection-string')",
         "log-level": {
            "Default": "Warning",
            "Microsoft": "Error",
            "Microsoft.AspNetCore": "Information"
         }
      },
      "console": {
        "enabled": true
      },
      "file": {
        "enabled": true,
        "path": "/logs/dab-log.txt",
        "rollingInterval": "Day",
        "retainedFileCountLimit": 7,
        "fileSizeLimitBytes": 10485760
      }
    }
    ...
  }
}
```

#### Console sink

The console sink, which is currently assumed, will be explicitly listed so it can be enabled and configured individually.

#### Global `log-level`

The global log-level, which is similar to how global cache TTL works, can be inherited by a sink.

#### Move Global `log-level`

The current (unadvertised) global log-level will be moved from under `/runtime` to under `/runtime/telemetry`.

#### Hide "current" `log-level`

The current (unadvertised) global log-level will be removed from the JSON schema to prevent breaking changes later. 

#### Per-sink `log-level` & precedence

A sink-specific log-level would override any global level settings. This is still under discussion, however. 

### File Explanation:

- **File Logging**:
  The **file** sink is enabled to store logs locally in the path `/logs/dab-log.txt`. It includes additional configurations for file management:
[...]

PR #2818, opened 2025-08-14
Adding 'Configure' options to CLI for File Sink

RubenCerna2079:
## Why make this change?

- Fixes issue #2577 
  - We want to allow the file sink properties to be configurable through the CLI command `dab configure`.

## What is this change?

This change adds the file sink properties to the configure command and allows the user to change those properties through the CLI. It also ensures that the path property exists if file sink is enabled.
- `ConfigOptions.cs`: Adds file sink properties to CLI command so that they can be configured by the user.
- `ConfigGenerator.cs`: Writes the file sink properties to the config file, and errors out if the user tries to add an invalid value.
- `RuntimeConfigValidator.cs`: Validates that `runtime.telemetry.file.path` is not empty or null if the file sink is enabled.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

Added tests that ensure the configure commands and validation work.
Also refactored the validation tests that followed the same pattern.

## Sample Request(s)
--runtime.telemetry.file.enabled
--runtime.telemetry.file.path
--runtime.telemetry.file.rolling-interval
--runtime.telemetry.file.retained-file-count-limit
--runtime.telemetry.file.file-size-limit-bytes

### DIFF
--- a/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
+++ b/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
@@ -698,6 +698,7 @@ public class ConfigurationHotReloadTests
     /// Invalid change that was added is a schema file that is not complete, which should be
     /// catched by the validator.
     /// </summary>
+    [Ignore]
     [TestCategory(MSSQL_ENVIRONMENT)]
     [TestMethod]
     public void HotReloadValidationFail()


schemas/dab.draft.schema.json
```diff
@@ -501,7 +501,7 @@
                   "type": "string",
                   "description": "Rolling interval for log files.",
                   "default": "Day",
-                  "enum": ["Minute", "Hour", "Day", "Month", "Year", "Infinite"]
+                  "enum": [ "Minute", "Hour", "Day", "Month", "Year", "Infinite" ]
                 },
                 "retained-file-count-limit": {
```
Aniruddh25, line 506: the original Enhancement lists out property names as camelCased e.g. `retainedFileCountLimit`. I agree that it should be the way you have it here as kebab-case. Can you please confirm the names with @JerryNixon and fix the original enhancement: https://github.com/Azure/data-api-builder/issues/2371 to match with what you have done?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -177,6 +178,24 @@ public void ValidateAzureLogAnalyticsAuth(RuntimeConfig runtimeConfig)
...
 
+    /// <summary>
+    /// The path in File Sink is required if it is enabled.
+    /// </summary>
+    public void ValidateFileSinkPath(RuntimeConfig runtimeConfig)
+    {
+        if (runtimeConfig.Runtime!.Telemetry is not null && runtimeConfig.Runtime.Telemetry.File is not null)
+        {
+            FileSinkOptions fileSinkOptions = runtimeConfig.Runtime.Telemetry.File;
+            if (fileSinkOptions.Enabled && string.IsNullOrEmpty(fileSinkOptions.Path))
```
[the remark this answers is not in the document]
Aniruddh25: @RubenCerna2079 this is a good suggestion

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -188,6 +188,44 @@ public void TestAddAzureLogAnalyticsOptions()
...
+        {
+            // Arrange
+            _fileSystem!.AddFile(TEST_RUNTIME_CONFIG_FILE, new MockFileData(INITIAL_CONFIG));
+
+            Assert.IsTrue(_fileSystem!.File.Exists(TEST_RUNTIME_CONFIG_FILE));
+
+            // Act: Attempts to add file options
+            ConfigureOptions options = new(
+                fileSinkEnabled: CliBool.True,
+                fileSinkPath: "/custom/log/path.txt",
```
Aniruddh25: The expected value `"/custom/log/path.txt"` tested here can be defined as a constant, so it can be reused in the Assert below. This way any change needed has to made only at one place.

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -319,24 +302,53 @@ public async Task TestValidateAKVOptionsWithoutEndpointFails()
...
+            fileSinkFileSizeLimitBytes: 1024,
+            config: TEST_RUNTIME_CONFIG_FILE
+        );
+
+        // Act
+        await TestValidatePropertyOptionsFails(options);
+    }
+
+    /// <summary>
+    /// Base test that ensures properties with missing options fail validation.
```
Aniruddh25: This is not a test in itself since it is not marked with the TestMethod attribute, but a helper function. Please rename the function to simply ValidatePropertyOptionFails and fix the summary here.

src/Cli/ConfigGenerator.cs
```diff
@@ -1199,6 +1218,74 @@ private static bool TryUpdateConfiguredAzureLogAnalyticsOptions(
...
+                if (options.FileSinkPath is not null)
+                {
+                    fileOptions = fileOptions with { Path = options.FileSinkPath, UserProvidedPath = true };
+                    _logger.LogInformation($"Updated configuration with runtime.telemetry.file.path as '{options.FileSinkPath}'");
+                }
+
+                // Runtime.Telemetry.File.RollingInterval
+                if (options.FileSinkRollingInterval is not null)
+                {
+                    fileOptions = fileOptions with { RollingInterval = options.FileSinkRollingInterval, UserProvidedRollingInterval = true };
```
Aniruddh25: Will this give error if the rolling interval specified by the user is not one of the expected enums?

Aniruddh25 (approved):
LGTM, left some questions and suggestions.

src/Cli/ConfigGenerator.cs
RubenCerna2079: Since the input was a string it allowed invalid enums, changed it works as expected.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -177,6 +178,61 @@ public void ValidateAzureLogAnalyticsAuth(RuntimeConfig runtimeConfig)
...
+                string? directoryName = System.IO.Path.GetDirectoryName(fileSinkOptions.Path);
+                if (directoryName is not null && directoryName.IndexOfAny(System.IO.Path.GetInvalidPathChars()) != -1)
+                {
+                    HandleOrRecordException(new DataApiBuilderException(
+                        message: "File option 'path' cannot have invalid characters in its directory or file name.",
+                        statusCode: HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+                }
+
+                System.IO.Path.GetFullPath(fileSinkOptions.Path);
```
souvikghosh04, line 224: how are we using this full path?

souvikghosh04 (approved):
LGTM

src/Core/Configurations/RuntimeConfigValidator.cs
RubenCerna2079: We don't need to use the value it returns, we just use it as a check to ensure that the path is valid, if it throws an error then we can tell the user that there was an error with their path.

anushakolan (approved):

anushakolan (approved):

PR #2825, opened 2025-08-19
Add logic to send logs through File Sink

RubenCerna2079:
## Why make this change?

- Closes issue #2578
  - In order to complete the File Sink feature, the logic that sends the logs created by DAB to the file is required.

## What is this change?

- This change implements the logic that sends all the logs from DAB to the file through the path the user requests. This is done with the help of Serilog which is a logging library that simplifies the creation of file sinks.
- The `Startup.cs` program creates the Serilog logger pipeline and adds it as part of the services so that it is used later by the `Program.cs` to set the different loggers with the Serilog pipeline and allow the logs to be sent to the file sink.
- We also deleted the `RollingIntervalMode.cs` since we discovered that Serilog has its own rolling interval enum class, which makes the one implemented in DAB obsolete.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

Created tests that check if the services needed for the File Sink exist when the File Sink property is enabled. 
Also, created test to check if the file sink with the appropriate name is created when the property is enabled.

## Sample Request(s)
image
image

### DIFF
--- a/src/Service.Tests/Configuration/Telemetry/FileSinkTests.cs
+++ b/src/Service.Tests/Configuration/Telemetry/FileSinkTests.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.IO;
-using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Azure.DataApiBuilder.Config.ObjectModel;


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4170,21 +4171,21 @@ public void AzureLogAnalyticsSerialization(
...
+        [DataRow(true, "/test/path.csv", RollingInterval.Hour, 10, 3000, true, "/test/path.csv", RollingInterval.Hour, 10, 3000)]
+        [DataRow(false, "C://absolute/file/path.log", RollingInterval.Month, 2147483647, 2048, false, "C://absolute/file/path.log", RollingInterval.Month, 2147483647, 2048)]
+        [DataRow(false, "D://absolute/test/path.txt", RollingInterval.Year, 10, 2147483647, false, "D://absolute/test/path.txt", RollingInterval.Year, 10, 2147483647)]
+        [DataRow(false, "", RollingInterval.Infinite, 5, 512, false, "", RollingInterval.Infinite, 5, 512)]
+        [DataRow(null, null, null, null, null, false, "/logs/dab-log.txt", RollingInterval.Day, 1, 1048576)]
         public void FileSinkSerialization(
             bool? enabled,
             string? path,
-            RollingIntervalMode? rollingInterval,
+            RollingInterval? rollingInterval,
```
souvikghosh04, line 4183: any reason this is nullable but `expectedRollingInterval` non-nullable?

src/Service/Startup.cs
```diff
@@ -941,6 +963,44 @@ private void ConfigureAzureLogAnalytics(RuntimeConfig runtimeConfig)
...
+        /// <summary>
+        /// Configure File Sink based on the loaded runtime configuration. If File Sink
+        /// is enabled, we can track different events and metrics.
+        /// </summary>
+        /// <param name="app">The application builder.</param>
+        /// <param name="runtimeConfig">The provider used to load runtime configuration.</param>
+        private void ConfigureFileSink(IApplicationBuilder app, RuntimeConfig runtimeConfig)
+        {
+            if (runtimeConfig?.Runtime?.Telemetry is not null
+               && runtimeConfig.Runtime.Telemetry.File is not null)
```
souvikghosh04, line 975: we should validate the configured file name and path is in expected and valid format

src/Service/Startup.cs
```diff
@@ -941,6 +963,44 @@ private void ConfigureAzureLogAnalytics(RuntimeConfig runtimeConfig)
...
+                    _logger.LogError("Logs won't be sent to File because the Path is not available in the config file.");
+                    return;
+                }
+
+                Logger? serilogLogger = app.ApplicationServices.GetService<Logger>();
+                if (serilogLogger is null)
+                {
+                    _logger.LogError("Serilog Logger Configuration is not set.");
+                    return;
+                }
```
souvikghosh04, line 996: do we need null check here? when can this be null? is there any silent exception scenario based on which this can be null?

src/Service/Program.cs
```diff
@@ -209,6 +215,20 @@ public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, Tele
...
+
+                        if (logLevelInitializer is null)
+                        {
+                            builder.AddFilter<SerilogLoggerProvider>(category: string.Empty, logLevel);
+                        }
+                        else
+                        {
+                            builder.AddFilter<SerilogLoggerProvider>(category: string.Empty, level => level >= logLevelInitializer.MinLogLevel);
+                        }
+                    }
```
souvikghosh04, line 230: nit: when you have time, please consider refactoring this function. for example- we are checking `logLevelInitializer` is null multiple times.

souvikghosh04 (approved):
Approved with suggesstions.

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: The `rollingInterval` represents the input a user can give, which also includes null. On the other hand, `expectedRollingInterval` is what is saved inside the FileSink object, which does not expect any nulls since there are default values for those cases.

src/Service/Startup.cs
RubenCerna2079: The validation is done in the PR
- #2818

src/Service/Startup.cs
RubenCerna2079: While it shouldn't be possible for it to be null, I think it is a good idea to have this check if there is an unforeseen case in which the serilog logger is not created.

src/Service/Program.cs
RubenCerna2079: Will create a new task in order to refactor certain aspects of the telemetry.

src/Service/Startup.cs
```diff
@@ -192,6 +195,18 @@ public void ConfigureServices(IServiceCollection services)
...
             }
 
+            if (runtimeConfigAvailable
+                && runtimeConfig?.Runtime?.Telemetry?.File is not null
+                && runtimeConfig.Runtime.Telemetry.File.Enabled)
+            {
+                services.AddSingleton(sp =>
+                {
+                    FileSinkOptions options = runtimeConfig.Runtime.Telemetry.File;
+                    return new LoggerConfiguration().WriteTo.File(path: options.Path, rollingInterval: (RollingInterval)Enum.Parse(typeof(RollingInterval), options.RollingInterval), retainedFileCountLimit: options.RetainedFileCountLimit, fileSizeLimitBytes: options.FileSizeLimitBytes, rollOnFileSizeLimit: true);
```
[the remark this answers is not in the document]
RubenCerna2079: Enum values are already validated beforehand

src/Service/Program.cs
```diff
@@ -132,9 +136,11 @@ private static ParseResult GetParseResult(Command cmd, string[] args)
         /// <summary>
         /// Creates a LoggerFactory and add filter with the given LogLevel.
         /// </summary>
-        /// <param name="logLevel">minimum log level.</param>
+        /// <param name="logLevel">Minimum log level.</param>
         /// <param name="appTelemetryClient">Telemetry client</param>
-        public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, TelemetryClient? appTelemetryClient = null, LogLevelInitializer? logLevelInitializer = null)
+        /// <param name="logLevelInitializer">Hot-reloadable log level</param>
+        /// <param name="serilogLogger">Core Serilog logging pipeline</param>
+        public static ILoggerFactory GetLoggerFactoryForLogLevel(LogLevel logLevel, TelemetryClient? appTelemetryClient = null, LogLevelInitializer? logLevelInitializer = null, Logger? serilogLogger = null)
```
Aniruddh25, line 143: Did you verify multiple sinks continue to operate simultaneously if we have both file sink and logAnalytics enabled?

src/Service.Tests/Configuration/Telemetry/FileSinkTests.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    private const string CONFIG_WITH_TELEMETRY = "dab-file-sink-test-config.json";
+    private const string CONFIG_WITHOUT_TELEMETRY = "dab-no-file-sink-test-config.json";
+    private static RuntimeConfig _configuration;
+
+    /// <summary>
+    /// This is a helper function that creates runtime config file with specified telemetry options.
+    /// </summary>
+    /// <param name="configFileName">Name of the config file to be created.</param>
+    /// <param name="isFileSinkEnabled">Whether telemetry is enabled or not.</param>
```
Aniruddh25: should say whether File Sink is enabled or not

src/Service.Tests/Configuration/Telemetry/FileSinkTests.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// <param name="fileSinkPath">Path where logs will be sent to.</param>
+    /// <param name="rollingInterval">Time it takes for logs to roll over to next file.</param>
+    public static void SetUpTelemetryInConfig(string configFileName, bool isFileSinkEnabled, string fileSinkPath, RollingInterval? rollingInterval = null)
```
Aniruddh25: why does this function need to be public? Should be private if its used only within this test

src/Service.Tests/Configuration/Telemetry/FileSinkTests.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        // Act
+        using (HttpClient client = server.CreateClient())
+        {
+            HttpRequestMessage restRequest = new(HttpMethod.Get, "/api/Book");
+            await client.SendAsync(restRequest);
+        }
+
+        // Assert
+        Assert.IsTrue(File.Exists(fileName));
```
Aniruddh25, line 123: In addition to checking for the existence of the file, can you also please check the contents of the file to verify it has the basic Logging information.. anything that is gauranteed to be present like the DAB output name etc.

Aniruddh25 (approved):
Approved with some suggestions to add additional test checks

src/Service.Tests/Configuration/Telemetry/FileSinkTests.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// Tests if the logs are flushed to the proper path when File Sink is enabled.
+    /// </summary>
+    /// <summary>
+    /// Tests if the logs are flushed to the proper path when File Sink is enabled.
+    /// </summary>
+    [DataTestMethod]
+    [DataRow("file-sink-test-file.txt")]
+    [DataRow("file-sink-test-file.log")]
+    [DataRow("file-sink-test-file.csv")]
+    public async Task TestFileSinkSucceed(string fileName)
```
aaronburtle, line 102: Should we be able to write to multiple file syncs? Should we add that case here?

RubenCerna2079: No, it should be able to write only to one file sink at a time, but it should be able to write the logs to different types of files.

src/Service/Program.cs
RubenCerna2079: Yes

Aniruddh25 (approved):

Issue #2856, opened 2025-08-27
[Bug] Stored procedures with unnamed aggregate columns fail GraphQL schema generation


ranishan:
### Summary
When a stored procedure returns a result set containing a column without a name, the Data API Builder fails to start. This is because the GraphQL schema generation process cannot handle unnamed columns, leading to an "Internal Server Error" with the message: "The value of a name node cannot be null or empty."

This typically occurs when using aggregate functions (e.g., `MAX()`, `COUNT()`) without providing an explicit column alias in the SQL query.

### Reproduction Steps

1.  **Database Setup:**
    *   In a SQL Server instance, create a table and a stored procedure that uses an aggregate function without an alias.
    ```sql
    CREATE TABLE books (Id INT, Name VARCHAR(255));
    GO

    CREATE PROCEDURE [dbo].[GetMaxPrc]
    AS
    BEGIN
        -- This query returns an unnamed column
        SELECT MAX(Id)
        FROM dbo.books;
    END
    GO
    ```

2.  **Configuration:**
    *   Configure the `dab-config.json` file to use this stored procedure as an entity.
    ```json
    {
      "entities": {
        "GetMaxId": {
          "source": {
            "object": "dbo.GetMaxPrc",
            "type": "stored-procedure"
          },
          "graphql": { "type": "query" },
          "permissions": [
            {
              "role": "anonymous",
              "actions": [ "execute" ]
            }
          ]
        }
      }
    }
    ```

3.  **Run the Service:**
    *   Start the Data API Builder service.

### Expected Behavior
The service should either:
-   Start successfully by assigning a default name to the unnamed column. - Suggested by copilot
-   Fail to start but provide a clear, user-friendly error message explaining that all columns in a stored procedure's result set must have a name, and guide the user to add an alias (e.g., `SELECT MAX(Id) AS MaxId`).

### Actual Behavior
The service fails to start, and the GraphQL endpoint is unavailable. The logs show an internal server error.

### Error Details
**Message:**
`{"code":"InternalError","subCode":0,"message":"The value of a name node cannot be null or empty. (Parameter 'value')"}`

**Call Stack:**
```
   at Azure.DataApiBuilder.Service.GraphQLBuilder.Sql.SchemaConverter.GenerateFieldForColumn(Entity configEntity, String columnName, ColumnDefinition column, List\`1 directives, IEnumerable\`1 roles)
[...]

PR #2857, opened 2025-08-29
Throw user error when Stored procedures with unnamed aggregate columns fails due to column name returning as null

ranishan:
## Why make this change?
The change resolves the issue #2856 where the call to get stored procedure schema fails when the SQL Server stored procedure has columns containing values from aggregate functions that do not have an alias.
e.g.

CREATE TABLE books (Id INT, Name VARCHAR(255));
GO

CREATE PROCEDURE [dbo].[GetMaxPrc]
AS
BEGIN
    -- This query returns an unnamed column
    SELECT MAX(Id)
    FROM dbo.books;
END
GO

**Root Cause**
The code uses the table sys.dm_exec_describe_first_result_set_for_object  to identify the name of the column, type etc.
As mentioned in the following doc, the "name" column in this table can be null if the column name cannot be determined.
https://learn.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-exec-describe-first-result-set-for-object-transact-sql?view=sql-server-ver17

## What is this change?
The change handles the failure by throwing a user exception with an appropriate message to help user fix the error.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

## Sample Request(s)

e.g. dab-config for the stored procedure mentioned earlier:
{
  "entities": {
    "GetMaxId": {
      "source": {
        "object": "dbo.GetMaxPrc",
        "type": "stored-procedure"
      },
      "graphql": { "type": "query" },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [ "execute" ]
        }
      ]
    }
  }
}

### DIFF
--- a/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
@@ -1151,6 +1151,19 @@ namespace Azure.DataApiBuilder.Core.Services
                 Type resultFieldType = SqlToCLRType(element.GetProperty(BaseSqlQueryBuilder.STOREDPROC_COLUMN_SYSTEMTYPENAME).ToString());
                 bool isResultFieldNullable = element.GetProperty(BaseSqlQueryBuilder.STOREDPROC_COLUMN_ISNULLABLE).GetBoolean();
 
+                // Validate that the stored procedure returns columns with proper names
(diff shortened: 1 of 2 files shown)


vadeveka (approved):

RubenCerna2079 (approved):
LGTM

JerryNixon:
+1

JerryNixon (approved):

Issue #2834, opened 2025-08-21
[Enh]: Add entity `description` (metadata)


JerryNixon:
## What?

Add a `description` property to entities.

## Why?

Provide semantic meaning to entities.

* Useful in documentation
* Useful in client code generation
* Useful in MCP tools & to AI models

## How?

- [x] Add `description` property to configuration. 
- [x] Update the JSON schema. 
- [x] Add `description` to `dab add` in CLI. 
- [x] Add `description` to `dab update` in CLI. 
- [ ] Add `description` to `dab validate` in CLI (warn when missing and MCP enabled). 
- [x] Add `description` to DAB's OpenAPI document.
- [x] Add `description` to DAB's GraphQL schema.

## Configuration

```json
{
  "entities": {
    "entity-name": {
      "description": "optional string"
    }
  }
}
```

### Example

```json
{
  "entities": {
    "User": {
      "source": "dbo.Users",
      "description": "Represents a user in the system"
    },
    "Order": {
      "source": "dbo.Orders",
      "description": "Represents a customer order"
    }
  }
}
```

### Command line

Add `dab add User --description "Represents a user in the system"`
Add `dab update User --description "Represents a user in the system"`

anushakolan:
Validation logic to check if,
1. MCP is enabled.
2. Description is missing for entities.
3. Log a warning to the user.

This code is not being added as part of this task, as this requiers the mcp endpoint to be available. Will get back to adding this after getting the mcp endpoint changes merged. `RuntimeCongValidator.cs`-> ValidateEntityConfiguration()

anushakolan:
// Warn if description is missing for entities when MCP is enabled
            if (runtimeConfig.IsMcpEnabled && runtimeConfig.Entities != null)
            {
                foreach (KeyValuePair<string, Entity> kvp in runtimeConfig.Entities)
                {
                    if (string.IsNullOrWhiteSpace(kvp.Value.Description))
                    {
                        _logger.LogWarning("Entity '{entityName}' is missing a description. It is recommended to add a description when MCP is enabled.", kvp.Key);
                    }
                }
            }

anushakolan:
/// <summary>
    /// Verifies that a warning is raised when validating a config with a missing entity description and MCP is enabled.
    /// </summary>
    [TestMethod]
    public void TestValidateConfigWarnsOnMissingDescriptionWithMcpEnabled()
    {
        
[...]

PR #2861, opened 2025-09-04
[DAB-MCP] Added description property to entities and GraphQL Schema.

anushakolan:
## Why make this change?
  - Adds support for entity-level descriptions in Data API Builder GraphQL schema.
  - This feature request aims to surface entity descriptions from the config file in the generated GraphQL schema, improving API documentation and discoverability.
  - See related discussion: [https://github.com/Azure/data-api-builder/issues/2834]

## What is this change?
- Adds a `description` property to the entity model and ensures it is deserialized from the config.
- Updates the GraphQL schema generator and converter to include entity descriptions as comments in the SDL and as HotChocolate type descriptions.
- Adds unit tests to verify that entity descriptions appear in the generated schema.

## How was this tested?

- [x] Unit Tests: Added tests to check for presence of entity description in generated GraphQL schema.
- [x] Manual verification: Ran GraphQL introspection queries and checked schema SDL output.

## Sample Request(s)

**GraphQL Introspection Query:**
```graphql
{
  __type(name: "Todo") {
    name
    description
  }
}
```
**Sample Query Response:**
```
{
  "data": {
    "__type": {
      "name": "Todo",
      "description": "Represents a todo item in the system"
    }
  }
}
```

**Sample SDL output:**
```
"""Represents a todo item in the system"""
type Todo {
  ...
}
```

### DIFF
--- a/src/Core/Generator/SchemaGenerator.cs
+++ b/src/Core/Generator/SchemaGenerator.cs
@@ -28,10 +28,15 @@ namespace Azure.DataApiBuilder.Core.Generator
 
         // List of JSON documents to process.
         private List<JsonDocument> _data;
+
         // Name of the Azure Cosmos DB container from which the JSON data is obtained.
         private string _containerName;
+
         // Dictionary mapping plural entity names to singular names based on the provided configuration.
         private Dictionary<string, string> _entityAndSingularNameMapping = new();
+
+        // Entities from config for description lookup
+        private IReadOnlyDictionary<string, Entity>? _entities;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SchemaGenerator"/> class.
@@ -57,6 +62,9 @@ namespace Azure.DataApiBuilder.Core.Generator
                 {
                     _entityAndSingularNameMapping.Add(item.Value.GraphQL.Singular.Pascalize(), item.Key);
                 }
+
+                // Convert RuntimeEntities to Dictionary for description lookup
+                _entities = config.Entities.ToDictionary(x => x.Key, x => x.Value);
             }
         }
 
@@ -128,6 +136,22 @@ namespace Azure.DataApiBuilder.Core.Generator
             {
                 // Determine if the entity is the root entity.
                 bool isRoot = entity.Key == _containerName.Pascalize();
+
+                // Get description from config if available
+                string? description = null;
+                if (_entityAndSingularNameMapping.ContainsKey(entity.Key) && _entities != null)
+                {
+                    string configEntityName = _entityAndSingularNameMapping[entity.Key];
+                    if (_entities.ContainsKey(configEntityName))
+                    {
+                        description = _entities[configEntityName].Description;
+                    }
+                }
+
+                if (!string.IsNullOrEmpty(description))
+                {
+                    sb.AppendLine($"\"\"\"{description}\"\"\"");
+                }
 
                 sb.Append($"type {entity.Key} ");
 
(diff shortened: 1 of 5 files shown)


anushakolan:
@microsoft-github-policy-service agree company="Microsoft"

souvikghosh04 (approved):
LGTM

src/Core/Generator/SchemaGenerator.cs
```diff
@@ -129,6 +137,22 @@ private string GenerateGQLSchema()
...
+                if (_entityAndSingularNameMapping.ContainsKey(entity.Key) && _entities != null)
+                {
+                    string configEntityName = _entityAndSingularNameMapping[entity.Key];
+                    if (_entities.ContainsKey(configEntityName))
+                    {
+                        description = _entities[configEntityName].Description;
+                    }
+                }
+
+                if (!string.IsNullOrEmpty(description))
```
aaronburtle: So if the description is the empty string we just don't print that out into the schema?

anushakolan: Yes, it is optional parameter. So, if it is empty, we just don't print it.

RubenCerna2079: I think it might be better to use `IsNullOrWhiteSpace` which covers the cases where it is null and empty, but it also would cover the case in which it was just filled with spaces.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -78,6 +78,18 @@ public static ObjectTypeDefinitionNode GenerateObjectTypeDefinitionForDatabaseOb
                         subStatusCode: DataApiBuilderException.SubStatusCodes.NotSupported);
             }
 
+            StringValueNode? descriptionNode = null;
+            if (!string.IsNullOrEmpty(configEntity.Description))
```
RubenCerna2079: Same as previous comment

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -122,14 +134,20 @@ private static ObjectTypeDefinitionNode CreateObjectTypeDefinitionForStoredProce
                 }
             }
 
+            StringValueNode? descriptionNode = null;
+            if (!string.IsNullOrEmpty(configEntity.Description))
```
RubenCerna2079: Same as previous comment

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -213,14 +231,20 @@ private static ObjectTypeDefinitionNode CreateObjectTypeDefinitionForTableOrView
...
```
RubenCerna2079: Same as previous comment

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -779,6 +779,43 @@ public override async Task TestNoAggregationOptionsForTableWithoutNumericFields(
...
+                source,
+                gqlOptions,
+                restOptions,
+                [],
+                null,
+                null,
+                null,
+                false,
+                null,
+                Description: "This is a test entity description for MSSQL."
```
RubenCerna2079: I think we should also test the scenarios in which the description is invalid. I.E. null, empty, or white spaces.

PR #2866, opened 2025-09-05
DAB MCP Runtime

souvikghosh04:
## Why make this change?

- Introduces a dedicated runtime layer for MCP

## What is this change?

- Adds a runtime initialization module that centralizes configuration load, validation, DI registration, and service/materialization order.
- Refactors existing startup code into clearer phases; removes duplicated wiring.
- Improves logging structure and prepares for future feature flags / layered config overrides.
- Cleans up deprecated code paths (overall net +901 / −130 LOC).
- No intentional breaking changes (verify before merge).

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

- Discover tools: POST `http://localhost:5000/mcp`

JSON body

```
{
  "jsonrpc": "2.0",
  "id": "1",
  "method": "tools/list",
  "params": {}
}
```

- List entities: POST `http://localhost:5000/mcp`

JSON body

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "tools/call",
  "params": {
    "name": "list_entities"
  }
}
```


schemas/dab.draft.schema.json
```diff
@@ -239,6 +239,63 @@
...
+              "description": "Global DML tools configuration",
+              "properties": {
+                "enabled": {
+                  "type": "boolean",
+                  "description": "Allow enabling/disabling DML tools for all entities.",
+                  "default": false
+                },
+                "describe-entities": {
+                  "type": "boolean",
+                  "description": "Allow enabling/disabling the describe-entities command.",
```
anushakolan: Is it, 
`"Allow enabling/disabling the describe-entities command.`
or
`"Allow enabling/disabling the describe-entities tool.`?

src/Azure.DataApiBuilder.Mcp/Tools/Extensions.cs
```diff
@@ -12,10 +12,45 @@ public static class Extensions
 {
     public static IServiceProvider? ServiceProvider { get; set; }
 
-    public static void AddDmlTools(this IServiceCollection services, McpOptions mcpOptions)
+    public static void AddDmlTools(this IServiceCollection services, McpRuntimeOptions? mcpOptions)
     {
-        HashSet<string> DmlToolNames = mcpOptions.DmlTools
-            .Select(x => x.ToString()).ToHashSet();
+        if (mcpOptions?.DmlTools == null || !mcpOptions.DmlTools.Enabled)
+        {
```
anushakolan, line 18: Can we add a log here, that the tools are not enabled or no tools available?

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -12,19 +13,32 @@
...
-        private static McpOptions _mcpOptions = default!;
+        private static McpRuntimeOptions? _mcpOptions;
 
         public static IServiceCollection AddDabMcpServer(this IServiceCollection services, RuntimeConfigProvider runtimeConfigProvider)
         {
-            if (runtimeConfigProvider.TryGetConfig(out RuntimeConfig? runtimeConfig))
+            if (!Debugger.IsAttached)
             {
-                _mcpOptions = runtimeConfig?.Ai?.Mcp ?? throw new NullReferenceException("Configuration is required.");
+                Debugger.Launch(); // Forces Visual Studio/VS Code to attach
```
anushakolan: Asking for my knowledge: Why are we launching the debugger here?

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -73,21 +86,34 @@ public static IServiceCollection AddDabMcpServer(this IServiceCollection service
...
                                     : messageEl.ToString();
 
-                                return ValueTask.FromResult(new CallToolResult
+                                return new CallToolResult
                                 {
                                     Content = [new TextContentBlock { Type = "text", Text = $"Echo: {msg}" }]
-                                });
+                                };
                             }
                             else if (request.Params?.Name == "list_entities")
```
anushakolan: Is the name `list_entities` or `describe-entities`? I think we are no longer using list_entities right?

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -73,21 +86,34 @@ public static IServiceCollection AddDabMcpServer(this IServiceCollection service
...
+
+                                // Create a scope to resolve scoped services
+                                using IServiceScope scope = serviceProvider.CreateScope();
+                                IServiceProvider scopedProvider = scope.ServiceProvider;
+
+                                // Set the service provider for DmlTools
+                                Azure.DataApiBuilder.Mcp.Tools.Extensions.ServiceProvider = scopedProvider;
+
+                                // Call the ListEntities tool method
+                                string entitiesJson = await DmlTools.ListEntities();
```
anushakolan: Same here, it should be `CreateEntities` right?

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -100,10 +126,37 @@ public static IServiceCollection AddDabMcpServer(this IServiceCollection service
...
+            }
+
+            // Get the MCP path with proper null handling and default
+            string mcpPath = mcpOptions.Path ?? McpRuntimeOptions.DEFAULT_PATH;
+            
+            // Map the MCP endpoint
+            endpoints.MapMcp(mcpPath);
+            
+            // Map health checks relative to the MCP path
+            string healthPath = mcpPath.TrimEnd('/') + "/health";
```
anushakolan, line 146: Why are we trimming `'/'` and again adding it with `'/health'`? Can't we instead directly use `mapPath + "health"`?

Also, how does the sample `mcpPath` look like?

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -100,10 +126,37 @@ public static IServiceCollection AddDabMcpServer(this IServiceCollection service
             return services;
         }
 
-        public static IEndpointRouteBuilder MapDabMcp(this IEndpointRouteBuilder endpoints, [StringSyntax("Route")] string pattern = "")
+        public static IEndpointRouteBuilder MapDabMcp(this IEndpointRouteBuilder endpoints, RuntimeConfigProvider runtimeConfigProvider, [StringSyntax("Route")] string pattern = "")
```
anushakolan, line 123: Is this function to map all paths or only health path? Like will `create_entities` also gets mapped here?

anushakolan, line 123: Can we add function summary comments for the functions so that it is easier to get the code context?

src/Cli/Commands/InitOptions.cs
```diff
@@ -35,8 +35,11 @@ public InitOptions(
             bool restDisabled = false,
             string graphQLPath = GraphQLRuntimeOptions.DEFAULT_PATH,
             bool graphqlDisabled = false,
+            string mcpPath = McpRuntimeOptions.DEFAULT_PATH,
+            bool mcpDisabled = false,
```
anushakolan, line 39: Does this mean, even after we add MCP endpoint support by default it will be disabled, and the customer should explicitly enable it if he wants to use it? I see it is the same for `GraphQL` and `Rest` endpoints, so trying to understand what `false` means.

src/Cli/ConfigGenerator.cs
```diff
@@ -943,6 +963,126 @@ private static bool TryUpdateConfiguredGraphQLValues(
...
         }
 
+        /// <summary>
+        /// Attempts to update the Config parameters in the Mcp runtime settings based on the provided value.
+        /// Validates that any user-provided values are valid and then returns true if the updated Mcp options
+        /// need to be overwritten on the existing config parameters
+        /// </summary>
+        /// <param name="options">options.</param>
+        /// <param name="updatedMcpOptions">updatedMcpOptions</param>
+        /// <returns>True if the value needs to be updated in the runtime config, else false</returns>
```
anushakolan, line 973: I notice a mismatch between the method behavior and its XML summary:
`/// <returns>True if the value needs to be updated in the runtime config, else false</returns>`

However, the method currently returns true even when no updates are made, i.e., when all `updatedValues` are null. 

It only returns false if:
1. RuntimeMcpPath fails validation, or
2. An exception is thrown.

The code needs to be updated accordingly to return false when all `updatedValues` are `null`.

src/Cli/ConfigGenerator.cs
```diff
@@ -943,6 +963,126 @@ private static bool TryUpdateConfiguredGraphQLValues(
...
+                    else
+                    {
+                        _logger.LogError("Failed to update Runtime.Mcp.Path as '{updatedValue}' due to exception message: {exceptionMessage}", updatedValue, exceptionMessage);
+                        return false;
+                    }
+                }
+
+                // Runtime.Mcp.Dml-Tools.Enabled
+                updatedValue = options?.RuntimeMcpDmlToolsEnabled;
+                if (updatedValue != null)
```
anushakolan, line 1009: There is a lot of repeated code from 1008 to 1075, can we extract it into a generic function and pass values accordingly. That way a lot of the redundant code can be removed.

src/Cli/ConfigGenerator.cs
```diff
@@ -943,6 +963,126 @@ private static bool TryUpdateConfiguredGraphQLValues(
...
+        /// Attempts to update the Config parameters in the Mcp runtime settings based on the provided value.
+        /// Validates that any user-provided values are valid and then returns true if the updated Mcp options
+        /// need to be overwritten on the existing config parameters
+        /// </summary>
+        /// <param name="options">options.</param>
+        /// <param name="updatedMcpOptions">updatedMcpOptions</param>
+        /// <returns>True if the value needs to be updated in the runtime config, else false</returns>
+        private static bool TryUpdateConfiguredMcpValues(
+            ConfigureOptions options,
+            ref McpRuntimeOptions? updatedMcpOptions)
```
anushakolan, line 976: Are we checking anywhere in the code before if `updatedMcpOptions` is not null before passing. The reason I ask is because we are using the `!` operator below, 
`updatedMcpOptions = updatedMcpOptions! with { Enabled = (bool)updatedValue };`

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -146,6 +146,25 @@ public string GraphQLPath
...
 
+    /// <summary>
+    /// The path at which MCP API is available
+    /// </summary>
+    [JsonIgnore]
+    public string McpPath
+    {
+        get
+        {
+            if (Runtime is null || Runtime.Mcp is null)
```
anushakolan: what about `Runtime.Mcp.path` is null? Is that case not possible?

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -53,25 +53,42 @@ public void CleanupAfterEachTest()
...
+                enableEntityRest,
+                enableEntityGraphQL);
+
             WriteToCustomConfigFile(runtimeConfig);
 
             string[] args = new[]
             {
-                $"--ConfigFileName={CUSTOM_CONFIG_FILENAME}"
-            };
+        $"--ConfigFileName={CUSTOM_CONFIG_FILENAME}"
```
anushakolan, line 90: The indentation needs to be fixed.

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -101,6 +118,7 @@ public async Task ComprehensiveHealthEndpoint_ValidateContents(bool enableGlobal
             }
         }
 
+
```
anushakolan: Extra line needs to be removed.

src/Service.Tests/Configuration/HealthEndpointTests.cs
```diff
@@ -443,6 +461,7 @@ private static void ValidateConfigurationDetailsHealthCheckResponse(Dictionary<s
                 ValidateConfigurationIsCorrectFlag(configPropertyValues, "rest", enableGlobalRest);
                 ValidateConfigurationIsNotNull(configPropertyValues, "graphql");
                 ValidateConfigurationIsCorrectFlag(configPropertyValues, "graphql", enableGlobalGraphQL);
+                ValidateConfigurationIsCorrectFlag(configPropertyValues, "mcp", enableGlobalMcp);
```
anushakolan, line 464: Why are we not checking `ValidateConfigurationIsNotNull` similar to `graphql`?

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -1440,20 +1449,27 @@ public void ValidateValidEntityDefinitionsDoesNotGenerateDuplicateQueries(Databa
...
         /// <param name="expectError">Exception expected</param>
         [DataTestMethod]
-        [DataRow("/graphql", "/graphql", true)]
-        [DataRow("/api", "/api", true)]
-        [DataRow("/graphql", "/api", false)]
-        public void TestGlobalRouteValidation(string graphQLConfiguredPath, string restConfiguredPath, bool expectError)
+        [DataRow("/graphql", "/graphql", "/mcp", true, DisplayName = "GraphQL and REST conflict (same path), MCP different.")]
+        [DataRow("/api", "/api", "/mcp", true, DisplayName = "REST and GraphQL conflict (same path), MCP different.")]
+        [DataRow("/graphql", "/api", "/mcp", false, DisplayName = "GraphQL and REST distinct, MCP different.")]
+        // Extra case: conflict with MCP
```
anushakolan, line 1458: Why are we not adding the extra case now?

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -1715,19 +1738,26 @@ public void ValidateApiURIsAreWellFormed(
...
         [DataTestMethod]
         public void EnsureFailureWhenBothRestAndGraphQLAreDisabled(
             bool restEnabled,
             bool graphqlEnabled,
+            bool mcpEnabled,
             bool expectError)
         {
             GraphQLRuntimeOptions graphQL = new(Enabled: graphqlEnabled);
             RestRuntimeOptions rest = new(Enabled: restEnabled);
+            McpRuntimeOptions mcp = new(Enabled: false);
```
anushakolan: Why is this directly `false`, instead of `mcpEnabled`?

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
```diff
@@ -399,6 +403,100 @@ public async Task ValidateInferredRelationshipInfoForPgSql()
...
+                Assert.Fail($"Expected DataApiBuilderException was not thrown for invalid resultFieldName: '{invalidFieldName}'.");
+            }
+            catch (DataApiBuilderException ex)
+            {
+                Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
+                Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ErrorInInitialization, ex.SubStatusCode);
+                Assert.IsTrue(ex.Message.Contains("returns a column without a name"));
+            }
+
+            TestHelper.UnsetAllDABEnvironmentVariables();
```
anushakolan: If the test fails before this line, the cleanup won’t run. It is good to wrap it in a `finally` block:

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
```diff
@@ -399,6 +403,100 @@ public async Task ValidateInferredRelationshipInfoForPgSql()
...
         }
 
+        /// <summary>
+        /// Data-driven test to validate that DataApiBuilderException is thrown for various invalid resultFieldName values
+        /// during stored procedure result set definition population.
+        /// </summary>
+        [DataTestMethod, TestCategory(TestCategory.MSSQL)]
+        [DataRow(null, DisplayName = "Null result field name")]
+        [DataRow("", DisplayName = "Empty result field name")]
+        [DataRow("   ", DisplayName = "Multiple spaces result field name")]
```
anushakolan, line 413: How does the valid `resultFieldName`  values look like?

anushakolan, line 413: Is this test only validating empty or null fields, what if the value is invalid of some SQL reserved keyword, can this test validate them as well?

anushakolan (changes requested):

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -23,6 +23,31 @@
         }
       }
     },
+    "mcp": {
+      "enabled": false,
```
anushakolan: Sam question, is `MCP` disabled by default? If so, why are we disabling it?

src/Service/Telemetry/AzureLogAnalyticsCustomLogCollector.cs
```diff
@@ -53,14 +53,18 @@ await _logs.Writer.WriteAsync(
...
         if (await _logs.Reader.WaitToReadAsync())
         {
-            while (_logs.Reader.TryRead(out AzureLogAnalyticsLogs? item))
+            Stopwatch time = Stopwatch.StartNew();
+
+            while (true)
             {
-                item.Identifier = dabIdentifier;
-                list.Add(item);
+                if (_logs.Reader.TryRead(out AzureLogAnalyticsLogs? item))
```
anushakolan, line 63: How are the Azure Log Analytics changes relevant here?

schemas/dab.draft.schema.json
```diff
@@ -239,6 +239,63 @@
...
+            "path": {
+              "default": "/mcp",
+              "type": "string"
+            },
+            "enabled": {
+              "type": "boolean",
+              "description": "Allow enabling/disabling MCP requests for all entities.",
+              "default": false
+            },
+            "dml-tools": {
```
JerryNixon, line 256: This property is intended to be one of two types. Either a `boolean` (which switches ALL of the nested properties in a single go) or `object` where you can individually set each nested property. Something like this:

```json
{
  "type": "object",
  "properties": {
    "dml-tools": {
      "anyOf": [
        { "type": "boolean" },
        { "type": "object" }
      ]
    }
  }
}
```

schemas/dab.draft.schema.json
JerryNixon: Tool, for sure.

src/Azure.DataApiBuilder.Mcp/Extensions.cs
JerryNixon: This is left over from our prototype and should be removed.

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -33,18 +47,18 @@ public static IServiceCollection AddDabMcpServer(this IServiceCollection service
             // Register MCP server with dynamic tool handlers
             services.AddMcpServer(options =>
             {
-                options.ServerInfo = new Implementation { Name = "MyServer", Version = "1.0.0" };
+                options.ServerInfo = new() { Name = "Data API Builder MCP Server", Version = "1.0.0" };
```
JerryNixon, line 44: Should we come up with a reason for this version number or surface it from somewhere else?

src/Azure.DataApiBuilder.Mcp/Extensions.cs
JerryNixon: It should be `DescribeEntities()`

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -56,15 +70,14 @@ public static IServiceCollection AddDabMcpServer(this IServiceCollection service
                                             }"
                                         )
                                     },
-                                    new Tool
+                                    new()
                                     {
                                         Name = "list_entities",
```
JerryNixon: `describe_entities` is the updated name.

src/Azure.DataApiBuilder.Mcp/Extensions.cs
```diff
@@ -33,18 +47,18 @@ public static IServiceCollection AddDabMcpServer(this IServiceCollection service
...
                     {
                         ListToolsHandler = (request, ct) =>
                             ValueTask.FromResult(new ListToolsResult
                             {
                                 Tools =
                                 [
-                                    new Tool
+                                    new()
                                     {
                                         Name = "echonew",
```
JerryNixon, line 57: Does it make sense to include `Echo` anymore?

src/Azure.DataApiBuilder.Mcp/Extensions.cs
JerryNixon: Because we are appending `/health` we avoid `path//health` which is invalid. We could use `Uri` if we wanted.

src/Azure.DataApiBuilder.Mcp/Extensions.cs
souvikghosh04: I think we can skip or keep this as is for now until we figure out a good use case to have a Version exposed for MCP

src/Azure.DataApiBuilder.Mcp/Extensions.cs
souvikghosh04: `create_entities` and other tools are not part of the API path. we use JSON payload body in the POST request

src/Azure.DataApiBuilder.Mcp/Extensions.cs
souvikghosh04: we can expose /health as an MCP tool itself rather than a separate endpoint since /mcp is the only endpoint and we use the JSON payload in request body to use them. for now, this function maps both

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
souvikghosh04: the code below that comment already does that and this comment seems unnecessary. will have it removed in next PR

Issue #2863, opened 2025-09-04
[Enh]: Support for `@file()`


JerryNixon:
## What?

In the same way `@env()` and `@akv()` allow for string replacements in the DAB configuration, introduce `@file()` which will read from a designated JSON file with a simple name/value dictionary `<string, string>`.

### Configuration

```json
{
  "metadata-file": {
    "path": "./metadata.json"
  }
}
```

#### Rules

These rules govern the implementation.

1. Any property name or value can be set using `@file()`.
2. The `@file()` function is a string replacement in the same way `@env()` and `@akv()` are.
3. The metadata file cannot have duplicate keys (this is a core feature of a dictionary already).
4. The metadata file is read before configuration parsing and is not hot-reloaded.

#### Order of operation

For string replacement in the configuration:

1. `@env()` is first and optional.
2. `@akv()` is second and optional.
3. `@file()` is third and optional.

```mermaid
sequenceDiagram
  actor Engine as Engine
  participant ConfigInMem as ConfigInMem
  participant Environment as Environment
  participant AKV as AKV
  participant File as MetadataFile
  participant Config as ConfigFile

  Engine ->> Engine: Pre-Start
  Engine ->> Config: Load Config
  Config -->> Engine: Config Data
  Engine ->> ConfigInMem: Create In-Memory Config

  Note over Engine: Perform Config Replacements

  activate Engine
  ConfigInMem -->> Engine: Parse @env Values
  Engine ->> Environment: Get
  Environment -->> Engine: Values
  Engine ->> ConfigInMem: Replace @env Values
  deactivate Engine

  Note over Engine: Use Config With ENV Replacements

  activate Engine
  ConfigInMem -->> Engine: Parse @akv Values
  Engine ->> AKV: Request
  AKV -->> Engine: Secrets
  Engine ->> ConfigInMem: Replace @akv Values
  deactivate Engine

  Note over Engine: Use Config With AKV Replacements

  activate Engine
  ConfigInMem -->> Engine: Parse @file Values
  Engine ->> File: Read
  File -->> Engine: Key/Value Pairs
  Engine ->> ConfigInMem: Replace @file Values
  deactivate Engine

  Note over Engine: Use Config With FILE Replacements

  Engine ->> Engine: Start
```

Now: this means `metadata-file.path` could be set using `@env()` or `@akv` if the developer wants to.

#### JSON schema

1. `metadata-file` is an optional property of type `object`.
[...]

PR #2882, opened 2025-09-18
Update variable replacement during deserialization to use replacement settings class and add AKV replacement logic.

### LINKED ISSUE #2708: [Enhancement]: Support AKV
For developers who want their secrets in Azure Key Vault, support a syntax in the config that allows it.

## Added properties

```json
{
  "azure-key-vault" : {
    "endpoint": "url", (string, required when @akv() is present)
    "retry-policy": {
        "mode": "fixed | exponential", (enum, default: exponential)
        "max-count": 3, (integer, default: 3)
        "delay-seconds": 1, (integer, default: 1)
        "max-delay-seconds": 100 (integer, default: 60),
        "network-timeout-seconds": 100 (integer, default: 60),
    }
  }
}
```

> Based on: https://github.com/Azure/azure-sdk-for-net/blob/dc70df6e1bb2595cf351269fdbba0ddd84515de5/sdk/core/Azure.Core/src/RetryOptions.cs#L67C42-L67C46

| Setting                          | Meaning                       | Default          |  Min | Max |
| -------------------------------- | ----------------------------- | ---------------- | -| -| 
| `endpoint`                       | Key Vault URL                 | *(not required)* | 
| `retry-policy / mode`              | Retry strategy                | `exponential`    |  - | - |
| `retry-policy / max-count`         | Max retry attempts            | `3` (for both strategies) |  0 | `int.MaxValue` |
| `retry-policy / delay-seconds`     | Initial delay between retries | `1` (for both strategies) |  1 | `int.MaxValue` |
| `retry-policy / max-delay-seconds` | Max delay cap                 | `60` (only for `exponential`) |  1 | `int.MaxValue` |
|`retry-policy / network-timeout-seconds`   | Max network timeout per try |  `60` | 1 | `int.MaxValue` |

### JSON Schema

Update our JSON schema. Add properties and constraints and defaults.

1. `azure-key-vault` is not required. 
2. `azure-key-vault.endpoint` is required when `azure-key-vault` is present. 

### CLI updates

Add command line support.

1. Add `dab configure --azure-key-vault.endpoint`
2. Add `dab configure --azure-key-vault.retry-policy.mode`
3. Add `dab configure --azure-key-vault.retry-policy.max-count`
4. Add `dab configure --azure-key-vault.retry-policy.delay-seconds`
5. Add `dab configure --azure-key-vault.retry-policy.max-delay-seconds`
6. Add `dab configure --azure-key-vault.retry-policy.network-timeout-seconds`

## Added syntax

Support new `@akv()` method to replace property values in our DAB configuration.

```json
{
  "data-source": {
    "connection-string": "@akv('my-connection-string')"
  }
}
```

This works. Because `env` runs first. 

```json
{
  "data-source": {
    "connection-string": "@akv('my-connection-string')"
  },
  "azure-key-vault": {
    "endpoint": "@env('my-akv-endpoint')"
  }
}
```

## Details

1. `@env()` executes before `@akv()`. So, `azure-key-vault.endpoint` can be set by `@env()`.
2. For `@akv('key-value')`, "key-value" is the secret name. 
3. Like `@env()` DAB supports `@akv()` for any config property. 
4. Using `@akv()` requires either system-assigned or user-assigned managed identity.
5. The `@akv()` function works in both mode `development` and `production`. 
6. The syntax `@env('@akv('value')')` is _not_ supported and would look for an env named `@akv('value')`.
7. However, the syntax `@akv('@env('value')')` is supported. 
8. The DAB config does not monitor, listen to or poll AKV for secret changes. Requires restart. 

## Key rules (based on AKV docs)

| Rule          | Constraint                                        |
| ------------- | ------------------------------------------------- |
| Allowed chars | Alphanumeric and `-` (hyphen)                     |
| Disallowed    | No spaces, no `_`, `/`, `@`, `.`, etc.            |
| Start/end     | Must start and end with an alphanumeric character |
| Length        | 1 to 127 characters                               |
| Case          | Case-insensitive (`MySecret` = `mysecret`)        |
| Uniqueness    | Must be unique within the Key Vault               |

## Multiple configs

DAB supports multiple configuration files, where the top-level config is the master and others inherit `runtime` settings from it. Child configs define their own connection strings in the `data-source` property and may also have unique `azure-key-vault.endpoint` values. If a child omits the endpoint, it inherits the setting from the master config, which is required only if the child omits it.

1. Each config file can define a unique `azure-key-vault.endpoint` value.
2. If a child omits `azure-key-vault.endpoint`, it inherits from the master.
3. A child may use `@akv()` even if the master does not define an endpoint. However, in this case, the child configuration must define an endpoint value. 

## Consideration

1. Include an OTEL `activity` wrapping the replacement. 
2. (would be nice) Include an OTEL `activity` wrapping the replacement for `@env()`. 

## Errors

1. When `@akv()` appears without `azure-key-vault.endpoint`, we will log (as error) and fail to start.
2. When `@akv()` connection fails (for any reason), we will log the reason (as error) and fail to start.
3. When `@akv()` secret not found, we will log "secret not found: {property-name}" (as error) and fail to start.
4. When `@akv()` fails during hot reload, we will log "@akv error: {message}", then return to "last known good". 

## Flow

```mermaid
sequenceDiagram
  actor Engine as Engine
  participant ConfigInMem as ConfigInMem
  participant Environment as Environment
  participant AKV as AKV
  participant Config as ConfigFile

  Engine ->> Engine: Pre-Start
  Engine ->> Config: Load Config
  Config -->> Engine: Config Data
  Engine ->> ConfigInMem: Create In-Memory Config

  Note over Engine: Perform Config Replacements

  activate Engine
  ConfigInMem -->> Engine: Parse @env Values
  Engine ->> Environment: Get
  Environment -->> Engine: Values
  Engine ->> ConfigInMem: Replace @env Values
  deactivate Engine

  Note over Engine: Use Config With ENV Replacements

  activate Engine
  ConfigInMem -->> Engine: Parse @akv Values
  Engine ->> AKV: Request 
  AKV -->> Engine: Secrets
  Engine ->> ConfigInMem: Replace @akv Values
  deactivate Engine

  Note over Engine: Use Config With AKV Replacements

  Engine ->> Engine: Start
```

### LINKED ISSUE #2748: Support `.akv` files just like `.env` files.
Support `.akv` files just like `.env` files.

aaronburtle:
## Why make this change?

Adds AKV variable replacement and expands our design for doing variable replacements to be more extensible when new variable replacement logic is added.

Closes #2708
Closes #2748
Related to #2863

## What is this change?

Change the way that variable replacement is handled to instead of simply using a `bool` to indicate that we want env variable replacement, we add a class which holds all of the replacement settings. This will hold whether or not we will do replacement for each kind of variable that we will handle replacement for during deserialization. We also include the replacement failure mode, and put the logic for handling the replacements into a strategy dictionary which pairs the replacement variable type with the strategy for doing that replacement.

Because Azure Key Vault secret replacement requires having the retry and connection settings in order to do the AKV replacement, we must do a first pass where we only do non-AKV replacement and get the required settings so that if AKV replacement is used we have the required settings to do that replacement.

We also have to keep in mind that the legacy of the `Configuration Controller` will ignore all variable replacement, so we construct the replacement settings for this code path to not use any variable replacement at all.

## How was this tested?

We have updated the logic for the tests to use the new system, however manual testing using an actual AKV is still required.

## Sample Request(s)

### DIFF
--- a/src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
+++ b/src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
@@ -188,7 +188,7 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
 
             _sourceDefinition.SourceEntityRelationshipMap.Add("persons", metadata);
 
-            // In serialization options we need  ReferenceHandler = ReferenceHandler.Preserve, or else it doesnot seiliate objects with cycle references
+            // In serialization options we need  ReferenceHandler = ReferenceHandler.Preserve, or else it does not serialize objects with cycle references
             // SourceDefinition -> RelationShipMetadata -> ForeignKeyDefinition RelationshipPair ->DatabaseTable -> SourceDefinition
             Assert.ThrowsException<JsonException>(() =>
             {


src/Config/RuntimeConfigLoader.cs
```diff
@@ -129,25 +129,91 @@ protected void SignalConfigChanged(string message = "")
...
+            using JsonDocument doc = JsonDocument.Parse(json);
+            if (doc.RootElement.TryGetProperty("azure-key-vault", out JsonElement akvElement))
+            {
+                return JsonSerializer.Deserialize<AzureKeyVaultOptions>(akvElement.GetRawText(), options);
+            }
+        }
+        catch
+        {
+            // If we can't extract AKV options, return null and proceed without AKV variable replacement
+            return null;
```
souvikghosh04, line 171: I don't see anything explicitly handled in catch. It returns null as same outside the catch

src/Config/RuntimeConfigLoader.cs
```diff
@@ -180,11 +246,11 @@ public static bool TryParseConfig(string json,
                 DataSource ds = config.GetDataSourceFromDataSourceName(dataSourceKey);
 
                 // Add Application Name for telemetry for MsSQL or PgSql
-                if (ds.DatabaseType is DatabaseType.MSSQL && replaceEnvVar)
+                if (ds.DatabaseType is DatabaseType.MSSQL && replacementSettings?.DoReplaceEnvVar == true)
                 {
                     updatedConnection = GetConnectionStringWithApplicationName(connectionValue);
                 }
-                else if (ds.DatabaseType is DatabaseType.PostgreSQL && replaceEnvVar)
+                else if (ds.DatabaseType is DatabaseType.PostgreSQL && replacementSettings?.DoReplaceEnvVar == true)
```
souvikghosh04, line 248: nit: `replacementSettings?.DoReplaceEnvVar is true` ?

src/Config/Converters/AKVRetryPolicyOptionsConverterFactory.cs
```diff
@@ -12,9 +12,8 @@ namespace Azure.DataApiBuilder.Config.Converters;
 /// </summary>
 internal class AKVRetryPolicyOptionsConverterFactory : JsonConverterFactory
 {
-    // Determines whether to replace environment variable with its
-    // value or not while deserializing.
-    private bool _replaceEnvVar;
+    // Settings for variable replacement during deserialization.
```
RubenCerna2079, line 15: Maybe make the comment a bit more descriptive, to show that it can replace the variable with either AKV, an environment variable, or nothing during deserialization.

src/Config/Converters/AKVRetryPolicyOptionsConverterFactory.cs
```diff
@@ -25,27 +24,26 @@ public override bool CanConvert(Type typeToConvert)
...
-        _replaceEnvVar = replaceEnvVar;
+        _replacementSettings = replacementSettings;
     }
 
     private class AKVRetryPolicyOptionsConverter : JsonConverter<AKVRetryPolicyOptions>
     {
-        // Determines whether to replace environment variable with its
-        // value or not while deserializing.
-        private bool _replaceEnvVar;
+        // Settings for variable replacement during deserialization.
```
RubenCerna2079, line 40: Same as comment above, this applies to all the places where the variable type was changed

src/Config/Converters/EntityCacheOptionsConverterFactory.cs
```diff
@@ -25,27 +25,27 @@ public override bool CanConvert(Type typeToConvert)
...
     public override JsonConverter? CreateConverter(Type typeToConvert, JsonSerializerOptions options)
     {
-        return new EntityCacheOptionsConverter(_replaceEnvVar);
+        return new EntityCacheOptionsConverter(_replacementSettings);
     }
 
     /// <param name="replaceEnvVar">Whether to replace environment variable with its
     /// value or not while deserializing.</param>
-    internal EntityCacheOptionsConverterFactory(bool replaceEnvVar)
+    internal EntityCacheOptionsConverterFactory(DeserializationVariableReplacementSettings? replacementSettings)
```
RubenCerna2079, line 32: The `param name` comment should also be changed.

src/Config/Converters/EntityCacheOptionsConverterFactory.cs
```diff
@@ -25,27 +25,27 @@ public override bool CanConvert(Type typeToConvert)
...
     {
         // Determines whether to replace environment variable with its
         // value or not while deserializing.
-        private bool _replaceEnvVar;
+        private DeserializationVariableReplacementSettings? _replacementSettings;
 
         /// <param name="replaceEnvVar">Whether to replace environment variable with its
         /// value or not while deserializing.</param>
-        public EntityCacheOptionsConverter(bool replaceEnvVar)
+        public EntityCacheOptionsConverter(DeserializationVariableReplacementSettings? replacementSettings)
```
RubenCerna2079, line 44: The `param name` comment should also be changed.

src/Config/Converters/StringJsonConverterFactory.cs
```diff
@@ -28,40 +27,35 @@ public override bool CanConvert(Type typeToConvert)
...
+                    return value;
+                }
+
+                // Apply all replacement strategies configured in the settings
+                foreach (KeyValuePair<Regex, Func<Match, string>> strategy in _replacementSettings.ReplacementStrategies)
+                {
+                    value = strategy.Key.Replace(value, new MatchEvaluator(strategy.Value));
+                }
+
+                return value;
```
RubenCerna2079, line 58: I am confused with how this section works, wouldn't this just give you the last replacement strategy since they are all saved in the same value?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -226,7 +226,8 @@ public bool TryLoadConfig(
                 }
             }
 
-            if (!string.IsNullOrEmpty(json) && TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar))
+            if (!string.IsNullOrEmpty(json) && TryParseConfig(json, out RuntimeConfig, 
+                new DeserializationVariableReplacementSettings(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAKVVar: true), logger: null, connectionString: _connectionString))
```
RubenCerna2079: We also need to delete the parameter `replaceEnvVar` from the `TryLoadConfig` function since we don't use it anymore

src/Config/RuntimeConfigLoader.cs
```diff
@@ -129,25 +129,91 @@ protected void SignalConfigChanged(string message = "")
...
+            envFailureMode: replacementFailureMode);
+        options.Converters.Add(new StringJsonConverterFactory(envOnlySettings));
+        options.Converters.Add(new EnumMemberJsonEnumConverterFactory());
+        options.Converters.Add(new AzureKeyVaultOptionsConverterFactory(replacementSettings: envOnlySettings));
+        options.Converters.Add(new AKVRetryPolicyOptionsConverterFactory(replacementSettings: envOnlySettings));
+
+        // Add environment variable replacement if enabled
+        if (enableEnvReplacement)
+        {
+
```
RubenCerna2079: I assume that there must be something missing inside this if

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
```diff
@@ -0,0 +1,113 @@
...
+                            if (reader.TokenType is JsonTokenType.StartObject)
+                            {
+                                // Pass the replaceEnvVar setting to the retry policy converter
+                                retryPolicy = JsonSerializer.Deserialize<AKVRetryPolicyOptions>(ref reader, options);
+                            }
+
+                            break;
+
+                        default:
+                            reader.Skip();
```
RubenCerna2079: Why do we want to skip an incorrect property? Shouldn't we want to throw an error?

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
```diff
@@ -0,0 +1,113 @@
...
+                            break;
+                    }
+                }
+            }
+            else if (reader.TokenType is JsonTokenType.Null)
```
RubenCerna2079: nit: I think it would be better to put this on top so it is consistent with other ConverterFactories

src/Config/Converters/StringJsonConverterFactory.cs
```diff
@@ -28,40 +27,35 @@ public override bool CanConvert(Type typeToConvert)
...
+            _replacementSettings = replacementSettings;
         }
 
         public override string? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
             if (reader.TokenType == JsonTokenType.String)
             {
                 string? value = reader.GetString();
-                return Regex.Replace(value!, ENV_PATTERN, new MatchEvaluator(ReplaceMatchWithEnvVariable));
+                if (string.IsNullOrEmpty(value))
```
RubenCerna2079: Is it possible for the value to be white spaces? If so I it would be better to use `IsNullOrWhiteSpace`

src/Config/Converters/Utf8JsonReaderExtensions.cs
```diff
@@ -19,8 +19,7 @@ static internal class Utf8JsonReaderExtensions
     /// <returns>The result of deserialization.</returns>
     /// <exception cref="JsonException">Thrown if the <see cref="JsonTokenType"/> is not String.</exception>
     public static string? DeserializeString(this Utf8JsonReader reader,
-        bool replaceEnvVar,
-        EnvironmentVariableReplacementFailureMode replacementFailureMode = EnvironmentVariableReplacementFailureMode.Throw)
+        DeserializationVariableReplacementSettings? replacementSettings)
```
RubenCerna2079, line 21: Need to update the parameters in the summary section

src/Config/RuntimeConfigLoader.cs
```diff
@@ -51,7 +51,7 @@ public RuntimeConfigLoader(HotReloadEventHandler<HotReloadEventArgs>? handler =
     /// <returns>DabChangeToken</returns>
 #pragma warning disable CA1024 // Use properties where appropriate
     public IChangeToken GetChangeToken()
-#pragma warning restore CA1024 // Use properties where appropriate
+#pragma warning restore CA1024 // Use properties where Appropriate
```
RubenCerna2079: nit: Change this back to lower case

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                return null;
+            }
+
+            throw new JsonException("Invalid AzureKeyVaultOptions format");
+        }
+
+        public override void Write(Utf8JsonWriter writer, AzureKeyVaultOptions value, JsonSerializerOptions options)
+        {
+            JsonSerializer.Serialize(writer, value, options);
```
RubenCerna2079: Do we always want to write the options back into the config? I would expect to only write values if the user wants to do it in the first place.

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+            EnvFailureMode = envFailureMode;
+
+            if (DoReplaceEnvVar)
+            {
+                ReplacementStrategies.Add(
+                    new Regex(INNER_ENV_PATTERN, RegexOptions.Compiled),
+                    ReplaceEnvVariable);
+            }
+
+            if (DoReplaceAKVVar && _azureKeyVaultOptions is not null)
```
RubenCerna2079: This function prioritizes AKV over env. Do we want it to work that way? Or is it expected that only one should work in any given case?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+                        statusCode: System.Net.HttpStatusCode.ServiceUnavailable,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
+            }
+            else
+            {
+                return value ?? match.Value;
+            }
+        }
+
+        private string ReplaceAKVVariable(Match match)
```
RubenCerna2079: nit: Change name of function to `ReplaceAkvVariable`

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            // strips first and last characters, ie: '''hello'' --> ''hello'
+            string name = Regex.Match(match.Value, INNER_AKV_PATTERN).Value[1..^1];
+            string? value = GetAKVVariable(name);
+            if (EnvFailureMode == EnvironmentVariableReplacementFailureMode.Throw)
+            {
+                return value is not null ? value :
+                    throw new DataApiBuilderException(message: $"Azure Key Vault Variable, {name}, not found.",
```
RubenCerna2079: nit:
```suggestion
                    throw new DataApiBuilderException(message: $"Azure Key Vault Variable '{name}' not found.",
```

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        private static SecretClient CreateSecretClient(AzureKeyVaultOptions options)
+        {
+            if (string.IsNullOrWhiteSpace(options.Endpoint))
+            {
+                throw new DataApiBuilderException(
+                    "Azure Key Vault endpoint must be specified.",
+                    System.Net.HttpStatusCode.InternalServerError,
+                    DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
```
RubenCerna2079, line 170: nit:
```suggestion
                throw new DataApiBuilderException(
                    "Missing 'endpoint' property is required to connect to Azure Key Value.",
                    System.Net.HttpStatusCode.InternalServerError,
                    DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
```

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+                clientOptions.Retry.MaxRetries = options.RetryPolicy.MaxCount ?? 3;
+                clientOptions.Retry.Delay = TimeSpan.FromSeconds(options.RetryPolicy.DelaySeconds ?? 1);
+                clientOptions.Retry.MaxDelay = TimeSpan.FromSeconds(options.RetryPolicy.MaxDelaySeconds ?? 16);
+                clientOptions.Retry.NetworkTimeout = TimeSpan.FromSeconds(options.RetryPolicy.NetworkTimeoutSeconds ?? 30);
+            }
+
+            return new SecretClient(new Uri(options.Endpoint), new DefaultAzureCredential(), clientOptions);
+        }
+
+        private string? GetAKVVariable(string name)
```
RubenCerna2079: nit: Change `GetAKVVariable` to `GetAkvVariable`

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+
+        private static SecretClient CreateSecretClient(AzureKeyVaultOptions options)
+        {
+            if (string.IsNullOrWhiteSpace(options.Endpoint))
+            {
+                throw new DataApiBuilderException(
+                    "Azure Key Vault endpoint must be specified.",
+                    System.Net.HttpStatusCode.InternalServerError,
```
RubenCerna2079, line 169: I am not sure if that is the correct error that we should have since this is an error due to a missing property in the config file

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+                    AKVRetryPolicyMode.Exponential => RetryMode.Exponential,
+                    null => RetryMode.Exponential,
+                    _ => RetryMode.Exponential
+                };
+
+                clientOptions.Retry.Mode = retryMode;
+                clientOptions.Retry.MaxRetries = options.RetryPolicy.MaxCount ?? 3;
+                clientOptions.Retry.Delay = TimeSpan.FromSeconds(options.RetryPolicy.DelaySeconds ?? 1);
+                clientOptions.Retry.MaxDelay = TimeSpan.FromSeconds(options.RetryPolicy.MaxDelaySeconds ?? 16);
+                clientOptions.Retry.NetworkTimeout = TimeSpan.FromSeconds(options.RetryPolicy.NetworkTimeoutSeconds ?? 30);
```
RubenCerna2079: Shouldn't we use the `DEFAULT` variables found in `AKVRetryPolicyOptions` instead of using numbers?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -129,25 +129,91 @@ protected void SignalConfigChanged(string message = "")
...
+                return JsonSerializer.Deserialize<AzureKeyVaultOptions>(akvElement.GetRawText(), options);
+            }
+        }
+        catch
+        {
+            // If we can't extract AKV options, return null and proceed without AKV variable replacement
+            return null;
+        }
+
+        return null;
```
RubenCerna2079, line 174: nit: Returning the nulls like this seems a bit redundant, do you think there is a better way to do this?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+using Azure.DataApiBuilder.Config.Converters;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Service.Exceptions;
+
+namespace Azure.DataApiBuilder.Config
+{
+    public class DeserializationVariableReplacementSettings
+    {
+        public bool DoReplaceEnvVar { get; set; } = true;
+        public bool DoReplaceAKVVar { get; set; } = true;
```
RubenCerna2079:
```suggestion
        public bool DoReplaceAkvVar { get; set; } = true;
```

src/Config/RuntimeConfigLoader.cs
```diff
@@ -129,25 +129,91 @@ protected void SignalConfigChanged(string message = "")
...
+            PropertyNameCaseInsensitive = false,
+            PropertyNamingPolicy = new HyphenatedNamingPolicy(),
+            ReadCommentHandling = JsonCommentHandling.Skip
+        };
+        DeserializationVariableReplacementSettings envOnlySettings = new(
+            azureKeyVaultOptions: null,
+            doReplaceEnvVar: enableEnvReplacement,
+            doReplaceAKVVar: false,
+            envFailureMode: replacementFailureMode);
+        options.Converters.Add(new StringJsonConverterFactory(envOnlySettings));
```
RubenCerna2079, line 155: nit:
```suggestion
            envFailureMode: replacementFailureMode);
            
        options.Converters.Add(new StringJsonConverterFactory(envOnlySettings));
```

src/Config/RuntimeConfigLoader.cs
```diff
@@ -129,25 +129,91 @@ protected void SignalConfigChanged(string message = "")
...
+
+        }
+
+        try
+        {
+            using JsonDocument doc = JsonDocument.Parse(json);
+            if (doc.RootElement.TryGetProperty("azure-key-vault", out JsonElement akvElement))
+            {
+                return JsonSerializer.Deserialize<AzureKeyVaultOptions>(akvElement.GetRawText(), options);
+            }
```
RubenCerna2079, line 166: Why do we want to deserialize the `azure-key-vault` here? Don't we already do that with the ConverterFactory?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+                    AKVRetryPolicyMode.Fixed => RetryMode.Fixed,
+                    AKVRetryPolicyMode.Exponential => RetryMode.Exponential,
+                    null => RetryMode.Exponential,
+                    _ => RetryMode.Exponential
+                };
+
+                clientOptions.Retry.Mode = retryMode;
+                clientOptions.Retry.MaxRetries = options.RetryPolicy.MaxCount ?? 3;
+                clientOptions.Retry.Delay = TimeSpan.FromSeconds(options.RetryPolicy.DelaySeconds ?? 1);
+                clientOptions.Retry.MaxDelay = TimeSpan.FromSeconds(options.RetryPolicy.MaxDelaySeconds ?? 16);
```
souvikghosh04: should the MaxDelay be more than the NetworkTimeout? also are these configurable or going to be fixed always?

src/Config/RuntimeConfigLoader.cs
```diff
@@ -129,25 +129,91 @@ protected void SignalConfigChanged(string message = "")
...
+    /// <summary>
+    /// Extracts AzureKeyVaultOptions from JSON string with configurable variable replacement.
+    /// </summary>
+    /// <param name="json">JSON that represents the config file.</param>
+    /// <param name="enableEnvReplacement">Whether to enable environment variable replacement during extraction.</param>
+    /// <param name="replacementFailureMode">Failure mode for environment variable replacement if enabled.</param>
+    /// <returns>AzureKeyVaultOptions if present, null otherwise.</returns>
+    private static AzureKeyVaultOptions? ExtractAzureKeyVaultOptions(string json, 
+        bool enableEnvReplacement,
+        Azure.DataApiBuilder.Config.Converters.EnvironmentVariableReplacementFailureMode replacementFailureMode = Azure.DataApiBuilder.Config.Converters.EnvironmentVariableReplacementFailureMode.Throw)
```
Aniruddh25: nit: do not need the prefix `Azure.DataApiBuilder.Config` since it is part of the same namespace.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -840,7 +840,7 @@ public void MsSqlConnStringSupplementedWithAppNameProperty(
             bool configParsed = RuntimeConfigLoader.TryParseConfig(
                 runtimeConfig.ToJson(),
                 out RuntimeConfig updatedRuntimeConfig,
-                replaceEnvVar: true);
+                new ());
```
Aniruddh25: nit: name the argument like

`replacementSettings: new()`

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -79,18 +78,18 @@ public void CheckConfigEnvParsingTest(
                 if (replaceEnvVar)
                 {
                     Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
-                        GetModifiedJsonString(repValues, @"""postgresql"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
+                        GetModifiedJsonString(repValues, @"""postgresql"""), out expectedConfig, replacementSettings: new DeserializationVariableReplacementSettings(azureKeyVaultOptions: null, doReplaceEnvVar: replaceEnvVar, doReplaceAKVVar: true)),
```
Aniruddh25: We should add similar tests for AKV replacement checks

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
```diff
@@ -0,0 +1,113 @@
...
+                        default:
+                            throw new JsonException($"Unexpected property {property}");
+                    }
+                }
+            }
+
+            throw new JsonException("Invalid AzureKeyVaultOptions format");
+        }
+
+        public override void Write(Utf8JsonWriter writer, AzureKeyVaultOptions value, JsonSerializerOptions options)
```
Aniruddh25, line 109: What changed in this file?

Aniruddh25 (changes requested):
The changes look good to me. This is a good refactor. However, unit testing still remains and so does the test in AKV environment.

src/Config/Converters/StringJsonConverterFactory.cs
aaronburtle: its applying the strategy and overwriting value with it, so we are getting a new string with each strategy applied in order.

src/Config/RuntimeConfigLoader.cs
aaronburtle: old code no longer needed.

src/Service.Tests/UnitTests/MySqlQueryExecutorUnitTests.cs
```diff
@@ -80,7 +81,8 @@ await provider.Initialize(
                         provider.GetConfig().ToJson(),
                         graphQLSchema: null,
                         connectionString: connectionString,
-                        accessToken: CONFIG_TOKEN);
+                        accessToken: CONFIG_TOKEN,
+                        new());
```
Aniruddh25: please name the argument - what is this new() supplied to?

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -114,7 +115,8 @@ await provider.Initialize(
                         provider.GetConfig().ToJson(),
                         graphQLSchema: null,
                         connectionString: connectionString,
-                        accessToken: CONFIG_TOKEN);
+                        accessToken: CONFIG_TOKEN,
+                        new());
```
Aniruddh25: name the parameter, and apply the same suggestion every other point

src/Service.Tests/UnitTests/PostgreSqlQueryExecutorUnitTests.cs
```diff
@@ -88,7 +89,8 @@ await provider.Initialize(
                         provider.GetConfig().ToJson(),
                         graphQLSchema: null,
                         connectionString: connectionString,
-                        accessToken: CONFIG_TOKEN);
+                        accessToken: CONFIG_TOKEN,
+                        new());
```
Aniruddh25: name the argument

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: they are independant, and I think based on the logic now we will do the env replacement first.

src/Config/RuntimeConfigLoader.cs
aaronburtle: for now will leave like this.

src/Config/RuntimeConfigLoader.cs
aaronburtle: we do it here because we cant do a replacement of the AKV variables for this section, so its done in a first pass.

src/Config/RuntimeConfigLoader.cs
aaronburtle: Yup, since this is part of a preprocess step only for the AKV section, we dont want an error to terminate progress, so we just return null.

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: theyre configurable, these are just the default values, ive aligned them to the issue.

https://github.com/Azure/data-api-builder/issues/2708

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
aaronburtle: the whole file is new.

src/Cli/Exporter.cs
```diff
@@ -62,9 +63,12 @@ public static bool Export(ExportOptions options, ILogger logger, FileSystemRunti
                 {
                     try
                     {
-                        ExportGraphQL(options, runtimeConfig, fileSystem, loader, logger).Wait();
-                        isSuccess = true;
-                        break;
+                        if (runtimeConfig is not null)
```
Aniruddh25: Why this change? The fact that TryLoadConfig returned true means runtimeConfig is going to be not null. Why do we need to check again here?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -116,21 +116,25 @@ public void TestInitializingRestAndGraphQLGlobalSettings()
...
+        DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true);
         Assert.IsTrue(_runtimeConfigLoader!.TryLoadConfig(
             TEST_RUNTIME_CONFIG_FILE,
             out RuntimeConfig? runtimeConfig,
-            replaceEnvVar: true));
+            replacementSettings: replacementSettings));
 
-        SqlConnectionStringBuilder builder = new(runtimeConfig.DataSource.ConnectionString);
-        Assert.AreEqual(ProductInfo.GetDataApiBuilderUserAgent(), builder.ApplicationName);
+        if (runtimeConfig is not null)
```
Aniruddh25: Same here, we are already Assert for true above. If the assert fails, it will error out. so runtimeConfig can never be null here.

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.ObjectModel;
```
Aniruddh25, line 6: This seems to be an unnecessary using statement

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+        return new AzureKeyVaultOptionsConverter(_replacementSettings);
+    }
+
+    private class AzureKeyVaultOptionsConverter : JsonConverter<AzureKeyVaultOptions>
+    {
+        // Determines whether to replace environment variable with its
+        // value or not while deserializing.
+        private readonly DeserializationVariableReplacementSettings? _replacementSettings;
+
+        /// <param name="replaceEnvVar">Whether to replace environment variable with its
```
Aniruddh25, line 43:
```suggestion
        /// <param name="replacementSettings">Whether to replace environment variable with its
```

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+                            {
+                                endpoint = reader.DeserializeString(_replacementSettings);
+                            }
+
+                            break;
+
+                        case "retry-policy":
+                            if (reader.TokenType is JsonTokenType.StartObject)
+                            {
+                                // Pass the replaceEnvVar setting to the retry policy converter
```
Aniruddh25: how is this setting passed to the retry policy converter?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+        // a valid environment variable name in certain shells.
+        public const string OUTER_ENV_PATTERN = @"@env\('.*?(?='\))'\)";
+        public const string OUTER_AKV_PATTERN = @"@AKV\('.*?(?='\))'\)";
+
+        // [^@env\(]   :  any substring that is not @env(
+        // [^@AKV\(]   :  any substring that is not @AKV(
+        // .*          :  any char except newline any number of times
+        // (?=\))      :  look ahead for end char of )
+        // This pattern greedy matches all characters that are not a part of @env()
+        // ie: @env('hello@env('goodbye')world') match: 'hello@env('goodbye')world'
```
Aniruddh25, line 40: In the match here, the substring @env( is a substring but the pattern [^@env\(] expects any substring that is not @env(. so how did it match?

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -172,7 +172,7 @@ private void OnNewFileContentsDetected(object? sender, EventArgs e)
         catch (Exception ex)
         {
             // Need to remove the dependencies in startup on the RuntimeConfigProvider
-            // before we can have an ILogger here.
+            // before we can have anILogger here.
```
Aniruddh25: nit: 
```suggestion
            // before we can have an ILogger here.
```

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -226,7 +225,15 @@ public bool TryLoadConfig(
                 }
             }
 
-            if (!string.IsNullOrEmpty(json) && TryParseConfig(json, out RuntimeConfig, connectionString: _connectionString, replaceEnvVar: replaceEnvVar))
+            // Use default replacement settings if none provided
+            replacementSettings ??= new DeserializationVariableReplacementSettings(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true);
```
Aniruddh25: Default for this function was to keep replaceEnvVar to false. But, here I see it is `true`, why is this change needed and if needed, did we ensure all the callers are updated accordingly?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -298,13 +298,19 @@ public RuntimeConfig(
...
+                DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true);
+
+                if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config, replacementSettings: replacementSettings))
                 {
                     try
                     {
-                        _dataSourceNameToDataSource = _dataSourceNameToDataSource.Concat(config._dataSourceNameToDataSource).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
-                        _entityNameToDataSourceName = _entityNameToDataSourceName.Concat(config._entityNameToDataSourceName).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
-                        allEntities = allEntities.Concat(config.Entities.AsEnumerable());
+                        if (config is not null)
```
Aniruddh25: why do we need to check for config is not null when TryLoadConfig is true?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2346,7 +2346,12 @@ public async Task TestSPRestDefaultsForManuallyConstructedConfigs(
            HttpStatusCode expectedResponseStatusCode)
         {
             string configJson = TestHelper.AddPropertiesToJson(TestHelper.BASE_CONFIG, entityJson);
-            RuntimeConfigLoader.TryParseConfig(configJson, out RuntimeConfig deserializedConfig, logger: null, GetConnectionStringFromEnvironmentConfig(environment: TestCategory.MSSQL));
+            RuntimeConfigLoader.TryParseConfig(
```
Aniruddh25, line 2349: for TryParseConfig, the default value of replaceEnvVar was false. But with the addition of the replacementSettings, the default value for this call ends up being true. This will break the previous assumptions. In case of refactors, the best idea is to maintain previous behavior to avoid regressions.

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
aaronburtle: needed for the `AzureKeyVaultOptions` type to be visible.

src/Config/Converters/AzureKeyVaultOptionsConverterFactory.cs
aaronburtle: comment isnt very clear here the settings are already registered, ive updated the comment to be more clear.

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: This is the pattern for collecting what is inside of @env('<string>'), the match initially uses the outer pattern, then we use the inner pattern to get out what is inside.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: since it is not null when true we dont need this check, removing.

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle: changed default behavior to false.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: will come with the file replacement logic.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -182,17 +182,16 @@ private void OnNewFileContentsDetected(object? sender, EventArgs e)
     /// </summary>
     /// <param name="path">The path to the dab-config.json file.</param>
     /// <param name="config">The loaded <c>RuntimeConfig</c>, or null if none was loaded.</param>
-    /// <param name="replaceEnvVar">Whether to replace environment variable with its
-    /// value or not while deserializing.</param>
     /// <param name="logger">ILogger for logging errors.</param>
     /// <param name="isDevMode">When not null indicates we need to overwrite mode and how to do so.</param>
+    /// <param name="replacementSettings">Settings for variable replacement during deserialization. If null, uses default settings with environment variable replacement enabled.</param>
```
Aniruddh25: With the latest changes, if this is null, the default settings for environment variable replacement is false. The comment should be fixed here.

src/Config/FileSystemRuntimeConfigLoader.cs
```diff
@@ -292,12 +299,16 @@ public bool TryLoadConfig(
...
-    /// <param name="replaceEnvVar">Whether to replace environment variable with its
-    /// value or not while deserializing.</param>
+    /// <param name="replacementSettings">Settings for variable replacement during deserialization. If null, uses default settings with environment variable replacement disabled.</param>
     /// <returns>True if the config was loaded, otherwise false.</returns>
     public override bool TryLoadKnownConfig([NotNullWhen(true)] out RuntimeConfig? config, bool replaceEnvVar = false)
     {
-        return TryLoadConfig(ConfigFilePath, out config, replaceEnvVar);
+        // Convert legacy replaceEnvVar parameter to replacement settings for backward compatibility
+        DeserializationVariableReplacementSettings? replacementSettings = replaceEnvVar
+            ? new DeserializationVariableReplacementSettings(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true)
```
Aniruddh25: cant this be simplified by 
`new DeserializationVariableReplacementSettings(azureKeyVaultOptions: null, doReplaceEnvVar: replaceEnvVar, doReplaceAkvVar: replaceEnvVar)`

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -79,18 +78,18 @@ public void CheckConfigEnvParsingTest(
...
                 {
                     Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
-                        GetModifiedJsonString(repKeys, @"""postgresql"""), out expectedConfig, replaceEnvVar: replaceEnvVar),
+                        GetModifiedJsonString(repKeys, @"""postgresql"""), out expectedConfig, replacementSettings: new DeserializationVariableReplacementSettings(azureKeyVaultOptions: null, doReplaceEnvVar: replaceEnvVar, doReplaceAkvVar: false)),
                         "Should read the expected config");
                 }
 
                 Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(
-                    GetModifiedJsonString(repKeys, @"""@env('enumVarName')"""), out RuntimeConfig actualConfig, replaceEnvVar: replaceEnvVar),
+                    GetModifiedJsonString(repKeys, @"""@env('enumVarName')"""), out RuntimeConfig actualConfig, replacementSettings: new DeserializationVariableReplacementSettings(azureKeyVaultOptions: null, doReplaceEnvVar: replaceEnvVar, doReplaceAkvVar: true)),
```
Aniruddh25: Are we testing replacement of akv variables here? If not, doReplaceAkvVar should be false here.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -302,10 +301,7 @@ public void CheckConfigEnvParsingThrowExceptions(string invalidEnvVarName)
         {
             string json = @"{ ""foo"" : ""@env('envVarName'), @env('" + invalidEnvVarName + @"')"" }";
             SetEnvVariables();
-            StringJsonConverterFactory stringConverterFactory = new(EnvironmentVariableReplacementFailureMode.Throw);
-            JsonSerializerOptions options = new() { PropertyNameCaseInsensitive = true };
```
Aniruddh25, line 306: Why dont you need these options anymore?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+        public const string INNER_ENV_PATTERN = @"[^@env\(].*(?=\))";
+        public const string INNER_AKV_PATTERN = @"[^@AKV\(].*(?=\))";
```
Aniruddh25: this should be `akv` in lower case as per the enhancement. same comment for the outer pattern.

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+            else
+            {
+                return value ?? match.Value;
+            }
+        }
+
+        private string ReplaceAkvVariable(Match match)
+        {
+            // strips first and last characters, ie: '''hello'' --> ''hello'
+            string name = Regex.Match(match.Value, INNER_AKV_PATTERN).Value[1..^1];
```
Aniruddh25, line 98: the name of the AKV variable here should follow certain rules as per the enhancement: https://github.com/Azure/data-api-builder/issues/2708

Look at section: `Key rules (based on AKV docs)`

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+            else
+            {
+                return value ?? match.Value;
+            }
+        }
+
+        private static SecretClient CreateSecretClient(AzureKeyVaultOptions options)
+        {
+            if (string.IsNullOrWhiteSpace(options.Endpoint))
```
Aniruddh25, line 165: Need to test if child configs do infact inherit the akv endpoint values from the parent config - see the  multiple configs section of the enhancement.

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -0,0 +1,167 @@
...
+            EnvFailureMode = envFailureMode;
+
+            if (DoReplaceEnvVar)
+            {
+                ReplacementStrategies.Add(
+                    new Regex(OUTER_ENV_PATTERN, RegexOptions.Compiled),
+                    ReplaceEnvVariable);
+            }
+
+            if (DoReplaceAkvVar && _azureKeyVaultOptions is not null)
```
Aniruddh25, line 67: A consideration in the enhancement is to add an OTEL activity wrapping the replacement of the AKV variable. Can you check if this can be done? Could be in a followup task.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: we do, added back.

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -18,32 +18,28 @@ public class DeserializationVariableReplacementSettings
...
-        // [^@AKV\(]   :  any substring that is not @AKV(
+        // [^@akv\(]   :  any substring that is not @akv(
         // .*          :  any char except newline any number of times
         // (?=\))      :  look ahead for end char of )
-        // This pattern greedy matches all characters that are not a part of @env()
+        // This pattern greedy matches all characters that are not a part of @env() / @akv()
         // ie: @env('hello@env('goodbye')world') match: 'hello@env('goodbye')world'
         public const string INNER_ENV_PATTERN = @"[^@env\(].*(?=\))";
-        public const string INNER_AKV_PATTERN = @"[^@AKV\(].*(?=\))";
+        public const string INNER_AKV_PATTERN = @"[^@akv\(].*(?=\))"; // changed from [^@AKV\(]
```
Aniruddh25:
```suggestion
        public const string INNER_AKV_PATTERN = @"[^@akv\(].*(?=\))"; 
```

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -18,32 +18,28 @@ public class DeserializationVariableReplacementSettings
...
-        // a valid environment variable name in certain shells.
+        // This pattern lazy matches any string that starts with @env(' and ends with ') OR @akv(' and ends with ')
+        // Example: fooBAR@env('hello-world')bash)FOO')  match: @env('hello-world')
+        // Example: fooBAR@akv('secret-name')bash)FOO') match: @akv('secret-name')
+        // This matching pattern allows for the @env('<match>') / @akv('<match>') to be safely nested
+        // within strings that contain ')' after our match.
+        // Note that there is no escape character currently for ')' to exist within the name of the variable.
         public const string OUTER_ENV_PATTERN = @"@env\('.*?(?='\))'\)";
-        public const string OUTER_AKV_PATTERN = @"@AKV\('.*?(?='\))'\)";
+        public const string OUTER_AKV_PATTERN = @"@akv\('.*?(?='\))'\)"; // changed from @AKV to @akv
```
Aniruddh25:
```suggestion
        public const string OUTER_AKV_PATTERN = @"@akv\('.*?(?='\))'\)";
```
seems like copilot made those changes and left comments in your code, which should have been cleaned up before publishing for PR.

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: Manually verified this works and ive added a unit test for it as well now.

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: added validation logic for these rules on the key

src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
```diff
@@ -187,7 +188,7 @@ public void TestSourceDefinitionCyclicObjectsSerializationDeserialization()
 
             _sourceDefinition.SourceEntityRelationshipMap.Add("persons", metadata);
 
-            // In serialization options we need  ReferenceHandler = ReferenceHandler.Preserve, or else it doesnot seialize objects with cycle references
+            // In serialization options we need  ReferenceHandler = ReferenceHandler.Preserve, or else it doesnot seiliate objects with cycle references
```
Aniruddh25:
```suggestion
            // In serialization options we need  ReferenceHandler = ReferenceHandler.Preserve, or else it does not serialize objects with cycle references
```

Aniruddh25 (approved):
LGTM.

RubenCerna2079 (approved):

PR #2887, opened 2025-09-25
Fix for stored procedure empty cell bug

gilemos:
## Why make this change?
 We have recently noticed that, if we have a column of type NVARCHAR or VARCHAR and we try to run a stored procedure that reads a row in which that column has an empty string value, we had an internal server error. 

This error happens when we try to run the method GetChars passing in a buffer with length 0

This PR aims to fix this problem

## What is this change?

We have added a small change to the method that was throwing the exception. 
If we find that resultFieldSize is 0 - which means that the data in the cell we are reading has a length of 0 - we will not call the method GetChars to read the data, but assume the data is empty and return the size of the data read in bytes as 0.

As you can see in the example bellow, that fixes the issue.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)

We have a table with a column of type NVARCHAR called "Description". In one of the rows, Description is an empty string

image

**Before the changes:**
If we try to run a stored procedure that reads that empty cell, we get an error
image

**After changes**
Stored procedure runs as expected
image

### DIFF
--- a/src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
+++ b/src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
@@ -3,7 +3,6 @@
 
 using System.Text.Json;
 using System.Threading.Tasks;
-using Azure.DataApiBuilder.Config.ObjectModel;
 using Azure.DataApiBuilder.Core.Resolvers;
 using Azure.DataApiBuilder.Service.Exceptions;
 using Azure.DataApiBuilder.Service.GraphQLBuilder.Queries;


src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -740,6 +740,12 @@ internal int StreamCharData(DbDataReader dbDataReader, long availableSize, Strin
             // else we throw exception.
             ValidateSize(availableSize, resultFieldSize);
 
+            // If the cell is empty, don't append anything to the resultJsonString and return 0.
+            if (resultFieldSize == 0)
```
vadeveka, line 744: Can we include a unit test or integration test to cover this scenario?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -740,6 +740,12 @@ internal int StreamCharData(DbDataReader dbDataReader, long availableSize, Strin
             // else we throw exception.
             ValidateSize(availableSize, resultFieldSize);
 
+            // If the cell is empty, don't append anything to the resultJsonString and return 0.
```
vadeveka, line 743: This was w.r.t. query on Stored procedure, is this handled well for table/view queries?

src/Core/Resolvers/QueryExecutor.cs
gilemos: Added the test!

src/Core/Resolvers/QueryExecutor.cs
gilemos: Tested a table query - will test the views for sanity checking :)

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -740,6 +740,12 @@ internal int StreamCharData(DbDataReader dbDataReader, long availableSize, Strin
...
+            if (resultFieldSize == 0)
+            {
+                return 0;
```
Aniruddh25, line 746: Can you please do a similar fix in the `StreamByteData` function below?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -740,6 +740,12 @@ internal int StreamCharData(DbDataReader dbDataReader, long availableSize, Strin
...
+            }
+
```
Aniruddh25, line 748: Could you also add a regression test in our test pipeline? You can take a look at existing tests on how to add new. Please reach out to Ruben Cerna if you need help.

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -623,6 +623,50 @@ public void ValidateStreamingLogicForStoredProcedures(int readDataLoops, bool ex
...
+        [DataTestMethod, TestCategory(TestCategory.MSSQL)]
+        public void ValidateStreamingLogicForEmptyCellsAsync()
+        {
+            TestHelper.SetupDatabaseEnvironment(TestCategory.MSSQL);
+            FileSystem fileSystem = new();
+            FileSystemRuntimeConfigLoader loader = new(fileSystem);
+            RuntimeConfig runtimeConfig = new(
+                Schema: "UnitTestSchema",
+                DataSource: new DataSource(DatabaseType: DatabaseType.MSSQL, "", Options: null),
+                Runtime: new(
```
RubenCerna2079, line 638:
```suggestion
                Runtime: new(
                        Rest: new(),
                        GraphQL: new(),
                        Mcp: new(),
                        Host: new(Cors: null, Authentication: null, MaxResponseSizeMB: 5)
                    ),
```

src/Core/Resolvers/QueryExecutor.cs
gilemos: Done!

src/Core/Resolvers/QueryExecutor.cs
gilemos: Added a test :)

src/Service.Tests/DatabaseSchema-MySql.sql
```diff
@@ -388,7 +388,8 @@ INSERT INTO books(id, title, publisher_id)
         (17, 'CONN%_CONN', 1234),
         (18, '[Special Book]', 1234),
         (19, 'ME\\YOU', 1234),
-        (20, 'C:\\\\LIFE', 1234);
+        (20, 'C:\\\\LIFE', 1234),
```
vadeveka, line 391: It would be more relevant to update mssql test which is where this issue was discovered, we can keep this for mysql if we want

gilemos: oh that was my mistake! I meant to update the MsSql

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
```diff
@@ -244,7 +244,7 @@ public async Task TestStoredProcedureMutationForInsertion(string dbQuery)
         /// <code>Check: </code> If the intended book is deleted from the DB and
         /// verifies the response.
         /// </summary>
-        public async Task TestStoredProcedureMutationForDeletion(string dbQueryToVerifyDeletion)
+        public async Task TestStoredProcedureMutationForDeletion(string dbQueryToVerifyDeletion, int expectedOriginalMaxId, int expectedFinalMaxId)
```
vadeveka: Touching existing tests might be hard to track. It might be useful to keep the tests and the concerns separate

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -84,94 +85,193 @@ public async Task RequestMaxUsingNegativeOne()
...
+                      ""id"": 20,
+                      ""title"": ""C:\\\\LIFE""
+                    }
+                  ],
+                  ""endCursor"": null,
+                  ""hasNextPage"": false
+                }";
+            }
+
+            // Note: The max page size is 21 for MsSql and 20 for all other data sources, so when using -1
```
Aniruddh25, line 178: why not make the test change for the other dbs as well? It will be a quick update to the schema files and the tests should work. This way we avoid hard coding two different constant expected depending on the db type.

src/Service.Tests/SqlTests/GraphQLMutationTests/GraphQLMutationTestBase.cs
gilemos: Fixed!

Aniruddh25 (approved):

src/Service.Tests/SqlTests/GraphQLPaginationTests/GraphQLPaginationTestBase.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Text.Json;
 using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config.ObjectModel;
```
RubenCerna2079: The pipeline is currently failing due to a formatting error. Delete this unused library.

gilemos: Good catch! Done

Issue #2829, opened 2025-08-20
[Enh]: Add DML tool: update_entity


JerryNixon:
### update_entity

A tool to allow an agent to update a single record.

### Behavior

1. **Behaves most like the GraphQL flow**

2. **Entity permissions are enforced**

   * Never bypass permissions.
   * On violation, return a standard DAB response payload with an error message (e.g. `"You do not have permission to update this entity"`).
   * Do **not** signal errors via HTTP status codes.

3. **Supports partial field lists**

   * Requests may specify only a subset of fields to update.
   * Fields not included in the request remain unchanged in the database.
   * If supplied fields violate constraints (nulls, types, etc.), allow the database to fail and return the resulting error in a standard DAB response payload.

4. **Returns the updated entity**

   * The response includes only the combination of keys and fields supplied in the request.
   * Example: a table with 100 columns, request with 3 fields → response contains only those 3 fields plus the key(s).

5. **Key handling**

   * Keys use the entity’s mapped aliases.
   * Keys are required in the request.
   * Keys uniquely identify the record to update.
   * If no record matches the supplied keys, return a DAB error payload (e.g. `"No entity found with the given key"` or whatever our current message is).

6. **Field handling**

   * Fields use the entity’s mapped aliases.

### Input payload

```json
{
    "entity": "User",
    "keys": {
        "Id": 123
    },
    "fields": {
        "email": "new@email.com",
        "name": "Updated Jerry"
    }
}
```

### Output payload

This is a standard Data API builder (DAB) payload which includes errors.

## Why?

This allows agents to update existing records.

## Configuration

```json
{
   "ai": {
      "mcp": {
         "dml-tools": {
            "update-entity": {
               "enabled": "true | false" // (bool) default: true
            }
         }
      }
   }
}
```

### Command line

Add `dab --configure ai.mcp.dml-tools.update-entity.enabled true`

Issue #1816, opened 2023-10-16
[Enhancement] Bulk Insert via REST 


severussundar:
### Abstract

Support the insertion of multiple items to a single table in a single POST request

### Additional Details on Request Header and Body 

To be Added

severussundar:
Moving this to milestone 1.1 as this is planned post GA

JerryNixon:
I believe `InsertMany` is all that is necessary here.

1. New permission next to `Create` is `BulkInsert`. This must be a permission because bulk inserts remove constraints before inserting which isn't what you want to allow just anyone to do. 
2. Not sure how to indicate to the endpoint that a bulk operation is being passed in instead of a single operation. This is an implementation detail that will have to be figured out.

PR #2888, opened 2025-09-26
[MCP] Added the update-record tool implementation.

anushakolan:
## Why make this change?

This PR,
- Adds the tool implementation for the `update-record` MCP tool in `Azure.DataApiBuilder.Mcp` to support updating records via MCP using keys and field payloads with full validation and authorization.
- Adds a README to guide contributors on how to test MCP tools using the MCP Inspector.
- Improves predicate construction logic in `SqlUpdateQueryStructure.cs` to support mapped column fields.

## What is this change?
- Adds `docs/Testing/mcp-inspector-testing.md` with step-by-step instructions for running MCP Inspector locally, bypassing TLS, and connecting to MCP endpoints.
- Introduces `UpdateRecordTool.cs` under BuiltInTools, enabling record updates via MCP with full validation, authorization, and mutation engine integration.
- Updates `SqlUpdateQueryStructure.cs` to construct predicates using parameterized values for safer and more flexible SQL generation.

Exceptions are thrown when:
1. Entity is null or empty.
2. Keys and Fields are not JSON objects.
3. Failed to parse Keys and Fields.
4. 0 keys provided.
5. 0 fields provided.
6. Key value is null or empty
7. field value is null or empty.

Error:
1. PermissionDenied - No permssions to execute.
2. InvalidArguments - No arguments provided.
3. InvalidArguments - Some missing arguments.
4. EntityNotFound - Entity not defined in the configuration.
5. InvalidArguments - No entity found with given key.
6. UnexpectedError - Any other UnexpectedError.

## How was this tested?

- [x] Manual testing via MCP Inspector
- [ ] Integration Tests
- [ ] Unit Tests

These scenarios were manually tested using MCP Inspector, as automated tests are not yet implemented.

### Valid Cases ###

1. Successful Update
- Provided valid entity, keys, and fields.
- Verified that values were correctly updated in the database.
- Confirmed that the response includes all the column values similar to REST API calls.

2. Permission Enforcement
- Modified role permissions and verified that access control is enforced correctly.

3. Composite Key Handling
- Tested entities requiring multiple keys.
- Verified correct update behavior and response formatting.

4. Alias Mapping
- Used mapped field name (e.g., order instead of position) and confirmed that only mapped names are accepted.

### Failure Cases ###

1. Empty Keys
- Provided an empty keys object.
- Received appropriate error indicating missing key values.

2. Empty Fields
- Provided an empty fields object.
- Received error indicating that at least one field is required.

3. Invalid Entity Name
- Used a non-existent entity name.
- Received EntityNotFound error.

4. Empty Entity String
- Provided an empty string for entity.
- Received InvalidArguments error.

5. Entity Not Configured
- Used an entity not present in the configuration (e.g., Todos).
- Received EntityNotFound error.

6. Null Entity
- Provided null for the entity field.
- Received InvalidArguments error.

8. Valid Key Format but Not in DB
- Used correct key structure but no matching record exists.
- Received error indicating no entity found with the given key.

9. Invalid or Null Key Values
- Provided null or malformed key values.
- Received InvalidArguments error.

10. Unauthorized Role Context
- Removed or misconfigured role header.
- Received PermissionDenied error.

## Sample Request(s)

`{ "id": "00000000-0000-0000-0000-000000000001" }`
`{ "title": "New Title", "order": 7 }`

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,7 +1,7 @@
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
 # review when someone opens a pull request.
-*       @Aniruddh25 @aaronburtle @anushakolan @RubenCerna2079 @souvikghosh04 @akashkumar58 @neeraj-sharma2592 @sourabh1007 @vadeveka @Alekhya-Polavarapu @rusamant @stuartpa
+*       @Aniruddh25 @aaronburtle @anushakolan @RubenCerna2079 @souvikghosh04 @neeraj-sharma2592 @sourabh1007 @vadeveka @Alekhya-Polavarapu @rusamant @stuartpa
 
 code_of_conduct.md @jerrynixon
 contributing.md @jerrynixon
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
@@ -1,0 +1,346 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Data.Common;
+using System.Text.Json;
+using Azure.DataApiBuilder.Auth;
+using Azure.DataApiBuilder.Config.DatabasePrimitives;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Core.Configurations;
+using Azure.DataApiBuilder.Core.Models;
+using Azure.DataApiBuilder.Core.Resolvers;
+using Azure.DataApiBuilder.Core.Resolvers.Factories;
+using Azure.DataApiBuilder.Core.Services;
+using Azure.DataApiBuilder.Core.Services.MetadataProviders;
+using Azure.DataApiBuilder.Mcp.Model;
+using Azure.DataApiBuilder.Mcp.Utils;
+using Azure.DataApiBuilder.Service.Exceptions;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Protocol;
+using static Azure.DataApiBuilder.Mcp.Model.McpEnums;
+
+namespace Azure.DataApiBuilder.Mcp.BuiltInTools
+{
+    /// <summary>
+    /// Tool to delete records from a table/view entity configured in DAB.
+    /// Supports both simple and composite primary keys.
+    /// </summary>
+    public class DeleteRecordTool : IMcpTool
+    {
+        /// <summary>
+        /// Gets the type of the tool, which is BuiltIn for this implementation.
+        /// </summary>
+        public ToolType ToolType { get; } = ToolType.BuiltIn;
+
+        /// <summary>
+        /// Gets the metadata for the delete-record tool, including its name, description, and input schema.
+        /// </summary>
+        public Tool GetToolMetadata()
+        {
+            return new Tool
+            {
+                Name = "delete_record",
+                Description = "Deletes a record from a table based on primary key or composite key",
+                InputSchema = JsonSerializer.Deserialize<JsonElement>(
+                    @"{
+                        ""type"": ""object"",
+                        ""properties"": {
+                            ""entity"": {
+                                ""type"": ""string"",
+                                ""description"": ""The name of the entity (table) as configured in dab-config. Required.""
+                            },
+                            ""keys"": {
+                                ""type"": ""object"",
+                                ""description"": ""Primary key values to identify the record to delete. For composite keys, provide all key columns as properties. Required.""
+                            }
+                        },
+                        ""required"": [""entity"", ""keys""]
+                    }"
+                )
+            };
+        }
+
+        /// <summary>
+        /// Executes the delete-record tool, deleting an existing record in the specified entity using provided keys.
+        /// </summary>
+        public async Task<CallToolResult> ExecuteAsync(
+            JsonDocument? arguments,
+            IServiceProvider serviceProvider,
+            CancellationToken cancellationToken = default)
+        {
+            ILogger<DeleteRecordTool>? logger = serviceProvider.GetService<ILogger<DeleteRecordTool>>();
+
+            try
+            {
+                // Cancellation check at the start
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // 1) Resolve required services & configuration
+                RuntimeConfigProvider runtimeConfigProvider = serviceProvider.GetRequiredService<RuntimeConfigProvider>();
+                RuntimeConfig config = runtimeConfigProvider.GetConfig();
+
+                // 2) Check if the tool is enabled in configuration before proceeding
+                if (config.McpDmlTools?.DeleteRecord != true)
+                {
+                    return McpResponseBuilder.BuildErrorResult(
+                        "ToolDisabled",
+                        $"The {this.GetToolMetadata().Name} tool is disabled in the configuration.",
+                        logger);
+                }
+
+                // 3) Parsing & basic argument validation
+                if (arguments is null)
+                {
+                    return McpResponseBuilder.BuildErrorResult("InvalidArguments", "No arguments provided.", logger);
+                }
+
+                if (!McpArgumentParser.TryParseEntityAndKeys(arguments.RootElement, out string entityName, out Dictionary<string, object?> keys, out string parseError))
+                {
+                    return McpResponseBuilder.BuildErrorResult("InvalidArguments", parseError, logger);
+                }
+
+                IMetadataProviderFactory metadataProviderFactory = serviceProvider.GetRequiredService<IMetadataProviderFactory>();
+                IMutationEngineFactory mutationEngineFactory = serviceProvider.GetRequiredService<IMutationEngineFactory>();
+
+                // 4) Resolve metadata for entity existence check
+                string dataSourceName;
+                ISqlMetadataProvider sqlMetadataProvider;
+
+                try
+                {
+                    dataSourceName = config.GetDataSourceNameFromEntityName(entityName);
+                    sqlMetadataProvider = metadataProviderFactory.GetMetadataProvider(dataSourceName);
+                }
+                catch (Exception)
+                {
+                    return McpResponseBuilder.BuildErrorResult("EntityNotFound", $"Entity '{entityName}' is not defined in the configuration.", logger);
+                }
+
+                if (!sqlMetadataProvider.EntityToDatabaseObject.TryGetValue(entityName, out DatabaseObject? dbObject) || dbObject is null)
+                {
+                    return McpResponseBuilder.BuildErrorResult("EntityNotFound", $"Entity '{entityName}' is not defined in the configuration.", logger);
+                }
+
+                // Validate it's a table or view
+                if (dbObject.SourceType != EntitySourceType.Table && dbObject.SourceType != EntitySourceType.View)
+                {
+                    return McpResponseBuilder.BuildErrorResult("InvalidEntity", $"Entity '{entityName}' is not a table or view. Use 'execute-entity' for stored procedures.", logger);
+                }
+
+                // 5) Authorization
+                IAuthorizationResolver authResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
+
+                if (!McpAuthorizationHelper.ValidateRoleContext(httpContext, authResolver, out string roleError))
+                {
+                    return McpResponseBuilder.BuildErrorResult("PermissionDenied", $"Permission denied: {roleError}", logger);
+                }
+
+                if (!McpAuthorizationHelper.TryResolveAuthorizedRole(
+                    httpContext!,
+                    authResolver,
+                    entityName,
+                    EntityActionOperation.Delete,
+                    out string? effectiveRole,
+                    out string authError))
+                {
+                    return McpResponseBuilder.BuildErrorResult("PermissionDenied", $"Permission denied: {authError}", logger);
+                }
+
+                // 6) Build and validate Delete context
+                RequestValidator requestValidator = new(metadataProviderFactory, runtimeConfigProvider);
+
+                DeleteRequestContext context = new(
+                    entityName: entityName,
+                    dbo: dbObject,
+                    isList: false);
+
+                foreach (KeyValuePair<string, object?> kvp in keys)
+                {
+                    if (kvp.Value is null)
+                    {
+                        return McpResponseBuilder.BuildErrorResult("InvalidArguments", $"Primary key value for '{kvp.Key}' cannot be null.", logger);
+                    }
+
+                    context.PrimaryKeyValuePairs[kvp.Key] = kvp.Value;
+                }
+
+                requestValidator.ValidatePrimaryKey(context);
+
+                // 7) Execute
+                DatabaseType dbType = config.GetDataSourceFromDataSourceName(dataSourceName).DatabaseType;
+                IMutationEngine mutationEngine = mutationEngineFactory.GetMutationEngine(dbType);
+
+                IActionResult? mutationResult = null;
+                try
+                {
+                    // Cancellation check before executing
+                    cancellationToken.ThrowIfCancellationRequested();
+                    mutationResult = await mutationEngine.ExecuteAsync(context).ConfigureAwait(false);
+                }
+                catch (DataApiBuilderException dabEx)
+                {
+                    // Handle specific DAB exceptions
+                    logger?.LogError(dabEx, "Data API Builder error deleting record from {Entity}", entityName);
+
+                    string message = dabEx.Message;
+
+                    // Check for specific error patterns
+                    if (message.Contains("Could not find item with", StringComparison.OrdinalIgnoreCase))
+                    {
+                        string keyDetails = McpJsonHelper.FormatKeyDetails(keys);
+                        return McpResponseBuilder.BuildErrorResult(
+                            "RecordNotFound",
+                            $"No record found with the specified primary key: {keyDetails}",
+                            logger);
+                    }
+                    else if (message.Contains("violates foreign key constraint", StringComparison.OrdinalIgnoreCase) ||
+                             message.Contains("REFERENCE constraint", StringComparison.OrdinalIgnoreCase))
+                    {
+                        return McpResponseBuilder.BuildErrorResult(
+                            "ConstraintViolation",
+                            "Cannot delete record due to foreign key constraint. Other records depend on this record.",
+                            logger);
+                    }
+                    else if (message.Contains("permission", StringComparison.OrdinalIgnoreCase) ||
+                             message.Contains("authorization", StringComparison.OrdinalIgnoreCase))
+                    {
+                        return McpResponseBuilder.BuildErrorResult(
+                            "PermissionDenied",
+                            "You do not have permission to delete this record.",
+                            logger);
+                    }
+                    else if (message.Contains("invalid", StringComparison.OrdinalIgnoreCase) &&
+                             message.Contains("type", StringComparison.OrdinalIgnoreCase))
+                    {
+                        return McpResponseBuilder.BuildErrorResult(
+                            "InvalidArguments",
+                            "Invalid data type for one or more key values.",
+                            logger);
+                    }
+
+                    // For any other DAB exceptions, return the message as-is
+                    return McpResponseBuilder.BuildErrorResult(
+                        "DataApiBuilderError",
+                        dabEx.Message,
+                        logger);
+                }
+                catch (SqlException sqlEx)
+                {
+                    // Handle SQL Server specific errors
+                    logger?.LogError(sqlEx, "SQL Server error deleting record from {Entity}", entityName);
+                    string errorMessage = sqlEx.Number switch
+                    {
+                        547 => "Cannot delete record due to foreign key constraint. Other records depend on this record.",
+                        2627 or 2601 => "Cannot delete record due to unique constraint violation.",
+                        229 or 262 => $"Permission denied to delete from table '{dbObject.FullName}'.",
+                        208 => $"Table '{dbObject.FullName}' not found in the database.",
+                        _ => $"Database error: {sqlEx.Message}"
+                    };
+                    return McpResponseBuilder.BuildErrorResult("DatabaseError", errorMessage, logger);
+                }
+                catch (DbException dbEx)
+                {
+                    // Handle generic database exceptions (works for PostgreSQL, MySQL, etc.)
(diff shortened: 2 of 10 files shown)


docs/Testing/mcp-inspector-testing.md
```diff
@@ -0,0 +1,20 @@
...
+
+### 2. **Bypass TLS Verification (For Local Testing)**
+set NODE_TLS_REJECT_UNAUTHORIZED=0
+
+### 3. ** Open the inspector with pre-filled token.**
+http://localhost:6274/?MCP_PROXY_AUTH_TOKEN=<token>
+
+### 4. **How to use the tool**
+- Set the transport type "Streamable HTTP".
+- Set the URL "http://localhost:5000/mcp" and hit connect.
```
Aniruddh25, line 20: Mention as pre-requisite that dab mcp-server should be running before hitting connect here.

docs/Testing/mcp-inspector-testing.md
```diff
@@ -0,0 +1,20 @@
...
+Steps to run and test MCP tools using the https://www.npmjs.com/package/@modelcontextprotocol/inspector.
+
+### 1. **Install MCP Inspector** 
+npx @modelcontextprotocol/inspector
+
+### 2. **Bypass TLS Verification (For Local Testing)**
+set NODE_TLS_REJECT_UNAUTHORIZED=0
+
+### 3. ** Open the inspector with pre-filled token.**
+http://localhost:6274/?MCP_PROXY_AUTH_TOKEN=<token>
```
Aniruddh25, line 16: what is this token used for?

docs/Testing/mcp-inspector-testing.md
```diff
@@ -0,0 +1,20 @@
+
+# MCP Inspector Testing Guide
+
+Steps to run and test MCP tools using the https://www.npmjs.com/package/@modelcontextprotocol/inspector.
+
+### 1. **Install MCP Inspector** 
+npx @modelcontextprotocol/inspector
```
Aniruddh25, line 10: is nodejs a pre-requisite to be installed?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+        /// <summary>
+        /// Gets the metadata for the update_record tool, including its name, description, and input schema.
+        /// </summary>
+        public Tool GetToolMetadata()
+        {
+            return new Tool
+            {
+                Name = "update_record",
+                Description = "Updates an existing record in the specified entity. Requires 'keys' to locate the record and 'fields' to specify new values.",
+                InputSchema = JsonSerializer.Deserialize<JsonElement>(
```
Aniruddh25, line 50: https://github.com/Azure/data-api-builder/issues/2829 - please update all the references of `update-entity` in the issue to `update_record`

src/Core/Resolvers/Sql Query Structures/SqlUpdateQueryStructure.cs
```diff
@@ -185,9 +185,9 @@ private Predicate CreatePredicateForParam(KeyValuePair<string, object?> param)
             {
                 predicate = new(
                     new PredicateOperand(
-                        new Column(tableSchema: DatabaseObject.SchemaName, tableName: DatabaseObject.Name, param.Key)),
+                        new Column(tableSchema: DatabaseObject.SchemaName, tableName: DatabaseObject.Name, backingColumn)),
```
Aniruddh25, line 188: this seems to have been a bug here. Good catch!

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+            }
+            catch (Exception ex)
+            {
+                ILogger<UpdateRecordTool>? innerLogger = serviceProvider.GetService<ILogger<UpdateRecordTool>>();
+                innerLogger?.LogError(ex, "Unexpected error in UpdateRecordTool.");
+
+                return BuildErrorResult(
+                    "UnexpectedError",
+                    ex.Message ?? "An unexpected error occurred during the update operation.",
+                    logger: null);
```
Aniruddh25: why pass a null to the logger here? why not pass the `logger` obtained on line 85

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+            }
+
+            if (fields.Count == 0)
+            {
+                throw new ArgumentException("At least one field must be provided to update an entity", nameof(fields));
+            }
+
+            foreach (KeyValuePair<string, object?> kv in keys)
+            {
+                if (kv.Value is null || (kv.Value is string str && string.IsNullOrWhiteSpace(str)))
```
Aniruddh25, line 296: We should be ok if the key value is null or whitespace in the edge case scenario when the update is wanting to set the column value to null or whitespace.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                IAuthorizationResolver authResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
+
+                if (httpContext is null || !authResolver.IsValidRoleContext(httpContext))
+                {
+                    return BuildErrorResult("PermissionDenied", "Permission denied: unable to resolve a valid role context for update operation.", logger);
+                }
+
+                if (!TryResolveAuthorizedRole(httpContext, authResolver, entityName, out string? effectiveRole, out string authError))
```
aaronburtle: What is `effectiveRole` being used for here?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                {
+                    return BuildErrorResult("EntityNotFound", $"Entity '{entityName}' is not defined in the configuration.", logger);
+                }
+
+                if (!sqlMetadataProvider.EntityToDatabaseObject.TryGetValue(entityName, out DatabaseObject? dbObject) || dbObject is null)
+                {
+                    return BuildErrorResult("EntityNotFound", $"Entity '{entityName}' is not defined in the configuration.", logger);
+                }
+
+                // 4) Authorization after we have a known entity
```
Aniruddh25: We need to ensure the ClientRoleHeaderAuthenticationHeaderMiddleware is invoked https://github.com/Azure/data-api-builder/blob/a86996d073ffb306ecea9ae615065a4359de017d/src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs#L26 
before Authorization resolver kicks in here. 

You should perform some testing using the simulator to check what happens for `authenticated` roles. `anonymous` role is by default.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+            out string error)
+        {
+            effectiveRole = null;
+            error = string.Empty;
+
+            string roleHeader = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER].ToString();
+
+            if (string.IsNullOrWhiteSpace(roleHeader))
+            {
+                error = "Client role header is missing or empty.";
```
Aniruddh25, line 319: If the `ClientRoleHeaderAuthenticationMiddleware` isn't invoked, the roleheader here will be empty. Please test that out.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                if (kv.Value is null || (kv.Value is string str && string.IsNullOrWhiteSpace(str)))
+                {
+                    throw new ArgumentException("Keys are required to update an entity");
+                }
+            }
+
+            return true;
+        }
+
+        private static bool TryResolveAuthorizedRole(
```
Aniruddh25:
```suggestion
        private static bool TryResolveAuthorizedRoleHasPermission(
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
Aniruddh25: nvm this is for the Primary Key which should be non-null. So, you can keep this error as is.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+
+                context.UpdateReturnFields(keys.Keys.Concat(fields.Keys).Distinct().ToList());
+
+                foreach (KeyValuePair<string, object?> kvp in keys)
+                {
+                    if (kvp.Value is null)
+                    {
+                        return BuildErrorResult("InvalidArguments", $"Primary key value for '{kvp.Key}' cannot be null.", logger);
+                    }
+
```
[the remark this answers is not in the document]
Aniruddh25: this suggestion from copilot is valid and should be addressed

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                    context.PrimaryKeyValuePairs[kvp.Key] = kvp.Value;
+                }
+
+                if (context.DatabaseObject.SourceType is EntitySourceType.Table)
```
Aniruddh25, line 174: Why is this validation restricted to entity type of Table?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                JsonElement upsertPayloadRoot = RequestValidator.ValidateAndParseRequestBody(JsonSerializer.Serialize(fields));
+                RequestValidator requestValidator = new(metadataProviderFactory, runtimeConfigProvider);
+
+                UpsertRequestContext context = new(
+                    entityName: entityName,
+                    dbo: dbObject,
+                    insertPayloadRoot: upsertPayloadRoot,
+                    operationType: EntityActionOperation.UpdateIncremental);
+
+                context.UpdateReturnFields(keys.Keys.Concat(fields.Keys).Distinct().ToList());
```
Aniruddh25: I thought we decided to return the entire record in the response to align with what we have in REST. In REST, do we only return the keys?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                }
+                catch (Exception ex)
+                {
+                    string errorMsg = ex.Message ?? string.Empty;
+
+                    if (errorMsg.Contains("No Update could be performed, record not found", StringComparison.OrdinalIgnoreCase))
+                    {
+                        return BuildErrorResult(
+                            "InvalidArguments",
+                            "No entity found with the given key.",
```
Aniruddh25:
```suggestion
                            "No record found with the given key.",
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                {
+                    new TextContentBlock { Type = "text", Text = output }
+                }
+            };
+        }
+
+        /// <summary>
+        /// Converts JsonElement to .NET object dynamically.
+        /// </summary>
+        private static object? GetJsonValue(JsonElement element)
```
Aniruddh25, line 399: This function GetJsonValue does not have any references, please remove. even Copilot did not find it.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+        {
+            // Extract only requested keys and updated fields from engineRootElement
+            Dictionary<string, object?> filteredResult = new();
+
+            // Navigate to "value" array in the engine result
+            if (engineRootElement.TryGetProperty("value", out JsonElement valueArray) &&
+                valueArray.ValueKind == JsonValueKind.Array &&
+                valueArray.GetArrayLength() > 0)
+            {
+                JsonElement firstItem = valueArray[0];
```
Aniruddh25, line 367: This is assuming there will only be one element in the Value Array.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                {
+                    filteredResult[prop.Name] = GetJsonValue(prop.Value);
+                }
+            }
+
+            // Build normalized response
+            Dictionary<string, object?> normalized = new()
+            {
+                ["status"] = "success",
+                ["result"] = filteredResult // only requested values
```
Aniruddh25: We are no longer sending back only the requested values. The comment should be updated.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+                requestValidator.ValidatePrimaryKey(context);
+
+                // 6) Execute
+                DatabaseType dbType = config.GetDataSourceFromDataSourceName(dataSourceName).DatabaseType;
+                IMutationEngine mutationEngine = mutationEngineFactory.GetMutationEngine(dbType);
+
+                IActionResult? mutationResult = null;
+                try
+                {
+                    mutationResult = await mutationEngine.ExecuteAsync(context).ConfigureAwait(false);
```
Aniruddh25, line 188: This ExecuteAsync is specifically creating the response as expected in a Rest Request. We should understand it is an MCP response and craft it as per the MCP endpoint needs instead of deserializing and serializing again.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
Aniruddh25: I see it now in BuildSuccessResult, closing this comment.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
Aniruddh25: We should use the `SqlResponseHelper` to create a response that MCP endpoint needs to avoid the duplicate serialization and deserialization.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+
+            string output = JsonSerializer.Serialize(normalized, new JsonSerializerOptions { WriteIndented = true });
+
+            logger?.LogInformation("UpdateRecordTool success for entity {Entity}.", entityName);
+
+            return new CallToolResult
+            {
+                Content = new List<ContentBlock>
+                {
+                    new TextContentBlock { Type = "text", Text = output }
```
Aniruddh25, line 391: Should the tool call result be of Type  `text` or `json`?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,463 @@
...
+        /// <param name="arguments">The JSON arguments containing entity, keys, and fields.</param>
+        /// <param name="serviceProvider">The service provider for resolving dependencies.</param>
+        /// <param name="cancellationToken">A token to cancel the operation.</param>
+        /// <returns>A <see cref="CallToolResult"/> representing the outcome of the update operation.</returns>
+        public async Task<CallToolResult> ExecuteAsync(
+            JsonDocument? arguments,
+            IServiceProvider serviceProvider,
+            CancellationToken cancellationToken = default)
+        {
+            ILogger<UpdateRecordTool>? logger = serviceProvider.GetService<ILogger<UpdateRecordTool>>();
```
Aniruddh25, line 85: How do we ensure the configuration check to enable the tool is propagated to whether we expose the tool or not?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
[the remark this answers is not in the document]
anushakolan: This is needed because we will get a,

object? KeyValuePair<string, object?>.Value { get; }
Gets the value in the key/value pair.

Returns:
  A object that is the value of the KeyValuePair<TKey, TValue>.
'Value' may be null here.

error if we remove the check

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: The caller of `TryResolveAuthorizedRole` can know which role was actually authorized, not just that authorization succeeded.
This is important if the client has multiple roles, and you want to track or audit which one was used for the operation.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: The method `context.UpdateReturnFields(keys.Keys.Concat(fields.Keys).Distinct().ToList())`; is intended to tell the mutation engine which fields to include in the response.
However, in our code, `BuildSuccessResult` method, we are not using this list. Instead, weextract all properties from the result:
`foreach (JsonProperty prop in firstItem.EnumerateObject())
{
    filteredResult[prop.Name] = GetJsonValue(prop.Value);
}`

So, we are returning all values. I verified this behaviour with REST, REST also returns all values.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: If you tried to run this validation on a non-table entity, you could encounter errors or incorrect behavior because the  logic in` ValidateUpsertRequestContext` is tightly coupled to the concept of database tables and their columns.

This is similar to how REST handled it.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: `ValueKind = Array : "[{"id":"00000000-0000-0000-0000-000000000004","title":"Update project documentation","completed":false,"owner_id":"Anusha Kolan","order":3}]"`

This is the value array, this will always be 1. Because in DAB or in the update operation we can only update 1 record at once, the one record that is uniquely identified by the key.

`ValueKind = Array : "[{"BookID":3,"AuthorID":104,"ContributionRole":"Author2"}]"` even in case of a composite key there will be only element which is a JSON of 2 keys.

`firstItem-> ValueKind = Object : "{"BookID":3,"AuthorID":104,"ContributionRole":"Author2"}"`

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: Will do part of a follow up PR.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: There is no `json` type, `text` is the only accepted type and this is working exactly as expected with MCP server like Claude.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: Good point, normally the tool should not be accessible from list tools itself when it is disabled, so this should not be an issue.
But I have also added an additional check to see if the tool is enabled and only then proceed.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: Will do as part of follow up PR.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
anushakolan: Will do as part of follow up PR.

docs/Testing/mcp-inspector-testing.md
anushakolan: Yes, will add it.

docs/Testing/mcp-inspector-testing.md
anushakolan: This token is used to authenticate your session and authorize access to MCP tools.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
```diff
@@ -0,0 +1,471 @@
...
+                }
+            };
+
+            string output = JsonSerializer.Serialize(errorObj);
+
+            logger?.LogWarning("UpdateRecordTool error {ErrorType}: {Message}", errorType, message);
+
+            return new CallToolResult
+            {
+                Content =
```
RubenCerna2079, line 433: There is a boolean variable called `IsError`, please use it in here so that it outputs the ToolResult as an error.

anushakolan: Done, thanks for sharing this.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/UpdateRecordTool.cs
Aniruddh25: We have a backlog item where we want to support doing bulk updates. When we add that functionality , we will have to make sure this piece of code is fixed as well. Please make a note of this - in the backlog item https://github.com/Azure/data-api-builder/issues/1816

anushakolan:
> LGTM. Thanks for addressing the comments and your patience. Please do create followup tasks for addressing the improvements.

Yes Ani, I will create the follow up tasks.

Issue #2830, opened 2025-08-20
[Enh]: Add DML tool: `delete_record`


JerryNixon:
### delete_record

A tool to allow an agent to delete a single record.

### Behavior

1. **Behaves most like the GraphQL flow**

2. **Entity permissions are enforced**

   * Never bypass permissions.
   * On violation, return a standard DAB response payload with an error message (e.g. `"You do not have permission to delete this record"`).
   * Do **not** signal errors via HTTP status codes.

3. **Key handling**

   * Keys use the entity’s mapped aliases.
   * Keys are required in the request.
   * Keys uniquely identify the record to delete.
   * If no record matches the supplied keys, return a DAB error payload (e.g. `"No record found with the given key"`).

4. **Returns the delete result**

   * Follow the current GraphQL delete behavior.
   * Standard response is `{ "status": "okay" }` on success.

5. **Field handling**

   * Fields are not valid for this request.

## How

* [ ] Add `delete_record` MCP tool through GraphQL flow
* [ ] Add config property (`runtime.mcp.dml-tools.delete-record.enabled=true`)
* [ ] Update JSON Schema with `runtime.mcp.dml-tools.delete-record.enabled`
* [ ] Obey configuration (`runtime.mcp.dml-tools.delete-record.enabled=true`)
* [ ] Add CLI `dab configure --runtime.mcp.dml-tools.delete-record.enabled true`
* [ ] Update `dab validate` (warn when `delete-record.enabled` and not `mcp.enabled`)

### Tool method

```csharp
/// <summary>
/// Deletes a single record from the specified entity. 
/// Keys must uniquely identify the record. 
/// Returns success or error. 
/// </summary>
Task<DabResponse> DeleteRecordAsync(
    [Description("The entity name of the record to delete. "
               + "Must match an entity returned by list_entities with delete=true. Required.")]
    string entity,

    [Description("A dictionary of key/value pairs that uniquely identify the record to delete. "
               + "Keys must match those defined in entity metadata. Required.")]
    IDictionary<string, object?> keys);
```

**Parameters**

* **entity** *(string, required)*

```csharp
if (string.IsNullOrWhiteSpace(entity))
    throw new ArgumentException("Entity is required", nameof(entity));
if (!entityMetadata.Delete)
    throw new UnauthorizedAccessException($"Entity {entity} cannot be deleted.");
```

* **keys** *(IDictionary<string, object?>, required)*

```csharp
if (keys == null || keys.Count == 0)
[...]

PR #2889, opened 2025-09-26
[MCP] Adding delete_record

souvikghosh04:
### Closes on
  - delete_record: #2830 

## What is this change?

This PR implements built-int tool `delete_record` as part of built-in tools to support delete operation on a table entity.

- deletes one record at a time
- Entity name and Keys should be specified
- Keys supports having a single primary key or composite key (refer sample request below)
- Operation is performed based on permissions as configured in dab-config
- Success or Failure message response is generated on execution of the delete operation

## How was this tested?

Functional testing using Insomnia client by running DAB in localhost and local SQL DB database
- MCP endpoint: http://localhost:5000/mcp
- JSON payload (details below)
- Querying and validating data in local database

## Sample Request(s)

Delete by single Primary Key

```
POST: http://localhost:5000/mcp

{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "tools/call",
  "params": {
    "name": "delete_record",
    "arguments": {
      "entity": "Book",
      "keys": {
        "id": 5009
      }
    }
  }
}


```

Delete by a composite key

```

{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "tools/call",
  "params": {
    "name": "delete_record",
    "arguments": {
      "entity": "Stock",
      "keys": {
        "categoryid": 10,
	     "pieceid": 1
      }
    }
  }
}

```

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
@@ -300,11 +300,28 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                     }
                 }
 
-                // Cancellation check before finalizing
-                cancellationToken.ThrowIfCancellationRequested();
-
                 // 8) Build response
-                return BuildDeleteSuccessResponse(entityName, keys, mutationResult, logger);
+                // Based on SqlMutationEngine, delete operations typically return NoContentResult
+                // We build a success response with just the operation details
+                Dictionary<string, object?> responseData = new()
+                {
+                    ["entity"] = entityName,
+                    ["keyDetails"] = McpJsonHelper.FormatKeyDetails(keys),
+                    ["message"] = "Record deleted successfully"
+                };
+
+                // If the mutation result is OkObjectResult (which would be unusual for delete),
+                // include the result value directly without re-serialization
+                if (mutationResult is OkObjectResult okObjectResult && okObjectResult.Value is not null)
+                {
+                    responseData["result"] = okObjectResult.Value;
+                }
+
+                return McpResponseBuilder.BuildSuccessResult(
+                    responseData,
+                    logger,
+                    $"DeleteRecordTool success for entity {entityName}."
+                );
             }
             catch (OperationCanceledException)
             {
@@ -324,40 +341,6 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                     "An unexpected error occurred during the delete operation.",
                     logger);
             }
-        }
-
-        private static CallToolResult BuildDeleteSuccessResponse(
-            string entityName,
-            Dictionary<string, object?> keys,
-            IActionResult? mutationResult,
-            ILogger? logger)
-        {
-            Dictionary<string, object?> responseData = new()
-            {
-                ["entity"] = entityName,
-                ["keyDetails"] = McpJsonHelper.FormatKeyDetails(keys),
-                ["message"] = "Record deleted successfully"
-            };
-
-            // Handle different result types
-            if (mutationResult is OkObjectResult okObjectResult)
-            {
-                string rawPayloadJson = McpResponseBuilder.ExtractResultJson(okObjectResult);
-                using JsonDocument resultDoc = JsonDocument.Parse(rawPayloadJson);
-                JsonElement root = resultDoc.RootElement;
-
-                Dictionary<string, object?> extractedData = McpJsonHelper.ExtractValuesFromEngineResult(root);
-                if (extractedData.Count > 0)
-                {
-                    responseData["result"] = extractedData;
-                }
-            }
-
-            return McpResponseBuilder.BuildSuccessResult(
-                responseData,
-                logger,
-                $"DeleteRecordTool success for entity {entityName}."
-            );
         }
     }
 }


src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+        public ToolType ToolType { get; } = ToolType.BuiltIn;
+
+        /// <summary>
+        /// Gets the metadata for the delete-record tool, including its name, description, and input schema.
+        /// </summary>
+        public Tool GetToolMetadata()
+        {
+            return new Tool
+            {
+                Name = "delete-record",
```
RubenCerna2079: I thought we agreed that the separations would be done with underscore `_`. I am wrong?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+                Description = "Deletes a record from a table based on primary key or composite key",
+                InputSchema = JsonSerializer.Deserialize<JsonElement>(
+                    @"{
+                        ""type"": ""object"",
+                        ""properties"": {
+                            ""entity"": {
+                                ""type"": ""string"",
+                                ""description"": ""The name of the entity (table) as configured in dab-config""
```
RubenCerna2079:
```suggestion
                                ""description"": ""The name of the entity (table) as configured in dab-config. Required.""
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+                            },
+                            ""keys"": {
+                                ""type"": ""object"",
+                                ""description"": ""Primary key values to identify the record to delete. For composite keys, provide all key columns as properties""
```
RubenCerna2079:
```suggestion
                                ""description"": ""Primary key values to identify the record to delete. For composite keys, provide all key columns as properties. Required.""
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+                {
+                    return McpResponseBuilder.BuildErrorResult("InvalidEntity", $"Entity '{entityName}' is not a table or view. Use 'execute-entity' for stored procedures.", logger);
+                }
+
+                // 4) Authorization
+                IAuthorizationResolver authResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
+
+                if (!McpAuthorizationHelper.ValidateRoleContext(httpContext, authResolver, out string roleError))
```
RubenCerna2079, line 139: Why should we give an `out string roleError` if there is only one possible error?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+            }
+            catch (OperationCanceledException)
+            {
+                return McpResponseBuilder.BuildErrorResult("OperationCanceled", "The delete operation was canceled.", logger: null);
+            }
+            catch (ArgumentException argEx)
+            {
+                return McpResponseBuilder.BuildErrorResult("InvalidArguments", argEx.Message, logger);
+            }
+            catch (Exception ex)
```
RubenCerna2079, line 334: Based on Jerry's suggestion we should also add a catch for `DbException`

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+            {
```
RubenCerna2079, line 335: I also think it would be a good idea to catch the `DataApiBuilderException` as we use many functions that were already implemented that throw this type of exception

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+                catch (Exception ex)
+                {
+                    string errorMsg = ex.Message ?? string.Empty;
+
+                    if (errorMsg.Contains("Could not find", StringComparison.OrdinalIgnoreCase) ||
+                        errorMsg.Contains("record not found", StringComparison.OrdinalIgnoreCase))
+                    {
+                        return McpResponseBuilder.BuildErrorResult(
+                            "RecordNotFound",
+                            "No entity found with the given key.",
```
RubenCerna2079: I think we should also return the key to the user in this message

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+                }
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // 7) Build response
+                return BuildDeleteSuccessResponse(entityName, keys, mutationResult, logger);
+            }
+            catch (OperationCanceledException)
+            {
+                return McpResponseBuilder.BuildErrorResult("OperationCanceled", "The delete operation was canceled.", logger: null);
```
RubenCerna2079: Is there a reason we want the logger to be null here?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,257 @@
...
+            }
+            catch (Exception ex)
+            {
+                ILogger<DeleteRecordTool>? innerLogger = serviceProvider.GetService<ILogger<DeleteRecordTool>>();
+                innerLogger?.LogError(ex, "Unexpected error in DeleteRecordTool.");
+
+                return McpResponseBuilder.BuildErrorResult(
+                    "UnexpectedError",
+                    ex.Message ?? "An unexpected error occurred during the delete operation.",
+                    logger: null);
```
RubenCerna2079: Is there a reason we want the logger to be null?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
souvikghosh04: Having out param allows us to extend `ValidateRoleContext()` handle different error scenarios and return the appropriate error message as out param. Currently, it has only one type of check and error message.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
souvikghosh04: you are right. the tool name itself will be exposed with an underscore naming convention, however, internally the tool will be configured with hyphen to follow the snake case convention in the code base. I have opened a task for this- https://github.com/Azure/data-api-builder/issues/2905

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,350 @@
...
+            try
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // 1) Parsing & basic argument validation
+                if (arguments is null)
+                {
+                    return McpResponseBuilder.BuildErrorResult("InvalidArguments", "No arguments provided.", logger);
+                }
+
```
RubenCerna2079, line 100: We need to check if the tool is enabled in the config file before we continue, you can check Anusha's or my PR as reference

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,350 @@
...
+                            logger);
+                    }
+                    else
+                    {
+                        // Re-throw unexpected exceptions
+                        throw;
+                    }
+                }
+
+                cancellationToken.ThrowIfCancellationRequested();
```
RubenCerna2079: Why do we need to have the cancellationToken twice?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
```diff
@@ -0,0 +1,350 @@
...
+                requestValidator.ValidatePrimaryKey(context);
+
+                // 6) Execute
+                DatabaseType dbType = config.GetDataSourceFromDataSourceName(dataSourceName).DatabaseType;
+                IMutationEngine mutationEngine = mutationEngineFactory.GetMutationEngine(dbType);
+
+                IActionResult? mutationResult = null;
+                try
+                {
+                    mutationResult = await mutationEngine.ExecuteAsync(context).ConfigureAwait(false);
```
RubenCerna2079, line 184: We should be using `SqlResponseHelper` after obtaining the `mutationResult` to create a response that MCP endpoint needs to avoid the duplicate serialization and deserialization.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
souvikghosh04: it serves different purposes-

- Beginning: Checks if cancellation was requested before starting any work. This is a quick fail-fast mechanism to avoid unnecessary processing.
- End: Checks if cancellation was requested after the mutation operation completes but before building the response. This ensures the operation respects cancellation even after the main work is done.

Ideally we should have a cancellation check just before execution as well which more important. I am making the changes in this PR for `delete-record` and other tools can follow.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
souvikghosh04: This is fixed for `delete-record` tool by removing `BuildDeleteSuccessResponse()`

Issue #2833, opened 2025-08-21
[Enh]: Add DML tool: `read_records`


JerryNixon:
### read\_records

A tool to allow an agent to read records from a single entity.

### Behavior

1. **Behaves most like the GraphQL flow**

2. **Entity permissions are enforced**

   * Never bypass permissions.
   * On violation, return a standard response payload with an error message.
   * Do not signal errors via HTTP status codes.

3. **Projection**

   * `select` is a CSV of field names to include in the response.
   * If `select` is null or empty, return all accessible fields.

   e.g.: `Id,Name,Email`

4. **Filtering**

   * `filter` is an expression string to reduce results.
   * Invalid filters should be passed through and return as an error payload.
   * ⚠️ Specific syntax needs team discussion!

5. **Paging**

   * `first` sets page size. It may be smaller or larger than the default, but never greater than the configured maximum.
   * `after` is a cursor token returned by a prior call. Paging is cursor-based.
   * If the anchor record for `after` was deleted between calls, the next page may be unavailable. Return a standard error payload.

6. **Sorting**

   * `orderby` is a list of fields with direction (`asc` or `desc`).
   * Sorting is recommended when paging to ensure stable page boundaries.

7. **Result shape**

   * Returns an array of projected entities. Projection follows `select`. If `select` is null, returns all accessible fields.

8. **Operation scope**

   * The result of this tool is ultimately a basic `SELECT` statement against the database.
   * Aggregate functions and advanced operations are supported by other tools that use the persisted document feature.

## How

* [ ] Add `read_records` MCP tool through graphql flow
* [x] Add config property (`runtime.mcp.dml-tools.read-records.enabled=true`) 
* [x] Update JSON Schema with `runtime.mcp.dml-tools.read-records.enabled`
* [ ] Obey configuration (`runtime.mcp.dml-tools.read-records.enabled=true`) 
* [ ] Add CLI `dab configure --runtime.mcp.dml-tools.read-records.enabled true`
* [ ] Update `dab validate` (warn when `read-records.enabled` and not `mcp.enabled`)

### Tool method

```csharp
/// <summary>
/// Reads records from the specified entity using projection, filtering, sorting, and paging. 
/// Returns an array of entities projected according to the request. 
/// </summary>
Task<DabResponse> ReadEntityAsync(
    [Description("The entity name to read from. Required.")]
    string entity,

    [Description("A CSV of field names to include in the response. If not provided, all fields are returned. Optional.")]
    string? select = null,

[...]

PR #2893, opened 2025-10-03
[MCP] Added read_records tool implementation

RubenCerna2079:
## Why make this change?
- #2833
One of the main DML tools that will be used is `read_records`, which will allow the MCP to read the records of an entity based on the needs of the user.

## What is this change?
- Create the `read_records` tool:
    - First it ensures that the MCP has the proper authorization to run the query
    - Then it creates the context with the parameters that were requested by the user
    - Lastly it calls the `SqlQueryEngine` in order to create and run the query and receive the results
- The `GenerateOrderByLists` function inside the `RequestParser.cs` file was changed from private to public in order to allow the `read_records` tool to also use it to generate the proper context for the query.
- Some functions inside of the `SqlResponseHelper.cs` file were changed to check if the query request comes from the `read_records` tool. This is done in order to output the correct information, right now the REST requests can also return a `nextLink` object which gives the API link necessary to get values in the case that not all of them were shown. We want to do something similar with the `read_records` tool, however we only want to return the `after` object which is the parameter that allows the query to know the exact place where it left from the previous query. This gets rid of unecessary information that can be found in the `nextLink` object.

Exceptions thrown when:
1. Entity is empty or null.
2. Parameters are not of the correct type.
3. Parameters are not correctly written.
4. Values inside `orderby` parameter are empty or null. (Note: `orderby` is an optional value as a whole, but the individual values it contains need to exist)
5. Not having necessary permissions.

Errors:
1. PermissionDenied - No permssions to execute.
2. InvalidArguments - No arguments provided.
6. InvalidArguments - Some missing arguments.
7. EntityNotFound - Entity not defined in the configuration.
8. UnexpectedError - Any other UnexpectedError.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests
- [x] Manual testing via MCP Inspecto

## Sample Request(s)
`{ Entity: Book }`
`{ Select: title,publisher_id }`
`{ First: 3 }`
`{ Orderby: ["publisher_id asc", "title desc"] }`

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
@@ -1,0 +1,379 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using Azure.DataApiBuilder.Auth;
+using Azure.DataApiBuilder.Config.DatabasePrimitives;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Core.Authorization;
+using Azure.DataApiBuilder.Core.Configurations;
+using Azure.DataApiBuilder.Core.Models;
+using Azure.DataApiBuilder.Core.Parsers;
+using Azure.DataApiBuilder.Core.Resolvers;
+using Azure.DataApiBuilder.Core.Resolvers.Factories;
+using Azure.DataApiBuilder.Core.Services;
+using Azure.DataApiBuilder.Core.Services.MetadataProviders;
+using Azure.DataApiBuilder.Mcp.Model;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Protocol;
+using static Azure.DataApiBuilder.Mcp.Model.McpEnums;
+
+namespace Azure.DataApiBuilder.Mcp.BuiltInTools
+{
+    public class ReadRecordsTool : IMcpTool
+    {
+        // private readonly IMetadataProviderFactory _metadataProviderFactory;
+        // private readonly IQueryEngineFactory _queryEngineFactory;
+
+        public ToolType ToolType { get; } = ToolType.BuiltIn;
+
+        public Tool GetToolMetadata()
+        {
+            return new Tool
+            {
+                Name = "read_records",
+                Description = "Reads the records from the specified entity.",
+                InputSchema = JsonSerializer.Deserialize<JsonElement>(
+                    @"{
+                        ""type"": ""object"",
+                        ""properties"": {
+                            ""entity"": {
+                                ""type"": ""string"",
+                                ""description"": ""The entity name to read from. Required.""
+                            },
+                            ""select"": {
+                                ""type"": ""string"",
+                                ""description"": ""A CSV of field names to include in the response. If not provided, all fields are returned. Optional.""
+                            },
+                            ""filter"": {
+                                ""type"": ""string"",
+                                ""description"": ""A filter expression string to restrict results. Optional.""
+                            },
+                            ""first"": {
+                                ""type"": ""integer"",
+                                ""description"": ""The maximum number of records to return in this page. Optional.""
+                            },
+                            ""orderby"": {
+                                ""type"": ""array"",
+                                ""items"": { ""type"": ""string"" },
+                                ""description"": ""A list of field names and directions for sorting (e.g., \""name asc\""). Optional.""
+                            },
+                            ""after"": {
+                                ""type"": ""string"",
+                                ""description"": ""A cursor token for retrieving the next page of results. Optional.""
+                            }
+                        }
+                    }"
+                )
+            };
+        }
+
+        public async Task<CallToolResult> ExecuteAsync(
+            JsonDocument? arguments,
+            IServiceProvider serviceProvider,
+            CancellationToken cancellationToken = default)
+        {
+            ILogger<ReadRecordsTool>? logger = serviceProvider.GetService<ILogger<ReadRecordsTool>>();
+
+            if (arguments == null)
+            {
+                return BuildErrorResult("InvalidArguments", "No arguments provided.", logger);
+            }
+
+            try
+            {
+                string entityName;
+                string? select = null;
+                string? filter = null;
+                int? first = null;
+                IEnumerable<string>? orderby = null;
+                string? after = null;
+
+                // Extract arguments
+                JsonElement root = arguments.RootElement;
+
+                if (!root.TryGetProperty("entity", out JsonElement entityElement) || string.IsNullOrWhiteSpace(entityElement.GetString()))
+                {
+                    return BuildErrorResult("InvalidArguments", "Missing requirement argument 'entity'.", logger);
+                }
+
+                entityName = entityElement.GetString()!;
+
+                if (root.TryGetProperty("select", out JsonElement selectElement))
+                {
+                    select = selectElement.GetString();
+                }
+
+                if (root.TryGetProperty("filter", out JsonElement filterElement))
+                {
+                    filter = filterElement.GetString();
+                }
+
+                if (root.TryGetProperty("first", out JsonElement firstElement))
+                {
+                    first = firstElement.GetInt32();
+                }
+
+                if (root.TryGetProperty("orderby", out JsonElement orderbyElement))
+                {
+                    orderby = (IEnumerable<string>?)orderbyElement.EnumerateArray().Select(e => e.GetString());
+                }
+
+                if (root.TryGetProperty("after", out JsonElement afterElement))
+                {
+                    after = afterElement.GetString();
+                }
+
+                // Get required services & configuration
+                IQueryEngineFactory queryEngineFactory = serviceProvider.GetRequiredService<IQueryEngineFactory>();
+                IMetadataProviderFactory metadataProviderFactory = serviceProvider.GetRequiredService<IMetadataProviderFactory>();
+                RuntimeConfigProvider runtimeConfigProvider = serviceProvider.GetRequiredService<RuntimeConfigProvider>();
+                RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
+
+                // Check metadata for entity exists
+                string dataSourceName;
+                ISqlMetadataProvider sqlMetadataProvider;
+
+                try
+                {
+                    dataSourceName = runtimeConfig.GetDataSourceNameFromEntityName(entityName);
+                    sqlMetadataProvider = metadataProviderFactory.GetMetadataProvider(dataSourceName);
+                }
+                catch (Exception)
+                {
+                    return BuildErrorResult("EntityNotFound", $"Entity '{entityName}' is not defined in the configuration.", logger);
+                }
+
+                if (!sqlMetadataProvider.EntityToDatabaseObject.TryGetValue(entityName, out DatabaseObject? dbObject) || dbObject is null)
+                {
+                    return BuildErrorResult("EntityNotFound", $"Entity '{entityName}' is not defined in the configuration.", logger);
+                }
+
+                // Authorization check in the existing entity
+                IAuthorizationResolver authResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
+
+                if (httpContext is null || !authResolver.IsValidRoleContext(httpContext))
+                {
+                    return BuildErrorResult("PermissionDenied", "You do not have permission to read records for this entity.", logger);
+                }
+
+                if (!TryResolveAuthorizedRole(httpContext, authResolver, entityName, out string? effectiveRole, out string authError))
+                {
+                    return BuildErrorResult("PermissionDenied", authError, logger);
+                }
+
+                // Build and validate Find context
+                RequestValidator requestValidator = new(metadataProviderFactory, runtimeConfigProvider);
+                FindRequestContext context = new(entityName, dbObject, true);
+
+                requestValidator.ValidateEntity(entityName);
+
+                IEnumerable<string> fieldsReturnedForFind;
+                if (!string.IsNullOrWhiteSpace(select))
+                {
+                    fieldsReturnedForFind = select.Split(",").ToList();
+                }
+                else
+                {
+                    fieldsReturnedForFind = authResolver.GetAllowedExposedColumns(context.EntityName, effectiveRole!, context.OperationType);
+                }
+
+                context.UpdateReturnFields(fieldsReturnedForFind);
+
+                if (!string.IsNullOrWhiteSpace(filter))
+                {
+                    string filterQueryString = $"?{RequestParser.FILTER_URL}={filter}";
+                    context.FilterClauseInUrl = sqlMetadataProvider.GetODataParser().GetFilterClause(filterQueryString, $"{context.EntityName}.{context.DatabaseObject.FullName}");
+                }
+
+                if (orderby is not null)
+                {
+                    string sortQueryString = $"?{RequestParser.SORT_URL}=";
+                    foreach (string param in orderby)
+                    {
+                        if (string.IsNullOrWhiteSpace(param))
+                        {
+                            return BuildErrorResult("InvalidArguments", "Parameters inside 'orderby' argument cannot be empty or null.", logger);
+                        }
+
+                        sortQueryString += $"{param}, ";
+                    }
+
+                    sortQueryString = sortQueryString.Substring(0, sortQueryString.Length - 2);
+                    (context.OrderByClauseInUrl, context.OrderByClauseOfBackingColumns) = RequestParser.GenerateOrderByLists(context, sqlMetadataProvider, sortQueryString);
+                }
+
+                context.First = first;
+                context.After = after;
+
+                // Execute
+                IQueryEngine queryEngine = queryEngineFactory.GetQueryEngine(sqlMetadataProvider.GetDatabaseType());
+                JsonDocument? queryResult = await queryEngine.ExecuteAsync(context);
+                IActionResult actionResult = queryResult is null ? SqlResponseHelpers.FormatFindResult(JsonDocument.Parse("[]").RootElement.Clone(), context, metadataProviderFactory.GetMetadataProvider(dataSourceName), runtimeConfigProvider.GetConfig(), httpContext, true)
+                                               : SqlResponseHelpers.FormatFindResult(queryResult.RootElement.Clone(), context, metadataProviderFactory.GetMetadataProvider(dataSourceName), runtimeConfigProvider.GetConfig(), httpContext, true);
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // Normalize response
+                string rawPayloadJson = ExtractResultJson(actionResult);
+                using JsonDocument result = JsonDocument.Parse(rawPayloadJson);
+                JsonElement queryRoot = result.RootElement;
+
+                return BuildSuccessResult(
+                    entityName,
+                    queryRoot.Clone(),
+                    logger);
+            }
+            catch (OperationCanceledException)
+            {
+                return BuildErrorResult("OperationCanceled", "The read operation was canceled.", logger: null);
+            }
+            catch (ArgumentException argEx)
+            {
+                return BuildErrorResult("InvalidArguments", argEx.Message, logger);
+            }
+            catch (Exception ex)
+            {
+                ILogger<ReadRecordsTool>? innerLogger = serviceProvider.GetService<ILogger<ReadRecordsTool>>();
+                innerLogger?.LogError(ex, "Unexpected error in ReadRecordTool.");
+
+                return BuildErrorResult(
+                    "UnexpectedError",
+                    "An unexpected error occurred while reading the record.",
+                    logger: null);
+            }
+        }
+
+        private static bool TryResolveAuthorizedRole(
+            HttpContext httpContext,
+            IAuthorizationResolver authorizationResolver,
+            string entityName,
+            out string? effectiveRole,
+            out string error)
+        {
+            effectiveRole = null;
+            error = string.Empty;
+
+            string roleHeader = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER].ToString();
+
+            if (string.IsNullOrWhiteSpace(roleHeader))
+            {
+                error = "Client role header is missing or empty.";
+                return false;
+            }
+
+            string[] roles = roleHeader
+                .Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                .Distinct(StringComparer.OrdinalIgnoreCase)
+                .ToArray();
+
+            if (roles.Length == 0)
+            {
+                error = "Client role header is missing or empty.";
+                return false;
+            }
+
+            foreach (string role in roles)
+            {
+                bool allowed = authorizationResolver.AreRoleAndOperationDefinedForEntity(
+                    entityName, role, EntityActionOperation.Read);
+
+                if (allowed)
+                {
+                    effectiveRole = role;
+                    return true;
+                }
+            }
+
+            error = "You do not have permission to read records for this entity.";
+            return false;
+        }
+
+        private static CallToolResult BuildSuccessResult(
+            string entityName,
+            JsonElement engineRootElement,
+            ILogger? logger)
+        {
+            // Build normalized response
+            Dictionary<string, object?> normalized = new()
+            {
+                ["status"] = "success",
+                ["result"] = engineRootElement // only requested values
+            };
+
+            string output = JsonSerializer.Serialize(normalized, new JsonSerializerOptions { WriteIndented = true });
+
+            logger?.LogInformation("UpdateRecordTool success for entity {Entity}.", entityName);
+
+            return new CallToolResult
+            {
+                Content = new List<ContentBlock>
+                {
+                    new TextContentBlock { Type = "text", Text = output }
+                }
+            };
+        }
+
+        private static CallToolResult BuildErrorResult(
+            string errorType,
+            string message,
+            ILogger? logger)
+        {
+            Dictionary<string, object?> errorObj = new()
+            {
+                ["status"] = "error",
+                ["error"] = new Dictionary<string, object?>
+                {
+                    ["type"] = errorType,
+                    ["message"] = message
+                }
+            };
+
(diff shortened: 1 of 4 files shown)


src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                }
+
+                if (root.TryGetProperty("first", out JsonElement firstElement))
+                {
+                    first = firstElement.GetInt32();
+                }
+
+                if (root.TryGetProperty("orderby", out JsonElement orderbyElement))
+                {
+                    orderby = (IEnumerable<string>?)orderbyElement.EnumerateArray().Select(e => e.GetString());
```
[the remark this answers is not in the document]
RubenCerna2079: We already check if the value is null and it is expected to return an error

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Protocol;
+using static Azure.DataApiBuilder.Mcp.Model.McpEnums;
+
+namespace Azure.DataApiBuilder.Mcp.BuiltInTools
+{
+    public class ReadRecordsTool : IMcpTool
+    {
+        // private readonly IMetadataProviderFactory _metadataProviderFactory;
```
anushakolan: Please remove these lines since they are no longer used.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        public async Task<CallToolResult> ExecuteAsync(
+            JsonDocument? arguments,
+            IServiceProvider serviceProvider,
+            CancellationToken cancellationToken = default)
+        {
+            ILogger<ReadRecordsTool>? logger = serviceProvider.GetService<ILogger<ReadRecordsTool>>();
+
+            if (arguments == null)
```
anushakolan: nit: Maybe we can move this inside try as well.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                                ""type"": ""string"",
+                                ""description"": ""The entity name to read from. Required.""
+                            },
+                            ""select"": {
+                                ""type"": ""string"",
+                                ""description"": ""A CSV of field names to include in the response. If not provided, all fields are returned. Optional.""
+                            },
+                            ""filter"": {
+                                ""type"": ""string"",
+                                ""description"": ""A filter expression string to restrict results. Optional.""
```
JerryNixon: Let's update this to be more expressive:

> A case-insensitive OData-like expression that defines a query predicate. Supports logical grouping with parentheses and the operators eq, ne, gt, ge, lt, le, and, or, not. Examples: `year ge 1990`, `date lt 2025-01-01T00:00:00Z`, `(title eq 'Foundation') and (available ne false)`.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+            {
+                Name = "read_records",
+                Description = "Reads the records from the specified entity.",
+                InputSchema = JsonSerializer.Deserialize<JsonElement>(
+                    @"{
+                        ""type"": ""object"",
+                        ""properties"": {
+                            ""entity"": {
+                                ""type"": ""string"",
+                                ""description"": ""The entity name to read from. Required.""
```
JerryNixon: Let's update this to be more expressive:
> The name of the entity to read, as provided by the `describe_entities` tool. Required.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                            },
+                            ""select"": {
+                                ""type"": ""string"",
+                                ""description"": ""A CSV of field names to include in the response. If not provided, all fields are returned. Optional.""
```
JerryNixon: Let's update this to be more expressive: 
> A comma-separated list of field names to include in the response. If omitted, all fields are returned. Optional.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+        // private readonly IQueryEngineFactory _queryEngineFactory;
+
+        public ToolType ToolType { get; } = ToolType.BuiltIn;
+
+        public Tool GetToolMetadata()
+        {
+            return new Tool
+            {
+                Name = "read_records",
+                Description = "Reads the records from the specified entity.",
```
JerryNixon: Let's update this to be more expressive:
> Retrieves records from a given entity.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                                ""type"": ""string"",
+                                ""description"": ""A CSV of field names to include in the response. If not provided, all fields are returned. Optional.""
+                            },
+                            ""filter"": {
+                                ""type"": ""string"",
+                                ""description"": ""A filter expression string to restrict results. Optional.""
+                            },
+                            ""first"": {
+                                ""type"": ""integer"",
+                                ""description"": ""The maximum number of records to return in this page. Optional.""
```
JerryNixon: Let's update this to be more expressive: 
> The maximum number of records to return in the current page. Optional.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                            },
+                            ""orderby"": {
+                                ""type"": ""array"",
+                                ""items"": { ""type"": ""string"" },
+                                ""description"": ""A list of field names and directions for sorting (e.g., \""name asc\""). Optional.""
```
JerryNixon: Let's update this to be more expressive: 
> A list of field names and directions for sorting, for example "name asc" or "year desc". Optional.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                            },
+                            ""after"": {
+                                ""type"": ""string"",
+                                ""description"": ""A cursor token for retrieving the next page of results. Optional.""
```
JerryNixon: Let's update this to be more expressive:
> A cursor token for retrieving the next page of results. Returned as `nextLink` in the previous response. Optional.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+            }
+            catch (OperationCanceledException)
+            {
+                return BuildErrorResult("OperationCanceled", "The read operation was canceled.", logger: null);
+            }
+            catch (ArgumentException argEx)
+            {
+                return BuildErrorResult("InvalidArguments", argEx.Message, logger);
+            }
+            catch (Exception ex)
```
JerryNixon: Can you add an additional catch for `SqlException`? This is a question, not a request. I am not sure the database exception bubbles up here or not. If it does, then we want to capture it if we can. This would include connection issues and missing database objects, etc. Obviously, we would want `DbException` or something generic so it is not only SQL, unless SQL is all we *can* support, in that case, SQL-only would be acceptable. My main request here is, if you can capture MORE explicit types of exceptions, I would want that.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                innerLogger?.LogError(ex, "Unexpected error in ReadRecordsTool.");
+
+                return BuildErrorResult(
+                    "UnexpectedError",
+                    "An unexpected error occurred while reading the record.",
+                    logger: null);
+            }
+        }
+
+        private static bool TryResolveAuthorizedRole(
```
JerryNixon, line 282: It appears to me that this is nearly an exact repeat from the other tools. Somehow, we need to centralize this repetitive code in a refactoring effort after these tools have been merged. If you want to help start that, you could refactor yours here into a reusable class the other tools can refactor into later. Speak to Anusha, she may have already done this.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+            }
+
+            string[] roles = roleHeader
+                .Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                .Distinct(StringComparer.OrdinalIgnoreCase)
+                .ToArray();
+
+            if (roles.Length == 0)
+            {
+                error = "Client role header is missing or empty.";
```
JerryNixon: Provide the header name in the message so the error can help the caller resolve the error.
> AuthorizationResolver.CLIENT_ROLE_HEADER

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+            out string error)
+        {
+            effectiveRole = null;
+            error = string.Empty;
+
+            string roleHeader = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER].ToString();
+
+            if (string.IsNullOrWhiteSpace(roleHeader))
+            {
+                error = "Client role header is missing or empty.";
```
JerryNixon: Provide the header name in the message so the error can help the caller resolve the error.
> AuthorizationResolver.CLIENT_ROLE_HEADER

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                return BuildErrorResult("OperationCanceled", "The read operation was canceled.", logger: null);
+            }
+            catch (ArgumentException argEx)
+            {
+                return BuildErrorResult("InvalidArguments", argEx.Message, logger);
+            }
+            catch (Exception ex)
+            {
+                ILogger<ReadRecordsTool>? innerLogger = serviceProvider.GetService<ILogger<ReadRecordsTool>>();
+                innerLogger?.LogError(ex, "Unexpected error in ReadRecordsTool.");
```
JerryNixon: Why are you logging here when you are ALSO logging in `BuildErrorResult`?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                    entityName, role, EntityActionOperation.Read);
+
+                if (allowed)
+                {
+                    effectiveRole = role;
+                    return true;
+                }
+            }
+
+            error = "You do not have permission to read records for this entity.";
```
JerryNixon: You know the entity name so include it in the error message. Remember, error messages are intended to inform the caller with as much information as possible without leaking security so that the caller can correct the problem.
> entityName

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+            // Build normalized response
+            Dictionary<string, object?> normalized = new()
+            {
+                ["status"] = "success",
+                ["result"] = engineRootElement // only requested values
+            };
+
+            string output = JsonSerializer.Serialize(normalized, new JsonSerializerOptions { WriteIndented = true });
+
+            logger?.LogInformation("ReadRecordsTool success for entity {Entity}.", entityName);
```
JerryNixon, line 347: Logs are nice, but where is the OTEL wrapper? This is important.
> mcp_read_total_count, mcp_read_active_count, etc. 

We want to ensure the developer experience across all 3 endpoints is the same.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        private static CallToolResult BuildErrorResult(
+            string errorType,
+            string message,
+            ILogger? logger)
+        {
+            Dictionary<string, object?> errorObj = new()
+            {
+                ["status"] = "error",
```
JerryNixon, line 371: Curious. Why are we not setting `Success = true` and `Success = false`?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                    new TextContentBlock { Type = "text", Text = output }
+                ]
+            };
+        }
+
+        /// <summary>
+        /// Extracts a JSON string from a typical IActionResult.
+        /// Falls back to "{}" for unsupported/empty cases to avoid leaking internals.
+        /// </summary>
+        private static string ExtractResultJson(IActionResult? result)
```
JerryNixon, line 397: This is another example of a reusable method.

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -89,27 +91,43 @@ public static OkObjectResult FormatFindResult(
...
+                    new { after = after }
+                });
+                JsonElement afterElement = JsonSerializer.Deserialize<JsonElement>(jsonString);
+
+                rootEnumerated.Add(afterElement);
+            }
+            // Create a 'nextLink' object if the request comes from REST endpoint.
+            else
+            {
+                string basePaginationUri = SqlPaginationUtil.ConstructBaseUriForPagination(httpContext, runtimeConfig.Runtime?.BaseRoute);
```
JerryNixon, line 114: I cannot tell by looking, you might just dismiss this comment, but does this line obey this?

```json
{
  "runtime": {
    "rest": {
      "next-link-relative": true // default is false
    }
  }
}
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                if (!string.IsNullOrWhiteSpace(select))
+                {
+                    fieldsReturnedForFind = select.Split(",").ToList();
+                }
+                else
+                {
+                    fieldsReturnedForFind = authResolver.GetAllowedExposedColumns(context.EntityName, effectiveRole!, context.OperationType);
+                }
+
+                context.UpdateReturnFields(fieldsReturnedForFind);
```
anushakolan: nit: we can add a comment on what this call is doing.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                IQueryEngine queryEngine = queryEngineFactory.GetQueryEngine(sqlMetadataProvider.GetDatabaseType());
+                JsonDocument? queryResult = await queryEngine.ExecuteAsync(context);
+                IActionResult actionResult = queryResult is null ? SqlResponseHelpers.FormatFindResult(JsonDocument.Parse("[]").RootElement.Clone(), context, metadataProviderFactory.GetMetadataProvider(dataSourceName), runtimeConfigProvider.GetConfig(), httpContext, true)
+                                               : SqlResponseHelpers.FormatFindResult(queryResult.RootElement.Clone(), context, metadataProviderFactory.GetMetadataProvider(dataSourceName), runtimeConfigProvider.GetConfig(), httpContext, true);
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // Normalize response
+                string rawPayloadJson = ExtractResultJson(actionResult);
+                using JsonDocument result = JsonDocument.Parse(rawPayloadJson);
```
anushakolan: nit: Can we declare this using in the imports section, start of the document?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
anushakolan, line 282: Add the function summary.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                {
+                    effectiveRole = role;
+                    return true;
+                }
+            }
+
+            error = "You do not have permission to read records for this entity.";
+            return false;
+        }
+
```
anushakolan, line 326: Add the function summary.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+            return new CallToolResult
+            {
+                Content = new List<ContentBlock>
+                {
+                    new TextContentBlock { Type = "text", Text = output }
+                }
+            };
+        }
+
```
anushakolan, line 357: Add the function summary

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -89,27 +91,43 @@ public static OkObjectResult FormatFindResult(
...
-
             // When there are extra fields present, they are removed before returning the response.
             if (extraFieldsInResponse.Count > 0)
             {
                 rootEnumerated = RemoveExtraFieldsInResponseWithMultipleItems(rootEnumerated, extraFieldsInResponse);
             }
 
-            rootEnumerated.Add(nextLink);
-            return OkResponse(JsonSerializer.SerializeToElement(rootEnumerated));
+            // Create an 'after' object if the request comes from MCP endpoint.
```
anushakolan, line 100: what is an 'after' object? This is more of a question, I am trying to understand the context here.

src/Core/Parsers/FilterParser.cs
```diff
@@ -44,7 +44,6 @@ public FilterClause GetFilterClause(string filterQueryString, string resourcePat
         {
             if (_model is null)
             {
```
anushakolan, line 46: nit: Is this new line introduced on purpose, if not we can revert it.

anushakolan (changes requested):

anushakolan:
nit: Fix the description, `Manual testing via MCP Inspecto`

src/Core/Resolvers/SqlResponseHelpers.cs
souvikghosh04: its a Base64Encoded link to next set of results in pagination if the results are longer to acommodate at once or as configured in PaginationOptions in config.

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -200,20 +219,34 @@ public static OkObjectResult OkResponse(JsonElement jsonResult)
...
-                Dictionary<string, object> nextLink = JsonSerializer.Deserialize<Dictionary<string, object>>(nextLinkJsonString[1..^1])!;
+                // [{object1}, {object2},...{objectlimit}, [{nextLinkObject/afterObject}]]
+                // if the last element is of type array, we know it is 'nextLink'
+                // if the request is done through the REST endpoint and it is
+                // 'after' if the request is done through the MCP endpoint,
+                // we strip the "[" and "]" and then save the element
+                // into a dictionary with a key of "nextLinkAfter" and a value that
+                // represents the nextLink/after data we require.
+                string nextLinkAfterJsonString = JsonSerializer.Serialize(resultEnumerated[resultEnumerated.Count - 1]);
+                Dictionary<string, object> nextLinkAfter = JsonSerializer.Deserialize<Dictionary<string, object>>(nextLinkAfterJsonString[1..^1])!;
```
souvikghosh04, line 232: instead of removing the first and last characters, can we have a more deterministic way of removing them only after validating there exists `[` and `]` in exact places? may be using a Regex?
also, nit- may be add a comment about this

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
souvikghosh04: yes, we should move them into Utils and use from there. as part of delete-record tool PR https://github.com/Azure/data-api-builder/pull/2889 where I have refactored these by moving them inside Utils. may be we can do that same here once my PR merges to main.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -0,0 +1,379 @@
...
+                }
+
+                // Build and validate Find context
+                RequestValidator requestValidator = new(metadataProviderFactory, runtimeConfigProvider);
+                FindRequestContext context = new(entityName, dbObject, true);
+
+                requestValidator.ValidateEntity(entityName);
+
+                IEnumerable<string> fieldsReturnedForFind;
+                if (!string.IsNullOrWhiteSpace(select))
```
souvikghosh04, line 190: what happens when user tries to query restricted columns? should this be allowed to fail naturally or needs additional check and error handling/logging.

src/Core/Resolvers/SqlResponseHelpers.cs
RubenCerna2079: When you create a read request to only get a specific number of results, you will be using pagination. In order to get the results that follow you need to use add a property called `after` which is a string that allows DAB to know the position of the following result of the query. However, in the case of REST we construct the entire api link and call it `nextLink`, which is not something that we want to do for the MCP.
In short, the `after` object is the string that DAB returns that allows it to know the position to start returning the next set of results from the entity.

src/Core/Parsers/FilterParser.cs
RubenCerna2079: I am not following, I deleted the line that shouldn't be there. I did not introduce a new line.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
RubenCerna2079: We already talked about this between Anusha, Aaron, and me. We came to the consensus that we refactor these reusable classes after we merge the tools into main.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
RubenCerna2079: We only return the `after` string for MCP, so I am changing `nextLink` to `after`

src/Core/Resolvers/SqlResponseHelpers.cs
RubenCerna2079: As discussed offline, we only return the `after` string for the MCP so this property is irrelevant in this case.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
RubenCerna2079: I think this is something that needs to be done in a more general area, since we want all of the tools to be tracked inside of the OTEL wrapper

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
RubenCerna2079: Didn't know we could do that :)

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
RubenCerna2079: There was a missing check for the excluded fields, added the final checks that were necessary

src/Core/Resolvers/SqlResponseHelpers.cs
RubenCerna2079: I think this should be a separate task since this is how it was done from the start and it is out of scope for this PR, I will create the task for it.

RubenCerna2079: Task was created: https://github.com/Azure/data-api-builder/issues/2901

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ReadRecordsTool.cs
```diff
@@ -78,11 +78,6 @@ public async Task<CallToolResult> ExecuteAsync(
         {
             ILogger<ReadRecordsTool>? logger = serviceProvider.GetService<ILogger<ReadRecordsTool>>();
```
anushakolan, line 80: We need to add a check to see if the tool is enabled or not, if it is not enabled, we should not return any response but give an error message.

RubenCerna2079:
> I received 2 important comments from Aniruddh on the update record tool which will also apply for all other tools.
> 
> 1. We should use the `SqlResponseHelper` to create a response that MCP endpoint needs to avoid the duplicate serialization and deserialization. (I think this is addressed in your PR`
> 2. If the `ClientRoleHeaderAuthenticationMiddleware` isn't invoked, the roleheader here will be empty. Please test that out.

1. You are correct, my PR already uses `SqlResponseHelper` you can see it under the `// Execute` comment.
2. Tested to make sure that the role header isn't empty. Tests were successful.

anushakolan:
> > I received 2 important comments from Aniruddh on the update record tool which will also apply for all other tools.
> > 
> > 1. We should use the `SqlResponseHelper` to create a response that MCP endpoint needs to avoid the duplicate serialization and deserialization. (I think this is addressed in your PR`
> > 2. If the `ClientRoleHeaderAuthenticationMiddleware` isn't invoked, the roleheader here will be empty. Please test that out.
> 
> 1. You are correct, my PR already uses `SqlResponseHelper` you can see it under the `// Execute` comment.
> 2. Tested to make sure that the role header isn't empty. Tests were successful.

Nice work Ruben!! Please update the testing section in the PR description with point no. 2. Everything else LGTM.

Issue #2831, opened 2025-08-20
[Enh]: Add DML tool: `execute_entity`


JerryNixon:
### execute_entity

A tool to allow an agent to execute a stored procedure or function exposed as a DAB entity.

### Behavior

1. **Behaves most like the GraphQL flow**

2. **Entity permissions are enforced**

   * Never bypass permissions.
   * On violation, return a standard DAB response payload with an error message (e.g. `"You do not have permission to execute this entity"`).
   * Do **not** signal errors via HTTP status codes.

3. **Parameters**

   * Uses the parameter definitions exposed in `describe_entities`.
   * All Parameters are required. 
   * Output Parameters are not supported. (This is DAB standard)
   * Types must match GraphQL data types defined in metadata.
   * If provided values violate constraints, allow the database to fail and return the resulting error in a standard DAB response payload.

4. **Returns the execution result**

   * The response is a standard DAB payload with the returned dataset (if any).
   * If the procedure returns multiple result sets, return the first one. (This is DAB standard)

5. **Key and field handling**

   * Keys and fields are not part of this request. Only parameters defined in metadata are valid.

## How

* [ ] Add `execute_entity` MCP tool through graphql flow
* [ ] Add config property (`runtime.mcp.dml-tools.execute-entity.enabled=true`) 
* [ ] Update JSON Schema with `runtime.mcp.dml-tools.execute-entity.enabled`
* [ ] Obey configuration (`runtime.mcp.dml-tools.execute-entity.enabled=true`) 
* [ ] Add CLI `dab configure --runtime.mcp.dml-tools.execute-entity.enabled true`
* [ ] Update `dab validate` (warn when `execute-entity.enabled` and not `mcp.enabled`)

### Tool method

```csharp
/// <summary>
/// Executes a stored procedure or function returns the results (if any). 
/// </summary>
Task<DabResponse> ExecuteEntityAsync(
    [Description("The entity name of the procedure or function to execute. "
               + "Must match an entity returned by list_entities with execute=true. Required.")]
    string entity,

    [Description("A dictionary of parameter names and values to pass to the procedure. "
               + "Parameters must match those defined in entity metadata. Optional if no parameters.")]
    IDictionary<string, object?>? parameters = null);
```

**Parameters**

* **entity** *(string, required)*

  ```csharp
  if (string.IsNullOrWhiteSpace(entity))
      throw new ArgumentException("Entity is required", nameof(entity));
  if (!entityMetadata.Execute)
[...]

PR #2902, opened 2025-10-10
[MCP] Fixes and Refactoring execute_entity

souvikghosh04:
### Closes on
- `execute_entity` https://github.com/Azure/data-api-builder/issues/2831

## What is this change?
This PR implements built-int tool execute_entity as part of built-in tools to support execution of stored procedures or functions.

- Performs execute operation on a stored procedure or function entity
- Supports both querying and mutation
- `entity` (required) name and `parameters` (optional based on entity parameters signature) should be specified (see sample below)
- Operation is performed based on permissions as configured in dab-config
- Success or Failure message response is generated on execution of the delete operation

## How was this tested?

- [x] Manual functional test using MCP inpector/HTTP client
- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

Execute stored procedure to read queries-

```

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "execute_entity",
    "arguments": {
      "entity": "GetBook",
      "parameters": {
        "id": 1
      }
    }
  }
}
```

Execute stored procedure without parameters-

```

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "execute_entity",
    "arguments": {
      "entity": "GetBooks"
    }
  }
}
```

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
@@ -232,7 +232,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                 catch (DataApiBuilderException dabEx)
                 {
                     // Handle specific DAB exceptions
-                    logger?.LogError(dabEx, "Data API Builder error executing stored procedure {StoredProcedure}", entity);
+                    logger?.LogError(dabEx, "Data API builder error executing stored procedure {StoredProcedure}", entity);
 
                     string message = dabEx.Message;
 


src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+                    @"{
+                        ""type"": ""object"",
+                        ""properties"": {
+                            ""entity"": {
+                                ""type"": ""string"",
+                                ""description"": ""The entity name of the procedure or function to execute. Must match a stored-procedure entity as configured in dab-config. Required.""
+                            },
+                            ""parameters"": {
+                                ""type"": ""object"",
+                                ""description"": ""A dictionary of parameter names and values to pass to the procedure. Parameters must match those defined in dab-config. Optional if no parameters.""
```
anushakolan, line 58: The parameter will now be a list of objects and dictionary will be deprecated with this PR, https://github.com/Azure/data-api-builder/pull/2904.

Based on the order of merging changes, this description will need to be updated later.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+                }
+
+                // Entity is required
+                if (string.IsNullOrWhiteSpace(entity))
+                {
+                    return BuildDabResponse(false, null, "Entity is required", logger);
+                }
+
+                // 2) Resolve required services & configuration
+                RuntimeConfigProvider runtimeConfigProvider = serviceProvider.GetRequiredService<RuntimeConfigProvider>();
```
anushakolan: Similar to Ani's comment in update entity PR, 2 things need to be checked,

1. We should use the SqlResponseHelper to create a response that MCP endpoint needs to avoid the duplicate serialization and deserialization.
2. If the ClientRoleHeaderAuthenticationMiddleware isn't invoked, the roleheader here will be empty. Please test that out.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+
+                // 3) Validate entity exists and is a stored procedure
+                if (!config.Entities.TryGetValue(entity, out Entity? entityConfig))
+                {
+                    return BuildDabResponse(false, null, $"Entity '{entity}' not found in configuration.", logger);
+                }
+
+                if (entityConfig.Source.Type != EntitySourceType.StoredProcedure)
+                {
+                    return BuildDabResponse(false, null, $"Entity {entity} cannot be executed.", logger);
```
anushakolan: nit: The error can be more detailed on the source type is not a stored procedure and cannot be executed.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+        }
+
+        /// <summary>
+        /// Executes a stored procedure or function, returns the results (if any).
+        /// </summary>
+        public async Task<CallToolResult> ExecuteAsync(
+            JsonDocument? arguments,
+            IServiceProvider serviceProvider,
+            CancellationToken cancellationToken = default)
+        {
```
anushakolan, line 74: We should another check to see if the tool is exposed or not by reading the config.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+                }
+
+                // 5) Authorization - Never bypass permissions
+                IAuthorizationResolver authResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
+
+                if (httpContext is null || !authResolver.IsValidRoleContext(httpContext))
+                {
+                    return BuildDabResponse(false, null, "You do not have permission to execute this entity", logger);
```
anushakolan: nit: We can be more specific in the error message to avoid the confusion between error messages in line 142 and 153.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+                // First, add user-provided parameters to the context
+                if (requestPayloadRoot != null)
+                {
+                    foreach (JsonProperty property in requestPayloadRoot.Value.EnumerateObject())
+                    {
+                        context.FieldValuePairsInBody[property.Name] = GetParameterValue(property.Value);
+                    }
+                }
+
+                // Then, add default parameters from configuration (only if not already provided by user)
```
anushakolan, line 205: What are default parameters here? Normally user mentions on what fields to return in the sp, when that is not mentioned what are we returning here?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+                    // Allow the database to fail and return the resulting error
+                    return BuildDabResponse(false, null, dabEx.Message, logger);
+                }
+                catch (SqlException sqlEx)
+                {
+                    // Handle SQL Server specific errors
+                    logger?.LogError(sqlEx, "SQL Server error executing stored procedure {StoredProcedure}", entity);
+                    string errorMessage = sqlEx.Number switch
+                    {
+                        2812 => $"Stored procedure '{entityConfig.Source.Object}' not found in the database.",
```
anushakolan, line 269: Curious, is this data not already present in the exception message?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+                    };
+                    return BuildDabResponse(false, null, errorMessage, logger);
+                }
+                catch (DbException dbEx)
+                {
+                    // Handle generic database exceptions (works for PostgreSQL, MySQL, etc.)
+                    logger?.LogError(dbEx, "Database error executing stored procedure {StoredProcedure}", entity);
+                    return BuildDabResponse(false, null, $"Database error: {dbEx.Message}", logger);
+                }
+                catch (InvalidOperationException ioEx) when (ioEx.Message.Contains("connection", StringComparison.OrdinalIgnoreCase))
```
anushakolan, line 284: InvalidOperationError and "failed to connect to db" are not the same. This error message seems confusing.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,554 @@
...
+                return BuildDabResponse(false, null, "The execute operation was canceled.", logger);
+            }
+            catch (ArgumentException argEx)
+            {
+                return BuildDabResponse(false, null, argEx.Message, logger);
+            }
+            catch (Exception ex)
+            {
+                logger?.LogError(ex, "Unexpected error in ExecuteEntityTool.");
+                return BuildDabResponse(false, null, "An unexpected error occurred during the execute operation.", logger);
```
anushakolan: To be consistent with the other tools, so that later refactoring will be easy it would be good to rename these functions to, `BuildSuccessResult` and `BuildErrorResult`.

anushakolan:
The tests are failing because of formatting issues, just for reference, dotnet format "file_name"

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: the response is currently a standard JSON serialized value of the entity metadata. there is no double deserialization, however, we need to test this in different setup for stdio and http clients. regarding the auth resolver, yes this has been tested. for example, if you use `anonymous` role in `GetBooksAuth` entity it will fail- 

image

image


it should succeed when we set correct client headers- 
image

this can be tested with the expected role and claims and base64 encoding the value in powershell- $principal = '{"identityProvider":"github","userId":"user123","userDetails":"user@example.com","userRoles":["teststoredprocauth"],"claims":[]}'
$bytes = [System.Text.Encoding]::UTF8.GetBytes($principal)
$encoded = [Convert]::ToBase64String($bytes)
Write-Host $encoded

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: added the check similar to delete-record

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: the default parameters are configured in the dab-config. I have been thinking about to remove this since it unnecessary and confusing so will have it removed in this PR.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: in another PR, there was an ask from Jerry to handle database exception explicitly if possible. this catch block specifically handles those.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: this has been refactored in `delete_record` PR for now. there is also a separate task created for refactoring and this one of the first thing to touch upon. for now marking this to close.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: yes this will change once your PR merges

anushakolan (approved):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,546 @@
...
+                }
+
+                // 2) Resolve required services & configuration
+                RuntimeConfigProvider runtimeConfigProvider = serviceProvider.GetRequiredService<RuntimeConfigProvider>();
+                RuntimeConfig config = runtimeConfigProvider.GetConfig();
+
+                // 3) Check if ExecuteEntity tool is enabled
+                if (config.McpDmlTools?.ExecuteEntity != true)
+                {
+                    return BuildDabResponse(false,
```
RubenCerna2079: nit:
```suggestion
                    return BuildDabResponse(
                        false,
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,546 @@
...
+
+                if (!sqlMetadataProvider.EntityToDatabaseObject.TryGetValue(entity, out DatabaseObject? dbObject) || dbObject is null)
+                {
+                    return BuildDabResponse(false, null, $"Failed to resolve database object for entity '{entity}'.", logger);
+                }
+
+                // 6) Authorization - Never bypass permissions
+                IAuthorizationResolver authResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
```
RubenCerna2079, line 148: Does it matter what type of http request it is?
For example in the `read_records` we need to change the request to `GET` for the authorization and proper execution, this needs to be done because all the MCP requests are done with the `POST` request.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,546 @@
...
+
+                if (httpContext is null || !authResolver.IsValidRoleContext(httpContext))
+                {
+                    return BuildDabResponse(false, null, "You do not have permission to execute this entity", logger);
+                }
```
RubenCerna2079: nit:
```suggestion
                    return BuildDabResponse(false, null, "You do not have permission to execute this entity.", logger);
                }
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,546 @@
...
+                    entityName, role, operation);
+
+                if (allowed)
+                {
+                    effectiveRole = role;
+                    return true;
+                }
+            }
+
+            error = $"Failed in TryResolveAuthorizedRole. You do not have permission to execute entity '{entityName}'.";
```
RubenCerna2079: Why do we need to specify that it failed in `TryResolveAuthorizedRole`?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,546 @@
...
+                {
+                    ["message"] = "You do not have permission to execute this entity"
+                };
+            }
+            else
+            {
+                // Empty or unknown result
+                response["value"] = JsonSerializer.SerializeToElement(Array.Empty<object>());
+                logger?.LogInformation("ExecuteEntityTool completed with empty result.");
+            }
```
RubenCerna2079: Is it possible for the result to be an error? I would expect it to throw an error that can be catched rather than a result that contains the error inside.

RubenCerna2079 (approved):
Just left a few comments. I'll be waiting for the next PR with the refactoring :)

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: the default is GET

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -201,6 +202,18 @@ public async Task<CallToolResult> ExecuteAsync(
                     }
                 }
 
+                // Then, add default parameters from configuration (only if not already provided by user)
+                if (entityConfig.Source.Parameters != null)
```
RubenCerna2079: I don't see any check to see if the user already provided to parameters.

anushakolan:
LGTM!!

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,453 @@
...
+                try
+                {
+                    // Cancellation check before executing
+                    cancellationToken.ThrowIfCancellationRequested();
+                    queryResult = await queryEngine.ExecuteAsync(context, dataSourceName).ConfigureAwait(false);
+                }
+                catch (DataApiBuilderException dabEx)
+                {
+                    // Handle specific DAB exceptions
+                    logger?.LogError(dabEx, "Data API Builder error executing stored procedure {StoredProcedure}", entity);
```
Aniruddh25:
```suggestion
                    logger?.LogError(dabEx, "Data API builder error executing stored procedure {StoredProcedure}", entity);
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -0,0 +1,453 @@
...
+                    dbo: dbObject,
+                    requestPayloadRoot: requestPayloadRoot,
+                    operationType: EntityActionOperation.Execute);
+
+                // First, add user-provided parameters to the context
+                if (requestPayloadRoot != null)
+                {
+                    foreach (JsonProperty property in requestPayloadRoot.Value.EnumerateObject())
+                    {
+                        context.FieldValuePairsInBody[property.Name] = GetParameterValue(property.Value);
```
Aniruddh25, line 201: We are deserializing the requestPayloadRoot here into a Dictionary of string, value names. But couldnt we simply have assigned the parsed `parameters` to `FieldValuePaisInBody`? 

I think we are doing unnecessary serialization of `parameters` passed to the MCP endpoint and deserializing them again to conform to the REST request requirements.

Aniruddh25: While this may be fine for now, to improve performance - we should eliminate unnecessary serialization and deserialization.

anushakolan (approved):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
Aniruddh25: resolving to follow up in a subsequent refactoring PR - @anushakolan for reference.

Issue #2828, opened 2025-08-20
[Enh]: Add DML tool: `create_record`


JerryNixon:
NOTE: Before you start, lets discuss calling this tool `create_record`.

### create\_entity

A tool to allow an agent to create a single record.

### Behavior

1. **Behaves most like the GraphQL flow**

2. **Entity permissions are enforced**

   * Never bypass permissions.
   * On violation, return a standard DAB response payload with an error message (e.g. `"You do not have permission to create this entity"`).
   * Do **not** signal errors via HTTP status codes.

3. **Supports partial field lists**

   * Requests may specify only a subset of fields.
   * Whether a field is required is not determined by metadata. A request may omit required fields in error.
   * If required fields are missing, allow the database to fail and return the resulting error in a standard DAB response payload.

4. **Returns the created entity**

   * The response includes only the combination of keys and fields supplied in the request.
   * If an autogen key was not supplied, include it in the response.
   * Example: a table with 100 columns, request with 10 fields → response contains those 10 fields (plus autogen keys, if any).

5. **Key handling**

   * Keys use the entity’s mapped aliases.
   * Non-autogen keys are required in the request.
   * Autogen keys are optional in the request.
   * Autogen keys can be included in the request as `null`.

6. **Field handling**

   * Fields use the entity’s mapped aliases.

## How

* [ ] Add `create_entity` MCP tool through graphql flow
* [ ] Add config property (`runtime.mcp.dml-tools.create-entity.enabled=true`)
* [ ] Update JSON Schema with `runtime.mcp.dml-tools.create-entity.enabled`
* [ ] Obey configuration (`runtime.mcp.dml-tools.create-entity.enabled=true`)
* [ ] Add CLI `dab configure --runtime.mcp.dml-tools.create-entity.enabled true`
* [ ] Update `dab validate` (warn when `create-entity.enabled` and not `mcp.enabled`)

### Tool method

```csharp
/// <summary>
/// Creates a new record in the specified entity. 
/// Returns only the requested keys and fields (plus any autogen keys, if applicable).
/// </summary>
Task<DabResponse> CreateEntityAsync(
    [Description("The entity name of the record to create. Must match an entity returned from list_entities with create=true. Required.")]
    string entity,

    [Description("A dictionary of key/value pairs identifying the record’s keys. Keys must match those defined in entity metadata. Optional for autogen keys.")]
    IDictionary<string, object?>? keys,

[...]

PR #2906, opened 2025-10-13
[MCP] Add the Create_Record Built in tool

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2828

By adding the `Create_Record` tool to the MCP endpoint.

## What is this change?
Add a built in tool, `Create_Record` which uses the `rest` infrastructure to form and validate the query needed to complete a create action. Behavior is at parity with the rest equivalent create operation.

## How was this tested?

Manually tested Insomnia using post.
image

image

image

## Sample Request(s)

```
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "tools/call",
  "params": {
    "name": "create_record",
    "arguments": {
      "entity": "Broker",
      "data": {
        "ID Number": 3,
				"First Name": "Michael",
				"Last Name": "Jordan"
      }
    }
  }
}
 
```

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
@@ -173,8 +173,8 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                         new Dictionary<string, object?>
                         {
                             ["entity"] = entityName,
-                            ["operation"] = "create",
-                            ["result"] = createdResult.Value
+                            ["result"] = createdResult.Value,
+                            ["message"] = $"Successfully created record in entity '{entityName}'"
                         },
                         logger,
                         $"Successfully created record in entity '{entityName}'");
@@ -195,12 +195,11 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                             new Dictionary<string, object?>
                             {
                                 ["entity"] = entityName,
-                                ["operation"] = "create",
                                 ["result"] = objectResult.Value,
-                                ["note"] = "Unable to perform read-back of inserted records"
+                                ["message"] = $"Successfully created record in entity '{entityName}'. Unable to perform read-back of inserted records."
                             },
                             logger,
-                            $"Successfully created record in entity '{entityName}' (read-back unavailable)");
+                            $"Successfully created record in entity '{entityName}'. Unable to perform read-back of inserted records.");
                     }
                 }
                 else
@@ -218,9 +217,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                             new Dictionary<string, object?>
                             {
                                 ["entity"] = entityName,
-                                ["operation"] = "create",
-                                ["result_type"] = result.GetType().Name,
-                                ["note"] = "Create operation completed with unexpected result type"
+                                ["message"] = $"Create operation completed with unexpected result type: {result.GetType().Name}"
                             },
                             logger,
                             $"Create operation completed for entity '{entityName}' with unexpected result type: {result.GetType().Name}");


src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
             };
         }
 
-        public Task<CallToolResult> ExecuteAsync(
+        public async Task<CallToolResult> ExecuteAsync(
             JsonDocument? arguments,
             IServiceProvider serviceProvider,
             CancellationToken cancellationToken = default)
         {
```
anushakolan, line 58: We need to add a check to see if the tool is exposed or not, and return an error message accordingly.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+                    try
+                    {
+                        requestValidator.ValidateInsertRequestContext(insertRequestContext);
+                    }
+                    catch (DataApiBuilderException ex)
+                    {
+                        return new CallToolResult
+                        {
+                            Content = [new TextContentBlock { Type = "text", Text = $"Error: Request validation failed: {ex.Message}" }]
+                        };
```
anushakolan: Add `IsError = true` to return it as error instead of success.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+                else if (result is ObjectResult objectResult)
                 {
-                    Content = [new TextContentBlock { Type = "text", Text = result }]
-                });
+                    return new CallToolResult
+                    {
+                        Content = [new TextContentBlock 
+                        { 
+                            Type = "text", 
+                            Text = $"Record creation completed with status {objectResult.StatusCode}. Result: {JsonSerializer.Serialize(objectResult.Value)}"
```
anushakolan: Just to confirm, are we returning the same response of hoe REST returns for create_record?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+                ["error"] = new Dictionary<string, object?>
+                {
+                    ["type"] = errorType,
+                    ["message"] = message
+                }
+            };
+
+            string output = JsonSerializer.Serialize(errorObj);
+
+            logger?.LogWarning("UpdateRecordTool error {ErrorType}: {Message}", errorType, message);
```
anushakolan: This should be create_record tool

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+
+            return new CallToolResult
+            {
+                Content =
+                [
+                    new TextContentBlock { Type = "text", Text = output }
+                ]
```
anushakolan: Same, add `IsError = true`

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+                    insertPayloadRoot,
+                    EntityActionOperation.Insert);
 
-                // TODO: Implement actual create logic using DAB's internal services
-                // For now, return a placeholder response
-                string result = $"Would create record in entity '{entityName}' with data: {dataElement.GetRawText()}";
+                RequestValidator requestValidator = serviceProvider.GetRequiredService<RequestValidator>();
+                
+                // Only validate tables
+                if (dbObject.SourceType is EntitySourceType.Table)
```
anushakolan, line 145: When the type is not table, may be we could error out with that reason, to give more context for the MCP Client.

anushakolan:
The current PR description only covers the success case. Since the tool should also be validated against edge cases and other standard scenarios, I recommend expanding the testing section to include details on the various test cases executed to confirm the tool’s expected behavior.

anushakolan:
The unit tests are failing, please run dotnet format and that should fix most of the syntax issues.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
             CancellationToken cancellationToken = default)
         {
+            ILogger<CreateRecordTool>? logger = serviceProvider.GetService<ILogger<CreateRecordTool>>();
             if (arguments == null)
             {
-                return Task.FromResult(new CallToolResult
-                {
-                    Content = [new TextContentBlock { Type = "text", Text = "Error: No arguments provided" }]
-                });
+                return BuildErrorResult("Invalid Arguments", "No arguments provided", logger);
```
RubenCerna2079: We already merged the PR that has the global functions. Will you change it in this PR or will you refactor it in a follow up?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
                 {
-                    return Task.FromResult(new CallToolResult
-                    {
-                        Content = [new TextContentBlock { Type = "text", Text = "Error: Missing required arguments 'entity' or 'data'" }]
-                    });
+                    return BuildErrorResult("Invalid Arguments", "Missing required arguments 'entity' or 'data'", logger);
                 }
 
                 string entityName = entityElement.GetString() ?? string.Empty;
+                if (string.IsNullOrEmpty(entityName))
```
RubenCerna2079: I think it woud be better to check if it is whitespace as well
```suggestion
                if (string.IsNullOrWhitespace(entityName))
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+                {
+                    return BuildErrorResult("Invalid Arguments", "Entity name cannot be empty", logger);
```
RubenCerna2079: In the other PRs we have the `ErrorType` without any spaces. Please change it for this one an all the others
```suggestion
                    return BuildErrorResult("InvalidArguments", "Entity name cannot be empty", logger);
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+                {
+                    return BuildErrorResult("Invalid Configuration", "Runtime configuration not available", logger);
+                }
+
+                string dataSourceName;
+                try
+                {
+                    dataSourceName = runtimeConfig.GetDataSourceNameFromEntityName(entityName);
+                }
+                catch (DataApiBuilderException)
```
RubenCerna2079: Is a `DataApiBuilderException` the only possible exception? I think it would be better to also at a `catch (Exception)`

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +53,218 @@ public Tool GetToolMetadata()
...
+                
+
+                // Create an HTTP context for authorization
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext httpContext = httpContextAccessor.HttpContext ?? new DefaultHttpContext();
+                IAuthorizationResolver authorizationResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+
+                if (httpContext is null || !authorizationResolver.IsValidRoleContext(httpContext))
+                {
+                    return BuildErrorResult("PermissionDenied", "Permission denied: unable to resolve a valid role context for update operation.", logger);
```
RubenCerna2079: nit: 
```suggestion
                    return BuildErrorResult("PermissionDenied", "Permission denied: Unable to resolve a valid role context for update operation.", logger);
```

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +52,235 @@ public Tool GetToolMetadata()
...
+                // Only validate tables
+                if (dbObject.SourceType is EntitySourceType.Table)
                 {
-                    Content = [new TextContentBlock { Type = "text", Text = result }]
-                });
+                    try
+                    {
+                        requestValidator.ValidateInsertRequestContext(insertRequestContext);
+                    }
+                    catch (DataApiBuilderException ex)
```
RubenCerna2079: Same as comment above. Is it only possible for the function to return a `DataApiBuilderException`?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +52,235 @@ public Tool GetToolMetadata()
...
+                    {
+                        return new CallToolResult
```
RubenCerna2079: I think we should use the `BuildErrorResult` to output this error.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +52,235 @@ public Tool GetToolMetadata()
...
+                        }]
+                    };
+                }
+                else if (result is ObjectResult objectResult)
+                {
+                    return new CallToolResult
+                    {
+                        Content = [new TextContentBlock 
+                        { 
+                            Type = "text",
```
RubenCerna2079: Is this expected to be an error or can it also be something else?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +52,235 @@ public Tool GetToolMetadata()
...
+                    return new CallToolResult
+                    {
+                        Content = [new TextContentBlock { Type = "text", Text = $"Successfully created record in entity '{entityName}'" }]
+                    };
+                }
             }
             catch (Exception ex)
             {
-                return Task.FromResult(new CallToolResult
+                return new CallToolResult
```
RubenCerna2079: This should also be called by `BuildErrorResult`

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +52,235 @@ public Tool GetToolMetadata()
...
+                    };
+                }
+                else if (result is ObjectResult objectResult)
+                {
+                    return new CallToolResult
+                    {
+                        Content = [new TextContentBlock 
+                        { 
+                            Type = "text", 
+                            Text = $"Record creation completed with status {objectResult.StatusCode}. Result: {JsonSerializer.Serialize(objectResult.Value)}"
```
Aniruddh25: If this is a success case, why is the message different here. Why is it not: "Successfully created record in entity"

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
Aniruddh25: Along the same lines, if its an error, IsError = true should be set.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +52,235 @@ public Tool GetToolMetadata()
...
+                    return true;
+                }
+            }
+
+            error = "You do not have permission to create records for this entity.";
+            return false;
+        }
+
+        private static CallToolResult BuildErrorResult(
+        string errorType,
```
Aniruddh25: nit: should be tabs before each argument declaration - similar to other function definitions.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +52,235 @@ public Tool GetToolMetadata()
...
+                return new CallToolResult
                 {
                     Content = [new TextContentBlock { Type = "text", Text = $"Error: {ex.Message}" }]
-                });
+                };
             }
         }
+
+        private static bool TryResolveAuthorizedRole(
+        HttpContext httpContext,
```
Aniruddh25: nit: should have tabs before the argument names

Aniruddh25 (approved):
Need to reuse the refactored helper methods (perhaps in a followup - please create a task for that), but can be merged as long as comments are addressed.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: Ill use the one we already have in main.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: This may or may not be an error, so I now have it looking at the status code to determine if IsError should be true or not, and a more generic message.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: It might not be a success case.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +51,200 @@ public Tool GetToolMetadata()
...
+                        Content = [new TextContentBlock 
+                        { 
+                            Type = "text", 
+                            Text = $"Successfully created record in entity '{entityName}'. Result: {JsonSerializer.Serialize(createdResult.Value)}"
+                        }]
+                    };
+                }
+                else if (result is ObjectResult objectResult)
+                {
+                    // Check if this is an error status code (400+ range)
```
Aniruddh25: Are all the ObjectResult cases where the record has been created? Is there additional information that can be appended to the successful creation message?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -37,51 +51,200 @@ public Tool GetToolMetadata()
...
+                    return new CallToolResult
+                    {
+                        Content = [new TextContentBlock
+                        {
+                            Type = "text",
+                            Text = $"Successfully created record in entity '{entityName}'. Result: {JsonSerializer.Serialize(objectResult.Value)}"
```
Aniruddh25: If its not an ObjectResult, is it still a successful case?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
Aniruddh25: Don't we need to check for different error status codes before implying it was always a success?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
anushakolan: > Just to confirm, are we returning the same response of hoe REST returns for create_record?

Waiting on this, rest looks good to me.

anushakolan:
The unit tests are failing because of formatting issues, fix the formatting and re-run the tests. Approving the changes to unblock the PR, but waiting on
1. confirming if the tool response matches the REST create record response.
2. All tests are successfully passing.

anushakolan (approved):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: yes we take the same results as the REST return and simply wrap it in the MCP format for the calltoolresult.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: added the chceks for these edge cases.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: added a null case here that is an error, otherwise it could be a complete operation with unexpected results, in which case we indicate that in our response,

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -172,31 +171,49 @@ public async Task<CallToolResult> ExecuteAsync(
                 {
                     return new CallToolResult
                     {
-                        Content = [new TextContentBlock 
-                        { 
-                            Type = "text", 
+                        Content = [new TextContentBlock
```
Aniruddh25: Should use the McpResponseBuilder.BuildSuccessResult.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -172,31 +171,49 @@ public async Task<CallToolResult> ExecuteAsync(
...
                 }
                 else if (result is ObjectResult objectResult)
                 {
-                    // Check if this is an error status code (400+ range)
+                    bool isError = objectResult.StatusCode.HasValue && objectResult.StatusCode.Value >= 400 && objectResult.StatusCode.Value != 403;
+                    string text = isError
+                        ? $"Failed to create record in entity '{entityName}'. Error: {JsonSerializer.Serialize(objectResult.Value)}"
+                        : $"Successfully created record in entity '{entityName}'. Result: {JsonSerializer.Serialize(objectResult.Value)}. Unable to perform " +
+                        $"read-back of inserted records";
                     return new CallToolResult
```
Aniruddh25: Should use BuildErrorResult in case of `isError` otherwise use BuildSuccessResult

Aniruddh25 (approved):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -169,30 +169,39 @@ public async Task<CallToolResult> ExecuteAsync(
...
+                    }
+                    else
+                    {
+                        return Utils.McpResponseBuilder.BuildSuccessResult(
+                            new Dictionary<string, object?>
+                            {
+                                ["entity"] = entityName,
+                                ["operation"] = "create",
+                                ["result"] = objectResult.Value,
+                                ["note"] = "Unable to perform read-back of inserted records"
```
Aniruddh25: is `note` a standard expected response property?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -205,14 +214,16 @@ public async Task<CallToolResult> ExecuteAsync(
...
+                        return Utils.McpResponseBuilder.BuildSuccessResult(
+                            new Dictionary<string, object?>
                             {
-                                Type = "text",
-                                Text = $"Create operation completed for entity '{entityName}' with unexpected result type: {result.GetType().Name}"
-                            }]
-                        };
+                                ["entity"] = entityName,
+                                ["operation"] = "create",
+                                ["result_type"] = result.GetType().Name,
```
Aniruddh25: Is `result_type` a standard expected response?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: Actually looks like we are using "message" here in delete records, so I changed now to align with that.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
aaronburtle: I removed this and appended it to the "message"

Aniruddh25 (approved):

PR #2916, opened 2025-10-21
[MCP] Added fields section for entities

anushakolan:
## Why make this change?
This change introduces richer metadata for entity fields and keys in the Data API Builder configuration. The goal is to enable semantic metadata for documentation, client generation, and MCP tools, and to begin deprecating the legacy mappings and source.key-fields properties.

## What is this change?
- Introduces a new `fields` property for entities in the config schema.
- Each field supports:
  - `name`: database column name (required)
  - `alias`: exposed name for the field (optional)
  - `description`: field description (optional)
  - `primary-key`: whether the field is a primary key (optional)
- Updates the JSON schema to enforce that `fields` cannot be used with legacy `mappings` or `source.key-fields`.
- Updates CLI commands (`dab add`, `dab update`) to support specifying field alias, description, and primary-key.
- Updates `dab validate` to warn if fields are missing and MCP is enabled.
- Updates OpenAPI, GraphQL, and MCP `describe_entities` responses to include field descriptions.
- Adds auto-migration logic to convert legacy `mappings` and `source.key-fields` to the new `fields` format when relevant CLI flags are used.
- Maintains backward compatibility: legacy properties are still supported, but validation enforces that `fields` and legacy props are not mixed.

## How was this tested?
Manually tested with following queries:

1. Update Entity with Field Mappings and Key Fields
`dotnet C:\DAB\data-api-builder\src\out\cli\net8.0\Microsoft.DataApiBuilder.dll update BookAuthor --map "BookID:book_id,AuthorID:author_id" --source.key-fields "BookID,AuthorID" --config "C:\DAB\data-api-builder\src\Service\dab-config.json"`

------

Validate the Configuration

`dotnet C:\DAB\data-api-builder\src\out\cli\net8.0\Microsoft.DataApiBuilder.dll validate -c "C:\DAB\data-api-builder\src\Service\dab-config.json"`

-----

Update Entity Field Metadata (Primary Key)

`dotnet C:\DAB\data-api-builder\src\out\cli\net8.0\Microsoft.DataApiBuilder.dll update Todo --fields.name owner_id --fields.primary-key True --config C:\DAB\data-api-builder\src\Service\dab-config.json`

----

Update Entity Key Fields Only

`dotnet C:\DAB\data-api-builder\src\out\cli\net8.0\Microsoft.DataApiBuilder.dll update BookAuthor --source.key-fields "BookID"`

----

Sample new format

`"fields": [
        {
          "name": "id",
          "alias": "id",
          "description": "The unique identifier for a todo item",
          "primary-key": true
        },
        {
          "name": "title",
          "alias": "title",
          "description": "The title of the todo item",
          "primary-key": false
        },
        {
          "name": "completed",
          "alias": "completed",
          "description": "Indicates whether the todo item is completed",
          "primary-key": false
        },
        {
          "name": "owner_id",
          "alias": "owner",
          "description": "Hello",
          "primary-key": true
        },
        {
          "name": "position",
          "alias": "position",
          "description": "The position of the todo item in the list",
          "primary-key": false
        }
      ],`

-----

GraphQL Introspection Example

Query

`{
  __type(name: "Todo") {
    name
    description
    fields {
      name
      description
    }
  }
}`

Result
`{
  "data": {
    "__type": {
      "name": "Todo",
      "description": "Represents a todo item in the system",
      "fields": [
        { "name": "id", "description": "The unique identifier for a todo item" },
        { "name": "title", "description": "The title of the todo item" },
        { "name": "completed", "description": "Indicates whether the todo item is completed" },
        { "name": "owner", "description": "Hello" },
        { "name": "position", "description": "The position of the todo item in the list" }
      ]
    }
  }
}`

------

OpenAPI Schema Example

`"Todo": {
  "type": "object",
  "properties": {
    "id": { "type": "string", "description": "The unique identifier for a todo item", "format": "" },
    "title": { "type": "string", "description": "The title of the todo item", "format": "" },
    "completed": { "type": "boolean", "description": "Indicates whether the todo item is completed", "format": "" },
    "owner_id": { "type": "string", "description": "Hello", "format": "" },
    "position": { "type": "number", "description": "The position of the todo item in the list", "format": "" }
  },
  "description": "Represents a todo item in the system"
}`

### DIFF
--- a/src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMsSql.verified.txt
+++ b/src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMsSql.verified.txt
@@ -1996,11 +1996,14 @@
       books_view_all: {
         Source: {
           Object: books_view_all,
-          Type: View,
-          KeyFields: [
-            id
-          ]
+          Type: View
         },
+        Fields: [
+          {
+            Name: id,
+            PrimaryKey: true
+          }
+        ],
         GraphQL: {
           Singular: books_view_all,
           Plural: books_view_alls,
@@ -2075,12 +2078,18 @@
       stocks_view_selected: {
         Source: {
           Object: stocks_view_selected,
-          Type: View,
-          KeyFields: [
-            categoryid,
-            pieceid
-          ]
+          Type: View
         },
+        Fields: [
+          {
+            Name: categoryid,
+            PrimaryKey: true
+          },
+          {
+            Name: pieceid,
+            PrimaryKey: true
+          }
+        ],
         GraphQL: {
           Singular: stocks_view_selected,
           Plural: stocks_view_selecteds,
@@ -2122,12 +2131,18 @@
       books_publishers_view_composite: {
         Source: {
           Object: books_publishers_view_composite,
-          Type: View,
-          KeyFields: [
-            id,
-            pub_id
-          ]
+          Type: View
         },
+        Fields: [
+        {
+          Name: id,
+          PrimaryKey: true
+        },
+        {
+          Name: pub_id,
+          PrimaryKey: true
+        }
+      ],
         GraphQL: {
           Singular: books_publishers_view_composite,
           Plural: books_publishers_view_composites,


anushakolan:
One pending follow up task, is to check if the given field is a valid column in the entity. Will add this check as part of a follow up PR.

src/Cli.Tests/Snapshots/UpdateEntityTests.TestUpdateSourceStringToDatabaseSourceObject_574e1995f787740f.verified.txt
```diff
@@ -27,12 +27,18 @@
...
         Source: {
           Object: s001.book,
-          Type: View,
-          KeyFields: [
-            col1,
-            col2
-          ]
+          Type: View
         },
+        Fields: [
```
Aniruddh25, line 32: We should ensure the new `fields` section is well-documented. Please create a doc task for @JerryNixon

src/Config/ObjectModel/FieldMetadata.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// </summary>
+        public string? Alias { get; set; }
+
+        /// <summary>
+        /// The description for the field.
+        /// </summary>
+        public string? Description { get; set; }
+
+        /// <summary>
+        /// Whether this field is a key (must be unique).
```
Aniruddh25, line 24: Do we validate for views and for tables which dont have a primary key defined in the SQL metadata - at least 1 field from the collection has PrimraryKey = true?

src/Config/ObjectModel/FieldMetadata.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    public class FieldMetadata
+    {
+        /// <summary>
+        /// The name of the field (must match a database column).
+        /// </summary>
+        public string Name { get; set; } = string.Empty;
+
+        /// <summary>
+        /// The alias for the field (must be unique per entity).
```
Aniruddh25, line 14: Do we validate alias is unique across all fields?

src/Cli/ConfigGenerator.cs
```diff
@@ -453,6 +453,7 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
             // Create new entity.
             Entity entity = new(
                 Source: source,
+                Fields: null,
```
Aniruddh25, line 456: are we not handling adding fields to the entity using the `dab add` command? Why are there no fieldOptions here?

src/Cli/ConfigGenerator.cs
```diff
@@ -2473,5 +2609,69 @@ private static bool TryUpdateConfiguredAzureKeyVaultOptions(
...
+                        PrimaryKey = keys.Count > i && keys[i],
+                    });
+                }
+            }
+
+            return fields;
+        }
+
+        /// <summary>
+        /// Validates that the provided fields are valid against the database columns and constraints.
```
Aniruddh25, line 2687: this doesnt seem to validate against the database columns. What if the field names provided dont match the column names in the database?

If we dont have the metadata info from the database, we should change the summary to say validating uniqueness only.

src/Cli/ConfigGenerator.cs
```diff
@@ -1646,24 +1647,137 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
+                }
+
+                // Remove legacy props
+                updatedMappings = null;
+                updatedSource = updatedSource with { KeyFields = null };
+            }
+            else
+            {
+                fields = [];
+                _logger.LogWarning("Using legacy 'mappings' and 'key-fields' properties. Consider using 'fields' for new entities.");
```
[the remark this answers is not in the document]
Aniruddh25: This is a good suggestion from Copilot

src/Cli/ConfigGenerator.cs
```diff
@@ -1646,24 +1647,137 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
+                foreach (FieldMetadata field in existingFields)
+                {
+                    if (updatedFieldsDict.TryGetValue(field.Name, out FieldMetadata? updatedField))
+                    {
+                        mergedFields.Add(new FieldMetadata
+                        {
+                            Name = updatedField.Name,
+                            Alias = updatedField.Alias ?? field.Alias,
+                            Description = updatedField.Description ?? field.Description,
+                            PrimaryKey = updatedField.PrimaryKey || field.PrimaryKey
```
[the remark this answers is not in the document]
Aniruddh25: Again, copilot is correct here.

src/Cli/ConfigGenerator.cs
```diff
@@ -1646,24 +1647,137 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
+                {
+                    if (options.Map is null || !TryParseMappingDictionary(options.Map, out updatedMappings))
+                    {
+                        _logger.LogError("Failed to parse mappings from --map option.");
+                        return false;
+                    }
+
+                    foreach (KeyValuePair<string, string> mapping in updatedMappings)
+                    {
+                        fields.Add(new FieldMetadata
```
Aniruddh25: What if there were `existingFields`? We need to merge them with the fields created using the mappings.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -424,16 +424,21 @@ private static FieldDefinitionNode GenerateFieldForColumn(Entity configEntity, s
             }
 
             string exposedColumnName = columnName;
-            if (configEntity.Mappings is not null && configEntity.Mappings.TryGetValue(key: columnName, out string? columnAlias))
+            FieldMetadata? fieldMetadata = null;
+            if (configEntity.Fields is not null)
```
Aniruddh25, line 435: We should continue to check for configEntity.Mappings is not null. Both cant be true but either could be.

src/Service.Tests/TestHelper.cs
```diff
@@ -80,6 +80,7 @@ public static RuntimeConfig AddMissingEntitiesToConfig(RuntimeConfig config, str
         {
             Entity entity = new(
                 Source: new(entityName, EntitySourceType.Table, null, keyfields),
+                Fields: new List<FieldMetadata> { new() { Name = keyfields?[0], PrimaryKey = true } },
```
Aniruddh25: is it possible that Name gets initialized to a null value given that keyFields could be null? If it gets set as null, we will end up with a Fields list with 1 element where Name is null, PrimaryKey is true.

Also, this is assuming when keyFields is not null, there will just be 1 key

Aniruddh25 (changes requested):
Waiting for addressing some comments where from Copilot

src/Config/ObjectModel/FieldMetadata.cs
anushakolan: Yes, we do

src/Cli/ConfigGenerator.cs
anushakolan: We do not have add supporting mappings either. I confirmed with Jerry that we only do this for update and not add.

src/Config/ObjectModel/FieldMetadata.cs
anushakolan: We don't validate it, but later in the code the test fails in LogPrimaryKeys if there is not even 1 pk set.

src/Cli/ConfigGenerator.cs
anushakolan: This is the pending task, I had to skip that as it is breaking more things and want to check in this code first.

src/Cli/ConfigGenerator.cs
anushakolan: Updated

src/Service.Tests/TestHelper.cs
anushakolan: Absolutely right, good catch. This caused the test failures related to alias not being set correctly. Fixed it.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
anushakolan: Done

schemas/dab.draft.schema.json
```diff
@@ -785,6 +785,21 @@
...
+            "fields": {
+              "type": "array",
+              "description": "Defines the fields (columns) exposed for this entity, with metadata.",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "name": { "type": "string", "description": "Database column name." },
+                  "alias": { "type": "string", "description": "Exposed name for the field." },
+                  "description": { "type": "string", "description": "Field description." },
+                  "primary-key": { "type": "boolean", "description": "Is this field a primary key?" }
```
RubenCerna2079, line 797: Is this a question for someone? I think we should change this description

src/Cli.Tests/Snapshots/UpdateEntityTests.TestUpdateSourceStringToDatabaseSourceObject_a13a9ca73b21f261.verified.txt
```diff
@@ -27,12 +27,18 @@
       MyEntity: {
         Source: {
           Object: s001.book,
-          Type: Table,
-          KeyFields: [
```
RubenCerna2079, line 31: Is `KeyFields` a different name for mappings just to specify that it is a primary key? Please also mention this parameter in the PR description

src/Cli.Tests/AddEntityTests.cs
```diff
@@ -45,7 +45,11 @@ public Task AddNewEntityWhenEntitiesEmpty()
                 cacheTtl: null,
                 config: TEST_RUNTIME_CONFIG_FILE,
                 restMethodsForStoredProcedure: null,
-                graphQLOperationForStoredProcedure: null
+                graphQLOperationForStoredProcedure: null,
+                fieldsNameCollection: [],
+                fieldsAliasCollection: [],
+                fieldsDescriptionCollection: [],
+                fieldsPrimaryKeyCollection: []
```
RubenCerna2079, line 52: Is there a reason to make these parameters empty and not null?

RubenCerna2079: The same question applies to the other instances where this change was made.

src/Cli/Commands/EntityOptions.cs
```diff
@@ -25,8 +25,13 @@ public EntityOptions(
...
             string? cacheTtl,
-            string? config,
-            string? description)
+            string? description,
+            IEnumerable<string>? fieldsNameCollection,
+            IEnumerable<string>? fieldsAliasCollection,
+            IEnumerable<string>? fieldsDescriptionCollection,
+            IEnumerable<bool>? fieldsPrimaryKeyCollection,
+            string? config
+            )
```
RubenCerna2079, line 34: nit: 
```suggestion
            string? config)
```

src/Cli/ConfigGenerator.cs
```diff
@@ -1646,24 +1647,137 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
+                }
+
+                // Remove legacy props
+                updatedSource = updatedSource with { KeyFields = null };
+                updatedMappings = null;
+            }
+            else
+            {
+                fields = [];
+                _logger.LogWarning("Using legacy 'mappings' and 'key-fields' properties. Consider using 'fields' for new entities.");
```
RubenCerna2079: I think this warning should be in the `else if (hasMappings || hasKeyFields)` section

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1094,22 +1094,72 @@ await PopulateResultSetDefinitionsForStoredProcedureAsync(
...
+                    // If still empty, fallback to DB schema PKs
+                    if (pkFields == null || pkFields.Count == 0)
+                    {
+                        DataTable dataTable = await GetTableWithSchemaFromDataSetAsync(
+                            entityName,
+                            GetSchemaName(entityName),
+                            GetDatabaseObjectName(entityName));
+
+                        pkFields = dataTable.PrimaryKey.Select(pk => pk.ColumnName).ToList();
+                    }
```
RubenCerna2079, line 1200: I see that this code is the same as the one above. Couldn't we make a helper function out of it in order to not have repeated code?

PR #2925, opened 2025-10-27
Partial fix for MCP test

RubenCerna2079:
## Why make this change?
- #2874
There is a test that is failing and it needs to be fixed in order to comply with the creation of the new MCP endpoint. The test already existed and covered the scenarios for REST and GraphQL endpoints. It will only be partially fixed, only fixing the issues related to the non-hosted scenario, this is due to the fact that fixing the hosted-scenario will need more changes than expected.

## What is this change?
This change partially fixes the test that was failing, the non-hosted scenario was failing because the MCP server was not able to start in time before the test tried to access the MCP endpoint. In order to fix it we added a delay so the server is available before the test tries to access the endpoint.
On the other hand, the hosted scenario is failing because of the way that DAB initializes its MCP service, which means that the base needs to be changed. Which is a bigger task than what is expected of this PR.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

### DIFF
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -695,7 +695,7 @@
     },
     "entities": {
       "type": "object",
-      "description": "Entities that will be exposed via REST and/or GraphQL",
+      "description": "Entities that will be exposed via REST, GraphQL and/or MCP",
       "patternProperties": {
         "^.*$": {
           "type": "object",
@@ -961,6 +961,31 @@
                   "default": 5
                 }
               }
+            },
+            "mcp": {
+              "oneOf": [
+                {
+                  "type": "boolean",
+                  "description": "Boolean shorthand: true enables dml-tools only (custom-tool remains false), false disables all MCP functionality."
+                },
+                {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "dml-tools": {
+                      "type": "boolean",
+                      "description": "Enable MCP DML (Data Manipulation Language) tools for this entity. Allows CRUD operations via MCP.",
+                      "default": true
+                    },
+                    "custom-tool": {
+                      "type": "boolean",
+                      "description": "Enable MCP custom tool for this entity. Only valid for stored procedures.",
+                      "default": false
+                    }
+                  }
+                }
+              ],
+              "description": "Model Context Protocol (MCP) configuration for this entity. Controls whether the entity is exposed via MCP tools."
             }
           },
           "if": {
@@ -1144,6 +1169,33 @@
                     { "properties": { "source": { "properties": { "key-fields": { } }, "required": ["key-fields"] } } }
                   ]
                 }
+              }
+            },
+            {
+              "if": {
+                "properties": {
+                  "mcp": {
+                    "properties": {
+                      "custom-tool": {
+                        "const": true
+                      }
+                    }
+                  }
+                },
+                "required": ["mcp"]
+              },
+              "then": {
+                "properties": {
+                  "source": {
+                    "properties": {
+                      "type": {
+                        "const": "stored-procedure"
+                      }
+                    },
+                    "required": ["type"]
+                  }
+                },
+                "errorMessage": "custom-tool can only be enabled for entities with source type 'stored-procedure'."
               }
             }
           ]
--- a/src/Cli.Tests/AddEntityTests.cs
+++ b/src/Cli.Tests/AddEntityTests.cs
@@ -633,5 +633,259 @@ namespace Cli.Tests
 
             return Verify(updatedRuntimeConfig, settings);
         }
+
+        #region MCP Entity Configuration Tests
+
+        /// <summary>
+        /// Test adding table entity with MCP dml-tools enabled or disabled
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("true", "books", "Book", DisplayName = "AddTableEntityWithMcpDmlToolsEnabled")]
+        [DataRow("false", "authors", "Author", DisplayName = "AddTableEntityWithMcpDmlToolsDisabled")]
+        public Task AddTableEntityWithMcpDmlTools(string mcpDmlTools, string source, string entity)
+        {
+            AddOptions options = new(
+                source: source,
+                permissions: new string[] { "anonymous", "*" },
+                entity: entity,
+                description: null,
+                sourceType: "table",
+                sourceParameters: null,
+                sourceKeyFields: null,
+                restRoute: null,
+                graphQLType: null,
+                fieldsToInclude: Array.Empty<string>(),
+                fieldsToExclude: Array.Empty<string>(),
+                policyRequest: null,
+                policyDatabase: null,
+                cacheEnabled: null,
+                cacheTtl: null,
+                config: TEST_RUNTIME_CONFIG_FILE,
+                restMethodsForStoredProcedure: null,
+                graphQLOperationForStoredProcedure: null,
+                parametersNameCollection: null,
+                parametersDescriptionCollection: null,
+                parametersRequiredCollection: null,
+                parametersDefaultCollection: null,
+                fieldsNameCollection: [],
+                fieldsAliasCollection: [],
+                fieldsDescriptionCollection: [],
+                fieldsPrimaryKeyCollection: [],
+                mcpDmlTools: mcpDmlTools,
+                mcpCustomTool: null
+            );
+
+            VerifySettings settings = new();
+            settings.UseParameters(mcpDmlTools, source);
+            return ExecuteVerifyTest(options, settings: settings);
+        }
+
+        /// <summary>
+        /// Test adding stored procedure with MCP custom-tool enabled (should serialize as object)
+        /// </summary>
+        [TestMethod]
+        public Task AddStoredProcedureWithMcpCustomToolEnabled()
+        {
+            AddOptions options = new(
+                source: "dbo.GetBookById",
+                permissions: new string[] { "anonymous", "execute" },
+                entity: "GetBookById",
+                description: null,
+                sourceType: "stored-procedure",
+                sourceParameters: null,
+                sourceKeyFields: null,
+                restRoute: null,
+                graphQLType: null,
+                fieldsToInclude: Array.Empty<string>(),
+                fieldsToExclude: Array.Empty<string>(),
+                policyRequest: null,
+                policyDatabase: null,
+                cacheEnabled: null,
+                cacheTtl: null,
+                config: TEST_RUNTIME_CONFIG_FILE,
+                restMethodsForStoredProcedure: null,
+                graphQLOperationForStoredProcedure: null,
+                parametersNameCollection: null,
+                parametersDescriptionCollection: null,
+                parametersRequiredCollection: null,
+                parametersDefaultCollection: null,
+                fieldsNameCollection: [],
+                fieldsAliasCollection: [],
+                fieldsDescriptionCollection: [],
+                fieldsPrimaryKeyCollection: [],
+                mcpDmlTools: null,
+                mcpCustomTool: "true"
+            );
+            return ExecuteVerifyTest(options);
+        }
+
+        /// <summary>
+        /// Test adding stored procedure with both MCP properties set to different values (should serialize as object with both)
+        /// </summary>
+        [TestMethod]
+        public Task AddStoredProcedureWithBothMcpProperties()
+        {
+            AddOptions options = new(
+                source: "dbo.UpdateBook",
+                permissions: new string[] { "anonymous", "execute" },
+                entity: "UpdateBook",
+                description: null,
+                sourceType: "stored-procedure",
+                sourceParameters: null,
+                sourceKeyFields: null,
+                restRoute: null,
+                graphQLType: null,
+                fieldsToInclude: Array.Empty<string>(),
+                fieldsToExclude: Array.Empty<string>(),
+                policyRequest: null,
+                policyDatabase: null,
+                cacheEnabled: null,
+                cacheTtl: null,
+                config: TEST_RUNTIME_CONFIG_FILE,
+                restMethodsForStoredProcedure: null,
+                graphQLOperationForStoredProcedure: null,
+                parametersNameCollection: null,
+                parametersDescriptionCollection: null,
+                parametersRequiredCollection: null,
+                parametersDefaultCollection: null,
+                fieldsNameCollection: [],
+                fieldsAliasCollection: [],
+                fieldsDescriptionCollection: [],
+                fieldsPrimaryKeyCollection: [],
+                mcpDmlTools: "false",
+                mcpCustomTool: "true"
+            );
+            return ExecuteVerifyTest(options);
+        }
+
+        /// <summary>
+        /// Test adding stored procedure with both MCP properties enabled (common use case)
+        /// </summary>
+        [TestMethod]
+        public Task AddStoredProcedureWithBothMcpPropertiesEnabled()
+        {
+            AddOptions options = new(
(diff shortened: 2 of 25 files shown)


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2570,17 +2581,35 @@ public async Task TestGlobalFlagToEnableRestAndGraphQLForHostedAndNonHostedEnvir
...
+                    jsonrpc = "2.0",
+                    id = 1,
+                    method = "tools/list"
+                };
+                HttpRequestMessage mcpRequest = new(HttpMethod.Post, "/mcp")
+                {
+                    Content = JsonContent.Create(mcpPayload)
+                };
+                mcpRequest.Headers.Add("Accept", "*/*");
+                HttpResponseMessage mcpResponse = await client.SendAsync(mcpRequest);
```
Aniruddh25: why cant we use the same `GetMcpResponsePostConfigHydration` method here to account for the retry logic?

Aniruddh25 (changes requested):
Looking for retry logic

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2570,17 +2581,35 @@ public async Task TestGlobalFlagToEnableRestAndGraphQLForHostedAndNonHostedEnvir
...
+                Assert.AreEqual(expectedStatusCodeForGraphQL, graphQLResponse.StatusCode, "The GraphQL response is different from the expected result.");
 
+                // REST request
                 HttpRequestMessage restRequest = new(HttpMethod.Get, "/api/Book");
                 HttpResponseMessage restResponse = await client.SendAsync(restRequest);
-                Assert.AreEqual(expectedStatusCodeForREST, restResponse.StatusCode);
+                Assert.AreEqual(expectedStatusCodeForREST, restResponse.StatusCode, "The REST response is different from the expected result.");
+
+                // MCP request
+                await Task.Delay(2000);
```
souvikghosh04: should we add a try-wait condition (within a limit- 5 seconds?) until he `/mcp` endpoint is available instead of hard-coded 2 seconds of wait?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5330,6 +5358,50 @@ private static async Task<HttpStatusCode> GetGraphQLResponsePostConfigHydration(
...
+                mcpRequest.Headers.Add("Accept", "*/*");
+
+                HttpResponseMessage mcpResponse = await httpClient.SendAsync(mcpRequest);
+                responseCode = mcpResponse.StatusCode;
+
+                if (responseCode == HttpStatusCode.ServiceUnavailable || responseCode == HttpStatusCode.NotFound)
+                {
+                    retryCount--;
+                    Thread.Sleep(TimeSpan.FromSeconds(RETRY_WAIT_SECONDS));
+                    continue;
```
souvikghosh04: The function is asynchronous but using `Thread.Sleep()` you can block the current thread. Instead use `Task.Await()`. Additionally, please check if there is a deterministic way to wait for `/mcp` endpoint within a given limit, instead of a hard-coded value which might vary in different envrionments (ADO pipelines) or scenarios.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2570,17 +2581,35 @@ public async Task TestGlobalFlagToEnableRestAndGraphQLForHostedAndNonHostedEnvir
...
+
+                // MCP request
+                await Task.Delay(2000);
+                object mcpPayload = new
+                {
+                    jsonrpc = "2.0",
+                    id = 1,
+                    method = "tools/list"
+                };
+                HttpRequestMessage mcpRequest = new(HttpMethod.Post, "/mcp")
```
souvikghosh04: can we get the MCP path from `RuntimeConfig` or `McpRuntimeOptions` itself since that is where it will be configured?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2578,9 +2590,25 @@ public async Task TestGlobalFlagToEnableRestAndGraphQLForHostedAndNonHostedEnvir
...
+                {
+                    jsonrpc = "2.0",
+                    id = 1,
+                    method = "tools/call"
+                };
+                HttpRequestMessage mcpRequest = new(HttpMethod.Post, "/mcp")
+                {
+                    Content = JsonContent.Create(mcpPayload)
+                };
+                HttpResponseMessage mcpResponse = await client.SendAsync(mcpRequest);
```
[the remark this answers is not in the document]
RubenCerna2079: Something similar will be implemented

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2578,9 +2590,25 @@ public async Task TestGlobalFlagToEnableRestAndGraphQLForHostedAndNonHostedEnvir
...
+                Assert.AreEqual(expectedStatusCodeForMcp, mcpResponse.StatusCode);
```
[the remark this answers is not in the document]
RubenCerna2079: They will be changed and will not be duplicated anymore

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Will change it for Rest and GraphQL as well

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Changed `Thread.Sleep()` to `Task.Delay()`

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: We will use the `GetMcpResponsePostConfigHydration` method which is also used later in order to address the retry logic

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Used the same method for retry logic, with some changes

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5243,14 +5259,14 @@ private static JsonContent GetPostStartupConfigParams(string environment, Runtim
         /// <param name="httpClient">Client used for request execution.</param>
         /// <param name="config">Post-startup configuration</param>
```
Aniruddh25: nit: need to define the new param in summary here.

src/Service/dab-config.json
```diff
@@ -17,6 +17,9 @@
         "path": "/graphql",
         "allow-introspection": true
       },
+      "mcp": {
```
Aniruddh25: Why is this change needed?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5320,7 +5336,50 @@ private static async Task<HttpStatusCode> GetGraphQLResponsePostConfigHydration(
                 if (responseCode == HttpStatusCode.ServiceUnavailable)
                 {
                     retryCount--;
-                    Thread.Sleep(TimeSpan.FromSeconds(RETRY_WAIT_SECONDS));
+                    await Task.Delay(TimeSpan.FromSeconds(RETRY_WAIT_SECONDS));
```
Aniruddh25: Consider adding an exponential increase in wait seconds for e.g. 1, 2, 4, 8

RubenCerna2079:
@copilot Do not make any more changes to this PR even if other users make comments

src/Service/dab-config.json
RubenCerna2079: They are not needed

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5256,41 +5274,44 @@ private static JsonContent GetPostStartupConfigParams(string environment, Runtim
...
-                    await httpClient.GetAsync($"api/{POST_STARTUP_CONFIG_ENTITY}");
+                    await httpClient.GetAsync($"{rest.Path}/{POST_STARTUP_CONFIG_ENTITY}");
                 responseCode = postConfigHydrationResult.StatusCode;
 
                 if (postConfigHydrationResult.StatusCode == HttpStatusCode.ServiceUnavailable)
                 {
-                    retryCount--;
-                    Thread.Sleep(TimeSpan.FromSeconds(RETRY_WAIT_SECONDS));
+                    retryCount++;
+                    await Task.Delay(TimeSpan.FromSeconds(Math.Pow(RETRY_WAIT_SECONDS, retryCount)));
```
souvikghosh04, line 5318: The incremental progression (2, 4, ..., 32, totalling to 64 seconds?) seems excess for a test and yet it can potentially fail beyond that? We should justify this and add additional comments and documentation regarding this delay.

souvikghosh04: Also, please consider moving this to a reusable test utility as we can reuse for other MCP tests and rely on a single source of mcp endpoint availability check.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5334,8 +5355,51 @@ private static async Task<HttpStatusCode> GetGraphQLResponsePostConfigHydration(
...
+                HttpRequestMessage mcpRequest = new(HttpMethod.Post, mcp.Path)
+                {
+                    Content = JsonContent.Create(payload)
+                };
+                mcpRequest.Headers.Add("Accept", "*/*");
+
+                HttpResponseMessage mcpResponse = await httpClient.SendAsync(mcpRequest);
+                responseCode = mcpResponse.StatusCode;
+
+                if (responseCode == HttpStatusCode.ServiceUnavailable || responseCode == HttpStatusCode.NotFound)
```
souvikghosh04, line 5411: when do we get `NotFound` and how different is it from `ServiceUnavailable`?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2600,18 +2617,19 @@ public async Task TestGlobalFlagToEnableRestAndGraphQLForHostedAndNonHostedEnvir
...
-                HttpStatusCode graphqlResponseCode = await GetGraphQLResponsePostConfigHydration(client);
-
-                Assert.AreEqual(expected: expectedStatusCodeForGraphQL, actual: graphqlResponseCode);
+                HttpStatusCode graphqlResponseCode = await GetGraphQLResponsePostConfigHydration(client, configuration.Runtime.GraphQL);
+                Assert.AreEqual(expected: expectedStatusCodeForGraphQL, actual: graphqlResponseCode, "The GraphQL hydration post-response is different from the expected result.");
 
+                // TODO: Currently DAB is unable to start MCP with the hydration post-response.
+                // This needs to be fixed before uncommenting the MCP check
+                // HttpStatusCode mcpResponseCode = await GetMcpResponsePostConfigHydration(client, configuration.Runtime.Mcp);
+                // Assert.AreEqual(expected: expectedStatusCodeForMcp, actual: mcpResponseCode, "The MCP hydration post-response is different from the expected result.");
```
souvikghosh04: could you please attach the issue/task link here so we don't lose the track until this gets addressed? having the link/issue number here will allow us to track even if the item is closed/moved out of sprint.

souvikghosh04 (commented):
Added some comments mainly about the exponential delay and moving it to a reusable utility. Suggested to attach the issue number for TODO task.

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: It would actually be total of 32 seconds, since we go 1, 2, 4, 8, 16. And if I remember correctly, we do this type of delay in other tests, so I don't see the reason to justify this and add additional comments.

RubenCerna2079: Regarding your second comment, do you mean we should move only the retry count logic to a separate function or do you mean the entire `GetMcpResponse` function?

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: When MCP is unavailable because it is booting up it can try and go through the REST endpoint, which would cause an unwanted error. This was changed so that if the endpoint it is trying to access is the MCP endpoint then it would return a `NotFound` error.

src/Service.Tests/Configuration/ConfigurationTests.cs
souvikghosh04: we should justify with some comments otherwise it seems like an assumption. it could be hard to tell why 32 seconds and why not more or less. this could help if anything goes wrong (like flaky environments and pipelines) in future or to refactor later. will leave it up to you.

about the second comment, I meant the `GetMcpResponse()` which can be reused for other MCP tests requiring to wait until MCP endpoint is available.

src/Service/dab-config.json
```diff
@@ -17,9 +17,6 @@
         "path": "/graphql",
         "allow-introspection": true
       },
-      "mcp": {
```
Aniruddh25: why this change?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2600,18 +2617,19 @@ public async Task TestGlobalFlagToEnableRestAndGraphQLForHostedAndNonHostedEnvir
...
 
-                HttpStatusCode graphqlResponseCode = await GetGraphQLResponsePostConfigHydration(client);
-
-                Assert.AreEqual(expected: expectedStatusCodeForGraphQL, actual: graphqlResponseCode);
+                HttpStatusCode graphqlResponseCode = await GetGraphQLResponsePostConfigHydration(client, configuration.Runtime.GraphQL);
+                Assert.AreEqual(expected: expectedStatusCodeForGraphQL, actual: graphqlResponseCode, "The GraphQL hydration post-response is different from the expected result.");
 
+                // TODO: Issue #3012 - Currently DAB is unable to start MCP with the hydration post-response.
+                // This needs to be fixed before uncommenting the MCP check
+                // HttpStatusCode mcpResponseCode = await GetMcpResponsePostConfigHydration(client, configuration.Runtime.Mcp);
```
Aniruddh25: If this is commented out, what is this PR really fixing? Is there any other MCP test that is getting fixed at all here? 

If the PR is NOT really fixing the MCP test - its description and title should change to only mention that it is adding exponential delay for better stability - otherwise the purpose of the PR is unclear. 

If you want to keep the same title, I expect the code here to be uncommented and fixed in the same PR instead of a separate issue.

Aniruddh25 (changes requested):
Purpose of this PR is unclear

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Just changed the PR description to reflect the changes that are expected of this PR. This PR does actually fix part of the test with the added delay, specifically the non-hosted scenario. The hosted scenario still needs more changes in order to be fixed.

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Added comment to give more context on the exponential retry logic.

For the second comment, I made the `GetMcpResponse` function public so that it can be used in other MCP tests.

PR #2927, opened 2025-10-28
Bug fix for setting dml tools based on bool value as set in config

souvikghosh04:
## Why make this change?

Bug fix for setting dml tools correctly.

Closes on - #2942 

## What is this change?

This fix addresses the inconsistency in config where `dml-tools` can be either set as bool or a dictionary object containing individual tool names and their bool values. Without this fix, we need to individually set each `dml-tools` to true currently.

With this change, we will have the following scenarios-

- set `dml-tools` to `true` will enable all dml tools (default is true)
Sample- ```"dml-tools": true```

- set `dml-tools` to `false` will disable all dml tools
Sample-```"dml-tools": false```

- individual tool names can be specified as dictionary object for `dml-tools` to turn them on (default is true if a tool is not specified).

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests
- [x] Manual testing using various combinations and scenarios

**Scenario 1:** Enable all tools using a single boolean value

`"dml-tools": true`

Note: default is true so even if `dml-tools` unspecified it will default to true.

**Scenario 2:** Disable `execute-entities` only and keep other tools enabled as default.

```
      "dml-tools": {
        "execute-entities":  false
      }
```

**Scenario 3:** Use full list of tools and enable or disable them

```
      "dml-tools": {
        "execute-entity": true,
        "delete-record": false,
        "update-record": false,
        "read-records": false,
        "describe-entities": true
      }
```

### DIFF
--- a/src/Config/ObjectModel/McpRuntimeOptions.cs
+++ b/src/Config/ObjectModel/McpRuntimeOptions.cs
@@ -51,7 +51,7 @@ public record McpRuntimeOptions
         // if DmlTools is null, set All tools enabled by default
         if (DmlTools is null)
         {
-            this.DmlTools = DmlToolsConfig.FromBoolean(true);
+            this.DmlTools = DmlToolsConfig.FromBoolean(DmlToolsConfig.DEFAULT_ENABLED);
         }
         else
         {


src/Config/ObjectModel/DmlToolsConfig.cs
```diff
@@ -116,12 +116,12 @@ public static DmlToolsConfig FromBoolean(bool enabled)
         return new DmlToolsConfig
         {
             AllToolsEnabled = enabled,
-            DescribeEntities = null,
-            CreateRecord = null,
-            ReadRecords = null,
-            UpdateRecord = null,
-            DeleteRecord = null,
-            ExecuteEntity = null
+            DescribeEntities = enabled,
```
Aniruddh25: Why is the setting `AllToolsEnabled` not sufficient to achieve `"dml-tools" :true` ? 
If its not useful, why do we keep this property: `AllToolsEnabled`?

Aniruddh25 (approved):
Approving since this is not harmful, but also waiting on clarification on need of `AllToolsEnabled`

Aniruddh25:
>> Manual testing using various combinations and scenarios

Can you please list out what combinations were tested here to understand if we missed a scenario?

src/Config/ObjectModel/DmlToolsConfig.cs
souvikghosh04: this is actually needed to denote whether how dml-tools was set- specific tools or all at once. based on this we also set `UserProvidedAllToolsEnabled` in `DmlToolsConfig.cs` and then use this in `DmlToolsConfigConverter.cs` to write the config properties accordingly.

src/Config/ObjectModel/DmlToolsConfig.cs
```diff
@@ -116,12 +116,12 @@ public static DmlToolsConfig FromBoolean(bool enabled)
...
+            CreateRecord = enabled,
+            ReadRecords = enabled,
+            UpdateRecord = enabled,
+            DeleteRecord = enabled,
+            ExecuteEntity = enabled
         };
```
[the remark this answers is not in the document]
souvikghosh04: Implemented the suggesstion.

anushakolan:
Since, we identified this as a bug, can you please provide more details on the testing. Waiting for the description section to be updated with more testing info.

souvikghosh04:
> Approving. But the description is not clear - it says default is false if not specified. I thought default was `true` for all the dml-tools even when we use the dictionary format of object for the `dml-tools` property.

this we are handling it by in `McpRuntimeOptions.cs` constructor-

```
        // if DmlTools is null, set All tools enabled by default
        if (DmlTools is null)
        {
            this.DmlTools = DmlToolsConfig.FromBoolean(true);
        }
        else
        {
            this.DmlTools = DmlTools;
        }
```

Aniruddh25:
> > Approving. But the description is not clear - it says default is false if not specified. I thought default was `true` for all the dml-tools even when we use the dictionary format of object for the `dml-tools` property.
> 
> this we are handling it by in `McpRuntimeOptions.cs` constructor-
> 
> ```
>         // if DmlTools is null, set All tools enabled by default
>         if (DmlTools is null)
>         {
>             this.DmlTools = DmlToolsConfig.FromBoolean(true);
>         }
>         else
>         {
>             this.DmlTools = DmlTools;
>         }
> ```

so the description needs to be fixed, It shouldnt say `default is false if not specified.`

souvikghosh04:
> > > Approving. But the description is not clear - it says default is false if not specified. I thought default was `true` for all the dml-tools even when we use the dictionary format of object for the `dml-tools` property.
> > 
> > 
> > this we are handling it by in `McpRuntimeOptions.cs` constructor-
> > ```
> >         // if DmlTools is null, set All tools enabled by default
> >         if (DmlTools is null)
> >         {
> >             this.DmlTools = DmlToolsConfig.FromBoolean(true);
> >         }
> >         else
> >         {
> >             this.DmlTools = DmlTools;
> >         }
> > ```
> 
> so the description needs to be fixed, It shouldnt say `default is false if not specified.`

the default here refers to the unspecified tools which will be set to `false` by default. the default value of `dml-tools` is `true` if `unspecified` but default value of individual tools will be `false` if `unspecified`. I have the updated the description and example.

Aniruddh25:
> > > > Approving. But the description is not clear - it says default is false if not specified. I thought default was `true` for all the dml-tools even when we use the dictionary format of object for the `dml-tools` property.
> > > 
> > > 
> > > this we are handling it by in `McpRuntimeOptions.cs` constructor-
> > > ```
> > >         // if DmlTools is null, set All tools enabled by default
> > >         if (DmlTools is null)
> > >         {
> > >             this.DmlTools = DmlToolsConfig.FromBoolean(true);
> > >         }
> > >         else
> > >         {
> > >             this.DmlTools = DmlTools;
> > >         }
> > > ```
> > 
> > 
> > so the description needs to be fixed, It shouldnt say `default is false if not specified.`
> 
> the default here refers to the unspecified tools which will be set to `false` by default. the default value of `dml-tools` is `true` if `unspecified` but default value of individual tools will be `false` if `unspecified`. I have the updated the description and example.

As discussed offline, this is incorrect behavior. Default value of unspecified tools in the dictionary object should be `true`. The PR needs to be fixed to reflect that.

src/Config/ObjectModel/McpRuntimeOptions.cs
```diff
@@ -48,7 +48,15 @@ public McpRuntimeOptions(
             this.Path = DEFAULT_PATH;
         }
 
-        this.DmlTools = DmlTools;
+        // if DmlTools is null, set All tools enabled by default
+        if (DmlTools is null)
+        {
+            this.DmlTools = DmlToolsConfig.FromBoolean(true);
```
Aniruddh25:
```suggestion
            this.DmlTools = DmlToolsConfig.FromBoolean(DmlToolsConfig.DEFAULT_ENABLED);
```

Aniruddh25 (approved):
LGTM, please fix the description.

src/Cli.Tests/ConfigGeneratorTests.cs
```diff
@@ -165,8 +165,16 @@ public void TestSpecialCharactersInConnectionString()
                  },
                  ""mcp"": {
                  ""enabled"": true,
-                 ""path"": ""/mcp""
-                 },
+                 ""path"": ""/mcp"",
+                 ""dml-tools"":{
```
anushakolan, line 169: When default is true, why are we explicitly setting everything to true here?

anushakolan:
LGTM, but the MS SQL Integration tests are still failing.

src/Cli.Tests/ConfigGeneratorTests.cs
souvikghosh04: some tests rely on a validating the runtime config in an expected format. this test is not checking anything specific to mcp or dml-tools but has existing validations which has a specific format of the config required to assert.

Issue #2644, opened 2025-04-07
Change default auth provider to `AppService` from `StaticWebApps`


JerryNixon:
## Ensure `EntraID` is identical to `StaticWebApp`

The goal is to ensure `AppService` is identical to `StaticWebApp` because we want to stop using `StaticWebApp` and start using `AppService`.

https://github.com/Azure/data-api-builder/blob/ebbe98983bb02a0cd1f099dede20e691db3cd9a7/src/Config/ObjectModel/EasyAuthType.cs#L6-L10

⚠️ None of this should be done if they are not identical.

## Modify the Engine default references to `StaticWebApps`

These references are scattered throughout the codebase. We're shifting to `AppService` as the default, so wherever `StaticWebApps` appears in code or comments, it should be updated. We're not removing support for `StaticWebApps`, just removing it as the default.

Comment example:  
https://github.com/Azure/data-api-builder/blob/ebbe98983bb02a0cd1f099dede20e691db3cd9a7/src/Config/ObjectModel/AuthenticationOptions.cs#L9

Code example:  
https://github.com/Azure/data-api-builder/blob/ebbe98983bb02a0cd1f099dede20e691db3cd9a7/src/Config/ObjectModel/AuthenticationOptions.cs#L14

## Modify the CLI default references to `StaticWebApps`

This primarily applies to `dab init`. The goal is to ensure the CLI no longer defaults to `StaticWebApps` as the auth provider but to `AppService`. Users can still configure `StaticWebApps`, but they must do so manually.

## Modify the JSON Schema

Update the schema to change the default value to `AppService`.

https://github.com/Azure/data-api-builder/blob/ebbe98983bb02a0cd1f099dede20e691db3cd9a7/schemas/dab.draft.schema.json#L275-L283

Also update the schema to restrict the provider property to only supported values (including `StaticWebApps`).

## Modify `DAB Validate`

Update `ValidateAuthenticationOptions` to log a warning if the runtime host is set to `StaticWebApps`.

```csharp
if (Enum.TryParse<EasyAuthType>(runtimeConfig.Runtime.Host.Authentication.Provider, ignoreCase: true, out var provider) &&
    provider == EasyAuthType.StaticWebApps)
{
    _logger.LogWarning("The 'StaticWebApps' authentication provider is deprecated.");
}
```

[...]

PR #2943, opened 2025-10-31
Changed the default auth provider from SWA to AppService

anushakolan:
## Why make this change?

- Closes [#2943](https://github.com/Azure/data-api-builder/issues/2644) Change default auth provider to AppService from StaticWebApps. Azure Static Web Apps EasyAuth is being deprecated, so DAB should no longer default to [StaticWebApps](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) as its authentication provider.
- Moving the default to `AppService` aligns DAB with the long‑term supported `EasyAuth` path while keeping behavior equivalent for existing workloads. `StaticWebApps` remains supported when explicitly configured, but new configurations and `dab init` flows should guide users toward `AppService` instead of a deprecated option.

## What is this change?

-Config and runtime behavior
 - Changed the default authentication provider from `Static Web Apps` to `App Service` in the core configuration model and JSON schema.
 - Added validation that logs a warning when Static Web Apps is explicitly selected (since it’s deprecated as a default).

-CLI and `dab init`
 - Updated `dab init` so that, when no auth provider is specified, it now generates configs using App Service as the provider instead of Static Web Apps.
 - Adjusted CLI configuration generation and option handling so any “default provider” usage now points to App Service.
 - Updated end-to-end CLI tests and initialization tests so their expected configurations and arguments reference App Service as the default.

-Schema, samples, and built‑in configs
 - Updated the JSON schema to set the default of the `authentication.provider` property to `AppService`.
 - Updated sample configuration snippets in the main documentation to show App Service as the provider.
 - Updated the built‑in `dab-config` JSON files (for all supported databases and multi‑DAB scenarios) so their runtime host sections use App Service.

-Engine tests and helpers
 - Updated test helpers to generate EasyAuth principals appropriate for the configured provider, and to treat App Service as the default in REST and GraphQL integration tests.
 - Adjusted configuration and health‑endpoint tests to no longer assume Static Web Apps as the implicit provider and to accept App Service as the default.
 
-Snapshots and expected outputs
 - updated a large set of snapshot files (CLI snapshots, configuration snapshots, entity update/add snapshots) so that anywhere the authentication section previously showed Static Web Apps as the provider, it now shows App Service.

-Note
We updated `AddEnvDetectedEasyAuth` so that it always registers both the `App Service` and `Static Web Apps` `EasyAuth` schemes in development mode, instead of only adding App Service when certain environment variables are present. This aligns with the new default of using App Service as the primary `EasyAuth` provider and makes dev/test/CI behavior deterministic, while still letting configuration (runtime.host.authentication.provider) choose which scheme is actually used.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests

## Sample Request(s)

`dab init --database-type mssql --connection-string "<conn-string>"`

Generates,

`"runtime": {
  "host": {
    "authentication": {
      "provider": "AppService"
    }
  }
}`

Users who still want Static Web Apps can override:

`dab init --database-type mssql --connection-string "<conn-string>" --auth.provider StaticWebApps`

### DIFF
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -4337,6 +4337,176 @@ type Planet @model(name:""PlanetAlias"") {
             }
         }
 
+        /// <summary>
+        /// Test validates that autoentities section can be deserialized and serialized correctly.
+        /// </summary>
+        [DataTestMethod]
+        [TestCategory(TestCategory.MSSQL)]
+        [DataRow(null, null, null, null, null, null, null, null, null, "anonymous", EntityActionOperation.Read)]
+        [DataRow(new[] { "%.%" }, new[] { "%.%" }, "{object}", true, true, true, false, 5, EntityCacheLevel.L1L2, "anonymous", EntityActionOperation.Read)]
+        [DataRow(new[] { "books.%" }, new[] { "books.pages.%" }, "books_{object}", false, false, false, true, 2147483647, EntityCacheLevel.L1, "test-user", EntityActionOperation.Delete)]
+        [DataRow(new[] { "books.%" }, null, "books_{object}", false, null, false, null, 2147483647, null, "test-user", EntityActionOperation.Delete)]
+        [DataRow(null, new[] { "books.pages.%" }, null, null, false, null, true, null, EntityCacheLevel.L1, "test-user", EntityActionOperation.Delete)]
+        [DataRow(new[] { "title.%", "books.%", "names.%" }, new[] { "names.%", "%.%" }, "{schema}.{object}", true, false, false, true, 1, null, "second-test-user", EntityActionOperation.Create)]
+        public void TestAutoEntitiesSerializationDeserialization(
+            string[]? include,
+            string[]? exclude,
+            string? name,
+            bool? restEnabled,
+            bool? graphqlEnabled,
+            bool? healthCheckEnabled,
+            bool? cacheEnabled,
+            int? cacheTTL,
+            EntityCacheLevel? cacheLevel,
+            string role,
+            EntityActionOperation entityActionOp)
+        {
+            TestHelper.SetupDatabaseEnvironment(MSSQL_ENVIRONMENT);
+
+            Dictionary<string, Autoentity> createdAutoentity = new();
+            createdAutoentity.Add("test-entity",
+                new Autoentity(
+                    Patterns: new AutoentityPatterns(include, exclude, name),
+                    Template: new AutoentityTemplate(
+                        Rest: restEnabled == null ? null : new EntityRestOptions(Enabled: (bool)restEnabled),
+                        GraphQL: graphqlEnabled == null ? null : new EntityGraphQLOptions(Singular: string.Empty, Plural: string.Empty, Enabled: (bool)graphqlEnabled),
+                        Health: healthCheckEnabled == null ? null : new EntityHealthCheckConfig(healthCheckEnabled),
+                        Cache: (cacheEnabled == null && cacheTTL == null && cacheLevel == null) ? null : new EntityCacheOptions(Enabled: cacheEnabled, TtlSeconds: cacheTTL, Level: cacheLevel)
+                    ),
+                    Permissions: new EntityPermission[1]));
+
+            EntityAction[] entityActions = new EntityAction[] { new(entityActionOp, null, null) };
+            createdAutoentity["test-entity"].Permissions[0] = new EntityPermission(role, entityActions);
+            RuntimeAutoentities autoentities = new(createdAutoentity);
+
+            FileSystemRuntimeConfigLoader baseLoader = TestHelper.GetRuntimeConfigLoader();
+            baseLoader.TryLoadKnownConfig(out RuntimeConfig? baseConfig);
+
+            RuntimeConfig config = new(
+                Schema: baseConfig!.Schema,
+                DataSource: baseConfig.DataSource,
+                Runtime: new(
+                    Rest: new(),
+                    GraphQL: new(),
+                    Mcp: new(),
+                    Host: new(null, null),
+                    Telemetry: new()
+                ),
+                Entities: baseConfig.Entities,
+                Autoentities: autoentities
+            );
+
+            string configWithCustomJson = config.ToJson();
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(configWithCustomJson, out RuntimeConfig? deserializedRuntimeConfig));
+
+            string serializedConfig = deserializedRuntimeConfig.ToJson();
+
+            using (JsonDocument parsedDocument = JsonDocument.Parse(serializedConfig))
+            {
+                JsonElement root = parsedDocument.RootElement;
+                JsonElement autoentitiesElement = root.GetProperty("autoentities");
+
+                bool entityExists = autoentitiesElement.TryGetProperty("test-entity", out JsonElement entityElement);
+                Assert.AreEqual(expected: true, actual: entityExists);
+
+                // Validate patterns properties and their values exists in autoentities
+                bool expectedPatternsExist = include != null || exclude != null || name != null;
+                bool patternsExists = entityElement.TryGetProperty("patterns", out JsonElement patternsElement);
+                Assert.AreEqual(expected: expectedPatternsExist, actual: patternsExists);
+
+                if (patternsExists)
+                {
+                    bool includeExists = patternsElement.TryGetProperty("include", out JsonElement includeElement);
+                    Assert.AreEqual(expected: (include != null), actual: includeExists);
+                    if (includeExists)
+                    {
+                        CollectionAssert.AreEqual(expected: include, actual: includeElement.EnumerateArray().Select(e => e.GetString()).ToArray());
+                    }
+
+                    bool excludeExists = patternsElement.TryGetProperty("exclude", out JsonElement excludeElement);
+                    Assert.AreEqual(expected: (exclude != null), actual: excludeExists);
+                    if (excludeExists)
+                    {
+                        CollectionAssert.AreEqual(expected: exclude, actual: excludeElement.EnumerateArray().Select(e => e.GetString()).ToArray());
+                    }
+
+                    bool nameExists = patternsElement.TryGetProperty("name", out JsonElement nameElement);
+                    Assert.AreEqual(expected: (name != null), actual: nameExists);
+                    if (nameExists)
+                    {
+                        Assert.AreEqual(expected: name, actual: nameElement.GetString());
+                    }
+                }
+
+                // Validate template properties and their values exists in autoentities
+                bool expectedTemplateExist = restEnabled != null || graphqlEnabled != null || healthCheckEnabled != null
+                    || cacheEnabled != null || cacheLevel != null || cacheTTL != null;
+                bool templateExists = entityElement.TryGetProperty("template", out JsonElement templateElement);
+                Assert.AreEqual(expected: expectedTemplateExist, actual: templateExists);
+
+                if (templateExists)
+                {
+                    bool restPropertyExists = templateElement.TryGetProperty("rest", out JsonElement restElement);
+                    Assert.AreEqual(expected: (restEnabled != null), actual: restPropertyExists);
+                    if (restPropertyExists)
+                    {
+                        Assert.IsTrue(restElement.TryGetProperty("enabled", out JsonElement restEnabledElement));
+                        Assert.AreEqual(expected: restEnabled, actual: restEnabledElement.GetBoolean());
+                    }
+
+                    bool graphqlPropertyExists = templateElement.TryGetProperty("graphql", out JsonElement graphqlElement);
+                    Assert.AreEqual(expected: (graphqlEnabled != null), actual: graphqlPropertyExists);
+                    if (graphqlPropertyExists)
+                    {
+                        Assert.IsTrue(graphqlElement.TryGetProperty("enabled", out JsonElement graphqlEnabledElement));
+                        Assert.AreEqual(expected: graphqlEnabled, actual: graphqlEnabledElement.GetBoolean());
+                    }
+
+                    bool healthPropertyExists = templateElement.TryGetProperty("health", out JsonElement healthElement);
+                    Assert.AreEqual(expected: (healthCheckEnabled != null), actual: healthPropertyExists);
+                    if (healthPropertyExists)
+                    {
+                        Assert.IsTrue(healthElement.TryGetProperty("enabled", out JsonElement healthEnabledElement));
+                        Assert.AreEqual(expected: healthCheckEnabled, actual: healthEnabledElement.GetBoolean());
+                    }
+
+                    bool expectedCacheExist = cacheEnabled != null || cacheTTL != null || cacheLevel != null;
+                    bool cachePropertyExists = templateElement.TryGetProperty("cache", out JsonElement cacheElement);
+                    Assert.AreEqual(expected: expectedCacheExist, actual: cachePropertyExists);
+                    if (cacheEnabled != null)
+                    {
+                        Assert.IsTrue(cacheElement.TryGetProperty("enabled", out JsonElement cacheEnabledElement));
+                        Assert.AreEqual(expected: cacheEnabled, actual: cacheEnabledElement.GetBoolean());
+                    }
+
+                    if (cacheTTL != null)
+                    {
+                        Assert.IsTrue(cacheElement.TryGetProperty("ttl-seconds", out JsonElement cacheTtlElement));
+                        Assert.AreEqual(expected: cacheTTL, actual: cacheTtlElement.GetInt32());
+                    }
+
+                    if (cacheLevel != null)
+                    {
+                        Assert.IsTrue(cacheElement.TryGetProperty("level", out JsonElement cacheLevelElement));
+                        Assert.IsTrue(string.Equals(cacheLevel.ToString(), cacheLevelElement.GetString(), StringComparison.OrdinalIgnoreCase));
+                    }
+                }
+
+                // Validate permissions properties and their values exists in autoentities
+                JsonElement permissionsElement = entityElement.GetProperty("permissions");
+
+                bool roleExists = permissionsElement[0].TryGetProperty("role", out JsonElement roleElement);
+                Assert.AreEqual(expected: true, actual: roleExists);
+                Assert.AreEqual(expected: role, actual: roleElement.GetString());
+
+                bool entityActionsExists = permissionsElement[0].TryGetProperty("actions", out JsonElement entityActionsElement);
+                Assert.AreEqual(expected: true, actual: entityActionsExists);
+                bool entityActionOpExists = entityActionsElement[0].TryGetProperty("action", out JsonElement entityActionOpElement);
+                Assert.AreEqual(expected: true, actual: entityActionOpExists);
+                Assert.IsTrue(string.Equals(entityActionOp.ToString(), entityActionOpElement.GetString(), StringComparison.OrdinalIgnoreCase));
+            }
+        }
+
 #nullable disable
 
         /// <summary>
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -693,6 +693,166 @@
         }
       }
     },
+    "autoentities": {
+      "type": "object",
+      "description": "Defines automatic entity generation rules for MSSQL tables based on include/exclude patterns and defaults.",
+      "patternProperties": {
+        "^.*$": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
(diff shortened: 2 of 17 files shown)


src/Service/Startup.cs
Aniruddh25, line 801: We should change this exception to a `LogWarning`, since going forward AppService is the default, but customers might still use it in a non-App service environment for anonymous use cases for example in which case, it should not error out. But, they should just know that authenticated usecases wont work in that environment.

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
```diff
@@ -75,8 +75,19 @@ public static async Task<IHost> CreateWebHost(
...
+                                // When EasyAuth provider is AppService, set the default scheme accordingly so
+                                // AppServiceAuthentication is registered and available during tests.
+                                if (easyAuthProvider == EasyAuthType.AppService)
+                                {
+                                    services.AddAuthentication(EasyAuthAuthenticationDefaults.APPSERVICEAUTHSCHEME)
+                                        .AddEasyAuthAuthentication(easyAuthProvider);
+                                }
+                                else
+                                {
+                                    services.AddAuthentication()
```
Aniruddh25: how is the else part different here? what scheme gets registered when we dony specify any argument to `AddAuthentication`?

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -654,7 +654,7 @@ private static bool TryParseAndAssertOnDefaults(string json, out RuntimeConfig p
             Assert.AreEqual(McpRuntimeOptions.DEFAULT_PATH, parsedConfig.McpPath);
             Assert.IsTrue(parsedConfig.AllowIntrospection);
             Assert.IsFalse(parsedConfig.IsDevelopmentMode());
-            Assert.IsTrue(parsedConfig.IsStaticWebAppsIdentityProvider);
+            Assert.IsTrue(parsedConfig.IsAppServiceIdentityProvider);
```
Aniruddh25, line 659: If there are no more callers of `IsStaticWebAppsIdentityProvider`, we should remove this function.

src/Service/Program.cs
```diff
@@ -269,9 +269,21 @@ public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
...
+        // host, including App Service as the default EasyAuth provider where
+        // configured.
         public static IWebHostBuilder CreateWebHostFromInMemoryUpdatableConfBuilder(string[] args) =>
-            WebHost.CreateDefaultBuilder(args)
-            .UseStartup<Startup>();
+            WebHost
+                .CreateDefaultBuilder(args)
+                .ConfigureAppConfiguration((_, builder) =>
+                {
+                    AddConfigurationProviders(builder, args);
```
Aniruddh25: When StaticWebApps was the default auth provider, we did not need to add the same configuration providers here. What changed when we moved to AppService, that these providers are required now even for in-memory updateable configuration?

src/Service/Startup.cs
```diff
@@ -798,8 +798,16 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
 
-                    services.AddAuthentication(EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME)
+                    string scheme = easyAuthType switch
+                    {
+                        EasyAuthType.AppService => EasyAuthAuthenticationDefaults.APPSERVICEAUTHSCHEME,
+                        EasyAuthType.StaticWebApps => EasyAuthAuthenticationDefaults.SWAAUTHSCHEME,
+                        _ => EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME
+                    };
+
+                    services.AddAuthentication(scheme)
```
Aniruddh25: Previously, when SWAAUTHSCHEME was the scheme, we did not need to provide that as the value to the AddAuthentication function. the generic value of EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME = EasyAuthAuthentication was sufficient. What changed with AppService, that we need to specify the explicit value of the scheme to the AddAuthentication function?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3377,6 +3389,23 @@ public async Task TestEngineSupportViewsWithoutKeyFieldsInConfigForMsSQL()
 
             RuntimeConfig configuration = InitMinimalRuntimeConfig(dataSource, new(), new(), new(), viewEntity, "books_view_all");
 
+            // For this test we only care that views without key fields
+            // are supported; use a provider with a registered handler
```
Aniruddh25: Why is the registered handler StaticWebApps? Is the function InitMinimalRuntimeConfig adding that provider? If the default now is AppService, we should not need to add StaticWebApps even for just the test failures - if we are required to do so, it means somewhere we are not adding the default handler accurately.

src/Service/Startup.cs
```diff
@@ -798,8 +798,16 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
                         }
```
Aniruddh25: Given that the default authProvider is now AppService, we should always simply log a warning even if its production mode and appservice environment config is missing. The exception above may restrict usage of DAB in case of anonymous access outside of appservice environment which we should avoid. Having said that, the above exception with `message: AppServiceAuthenticationInfo.APPSERVICE_PROD_MISSING_ENV_CONFIG,` should be removed.

src/Service/Startup.cs
```diff
@@ -798,8 +798,16 @@ private void ConfigureAuthentication(IServiceCollection services, RuntimeConfigP
...
+                    {
+                        EasyAuthType.AppService => EasyAuthAuthenticationDefaults.APPSERVICEAUTHSCHEME,
+                        EasyAuthType.StaticWebApps => EasyAuthAuthenticationDefaults.SWAAUTHSCHEME,
+                        _ => EasyAuthAuthenticationDefaults.AUTHENTICATIONSCHEME
+                    };
+
+                    services.AddAuthentication(scheme)
                         .AddEasyAuthAuthentication(easyAuthAuthenticationProvider: easyAuthType);
+                    _logger.LogInformation($"Registered EasyAuth scheme: {scheme}");
                 }
```
Aniruddh25, line 811: As per the Description, `AddEnvDetectedEasyAuth` should be modified to add AppService irrespective of detection of the AppService environment. But that function has not been modified. We should fix that function to depend on the boolean `appServiceEnvironmentDetected`, instead always add the AppService provider.

aaronburtle:
In the description where it says `closes` it self-references this PR, I think you meant to use this link to the issue https://github.com/Azure/data-api-builder/issues/2644

aaronburtle:
Since SWA is a deprecated option now, should we log that it is deprecated for the user if they are using it?

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
aaronburtle: I think because in the tests we are calling 

`SendRequestAndGetHttpContextState(string? token, EasyAuthType easyAuthType, ...)`

With the `EasyAuthType` and that is used in

`CreateWebHost(easyAuthType.ToString(), useAuthorizationMiddleware)`

We don't need the `EasyAuthAuthenticationDefaults.APPSERVICEAUTHSCHEME` in the call to AddAuthentication for the tests as the comment above the branch suggests, because that type is only needed in `AddAuthentication` when you want the handler to invoke said type when none is provided. But we are providing that type in the tests, both for `SWA` and `AppService`. If there's another reason we need it, then the comment should reflect that, but I don't see one.

src/Service/Startup.cs
anushakolan: I did not get this Ani, that function has been modified to always add `AppService` provider. Let me know if I am missing something.

src/Service/Startup.cs
anushakolan: 1. Previously there was only one `EasyAuth `handler, so a single generic scheme name was enough.
2. Now we have two distinct `EasyAuth` handlers, so we:
Register both, and
Tell ASP.NET Core exactly which named scheme should be the default (`StaticWebAppsAuthentication` or `AppServiceAuthentication`).

src/Service/Startup.cs
anushakolan: We want both `EasyAuth` schemes wired every time irrespective of `appServiceEnvironmentDetected`, so that:
`ClientRoleHeaderAuthenticationMiddleware` can resolve either `StaticWebAppsAuthentication` or `AppServiceAuthentication`.
Tests or callers asking for `StaticWebAppsAuthentication` don’t throw “no handler registered”.

src/Service/Startup.cs
anushakolan: Done

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
anushakolan: No, it is still used in a few places.

src/Service/Program.cs
anushakolan: Removed

src/Service.Tests/Configuration/ConfigurationTests.cs
anushakolan: Removed

anushakolan:
> Since SWA is a deprecated option now, should we log that it is deprecated for the user if they are using it?

Weare adding it already.

src/Service.Tests/Authentication/Helpers/WebHostBuilderHelper.cs
anushakolan: No longer needed, reverted the change.

anushakolan:
> Waiting on answers to some questions.

Answered the questions, please re-review it.

aaronburtle (approved):
Looks good!

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -762,10 +796,20 @@ public async Task CanPatchItemWithoutVariables()
     }}
 }}";
             JsonElement response = await ExecuteGraphQLRequestAsync("patchPlanet", mutation, variables: new());
+
+            JsonElement result = response;
+
+            // In some environments ExecuteGraphQLRequestAsync may return the items array,
```
Aniruddh25: Whether GraphQLRequestAsync returns items as an array or not, should be deterministic. Why would it differ when the auth provider is different?

anushakolan: This change is not needed, reverting.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -62,10 +62,31 @@ public async Task InvokeAsync(HttpContext httpContext)
...
+            }
+            else
+            {
+                scheme = EasyAuthAuthenticationDefaults.SWAAUTHSCHEME;
+            }
+        }
+        else
         {
-            AuthenticationOptions? dabAuthNOptions = _runtimeConfigProvider.GetConfig().Runtime?.Host?.Authentication;
+            var dabAuthNOptions = _runtimeConfigProvider.GetConfig().Runtime?.Host?.Authentication;
```
Aniruddh25: using `var` is discouraged unless the type is completely obvious from the RHS.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -62,10 +62,31 @@ public async Task InvokeAsync(HttpContext httpContext)
         // Determine the authentication scheme to use based on dab-config.json.
         // Compatible with both ConfigureAuthentication and ConfigureAuthenticationV2 in startup.cs.
         // This means that this code is resilient to whether or not the default authentication scheme is set in startup.
-        string scheme = EasyAuthAuthenticationDefaults.SWAAUTHSCHEME;
-        if (!_runtimeConfigProvider.IsLateConfigured)
+        string scheme;
+
+        if (_runtimeConfigProvider.IsLateConfigured)
+        {
+            // Hosted/late-config: prefer App Service when its EasyAuth is enabled via env;
```
Aniruddh25: Even in the hosted/late configuration case, we should not rely on setup of the environment variable. Always let app service to be used. 

Is there any restriction to use AppService when the environment variable is missing?

src/Service.Tests/CosmosTests/MutationTests.cs
```diff
@@ -797,19 +797,17 @@ public async Task CanPatchItemWithoutVariables()
...
-            JsonElement result = response;
+            Console.WriteLine($"patchPlanet raw response: {response}");
 
-            // In some environments ExecuteGraphQLRequestAsync may return the items array,
-            // so pick the first element when ValueKind is Array.
-            if (response.ValueKind == JsonValueKind.Array && response.GetArrayLength() > 0)
+            if (response.ValueKind == JsonValueKind.Array)
             {
-                result = response[0];
+                Assert.Fail($"Expected patchPlanet to return a single result object, but got an array: {response}");
```
Aniruddh25: Why are we changing the expectation of this test? Why do we expect a single element?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5270,12 +5270,34 @@ private static JsonContent GetPostStartupConfigParams(string environment, Runtim
...
+                Content = content
+            };
+
+            postRequest.Headers.Add(
+                Config.ObjectModel.AuthenticationOptions.CLIENT_PRINCIPAL_HEADER,
+                appServiceTokenPayload);
+
+            HttpResponseMessage postResult = await httpClient.SendAsync(postRequest);
+            string body = await postResult.Content.ReadAsStringAsync();
+            Assert.AreEqual(HttpStatusCode.Unauthorized, postResult.StatusCode, body);
```
Aniruddh25: why dont we expect HttpStatusCode.OK anymore?

src/Service.Tests/TestHelper.cs
```diff
@@ -367,21 +367,68 @@ public static void ConstructNewConfigWithSpecifiedHostMode(string configFileName
             RuntimeConfigProvider configProvider = GetRuntimeConfigProvider(GetRuntimeConfigLoader());
             RuntimeConfig config = configProvider.GetConfig();
 
+            // Decide whether to keep base route based on provider
+            AuthenticationOptions auth = config.Runtime?.Host?.Authentication;
+            bool isStaticWebApps = string.Equals(auth?.Provider, "StaticWebApps", StringComparison.OrdinalIgnoreCase);
```
Aniruddh25: Who determines which auth provider the config contains?

If its the respective test, that populates the provider, why cant we have the caller of `ConstructNewConfigWithSpecifiedHostModeStaticWebApps` specify SWA as the provider? instead of adding an explicit new function: `ConstructNewConfigWithSpecifiedHostMode`

Aniruddh25 (changes requested):
Still waiting for successful tests run and answer to some questions

src/Service.Tests/Configuration/ConfigurationTests.cs
anushakolan: Fixed this

src/Service.Tests/CosmosTests/MutationTests.cs
anushakolan: No we don't, this was written as part of test debugging and I missed claning it up.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
anushakolan: No, we don't have a restriction as of that. I added support to just in case if some use cases might still need SWA. Removing this.

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
```diff
@@ -62,7 +62,7 @@ public async Task InvokeAsync(HttpContext httpContext)
         // Determine the authentication scheme to use based on dab-config.json.
         // Compatible with both ConfigureAuthentication and ConfigureAuthenticationV2 in startup.cs.
         // This means that this code is resilient to whether or not the default authentication scheme is set in startup.
-        string scheme = EasyAuthAuthenticationDefaults.SWAAUTHSCHEME;
+        string scheme = EasyAuthAuthenticationDefaults.APPSERVICEAUTHSCHEME;
```
Aniruddh25, line 65: I cant see what changed in the latest commits - since there may have been a force push. To help my understanding, how did you resolve the last of the tests by setting this to AppService here? so that for late configured scenarios, it doesnt error out anymore? What really was the root cause of the failures.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2498,7 +2510,7 @@ public async Task TestRuntimeBaseRouteInNextLinkForPaginatedRestResponse()
         {
             const string CUSTOM_CONFIG = "custom-config.json";
             string runtimeBaseRoute = "/base-route";
-            TestHelper.ConstructNewConfigWithSpecifiedHostMode(CUSTOM_CONFIG, HostMode.Production, TestCategory.MSSQL, runtimeBaseRoute: runtimeBaseRoute);
+            TestHelper.ConstructNewConfigWithSpecifiedHostMode(CUSTOM_CONFIG, HostMode.Production, TestCategory.MSSQL, runtimeBaseRoute: runtimeBaseRoute, "StaticWebApps");
```
Aniruddh25, line 2513: Why do we still need StaticWebApps here?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3590,7 +3621,7 @@ type Planet @model(name:""PlanetAlias"") {
         [TestCategory(TestCategory.MSSQL)]
         [DataRow(HostMode.Development, EasyAuthType.AppService, false, false, DisplayName = "AppService Dev - No EnvVars - No Error")]
         [DataRow(HostMode.Development, EasyAuthType.AppService, true, false, DisplayName = "AppService Dev - EnvVars - No Error")]
-        [DataRow(HostMode.Production, EasyAuthType.AppService, false, true, DisplayName = "AppService Prod - No EnvVars - Error")]
```
Aniruddh25, line 3593: Provide a reason in comment why we dont expect error here anymore

src/Service.Tests/TestHelper.cs
```diff
@@ -29,6 +30,36 @@ public static void UnsetAllDABEnvironmentVariables()
             Environment.SetEnvironmentVariable(FileSystemRuntimeConfigLoader.RUNTIME_ENV_CONNECTION_STRING, null);
         }
 
+        /// <summary>
+        /// Sets environment variables to simulate an Azure App Service EasyAuth environment.
```
Aniruddh25: Since we no longer depend on the env vars to be setup, and dont error out if they are not, do we really need to set these environment variables in the TestHelper?

src/Core/AuthenticationHelpers/ClientRoleHeaderAuthenticationMiddleware.cs
anushakolan: I did not do a force push Ani. I see all the commits, but their time stamps were updated, this is weird, but I did not give a force push command as you explicitly mentioned not to use that earlier.

This was the root cause; to give you some context, this was one of your previous comments,
https://github.com/Azure/data-api-builder/pull/2943#discussion_r2629084523

We are setting the default auth provider as `AppService` everywhere in the config but here earlier it was relying on the environment to get App Service auth scheme. So, I first fixed the code by updating the test to explicitly set the `AppService` environment but later based on your comment I understood we should not set it based on the environment, instead the default should always be `AppService`. So, when I changed the default to `AppService` Scheme, the tests ran fine.

src/Service.Tests/Configuration/ConfigurationTests.cs
anushakolan: Okay

src/Service.Tests/Configuration/ConfigurationTests.cs
anushakolan: The runtime base route feature is only relevant in the context of Azure Static Web Apps (SWA) integration. When DAB is running in SWA mode, it expects to be behind the SWA reverse proxy, which may rewrite the base route (e.g., /base-route) before forwarding requests to DAB. The test is verifying that the generated nextLink in paginated REST responses includes the configured base route, which is a behavior specific to SWA integration scenarios.

PR #2944, opened 2025-10-31
Fix the Serialization/Deserialization issue with $ prefix columns

Alekhya-Polavarapu:
### Why make this change?
Serialization and deserialization of metadata currently fail when column names are prefixed with the $ symbol.

### Root cause
This issue occurs because we’ve enabled the ReferenceHandler flag in our System.Text.Json serialization settings. When this flag is active, the serializer treats $ as a reserved character used for special metadata (e.g., $id, $ref). As a result, any property name starting with $ is interpreted as metadata and cannot be deserialized properly.

### What is this change?
This update introduces custom logic in the converter’s Write and Read methods to handle $-prefixed column names safely.

- During serialization, columns beginning with $ are escaped as "_$".

- During deserialization, this transformation is reversed to restore the original property names.

### How was this tested

- [x] Unit tests

### DIFF
--- a/src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
+++ b/src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
@@ -318,7 +318,7 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
         [TestMethod]
         public void TestDatabaseViewSerializationDeserialization_WithDollarColumn()
         {
-            InitializeObjects(true);
+            InitializeObjects(generateDollaredColumn: true);
 
             TestTypeNameChanges(_databaseView, "DatabaseView");
 
@@ -345,7 +345,7 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
         [TestMethod]
         public void TestDatabaseStoredProcedureSerializationDeserialization_WithDollarColumn()
         {
-            InitializeObjects(true);
+            InitializeObjects(generateDollaredColumn: true);
 
             TestTypeNameChanges(_databaseStoredProcedure, "DatabaseStoredProcedure");
 


src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
```diff
@@ -17,6 +17,8 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
     public class DatabaseObjectConverter : JsonConverter<DatabaseObject>
     {
         private const string TYPE_NAME = "TypeName";
+        private const string DOLLAR_CHAR = "$";
+        private const string ESCAPED_DOLLARCHAR = "_$";
```
Alekhya-Polavarapu: Can we confidently escape it to "_$", may be there might be columns that are genuinely created with _$.

souvikghosh04: `$` or `_$` are allowed in some databases but used in different ways. for e.g., in PostgreSQL, the column should be used in within double-quotes. In MS SQL, its allowed but in box brackets I suppose. MySQL allows too. its better to handle these edge cases and also test it with different databases.

JerryNixon: We should consider if this is an edge case we care about.

src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
```diff
@@ -276,8 +276,96 @@ public void TestDictionaryDatabaseObjectSerializationDeserialization()
...
+
+        /// <summary>
+        /// Validates serialization and deserilization of Dictionary containing DatabaseStoredProcedure
+        /// The table will have dollar sign prefix ($) in the column name
+        /// this is how we serialize and deserialize metadataprovider.EntityToDatabaseObject dict.
+        /// </summary>
+        [TestMethod]
+        public void TestDatabaseStoredProcedureSerializationDeserialization_WithDollarColumn()
+        {
+            InitializeObjects(true);
```
Aniruddh25: nit: name the argument whose value is `true` here
```suggestion
            InitializeObjects( generateDollaredColumn: true);
```

src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
Aniruddh25: @Alekhya-Polavarapu - can you add the escaped char based on the database type, and just use such an escaped char that already is used by the respective db itself. So, in case it is `\` for MSSQL, use that. and distinguish usage of \ by escaping it with 2 backslashes `\\`

Aniruddh25: **Prompt [what are the valid characters in a database... | Try in Copilot Chat](https://m365.cloud.microsoft/chat/entity1-d870f6cd-4aa5-4d42-9626-ab690c041429/eyJpZCI6IlZYTmxjbFl4ZkdoMGRIQnpPaTh2YzNWaWMzUnlZWFJsTFdsdWRDNXZabVpwWTJVdVkyOXRMM3hQU1VRNllqSXlZekZoTmpRdFpqRmtNQzAwTlRZMExXSm1NbU10TW1WbU1qSXlPVEZsTUdNM2ZEUm1OVGt6TmpVeUxUaGpObUl0TkRjME5pMDRaV0ZtTFRCaE1qRTRNalk0TVdGa01ud3lNREkxTFRFeExUQTRWREF3T2pNME9qRXlMamswTWpjME9UUmEiLCJzY2VuYXJpbyI6InNoYXJlTGlua1ZpYVJpY2hDb3B5IiwicHJvcGVydGllcyI6eyJwcm9tcHRTb3VyY2UiOiJ1c2VyIiwiY2xpY2tUaW1lc3RhbXAiOiIyMDI1LTExLTA4VDAwOjM0OjEzLjk5OFoifSwiY2hhdFR5cGUiOiJ3b3JrIiwidmVyc2lvbiI6MS4xfQ)**
what are the valid characters in a database column name in Microsoft SQL Server?
**Response**
In **Microsoft SQL Server**, the rules for valid column names are:

---

### ✅ **Allowed Characters**
- **Letters**: `A–Z` and `a–z`
- **Digits**: `0–9` (but the name **cannot start with a digit**)
- **Underscore**: `_`
- **At sign**: `@` (only for local variables, not recommended for column names)
- **Hash**: `#` (used for temporary tables, not typical for columns)

---

### ❌ **Not Allowed**
- Spaces (unless enclosed in delimiters)
- Special characters like `!`, `?`, `%`, `^`, `&`, `*`, `(`, `)`, `+`, `=`, `/`, `\`, `|`, `~`
- Control characters or non-printable characters

---

### **Length**
- Maximum: **128 characters**

---

### **Delimiters for Exceptions**
If you need spaces or reserved keywords:
- Use **square brackets**:  
  ```sql
  [Column Name With Spaces]
  ```
- Or **double quotes** (if `QUOTED_IDENTIFIER` is ON):  
  ```sql
  "Column Name With Spaces"
  ```

---

### **Best Practices**
- Start with a letter or underscore
- Avoid reserved keywords (e.g., `SELECT`, `TABLE`)
- Stick to alphanumeric and underscore for portability

---

Do you want me to also include **examples of valid and invalid column names** for SQL Server? Or a **quick T-SQL script to check if a name is valid**?

Aniruddh25: For e.g. You can also use any of the not allowed characters as escape chars since that guarantees they wont be present in the column name.

Alekhya-Polavarapu: Had a discussion with @Aniruddh25  on this and updated the changes accordingly!

in nutshell: its hard to come up with escaping characters for each flavor of sql, because, MSSQL nearly accepts everything when its escaped between [...] and for PGSQL everything is accepted except NULL's(iirc)
So, went with this as ESCAPE CHAR for "$": **DAB_ESCAPE$**

**$ -> DAB_ESCAPE$** --> in serialization 
**DAB_ESCAPE$ -> $** -- in deserialization

src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
```diff
@@ -277,8 +277,96 @@ public void TestDictionaryDatabaseObjectSerializationDeserialization()
...
         {
+            InitializeObjects(generateDollaredColumn: true);
+
+            _options = new()
+            {
+                Converters = {
+                    new DatabaseObjectConverter(),
+                    new TypeConverter()
+                },
+                ReferenceHandler = ReferenceHandler.Preserve,
```
RubenCerna2079: Does there need to be a comma at the end of this line?

Aniruddh25 (approved):

PR #2952, opened 2025-11-05
User provided dml-tools property serialization

souvikghosh04:
## Why make this change?

This PR addresses the issue where with serialization of `dml-tools` property for user-provided (non-default) values. This is a follow up to the fix made for the functionality of `dml-tools` property- https://github.com/Azure/data-api-builder/pull/2927

## What is this change?

If `dml-tools` is not configured, this property should not be serialized. Serialization should only be done for user provided configuration. So, any value explicitly set or configured by the user for `dml-tools` will be serialized.
`dml-tools` can either have either of the below values-
- A boolean value- true or false. This is a global value to enable or disable all DML tools and `UserProvidedAllTools` is written in JSON
- A dictionary object of individual tool name and boolean values. This contains individual tool specific values and only the specified tools will be taken for JSON writing and unspecified tool names will be ignored.

## How was this tested?

- [x] Integration Tests
- [x] Unit Tests
- [x] Manual Tests

Sample scenarios for testing-

**Scenario 1:** Enable all tools using a single boolean value

`"dml-tools": true`

Note: default is true so even if `dml-tools` unspecified it will default to true.

**Scenario 2:** Disable `execute-entities` only and keep other tools enabled as default.

```
      "dml-tools": {
        "execute-entities":  false
      }
```

**Scenario 3:** Use full list of tools and enable or disable them

```
      "dml-tools": {
        "execute-entity": true,
        "delete-record": false,
        "update-record": false,
        "read-records": false,
        "describe-entities": true
      }
```

### DIFF
--- a/src/Config/Converters/DmlToolsConfigConverter.cs
+++ b/src/Config/Converters/DmlToolsConfigConverter.cs
@@ -144,8 +144,7 @@ internal class DmlToolsConfigConverter : JsonConverter<DmlToolsConfig>
             writer.WritePropertyName("dml-tools");
             writer.WriteBooleanValue(value.AllToolsEnabled);
         }
-
-        if (hasIndividualSettings)
+        else if (hasIndividualSettings)
         {
             writer.WritePropertyName("dml-tools");
 


src/Config/ObjectModel/McpRuntimeOptions.cs
```diff
@@ -51,7 +51,8 @@ public McpRuntimeOptions(
         // if DmlTools is null, set All tools enabled by default
         if (DmlTools is null)
         {
-            this.DmlTools = DmlToolsConfig.FromBoolean(DmlToolsConfig.DEFAULT_ENABLED);
+            // Use Default instead of FromBoolean to avoid setting UserProvided flags
+            this.DmlTools = DmlToolsConfig.Default;
```
Aniruddh25, line 55: Where does FromBoolean get called from? Is there any reference? Do we still need that function?

souvikghosh04: Yes its been used in `ConfigGenerator.cs` and `DmlToolsConfigConverter.cs`. The `FromBoolean()` method is used by both the JSON converter for deserializing boolean config values and the CLI for the runtime mcp dml-tools enabled option. It serves the Default property by marking that the user explicitly provided a boolean value, which affects serialization behavior.

src/Config/ObjectModel/DmlToolsConfig.cs
```diff
@@ -64,61 +64,79 @@ public DmlToolsConfig(
         if (allToolsEnabled is not null)
         {
             AllToolsEnabled = allToolsEnabled.Value;
-            UserProvidedAllToolsEnabled = true;
+            UserProvidedAllTools = true;
+
```
anushakolan, line 68: So, will use still provide all tools, when there is a value set set for allToolsEnabled? Like why will we have both of them?

souvikghosh04: These 2 are separate fields and serves different purpose. `AllToolsEnabled` is true by default. However, `UserProvidedAllTools` will be false, unless user adds `"dml-tools": true` or `"dml-tools": false` explictly. By default it won't be available in the config and still `AllToolsEnabled` will be true, as by default. We won't write to file unless `UserProvidedAllTools` is true.

anushakolan (approved):

PR #2900, opened 2025-10-09
[MCP] describe_entities tool fixes and refactoring

souvikghosh04:
### Closes on
- https://github.com/Azure/data-api-builder/issues/2827

Added fixes and refactored the describe_entities tool to support entity metadata discovery for AI agents and LLM clients before performing CRUD operations.

## What is this change?

- Expose entity descriptions with proper null handling
- Expose field description for entities of type stored-procedure with parameter metadata
- Expose permissions for all entities with distinct values
- Added nameOnly parameter to return lightweight response (entity names and descriptions only)
- Added entities parameter to filter results by specific entity names
- Implemented proper error handling using McpResponseBuilder pattern
- Added support for relationships, mappings, and primary key information
- Improved response structure with metadata section including mode and count

## How was this tested?

Functional testing using Insomnia client by running DAB in localhost and local SQL DB database

- MCP endpoint: http://localhost:5000/mcp
- JSON payload (details below)
- Validating the response

Expected (In-progress) format of entity description in response-

```
{
  "entities": [
    {
      "name": "entity-name",
      "description": "entity description",
      "type": "Table",
      "fields": [
        {
          "name": "field-name",
          "type": "unknown",
          "description": "field description"
        }
      ],
      "permissions": [
        "CREATE",
        "READ",
        "UPDATE",
        "DELETE"
      ],
      "primaryKey": ["id"],
      "relationships": [...],
      "mappings": {...}
    }
  ],
  "count": 1,
  "mode": "full"
}
```

## Sample Request(s)

```
POST http://localhost:5000/mcp

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "describe_entities"
  },
  "id": 1
}
```

```
POST http://localhost:5000/mcp

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "describe_entities",
    "arguments": {
      "nameOnly": true
    }
  },
  "id": 2
}
```

```
POST http://localhost:5000/mcp

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "describe_entities",
    "arguments": {
      "entities": ["Book", "Publisher"]
    }
  },
  "id": 1
}
```

PR #2956, opened 2025-11-06
Honoring incoming request role in determining allowed permissions for describe-entities MCP tool

souvikghosh04:
## Why make this change?

- Addresses follow ups to PR #2900
The `describe_entities` tool response format needed improvements to better align with MCP specifications and provide more accurate, user-scoped information. Key issues included non-specification compliant response fields, overly broad permission reporting across all roles, and inconsistent entity/field naming conventions that didn't prioritize user-friendly aliases.

## What is this change?

- **Removed non-spec fields from response**: Eliminated `mode` and `filter` fields that were not part of the MCP specification
- **Scoped permissions to current user's role**: Modified permissions logic to only return permissions available to the requesting user's role instead of all permissions across all roles
- **Implemented entity alias support**: Updated entity name resolution to prefer GraphQL singular names (aliases) over configuration names, falling back to entity name only when alias is absent
- **Fixed parameter metadata format**: Changed parameter default value key from `@default` to `default` in JSON response
- **Enhanced field name resolution**: Updated field metadata to use field aliases when available, falling back to field names when aliases are absent
- **Added proper authorization context**: Integrated HTTP context and authorization resolver to determine current user's role for permission filtering

## How was this tested?

- [x] Manual Tests

## Sample Request(s)

```

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "describe_entities"
  },
  "id": 1
}
```

```

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "describe_entities",
    "arguments": {
      "nameOnly": true
    }
  },
  "id": 2
}
```

```

{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "describe_entities",
    "arguments": {
      "entities": ["Book", "Publisher"]
    }
  },
  "id": 1
}
```

### DIFF
--- a/src/Cli.Tests/ConfigGeneratorTests.cs
+++ b/src/Cli.Tests/ConfigGeneratorTests.cs
@@ -165,15 +165,7 @@ public class ConfigGeneratorTests
                  },
                  ""mcp"": {
                  ""enabled"": true,
-                 ""path"": ""/mcp"",
-                 ""dml-tools"":{
-                   ""describe-entities"": true,
-                   ""create-record"": true,
-                   ""read-records"": true,
-                   ""update-record"": true,
-                   ""delete-record"": true,
-                   ""execute-entity"":true
-                 }
+                 ""path"": ""/mcp""
                 },
                 ""host"": {
                   ""cors"": {
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestAddingStoredProcedureWithRestMethodsAndGraphQLOperations.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestAddingStoredProcedureWithRestMethodsAndGraphQLOperations.verified.txt
@@ -27,13 +27,13 @@
         UpdateRecord: true,
         DeleteRecord: true,
         ExecuteEntity: true,
-        UserProvidedAllToolsEnabled: false,
-        UserProvidedDescribeEntities: true,
-        UserProvidedCreateRecord: true,
-        UserProvidedReadRecords: true,
-        UserProvidedUpdateRecord: true,
-        UserProvidedDeleteRecord: true,
-        UserProvidedExecuteEntity: true
+        UserProvidedAllTools: false,
+        UserProvidedDescribeEntities: false,
+        UserProvidedCreateRecord: false,
+        UserProvidedReadRecords: false,
+        UserProvidedUpdateRecord: false,
+        UserProvidedDeleteRecord: false,
+        UserProvidedExecuteEntity: false
       }
     },
     Host: {
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceAsStoredProcedure.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceAsStoredProcedure.verified.txt
@@ -27,13 +27,13 @@
         UpdateRecord: true,
         DeleteRecord: true,
         ExecuteEntity: true,
-        UserProvidedAllToolsEnabled: false,
-        UserProvidedDescribeEntities: true,
-        UserProvidedCreateRecord: true,
-        UserProvidedReadRecords: true,
-        UserProvidedUpdateRecord: true,
-        UserProvidedDeleteRecord: true,
-        UserProvidedExecuteEntity: true
+        UserProvidedAllTools: false,
+        UserProvidedDescribeEntities: false,
+        UserProvidedCreateRecord: false,
+        UserProvidedReadRecords: false,
+        UserProvidedUpdateRecord: false,
+        UserProvidedDeleteRecord: false,
+        UserProvidedExecuteEntity: false
       }
     },
     Host: {
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceWithDefaultType.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceWithDefaultType.verified.txt
@@ -27,13 +27,13 @@
         UpdateRecord: true,
         DeleteRecord: true,
         ExecuteEntity: true,
-        UserProvidedAllToolsEnabled: false,
-        UserProvidedDescribeEntities: true,
-        UserProvidedCreateRecord: true,
-        UserProvidedReadRecords: true,
-        UserProvidedUpdateRecord: true,
-        UserProvidedDeleteRecord: true,
-        UserProvidedExecuteEntity: true
+        UserProvidedAllTools: false,
+        UserProvidedDescribeEntities: false,
+        UserProvidedCreateRecord: false,
+        UserProvidedReadRecords: false,
+        UserProvidedUpdateRecord: false,
+        UserProvidedDeleteRecord: false,
+        UserProvidedExecuteEntity: false
       }
     },
     Host: {
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithoutIEnumerables.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithoutIEnumerables.verified.txt
@@ -27,13 +27,13 @@
         UpdateRecord: true,
         DeleteRecord: true,
         ExecuteEntity: true,
-        UserProvidedAllToolsEnabled: false,
-        UserProvidedDescribeEntities: true,
-        UserProvidedCreateRecord: true,
-        UserProvidedReadRecords: true,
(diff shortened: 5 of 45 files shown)


src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -80,6 +83,25 @@ public Task<CallToolResult> ExecuteAsync(
...
+                string? currentUserRole = null;
+                if (httpContext != null)
+                {
+                    string roleHeader = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER].ToString();
+                    if (!string.IsNullOrWhiteSpace(roleHeader))
+                    {
+                        // Take the first role if multiple roles are provided
+                        currentUserRole = roleHeader
+                            .Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                            .FirstOrDefault();
```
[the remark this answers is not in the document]
souvikghosh04: For the `DescribeEntitiesTool`, we don't have a specific operation like CREATE, READ, UPDATE, DELETE that we can check permissions against. The tool is more of a discovery tool that shows what's available.

However, the Copilot comment is valid - we should be consistent with the pattern used by other tools. The challenge is that `DescribeEntitiesTool` doesn't check permissions for a specific operation on each entity. So, the below points will be addressed.

- Added proper authorization resolver back to validate role context
- Added comprehensive comments explaining why we use the first role for discovery operations
- Used `IsValidRoleContext` check to ensure the role is valid before proceeding
- Documented that discovery tools like `describe_entities` behave differently from operation-specific tools

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -2,11 +2,14 @@
 // Licensed under the MIT License.
```
Aniruddh25, line 3: the PR title should reflect "what" the fixes are related to. The title goes into the Release notes as is. Saying fixes to describe-entities tool doesnt convey much.

It should mention - honoring incoming request role in determining allowed permissions.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -80,8 +83,35 @@ public Task<CallToolResult> ExecuteAsync(
...
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
+
+                // Get current user's role for permission filtering
+                // For discovery tools like describe_entities, we use the first valid role from the header
+                // This differs from operation-specific tools that check permissions per entity per operation
+                string? currentUserRole = null;
+                if (httpContext != null && authResolver.IsValidRoleContext(httpContext))
+                {
+                    string roleHeader = httpContext.Request.Headers[AuthorizationResolver.CLIENT_ROLE_HEADER].ToString();
```
Aniruddh25, line 97: Did you verify if the role header is populated with `anonymous` when request is not authenticated?

souvikghosh04: yes this is tested locally. here we can see the user's request is not authenticated and `roleHeader` is `Anonymous`.
image

souvikghosh04:
> LGTM as long as you have verified `anonymous` role value is set on the X_MS_API_ROLE header

confirming the same here after validating.
image

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -80,8 +83,35 @@ public Task<CallToolResult> ExecuteAsync(
...
+                    if (!string.IsNullOrWhiteSpace(roleHeader))
+                    {
+                        // For discovery operations, take the first role from comma-separated list
+                        // This provides a consistent view of available entities for the primary role
+                        currentUserRole = roleHeader
```
anushakolan: Shouldn't we throw error, if we are having multiple roles being sent, rather than taking the first role. It could be error prone.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -80,8 +83,35 @@ public Task<CallToolResult> ExecuteAsync(
...
+                            .Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                            .FirstOrDefault();
+                    }
+                }
+
                 (bool nameOnly, HashSet<string>? entityFilter) = ParseArguments(arguments, logger);
 
+                if (currentUserRole == null)
+                {
+                    logger?.LogWarning("Current user role could not be determined from HTTP context or role header. Returning empty permissions for all entities.");
```
anushakolan: I did not get this. 
1. Like does empty permissions mean anonymous permissions? May be that is not clear here.
2. We are not returning anything here, is it "Using" empty permissions?

In both cases, I think it is better if we update the warning message to reflect these.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -140,19 +170,13 @@ public Task<CallToolResult> ExecuteAsync(
...
                 };
 
-                if (entityFilter != null && entityFilter.Count > 0)
-                {
-                    responseData["filter"] = entityFilter.ToArray();
-                }
-
                 logger?.LogInformation(
-                    "DescribeEntitiesTool returned {EntityCount} entities in {Mode} mode.",
+                    "DescribeEntitiesTool returned {EntityCount} entities with nameOnly={NameOnly}.",
```
anushakolan: what is nameOnly here? Maybe we can be more descriptive here.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -348,27 +388,29 @@ private static List<object> BuildParameterMetadataInfo(List<ParameterMetadata>?
...
-                    result.Add(new
+                    Dictionary<string, object?> paramInfo = new()
                     {
-                        name = param.Name,
-                        required = param.Default == null, // required if no default
-                        @default = param.Default,
-                        description = param.Description ?? string.Empty
-                    });
+                        ["name"] = param.Name,
+                        ["required"] = param.Default == null, // required if no default
```
anushakolan: Isn't this `param.required`? I did not get this, how are setting required value based on Default and we are also not considering the `param.required` value.

anushakolan (changes requested):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
souvikghosh04: The `DescribeEntitiesTool` is a discovery tool that provides a consistent view of available entities and permissions, unlike operation-specific tools that check permissions per entity-operation pair. Using the first role ensures uniform perspective across all entities in the response. the enhanced logging now clearly warns users when multiple roles are detected and suggests using a single role for consistency. Throwing an error would reduce usability for legitimate scenarios where users want to understand permissions for their primary.
We need to keep the existing behavior (using first role) but made it transparent and documented.

I had addressed some additional details here- https://github.com/Azure/data-api-builder/pull/2956#discussion_r2498772141.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
souvikghosh04: Improved the logging here.
Please note- `currentUserRole` can not be empty since Auth middle sets it to `Anonymous` as default. This is additional check in case headers yields a null value or Auth middleware's `IsValidRoleContext()` is invalid.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
souvikghosh04: Added it here.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
souvikghosh04: This is needed since we want to explicitly mention the exact MCP tools schema field `nameOnly`. Added additional message alongside this.

PR #2957, opened 2025-11-06
Support for .akv files to AKV variable replacement.

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2748

## What is this change?

Adds the option to use a local .akv file instead of Azure Key Vault for @AKV('') replacement in the config file during deserialization. Similar to how we handle .env files.

## How was this tested?

A new test was added that verifies we are able to do the replacement and get the correct resultant configuration.

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
@@ -57,20 +57,22 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
             CancellationToken cancellationToken = default)
         {
             ILogger<CreateRecordTool>? logger = serviceProvider.GetService<ILogger<CreateRecordTool>>();
+            string toolName = GetToolMetadata().Name;
             if (arguments == null)
             {
-                return Utils.McpResponseBuilder.BuildErrorResult("Invalid Arguments", "No arguments provided", logger);
+                return Utils.McpResponseBuilder.BuildErrorResult(toolName, "Invalid Arguments", "No arguments provided", logger);
             }
 
             RuntimeConfigProvider runtimeConfigProvider = serviceProvider.GetRequiredService<RuntimeConfigProvider>();
             if (!runtimeConfigProvider.TryGetConfig(out RuntimeConfig? runtimeConfig))
             {
-                return Utils.McpResponseBuilder.BuildErrorResult("Invalid Configuration", "Runtime configuration not available", logger);
+                return Utils.McpResponseBuilder.BuildErrorResult(toolName, "Invalid Configuration", "Runtime configuration not available", logger);
             }
 
             if (runtimeConfig.McpDmlTools?.CreateRecord != true)
             {
                 return Utils.McpResponseBuilder.BuildErrorResult(
+                    toolName,
                     "ToolDisabled",
                     "The create_record tool is disabled in the configuration.",
                     logger);
@@ -84,13 +86,13 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                 if (!root.TryGetProperty("entity", out JsonElement entityElement) ||
                     !root.TryGetProperty("data", out JsonElement dataElement))
                 {
-                    return Utils.McpResponseBuilder.BuildErrorResult("InvalidArguments", "Missing required arguments 'entity' or 'data'", logger);
+                    return Utils.McpResponseBuilder.BuildErrorResult(toolName, "InvalidArguments", "Missing required arguments 'entity' or 'data'", logger);
                 }
 
                 string entityName = entityElement.GetString() ?? string.Empty;
                 if (string.IsNullOrWhiteSpace(entityName))
                 {
-                    return Utils.McpResponseBuilder.BuildErrorResult("InvalidArguments", "Entity name cannot be empty", logger);
+                    return Utils.McpResponseBuilder.BuildErrorResult(toolName, "InvalidArguments", "Entity name cannot be empty", logger);
                 }
 
                 string dataSourceName;
@@ -100,7 +102,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                 }
                 catch (Exception)
                 {
-                    return Utils.McpResponseBuilder.BuildErrorResult("InvalidConfiguration", $"Entity '{entityName}' not found in configuration", logger);
+                    return Utils.McpResponseBuilder.BuildErrorResult(toolName, "InvalidConfiguration", $"Entity '{entityName}' not found in configuration", logger);
                 }
 
                 IMetadataProviderFactory metadataProviderFactory = serviceProvider.GetRequiredService<IMetadataProviderFactory>();
@@ -113,7 +115,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                 }
                 catch (Exception)
                 {
-                    return Utils.McpResponseBuilder.BuildErrorResult("InvalidConfiguration", $"Database object for entity '{entityName}' not found", logger);
+                    return Utils.McpResponseBuilder.BuildErrorResult(toolName, "InvalidConfiguration", $"Database object for entity '{entityName}' not found", logger);
                 }
 
                 // Create an HTTP context for authorization
@@ -123,13 +125,13 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
 
                 if (httpContext is null || !authorizationResolver.IsValidRoleContext(httpContext))
                 {
-                    return Utils.McpResponseBuilder.BuildErrorResult("PermissionDenied", "Permission denied: Unable to resolve a valid role context for update operation.", logger);
+                    return Utils.McpResponseBuilder.BuildErrorResult(toolName, "PermissionDenied", "Permission denied: Unable to resolve a valid role context for update operation.", logger);
                 }
 
                 // Validate that we have at least one role authorized for create
                 if (!TryResolveAuthorizedRole(httpContext, authorizationResolver, entityName, out string authError))
                 {
-                    return Utils.McpResponseBuilder.BuildErrorResult("PermissionDenied", authError, logger);
+                    return Utils.McpResponseBuilder.BuildErrorResult(toolName, "PermissionDenied", authError, logger);
                 }
 
                 JsonElement insertPayloadRoot = dataElement.Clone();
@@ -150,12 +152,13 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                     }
                     catch (Exception ex)
                     {
-                        return Utils.McpResponseBuilder.BuildErrorResult("ValidationFailed", $"Request validation failed: {ex.Message}", logger);
+                        return Utils.McpResponseBuilder.BuildErrorResult(toolName, "ValidationFailed", $"Request validation failed: {ex.Message}", logger);
                     }
                 }
                 else
                 {
                     return Utils.McpResponseBuilder.BuildErrorResult(
+                        toolName,
                         "InvalidCreateTarget",
                         "The create_record tool is only available for tables.",
                         logger);
@@ -185,6 +188,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                     if (isError)
                     {
                         return Utils.McpResponseBuilder.BuildErrorResult(
+                            toolName,
                             "CreateFailed",
                             $"Failed to create record in entity '{entityName}'. Error: {JsonSerializer.Serialize(objectResult.Value)}",
                             logger);
@@ -207,6 +211,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                     if (result is null)
                     {
                         return Utils.McpResponseBuilder.BuildErrorResult(
+                            toolName,
                             "UnexpectedError",
                             $"Mutation engine returned null result for entity '{entityName}'",
                             logger);
@@ -226,7 +231,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
             }
             catch (Exception ex)
             {
-                return Utils.McpResponseBuilder.BuildErrorResult("Error", $"Error: {ex.Message}", logger);
+                return Utils.McpResponseBuilder.BuildErrorResult(toolName, "Error", $"Error: {ex.Message}", logger);
             }
         }
 
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/DeleteRecordTool.cs
@@ -73,6 +73,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
             CancellationToken cancellationToken = default)
         {
             ILogger<DeleteRecordTool>? logger = serviceProvider.GetService<ILogger<DeleteRecordTool>>();
+            string toolName = GetToolMetadata().Name;
 
             try
             {
@@ -87,6 +88,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                 if (config.McpDmlTools?.DeleteRecord != true)
                 {
                     return McpResponseBuilder.BuildErrorResult(
+                        toolName,
                         "ToolDisabled",
                         $"The {this.GetToolMetadata().Name} tool is disabled in the configuration.",
                         logger);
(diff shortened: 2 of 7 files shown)


src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -70,13 +71,65 @@ public DeserializationVariableReplacementSettings(
...
             }
         }
 
+        // Checks if the endpoint is a path to a local .akv file.
+        private static bool IsLocalAkvFileEndpoint(string? endpoint)
+            => !string.IsNullOrWhiteSpace(endpoint)
+               && endpoint.EndsWith(".akv", StringComparison.OrdinalIgnoreCase)
+               && File.Exists(endpoint);
+
+        // Loads key=value pairs from a .akv file, similar to .env style. Lines starting with '#' are comments.
```
Aniruddh25, line 96: How do we read the .env file? Can we reuse the code?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -70,13 +71,65 @@ public DeserializationVariableReplacementSettings(
...
+                string line = rawLine.Trim();
+                if (string.IsNullOrEmpty(line) || line.StartsWith('#'))
+                {
+                    continue;
+                }
+
+                int eqIndex = line.IndexOf('=');
+                if (eqIndex <= 0)
+                {
+                    // Ignore malformed lines silently; could be enhanced to log.
```
anushakolan: We should log this.

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -70,13 +71,65 @@ public DeserializationVariableReplacementSettings(
...
+
+                // Remove optional surrounding quotes
+                if (value.Length >= 2 && ((value.StartsWith('"') && value.EndsWith('"')) || (value.StartsWith('\'') && value.EndsWith('\''))))
+                {
+                    value = value[1..^1];
+                }
+
+                if (!string.IsNullOrEmpty(key))
+                {
+                    secrets[key] = value;
```
anushakolan: Is there a possibility of duplicate keys? If so, we will be overring the previous value, is this the expected behavior? Or do we want to check against duplicate keys?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -70,13 +71,65 @@ public DeserializationVariableReplacementSettings(
...
+               && File.Exists(endpoint);
+
+        // Loads key=value pairs from a .akv file, similar to .env style. Lines starting with '#' are comments.
+        private static Dictionary<string, string> LoadAkvFileSecrets(string filePath)
+        {
+            Dictionary<string, string> secrets = new(StringComparer.OrdinalIgnoreCase);
+            foreach (string rawLine in File.ReadAllLines(filePath))
+            {
+                string line = rawLine.Trim();
+                if (string.IsNullOrEmpty(line) || line.StartsWith('#'))
```
anushakolan, line 103: Are multiple line comments possible, stating with `//`? If so we should also consider those, otherwise we might get wrong values.

src/Config/DeserializationVariableReplacementSettings.cs
anushakolan: We should use `secrets.TryAdd(key, value)` so that we can either log duplicate key values or throw exception.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -667,5 +667,52 @@ private static bool TryParseAndAssertOnDefaults(string json, out RuntimeConfig p
...
+            DeserializationVariableReplacementSettings replacementSettings = new(
+                azureKeyVaultOptions: null,
+                doReplaceEnvVar: false,
+                doReplaceAkvVar: true);
+            bool parsed = RuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig config, replacementSettings: replacementSettings);
+
+            // Assert
+            Assert.IsTrue(parsed, "Config should parse successfully with local AKV file replacement.");
+            Assert.IsNotNull(config, "Config should not be null.");
+            Assert.AreEqual(secretConnectionString, config.DataSource.ConnectionString, "Connection string should be replaced from AKV local file secret.");
```
anushakolan: Why are we not asserting for the `API_KEY`?

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: This is done in a different way, where for env files in the CLI project we eagerly add any values provided in a .env file to the environment variables, but for AKV we lazily add when provided that local file as the endpoint. Made more sense to keep these seperated.

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: This is unexpected, if they have duplicates, the last one will take precedent, but they shouldnt do that.

aaronburtle: will use try add as you suggested and log the duplicates

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: `API_KEY` isnt a part of the json config or the replacement logic its just a part of the file that we generate to parse the `.akv` variable from, so it goes unused in the deserialization process, but simply exists to show we can correctly parse the file even with unrelated lines.

src/Config/DeserializationVariableReplacementSettings.cs
aaronburtle: we shouldnt have to worry about this case, and we dont handle it for the .env files either.

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -47,19 +48,23 @@ public class DeserializationVariableReplacementSettings
 
         private readonly AzureKeyVaultOptions? _azureKeyVaultOptions;
         private readonly SecretClient? _akvClient;
+        private readonly Dictionary<string, string>? _akvFileSecrets; // Local .akv file secrets
+        private readonly ILogger? _logger; // logger for optional diagnostics (malformed lines, etc.)
```
souvikghosh04: nit: we can remove the comments unless really needed. we can rather have them where they used may be, like in method signatures or body.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -667,5 +667,52 @@ private static bool TryParseAndAssertOnDefaults(string json, out RuntimeConfig p
...
         record StubJsonType(string Foo);
+
+        /// <summary>
+        /// Test to verify Azure Key Vault variable replacement from local .akv file.
+        /// </summary>
+        [TestMethod]
+        public void TestAkvVariableReplacementFromLocalFile()
+        {
+            // Arrange: create a temporary .akv secrets file
+            string akvFilePath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString() + ".akv");
```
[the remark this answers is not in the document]
aaronburtle: newguid() returns a guid not a string.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3319,7 +3319,8 @@ public async Task ValidateStrictModeAsDefaultForRestRequestBody(bool includeExtr
                 HttpMethod httpMethod = SqlTestHelper.ConvertRestMethodToHttpMethod(SupportedHttpVerb.Post);
                 string requestBody = @"{
                         ""title"": ""Harry Potter and the Order of Phoenix"",
-                        ""publisher_id"": 1234";
+                        ""publisher_id"": 1234
+                    }";
```
Aniruddh25: do you need this change? Isnt line 3331 taking care of the closing bracket?

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -671,5 +671,57 @@ private static bool TryParseAndAssertOnDefaults(string json, out RuntimeConfig p
...
+
+            // Escape backslashes for JSON
+            string escapedPath = akvFilePath.Replace("\\", "\\\\");
+
+            string jsonConfig = $$"""
+            {
+              "$schema": "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch-alpha/dab.draft.schema.json",
+              "data-source": {
+                "database-type": "mssql",
+                "connection-string": "@AKV('DB_CONN')"
```
Aniruddh25: Should add a test with value of connection-string = `@env('Env_variable')`
and 
Env_variable=@akv('DB_CONN').

This should test the 2 pass algorithm you have.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -671,5 +671,57 @@ private static bool TryParseAndAssertOnDefaults(string json, out RuntimeConfig p
...
+              },
```
Aniruddh25, line 697: nit: I thought we look for `@akv` <- notice the lower case. NOT @AKV.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -671,5 +671,57 @@ private static bool TryParseAndAssertOnDefaults(string json, out RuntimeConfig p
...
+            string secretConnectionString = "Server=tcp:127.0.0.1,1433;Persist Security Info=False;Trusted_Connection=True;TrustServerCertificate=True;MultipleActiveResultSets=False;Connection Timeout=5;";
+            File.WriteAllText(akvFilePath, $"DB_CONN={secretConnectionString}\nAPI_KEY=abcd\n# Comment line should be ignored\n MALFORMEDLINE \n");
+
+            // Escape backslashes for JSON
+            string escapedPath = akvFilePath.Replace("\\", "\\\\");
+
+            string jsonConfig = $$"""
+            {
+              "$schema": "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch-alpha/dab.draft.schema.json",
+              "data-source": {
```
Aniruddh25, line 694: Should also have a test where DB_CONN in @akv('DB_CONN') has the value of @env('env'). 
This should not be resolving the env variable.

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: You're right, my IDE just complains about this, will remove.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: added test case.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: added test case.

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -66,13 +71,68 @@ public DeserializationVariableReplacementSettings(
...
                 ReplacementStrategies.Add(
                     new Regex(OUTER_AKV_PATTERN, RegexOptions.Compiled),
                     ReplaceAkvVariable);
             }
         }
 
+        // Checks if the endpoint is a path to a local .akv file.
+        private static bool IsLocalAkvFileEndpoint(string? endpoint)
+            => !string.IsNullOrWhiteSpace(endpoint)
+               && endpoint.EndsWith(".akv", StringComparison.OrdinalIgnoreCase)
```
RubenCerna2079: Is there a reason why we want to ignore the case? I don't think we want capital letters in the file such as `.AKV`, only `.akv`

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -723,5 +725,126 @@ public void TestAkvVariableReplacementFromLocalFile()
...
+        /// because replacement only runs once per original JSON token. Demonstrates that nested env patterns inside
+        /// AKV secret values are left intact.
+        /// </summary>
+        [TestMethod]
+        public void TestAkvSecretValueContainingEnvPatternIsNotEnvExpanded()
+        {
+            string akvFilePath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString() + ".akv");
+            // Valid MSSQL connection string which embeds an @env('env') pattern in the Database value.
+            // This pattern should NOT be expanded because replacement only runs once on the original JSON token (@akv('DBCONN')).
+            string secretValueWithEnvPattern = "Server=localhost;Database=@env('env');User Id=sa;Password=Pass@word1;";
```
Aniruddh25:
```suggestion
            string secretValueWithEnvPattern = "Server=localhost;Database=@env('env');User Id=sa;Password=XXXX;";
```
should not checkin a password even if it is for a sample. This will trigger S360 issues.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -723,5 +725,126 @@ public void TestAkvVariableReplacementFromLocalFile()
...
+            File.WriteAllText(akvFilePath, $"DBCONN={secretValueWithEnvPattern}\n");
+            string escapedPath = akvFilePath.Replace("\\", "\\\\");
+
+            // Set env variable to prove it would be different if expansion occurred.
+            Environment.SetEnvironmentVariable("env", "SHOULD_NOT_APPEAR");
```
Aniruddh25, line 745: Should assert that the string "SHOULD_NOT_APPEAR" does actually not appear in the actual string.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -723,5 +725,126 @@ public void TestAkvVariableReplacementFromLocalFile()
...
+
+        /// <summary>
+        /// Validates two-pass replacement where an env var resolves to an AKV pattern which then resolves to the secret value.
+        /// connection-string = @env('env_variable'), env_variable value = @akv('DBCONN'), AKV secret DBCONN holds the final connection string.
+        /// </summary>
+        [TestMethod]
+        public void TestEnvVariableResolvingToAkvPatternIsExpandedInSecondPass()
+        {
+            string akvFilePath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString() + ".akv");
+            string finalSecretValue = "Server=localhost;Database=Test;User Id=sa;Password=Pass@word1;";
```
Aniruddh25:
```suggestion
            string finalSecretValue = "Server=localhost;Database=Test;User Id=sa;Password=XXXX;";
```

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: Added the assert.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -627,7 +629,7 @@ private static void ClearEnvironmentVariablesFromDictionary(Dictionary<string, s
             { "DATABASE_NAME", "planet" },
             { "GRAPHQL_SCHEMA_PATH", "gql-schema.gql" },
             { "DATABASE_CONNECTION_STRING", "Data Source=<>;Initial Catalog=<>;User ID=<>;Password=<>;" },
-            { "DATABASE_CONNECTION_STRING_PGSQL", "Host=<>;Database=<>;username=<>;password=<>" }
+            { "DATABASE_CONNECTION_STRING_PGSQL", "Host=<>;Database=<>;username=<>;password<>" }
```
souvikghosh04: not sure why this was changed. should it be `password=<>`?

src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -195,6 +264,12 @@ private static SecretClient CreateSecretClient(AzureKeyVaultOptions options)
 
         private string? GetAkvVariable(string name)
         {
+            // If using local .akv file secrets, return from dictionary.
+            if (_akvFileSecrets is not null)
+            {
+                return _akvFileSecrets.TryGetValue(name, out string? value) ? value : null;
```
souvikghosh04, line 270: when returning null if secret doesn't exist, do you need to log this information somewhere?

aaronburtle: We do not, this will be captured in `ReplaceAkvVariable` if failure mode is set to throw, otherwise we return the original `@akv('<nonexistent secret name>')` as we do with env variable replacement as well.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
aaronburtle: typo, reverting, thanks

PR #2977, opened 2025-11-15
Added aspire sample

JerryNixon:
This is a simple sample that demonstrates how to use DAB in Aspire. 

I will add it to our docs/quick-starts when this is merged.

Please note there is no /src code change, just /samples.

### DIFF
--- a/.pipelines/azure-pipelines.yml
+++ b/.pipelines/azure-pipelines.yml
@@ -9,6 +9,18 @@ trigger:
     - gh-readonly-queue/main
     - release/*
 
+pr:
+  branches:
+    include:
+    - main
+    - release/*
+  paths:
+    exclude:
+    - samples/**
+    - docs/**
+    - '*.md'
+    - templates/**
+
 variables:
 #  DebugArguments: ' --filter "TestCategory!=ABC" --verbosity normal '
 #  ReleaseArguments: ' --filter "TestCategory!=ABC" --verbosity normal '
--- a/.pipelines/cosmos-pipelines.yml
+++ b/.pipelines/cosmos-pipelines.yml
@@ -13,6 +13,18 @@ trigger:
     - gh-readonly-queue/main
     - release/*
 
+pr:
+  branches:
+    include:
+    - main
+    - release/*
+  paths:
+    exclude:
+    - samples/**
+    - docs/**
+    - '*.md'
+    - templates/**
+
 strategy:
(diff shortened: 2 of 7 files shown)


JerryNixon:
We really need to add this to our tests in order to avoid requiring the full suite of integration tests when /samples is updated. In fact, there are many paths in this repo that should be set up to allow pull requests to skip integration tests. Please consider.

```yaml
on:
  pull_request:
    paths-ignore:
      - 'samples/**'
```

samples/aspire/apphost.cs
```diff
@@ -0,0 +1,71 @@
...
+#:package CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects@9.8.1-beta.420
+#:package CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder@9.8.1-beta.420
+#:package CommunityToolkit.Aspire.Hosting.McpInspector@9.8.0
+
+var builder = DistributedApplication.CreateBuilder(args);
+
+var options = new
+{
+    SqlScript = "database.sql",
+    SqlPassword = "P@ssw0rd!",
```
Aniruddh25: We cannot checkin Passwords even if they are in samples.

Aniruddh25 (approved):

samples/aspire/apphost.cs
```diff
@@ -0,0 +1,70 @@
+#:sdk Aspire.AppHost.Sdk@13.0.0
```
aaronburtle, line 4: nit: we normally include

for any .cs file

samples/aspire/global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "10.0.100"
```
aaronburtle, line 3: Aren't we normally in version 8?

samples/aspire/apphost.cs
Aniruddh25: @Copilot could you address this feedback?

samples/aspire/global.json
Aniruddh25: Thats true, we should upgrade dab to .NET 10 as well since it released on Nov 11, and is a LTS. Since, this is a sample, I'm resolving this comment and create a separate task to upgrade dab.

Aniruddh25 (approved):

PR #2983, opened 2025-11-17
[MCP] Added support for `--mcp-stdio` flag to `dab start`

anushakolan:
## Why make this change?

- Add MCP stdio support to Data API Builder and wire it through both the engine and CLI so DAB can be used as a Model Context Protocol (MCP) server.
- Ensures MCP sessions can run under a specific DAB authorization role, making it possible to test and use MCP tools with permissions from `dab-config.json`.

## What is this change?

Service entrypoint
 - Detects `--mcp-stdio` early, configures stdin/stdout encodings, and redirects all non‑MCP output to STDERR to keep STDOUT clean for MCP JSON.
 - Parses an optional `role:<name>` argument (e.g. role:anonymous, role:authenticated) and injects it into configuration as `MCP:Role`, defaulting to `anonymous` when omitted.
 - In MCP stdio mode, forces `Runtime:Host:Authentication:Provider = "Simulator"` via in‑memory configuration so the requested role is always available during MCP sessions.
 - Starts the full ASP.NET Core host, registers all MCP tools from DI, and runs the MCP stdio loop instead of the normal HTTP `host.Run(`).

CLI Integration
 - Adds `--mcp-stdio` to `dab start` to launch the engine in MCP stdio mode.
 - Adds an optional positional `role` argument (e.g. `role:anonymous`) captured as `StartOptions.McpRole`.
 - Keeps existing behavior for non‑MCP `dab start` unchanged.

Note
- `ExecuteEntityTool` now looks for MCP tool inputs under arguments (the standard MCP field) and falls back to the legacy parameters property only if arguments is missing. This aligns our server with how current MCP clients (like VS Code) actually send tool arguments, and preserves backward compatibility for any older clients that still use parameters.

## How was this tested?

Integration-like manual testing via MCP clients against:
 - Engine-based MCP server: `dotnet Azure.DataApiBuilder.Service.dll --mcp-stdio role:authenticated`.
 - CLI-based MCP server: `dab start --mcp-stdio role:authenticated`.

Manual verification of all MCP tools:
 - `describe_entities` shows correct entities and effective permissions for the active role.
 - `read_records`, `create_record`, `update_record`, `delete_record`, `execute_entity` succeed when the role has the appropriate permissions.

## Sample Request(s)

1. MCP server via CLI (dab)

`
{
  "mcpServers": {
    "dab-with-exe": {
      "command": "C:\\DAB\\data-api-builder\\out\\publish\\Debug\\net8.0\\win-x64\\dab\\Microsoft.DataApiBuilder.exe",
      "args": ["start", "--mcp-stdio", "role:authenticated", "--config",
				"C:\\DAB\\data-api-builder\\dab-config.json"],
      "env": {
        "DAB_ENVIRONMENT": "Development"
      }
  }
}
`
2. MCP server via engine DLL

`
{
  "mcpServers": {
    "dab": {
			"command": "dotnet",
			"args": [
				"C:\\DAB\\data-api-builder\\out\\publish\\Debug\\net8.0\\win-x64\\dab\\Azure.DataApiBuilder.Service.dll",
				"--mcp-stdio",
				"role:authenticated",
				"--config",
				"C:\\DAB\\data-api-builder\\dab-config.json"
			],
			"type": "stdio"
		}
  }
}
`


src/Cli/ConfigGenerator.cs
```diff
@@ -2359,6 +2359,17 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
 
+            // If MCP stdio was requested, append the stdio-specific switches.
+            if (options.McpStdio)
+            {
+                string effectiveRole = string.IsNullOrWhiteSpace(options.McpRole)
+                    ? "anonymous"
+                    : options.McpRole;
+
+                args.Add("--mcp-stdio");
+                args.Add(effectiveRole);
```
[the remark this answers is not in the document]
anushakolan: No, this will fail because the `effectiveRole` already comes with "role:" attached.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+            using StreamReader reader = new(stdin, utf8NoBom);
+            using StreamWriter writer = new(stdout, utf8NoBom) { AutoFlush = true };
+
+            // Redirect Console.Out to use our writer
+            Console.SetOut(writer);
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                string? line = await reader.ReadLineAsync();
+                if (string.IsNullOrWhiteSpace(line))
```
aaronburtle, line 61: Do we need to handle multiline or framed messages? #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+        /// </param>
+        /// <remarks>
+        /// This method constructs and writes the MCP "initialize" response to STDOUT. It uses the protocol version defined by <c>PROTOCOL_VERSION</c>
+        /// and includes supported capabilities and server information. No notifications are sent here; the server waits for the client to send
+        /// "notifications/initialized" before sending any notifications.
+        /// </remarks>
+        private static void HandleInitialize(JsonElement? id)
+        {
+            // Extract the actual id value from the request
+            int requestId = id.HasValue ? id.Value.GetInt32() : 0;
```
aaronburtle: In `WriteResult` we are using the `GetIdValue` helper to parse the id, since it can be a string, or float, or 64 bit int, but here we seem to assume it will be a 32 bit int, should we be using the helper here too?


```return id.ValueKind switch
            {
                JsonValueKind.String => id.GetString(),
                JsonValueKind.Number => id.TryGetInt64(out long l) ? l :
                                        id.TryGetDouble(out double d) ? d : null,
                _ => null
            };```
 #Resolved

File: src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs [hunk 4]
```diff
@@ -0,0 +1,458 @@
...
+    /// - Reads JSON-RPC requests (initialize, listTools, callTool) from STDIN
+    /// - Writes ONLY MCP JSON responses to STDOUT
+    /// - Writes diagnostics to STDERR (so STDOUT remains “pure MCP”)
+    /// </summary>
+    public class McpStdioServer : IMcpStdioServer
+    {
+        private readonly McpToolRegistry _toolRegistry;
+        private readonly IServiceProvider _serviceProvider;
+
+        private const string PROTOCOL_VERSION = "2025-06-18";
```
aaronburtle: Will we have to manually change this version every time it updated? We have `AddMcpServer` in the `McpServerConfguration` could this cause a divergence in version? Maybe we can centralize this somehow or allow other versions or overrides? #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+                    {
+                        WriteError(id, -32600, "Invalid Request");
+                        continue;
+                    }
+
+                    string method = methodEl.GetString() ?? string.Empty;
+
+                    try
+                    {
+                        switch (method)
```
aaronburtle, line 110: In `HandleInitialize` we include resources and prompts in the capabilities, but we don't handle them here in `RunAsync`, is that OK, or could this lead to a method not found error? Should we have stubs for those cases that return a message or something empty?

                    ```capabilities = new
                    {
                        tools = new { listChanged = true },
                        resources = new { subscribe = true, listChanged = true },
                        prompts = new { listChanged = true },
                        logging = new { }
                    },``` #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+                {
+                    continue;
+                }
+
+                JsonDocument doc;
+                try
+                {
+                    doc = JsonDocument.Parse(line);
+                }
+                catch (Exception)
```
aaronburtle: Is just catching generic exceptions OK here? Should we have more granular exception handling here? Same question for the other generic exceptions. #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+                    {
+                        name = "Data API Builder",
+                        version = "1.0.0"
+                    }
+                }
+            };
+
+            string json = JsonSerializer.Serialize(response);
+            Console.Out.WriteLine(json);
+            Console.Out.Flush();
```
aaronburtle: with `using StreamWriter writer = new(stdout, utf8NoBom) { AutoFlush = true };` won't we automatically flush here? Same comment for the other `Console.Out.Flush()` calls. #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+        /// </summary>
+        /// <param name="id">
+        /// The request identifier extracted from the incoming JSON-RPC request. Used to correlate the response with the request.
+        /// </param>
+        private void HandleListTools(JsonElement? id)
+        {
+            List<object> toolsWire = new();
+            int count = 0;
+
+            foreach (Tool tool in _toolRegistry.GetAllTools())
```
aaronburtle, line 204: Just to clarify, we won't ever be registering tools after requests are being handled right? Otherwise, we may need locking here #Resolved

src/Service/Program.cs
```diff
@@ -33,27 +37,87 @@ public class Program
...
+                string? roleArg = Array.Find(args, a => a != null && a.StartsWith("role:", StringComparison.OrdinalIgnoreCase));
+                if (!string.IsNullOrEmpty(roleArg))
+                {
+                    string roleValue = roleArg.Substring(roleArg.IndexOf(':') + 1);
+                    if (!string.IsNullOrWhiteSpace(roleValue))
+                    {
+                        mcpRole = roleValue;
+                    }
+                }
+            }
```
souvikghosh04: we should consider moving this to a utility since this is the `Main()` function and we should keep it modular.

src/Service/Program.cs
```diff
@@ -33,27 +37,87 @@ public class Program
...
+                    using IServiceScope scope = scopeFactory.CreateScope();
+                    IHostApplicationLifetime lifetime = scope.ServiceProvider.GetRequiredService<IHostApplicationLifetime>();
+                    Mcp.Core.IMcpStdioServer stdio = scope.ServiceProvider.GetRequiredService<Mcp.Core.IMcpStdioServer>();
+
+                    // Run the stdio loop until cancellation (Ctrl+C / process end)
+                    stdio.RunAsync(lifetime.ApplicationStopping).GetAwaiter().GetResult();
+
+                    host.StopAsync().GetAwaiter().GetResult();
+                    return true;
+                }
```
souvikghosh04, line 67: can we simplify this and make it modular inside a Utility similar to the logic we have in `Main()`?

src/Service/Program.cs
```diff
@@ -72,17 +136,48 @@ public static bool StartEngine(string[] args)
...
+                string? roleArg = Array.Find(args, a => a != null && a.StartsWith("role:", StringComparison.OrdinalIgnoreCase));
+                if (!string.IsNullOrEmpty(roleArg))
+                {
+                    string roleValue = roleArg[(roleArg.IndexOf(':') + 1)..];
+                    if (!string.IsNullOrWhiteSpace(roleValue))
+                    {
+                        mcpRole = roleValue;
+                    }
+                }
+            }
```
souvikghosh04: same recoemmendation as before. we have similar role parsing and conditional checks used in multiple places. please consider moving these to utility and reuse them to avoid bugs and ease of maintenance.

src/Service/Program.cs
```diff
@@ -33,27 +37,87 @@ public class Program
...
+                // MCP requires STDOUT to contain only protocol JSON; send all other text to STDERR
+                Console.SetOut(Console.Error);
+
+                // If caller provided an optional role token like `role:authenticated`, capture it and
+                // force the runtime to use the Simulator authentication provider for this session.
+                // This makes it easy to run MCP stdio sessions with a preconfigured permissions role.
+                string? roleArg = Array.Find(args, a => a != null && a.StartsWith("role:", StringComparison.OrdinalIgnoreCase));
+                if (!string.IsNullOrEmpty(roleArg))
+                {
+                    string roleValue = roleArg.Substring(roleArg.IndexOf(':') + 1);
```
souvikghosh04: in line 150, you have used parsing logic as `string roleValue = roleArg[(roleArg.IndexOf(':') + 1)..];` but here it is different. it may work most of the time as intended but can behave differently in some scenarios. please have a common parsing logic and reuse it where needed.

src/Service/Program.cs
```diff
@@ -33,27 +37,87 @@ public class Program
...
+                Console.OutputEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+                Console.InputEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+
+                // MCP requires STDOUT to contain only protocol JSON; send all other text to STDERR
+                Console.SetOut(Console.Error);
+
+                // If caller provided an optional role token like `role:authenticated`, capture it and
+                // force the runtime to use the Simulator authentication provider for this session.
+                // This makes it easy to run MCP stdio sessions with a preconfigured permissions role.
+                string? roleArg = Array.Find(args, a => a != null && a.StartsWith("role:", StringComparison.OrdinalIgnoreCase));
```
souvikghosh04: nit: can we set and use `"role:"` as a constant string value as we have in multiple places?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
souvikghosh04: although it won't break anything but versioning is upon us how we maintain and moving it to a central configuration is receommended.

src/Service/Program.cs
```diff
@@ -33,27 +37,87 @@ public class Program
...
+                    // host.Start() boots the host without blocking on Kestrel,
+                    // allowing the process to handle MCP requests over stdio instead
+                    // of serving HTTP traffic via host.Run().
+                    host.Start();
+
+                    Mcp.Core.McpToolRegistry registry = host.Services.GetRequiredService<Mcp.Core.McpToolRegistry>();
+                    IEnumerable<Mcp.Model.IMcpTool> tools = host.Services.GetServices<Mcp.Model.IMcpTool>();
+                    foreach (Mcp.Model.IMcpTool tool in tools)
+                    {
+                        Tool metadata = tool.GetToolMetadata();
```
souvikghosh04: it seems `Tool metadata` is initialized but not used.

src/Cli/ConfigGenerator.cs
```diff
@@ -2359,6 +2359,17 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
+            }
```
souvikghosh04, line 2371: nit: please centralize this, similar to the reviews in `Program.cs`

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+                string? stdioRole = configuration?.GetValue<string>("MCP:Role");
+                if (!string.IsNullOrWhiteSpace(stdioRole))
+                {
+                    IServiceScopeFactory scopeFactory = _serviceProvider.GetRequiredService<IServiceScopeFactory>();
+                    IServiceScope scope = scopeFactory.CreateScope();
+                    IServiceProvider scopedProvider = scope.ServiceProvider;
+
+                    // Create a default HttpContext and set the client role header
+                    DefaultHttpContext httpContext = new();
+                    httpContext.Request.Headers["X-MS-API-ROLE"] = stdioRole;
```
aaronburtle, line 292: nit: should we use a defined constant instead of hardcoding the string like in the auth resolver?

`public const string CLIENT_ROLE_HEADER = "X-MS-API-ROLE";` #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+
+                // Execute the tool.
+                // If a MCP stdio role override is set in the environment, create
+                // a request HttpContext with the X-MS-API-ROLE header so tools and authorization
+                // helpers that read IHttpContextAccessor will see the role. We also ensure the
+                // Simulator authentication handler can authenticate the user by flowing the
+                // Authorization header commonly used in tests/simulator scenarios.
+                CallToolResult callResult;
+                IConfiguration? configuration = _serviceProvider.GetService<IConfiguration>();
+                string? stdioRole = configuration?.GetValue<string>("MCP:Role");
```
aaronburtle, line 283: Should we validate that whatever this role is defined as is valid? Could it be `null`? #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+                if (!string.IsNullOrWhiteSpace(stdioRole))
+                {
+                    IServiceScopeFactory scopeFactory = _serviceProvider.GetRequiredService<IServiceScopeFactory>();
+                    IServiceScope scope = scopeFactory.CreateScope();
```
aaronburtle: I think we need to use a `using` key word here to ensure proper disposal of the IServiceScope.

`using IServiceScope scope = scopeFactory.CreateScope();` #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+                    if (httpContextAccessor is not null)
+                    {
+                        httpContextAccessor.HttpContext = httpContext;
+                    }
+
+                    // Execute the tool with the scoped service provider so any scoped services resolve correctly.
+                    callResult = await tool.ExecuteAsync(argsDoc, scopedProvider, ct);
+
+                    // Clear the accessor's HttpContext to avoid leaking across calls
+                    if (httpContextAccessor is not null)
```
aaronburtle: Because this is an `async` function, should we use try/finally here instead of just clearing, in case there are other async operations still running? #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+        /// Safely converts an object to its string representation, preferring JSON serialization for readability.
+        /// </summary>
+        /// <param name="obj">The object to convert to a string.</param>
+        /// <returns>A string representation of the object.</returns>
+        private static string SafeToString(object obj)
+        {
+            try
+            {
+                // Try JSON first for readability
+                return JsonSerializer.Serialize(obj);
```
aaronburtle: Do we need to worry about size caps for very large json objets? #Resolved

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,458 @@
...
+            using Stream stdout = Console.OpenStandardOutput();
+            using StreamReader reader = new(stdin, utf8NoBom);
+            using StreamWriter writer = new(stdout, utf8NoBom) { AutoFlush = true };
+
+            // Redirect Console.Out to use our writer
+            Console.SetOut(writer);
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                string? line = await reader.ReadLineAsync();
```
aaronburtle: Should we add some limit to the size of line we will read? What if the line is arbitrarily large? #Resolved

anushakolan: Added a size limit.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: In this case we are specifically parsing the `JsonDocument`, so there is only parsing error possible, so I used a generic exception.  I am not sure what other exceptions are possible here that needs to be specifically handled. Let me know if you think some other exceptions that are possible here, I will add them accordingly. I will take a look at other exceptions as well.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: We don't support prompts or resources now, so we are not handling those requests in `RunAsync()`, this won't give a problem instead it will send a generic method not found reply. I will remove these from the capabilities also.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: As per today, Tools are expected to be registered during application startup only.
If this ever changes and tools can be added/removed at runtime while requests are being handled, we may need to introduce locking here or have the registry return a thread-safe snapshot.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: We are validating it in the next line if it is null or whitespace.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Since we are using it only once throughout the code, I did not find the need to create a new constant.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Added size cap.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Updated the code to use a centralized source of truth. If a ProtocolVersion is configured via appsettings.json. it gets automatically picked up if not the default version will be used.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: This might be a future case if needed, but the vs code or the servers which we are using send one JSON-RPC object per line.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
aaronburtle: That makes sense, the only spot that stands out to me where another exception might bubble up is inside the try/catch where we call `await HandleCallToolAsync(id, root, cancellationToken);` since we might get a tool specific exception there, or mcp exception of some kind.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
aaronburtle: Gotcha on the `null`, but what about if the role is not valid in any entity? For example if someone has a typo and gives a role of `admn` instead of `admin` do we want to fail early and inform them that the role isnt valid for any entities?

aaronburtle (approved):
Looks good!

souvikghosh04 (changes requested):
Added some comments to have common role parsing and retrieval logic along with some refactoring/modularization.

src/Service/Program.cs
```diff
@@ -33,27 +37,84 @@ public class Program
...
+                        registry.RegisterTool(tool);
+                    }
+
+                    // Resolve and run the MCP stdio server from DI
+                    IServiceScopeFactory scopeFactory = host.Services.GetRequiredService<IServiceScopeFactory>();
+                    using IServiceScope scope = scopeFactory.CreateScope();
+                    IHostApplicationLifetime lifetime = scope.ServiceProvider.GetRequiredService<IHostApplicationLifetime>();
+                    Mcp.Core.IMcpStdioServer stdio = scope.ServiceProvider.GetRequiredService<Mcp.Core.IMcpStdioServer>();
+
+                    // Run the stdio loop until cancellation (Ctrl+C / process end)
```
souvikghosh04: are we saying we will have cancellation triggered when Ctrl + C is pressed or there are other ways we should allow cancellation as well? for e.g. cancelling out an unresponsive request? please also check we have cancellation support for stdio. we can probably test by retrieving a very large data from an MCP tool and see if cancellation works. this could be a real scenario where a user has a heavy operation triggerred through the MCP tool and a cancellation needs to done.

src/Cli/ConfigGenerator.cs
anushakolan: This logic is only used once to add the arguments. In ``Program.cs` there’s a lot of other code that could benefit from helper functions, centralizing this particular piece wouldn’t add much value.

src/Service/Program.cs
anushakolan: ### Process‑level cancellation (Ctrl+C / host shutdown)

1. In `RunMcpStdioHost` we run the `stdio` server with the host’s shutdown token:
 `stdio.RunAsync(lifetime.ApplicationStopping).GetAwaiter().GetResult();`
2. `lifetime.ApplicationStopping` is triggered when the DAB host is stopping (e.g., Ctrl+C in the DAB process, SIGTERM, or host.StopAsync()).
3. When that happens, the stdio loop observes the token, exits cleanly, and then we call `host.StopAsync`.
4. Ctrl+C (and other host shutdown signals) will cancel the stdio server and any in‑flight MCP operations as part of process shutdown.

### Per‑request cancellation for heavy / unresponsive MCP calls
1. Separately from host shutdown, I validated request‑level cancellation via the MCP integration:
    - I added a long‑running DAB tool `LongRunningTodoQuery` to `dab-config.json`.
    - invoked it through the DAB MCP server from VS Code Copilot.
    - While it was running, you used the Copilot chat cancel control (the “stop” on the chat/tool) and the call was aborted with: “The user cancelled the tool call.”
2. This shows that:
   - The stdio server and tools honor client‑initiated cancellation for an individual MCP request.
   - The server stays up; only that specific long‑running operation is stopped.

anushakolan:
> Added some comments to have common role parsing and retrieval logic along with some refactoring/modularization.

Addressed all comments.

aaronburtle (approved):

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,486 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using ModelContextProtocol.Protocol;
+
+namespace Azure.DataApiBuilder.Mcp.Core
+{
+    /// <summary>
+    /// MCP stdio server:
+    /// - Reads JSON-RPC requests (initialize, listTools, callTool) from STDIN
+    /// - Writes ONLY MCP JSON responses to STDOUT
+    /// - Writes diagnostics to STDERR (so STDOUT remains “pure MCP”)
```
Aniruddh25, line 19: Will STDOUT continue to contain REST/GraphQL diagnostics?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,486 @@
...
+        }
+
+        /// <summary>
+        /// Runs the MCP stdio server loop, reading JSON-RPC requests from STDIN and writing MCP JSON responses to STDOUT.
+        /// </summary>
+        /// <param name="cancellationToken">Token to signal cancellation of the server loop.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public async Task RunAsync(CancellationToken cancellationToken)
+        {
+            Console.Error.WriteLine("[MCP DEBUG] MCP stdio server started.");
```
Aniruddh25, line 46: Where does this WriteLine go to since we are setting out to our writer much later?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,486 @@
...
+            {
+                string? line = await reader.ReadLineAsync(cancellationToken);
+                if (string.IsNullOrWhiteSpace(line))
+                {
+                    continue;
+                }
+
+                if (line.Length > MAX_LINE_LENGTH)
+                {
+                    WriteError(id: null, code: -32600, message: "Request too large");
```
Aniruddh25, line 68: This says WriteError but still is writing to stdout in the function. Why is that? Shouldn't we write to STDERR?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,486 @@
...
+                {
+                    protocolVersion = _protocolVersion,
+                    capabilities = new
+                    {
+                        tools = new { listChanged = true },
+                        logging = new { }
+                    },
+                    serverInfo = new
+                    {
+                        name = "Data API Builder",
```
Aniruddh25, line 179: name should be `SQL MCP Server`

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
Aniruddh25, line 68: The error codes should be defined as constants in a common place, not hard coded as 326XX. How did we come up with these numbers anyway?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,486 @@
...
+                        version = "1.0.0"
+                    }
+                }
+            };
+
+            string json = JsonSerializer.Serialize(response);
+            Console.Out.WriteLine(json);
```
Aniruddh25, line 186: Why dont we use WriteResult here instead of simply Console.Out.WriteLine?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,486 @@
...
+                return;
+            }
+
+            JsonDocument? argsDoc = null;
+            try
+            {
+                if (@params.TryGetProperty("arguments", out JsonElement argsEl) && argsEl.ValueKind == JsonValueKind.Object)
+                {
+                    string rawArgs = argsEl.GetRawText();
+                    Console.Error.WriteLine($"[MCP DEBUG] callTool → tool: {toolName}, args: {rawArgs}");
```
Aniruddh25, line 267: Why is this an error?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -0,0 +1,486 @@
...
+                count++;
+                toolsWire.Add(new
+                {
+                    name = tool.Name,
+                    description = tool.Description,
+                    inputSchema = tool.InputSchema
+                });
+            }
+
+            WriteResult(id, new { tools = toolsWire });
```
Aniruddh25, line 215: Why isnt the stdioRole required to be added to the httpcontext when listing the tools? Is that information not controlled by our authZ logic currently?

src/Service/Program.cs
```diff
@@ -33,27 +33,41 @@ public class Program
...
             {
-                CreateHostBuilder(args).Build().Run();
+                IHost host = CreateHostBuilder(args, runMcpStdio, mcpRole).Build();
+
+                if (runMcpStdio)
+                {
+                    return McpStdioHelper.RunMcpStdioHost(host);
+                }
+
+                // Normal web mode
```
Aniruddh25, line 69: The enhancement requirement explicitly says that 

`REST and GraphQL endpoints are unaffected
http behavior of the MCP endpoint is unaffected`

However, the code here seems to not honor this. in stdio mode, we are restricted to stdio. Is that really the expectation?

src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// </summary>
+    internal static class McpStdioHelper
+    {
+        /// <summary>
+        /// Determines if MCP stdio mode should be run based on command line arguments.
+        /// </summary>
+        /// <param name="args"> The command line arguments.</param>
+        /// <param name="mcpRole"> The role for MCP stdio mode, if specified.</param>
+        /// <returns></returns>
+        public static bool ShouldRunMcpStdio(string[] args, out string? mcpRole)
```
Aniruddh25, line 20: You can annotate mcpRole with [NotNullWhen(true)] to help with static analysis,
but to be 100% accurate, also ensure assignment to mcpRole on all paths returning true.

src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            bool runMcpStdio = Array.Exists(
+                args,
+                a => string.Equals(a, "--mcp-stdio", StringComparison.OrdinalIgnoreCase));
+
+            if (!runMcpStdio)
+            {
+                return false;
+            }
+
```
Aniruddh25, line 32: set default mcpRole = anonymous when runMcpStdio is true.

src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+            host.Start();
+
+            Mcp.Core.McpToolRegistry registry =
+                host.Services.GetRequiredService<Mcp.Core.McpToolRegistry>();
+            IEnumerable<Mcp.Model.IMcpTool> tools =
+                host.Services.GetServices<Mcp.Model.IMcpTool>();
+
+            foreach (Mcp.Model.IMcpTool tool in tools)
+            {
+                _ = tool.GetToolMetadata();
```
Aniruddh25, line 79: If we are not using the metadata, why do we need to get it?

src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+                registry.RegisterTool(tool);
+            }
+
+            IServiceScopeFactory scopeFactory =
+                host.Services.GetRequiredService<IServiceScopeFactory>();
+            using IServiceScope scope = scopeFactory.CreateScope();
+            IHostApplicationLifetime lifetime =
+                scope.ServiceProvider.GetRequiredService<IHostApplicationLifetime>();
```
Aniruddh25, line 87: Why do we need to use a scope to get the required service? Can we not obtain it from `host.Services.GetRequiredService`?

Just trying to understand the reasoning.

src/Service/Program.cs
anushakolan: No, based on both the requirement and the current implementation, restricted to `stdio` is not the intended behavior, and the code actually does keep HTTP endpoints alive even in `stdio` mode.


1.`StartEngine` always builds the normal ASP.NET Core `IHost` with `ConfigureWebHostDefault` and `Startup` (REST/GraphQL/MCP HTTP are all wired as usual).
2. In stdio mode, `StartEngine` calls `McpStdioHelper.RunMcpStdioHost(host)` instead of `host.Run()`.
3. `RunMcpStdioHost` calls `host.Start()` first – this starts Kestrel and all HTTP endpoints exactly as in web mode, then runs the MCP stdio loop, and finally `host.StopAsync()` when stdio exits.
4. `Startup` still adds controllers, GraphQL, and `AddDabMcpServer`; the only stdio-specific change is forcing the Simulator auth provider and injecting MCP:Role.
5. So, even with `--mcp-stdio`, the process is not “stdio-only”: REST, GraphQL, and the HTTP MCP endpoint are still reachable on the configured URLs while the stdio MCP server is running. The only “restriction” in stdio mode is that logging is routed to STDERR and authentication is forced to Simulator with the configured MCP role.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: When DAB runs as an MCP stdio server, the MCP client expects that everything it reads from standard output is valid MCP JSON. If we mixed human‑readable debug messages into that same stream, the client could no longer parse it reliably.

To avoid that, we send only structured MCP responses on standard output, and we send all other logs (debug traces, “tool not found”, parse problems, etc.) to standard error. This way:

MCP clients get a clean, machine‑readable JSON stream and we can still see what went wrong by looking at standard error, without breaking the protocol.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Listing tools is just capability discovery. It answers: “What kinds of operations does this MCP server support in general?” That’s a static list (describe_entities, read_records, etc.) and is not currently filtered by role. Everyone sees the same menu of tools.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: When the server is started in MCP stdio mode, only MCP protocol messages should appear on standard output.
All REST and GraphQL logs, diagnostics, and error messages should appear on standard error instead, so they don’t mix with the MCP messages.
When the server is started normally (without MCP stdio mode), logs and diagnostics should appear on standard output as they did before.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: That line prints a message to the program’s error stream, not to the main output stream that carries MCP JSON.

Changing the program’s main output stream later in the method does not affect the error stream at all, so this message (and the other debug messages) always go to error output only and never mix with the MCP responses.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: protocol‑level errors go to the normal output stream (so the client can parse them); debug/diagnostic messages go to the error stream. The helper’s name makes it sound like “write to stderr”, but its job is actually “write a JSON error response to the MCP protocol channel”.

anushakolan: Those numbers aren’t arbitrary; they come from the JSON‑RPC 2.0 specification.

In JSON‑RPC 2.0, a handful of error codes are “well‑known” and reserved for specific meanings:

-32700  - “Parse error” (invalid JSON received)
-32600  - “Invalid Request” (bad JSON‑RPC envelope, not a valid request object)
-32601  - “Method not found”
-32602  - “Invalid params”
-32603  - “Internal error”
The server is using those spec‑defined values so that any JSON‑RPC‑aware client can recognize the error types without having to know anything specific about this project.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: This is an inconsistency, fixed it.

src/Service/Program.cs
Aniruddh25: thanks for the confirmation!

Issue #2948, opened 2025-11-04
Add mcp option at entity level


souvikghosh04:
To support custom tools, we need to have mcp option property to enable disable them at entity level. This is a pre-requisite to custom tools.

PR #2989, opened 2025-11-21
Add entity-level MCP configuration support

souvikghosh04:
## Why make this change?

This change allows entity-level MCP configuration to control which entities participate in MCP runtime tools, providing granular control over DML operations and custom tool exposure.

- Closes on #2948 

## What is this change?

This change introduces an optional mcp property at the entity level that controls participation in MCP's runtime tools. This is a prerequisite for custom tools support.

The MCP property supports two formats:

- **Boolean shorthand**: `"mcp": true` or `"mcp": false`
- **Object format**: `{"dml-tools": boolean, "custom-tool": boolean}`

Property Behavior:

1. Boolean Shorthand (`"mcp": true/false`)

- `"mcp": true`: Enables DML tools only; custom tools remain disabled.
- `"mcp": false`: Disables all MCP functionality for the entity.

2. Object Format `("mcp": { ... })`

- `{ "dml-tools": true, "custom-tool": true }`: Enables both (valid only for stored procedures).
- `{ "dml-tools": true, "custom-tool": false }`: DML only.
- `{ "dml-tools": false, "custom-tool": true }`: Custom tool only (stored procedures).
- `{ "dml-tools": false, "custom-tool": false }`: Fully disabled.

Single-property cases:

- `{"dml-tools": true}`: Enables DML only; auto-serializes to `"mcp": true`.
- `{"custom-tool": true}`: Enables custom tool only; serializes as given.

3. No MCP Configuration in Entity (default)
- `dml-tools` will still be enabled by default and no other change is behavior

## How was this tested?

- [x] Unit Tests
- [x] Integrations Tests
- [x] CLI Command Testing

Sample CLI commands:

Add table with DML tools enabled

`dab add Book --source books --permissions "anonymous:*" --mcp.dml-tools true`

Add stored procedure with custom tool enabled

`dab add GetBookById --source dbo.get_book_by_id --source.type stored-procedure --permissions "anonymous:execute" --mcp.custom-tool true`

Add stored procedure with both properties

`dab add UpdateBook --source dbo.update_book --source.type stored-procedure --permissions "anonymous:execute" --mcp.custom-tool true --mcp.dml-tools false`

### DIFF
--- a/src/Service.Tests/ModuleInitializer.cs
+++ b/src/Service.Tests/ModuleInitializer.cs
@@ -69,6 +69,10 @@ static class ModuleInitializer
         VerifierSettings.IgnoreMember<Entity>(entity => entity.IsLinkingEntity);
         // Ignore the UserProvidedTtlOptions. They aren't serialized to our config file, enforced by EntityCacheOptionsConverter.
         VerifierSettings.IgnoreMember<EntityCacheOptions>(cacheOptions => cacheOptions.UserProvidedTtlOptions);
+        // Ignore the UserProvidedCustomToolEnabled. They aren't serialized to our config file, enforced by EntityMcpOptionsConverterFactory.
+        VerifierSettings.IgnoreMember<EntityMcpOptions>(mcpOptions => mcpOptions.UserProvidedCustomToolEnabled);
+        // Ignore the UserProvidedDmlToolsEnabled. They aren't serialized to our config file, enforced by EntityMcpOptionsConverterFactory.
+        VerifierSettings.IgnoreMember<EntityMcpOptions>(mcpOptions => mcpOptions.UserProvidedDmlToolsEnabled);
         // Ignore the CosmosDataSourceUsed as that's unimportant from a test standpoint.
         VerifierSettings.IgnoreMember<RuntimeConfig>(config => config.CosmosDataSourceUsed);
         // Ignore the IsRequestBodyStrict as that's unimportant from a test standpoint.


config-generators/config-generator.ps1
```diff
@@ -70,4 +70,18 @@ foreach($databaseType in $databaseTypes){
         $commandToExecute = "dotnet " + $pathToDabDLL + " " + $command;
         Invoke-Expression $commandToExecute;
     }
+
+    # Post-process MsSql and DwSql configs to fix stored procedure GraphQL operations
+    # The CLI currently ignores --graphql.operation parameter for stored procedures,
+    # defaulting them to 'mutation'. We manually fix specific procedures that should be 'query'.
```
Aniruddh25: This sounds like a bug - when CLI ignore --graphql.operation. Is there an issue created for this? Once fixed, this post-processing step should not be required.

schemas/dab.draft.schema.json
```diff
@@ -875,18 +899,43 @@
...
                     }
                   ]
                 }
               ]
             },
+            "mcp": {
+              "oneOf": [
+                {
+                  "type": "boolean",
+                  "description": "Enable/disable DML tools. When true, all DML tools are enabled. When false, all DML tools are disabled. For stored procedures only: true also enables custom-tool as false by default."
```
Aniruddh25: `For stored procedures only: true also enables custom-tool as false by default.` => this isnt clear. Does it mean when mcp = true, custom-tool is false ? If yes, the consider rephrasing "true also enables custom-tool as false" -> "custom-tool stays disabled by default."

schemas/dab.draft.schema.json
```diff
@@ -747,7 +747,7 @@
                   "properties": {
                     "type": {
                       "type": "string",
-                      "enum": ["table", "view", "stored-procedure"],
+                      "enum": [ "table", "view", "stored-procedure" ],
```
Aniruddh25: `"description": "Entities that will be exposed via REST and/or GraphQL",`

      Should mention MCP too in the description above.

src/Cli.Tests/Snapshots/UpdateEntityTests.TestConversionOfSourceObject_036a859f50ce167c.verified.txt
```diff
@@ -27,18 +27,12 @@
       MyEntity: {
         Source: {
           Object: s001.book,
-          Type: View
+          Type: View,
+          KeyFields: [
```
Aniruddh25: this seems to be a regression. KeyFields was the old way, Fields is the new way to provide primary key. Same for all snapshot files.

src/Cli/ConfigGenerator.cs
```diff
@@ -1583,20 +1605,32 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
 
             // Validations to ensure that REST methods and GraphQL operations can be configured only
             // for stored procedures
-            if (options.GraphQLOperationForStoredProcedure is not null &&
-                !(isCurrentEntityStoredProcedure || doOptionsRepresentStoredProcedure))
+            if (options.GraphQLOperationForStoredProcedure is not null && !isCurrentEntityStoredProcedure)
```
Aniruddh25: Why remove the check for doOptionsRepresentStoredProcedure? We need that check when the new type is stored proc but old was table/view

src/Cli/ConfigGenerator.cs
```diff
@@ -1583,20 +1605,32 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
             {
                 _logger.LogError("--graphql.operation can be configured only for stored procedures.");
                 return false;
             }
 
             if ((options.RestMethodsForStoredProcedure is not null && options.RestMethodsForStoredProcedure.Any())
-                && !(isCurrentEntityStoredProcedure || doOptionsRepresentStoredProcedure))
+                && !isCurrentEntityStoredProcedure)
```
Aniruddh25: same why remove the doOptionsRepresentStoredProcedure check here?

src/Cli/ConfigGenerator.cs
```diff
@@ -1583,20 +1605,32 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
             {
                 _logger.LogError("--rest.methods can be configured only for stored procedures.");
                 return false;
             }
 
+            // Validate MCP custom-tool is only for stored procedures
+            if (options.McpCustomToolEnabled is not null and not CliBool.None)
+            {
+                bool willBeStoredProcedure = (isCurrentEntityStoredProcedure && options.SourceType is null) ||
+                                              (options.SourceType is not null && IsStoredProcedure(options));
```
Aniruddh25: could reuse the boolean `doOptionsRepresentStoredProcedure` here

Aniruddh25 (changes requested):
Posting comments so far

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
-            EntityActionFields? updatedFields = GetFieldsForOperation(options.FieldsToInclude, options.FieldsToExclude);
-            EntityCacheOptions? updatedCacheOptions = ConstructCacheOptions(options.CacheEnabled, options.CacheTtl);
+            EntityMcpOptions? updatedMcpOptions = ConstructUpdatedMcpOptions(entity, options);
 
-            if (!updatedGraphQLDetails.Enabled)
+            // Add these missing variable definitions
+            EntityActionPolicy? policy = GetPolicyForOperation(options.PolicyRequest, options.PolicyDatabase);
+            EntityActionFields? field = GetFieldsForOperation(options.FieldsToInclude, options.FieldsToExclude);
+
+            // Validate that permissions are provided when using field include/exclude
```
Aniruddh25: Similar validation for policy is now missing. It was happening before when we provided the message: `--permissions is mandatory with --policy-request and --policy-database.`

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
+            }
 
-                if (options.FieldsToInclude is not null && options.FieldsToInclude.Any()
-                    || options.FieldsToExclude is not null && options.FieldsToExclude.Any())
-                {
-                    _logger.LogInformation("--permissions is mandatory with --fields.include and --fields.exclude.");
-                    return false;
-                }
+            // Handle relationships
+            Dictionary<string, EntityRelationship>? updatedRelationships = null;
```
Aniruddh25: I dont understand why was any of this refactored. How is this related to the updatedMcpOptions change? This could have been avoided to make it easier for PR reviewer.

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
                 if (updatedPermissions is null)
                 {
-                    _logger.LogError("Failed to update permissions.");
+                    _logger.LogError("Failed to update entity permissions.");
                     return false;
                 }
             }
             else
             {
+                updatedPermissions = entity.Permissions;
```
Aniruddh25: We are no longer verifying the stored procedure permissions. Where is the call to: 

                `if (updatedSourceType is EntitySourceType.StoredProcedure &&
                    !VerifyPermissionOperationsForStoredProcedures(entity.Permissions))
                {
                    return false;
                }`

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
-                {
-                    _logger.LogInformation("--permissions is mandatory with --fields.include and --fields.exclude.");
-                    return false;
-                }
+            // Handle relationships
+            Dictionary<string, EntityRelationship>? updatedRelationships = null;
 
-                if (options.PolicyRequest is not null || options.PolicyDatabase is not null)
+            // Only verify and update relationships if relationship options are provided
+            if (options.Cardinality is not null || options.TargetEntity is not null)
```
Aniruddh25: why are we not checking options.Relationship any more?

schemas/dab.draft.schema.json
```diff
@@ -774,12 +774,24 @@
...
-                            "required": ["name"],
+                            "required": [ "name" ],
                             "properties": {
-                              "name": { "type": "string", "description": "Parameter name" },
-                              "required": { "type": "boolean", "description": "Is parameter required" },
-                              "default": { "type": ["string", "number", "boolean", "null"], "description": "Default value" },
-                              "description": { "type": "string", "description": "Parameter description. Since descriptions for multiple parameters are provided as a comma-separated string, individual parameter descriptions must not contain a comma (',')." }
+                              "name": {
+                                "type": "string",
+                                "description": "Parameter name"
```
RubenCerna2079: nit:
```suggestion
                                "description": "Parameter name."
```

schemas/dab.draft.schema.json
```diff
@@ -774,12 +774,24 @@
...
+                              },
+                              "required": {
+                                "type": "boolean",
+                                "description": "Is parameter required"
```
RubenCerna2079: nit: 
```suggestion
                                "description": "Is parameter required."
```

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
                 {
-                    _logger.LogInformation("--permissions is mandatory with --policy-request and --policy-database.");
-                    return false;
+                    EntityRelationship? newRelationship = CreateNewRelationshipWithUpdateOptions(options);
+                    if (newRelationship is not null)
+                    {
+                        updatedRelationships = entity.Relationships is null
+                            ? new Dictionary<string, EntityRelationship> { { options.Relationship ?? options.TargetEntity!, newRelationship } }
+                            : new Dictionary<string, EntityRelationship>(entity.Relationships) { [options.Relationship ?? options.TargetEntity!] = newRelationship };
+                    }
```
Aniruddh25: if newRelationship ends up being null, we returned false previously.see lines 1678-1681

schemas/dab.draft.schema.json
```diff
@@ -774,12 +774,24 @@
...
+                              },
+                              "default": {
+                                "type": [ "string", "number", "boolean", "null" ],
+                                "description": "Default value"
```
RubenCerna2079: nit: 
```suggestion
                                "description": "Default value."
```

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
             {
-                _logger.LogError(errorMessage);
-                return false;
+                updatedGraphQLDetails = updatedGraphQLDetails with { Operation = null };
             }
 
+            // Check for GraphQL multiple create options
+            bool isMultipleCreateEnabledForGraphQL = initialConfig.IsMultipleCreateOperationEnabled();
+
+            // We log a warning when enabling GraphQL for SP entity without multiple-create option as it may not be supported.
```
Aniruddh25: Where is this warning coming from? we dont need multiple create option for enabling GraphQL for a mutation

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
-
-                updatedRelationships[options.Relationship] = new_relationship;
+                updatedRelationships = entity.Relationships;
             }
 
-            bool hasFields = options.FieldsNameCollection != null && options.FieldsNameCollection.Count() > 0;
-            bool hasMappings = options.Map != null && options.Map.Any();
-            bool hasKeyFields = options.SourceKeyFields != null && options.SourceKeyFields.Any();
+            // Handle mappings
+            Dictionary<string, string>? updatedMappings;
```
Aniruddh25: Why is all the logic to check for hasMappings & hasKeyFields removed?

src/Cli.Tests/AddEntityTests.cs
```diff
@@ -597,8 +609,9 @@ public void TestAddEntityPermissionWithInvalidOperation(IEnumerable<string> perm
                 fieldsNameCollection: [],
                 fieldsAliasCollection: [],
                 fieldsDescriptionCollection: [],
-                fieldsPrimaryKeyCollection: []
-                );
+                fieldsPrimaryKeyCollection: [],
+                mcpCustomToolEnabled: null,
+                mcpDmlToolsEnabled: null);
```
RubenCerna2079: I think we should add a test that has the custom tool and the dml-tools enabled

src/Cli.Tests/Snapshots/EndToEndTests.TestUpdatingStoredProcedureWithRestMethods.verified.txt
```diff
@@ -1,4 +1,4 @@
-{
+﻿{
```
RubenCerna2079: I see you added multiple binary files, what are the files in the picture for?
image

src/Cli.Tests/Snapshots/UpdateEntityTests.TestConversionOfSourceObject_c26902b0e44f97cd.verified.txt
```diff
@@ -49,8 +39,15 @@
           Methods: [
             Post
           ],
+          Path: /api,
```
RubenCerna2079: Since you removed the `Fields` property, shouldn't the `KeyFields` property be added instead similar to other snapshots?

src/Cli.Tests/Snapshots/UpdateEntityTests.TestUpdateExistingMappings.verified.txt
```diff
@@ -70,7 +53,13 @@
               }
             ]
           }
-        ]
+        ],
+        Mappings: {
+          addr: Company Address,
+          id: Identity,
```
RubenCerna2079: Why do we add `id: Identity` if that is not in the original snapshot?

src/Cli.Tests/Snapshots/UpdateEntityTests.TestUpdateRestAndGraphQLSettingsForStoredProcedures_f4cadb897fc5b0fe.verified.txt
```diff
@@ -33,7 +33,7 @@
           Singular: book,
           Plural: books,
           Enabled: true,
-          Operation: Query
+          Operation: Mutation
```
RubenCerna2079: Is there a reason to change the operations from query to mutation?

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
+                if (updatedRestDetails.Path == null || string.IsNullOrEmpty(updatedRestDetails.Path) || !IsRouteValid(updatedRestDetails.Path))
                 {
-                    _logger.LogWarning("Using legacy 'mappings' and 'key-fields' properties. Consider using 'fields' for new entities.");
+                    updatedRestDetails = updatedRestDetails with { Path = RestRuntimeOptions.DEFAULT_PATH };
                 }
             }
 
-            if (!ValidateFields(fields, out string errorMessage))
+            // If entity is being converted from stored procedure, clear the GraphQL operation
+            if (isCurrentEntityStoredProcedure && !doOptionsRepresentStoredProcedure)
```
Aniruddh25: why is this change?

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
             }
             else
             {
-                fields = entity.Fields?.ToList() ?? new List<FieldMetadata>();
-                if (entity.Mappings is not null || entity.Source?.KeyFields is not null)
+                fields = entity.Fields;
+            }
+
+            // If entity is being converted to stored procedure, default the REST settings
+            if (IsEntityBeingConvertedToStoredProcedure(entity, options))
```
Aniruddh25: why is this change?

src/Cli/ConfigGenerator.cs
```diff
@@ -2862,5 +2816,120 @@ private static bool ValidateFields(
...
+            return null;
+        }
+
+        /// <summary>
+        /// Validates if a REST route is valid.
+        /// </summary>
+        private static bool IsRouteValid(string route)
+        {
+            // Basic validation for route
+            return !string.IsNullOrWhiteSpace(route) && route.StartsWith("/");
```
Aniruddh25: We have REST route validation already being done in the codebase. We should reuse that function.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -773,11 +773,11 @@ public void TestUpdateEntity()
...
         CollectionAssert.AreEqual(new string[] { "id" }, relationship.LinkingTargetFields);
 
-        Assert.IsNotNull(entity.Fields);
-        Assert.AreEqual(2, entity.Fields.Count);
-        Assert.AreEqual(entity.Fields[0].Alias, "identity");
-        Assert.AreEqual(entity.Fields[1].Alias, "Company Name");
-        Assert.IsNull(entity.Mappings);
+        Assert.IsNotNull(entity.Mappings);
+        Assert.AreEqual(2, entity.Mappings.Count);
+        Assert.AreEqual("identity", entity.Mappings["id"]);
```
Aniruddh25: Mappings has been deprecated. We should test for entity.Fields

src/Config/ObjectModel/EntityMcpOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// Indicates whether custom tools are enabled for this entity.
+        /// </summary>
+        [JsonPropertyName("custom-tool")]
+        public bool? CustomToolEnabled { get; init; } = false;
+
+        /// <summary>
+        /// Indicates whether DML tools are enabled for this entity.
+        /// </summary>
+        [JsonPropertyName("dml-tools")]
+        public bool? DmlToolEnabled { get; init; } = true;
```
Aniruddh25: Does this have to be nullable? It has a default value of true, right? same for CustomToolEnabled

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    public override JsonConverter CreateConverter(Type typeToConvert, JsonSerializerOptions options)
+    {
+        return new EntityMcpOptionsConverter();
+    }
+
+    /// <summary>
+    /// Converter for EntityMcpOptions that handles both boolean and object representations.
+    /// When boolean: true enables dml-tools, false disables dml-tools.
```
Aniruddh25: When boolean, what happens to custom-tool? Should mention it here.

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,127 @@
...
+                {
+                    writer.WriteBoolean("dml-tools", value.DmlToolEnabled ?? true);
+                }
+
+                writer.WriteEndObject();
+            }
+            else
+            {
+                // Nothing provided, write null (will be omitted by DefaultIgnoreCondition)
+                writer.WriteNullValue();
```
Aniruddh25: Why do we even write out the null? Cant we skip writing altogether?

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMySql.verified.txt
```diff
@@ -390,18 +390,6 @@
           Object: books,
           Type: Table
         },
-        Fields: [
-          {
-            Name: id,
-            Alias: id,
-            PrimaryKey: false
```
Aniruddh25, line 397: Removal of fields is not expected in these snapshots. This will be a regression.

src/Service.Tests/dab-config.PostgreSql.json
```diff
@@ -2697,7 +2728,9 @@
           ],
           "target.fields": [
             "accountkey"
-          ]
+          ],
+          "linking.source.fields": [],
```
Aniruddh25: if we are writing linking.source.fields even though they are empty, it signals a bug in CLI. Can you please create an issue tracking this?

src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -31,29 +35,23 @@
         "provider": "StaticWebApps"
       },
       "mode": "development"
-    },
-    "telemetry": {
```
RubenCerna2079, line 35: Is there a reason to delete the telemetry section?

src/Service.Tests/dab-config.CosmosDb_NoSql.json
```diff
@@ -800,4 +798,4 @@
       ]
     }
   }
-}
+}
```
RubenCerna2079: Missing end of line

src/Service.Tests/dab-config.MySql.json
```diff
@@ -2322,9 +2390,11 @@
...
-          ]
+          ],
+          "linking.source.fields": [],
+          "linking.target.fields": []
         }
       }
     }
   }
-}
+}
```
RubenCerna2079: Missing end of line

src/Cli/Commands/EntityOptions.cs
```diff
@@ -132,5 +137,11 @@ public EntityOptions(
 
         [Option("fields.primary-key", Required = false, Separator = ',', HelpText = "Set this field as a primary key.")]
         public IEnumerable<bool>? FieldsPrimaryKeyCollection { get; }
+
+        [Option("mcp.custom-tool", Required = false, HelpText = "Enable this stored procedure as a custom MCP tool. Valid values: true, false")]
```
RubenCerna2079: nit:
```suggestion
        [Option("mcp.custom-tool", Required = false, HelpText = "Enable entity as a custom MCP tool, only available for stored procedure entities. Valid values: true, false")]
```

src/Cli/ConfigGenerator.cs
```diff
@@ -450,12 +450,34 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
             EntityGraphQLOptions graphqlOptions = ConstructGraphQLTypeDetails(options.GraphQLType, graphQLOperationsForStoredProcedures);
             EntityCacheOptions? cacheOptions = ConstructCacheOptions(options.CacheEnabled, options.CacheTtl);
 
+            // Validate and construct MCP options
+            EntityMcpOptions? mcpOptions = null;
+
+            if (options.McpCustomToolEnabled is not null and not CliBool.None)
+            {
+                if (!isStoredProcedure)
+                {
```
RubenCerna2079: Couldn't all of this be only one `if` statement

src/Cli/ConfigGenerator.cs
```diff
@@ -450,12 +450,34 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
...
+                    return false;
+                }
+            }
+
+            // Construct MCP options if any MCP-related flags are provided
+            // For stored procedures: both custom-tool and dml-tools can be set
+            // For other entity types: only dml-tools can be set
+            if (options.McpCustomToolEnabled is not null and not CliBool.None ||
+                options.McpDmlToolsEnabled is not null and not CliBool.None)
+            {
```
RubenCerna2079: The `ConstructMcpOptions` function already checks if the the custom-tool and the dml-tools exist, so do we really need this if check?

src/Cli/ConfigGenerator.cs
```diff
@@ -1597,6 +1618,16 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
                 return false;
             }
 
+            // Validate MCP custom-tool is only for stored procedures
+            if (options.McpCustomToolEnabled is not null and not CliBool.None)
+            {
+                if (!(isCurrentEntityStoredProcedure || doOptionsRepresentStoredProcedure))
+                {
```
RubenCerna2079: Couldn't this be a single `if` statement?

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
 
             EntityRestOptions updatedRestDetails = ConstructUpdatedRestDetails(entity, options, initialConfig.DataSource.DatabaseType == DatabaseType.CosmosDB_NoSQL);
             EntityGraphQLOptions updatedGraphQLDetails = ConstructUpdatedGraphQLDetails(entity, options);
-            EntityPermission[]? updatedPermissions = entity!.Permissions;
```
RubenCerna2079, line 1617: I don't understand why the permissions, relationships, and mappings variables were deleted

src/Cli/ConfigGenerator.cs
```diff
@@ -1614,251 +1645,174 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
...
 
-                fields = mergedFields;
-
-                // If user didn't mark any PK in fields, carry over existing source key-fields
-                if (!fields.Any(f => f.PrimaryKey) && updatedSource.KeyFields is { Length: > 0 })
+                // Parse the mapping strings (format: "backendName:exposedName")
+                foreach (string mapping in options.Map)
                 {
-                    foreach (string k in updatedSource.KeyFields)
+                    string[] parts = mapping.Split(':');
```
RubenCerna2079: I don't understand why we split the mapping. Was this not done before? Or was it not saved as a dictionary of some sort?

src/Cli/ConfigGenerator.cs
```diff
@@ -2489,7 +2443,7 @@ private static bool TryAddGraphQLOperationForStoredProcedure(EntityOptions optio
         /// <param name="options">Input from update command</param>
         /// <returns>Boolean -> when the entity's REST configuration is true/false.
         /// RestEntitySettings -> when a non stored procedure entity is configured with granular REST settings (Path).
-        /// RestStoredProcedureEntitySettings -> when a stored procedure entity is configured with explicit SupportedRestMethods.
+        /// RestStoredProcedureEntitySettings-> when a stored procedure entity is configured with explicit SupportedRestMethods.
```
RubenCerna2079: niy: 
```suggestion
        /// RestStoredProcedureEntitySettings -> when a stored procedure entity is configured with explicit SupportedRestMethods.
```

src/Cli/ConfigGenerator.cs
```diff
@@ -2862,5 +2816,120 @@ private static bool ValidateFields(
 
             return true;
         }
+
+        /// <summary>
+        /// Constructs MCP options for an entity based on the provided CLI options.
+        /// Only applicable for stored-procedure entities.
```
RubenCerna2079: Does this mean that this function should only be called for stored-procedure entities? At least it doesn't seem to be used that way.

src/Cli/ConfigGenerator.cs
```diff
@@ -2862,5 +2816,120 @@ private static bool ValidateFields(
...
+        /// existing MCP configuration for an entity
+        /// </summary>
+        /// <param name="entity">Entity for which MCP settings are updated</param>
+        /// <param name="options">Input from update command</param>
+        /// <returns>Updated EntityMcpOptions or null</returns>
+        private static EntityMcpOptions? ConstructUpdatedMcpOptions(Entity entity, UpdateOptions options)
+        {
+            // Get the current and potential new entity types
+            bool isCurrentEntityStoredProcedure = IsStoredProcedure(entity);
+            bool isBecomingStoredProcedure = options.SourceType is not null && IsStoredProcedure(options);
```
RubenCerna2079: nit:
```suggestion
            bool isBecomingEntityStoredProcedure = options.SourceType is not null && IsStoredProcedure(options);
```

src/Cli/ConfigGenerator.cs
```diff
@@ -2862,5 +2816,120 @@ private static bool ValidateFields(
...
+
+            // MCP custom-tool is only applicable for stored procedures
+            if (!isCurrentEntityStoredProcedure && !isBecomingStoredProcedure)
+            {
+                // Not a stored procedure, MCP custom-tool not applicable
+                if (options.McpCustomToolEnabled is not null and not CliBool.None)
+                {
+                    _logger.LogWarning("--mcp.custom-tool is only applicable for stored procedures and will be ignored.");
```
RubenCerna2079: Why for the add MCP option do we throw an error and for the update option do we only throw a warning? I think both should throw an error.

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,127 @@
...
+                    {
+                        break;
+                    }
+
+                    if (reader.TokenType == JsonTokenType.PropertyName)
+                    {
+                        string? propertyName = reader.GetString();
+                        reader.Read(); // Move to the value
+
+                        if (propertyName == "custom-tool")
```
RubenCerna2079: Wouldn't it be better to have a switch case for this instead of if and else if statements? With the switch case, we can easily add more property names if it is required and also add a default case that throws an exception if the property name used is invalid.

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
RubenCerna2079: Agree with Ani, I also think this should apply to the `if (value == null)` at the start of the write function.

src/Config/ObjectModel/EntityMcpOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            if (dmlToolsEnabled is not null)
+            {
+                this.DmlToolEnabled = dmlToolsEnabled;
+                this.UserProvidedDmlToolsEnabled = true;
+            }
+            else
+            {
+                this.DmlToolEnabled = true;
```
RubenCerna2079: I think either this else statement is unnecessary or there needs to be an else statement for the `customToolEnabled` variable

src/Service.Tests/dab-config.DwSql.json
```diff
@@ -1,2721 +1,2871 @@
...
-          "linking.source.fields": [],
-          "linking.target.fields": []
-        }
-      }
-    }
-  }
-}
\ No newline at end of file
+﻿{
+    "$schema":  "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json",
```
RubenCerna2079: The only changes I see here are the mcp additionsto runtime. Please fix the file so that it only shows your changes

src/Service.Tests/dab-config.MsSql.json
```diff
@@ -1,3852 +1,4148 @@
...
-              "action": "execute"
-            }
-          ]
-        }
-      ]
-    }
-  }
+﻿{
+    "$schema":  "https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json",
+    "data-source":  {
```
RubenCerna2079: Same comment as with the `DwSql` file

config-generators/config-generator.ps1
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli.Tests/Snapshots/UpdateEntityTests.TestConversionOfSourceObject_036a859f50ce167c.verified.txt
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli/ConfigGenerator.cs
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli/ConfigGenerator.cs
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli/ConfigGenerator.cs
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli/ConfigGenerator.cs
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli/ConfigGenerator.cs
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli/ConfigGenerator.cs
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli.Tests/EndToEndTests.cs
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Config/ObjectModel/EntityMcpOptions.cs
souvikghosh04: Fixed

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
souvikghosh04: Fixed

src/Service.Tests/Snapshots/ConfigurationTests.TestReadingRuntimeConfigForMySql.verified.txt
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Service.Tests/dab-config.PostgreSql.json
souvikghosh04: Reverted and fixed as its not relevant for now.

schemas/dab.draft.schema.json
souvikghosh04: Reverted and fixed as its not relevant for now.

schemas/dab.draft.schema.json
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Cli.Tests/Snapshots/EndToEndTests.TestUpdatingStoredProcedureWithRestMethods.verified.txt
souvikghosh04: Reverted and fixed as its not relevant for now.

src/Service.Tests/dab-config.CosmosDb_NoSql.json
souvikghosh04: Reverted and fixed as its not relevant for now.

schemas/dab.draft.schema.json
souvikghosh04: I have updated the description based on the updated schema

src/Cli/ConfigGenerator.cs
souvikghosh04: This code is updated now and we don't have this change currently.

src/Cli/ConfigGenerator.cs
souvikghosh04: this is reverted and fixed.

src/Cli/ConfigGenerator.cs
souvikghosh04: This change is reverted and fixed.

src/Cli/ConfigGenerator.cs
souvikghosh04: This is fixed now.

src/Cli/ConfigGenerator.cs
souvikghosh04: This is fixed now.

src/Cli/ConfigGenerator.cs
souvikghosh04: This is fixed.

src/Cli/ConfigGenerator.cs
souvikghosh04: This is not required and the change is reverted.

schemas/dab.draft.schema.json
souvikghosh04: draft-schema changes has been updated and we don't have this anymore.

src/Cli/ConfigGenerator.cs
souvikghosh04: The `ConfigGenerator.cs` is updated and this change is outdated.

src/Cli/ConfigGenerator.cs
souvikghosh04: The `ConfigGenerator.cs` is updated and this change is outdated.

src/Cli/ConfigGenerator.cs
souvikghosh04: This file has been updated fully. Resolving this as its outdated and doesn't exist.

src/Cli/ConfigGenerator.cs
souvikghosh04: This file has been updated fully. Resolving this as its outdated and doesn't exist.

src/Cli/ConfigGenerator.cs
souvikghosh04: This has been fixed.

src/Cli.Tests/AddEntityTests.cs
souvikghosh04: I have separate task for that since there are many scenarios I want to cover- #3008

src/Cli/ConfigGenerator.cs
```diff
@@ -450,6 +450,18 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
...
+            if (options.McpDmlTools is not null || options.McpCustomTool is not null)
+            {
+                EntityMcpOptions? mcpOptions = ConstructMcpOptions(options.McpDmlTools, options.McpCustomTool, isStoredProcedure);
+                if (mcpOptions is null)
+                {
+                    _logger.LogError("Failed to construct MCP options.");
+                    return false;
+                }
+            }
+
```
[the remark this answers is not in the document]
souvikghosh04: Fixed. Thanks!

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,125 @@
...
+                        string? propertyName = reader.GetString();
+                        reader.Read(); // Move to the value
+
+                        switch (propertyName)
+                        {
+                            case "custom-tool":
+                                customToolEnabled = reader.TokenType == JsonTokenType.True;
+                                break;
+                            case "dml-tools":
+                                dmlToolsEnabled = reader.TokenType == JsonTokenType.True;
```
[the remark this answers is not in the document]
souvikghosh04: Fixed. Thanks!

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,125 @@
...
+            // Check if we should write as boolean shorthand
+            // Write as boolean if: only dml-tools is set (or custom-tool is default false)
+            bool writeAsBoolean = !value.UserProvidedCustomToolEnabled && value.UserProvidedDmlToolsEnabled;
+
+            if (writeAsBoolean)
+            {
+                // Write as boolean shorthand
+                writer.WriteBooleanValue(value.DmlToolEnabled);
+            }
+            else if (value.UserProvidedCustomToolEnabled || value.UserProvidedDmlToolsEnabled)
```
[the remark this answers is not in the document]
souvikghosh04: This is a false positive from Copilot's static analysis.

Why Copilot is wrong:
Copilot claims this is "always false" because it misunderstands the logic. The condition is specifically designed to detect when only dml-tools is provided (the boolean shorthand scenario like "mcp": true).

Proof from our testing:

The Author entity was created with only --mcp.dml-tools true
It correctly serialized as "mcp": true (boolean shorthand)
This proves the condition evaluated to true and took the boolean serialization path

schemas/dab.draft.schema.json
```diff
@@ -961,6 +961,31 @@
                   "default": 5
                 }
               }
+            },
+            "mcp": {
+              "oneOf": [
+                {
+                  "type": "boolean",
+                  "description": "Boolean shorthand: true enables dml-tools, false disables dml-tools."
```
RubenCerna2079: I think a bit more description should be added to specify that custom-tools don't apply to this case

schemas/dab.draft.schema.json
```diff
@@ -1145,6 +1170,33 @@
...
+                  "source": {
+                    "properties": {
+                      "type": {
+                        "const": "stored-procedure"
+                      }
+                    },
+                    "required": ["type"]
+                  }
+                },
+                "errorMessage": "custom-tool can only be enabled for entities with source type 'stored-procedure'"
```
RubenCerna2079:
```suggestion
                "errorMessage": "custom-tool can only be enabled for entities with source type 'stored-procedure'."
```

src/Cli/Commands/AddOptions.cs
```diff
@@ -43,7 +43,9 @@ public AddOptions(
             IEnumerable<string>? fieldsAliasCollection,
             IEnumerable<string>? fieldsDescriptionCollection,
             IEnumerable<bool>? fieldsPrimaryKeyCollection,
-            string? config
+            string? mcpDmlTools = null,
+            string? mcpCustomTool = null,
```
RubenCerna2079, line 47: Why are these variables set as null here? Shouldn't they follow the pattern that is already set?

src/Cli/Commands/EntityOptions.cs
```diff
@@ -34,7 +34,9 @@ public EntityOptions(
             IEnumerable<string>? fieldsAliasCollection,
             IEnumerable<string>? fieldsDescriptionCollection,
             IEnumerable<bool>? fieldsPrimaryKeyCollection,
-            string? config
+            string? mcpDmlTools = null,
+            string? mcpCustomTool = null,
```
RubenCerna2079, line 38: Same as previous comment

src/Cli/Commands/EntityOptions.cs
```diff
@@ -132,5 +136,11 @@ public EntityOptions(
 
         [Option("fields.primary-key", Required = false, Separator = ',', HelpText = "Set this field as a primary key.")]
         public IEnumerable<bool>? FieldsPrimaryKeyCollection { get; }
+
+        [Option("mcp.dml-tools", Required = false, HelpText = "Enable MCP DML (Data Manipulation Language) tools for this entity. Allows CRUD operations via MCP.")]
```
RubenCerna2079: nit:
```suggestion
        [Option("mcp.dml-tools", Required = false, HelpText = "Enable MCP DML (Data Manipulation Language) tools for this entity. Allows CRUD operations via MCP. Default value is true.")]
```

src/Cli/Commands/EntityOptions.cs
```diff
@@ -132,5 +136,11 @@ public EntityOptions(
...
+        public string? McpDmlTools { get; }
+
+        [Option("mcp.custom-tool", Required = false, HelpText = "Enable MCP custom tool for this entity. Only valid for stored procedures.")]
```
RubenCerna2079: nit:
```suggestion
        [Option("mcp.custom-tool", Required = false, HelpText = "Enable MCP custom tool for this entity. Only valid for stored procedures. Default value is false.")]
```

src/Cli/Commands/UpdateOptions.cs
```diff
@@ -51,7 +51,9 @@ public UpdateOptions(
             IEnumerable<string>? fieldsAliasCollection,
             IEnumerable<string>? fieldsDescriptionCollection,
             IEnumerable<bool>? fieldsPrimaryKeyCollection,
-            string? config)
+            string? mcpDmlTools = null,
+            string? mcpCustomTool = null,
```
RubenCerna2079, line 55: Same comment related to null value as before.

src/Cli/ConfigGenerator.cs
```diff
@@ -1621,6 +1633,26 @@ public static bool TryUpdateExistingEntity(UpdateOptions options, RuntimeConfig
             EntityActionFields? updatedFields = GetFieldsForOperation(options.FieldsToInclude, options.FieldsToExclude);
             EntityCacheOptions? updatedCacheOptions = ConstructCacheOptions(options.CacheEnabled, options.CacheTtl);
 
+            // Determine if the entity is or will be a stored procedure
+            bool isStoredProcedureAfterUpdate = doOptionsRepresentStoredProcedure || (isCurrentEntityStoredProcedure && options.SourceType is null);
```
RubenCerna2079, line 1638: Is there a reason to check the `options.SourceType`?

src/Cli/ConfigGenerator.cs
```diff
@@ -450,6 +450,17 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
             EntityGraphQLOptions graphqlOptions = ConstructGraphQLTypeDetails(options.GraphQLType, graphQLOperationsForStoredProcedures);
             EntityCacheOptions? cacheOptions = ConstructCacheOptions(options.CacheEnabled, options.CacheTtl);
 
+            EntityMcpOptions? mcpOptionsToUse = null;
```
RubenCerna2079:
```suggestion
            EntityMcpOptions? mcpOptions = null;
```

src/Config/ObjectModel/Entity.cs
```diff
@@ -39,6 +40,9 @@ public record Entity
     public EntityCacheOptions? Cache { get; init; }
     public EntityHealthCheckConfig? Health { get; init; }
 
+    [JsonConverter(typeof(EntityMcpOptionsConverterFactory))]
```
RubenCerna2079, line 43: Is there a reason you need to add the JsonConverter here? If the other variables don't need it, then why is this one different?

src/Config/ObjectModel/EntityMcpOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// </summary>
+        [JsonPropertyName("custom-tool")]
+        public bool CustomToolEnabled { get; init; } = false;
+
+        /// <summary>
+        /// Indicates whether DML tools are enabled for this entity.
+        /// Defaults to false when not explicitly provided.
+        /// </summary>
+        [JsonPropertyName("dml-tools")]
+        public bool DmlToolEnabled { get; init; } = false;
```
RubenCerna2079: Shouldn't dml-tools default to true?

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.ObjectModel;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+
+/// <summary>
+/// Factory for creating EntityMcpOptions converters.
+/// </summary>
+internal class EntityMcpOptionsConverterFactory : JsonConverterFactory
```
RubenCerna2079, line 13: I think this might have to be added to the `RuntimeConfigLoader.cs`

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,125 @@
...
+                            case "dml-tools":
+                                dmlToolsEnabled = reader.GetBoolean();
+                                break;
+                            default:
+                                throw new JsonException($"Unknown property '{propertyName}' in EntityMcpOptions");
+                        }
+                    }
+                }
+
+                return new EntityMcpOptions(customToolEnabled, dmlToolsEnabled);
```
RubenCerna2079: nit: Wouldn't it be better to move this return inside the `if (reader.TokenType == JsonTokenType.EndObject)`?

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
```diff
@@ -0,0 +1,125 @@
...
+        public override void Write(Utf8JsonWriter writer, EntityMcpOptions value, JsonSerializerOptions options)
+        {
+            if (value == null)
+            {
+                return;
+            }
+
+            // Check if we should write as boolean shorthand
+            // Write as boolean if: only dml-tools is set (or custom-tool is default false)
+            bool writeAsBoolean = !value.UserProvidedCustomToolEnabled && value.UserProvidedDmlToolsEnabled;
```
RubenCerna2079, line 97: Wouldn't this not allow the user to specify that the dml-tools are true and only set the shorthand every time?

What I am worried about is that there is no way to write the following structure.
```
entity: {
  mcp: {
    dml-tools: true
  }
}
```

src/Cli/Commands/AddOptions.cs
souvikghosh04: These parameters use `= null` to match the existing pattern for optional CLI parameters (e.g., `config = null`). The default indicates the parameter is optional and wasn't provided by the user.

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
souvikghosh04: The converter is already registered via the `[JsonConverter(typeof(EntityMcpOptionsConverterFactory))]` attribute on the `Mcp` property in `Entity.cs` (line 43). This attribute-based registration is sufficient for `System.Text.Json` to use the converter automatically. No additional registration in `RuntimeConfigLoader.cs` is needed.

schemas/dab.draft.schema.json
```diff
@@ -1145,6 +1170,33 @@
...
                 }
               }
+            },
+            {
+              "if": {
+                "properties": {
+                  "mcp": {
+                    "properties": {
+                      "custom-tool": {
+                        "const": true
```
Aniruddh25, line 1180: What does "const": true imply here? Shouldnt this be one of 'true' or 'false'?

src/Cli.Tests/AddEntityTests.cs
```diff
@@ -633,5 +633,293 @@ private Task ExecuteVerifyTest(AddOptions options, string config = INITIAL_CONFI
...
+                graphQLOperationForStoredProcedure: null,
+                parametersNameCollection: null,
+                parametersDescriptionCollection: null,
+                parametersRequiredCollection: null,
+                parametersDefaultCollection: null,
+                fieldsNameCollection: [],
+                fieldsAliasCollection: [],
+                fieldsDescriptionCollection: [],
+                fieldsPrimaryKeyCollection: [],
+                mcpDmlTools: "false",
```
Aniruddh25: Cant the above test and this one be combined into one DataTestMethod with DataRow argument for `mcpDmlTools`?

src/Cli.Tests/Snapshots/AddEntityTests.AddStoredProcedureWithBothMcpProperties.verified.txt
```diff
@@ -0,0 +1,63 @@
...
+              {
+                Action: Execute
+              }
+            ]
+          }
+        ],
+        Mcp: {
+          CustomToolEnabled: true,
+          DmlToolEnabled: false,
+          UserProvidedCustomToolEnabled: true,
```
Aniruddh25: UserProvidedCustomToolEnabled can be added to the properties to ignore in the snapshot. Look at similar other flags which get ignored.

src/Cli.Tests/UpdateEntityTests.cs
```diff
@@ -1211,5 +1215,461 @@ private Task ExecuteVerifyTest(string initialConfig, UpdateOptions options, Veri
...
+                mcpCustomTool: mcpCustomTool
+            );
+
+            string initialConfig = GetInitialConfigString() + "," + @"
+                    ""entities"": {
+                            ""MyEntity"": {
+                                ""source"": ""MyTable"",
+                                ""permissions"": [
+                                    {
+                                        ""role"": ""anonymous"",
```
Aniruddh25, line 1617: All the tests added in the updateEntityTests.cs seem to start with no mcp option in the initialConfig. This is as good as the AddEntityTests.cs. 
The update tests should start with an already existing mcp config option in the initial config and then apply the update options in such a way that the existing option is modified. For example, from disabled to enabled etc. This would really "test" the update scenario otherwise all the existing test in the latest iteration are simply testing the add scenario.

src/Cli/ConfigGenerator.cs
Aniruddh25: that is to identify if the options are about to change the source type to a table or view even if currently the entity is a stored procedure. If its null, that means no change to the source type is happening. If option.sourcetype is not null, and is a stored procedure, we capture that in the first flag- `doOptionsRepresentStoredProcedure`

src/Config/ObjectModel/EntityMcpOptions.cs
Aniruddh25: I agree, dml-tools should default to true.

src/Service.Tests/Configuration/EntityMcpConfigurationTests.cs
```diff
@@ -0,0 +1,504 @@
...
+            // Act
+            bool success = RuntimeConfigLoader.TryParseConfig(config, out RuntimeConfig runtimeConfig);
+
+            // Assert
+            Assert.IsTrue(success, "Config should parse successfully");
+            Assert.IsNotNull(runtimeConfig);
+            Assert.IsTrue(runtimeConfig.Entities.ContainsKey("Book"));
+
+            Entity bookEntity = runtimeConfig.Entities["Book"];
+            Assert.IsNull(bookEntity.Mcp, "MCP options should be null when not specified");
```
Aniruddh25: Even if the option is null, do we ensure that, that the dml tools are enabled at entity level?

I.e. Does this PR use the entity Mcp Options in the IsToolEnabled check or is that going to be in a separate PR?

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
RubenCerna2079: Why was this resolved if there is no change? Is there going to be a change related to this or not?

src/Service.Tests/Configuration/EntityMcpConfigurationTests.cs
```diff
@@ -0,0 +1,504 @@
...
+            Assert.IsFalse(bookEntity.Mcp.CustomToolEnabled, "CustomTool should be disabled (default)");
+        }
+
+        /// <summary>
+        /// Test that deserializing object format with both properties works correctly.
+        /// </summary>
+        [TestMethod]
+        public void DeserializeConfig_McpObject_SetsBothProperties()
+        {
+            // Arrange
```
RubenCerna2079, line 118: I think it could be better to create a function for this config string that is only has a few changes.

schemas/dab.draft.schema.json
souvikghosh04: The "const": true is valid JSON Schema syntax used in conditional validation (the if block). It checks if the custom-tool property is exactly true. This allows us to enforce that when custom-tool: true is set, the entity **must be a stored procedure**. The const keyword is specifically designed for this type of exact-value validation in JSON Schema.
We have similar pattern in other places, for e.g. this one- 

```
              "if": {
                "properties": {
                  "enabled": {
                    "const": true
                  }
                }
              },
              "then": {
                "required": [ "path" ]
              }
```

Reference- https://json-schema.org/understanding-json-schema/reference/const

src/Service.Tests/Configuration/EntityMcpConfigurationTests.cs
souvikghosh04: yes, the entity-level MCP configuration (Entity.Mcp.DmlToolEnabled) defaults to true when null/unspecified. 
However, this property is not currently being checked by the MCP tools when they execute. The tools only check the runtime-level configuration (RuntimeConfig.McpDmlTools.ReadRecords, etc. it will be addressed in a separate PR. added a task - https://github.com/Azure/data-api-builder/issues/3017

src/Cli.Tests/AddEntityTests.cs
souvikghosh04: Consolidated some of the tests. some tests like stored proc specific scenarios had multiple variations which are kept as is but refactored.

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
souvikghosh04: yes, this class is added there.

src/Service.Tests/Configuration/EntityMcpConfigurationTests.cs
souvikghosh04: The duplicate config templates are inside a utility functionality and additional test/scenario specific are appended to that.

src/Cli/Commands/EntityOptions.cs
souvikghosh04: as explained previously, this indicates the parameter is optional and not provided by user.

src/Cli/Commands/UpdateOptions.cs
souvikghosh04: duplicate comment. This is explained in AddOptions.cs and UpdateOptions.cs.

src/Config/Converters/EntityMcpOptionsConverterFactory.cs
souvikghosh04: This is supported but doesn't make sense. The default, if unspecified remains true and boolean shorthand can be used to turn off (false).

src/Cli.Tests/UpdateEntityTests.cs
souvikghosh04: This has been addressed. All update tests in UpdateEntityTests.cs now start with existing MCP configurations and test actual modification scenarios.

For example, TestUpdateTableEntityWithMcpDmlTools (Lines 1225-1284):

Tests both [DataRow("true", "false")] and [DataRow("false", "true")]
Initial config includes the existing MCP value: "mcp":  + initialMcpDmlTools
Then applies the update to transition from enabled→disabled or disabled→enabled

src/Config/ObjectModel/Entity.cs
souvikghosh04: `EntityMcpOptions` requires a custom converter because it accepts two different JSON formats: boolean shorthand ("mcp": true/false) or object format ("mcp": {"dml-tools": bool, "custom-tool": bool}. This is needed.

src/Config/ObjectModel/EntityMcpOptions.cs
souvikghosh04: This is set to true. The behavior is dml-tools will be true by default and custom-tool will be false.

src/Cli.Tests/Snapshots/AddEntityTests.AddStoredProcedureWithBothMcpProperties.verified.txt
souvikghosh04: This is fixed by adding these to ignore list in ModuleInitializer.cs.

src/Cli.Tests/ModuleInitializer.cs
```diff
@@ -65,6 +65,10 @@ public static void Init()
         VerifierSettings.IgnoreMember<Entity>(entity => entity.IsLinkingEntity);
         // Ignore the UserProvidedTtlOptions. They aren't serialized to our config file, enforced by EntityCacheOptionsConverter.
         VerifierSettings.IgnoreMember<EntityCacheOptions>(cacheOptions => cacheOptions.UserProvidedTtlOptions);
+        // Ignore the UserProvidedCustomToolEnabled. They aren't serialized to our config file, enforced by EntityMcpOptionsConverterFactory.
```
RubenCerna2079, line 68: Quick question did you also add this to the `ModuleInitializer` file for the `Service.Test`? Since this is only for `Cli.Tests`

souvikghosh04: Added it for `Service.Test` as well.

RubenCerna2079 (approved):
LGTM! Thanks for the changes :)

PR #3004, opened 2025-12-04
Suppress SM05137 in deserialization variable replacement settings 

aaronburtle:
## Why make this change?

Silences CodeQL flag.

## What is this change?

Adds the suppression language to the usage of `DefaultAzureCredential()`

## How was this tested?

Against usual test suite, no real code change, just a comment.

## Sample Request(s)

N/A

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/Core/McpProtocolDefaults.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Core/McpProtocolDefaults.cs
@@ -1,0 +1,30 @@
+using Microsoft.Extensions.Configuration;
+
+namespace Azure.DataApiBuilder.Mcp.Core
+{
+    /// <summary>
+    /// Centralized defaults and configuration keys for MCP protocol settings.
+    /// </summary>
+    public static class McpProtocolDefaults
+    {
+        /// <summary>
+        /// Default MCP protocol version advertised when no configuration override is provided.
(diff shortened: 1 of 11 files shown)


src/Config/DeserializationVariableReplacementSettings.cs
```diff
@@ -259,7 +259,7 @@ private static SecretClient CreateSecretClient(AzureKeyVaultOptions options)
                 clientOptions.Retry.NetworkTimeout = TimeSpan.FromSeconds(options.RetryPolicy.NetworkTimeoutSeconds ?? AKVRetryPolicyOptions.DEFAULT_NETWORK_TIMEOUT_SECONDS);
             }
 
-            return new SecretClient(new Uri(options.Endpoint), new DefaultAzureCredential(), clientOptions);
+            return new SecretClient(new Uri(options.Endpoint), new DefaultAzureCredential(), clientOptions); // CodeQL [SM05137] DefaultAzureCredential will use Managed Identity if available or fallback to default.
```
RubenCerna2079, line 262: Wouldn't this cause a security issue to come up?

aaronburtle: This is how you are supposed to suppress the issue. There are separate instructions for mitigating, which we may want to come back to in the future, but for now these are all suppressed.

Issue #2010, opened 2024-02-05
[Bug]: MSSQL pipeline Flakey


seantleonard:
### What happened?

MSSQL pipeline is flakely , typically in Linux tests, but sometimes windows, where all integration tests pass, but there is either a timeout or connection forcibly closed error. There may be a memory leak or inefficient use of resources occuring in test fixture which should be identified and remediated.

Why this is impactful:
1. Affects iteration velocity -> may allow quicker pace of merges because pipeline execution speed may be increased due to this and multiple runs don't need to be triggered

### Version

latest `main`

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

REST

### Relevant log output

```Text
Windows:

Passed!  - Failed:     0, Passed:  1010, Skipped:     2, Total:  1012, Duration: 11 m 55 sTest Run Aborted with error System.Exception: One or more errors occurred.
 ---> System.Exception: Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host..
 ---> System.Exception: An existing connection was forcibly closed by the remote host.
   at System.Net.Sockets.NetworkStream.Read(Span`1 buffer)
   --- End of inner exception stack trace ---
   at System.Net.Sockets.NetworkStream.Read(Span`1 buffer)
   at System.Net.Sockets.NetworkStream.ReadByte()
   at System.IO.BinaryReader.Read7BitEncodedInt()
   at System.IO.BinaryReader.ReadString()
   at Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.LengthPrefixCommunicationChannel.NotifyDataAvailable()
   at Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.TcpClientExtensions.MessageLoopAsync(TcpClient client, ICommunicationChannel channel, Action`1 errorHandler, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---.
 - Azure.DataApiBuilder.Service.Tests.dll (net6.0)
##[error]Error: The process 'C:\hostedtoolcache\windows\dotnet\dotnet.exe' failed with exit code 1
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

seantleonard:
[current] I'm adding some more diagnostics (via dotnet test --diagto try and identify the issue in addition to having :
[...]

PR #3018, opened 2025-12-15
Hot Reload test fix

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2010

Note that hot-reload of the graphQL schema is bugged currently, and tests are still ignored until this issue is resolved: https://github.com/Azure/data-api-builder/issues/3019

## What is this change?

As documented in the issue above, a number of hot-reload tests were failing due to relying on a simple timeout to allow the hot-reload to complete within tests. This creates a number of potential problems, including potential race conditions within the tests. We replace that strategy with a waiting process and change from syn to async in a number of functions. We also cleanup the config creation in the tests so that the new properties added since these tests were ignored fit into the flow and expectations of the tests.

## How was this tested?

The previously ignored tests are now run and have the ignore tag removed.

## Sample Request(s)

N/A

### DIFF
--- a/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
+++ b/src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
@@ -31,7 +31,7 @@ public class ConfigurationHotReloadTests
     private const string GQL_QUERY_NAME = "books";
     private const string HOT_RELOAD_SUCCESS_MESSAGE = "Validated hot-reloaded configuration file";
     private const string HOT_RELOAD_FAILURE_MESSAGE = "Unable to hot reload configuration file due to";
-    private const int HOT_RELOAD_TIMEOUT_SECONDS = 120; // Increased timeout for CI/CD environments where file watchers can be slow
+    private const int HOT_RELOAD_TIMEOUT_SECONDS = 120;
 
     private const string GQL_QUERY = @"{
                 books(first: 100) {
@@ -195,7 +195,6 @@ public class ConfigurationHotReloadTests
         // Arrange
         GenerateConfigFile(connectionString: $"{ConfigurationTests.GetConnectionStringFromEnvironmentConfig(TestCategory.MSSQL).Replace("\\", "\\\\")}");
 
-        // Add retry logic for test server initialization in CI/CD environments
         int maxRetries = 3;
         int retryDelayMs = 2000;
         Exception lastException = null;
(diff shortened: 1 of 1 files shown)


src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -376,10 +412,13 @@ public async Task HotReloadEntityGQLEnabledFlag()
     /// </summary>
     [TestCategory(MSSQL_ENVIRONMENT)]
     [TestMethod]
-    [Ignore]
+    [Ignore] // This test requires GraphQL schema reload
```
Aniruddh25: the PR description should mention why this test is still being ignored. I thought we successfully fixed hot reload for GraphQL tests.

aaronburtle: The tests which require SQL schema eviction will need to wait on the fix for this issue: https://github.com/Azure/data-api-builder/issues/3019

seems that GQL is not properly invalidating the schema cache when we evict the schema.

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -337,10 +365,13 @@ public async Task HotReloadConfigRuntimeGQLEnabledEndToEndTest()
     /// </summary>
     [TestCategory(MSSQL_ENVIRONMENT)]
     [TestMethod("Hot-reload gql disabled at entity level.")]
-    [Ignore]
+    [Ignore] // This test requires GraphQL schema reload
```
RubenCerna2079: I think it is a good idea to add the number of the issue that is holding this change back

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -746,23 +807,39 @@ public void HotReloadValidationFail()
...
+        _writer = new StringWriter();
+        Console.SetOut(_writer);
+
         RuntimeConfig lkgRuntimeConfig = _configProvider.GetConfig();
         Assert.IsNotNull(lkgRuntimeConfig);
 
+        // Capture properties to verify config hasn't changed
+        bool originalRestEnabled = lkgRuntimeConfig.Runtime.Rest.Enabled;
+        bool originalGraphQLEnabled = lkgRuntimeConfig.Runtime.GraphQL.Enabled;
+
```
RubenCerna2079, line 901: Why do we not check for the MCP enabled?

aaronburtle: because we don't attempt to change them in the modified config.

Issue #3026, opened 2025-12-19
[Bug]: KeyNotFoundException when querying multiple nested relationships


magnh:
## Problem

When querying multiple nested relationships in GraphQL, a `KeyNotFoundException` is thrown. This seems to occur because metadata keys for sibling relationships collide.

**The following queries will pass**

```graphql
query {
  books {
    items {
      id
      category {
        name
      }
      author {
        name
      }
    }
  }
}
```

```graphql
query {
  books {
    items {
      id
      category {
        name
        parent {
          name
        }
      }
    }
  }
}
```

```graphql
query {
  books {
    items {
      id
      author {
        name
        publisher {
          name
        }
      }
    }
  }
}
```

**When combining multiple nested relationships, the following query fails**

```graphql
query {
  books {
    items {
      id
      category {
        name
        parent {
          name
        }
      }
      author {
        name
        publisher {
          name
        }
      }
    }
  }
}
```

## Root Cause

In `ExecutionHelper.cs`, the `SetNewMetadataChildren` method creates metadata dictionary keys using only the GraphQL type name (e.g., `"Category"`). When two sibling relationships resolve to the same type, the second one overwrites the first in the dictionary.

Later, `GetMetadataObjectField` tries to look up metadata by type name but finds the wrong relationship's metadata, causing the mismatch.

## Proposed Solution

1. Create unique metadata keys by incorporating the **relationship path** from the root, not just the type name
2. Add a helper method `GetRelationshipPathSuffix` that builds a unique suffix from the selection path (e.g., `"category"` or `"author"`)
3. Modify `SetNewMetadataChildren` to append this suffix to metadata keys
4. Modify `GetMetadataObjectField` to use the same suffix logic when looking up metadata

See pull request: #3025 

This should ensure each relationship gets its own unique metadata entry regardless of the target type.

### Version

Latest (main-branch)

### What database are you using?

Azure SQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

GraphQL

### Relevant log output

```Text
System.Collections.Generic.KeyNotFoundException: The given key 'parent' was not present in the dictionary.

System.Collections.Generic.KeyNotFoundException: The given key 'publisher' was not present in the dictionary.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

fmrogers:
👍

magnh:
Perfect, thank you @anushakolan! Approximately when would this this be released?

Issue #2374, opened 2024-09-09
[Bug]: Nested entities in GraphQL query may cause System.Collections.Generic.KeyNotFoundException: The given key 'key' was not present in the dictionary.


marcusrangell:
### What happened?

**This is the scenario.**
A SQL database with a Person table.
A Person may have 0:M Addresses. Each Address has an AddressType.
A Person may have 0:M PhoneNumbers. Each PhoneNumber has a PhoneNumberType.

When querying using GraphQL, we can query for Person and related Addresses and PhoneNumbers. However, we can only add the related AddressType/PhoneNumberType to the _first_ related entity in the query. Retrieving a Person with Addresses and PhoneNumbers as well as the related AddressType and PhoneNumberType entities will fail, **nullifying the latter** of the two.

Here is a short example that shows the issue.

<details><summary>T-SQL statements to create tables and example data</summary>
<p>

```
-- Table to store person data
CREATE TABLE Person (
    PersonID INT IDENTITY(1,1) PRIMARY KEY,
    FirstName NVARCHAR(50),
    LastName NVARCHAR(50)
);

-- Table to store address types (e.g., Home, Work, etc.)
CREATE TABLE AddressType (
    AddressTypeID INT IDENTITY(1,1) PRIMARY KEY,
    TypeName NVARCHAR(50) NOT NULL
);

-- Table to store phone number types (e.g., Mobile, Home, Work)
CREATE TABLE PhoneNumberType (
    PhoneNumberTypeID INT IDENTITY(1,1) PRIMARY KEY,
    TypeName NVARCHAR(50) NOT NULL
);

-- Table to store addresses, which reference a person and an address type
CREATE TABLE Address (
    AddressID INT IDENTITY(1,1) PRIMARY KEY,
    PersonID INT,
    AddressTypeID INT,
    StreetAddress NVARCHAR(100),
    City NVARCHAR(50),
    State NVARCHAR(50),
    PostalCode NVARCHAR(10),
    CONSTRAINT FK_Person_Address FOREIGN KEY (PersonID) REFERENCES Person(PersonID) ON DELETE CASCADE,
    CONSTRAINT FK_AddressType_Address FOREIGN KEY (AddressTypeID) REFERENCES AddressType(AddressTypeID)
);

-- Table to store phone numbers, which reference a person and a phone number type
CREATE TABLE PhoneNumber (
    PhoneNumberID INT IDENTITY(1,1) PRIMARY KEY,
    PersonID INT,
[...]

PR #3029, opened 2025-12-20
Bug fix for pagination nested entities resulting key not found error.

anushakolan:
## Why make this change?
We are addressing 2 related issues in this PR:

Issue #2374 – Nested sibling relationships under books (websiteplacement, reviews, authors)

**Problem**: A nested query on books where a parent has multiple sibling relationships (for example, websiteplacement, reviews, and authors) could throw a `KeyNotFoundException` when RBAC or shape changes were involved. Pagination metadata was stored using only the root and the depth in the path, so different sibling relationships at the same depth could overwrite each other or look up the wrong entry.

**Solution**: We now key pagination metadata by both depth and the full relationship path (for example, “books → items → reviews” vs “books → items → authors”), so each sibling branch gets its own unique entry. Reads use the same full-path key, and if metadata for a branch is missing, we return an “empty” `PaginationMetadata` instead of throwing. This prevents collisions between sibling relationships and avoids runtime errors when a particular branch has no metadata.

Issue #3026 – Person's graph (AddressType / PhoneNumberType)

**Problem**: In the persons graph, a query selecting persons → addresses.items.AddressType and persons → phoneNumbers.items.PhoneNumberType could also throw a `KeyNotFoundException`. In some cases (for example, when RBAC removes a relationship or when that relationship is not paginated at all), there is legitimately no pagination metadata for that nested field, but the code assumed it always existed and indexed into the dictionary directly.

**Solution**: Metadata handling is now defensive in two places:
In the GraphQL execution helper, metadata lookups for object and list fields use safe TryGet-style access; if an entry isn’t present, we fall back to an empty PaginationMetadata instead of failing.
In the SQL query engine’s object resolver, we first check whether there is a subquery metadata entry for the field. If there isn’t, we treat the field as non‑paginated and return the JSON as-is rather than throwing.

Together, these changes fix both issues by
 (a) using full path-based keys, so sibling branches don’t conflict,
 (b) treating missing metadata as “no pagination here” rather than as a fatal error.

## What is this change?

1. In `SqlQueryEngine.ResolveObject`, instead of always doing `parentMetadata.Subqueries[fieldName]` (which crashed when RBAC caused that entry to be missing), it now uses `TryGetValue` and:
   - If metadata exists and `IsPaginated` is true -> wrap the JSON as a pagination connection.
   - If metadata is missing -> just return the JSON as-is (no exception).
2. Introduced `GetRelationshipPathSuffix(HotChocolate.Path path)` to build a relationship path suffix like:
    - `rel1` for `/entity/items[0]/rel1`
    - `rel1::nested` for `/entity/items[0]/rel1/nested`
3. `SetNewMetadataChildren`, now stores child metadata under keys of the form
     - `root_PURE_RESOLVER_CTX::depth::relationshipPath`, ensuring siblings at the same depth get distinct entries.
5. `GetMetadata` (used for list items fields):
    - For `Selection.ResponseName == "items"` and non-root paths, now looks up:
       a.  `GetMetadataKey(context.Path) + "::" + context.Path.Parent.Depth()`
 plus the relationship suffix from `GetRelationshipPathSuffix(context.Path.Parent)`.
       b. Uses `ContextData.TryGetValue(...)` and falls back to `PaginationMetadata.MakeEmptyPaginationMetadata()` when metadata is missing (e.g. Cosmos, pruned relationships).
6. `GetMetadataObjectField` (used for object fields like addresses, AddressType, PhoneNumberType):
Updated all branches (indexer, nested non-root, root) to:
   - Append the relationship suffix to the base key (so keys align with `SetNewMetadataChildren`).
   - Use `ContextData.TryGetValue(...)` instead of direct indexing, return `PaginationMetadata.MakeEmptyPaginationMetadata()` when no metadata exists, instead of throwing.
7. Added a new test case in `MsSqlGraphQLQueryTests`, an integration test which queries books with multiple sibling nested relationships (websiteplacement, reviews, authors) under the authenticated role to:
   - Assert no KeyNotFoundException,
   - Verify all nested branches return data.

## How was this tested?
Tested both manually and added an integration test (NestedReviewsConnection_WithSiblings_PaginatesMoreThanHundredItems).

Manually if we run this query without the bug fix:
`query {
  persons {
    items {
      PersonID
      FirstName
      LastName
      addresses {
        items {
          AddressID
          City
          AddressType {
            AddressTypeID
            TypeName
          }
        }
      }
      phoneNumbers {
        items {
          PhoneNumberID
          PhoneNumber
          PhoneNumberType {
            PhoneNumberTypeID
            TypeName
          }
        }
      }
    }
  }
}`

We get the following response:

`{
  "errors": [
    {
      "message": "The given key 'AddressType' was not present in the dictionary.",
      "locations": [
        {
          "line": 11,
          "column": 11
        }
      ],
      "path": [
        "persons",
        "items",
        0,
        "addresses",
        "items",
        1,
        "AddressType"
      ]
    },
    {
      "message": "The given key 'AddressType' was not present in the dictionary.",
      "locations": [
        {
          "line": 11,
          "column": 11
        }
      ],
      "path": [
        "persons",
        "items",
        0,
        "addresses",
        "items",
        0,
        "AddressType"
      ]
    },
    {
      "message": "The given key 'AddressType' was not present in the dictionary.",
      "locations": [
        {
          "line": 11,
          "column": 11
        }
      ],
      "path": [
        "persons",
        "items",
        1,
        "addresses",
        "items",
        0,
        "AddressType"
      ]
    }
  ],
  "data": {
    "persons": {
      "items": [
        {
          "PersonID": 1,
          "FirstName": "John",
          "LastName": "Doe",
          "addresses": {
            "items": [
              {
                "AddressID": 1,
                "City": "New York",
                "AddressType": null
              },
              {
                "AddressID": 2,
                "City": "New York",
                "AddressType": null
              }
            ]
          },
          "phoneNumbers": {
            "items": [
              {
                "PhoneNumberID": 1,
                "PhoneNumber": "123-456-7890",
                "PhoneNumberType": {
                  "PhoneNumberTypeID": 1,
                  "TypeName": "Mobile"
                }
              },
              {
                "PhoneNumberID": 2,
                "PhoneNumber": "111-222-3333",
                "PhoneNumberType": {
                  "PhoneNumberTypeID": 3,
                  "TypeName": "Work"
                }
              }
            ]
          }
        },
        {
          "PersonID": 2,
          "FirstName": "Jane",
          "LastName": "Smith",
          "addresses": {
            "items": [
              {
                "AddressID": 3,
                "City": "Los Angeles",
                "AddressType": null
              }
            ]
          },
          "phoneNumbers": {
            "items": [
              {
                "PhoneNumberID": 3,
                "PhoneNumber": "987-654-3210",
                "PhoneNumberType": {
                  "PhoneNumberTypeID": 2,
                  "TypeName": "Home"
                }
              }
            ]
          }
        }
      ]
    }
  }
}`

After the bug fix, we get,

`{
  "data": {
    "persons": {
      "items": [
        {
          "PersonID": 1,
          "FirstName": "John",
          "LastName": "Doe",
          "addresses": {
            "items": [
              {
                "AddressID": 1,
                "City": "New York",
                "AddressType": {
                  "AddressTypeID": 1,
                  "TypeName": "Home"
                }
              },
              {
                "AddressID": 2,
                "City": "New York",
                "AddressType": {
                  "AddressTypeID": 2,
                  "TypeName": "Work"
                }
              }
            ]
          },
          "phoneNumbers": {
            "items": [
              {
                "PhoneNumberID": 1,
                "PhoneNumber": "123-456-7890",
                "PhoneNumberType": {
                  "PhoneNumberTypeID": 1,
                  "TypeName": "Mobile"
                }
              },
              {
                "PhoneNumberID": 2,
                "PhoneNumber": "111-222-3333",
                "PhoneNumberType": {
                  "PhoneNumberTypeID": 3,
                  "TypeName": "Work"
                }
              }
            ]
          }
        },
        {
          "PersonID": 2,
          "FirstName": "Jane",
          "LastName": "Smith",
          "addresses": {
            "items": [
              {
                "AddressID": 3,
                "City": "Los Angeles",
                "AddressType": {
                  "AddressTypeID": 1,
                  "TypeName": "Home"
                }
              }
            ]
          },
          "phoneNumbers": {
            "items": [
              {
                "PhoneNumberID": 3,
                "PhoneNumber": "987-654-3210",
                "PhoneNumberType": {
                  "PhoneNumberTypeID": 2,
                  "TypeName": "Home"
                }
              }
            ]
          }
        }
      ]
    }
  }
}`

## Sample Request(s)


src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -229,6 +229,68 @@ FOR JSON PATH
...
+                response.Contains("KeyNotFoundException", StringComparison.OrdinalIgnoreCase) ||
+                response.Contains("The given key", StringComparison.OrdinalIgnoreCase),
+                "GraphQL response should not contain KeyNotFoundException when resolving nested sibling relationships.");
+
+            // 2) Ensure nested branches are actually materialized
+            Assert.IsTrue(
+                response.Contains("\"websiteplacement\"", StringComparison.OrdinalIgnoreCase) &&
+                response.Contains("\"reviews\"", StringComparison.OrdinalIgnoreCase) &&
+                response.Contains("\"authors\"", StringComparison.OrdinalIgnoreCase),
+                "Expected nested websiteplacement, reviews and authors data to be present in the GraphQL response.");
```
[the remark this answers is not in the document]
anushakolan: @copilot add a new commit to apply changes based on [this feedback](https://github.com/Azure/data-api-builder/pull/3029#discussion_r2636653194)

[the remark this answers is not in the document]
anushakolan: @copilot adda new commit to apply changes based on [this feedback](https://github.com/Azure/data-api-builder/pull/3029#discussion_r2636653194)

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -224,12 +224,18 @@ public JsonElement ResolveObject(JsonElement element, ObjectField fieldSchema, r
...
                 }
 
-                PaginationMetadata currentMetadata = parentMetadata.Subqueries[fieldSchema.Name];
-                metadata = currentMetadata;
-
-                if (currentMetadata.IsPaginated)
+                // In some scenarios (for example when RBAC removes a relationship
+                // or when multiple sibling nested entities are present), we may not
+                // have pagination metadata for the current field. In those cases we
+                // should simply return the element as-is instead of throwing.
```
Aniruddh25, line 230: If we return the element as-is, doesnt it indicate that we are missing the metadata? Wouldnt this impact the pagination connection element? if we dont create the connection element, how is the query still accurate? can we still paginate?

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -229,6 +229,71 @@ FOR JSON PATH
...
+            string graphQLQueryName = "books";
+            string graphQLQuery = @"query {
+                                        books(first: 2) {
+                                            items {
+                                            id
+                                            title
+                                            websiteplacement {
+                                                price
+                                            }
+                                            reviews {
```
Aniruddh25: Can you please test this by adding more than 100 reviews and see if we can still successfully paginate? First page should have 100 reviews and it should have an `endCursor` clause with the opaque id of the last review on page 1.

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -229,6 +229,71 @@ FOR JSON PATH
...
+                Assert.IsFalse(
+                    errorsText.Contains("KeyNotFoundException", StringComparison.OrdinalIgnoreCase) ||
+                    errorsText.Contains("The given key", StringComparison.OrdinalIgnoreCase),
+                    "GraphQL response should not contain KeyNotFoundException when resolving nested sibling relationships.");
+            }
+
+            string response = actual.ToString();
+            // 2) Ensure nested branches are actually materialized
+            Assert.IsTrue(
+                response.Contains("\"websiteplacement\"", StringComparison.OrdinalIgnoreCase) &&
```
Aniruddh25: Just verifying that the branches contain these nodes isnt sufficient. We should compare it with the expected JSON results. By comparing it with results of an equivalent SQL JSON query.

src/Core/Resolvers/SqlQueryEngine.cs
anushakolan: 1. When a field is modeled as a connection (needs items, cursors, etc.) the metadata builder always creates a `PaginationMetadata` entry for that field and sets `IsPaginated = true`. In those cases, `TryGetValue` will succeed and we still go into `CreatePaginationConnectionFromJsonElement(...) `exactly as before. Pagination behavior for those fields is unchanged.
2. If TryGetValue fails, it means the metadata tree does not have an entry for this field. That only happens when:
    - The field was not planned as paginated at all (it’s just a plain nested object/JSON subtree), or
    - Some upstream condition (like RBAC removing a relationship or trimming its selection) means we never built a subquery/metadata node for it, so there is no offset/limit/cursor information for that branch.
3. In those “no metadata” cases we cannot correctly construct a connection object anyway: we don’t know total count, whether there is a next page, and the original limit. Trying to index into `Subqueries[fieldName]` and build a connection would just throw, which is the bug we were seeing.
4. By returning the element as‑is, we’re effectively saying: “treat this nested piece as plain JSON, not a connection”. The rest of the query (including the top‑level books connection and any other fields that do have metadata) remains fully paginated and accurate. The only thing we lose is connection‑style pagination on that particular branch, in the rare cases where its metadata is missing.

anushakolan:
> In the description, `After the bug fix, we get,` - we still see an error message. not the response after fixing the bug,

Fixed

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -285,13 +287,182 @@ public async Task NestedSiblingRelationshipsWithRbac_DoNotThrowAndMaterialize()
...
+                    dataSourceName: string.Empty,
+                    parameters: null,
+                    dataReaderHandler: null);
+
+                string graphQLQueryName = "books";
+                string graphQLQuery = @"query {
+                    books(filter: { id: { eq: 1 } }) {
+                        items {
+                            id
+                            reviews(first: 100) {
```
Aniruddh25, line 288: Did you double check if we already have a test for pagination where the data is more than 100? My review comment was to club this scenario with the nested sibling usecase for which the bug has been filed. This test doesnt seem to combine the two.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -224,12 +224,18 @@ public JsonElement ResolveObject(JsonElement element, ObjectField fieldSchema, r
...
+                if (parentMetadata.Subqueries.TryGetValue(fieldSchema.Name, out PaginationMetadata? currentMetadata))
                 {
-                    return SqlPaginationUtil.CreatePaginationConnectionFromJsonElement(element, currentMetadata);
+                    metadata = currentMetadata;
```
RubenCerna2079, line 233: Do we do something with this `metadata` variable?

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -229,6 +231,240 @@ FOR JSON PATH
...
+                                                id
+                                                content
+                                                }
+                                            }
+                                            authors {
+                                                items {
+                                                name
+                                                }
+                                            }
+                                            }
```
RubenCerna2079: nit:
```suggestion
                                            items {
                                                id
                                                title
                                                websiteplacement {
                                                    price
                                                }
                                                reviews {
                                                    items {
                                                    id
                                                    content
                                                    }
                                                }
                                                authors {
                                                    items {
                                                    name
                                                    }
                                                }
                                            }
```

src/Core/Resolvers/SqlQueryEngine.cs
anushakolan: Yes, it is used as a reference variable in the function signature.

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
anushakolan: Addressed

src/Core/Services/ExecutionHelper.cs
```diff
@@ -614,6 +671,54 @@ private static string GetMetadataKey(HotChocolate.Path path)
...
+            {
+                if (current is IndexerPathSegment)
+                {
+                    // We've reached items[n], stop here
+                    break;
+                }
+
+                if (current is NamePathSegment nameSegment)
+                {
+                    // Don't include "items" in the path suffix
```
Aniruddh25: if we would break when its an IndexerPathSegment, would we ever reach here for items? Do we need to check again for `nameSegment.Name != QueryBuilder.PAGINATION_FIELD_NAME`?

src/Core/Services/ExecutionHelper.cs
```diff
@@ -657,6 +762,16 @@ private static void SetNewMetadataChildren(IResolverContext context, IMetadata?
             // is 3 because depth is 0-indexed.
             string contextKey = GetMetadataKey(context.Path) + "::" + context.Path.Depth();
```
Aniruddh25: contextKey should be a StringBuilder instead of string. This is because we are concatenating to it, and its more performant with StringBuilder. With a string, it will create multiple copies of strings.

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
Aniruddh25: this doesn't seem to have been resolved. 

`expected` results should be compared with `actual`. Calculate the expected results by executing a SQL query against the db. write the expected SQL query in code - `SELECT .. FROM AS JSON`

anushakolan: The change got reverted in a recent fix, let me add it back.

src/Service.Tests/SqlTests/GraphQLQueryTests/MsSqlGraphQLQueryTests.cs
```diff
@@ -344,11 +344,83 @@ await _queryExecutor.ExecuteQueryAsync<object>(
...
+                                    SELECT [a].[name] AS [name]
+                                    FROM [dbo].[authors] AS [a]
+                                    INNER JOIN [dbo].[book_author_link] AS [bal]
+                                        ON [bal].[author_id] = [a].[id]
+                                    WHERE [bal].[book_id] = [b].[id]
+                                    ORDER BY [a].[id] ASC
+                                    FOR JSON PATH, INCLUDE_NULL_VALUES
+                                )) AS [items]
+                            FOR JSON PATH, INCLUDE_NULL_VALUES, WITHOUT_ARRAY_WRAPPER
+                        )) AS [authors],
```
Aniruddh25: if the SQL statement that you craft here, is reordered to have `reviews` before `authors`, you dont need to project the GraphQL result to the same shape as the SQL JSON since they would essentially be the same. 

You are in control of how you write the SQL query so the expected json should match the shape of the actual GraphQL query written in the test. 

Please avoid the fullTree creation/serialization but rely on `actual` result obtained already to perform the comparison between expected and actual.

Aniruddh25 (changes requested):
Please link the issues that this PR will resolve in the Development section so that they are closed automatically. We should strive to maintain issue closure hygiene as much as possible, and automate it to avoid manual cleanup of our issue backlog. 

image

src/Core/Services/ExecutionHelper.cs
```diff
@@ -592,12 +627,35 @@ private static IMetadata GetMetadataObjectField(IResolverContext context)
                 // e.g. metadata for index 4 will not exist. only 3.
                 // Depth: /  0   / 1  /   2    /   3      /   4
                 // Path:  /books/items/items[0]/publishers/books
+                // 
+                // To handle arbitrary nesting depths with sibling relationships, we need to include
+                // the relationship field path in the key. For example:
+                // - /entity/items[0]/rel1/nested uses key ::3::rel1
+                // - /entity/items[0]/rel2/nested uses key ::3::rel2
+                // - /entity/items[0]/rel1/nested/deeper uses key ::4::rel1
```
RubenCerna2079: Why wouldn't this example use the key `::4::rel1::nested` similar to the one below?

anushakolan: It is a typo, fixed it.

RubenCerna2079 (approved):

PR #3035, opened 2025-12-30
[MCP] Centralize JSON‑RPC error handling, standardize MCP response envelope, and add default anonymous role for stdio mode.

anushakolan:
## Why make this change?

This is a follow up PR for, https://github.com/Azure/data-api-builder/pull/2983 to address some of the newer comments.

## What is this change?
1. Centralized JSON‑RPC error handling: all standard JSON‑RPC error codes are now defined in one place and used consistently throughout the MCP server (for things like parse errors, invalid requests, unknown methods, invalid parameters, and internal errors).
2. Standardized how responses are written: both successful results and errors now go through shared helpers, so every MCP response has a consistent JSON‑RPC envelope and is easy to change or audit.
3. Added an implicit default role to be `anonymous` when no role is provided in `mcp-stdio` mode.

## How was this tested?
Tested manually by calling all the CRUD tools and execute tool.

Added an MCP server in VS Code in `C:\DAB\data-api-builder\.vscode\mcp.json`

`"dab-with-exe": {
      "command": "C:\\DAB\\data-api-builder\\out\\publish\\Debug\\net8.0\\win-x64\\dab\\Microsoft.DataApiBuilder.exe",
      "args": ["start", "--mcp-stdio", "--config",
				"C:\\DAB\\data-api-builder\\dab-config.json"],
      "env": {
        "DAB_ENVIRONMENT": "Development"
      }
    }`

1. When calling tools with `Authenticated` all tools pass.
2. When calling tools with `Anonymous`, all tools except `describe-entities` fail with permission denied error.
5. When calling tools without any role parameter, all tools except `describe-entities` fail with permission denied error.

### DIFF
--- a/src/Service.Tests/SqlTests/SqlTestBase.cs
+++ b/src/Service.Tests/SqlTests/SqlTestBase.cs
@@ -514,7 +514,7 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests
                 request.Headers.Add(AuthorizationResolver.CLIENT_ROLE_HEADER, clientRoleHeader);
 
                 // Detect runtime auth provider once per call
-                var configProvider = _application.Services.GetRequiredService<RuntimeConfigProvider>();
+                RuntimeConfigProvider configProvider = _application.Services.GetRequiredService<RuntimeConfigProvider>();
                 string provider = configProvider.GetConfig().Runtime.Host.Authentication.Provider;
 
                 if (string.Equals(provider, nameof(EasyAuthType.AppService), StringComparison.OrdinalIgnoreCase))


src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -1,5 +1,6 @@
 using System;
```
Aniruddh25, line 4: Please add the copyright header to the file.

src/Azure.DataApiBuilder.Mcp/Core/JsonRpcErrorCodes.cs
```diff
@@ -0,0 +1,36 @@
+namespace Azure.DataApiBuilder.Mcp.Core
+{
+    /// <summary>
+    /// JSON-RPC 2.0 standard error codes used by the MCP stdio server.
+    /// These values come from the JSON-RPC 2.0 specification and are shared
+    /// so they are not hard-coded throughout the codebase.
+    /// </summary>
+    internal static class JsonRpcErrorCodes
```
souvikghosh04: please consider renaming this to something like McpStdioJsonRpcErrorCodes or something to reflect its related to only MCP. also, we should move this inside "../src/Azure.DataApiBuilder.Mcp/Model" where we already have other error codes and enums, specifically for MCP.

src/Azure.DataApiBuilder.Mcp/Core/JsonRpcErrorCodes.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// The JSON sent is not a valid Request object.
+        /// </summary>
+        public const int INVALIDREQUEST = -32600;
+
+        /// <summary>
+        /// The method does not exist / is not available.
+        /// </summary>
+        public const int METHODNOTFOUND = -32601;
```
souvikghosh04: please consider using existing format "SCREAMING_SNAKE_CASE or UPPER_SNAKE_CASE" for all constant variables. refer- ../src/Config/HealthCheck/HealthCheckConstants.cs

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -158,32 +158,22 @@ public async Task RunAsync(CancellationToken cancellationToken)
...
-                        name = "Data API Builder",
-                        version = "1.0.0"
-                    }
+                    tools = new { listChanged = true },
+                    logging = new { }
+                },
+                serverInfo = new
+                {
+                    name = "SQL MCP Server",
+                    version = "1.0.0"
```
souvikghosh04: can you please also update the name and version in ../src/Azure.DataApiBuilder.Mcp/Core/McpServerConfiguration.cs? for consistency, try to refactor it and use the same name and version through a configurable value or use the DAB version as the version will keep changing.
tagging @Aniruddh25 to confirm if we can use the DAB version itself or have a separate versioning for MCP server.

src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -15,9 +19,9 @@ internal static class McpStdioHelper
         /// Determines if MCP stdio mode should be run based on command line arguments.
         /// </summary>
         /// <param name="args"> The command line arguments.</param>
-        /// <param name="mcpRole"> The role for MCP stdio mode, if specified.</param>
-        /// <returns></returns>
-        public static bool ShouldRunMcpStdio(string[] args, out string? mcpRole)
+        /// <param name="mcpRole"> The role for MCP stdio mode. When this method returns true, the role is guaranteed to be non-null.</param>
```
souvikghosh04: nit: should mention mcpRole defaults to anonymous

src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -43,6 +47,11 @@ public static bool ShouldRunMcpStdio(string[] args, out string? mcpRole)
                 }
             }
 
+            // Ensure that when MCP stdio is enabled, mcpRole is always non-null.
+            // This matches the NotNullWhen(true) contract and avoids nullable warnings
+            // for callers while still allowing an implicit default when no role is provided.
+            mcpRole ??= "anonymous";
```
souvikghosh04, line 53: is this necessary? can we instead just initialize as `mcpRole = "anonymous";` instead of `mcpRole = null;` in the beginning at line: 26?

Aniruddh25 (approved):

souvikghosh04 (changes requested):
Added some comments for appropriate way of naming and versioning of MCP server and some naming/structuring conventions.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
Aniruddh25: Its not a different product, we can use the already configurable DAB version, please use ProductInfo.GetVersion function to get this version.

src/Service/Utilities/McpStdioHelper.cs
anushakolan: With the current code,

When --mcp-stdio is absent: method returns false and mcpRole is null.
When --mcp-stdio is present and no role: arg: method returns true and mcpRole becomes "anonymous" via mcpRole ??= "anonymous";.

If we change the start to:
mcpRole = "anonymous";

and remove the later mcpRole ??= "anonymous";, then:

When --mcp-stdio is absent: method returns false but mcpRole will be "anonymous" (non-null).
When --mcp-stdio is present and no role: arg: still returns true, mcpRole is "anonymous" – same as now.

If you want the out param to be meaningful only when the method returns true, the current pattern (mcpRole = null; then mcpRole ??= "anonymous"; after the runMcpStdio check) is clearer and preserves that behavior.

PR #3045, opened 2026-01-07
Fix serialization for StoredProcedureDefinition inheritance

Alekhya-Polavarapu:
## Why make this change?
- To apply correct serialization and deserialization logic for stored procedures. With the previous changes, serialization was not working correctly for the StoredProcedureDefinition type, which extends SourceDefinition. When the value type was passed explicitly for serialization, the parent type was used instead, causing some child-type properties to be omitted.
## What is this change?
Instead of manually specifying the value type during serialization, this change allows the library to infer the type automatically and perform the correct serialization.

## How was this tested?
- [x] Unit Tests

### DIFF
--- a/src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
+++ b/src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
@@ -34,7 +34,7 @@ namespace Azure.DataApiBuilder.Core.Services.MetadataProviders.Converters
 
                 DatabaseObject objA = (DatabaseObject)JsonSerializer.Deserialize(document, concreteType, options)!;
 
-                foreach (PropertyInfo prop in objA.GetType().GetProperties().Where(IsSourceDefinitionProperty))
+                foreach (PropertyInfo prop in objA.GetType().GetProperties().Where(IsSourceDefinitionOrDerivedClassProperty))
                 {
                     SourceDefinition? sourceDef = (SourceDefinition?)prop.GetValue(objA);
                     if (sourceDef is not null)


src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
```diff
@@ -73,25 +73,23 @@ public override void Write(Utf8JsonWriter writer, DatabaseObject value, JsonSeri
...
                 }
 
-                JsonSerializer.Serialize(writer, propVal, propType, options);
+                JsonSerializer.Serialize(writer, propVal, options);
             }
 
             writer.WriteEndObject();
         }
 
         private static bool IsSourceDefinitionProperty(PropertyInfo prop)
```
Aniruddh25: Shouldnt the name change to IsSourceDefinitionOrDerivedClassProperty?

src/Service.Tests/UnitTests/SerializationDeserializationTests.cs
```diff
@@ -322,14 +325,22 @@ public void TestDatabaseViewSerializationDeserialization_WithDollarColumn()
...
+
             // Test to catch if there is change in number of properties/fields
             // Note: On Addition of property make sure it is added in following object creation _databaseView and include in serialization
             // and deserialization test.
             int fields = typeof(DatabaseView).GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance).Length;
             Assert.AreEqual(fields, 6);
 
-            string serializedDatabaseView = JsonSerializer.Serialize(_databaseView, _options);
-            DatabaseView deserializedDatabaseView = JsonSerializer.Deserialize<DatabaseView>(serializedDatabaseView, _options)!;
+            string serializedDatabaseView = JsonSerializer.Serialize(dict, _options);
```
Aniruddh25, line 337: Does it matter if you serialize the dict or keep serializing the `_databaseView` to assert for the dollar sign in the column name?

Alekhya-Polavarapu: nothing specific, i have added to keep it consistent with DatabaseTable

src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
Alekhya-Polavarapu: done, updated

PR #3048, opened 2026-01-12
Implementation of Custom Tool in MCP

souvikghosh04:
## Why make this change?

- Closes on - #2877 

## What is this change?

This pull request introduces a new system for dynamically generating and registering custom MCP tools based on stored procedure entity configurations in the runtime configuration. The main changes are the implementation of the `DynamicCustomTool` class, a factory to create these tools from configuration, and the necessary service registration logic to ensure these custom tools are available at runtime.

**Dynamic custom MCP tool support:**

* Added the `DynamicCustomTool` class, which implements `IMcpTool` and provides logic for generating tool metadata, validating configuration, handling authorization, executing the underlying stored procedure, and formatting the response. This enables each stored procedure entity with `custom-tool` enabled to be exposed as a dedicated MCP tool.
* Introduced the `CustomMcpToolFactory` class, which scans the runtime configuration for stored procedure entities marked with `custom-tool` enabled and creates corresponding `DynamicCustomTool` instances.

**Dependency injection and service registration:**

* Updated the MCP server startup (`AddDabMcpServer`) to register custom tools generated from configuration by calling a new `RegisterCustomTools` method after auto-discovering static tools.
* Modified the `RegisterAllMcpTools` method to exclude `DynamicCustomTool` from auto-discovery (since these are created dynamically per configuration) and added the `RegisterCustomTools` method to register each generated custom tool as a singleton service.

## How was this tested?

- [x] Unit Tests
- [x] Manual Tests using Insomnia and VS code GHCP chat

## Sample Request(s)

1. List All Tools (also includes custom tool)
```
{
  "jsonrpc": "2.0",
  "method": "tools/list",
  "params": {},
  "id": 1
}
```

2. Get Books (no parameters)
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "get_books"
  },
  "id": 2
}
```

3. Get Book by ID
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "get_book",
    "arguments": {
      "id": 1
    }
  },
  "id": 3
}
```

4. Insert Book
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "insert_book",
    "arguments": {
      "title": "Test Book from MCP",
      "publisher_id": "1234"
    }
  },
  "id": 4
}
```

5. Count Books (no parameters)
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "count_books"
  },
  "id": 5
}
```

Error Scenarios
6. Missing Required Parameter
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "get_book"
  },
  "id": 6
}
```

7. Non-Existent Tool
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "non_existent_tool"
  },
  "id": 7
}
```

8. Invalid Foreign Key
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "insert_book",
    "arguments": {
      "title": "Test Book",
      "publisher_id": "999999"
    }
  },
  "id": 8
}
```

Edge Cases
9. SQL Injection Attempt
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "get_book",
    "arguments": {
      "id": "1; DROP TABLE books; --"
    }
  },
  "id": 9
}
```

11. Special Characters
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "insert_book",
    "arguments": {
      "title": "Test Book with 'quotes' and \"double quotes\" and <tags>",
      "publisher_id": "1234"
    }
  },
  "id": 10
}
```

12. Empty String
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "insert_book",
    "arguments": {
      "title": "",
      "publisher_id": "1234"
    }
  },
  "id": 11
}
```

13. Invalid Type (string for int)
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "get_book",
    "arguments": {
      "id": "not_a_number"
    }
  },
  "id": 12
}
```

14. Negative ID
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "get_book",
    "arguments": {
      "id": -1
    }
  },
  "id": 13
}
```

15. Maximum Integer
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "get_book",
    "arguments": {
      "id": 2147483647
    }
  },
  "id": 14
}
```

16. Case Sensitivity (should fail)
```
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "GET_BOOKS"
  },
  "id": 15
}
```

### DIFF
--- a/src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
+++ b/src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
@@ -18,20 +18,6 @@ namespace Azure.DataApiBuilder.Service.Tests.Mcp
     public class CustomMcpToolFactoryTests
     {
         /// <summary>
-        /// Test that CreateCustomTools returns empty collection when config is null.
-        /// </summary>
-        [TestMethod]
-        public void CreateCustomTools_ReturnsEmptyCollection_WhenConfigIsNull()
-        {
-            // Act
-            System.Collections.Generic.IEnumerable<Azure.DataApiBuilder.Mcp.Model.IMcpTool> tools = CustomMcpToolFactory.CreateCustomTools(null!, null);
-
-            // Assert
-            Assert.IsNotNull(tools);
-            Assert.AreEqual(0, tools.Count());
-        }
-
-        /// <summary>
         /// Test that CreateCustomTools returns empty collection when no entities exist.
         /// </summary>
         [TestMethod]


src/Azure.DataApiBuilder.Mcp/Core/McpServiceCollectionExtensions.cs
```diff
@@ -54,12 +57,25 @@ private static void RegisterAllMcpTools(IServiceCollection services)
...
+        /// <summary>
+        /// Registers custom MCP tools generated from stored procedure entity configurations.
+        /// </summary>
+        private static void RegisterCustomTools(IServiceCollection services, RuntimeConfig config)
+        {
+            // Create custom tools and register each as a singleton
+            foreach (IMcpTool customTool in CustomMcpToolFactory.CreateCustomTools(config))
+            {
+                services.AddSingleton<IMcpTool>(customTool);
+            }
```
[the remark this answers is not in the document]
souvikghosh04: this is already captured as a subsequent work item- #3049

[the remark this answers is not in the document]
souvikghosh04: this is already captured as a subsequent work item- https://github.com/Azure/data-api-builder/issues/3049

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+            catch (OperationCanceledException)
+            {
+                return McpResponseBuilder.BuildErrorResult(toolName, "OperationCanceled", "The operation was canceled.", logger);
+            }
+            catch (Exception ex)
+            {
+                logger?.LogError(ex, "Unexpected error in DynamicCustomTool for {EntityName}", _entityName);
+                return McpResponseBuilder.BuildErrorResult(toolName, "UnexpectedError", "An unexpected error occurred.", logger);
+            }
+        }
```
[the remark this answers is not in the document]
souvikghosh04: this is known and to be refactored in a separate PR.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+            {
+                responseData["value"] = JsonSerializer.SerializeToElement(Array.Empty<object>());
+            }
+
+            return McpResponseBuilder.BuildSuccessResult(
+                responseData,
+                logger,
+                $"Custom tool {toolName} executed successfully for entity {entityName}."
+            );
+        }
```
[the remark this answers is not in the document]
souvikghosh04: The helper methods BuildInputSchema and BuildExecuteSuccessResponse are lightweight, synchronous operations that process small in-memory collections (typically < 20 parameters). Adding cancellation checks would introduce overhead without meaningful benefit. Cancellation is already checked at the critical async points: before metadata resolution (line 96) and before database execution (line 205), which are the expensive operations that benefit from cancellation support.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+                // Add default parameters from configuration if not provided
+                if (entityConfig.Source.Parameters != null)
+                {
+                    foreach (ParameterMetadata param in entityConfig.Source.Parameters)
+                    {
+                        if (!context.FieldValuePairsInBody.ContainsKey(param.Name))
+                        {
+                            context.FieldValuePairsInBody[param.Name] = param.Default;
+                        }
+                    }
```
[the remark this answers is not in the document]
souvikghosh04: this is not needed since the current code is more readable

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+            }
+
+            return McpResponseBuilder.BuildSuccessResult(
+                responseData,
+                logger,
+                $"Custom tool {toolName} executed successfully for entity {entityName}."
+            );
+        }
+    }
+}
```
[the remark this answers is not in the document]
souvikghosh04: Added tests

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+            // Build input schema based on parameters
+            JsonElement inputSchema = BuildInputSchema();
+
+            return new Tool
+            {
+                Name = toolName,
+                Description = description,
+                InputSchema = inputSchema
+            };
+        }
```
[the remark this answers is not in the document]
souvikghosh04: Fixed

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+                    dbo: dbObject,
+                    requestPayloadRoot: requestPayloadRoot,
+                    operationType: EntityActionOperation.Execute);
+
+                // Add user-provided parameters
+                if (requestPayloadRoot != null)
+                {
+                    foreach (JsonProperty property in requestPayloadRoot.Value.EnumerateObject())
+                    {
+                        context.FieldValuePairsInBody[property.Name] = GetParameterValue(property.Value);
```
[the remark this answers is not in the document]
souvikghosh04: This will be refactored in a separate PR

src/Azure.DataApiBuilder.Mcp/Core/CustomMcpToolFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+                    {
+                        DynamicCustomTool tool = new(entityName, entity);
+                        customToolCount++;
+                        
+                        logger?.LogInformation(
+                            "Created custom MCP tool '{ToolName}' for stored procedure entity '{EntityName}'",
+                            tool.GetToolMetadata().Name,
+                            entityName);
+
+                        customTools.Add(tool);
```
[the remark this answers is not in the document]
souvikghosh04: Fixed

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+                JsonValueKind.String => element.GetString(),
+                JsonValueKind.Number =>
+                    element.TryGetInt64(out long longValue) ? longValue :
+                    element.TryGetDecimal(out decimal decimalValue) ? decimalValue :
+                    element.GetDouble(),
+                JsonValueKind.True => true,
+                JsonValueKind.False => false,
+                JsonValueKind.Null => null,
+                _ => element.ToString()
+            };
```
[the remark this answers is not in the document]
souvikghosh04: This will be refactored in a separate PR

src/Azure.DataApiBuilder.Mcp/Core/CustomMcpToolFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+                            entityName);
+                    }
+                }
+            }
+
+            logger?.LogInformation("Custom MCP tool generation complete. Created {Count} custom tools.", customToolCount);
+            return customTools;
+        }
+    }
+}
```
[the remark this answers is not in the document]
souvikghosh04: Added tests

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        /// <summary>
+        /// Converts entity name to tool name format (lowercase with underscores).
+        /// </summary>
+        private static string ConvertToToolName(string entityName)
+        {
+            // Convert PascalCase to snake_case
+            string result = Regex.Replace(entityName, "([a-z0-9])([A-Z])", "$1_$2");
+            return result.ToLowerInvariant();
```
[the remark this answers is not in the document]
souvikghosh04: This is currently planned in a separate PR

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+                Name = toolName,
+                Description = description,
+                InputSchema = inputSchema
+            };
+        }
+
+        /// <summary>
+        /// Executes the stored procedure represented by this custom tool.
+        /// </summary>
+        public async Task<CallToolResult> ExecuteAsync(
```
JerryNixon, line 90: # Important

Please add explicit OpenTelemetry tracing for custom stored-procedure MCP tools.

**Scope:** `DynamicCustomTool.ExecuteAsync(...)`

**Requirements:**
- Create an `Activity`/span around each tool invocation (e.g., span name `mcp.tool.call` or `dab.mcp.custom_tool.execute`).
- Add span attributes/tags at minimum:
  - `mcp.tool.name` = tool name
  - `dab.entity` = entity alias/name
  - `dab.operation` = `execute`
  - `db.procedure` (or equivalent) = stored procedure name / db object (schema-qualified if available)
- Record outcome on the span:
  - mark success vs failure (`ActivityStatusCode.Ok` / `Error`)
  - on exceptions, record the exception on the span and include an error message/code where available.

**Notes:**
- Logging via `ILogger` is not sufficient; the tool should emit tracing spans directly.
- Ensure spans are created for both success and failure paths (including auth/validation failures).

JerryNixon: # PLEASE TREAT TELEMETRY AS A BASE REQUIREMENT

JerryNixon: Please implement OpenTelemetry tracing for MCP tool execution **centrally** to avoid duplicating instrumentation in every tool.

**Goal:** Every MCP `tools/call` invocation (built-in DML tools and dynamic custom stored-proc tools) emits a consistent span.

**General idea:**
- Add instrumentation in the MCP dispatcher / tool execution entry point (the code path that resolves the tool by name and invokes `IMcpTool.ExecuteAsync`).
- Start an `Activity` span for each tool call (e.g., `mcp.tool.call` / `dab.mcp.tool.execute`) and ensure it wraps the entire execution including auth/validation and query execution.
- Apply standard tags (e.g., `mcp.tool.name`, `dab.entity` when applicable, `dab.operation`, `db.procedure` when applicable).
- Set span status to Ok/Error and record exceptions.

Aniruddh25: Agree on this - but we can add it in a separate subsequent PR - for easier PR review and merge process.

souvikghosh04: Added a work item for Feb to address this- #3060

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+{
+    /// <summary>
+    /// Dynamic custom MCP tool generated from stored procedure entity configuration.
+    /// Each custom tool represents a single stored procedure exposed as a dedicated MCP tool.
+    /// 
+    /// Note: The entity configuration is captured at tool construction time. If the RuntimeConfig
+    /// is hot-reloaded, GetToolMetadata() will return cached metadata (name, description, parameters)
+    /// from the original configuration. This is acceptable because:
+    /// 1. MCP clients typically call tools/list once at startup
+    /// 2. ExecuteAsync always validates against the current runtime configuration
```
Aniruddh25, line 36: will the validation fail in that case?

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+using static Azure.DataApiBuilder.Mcp.Model.McpEnums;
+
+namespace Azure.DataApiBuilder.Mcp.Core
+{
+    /// <summary>
+    /// Dynamic custom MCP tool generated from stored procedure entity configuration.
+    /// Each custom tool represents a single stored procedure exposed as a dedicated MCP tool.
+    /// 
+    /// Note: The entity configuration is captured at tool construction time. If the RuntimeConfig
+    /// is hot-reloaded, GetToolMetadata() will return cached metadata (name, description, parameters)
```
Aniruddh25, line 33: Do we plan to fix this hot reload behavior in the future? Is there a tracking issue for this?

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+                    properties[param.Name] = new Dictionary<string, object>
+                    {
+                        ["type"] = new[] { "string", "number", "boolean", "null" },
+                        ["description"] = param.Description ?? $"Parameter {param.Name}"
+                    };
+                }
+            }
+            else
+            {
+                schema["properties"] = new Dictionary<string, object>();
```
Aniruddh25: Why is there a need for a new properties Dictionary if we already initialize it on line 266 when we define the schema?

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+                var properties = (Dictionary<string, object>)schema["properties"];
+
+                foreach (var param in _entity.Source.Parameters)
+                {
+                    // Note: Parameter type information is not available in ParameterMetadata,
+                    // so we allow multiple JSON types to match the behavior of GetParameterValue
+                    // that handles string, number, boolean, and null values.
+                    properties[param.Name] = new Dictionary<string, object>
+                    {
+                        ["type"] = new[] { "string", "number", "boolean", "null" },
```
Aniruddh25, line 280: can we derive the parameter type from _entity.Source definition?

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+        private static CallToolResult BuildExecuteSuccessResponse(
+            string toolName,
+            string entityName,
+            Dictionary<string, object?>? parameters,
+            IActionResult? queryResult,
+            ILogger? logger)
+        {
+            Dictionary<string, object?> responseData = new()
+            {
+                ["entity"] = entityName,
```
Aniruddh25, line 320: have you double checked we expect these keys in the response as per the functional spec/enhancement description?

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: validations will work fine. for a long-running MCP session where config is hot-reloaded, clients operate with outdated tool metadata until they reconnect and call tools/list again.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: a work item is created to track it- https://github.com/Azure/data-api-builder/issues/3066, however, the general recommendation would be to have a clean restart for any changes in config such that MCP server. this is much robust and environment agnostic, and most software components follow that.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: ParameterMetadata doesn't expose type information. We could fetch from database metadata, but that would require making database calls during tool listing (performance concern) and breaks the cached metadata design. The generic type array matches our runtime GetParameterValue behavior which accepts any JSON primitive.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: This is kept similar to our ExecuteEntity tool since custom tool is essentially a speiallized version of ExecuteEntity. Keeping it consistent with the same and other built-in tools as well.

src/Azure.DataApiBuilder.Mcp/Core/CustomMcpToolFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /// <summary>
+        /// Creates custom MCP tools from entities configured with "mcp": { "custom-tool": true }.
+        /// </summary>
+        /// <param name="config">The runtime configuration containing entity definitions.</param>
+        /// <param name="logger">Optional logger for diagnostic information.</param>
+        /// <returns>Enumerable of custom tools generated from configuration.</returns>
+        public static IEnumerable<IMcpTool> CreateCustomTools(RuntimeConfig config, ILogger? logger = null)
+        {
+            if (config?.Entities == null)
```
Aniruddh25: is config nullable? doesnt seem to be looking at the parameter type.
```suggestion
            if (config.Entities == null)
```

So why do we have `config?`

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
```diff
@@ -0,0 +1,238 @@
...
+    public class CustomMcpToolFactoryTests
+    {
+        /// <summary>
+        /// Test that CreateCustomTools returns empty collection when config is null.
+        /// </summary>
+        [TestMethod]
+        public void CreateCustomTools_ReturnsEmptyCollection_WhenConfigIsNull()
+        {
+            // Act
+            var tools = CustomMcpToolFactory.CreateCustomTools(null!);
```
Aniruddh25: is there any scenario when config can be null?

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        /// <summary>
+        /// Test that CreateCustomTools returns empty collection when no entities exist.
+        /// </summary>
+        [TestMethod]
+        public void CreateCustomTools_ReturnsEmptyCollection_WhenNoEntities()
+        {
+            // Arrange
+            var config = CreateEmptyConfig();
```
Aniruddh25: use of `var` is discouraged unless it is apparent from the RHS. Since this is not an instantiation, recommend switching the type to `RuntimeConfig` instead of mentioning var here, Same in other tests.

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
```diff
@@ -0,0 +1,238 @@
...
+                ),
+                Entities: new(entities)
+            );
+        }
+
+        private static RuntimeConfig CreateConfigWithInvalidEntity()
+        {
+            var entities = new Dictionary<string, Entity>
+            {
+                // Valid SP
```
Aniruddh25: If this is a valid SP, what makes this a config with an invalid entity?

src/Service.Tests/Mcp/DynamicCustomToolTests.cs
```diff
@@ -0,0 +1,236 @@
...
+            // Arrange
+            Entity entity = CreateTestStoredProcedureEntity();
+            DynamicCustomTool tool = new("GetBooks", entity);
+
+            // Act
+            var metadata = tool.GetToolMetadata();
+
+            // Assert
+            Assert.IsNotNull(metadata.InputSchema);
+            var schemaObj = JsonDocument.Parse(metadata.InputSchema.GetRawText());
```
Aniruddh25: avoid usage of `var`

src/Service.Tests/Mcp/DynamicCustomToolTests.cs
```diff
@@ -0,0 +1,236 @@
...
+            Assert.IsTrue(schemaObj.RootElement.TryGetProperty("properties", out var props));
+            Assert.AreEqual(JsonValueKind.Object, props.ValueKind);
```
Aniruddh25, line 145: how are we checking that props is empty?

Aniruddh25 (changes requested):
Looks good overall, some confusion around what happens when a ConfigWithInvalidEntity is created.

JerryNixon (approved):

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
JerryNixon: This should be impossible. Maybe forgot `dotnet format`?

src/Azure.DataApiBuilder.Mcp/Core/CustomMcpToolFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+                logger?.LogWarning("No entities found in runtime configuration for custom tool generation.");
+                return Enumerable.Empty<IMcpTool>();
+            }
+
+            List<IMcpTool> customTools = new();
+
+            foreach ((string entityName, Entity entity) in config.Entities)
+            {
+                // Filter: Only stored procedures with custom-tool enabled
+                if (entity.Source.Type == EntitySourceType.StoredProcedure &&
```
anushakolan, line 35: You check entity.Source.Type, but entity.Source itself could be null if runtime config is malformed or partially loaded.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+
+                // 3) Validate entity still exists in configuration
+                if (!config.Entities.TryGetValue(_entityName, out Entity? entityConfig))
+                {
+                    return McpResponseBuilder.BuildErrorResult(toolName, "EntityNotFound", $"Entity '{_entityName}' not found in configuration.", logger);
+                }
+
+                if (entityConfig.Source.Type != EntitySourceType.StoredProcedure)
+                {
+                    return McpResponseBuilder.BuildErrorResult(toolName, "InvalidEntity", $"Entity {_entityName} is not a stored procedure.", logger);
```
anushakolan, line 124: We already validated in `CreateCustomTools` and in the constructor, is it needed to validated again?

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+            string result = Regex.Replace(entityName, "([a-z0-9])([A-Z])", "$1_$2");
+            return result.ToLowerInvariant();
+        }
+
+        /// <summary>
+        /// Builds the input schema for the tool based on entity parameters.
+        /// </summary>
+        private JsonElement BuildInputSchema()
+        {
+            var schema = new Dictionary<string, object>
```
anushakolan: nit: Let's try to avoid usage of var and use actual data types.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Converts entity name to tool name format (lowercase with underscores).
+        /// </summary>
+        private static string ConvertToToolName(string entityName)
+        {
+            // Convert PascalCase to snake_case
+            string result = Regex.Replace(entityName, "([a-z0-9])([A-Z])", "$1_$2");
```
anushakolan, line 254: ConvertToToolName converts PascalCase to snake_case, but two entities like `GenerateReport` and `Generate_Report` could collide. Maybe we could append a hash if duplicates occur

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -0,0 +1,375 @@
...
+    /// <summary>
+    /// Dynamic custom MCP tool generated from stored procedure entity configuration.
+    /// Each custom tool represents a single stored procedure exposed as a dedicated MCP tool.
+    /// 
+    /// Note: The entity configuration is captured at tool construction time. If the RuntimeConfig
+    /// is hot-reloaded, GetToolMetadata() will return cached metadata (name, description, parameters)
+    /// from the original configuration. This is acceptable because:
+    /// 1. MCP clients typically call tools/list once at startup
+    /// 2. ExecuteAsync always validates against the current runtime configuration
+    /// 3. Cached metadata improves performance for repeated metadata requests
```
anushakolan, line 37: What about happens if parameters change after hot-reload, MCP clients will see stale schemas?

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
```diff
@@ -0,0 +1,238 @@
...
+            // Arrange
+            var config = CreateConfigWithMixedEntities();
+
+            // Act
+            var tools = CustomMcpToolFactory.CreateCustomTools(config);
+
+            // Assert
+            Assert.IsNotNull(tools);
+            // Should only include GetBook (SP with custom-tool enabled)
+            Assert.AreEqual(1, tools.Count());
```
anushakolan, line 53: We can add `Assert.AreEqual(ToolType.Custom, tools.First().ToolType);`

src/Azure.DataApiBuilder.Mcp/Core/CustomMcpToolFactory.cs
souvikghosh04: `entity.Source` is a non-nullable property in `Entity.cs`, so the null-conditional check is not needed. The type system guarantees `Source` is always present.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: Both validations are necessary - one validates construction-time entity, the other validates runtime entity. This is part of the defensive design for hot-reload support.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: there is a separate work item planned for addressing duplicate and as already captured in https://github.com/Azure/data-api-builder/pull/3048#discussion_r2681848736.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: resolving this as its duplicate of this review comment- https://github.com/Azure/data-api-builder/pull/3048#discussion_r2692506351. general recommendation is to do a clean start for a config change but a separate work item planned as well.

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
souvikghosh04: in line 62 we create and assign only custom tools. since it's a compile-time constant (not configuration-driven), the assertion would always pass and doesn't test meaningful behavior. The current assertions on filtering logic and metadata are sufficient.

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
souvikghosh04: likely no but it's just for the sake of completeness of test scenario and coverage.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
souvikghosh04: this is addressed now across all places.

src/Service.Tests/Mcp/CustomMcpToolFactoryTests.cs
souvikghosh04: You are right- the helper creates a valid entity, so nothing would cause the try-catch to trigger. I have removed this test for now. Proper failure handling tests will be added in the comprehensive test PR mentioned in the class comment.

src/Service.Tests/Mcp/DynamicCustomToolTests.cs
souvikghosh04: added an assertion for that.

anushakolan (approved):

Aniruddh25 (changes requested):
Nullability of the `config` parameter type has inconsistent expectations.

PR #3054, opened 2026-01-13
#3053 Fix by using custom bool parsing that if a string is found then…

### LINKED ISSUE #3053: [Bug]: Application Insights can't be disabled by environment variables
### What happened?

There is no environment variable substitution on booleans thus you can't enabled and disable telemetry by environment variables.

Requires significant hacks to work around this at startup in the docker container. 

### Version

1.6+

Azure SQL

Container Apps

REST

### Relevant log output

```Text
Error that config isn't valid
```

### Code of Conduct

simonsabin:
… it uses the environment replacement and looks for a true false 1 or 0.

## Why make this change?

Fixes #3053 

Boolean values can't be set using environment variables. This allows that

## What is this change?

Using custom JsonConverter for bools that if a string is detected it uses the string serialiser that uses the environment replacement rules. 

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/Core/CustomMcpToolFactory.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Core/CustomMcpToolFactory.cs
@@ -1,0 +1,63 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Mcp.Model;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.DataApiBuilder.Mcp.Core
+{
+    /// <summary>
+    /// Factory for creating custom MCP tools from stored procedure entity configurations.
+    /// Scans runtime configuration and generates dynamic tools for entities marked with custom-tool enabled.
+    /// </summary>
+    public class CustomMcpToolFactory
+    {
+        /// <summary>
+        /// Creates custom MCP tools from entities configured with "mcp": { "custom-tool": true }.
+        /// </summary>
+        /// <param name="config">The runtime configuration containing entity definitions.</param>
+        /// <param name="logger">Optional logger for diagnostic information.</param>
+        /// <returns>Enumerable of custom tools generated from configuration.</returns>
+        public static IEnumerable<IMcpTool> CreateCustomTools(RuntimeConfig config, ILogger? logger = null)
+        {
+            if (config.Entities == null)
+            {
+                logger?.LogWarning("No entities found in runtime configuration for custom tool generation.");
+                return Enumerable.Empty<IMcpTool>();
+            }
+
+            List<IMcpTool> customTools = new();
+
+            foreach ((string entityName, Entity entity) in config.Entities)
+            {
+                // Filter: Only stored procedures with custom-tool enabled
+                if (entity.Source.Type == EntitySourceType.StoredProcedure &&
+                    entity.Mcp?.CustomToolEnabled == true)
+                {
+                    try
+                    {
+                        DynamicCustomTool tool = new(entityName, entity);
+
+                        logger?.LogInformation(
+                            "Created custom MCP tool '{ToolName}' for stored procedure entity '{EntityName}'",
+                            tool.GetToolMetadata().Name,
+                            entityName);
+
+                        customTools.Add(tool);
+                    }
+                    catch (Exception ex)
+                    {
+                        logger?.LogError(
+                            ex,
+                            "Failed to create custom tool for entity '{EntityName}'. Skipping.",
+                            entityName);
+                    }
+                }
+            }
+
+            logger?.LogInformation("Custom MCP tool generation complete. Created {Count} custom tools.", customTools.Count);
+            return customTools;
+        }
+    }
+}
--- a/src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
@@ -1,0 +1,375 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Data.Common;
+using System.Text.Json;
+using System.Text.RegularExpressions;
+using Azure.DataApiBuilder.Auth;
+using Azure.DataApiBuilder.Config.DatabasePrimitives;
+using Azure.DataApiBuilder.Config.ObjectModel;
+using Azure.DataApiBuilder.Core.Configurations;
+using Azure.DataApiBuilder.Core.Models;
+using Azure.DataApiBuilder.Core.Resolvers;
+using Azure.DataApiBuilder.Core.Resolvers.Factories;
+using Azure.DataApiBuilder.Core.Services;
+using Azure.DataApiBuilder.Mcp.Model;
+using Azure.DataApiBuilder.Mcp.Utils;
+using Azure.DataApiBuilder.Service.Exceptions;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Protocol;
+using static Azure.DataApiBuilder.Mcp.Model.McpEnums;
+
+namespace Azure.DataApiBuilder.Mcp.Core
+{
+    /// <summary>
+    /// Dynamic custom MCP tool generated from stored procedure entity configuration.
+    /// Each custom tool represents a single stored procedure exposed as a dedicated MCP tool.
+    /// 
+    /// Note: The entity configuration is captured at tool construction time. If the RuntimeConfig
+    /// is hot-reloaded, GetToolMetadata() will return cached metadata (name, description, parameters)
+    /// from the original configuration. This is acceptable because:
+    /// 1. MCP clients typically call tools/list once at startup
+    /// 2. ExecuteAsync always validates against the current runtime configuration
+    /// 3. Cached metadata improves performance for repeated metadata requests
+    /// </summary>
+    public class DynamicCustomTool : IMcpTool
+    {
+        private readonly string _entityName;
+        private readonly Entity _entity;
+
+        /// <summary>
+        /// Initializes a new instance of DynamicCustomTool.
+        /// </summary>
+        /// <param name="entityName">The entity name from configuration.</param>
+        /// <param name="entity">The entity configuration object.</param>
+        public DynamicCustomTool(string entityName, Entity entity)
+        {
+            _entityName = entityName ?? throw new ArgumentNullException(nameof(entityName));
+            _entity = entity ?? throw new ArgumentNullException(nameof(entity));
+
+            // Validate that this is a stored procedure
+            if (_entity.Source.Type != EntitySourceType.StoredProcedure)
+            {
+                throw new ArgumentException(
+                    $"Custom tools can only be created for stored procedures. Entity '{entityName}' is of type '{_entity.Source.Type}'.",
+                    nameof(entity));
+            }
+        }
+
+        /// <summary>
+        /// Gets the type of the tool, which is Custom for dynamically generated tools.
+        /// </summary>
+        public ToolType ToolType { get; } = ToolType.Custom;
+
+        /// <summary>
+        /// Gets the metadata for this custom tool, including name, description, and input schema.
+        /// </summary>
+        public Tool GetToolMetadata()
+        {
+            string toolName = ConvertToToolName(_entityName);
+            string description = _entity.Description ?? $"Executes the {toolName} stored procedure";
+
+            // Build input schema based on parameters
+            JsonElement inputSchema = BuildInputSchema();
+
+            return new Tool
+            {
+                Name = toolName,
+                Description = description,
+                InputSchema = inputSchema
+            };
+        }
+
+        /// <summary>
+        /// Executes the stored procedure represented by this custom tool.
+        /// </summary>
+        public async Task<CallToolResult> ExecuteAsync(
+            JsonDocument? arguments,
+            IServiceProvider serviceProvider,
+            CancellationToken cancellationToken = default)
+        {
+            ILogger<DynamicCustomTool>? logger = serviceProvider.GetService<ILogger<DynamicCustomTool>>();
+            string toolName = GetToolMetadata().Name;
+
+            try
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // 1) Resolve required services & configuration
+                RuntimeConfigProvider runtimeConfigProvider = serviceProvider.GetRequiredService<RuntimeConfigProvider>();
+                RuntimeConfig config = runtimeConfigProvider.GetConfig();
+
+                // 2) Parse arguments from the request
+                Dictionary<string, object?> parameters = new();
+                if (arguments != null)
+                {
+                    foreach (JsonProperty property in arguments.RootElement.EnumerateObject())
+                    {
+                        parameters[property.Name] = GetParameterValue(property.Value);
+                    }
+                }
+
+                // 3) Validate entity still exists in configuration
+                if (!config.Entities.TryGetValue(_entityName, out Entity? entityConfig))
+                {
+                    return McpResponseBuilder.BuildErrorResult(toolName, "EntityNotFound", $"Entity '{_entityName}' not found in configuration.", logger);
+                }
+
+                if (entityConfig.Source.Type != EntitySourceType.StoredProcedure)
+                {
+                    return McpResponseBuilder.BuildErrorResult(toolName, "InvalidEntity", $"Entity {_entityName} is not a stored procedure.", logger);
+                }
+
+                // 4) Resolve metadata
+                if (!McpMetadataHelper.TryResolveMetadata(
+                        _entityName,
+                        config,
+                        serviceProvider,
+                        out ISqlMetadataProvider sqlMetadataProvider,
+                        out DatabaseObject dbObject,
+                        out string dataSourceName,
+                        out string metadataError))
+                {
+                    return McpResponseBuilder.BuildErrorResult(toolName, "EntityNotFound", metadataError, logger);
+                }
+
+                // 5) Authorization check
+                IAuthorizationResolver authResolver = serviceProvider.GetRequiredService<IAuthorizationResolver>();
+                IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
+                HttpContext? httpContext = httpContextAccessor.HttpContext;
+
+                if (!McpAuthorizationHelper.ValidateRoleContext(httpContext, authResolver, out string roleError))
+                {
+                    return McpErrorHelpers.PermissionDenied(toolName, _entityName, "execute", roleError, logger);
+                }
+
+                if (!McpAuthorizationHelper.TryResolveAuthorizedRole(
+                    httpContext!,
+                    authResolver,
+                    _entityName,
+                    EntityActionOperation.Execute,
+                    out string? effectiveRole,
+                    out string authError))
+                {
+                    return McpErrorHelpers.PermissionDenied(toolName, _entityName, "execute", authError, logger);
+                }
+
+                // 6) Build request payload
+                JsonElement? requestPayloadRoot = null;
+                if (parameters.Count > 0)
+                {
+                    string jsonPayload = JsonSerializer.Serialize(parameters);
+                    using JsonDocument doc = JsonDocument.Parse(jsonPayload);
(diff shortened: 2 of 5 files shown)


src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+
+/// <summary>
+/// Custom string json converter factory to replace environment variables and other variable patterns
+/// during deserialization using the DeserializationVariableReplacementSettings.
+/// </summary>
+public class BoolJsonConverter : JsonConverter<bool>
```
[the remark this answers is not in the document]
Aniruddh25: If the filename was changed to ConverterFactory, the class name should be changed to BooleanJsonConverterFactory here as well.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -213,6 +213,71 @@ private static string GetExpectedPropertyValue(string envVarName, bool replaceEn
...
+                string configJson = @"{
+                    ""$schema"": ""https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch-alpha/dab.draft.schema.json"",
+                    ""data-source"": {
+                        ""database-type"": ""mssql"",
+                        ""connection-string"": ""Server=tcp:127.0.0.1,1433;Persist Security Info=False;Trusted_Connection=True;TrustServerCertificate=True;MultipleActiveResultSets=False;Connection Timeout=5;""
+                    },
+                    ""runtime"": {
+                        ""telemetry"": {
+                            ""application-insights"": {
+                                ""enabled"": " + configValue + @",
```
Aniruddh25, line 279: Should there be a test for the configValue itself be in double quotes? So, 

`"enabled" : "true"` or
`"enabled" : "@env('application_insights_enabled')"`

Aniruddh25: Without this, the current test is not testing the scenario when configValue is hardcoded to true because this code will translate to `"enabled": true,`

simonsabin: putting a load more tests

src/Config/Converters/BooleanJsonConverter.cs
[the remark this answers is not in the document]
Aniruddh25: In reality, this is not a factory, So the file name should be kept as BoolJsonConverter to match the existing class name `BoolJsonConverter`

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -213,6 +213,71 @@ private static string GetExpectedPropertyValue(string envVarName, bool replaceEn
...
+        [DataRow(true, true, DisplayName = "Config with enabled set to true ")]
+        [DataRow(true, false, DisplayName = "Config with enabled set to false")]
+        [DataRow(false, true,  DisplayName = "Replace environment variables containing boolean value with true")]
+        [DataRow(false, false, DisplayName = "Replace environment variables containing boolean value with false")]
+
+        public void TestTelemetryApplicationInsightsEnabledWithEnvironmentVariable(bool hardcoded, bool expected)
+        {
+            // Arrange
+            const string envVarName = "APP_INSIGHTS_ENABLED";
+             string envVarValue = expected.ToString();
```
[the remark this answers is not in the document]
simonsabin: Extra test cases have been added

src/Config/Converters/BooleanJsonConverter.cs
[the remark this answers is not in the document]
simonsabin: file renamed

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
simonsabin: a whole set of extra tests to handle string and non string booleans

simonsabin:
> Additional testing for double quoted value boolean value to test the string deserialization.

I think this is now catered for it you mean where the json is the following. 

```
{
 .... 
  "enabled": "true"
  ...
}
```
and also 
```
{
 .... 
   "enabled": true
  ...
}
```

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -213,6 +213,71 @@ private static string GetExpectedPropertyValue(string envVarName, bool replaceEn
...
+                Assert.IsNotNull(runtimeConfig.Runtime.Telemetry.ApplicationInsights, "ApplicationInsights section should not be null");
+                Assert.AreEqual("InstrumentationKey=test-key", runtimeConfig.Runtime.Telemetry.ApplicationInsights.ConnectionString, "Connection string should be preserved");
+                Assert.AreEqual(expected, runtimeConfig.Runtime.Telemetry.ApplicationInsights.Enabled, "ApplicationInsights enabled value should match expected value");
+            }
+            finally
+            {
+                // Cleanup
+                Environment.SetEnvironmentVariable(envVarName, null);
+            }
+        }
```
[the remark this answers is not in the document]
simonsabin: extra test cases have been added

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
simonsabin: test now check for true, false, 0, 1 natively, hardcoded within a string, or from an environment replacement. 
Assuming keyvault replacements would "just work"

JerryNixon:
I am not sure this resolves JSON Schema constraints/editing errors, does it?

simonsabin (commented):
> I am not sure this resolves JSON Schema constraints/editing errors, does it?

It doesn't but this would

simonsabin:
> I am not sure this resolves JSON Schema constraints/editing errors, does it?

@JerryNixon  It doesn't but I've just pushed change to the schema that would. 

If thats an acceptable approach I can apply to other booleans.

Interestinlythere does seem to be a mix of json schema validation jsonconverter validation. The jsonconverter stuff looks overly complicated. for some simple serialisation.

simonsabin:
Added tests to validate the schema approach.  for key enabled flags.

simonsabin:
forgot formatting, try again

schemas/dab.draft.schema.json
```diff
@@ -1391,7 +1391,15 @@
...
-          "required": ["singular"]
+          "required": [ "singular" ]
+        }
+      ]
+    },
+    "boolean-or-string": {
+      "oneOf":[
+        {
+          "type": [ "boolean", "string" ],
+          "pattern": "^(true|false|1|0|@env\\('.*'\\)|@akv\\('.*'\\))$"
```
JerryNixon:
```suggestion
          "pattern": "^(?:true|false|1|0|@.{3}\\('.*'\\))$"
```

src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+/// other converters (for example, the string converter) before the value is parsed here.
+public class BoolJsonConverter : JsonConverter<bool>
+{
+
+    public override bool Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        if (reader.TokenType is JsonTokenType.Null)
+        {
+
+            throw new JsonException();
```
JerryNixon:
```suggestion
            throw new JsonException("Unexpected null JSON token. Expected a boolean literal or a valid @expression.");
```

src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+                _ => throw new JsonException($"Invalid boolean value. Specify either true or false."),
+            };
+            return result;
+        }
+        else
+        {
+            return reader.GetBoolean();
+        }
+
+        throw new JsonException();
```
JerryNixon:
```suggestion
        throw new JsonException("Invalid JSON value. Expected a boolean literal or a valid @expression.");
```

simonsabin: This is pattern is repeated a few times in the converters.

CONTRIBUTING.md
```diff
@@ -112,17 +112,42 @@ if [ "\$(get_files)" = '' ]; then
 fi
 
 get_files |
-    xargs dotnet format src/Azure.DataApiBuilder.Service.sln \\
-        --check \\
-        --fix-whitespace --fix-style warn --fix-analyzers warn \\
+    xargs dotnet format src/Azure.DataApiBuilder.sln \\
```
Aniruddh25, line 115: why are we removing `--fix-whitespace --fix-style warn --fix-analyzers warn` options? Are those not useful?

simonsabin: They arent valid https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format

Not sure when they stopped being valid

schemas/dab.draft.schema.json
```diff
@@ -1399,7 +1399,7 @@
       "oneOf":[
         {
           "type": [ "boolean", "string" ],
-          "pattern": "^(true|false|1|0|@env\\('.*'\\)|@akv\\('.*'\\))$"
+          "pattern": "^(?:true|false|1|0|@.{3}\\('.*'\\))$"
```
Aniruddh25: wouldnt doing `.{3}` allow any pattern other than env or akv?

If yes, I think we should retain the previous pattern.

CONTRIBUTING.md
simonsabin: Happy to chat over, not sure where the arguments in the hook came from

schemas/dab.draft.schema.json
simonsabin: This was @JerryNixon 's suggestion. Assume it was a bit of forward planning.

simonsabin:
> Looks good, except for the pattern match for @env/@akv.

Can you discuss this with @JerryNixon and decide

simonsabin:
Integration tests aren't passing, is that an issue?

Aniruddh25:
> Integration tests aren't passing, is that an issue?

that's not. rerunning should fix it.

Aniruddh25:
> > Looks good, except for the pattern match for @env/@akv.
> 
> Can you discuss this with @JerryNixon and decide

checked with Jerry. your previous pattern is what we both recommend.

schemas/dab.draft.schema.json
simonsabin: pushed fix

Aniruddh25 (approved):

src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+                _ => throw new JsonException($"Invalid boolean value. Specify either true or false."),
+            };
+            return result;
+        }
+        else
+        {
+            return reader.GetBoolean();
+        }
+
+        throw new JsonException("Invalid JSON value. Expected a boolean literal or a valid @expression.");
```
souvikghosh04: the exception thrown at line 54 seems to be unreachable as all code paths return before reaching here

src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            return result;
+        }
+        else if (reader.TokenType == JsonTokenType.Number)
+        {
+            bool result = reader.GetInt32() switch
+            {
+                1 => true,
+                0 => false,
+                _ => throw new JsonException($"Invalid boolean value. Specify either true or false."),
```
souvikghosh04: nit: add a more specific message saying acceptable values are 1 and 0. something like - "Specify 1 for true or 0 for false"

src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Azure.DataApiBuilder.Config.Converters;
+
+/// <summary>
+/// JSON converter for boolean values that also supports string representations such as
+/// "true", "false", "1", and "0". Any environment variable replacement is handled by
+/// other converters (for example, the string converter) before the value is parsed here.
```
souvikghosh04, line 12: nit: closing `</summary>` tag missing

src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            throw new JsonException("Unexpected null JSON token. Expected a boolean literal or a valid @expression.");
+        }
+
+        if (reader.TokenType == JsonTokenType.String)
+        {
+
+            string? tempBoolean = JsonSerializer.Deserialize<string>(ref reader, options);
+
+            bool result = tempBoolean?.ToLower() switch
```
souvikghosh04, line 30: minor optimization- `ToLower()` has an allocation overhead as it creates a new string. use `StringComparison.OrdinalIgnoreCase` instead. Refer- https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1862#rule-description

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -213,6 +215,138 @@ private static string GetExpectedPropertyValue(string envVarName, bool replaceEn
...
+                configJson,
+                out RuntimeConfig runtimeConfig,
+                replacementSettings: new DeserializationVariableReplacementSettings(
+                    azureKeyVaultOptions: null,
+                    doReplaceEnvVar: true,
+                    doReplaceAkvVar: false));
+
+            // Assert
+            Assert.IsTrue(IsParsed);
+            Assert.AreEqual("InstrumentationKey=test-key", runtimeConfig.Runtime.Telemetry.ApplicationInsights.ConnectionString, "Connection string should be preserved");
```
souvikghosh04: move `InstrumentationKey=test-key"` into a CONST string and use instead of repeatedly using the same hard-coded value everywhere.

simonsabin: There are a large number of strings repeated. Not sure it adds anything to make this a const.

src/Config/Converters/BooleanJsonConverter.cs
simonsabin: That doesn't work with switch. This a in frequent action that I don't think one needs to be worried about allocations.

src/Service.Tests/UnitTests/EnvironmentTests copy.cs
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) Microsoft Corporation.
```
Aniruddh25: accidental addition of this file?

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
souvikghosh04: that is the exact reason to move to a CONST. I see your point but to clarify the benefit:  if Application Insights 
changes their connection string format (e.g., requiring additional 
parameters), we would need to update it in 8 places.  Extracting to a const 
ensures consistency and reduces the chance of missing one during updates. 

That said, I recognize this is test code and the value is a dummy but consistency is important.

souvikghosh04 (approved):
Approved with suggestions.

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
simonsabin: My point is there is no consistency. This is a test and use of a const I dont agree does anything. 
If you want to follow that you'd need to make the ALL the test configs to be based on a combination of consts.
I've followed the consistency in AddTelemetryTests.cs, EndToEndTests.cs.

src/Service.Tests/UnitTests/EnvironmentTests copy.cs
simonsabin: yep

src/Config/Converters/BooleanJsonConverter.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            return result;
+        }
+        else if (reader.TokenType == JsonTokenType.Number)
+        {
+            bool result = reader.GetInt32() switch
+            {
+                1 => true,
+                0 => false,
+                _ => throw new JsonException($"Invalid boolean value. Specify either true or false."),
```
souvikghosh04: since you are checking `(reader.TokenType == JsonTokenType.Number)`, the exception message should reflect the same. update the exception message accurately and keep it consistent to the checks and error message between line 30 and line 35

simonsabin: changed

src/Service.Tests/UnitTests/RuntimeConfigLoaderJsonDeserializerTests.cs
```diff
@@ -213,6 +215,138 @@ private static string GetExpectedPropertyValue(string envVarName, bool replaceEn
...
+        }
+
+        /// <summary>
+        ///  
+        /// </summary>
+        /// <param name="configValue">Value to set in the config to cause error</param>
+        /// <param name="message">Error message</param>
+        [TestMethod]
+        [DataRow("somenonboolean", "Invalid boolean value: somenonboolean. Specify either true or false.", DisplayName = "ApplicationInsights.Enabled invalid value should error")]
+        public void TestTelemetryApplicationInsightsEnabledShouldError(string configValue, string message)
```
souvikghosh04, line 309: this test is failing and blocking the merge. there are other tests as well and I can't capture all of them in review comment.

```
Failed ApplicationInsights.Enabled invalid value should error [8 ms]
  Error Message:
   Assert.AreEqual failed. Expected:<Invalid boolean value: somenonboolean. Specify either true or 1 for true, false or 0 for false>. Actual:<Invalid boolean value: somenonboolean. Specify either true or false.>. 
  Stack Trace:
     at Azure.DataApiBuilder.Service.Tests.UnitTests.RuntimeConfigLoaderJsonDeserializerTests.TestTelemetryApplicationInsightsEnabledShouldError(String configValue, String message) in D:\a\1\s\src\Service.Tests\UnitTests\RuntimeConfigLoaderJsonDeserializerTests.cs:line 347
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)

  Skipped InsertMutationForNonGraphQLTypeTable
  Skipped InsertIntoInsertableComplexView
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.InvalidOperationException: The requested operation requires an element of type 'Object', but the target element has type 'Null'.
   at System.Text.Json.ThrowHelper.ThrowJsonElementWrongTypeException(JsonTokenType expectedType, JsonTokenType actualType)
   at System.Text.Json.JsonDocument.TryGetNamedPropertyValue(Int32 index, ReadOnlySpan`1 propertyName, JsonElement& value)
   at System.Text.Json.JsonElement.TryGetProperty(String propertyName, JsonElement& value)
   at Azure.DataApiBuilder.Service.Tests.Configuration.ConfigurationTests.ValidateMutationSucceededAtDbLayer(TestServer server, HttpClient client, String query, String queryName, String authToken, String clientRoleHeader) in D:\a\1\s\src\Service.Tests\Configuration\ConfigurationTests.cs:line 3183
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
```

souvikghosh04 (commented):
some tests are failing which is blocking the merge. please ensure to run and validate the tests locally. for example- `..\CosmosTests\SchemaGeneratorTest.cs` has multiple failing tests.

simonsabin:
Tests were failing due to inconsistencies with CR and LF endings. 
It appears that schemas are generted with CRLFs. The code formatting is enforcing LFs in git. 
This makes the results slightly undeterminstic. Had changed the gitattributes to align the .cs files with the code formatting rules, but that breaks as files checked out get LF line endings which breaks the test.

simonsabin:
Are these test failues transient?

PR #3077, opened 2026-01-23
[MCP]Fixed CLI bug in `fields.primary-key` updation when using update entity command.

anushakolan:
## Why make this change?
When updating an entity with a command like: `dab update Incident --fields.name "Id" --fields.description "Unique Key"`, the existing `primary-key` flag on Id was being changed to `false`even though no `primary-key` option was specified.
This change removes that surprise so primary-key metadata only changes when explicitly requested.

## What is this change?
- Adjusts the entity update logic so that:
1. If the user does not provide any fields.primary-key option, existing primary-key flags for fields mentioned in fields.name are preserved.
2. If the user does provide fields.primary-key, those true/false values are still applied positionally to the corresponding fields exactly as before.

- Adds a regression test that:
1. Starts from a config where Id is already a primary key.
2. Runs an update that only changes the description for Id via `fields.name` and `fields.description`.
3. Verifies that the description changes and the primary-key flag on Id remains true.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

1. Existing CLI unit tests for entity update.
2. New regression test that validates `primary-key` is preserved when only the field description is updated.

## Sample Request(s)

- To demonstrate the fixed behavior:

1. Initial state: Id is configured as a primary key on the entity.
2. Command: `dab update Incident --fields.name "Id" --fields.description "Unique Key 2"`
3. Result after this change:
4. The description for Id becomes “Unique Key 2”.
5. The primary-key flag for Id stays true.

- To explicitly change primary keys (still supported):
1. Set Id as primary key:
` dab update Incident --fields.name "Id" --fields.primary-key true`

2. Clear Id as primary key:
`dab update Incident --fields.name "Id" --fields.primary-key false`

### DIFF
--- a/src/Cli.Tests/UpdateEntityTests.cs
+++ b/src/Cli.Tests/UpdateEntityTests.cs
@@ -1101,11 +1101,14 @@ namespace Cli.Tests
         }
 
         /// <summary>
-        /// Updating a field's description without --fields.primary-key
-        /// should not change its existing primary-key flag.
+        /// Updating a field's description should either preserve or clear
+        /// the existing primary-key flag depending on whether an explicit
+        /// --fields.primary-key value is provided.
         /// </summary>
-        [TestMethod]
-        public void TestUpdateFieldDescriptionPreservesPrimaryKeyWhenNoFlagProvided()
+        [DataTestMethod]
+        [DataRow(null, true, DisplayName = "No primary-key flag: preserve existing true")]
+        [DataRow(false, false, DisplayName = "Explicit primary-key false: clear existing true")]
+        public void TestUpdateFieldDescriptionPrimaryKeyBehavior(bool? primaryKeyFlag, bool expectedPrimaryKey)
         {
             string initialConfig = GetInitialConfigString() + "," + @"
                 ""entities"": {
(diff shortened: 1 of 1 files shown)


src/Cli.Tests/UpdateEntityTests.cs
```diff
@@ -1100,6 +1100,164 @@ public void TestUpdateEntityDescription()
             Assert.AreEqual("Updated description", updatedRuntimeConfig.Entities["MyEntity"].Description);
         }
 
+        /// <summary>
+        /// Updating a field's description without --fields.primary-key
+        /// should not change its existing primary-key flag.
+        /// </summary>
+        [TestMethod]
+        public void TestUpdateFieldDescriptionPreservesPrimaryKeyWhenNoFlagProvided()
```
RubenCerna2079: I see that both tests are basically the same, I think it would be better for them to be the same test and just have a variable that changes the `fieldsPrimaryKeyCollection` field.

souvikghosh04 (approved):
LGTM. The change is small and precise addressing the intended behavior. Good to merge after addressing the comments.

Issue #2966, opened 2025-11-10
Query `include` & `exclude` properties for autoentities


RubenCerna2079:
Implement the main logic for the wildcards which passes the `patterns.include` and `patterns.exclude` properties to the database as part of a `LIKE` query and returns the tables that will be added as entities inside of DAB.

Issue #2965, opened 2025-11-10
Add interpolation for schema/object


RubenCerna2079:
`schema` and `object` will be reserved words inside the config file when using brackets. They will allow the user to add the name of the schema and object automatically, without the need to find it themselves.

Issue #2794, opened 2025-07-30
[Enh]: Support autoentities


JerryNixon:
## What

Add entities automatically.

## Why

Super-simplify setup and onboarding.

> [!Note]
> Autoentities cannot conflict with explicit entities.

## Constraints

* Data-source type: `MSSQL`-only
* Object-source type: `Table`-only
* When at least one `autoentities` member is present, the top-level `entities` section becomes optional.

## How

Add new top-level section to the configuration.

```jsonc
{
  "runtime": {},
  "entities": {},
  "autoentities": {} // new
}
```

This defines the search pattern as well as the entity defaults.

> [!Important]
> Autoentities are in-memory only. The configuration file is never updated from this behavior.

### Configuration

```jsonc
{
  "autoentities": {
    "{definition-name}": {

      "patterns": {
        "include": [ "t1", "t2", "t3", "%.%" ], // default: null which means all. mssql syntax (T-SQL LIKE)
        "exclude": [ "t3" ], // default: null which means none. mssql syntax (T-SQL LIKE)
        "name": "{schema}{object}" // default: null which means "{object}". This is interpolation syntax
      }

      "template": {
        "mcp": { "dml-tool": "<boolean>" }, // default: true
        "rest": { "enabled": "<boolean>" }, // default: true
        "graphql": { "enabled": "<boolean>" }, // default: true
        "health": { "enabled": "<boolean>" }, // default: true
        "cache": {
          "enabled": "<boolean>", // default: false
          "ttl-seconds": "<integer>", // default: 5
          "level": "<enum>" // default: L1L2
        }
      },

      "permissions": [ // at least one is required
        {
          "role": "<string>",
          "actions": [
            "*",
            "create",
            "read",
            "update",
            "delete"
          ]
        }
      ]
    }
  }
}
```

### Definition and Patterns

* `{definition-name}` uniquely identifies a definition in `autoentities`, used by the CLI for upserts.
* `patterns.include`: `null` includes all (default). Uses MSSQL `LIKE` syntax validated at runtime.
* `patterns.exclude`: `null` excludes none (default). Uses MSSQL `LIKE` syntax validated at runtime.
* `patterns.name`: `null` equals `{object}` (default). Supports `{schema}` and `{object}` interpolation (for example `{schema}{object}`) and must be unique.
* The `patterns` section is optional and uses defaults when omitted.
* When an entity matches both include and exclude, exclude wins.

### Template and Permissions

[...]

PR #3082, opened 2026-01-27
Create and execute query to return tables and interpolated names for Autoentities

RubenCerna2079:
## Why make this change?
- #2966
- #2965 
We need to create a query and send it to the database in order to receive all of the tables that will be turned into entities from the autoentities configuration. This includes the `schema`, `object`, and `entity name`.

## What is this change?
- **IQueryBuilder & MsSqlQueryBuilder**: Adds new function that builds the query that returns all of the tables that will be turned into entities.
- **SqlMetadataProvider & MsSqlMetadataProvider**: Adds function that executes query and receives the information for all of the existing autoentities. The information that is received for each table is the `schema`, `object`, and `entity name`, an example of this would be `dbo`, `book`, `dbo.book`. The query also transforms the reserved words `{schema}` and `{object}` into the tables schema and object and returns it in the name.
- **SqlMetadataProviderUnitTests**: Adds testing to ensure that the query returns the expected output.

## How was this tested?
- [ ] Integration Tests
- [x] Unit Tests
- [x] Manual Testing

### DIFF
--- a/src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
@@ -292,19 +292,9 @@ namespace Azure.DataApiBuilder.Core.Services
         }
 
         /// <inheritdoc/>
-        // TODO: Finish implementation of autoentities generation in task #3052
         protected override async Task GenerateAutoentitiesIntoEntities()
         {
-            await Task.CompletedTask; // Temporary await to suppress build errors.
-
-            /*RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
-            if (runtimeConfig.Autoentities is not null)
-            {
-                foreach ((string name, Autoentity autoentity) in runtimeConfig.Autoentities.AutoEntities)
-                {
-                    JsonArray? resultArray = await QueryAutoentitiesAsync(autoentity);
-                }
-            }*/
+            await Task.CompletedTask;
         }
 
         public async Task<JsonArray?> QueryAutoentitiesAsync(Autoentity autoentity)


src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+                "( N'geography', N'geometry', N'hierarchyid', N'sql_variant', N'xml', N'rowversion', N'vector' ) ) THEN 1 ELSE 0 END AS contains_invalid_types FROM eligible_tables AS a WHERE " +
+                "a.is_system_object = 0 AND ( NOT EXISTS (SELECT 1 FROM exclude_patterns) OR NOT EXISTS ( SELECT 1 FROM exclude_patterns AS ep WHERE a.full_name LIKE ep.pattern " +
+                "COLLATE DATABASE_DEFAULT ESCAPE '\\' ) ) AND ( NOT EXISTS (SELECT 1 FROM include_patterns) OR EXISTS ( SELECT 1 FROM include_patterns AS ip WHERE a.full_name LIKE ip.pattern " +
+                "COLLATE DATABASE_DEFAULT ESCAPE '\\' ) ) AND ( @exclude_invalid_types = 0 OR NOT EXISTS ( SELECT 1 FROM sys.columns AS c JOIN sys.types AS ty ON c.user_type_id = ty.user_type_id " +
+                "WHERE c.object_id = a.object_id AND ty.name IN ( N'geography', N'geometry', N'hierarchyid', N'sql_variant', N'xml', N'rowversion', N'vector' ) ) ) ORDER BY a.schema_name, a.object_name;";
+
+            return query;
```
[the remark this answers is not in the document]
anushakolan: This is a valid point, please address it.

src/Core/Resolvers/IQueryBuilder.cs
```diff
@@ -90,5 +90,7 @@ public interface IQueryBuilder
         /// DB Connection Param.
         /// </summary>
         public string QuoteTableNameAsDBConnectionParam(string param);
+
+        public string BuildGetAutoentitiesQuery(string include, string exclude, string namePattern) => throw new NotImplementedException();
```
anushakolan: `public string BuildGetAutoentitiesQuery(string include, string exclude, string namePattern)
    => throw new NotSupportedException($"{GetType().Name} does not support Autoentities yet.");`

Prefer `NotSupportedException` (with a provider-specific message) when the intent is “this provider doesn’t support auto-entities yet,”

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+        /// Builds the query used to get the list of tables with the SQL LIKE
+        /// syntax that will be transformed into entities.
+        /// </summary>
+        /// <param name="include">Pattern for tables that will be included.</param>
+        /// <param name="exclude">Pattern for tables that will be excluded.</param>
+        /// <param name="namePattern">Pattern for naming the entities.</param>
+        /// <returns></returns>
+        public string BuildGetAutoentitiesQuery(string include, string exclude, string namePattern)
+        {
+            string query = $"DECLARE @include_pattern NVARCHAR(MAX) = '{include}' DECLARE @exclude_pattern NVARCHAR(MAX) = '{exclude}' DECLARE @name_pattern NVARCHAR(255) = '{namePattern}' " +
```
anushakolan: The entire query is constructed as one giant C# interpolated string. This makes it extremely hard to read, review, or safely maintain. Would strongly recommend rewriting using a verbatim @$" … " string with proper indentation or move the SQL into a helper method/constant. This reduces noise from escaping and makes the structure readable.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+
+        /// <summary>
+        /// Builds the query used to get the list of tables with the SQL LIKE
+        /// syntax that will be transformed into entities.
+        /// </summary>
+        /// <param name="include">Pattern for tables that will be included.</param>
+        /// <param name="exclude">Pattern for tables that will be excluded.</param>
+        /// <param name="namePattern">Pattern for naming the entities.</param>
+        /// <returns></returns>
+        public string BuildGetAutoentitiesQuery(string include, string exclude, string namePattern)
```
anushakolan: This implementation only uses `sys.tables` and therefore auto‑entities will not include views.
If views are deliberately excluded in v1, add a comment clarifying the same.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+        {
+            string query = $"DECLARE @include_pattern NVARCHAR(MAX) = '{include}' DECLARE @exclude_pattern NVARCHAR(MAX) = '{exclude}' DECLARE @name_pattern NVARCHAR(255) = '{namePattern}' " +
+                "DECLARE @exclude_invalid_types BIT = 1 SET NOCOUNT ON; WITH include_patterns AS ( SELECT LTRIM(RTRIM(value)) AS pattern " +
```
anushakolan: The include and exclude lists both use the same logic (split → trim → remove empty values).
Right now this logic is duplicated in two different CTEs.
It would be cleaner and safer to put this logic in one shared place so both lists are processed the same way.
This avoids differences in behavior later if someone updates one list but forgets to update the other.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+                "END AS entity_name, CASE WHEN EXISTS ( SELECT 1 FROM sys.columns AS c JOIN sys.types AS ty ON c.user_type_id = ty.user_type_id WHERE c.object_id = a.object_id AND ty.name IN " +
+                "( N'geography', N'geometry', N'hierarchyid', N'sql_variant', N'xml', N'rowversion', N'vector' ) ) THEN 1 ELSE 0 END AS contains_invalid_types FROM eligible_tables AS a WHERE " +
+                "a.is_system_object = 0 AND ( NOT EXISTS (SELECT 1 FROM exclude_patterns) OR NOT EXISTS ( SELECT 1 FROM exclude_patterns AS ep WHERE a.full_name LIKE ep.pattern " +
+                "COLLATE DATABASE_DEFAULT ESCAPE '\\' ) ) AND ( NOT EXISTS (SELECT 1 FROM include_patterns) OR EXISTS ( SELECT 1 FROM include_patterns AS ip WHERE a.full_name LIKE ip.pattern " +
+                "COLLATE DATABASE_DEFAULT ESCAPE '\\' ) ) AND ( @exclude_invalid_types = 0 OR NOT EXISTS ( SELECT 1 FROM sys.columns AS c JOIN sys.types AS ty ON c.user_type_id = ty.user_type_id " +
+                "WHERE c.object_id = a.object_id AND ty.name IN ( N'geography', N'geometry', N'hierarchyid', N'sql_variant', N'xml', N'rowversion', N'vector' ) ) ) ORDER BY a.schema_name, a.object_name;";
+
```
anushakolan, line 743: The query currently has ~8 tightly‑packed clauses, which makes the SQL hard to read and review. Splitting the logic into layered Common Table Expressions (WITH blocks) makes each step clear: normalize include patterns -> normalize exclude patterns  -> gather all tables -> filter eligible objects -> filter invalid types -> final SELECT.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -290,5 +291,34 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+            if (runtimeConfig.Autoentities is not null)
+            {
+                foreach ((string name, Autoentity autoentity) in runtimeConfig.Autoentities.AutoEntities)
+                {
+                    JsonArray? resultArray = await QueryAutoentitiesConfiguration(autoentity);
+                }
+            }
+        }
+
+        public async Task<JsonArray?> QueryAutoentitiesConfiguration(Autoentity autoentity)
```
anushakolan: Nit: `QueryAutoentitiesConfiguration` sounds like it mutates configuration.
Consider renaming to something like `FetchAutoentitiesAsync` / `QueryAutoentitiesAsync` to reflect behavior.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -290,5 +291,34 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+        {
+            string include = string.Join(",", autoentity.Patterns.Include);
+            string exclude = string.Join(",", autoentity.Patterns.Exclude);
+            string namePattern = autoentity.Patterns.Name;
+            string getAutoentitiesQuery = SqlQueryBuilder.BuildGetAutoentitiesQuery(include, exclude, namePattern);
```
anushakolan: Consider logging include/exclude/namePattern before executing the SQL.
This will make debugging broken queries or invalid patterns significantly easier.

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
```diff
@@ -588,5 +589,63 @@ private static async Task SetupTestFixtureAndInferMetadata()
...
+        /// <summary>
+        /// Ensures that the query that returns the tables that will be generated
+        /// into entities from the autoentites configuration return the expected result.
+        /// </summary>
+        [DataTestMethod, TestCategory(TestCategory.MSSQL)]
+        [DataRow(new string[] { "dbo.%book%" }, new string[] { }, "{schema}.{object}.books", new string[] { "book" }, "")]
+        [DataRow(new string[] { "dbo.%publish%" }, new string[] { }, "{schema}.{object}", new string[] { "publish" }, "")]
+        [DataRow(new string[] { "dbo.%book%" }, new string[] { "dbo.%books%" }, "{schema}_{object}_exclude_books", new string[] { "book" }, "books")]
+        [DataRow(new string[] { "dbo.%book%", "dbo.%publish%" }, new string[] { }, "{object}", new string[] { "book", "publish" }, "")]
+        public async Task CheckAutoentitiesQuery(string[] include, string[] exclude, string name, string[] includeObject, string excludeObject)
```
anushakolan, line 602: nit: 2 other test cases that could be added,
 1. empty include + non‑empty exclude
 2.  namePattern containing {schema} only

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
```diff
@@ -588,5 +589,63 @@ private static async Task SetupTestFixtureAndInferMetadata()
...
+            Assert.IsNotNull(resultArray);
+            foreach(JsonObject resultObject in resultArray)
+            {
+                bool includedObjectExists = false;
+                foreach (string included in includeObject)
+                {
+                    if (resultObject[1].ToString().Contains(included))
+                    {
+                        includedObjectExists = true;
+                        Assert.AreEqual(expected: "dbo", actual: resultObject[0].ToString(), "Query does not return expected schema.");
```
anushakolan: The test asserts against resultObject[0], resultObject[1], and resultObject[3] but these indexes correspond to schema/object/entityName positions implicitly. Adding named constants or unpacking into variables would make the intent clearer and reduce risk of index mistakes.

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
```diff
@@ -588,5 +589,63 @@ private static async Task SetupTestFixtureAndInferMetadata()
...
+                        Assert.AreNotEqual(name, resultObject[3].ToString(), "Name returned by query should not include {schema} or {object}.");
+                        if (exclude.Length > 0)
+                        {
+                            Assert.IsTrue(!resultObject[1].ToString().Contains(excludeObject), "Query returns pattern that should be excluded.");
+                        }
+                    }
+                }
+
+                Assert.IsTrue(includedObjectExists, "Query does not return expected object.");
```
anushakolan, line 649: The test checks only `includedObjectExists`, which confirms that one expected object is present, but it does not verify that the full expected set is returned.
For multi‑include cases (e.g.,` {book, publish}`) the query might return only book or might include extra objects, and the current test would still pass.
Consider adding a final `SetEquals / CollectionAssert.AreEquivalent` check to validate the complete result set and detect missing or extra objects.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -290,5 +291,34 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+        protected override async Task GenerateAutoentitiesIntoEntities()
+        {
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            if (runtimeConfig.Autoentities is not null)
+            {
+                foreach ((string name, Autoentity autoentity) in runtimeConfig.Autoentities.AutoEntities)
+                {
+                    JsonArray? resultArray = await QueryAutoentitiesConfiguration(autoentity);
+                }
+            }
```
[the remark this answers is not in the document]
RubenCerna2079: The function is made that way on purpose, as there will be a follow up PR that finishes this method. Will add a comment that explains the situation.

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
RubenCerna2079: 1. Will add a test case for it.
2. That test is not possible because all of the entities need to have a unique name, which the {schema} by itself doesn't provide.

src/Core/Resolvers/MsSqlQueryBuilder.cs
RubenCerna2079: Support for views will come in a later enhancement. Will add a comment that explains this.

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
```diff
@@ -588,5 +589,63 @@ private static async Task SetupTestFixtureAndInferMetadata()
...
+                    {
+                        includedObjectExists = true;
+                        Assert.AreEqual(expected: "dbo", actual: resultObject[0].ToString(), "Query does not return expected schema.");
+                        Assert.AreNotEqual(name, resultObject[3].ToString(), "Name returned by query should not include {schema} or {object}.");
+                        if (exclude.Length > 0)
+                        {
+                            Assert.IsTrue(!resultObject[1].ToString().Contains(excludeObject), "Query returns pattern that should be excluded.");
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
RubenCerna2079: @copilot I don't understand this comment

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -506,7 +506,7 @@ public string BuildStoredProcedureResultDetailsQuery(string databaseObjectName)
         /// 2. are computed based on other columns,
         /// are considered as read only columns. The query combines both the types of read-only columns and returns the list.
         /// </summary>
-        /// <param name="schemaOrDatabaseParamName">Param name of the schema/database.</param>
+        /// <param name="schemaParamName">Param name of the schema/database.</param>
```
Aniruddh25:
```suggestion
        /// <param name="schemaParamName">Param name of the schema.</param>
```

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+        /// Builds the query used to get the list of tables with the SQL LIKE
+        /// syntax that will be transformed into entities.
+        /// NOTE: Currently this query only returns Tables, support for Views will come later.
+        /// </summary>
+        /// <param name="include">Pattern for tables that will be included.</param>
+        /// <param name="exclude">Pattern for tables that will be excluded.</param>
+        /// <param name="namePattern">Pattern for naming the entities.</param>
+        public string BuildGetAutoentitiesQuery(string include, string exclude, string namePattern)
+        {
+            string query = $"DECLARE @include_pattern NVARCHAR(MAX) = '{include}'; DECLARE @exclude_pattern NVARCHAR(MAX) = '{exclude}'; DECLARE @name_pattern NVARCHAR(255) = '{namePattern}'; " +
```
Aniruddh25: avoid setting the parameter value directly here. instead, pass the values as  parameters of the cmd.ExecuteQuery call in the MsSqlExecutor.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -307,6 +307,11 @@ public string GetEntityName(string graphQLType)
         public async Task InitializeAsync()
         {
             System.Diagnostics.Stopwatch timer = System.Diagnostics.Stopwatch.StartNew();
+            if (GetDatabaseType() == DatabaseType.MSSQL)
+            {
+                await GenerateAutoentitiesIntoEntities();
+            }
```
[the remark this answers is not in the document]
RubenCerna2079: This needs to be done before the `_isValidateOnly` because we need to generate all of the entities inside autoentities before the `GenerateDatabaseObjectForEntities` function.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+                "DECLARE @exclude_invalid_types BIT = 1; SET NOCOUNT ON; WITH include_patterns AS ( SELECT LTRIM(RTRIM(value)) AS pattern " +
+                "FROM STRING_SPLIT(ISNULL(@include_pattern, N''), N',') WHERE LTRIM(RTRIM(value)) <> N'' ), " +
+                "exclude_patterns AS ( SELECT LTRIM(RTRIM(value)) AS pattern FROM STRING_SPLIT(ISNULL(@exclude_pattern, N''), N',') " +
```
Aniruddh25: Please also, format this query for readability using: 

`For SQL Server use https://poorsql.com/. Edit the default formatter settings by checking the "trailing commas" checkbox, and adding \s\s\s\s in the "indent string box".`

this is noted here: https://github.com/Azure/data-api-builder/blob/main/CONTRIBUTING.md#modifying-tests

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+                "WHERE LTRIM(RTRIM(value)) <> N'' ), all_tables AS ( SELECT s.name AS schema_name, t.name AS object_name, s.name + N'.' " +
+                "+ t.name AS full_name, N'table' AS object_type, t.object_id FROM sys.tables AS t JOIN sys.schemas AS s ON t.schema_id = s.schema_id " +
+                "WHERE EXISTS ( SELECT 1 FROM sys.key_constraints AS kc WHERE kc.parent_object_id = t.object_id AND kc.type = 'PK' ) ), eligible_tables AS " +
```
Aniruddh25:
```
        WHERE EXISTS
        (
            SELECT 1
            FROM sys.key_constraints AS kc
            WHERE kc.parent_object_id = t.object_id
              AND kc.type = 'PK'
        )
```
this condition is going to select only those tables which have PK constraint defined on them. Is that our intention? I dont think this should be what we do.

src/Core/Resolvers/MsSqlQueryBuilder.cs
```diff
@@ -560,5 +560,39 @@ protected override string BuildPredicates(SqlQueryStructure structure)
...
+        /// syntax that will be transformed into entities.
+        /// NOTE: Currently this query only returns Tables, support for Views will come later.
+        /// </summary>
+        /// <param name="include">Pattern for tables that will be included.</param>
+        /// <param name="exclude">Pattern for tables that will be excluded.</param>
+        /// <param name="namePattern">Pattern for naming the entities.</param>
+        public string BuildGetAutoentitiesQuery(string include, string exclude, string namePattern)
+        {
+            string query = $"DECLARE @include_pattern NVARCHAR(MAX) = '{include}'; DECLARE @exclude_pattern NVARCHAR(MAX) = '{exclude}'; DECLARE @name_pattern NVARCHAR(255) = '{namePattern}'; " +
+                "DECLARE @exclude_invalid_types BIT = 1; SET NOCOUNT ON; WITH include_patterns AS ( SELECT LTRIM(RTRIM(value)) AS pattern " +
```
Aniruddh25: why is @exclude_invalid_types = 1 if the query in the issue initializes this to 0? https://github.com/Azure/data-api-builder/issues/2794#issuecomment-3756630694

Aniruddh25: got the reason, because we want to exclude invalid types by default.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -290,5 +291,40 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+            string getAutoentitiesQuery = SqlQueryBuilder.BuildGetAutoentitiesQuery(include, exclude, namePattern);
+
+            _logger.LogInformation("Query for Autoentities is being executed with the following parameters.");
+            _logger.LogInformation($"Autoentities include pattern: {include}");
+            _logger.LogInformation($"Autoentities exclude pattern: {exclude}");
+            _logger.LogInformation($"Autoentities name pattern: {namePattern}");
+
+            JsonArray? resultArray = await QueryExecutor.ExecuteQueryAsync(
+                sqltext: getAutoentitiesQuery,
+                parameters: null!,
```
Aniruddh25: the parameters here should be filled with the include, exclude and name patterns. This is to avoid SQL Injection.

src/Core/Resolvers/MsSqlQueryBuilder.cs
RubenCerna2079: While I agree with you on this, I think we need to discuss it with @JerryNixon

JerryNixon: DAB requires keys. Tables without PK constraints will be rejected automatically during startup. Views, for the same reason, are not supported because they do not have keys. 

**Explicit validation** in `SqlMetadataProvider.cs`:

```csharp
sourceDefinition.PrimaryKey = [.. pkFields];

if (sourceDefinition.PrimaryKey.Count == 0)
{
           message: $"Primary key not configured on the given database object {tableName}",
}
```

### Why?

- GraphQL by-PK queries need them to fetch individual items
- REST point-get operations (`GET /api/entity/id`) require PK identification
- Update and Delete operations use PKs to target specific records
- Relationship resolution relies on PK/FK mapping

### Aside

While you can configure `key-fields` manually in the config for views or tables without database-enforced PKs, the underlying requirement for a logical primary key still exists for DAB to function properly. Auto-generated entities don't have any per-entity rules to identify the key, so tables without keys have to be skipped.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -293,25 +292,33 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
         }
 
         /// <inheritdoc/>
+        // TODO: Finish implementation of autoentities generation in task #3052
         protected override async Task GenerateAutoentitiesIntoEntities()
         {
-            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            await Task.CompletedTask; // Temporary await to suppress build errors.
+
+            /*RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
```
Aniruddh25: remove comments from code, even if they are required in future

anushakolan:
LGTM!

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
anushakolan: This is not addressed yet, please remove the code comments.

Issue #3094, opened 2026-02-02
[Bug]: Date filters are not supported for `date` in Postgres


marshall-lerner:
### What happened?

Say if the table has a column type as Date, and we're trying to execute a graphql query like below

```
query {
  todos(filter: { dueDate: { gte: "2024-01-01T00:00:00.000Z" } }) {
    items {
      dueDate
    }
  }
}
```
Then we will hit below error "operator does not exist: date >= text":
```
{
  "errors": [
    {
      "message": "42883: operator does not exist: date >= text\r\n\r\nPOSITION: 168",
      "locations": [
        {
          "line": 2,
          "column": 2
        }
      ],
      "path": [
        "todos"
      ],
      "extensions": {
        "code": "DatabaseOperationFailed"
      }
    }
  ],
  "data": null
}
```

It looks like eventually the SQL statements generated will convert the string literal "2024-01-01T00:00:00.000Z" into Text type in PGSQL, and such comparison is not supported. 
 
I can reproduce the same by running below SQL query directly with PGSQL

<img width="804" height="248" alt="Image" src="https://github.com/user-attachments/assets/5de2a786-900f-470f-a7eb-fb54e8e5fe01" />

### Version

0.2025.909.2

### What database are you using?

PostgreSQL

### What hosting model are you using?

Local (including CLI)

### Which API approach are you accessing DAB through?

GraphQL

### Relevant log output

```Text
2026-01-27 18:09:49.491 UTC [82] ERROR:  operator does not exist: timestamp with time zone = text at character 213
2026-01-27 18:09:49.491 UTC [82] HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
2026-01-27 18:09:49.491 UTC [82] STATEMENT:  SELECT COALESCE(jsonb_agg(to_jsonb("subq6")), '[]') AS "data" FROM ( SELECT "table0"."id" AS "id", "table0"."dueDate" AS "dueDate" FROM "public"."Todos" AS "table0" WHERE ($1 = "user_id") AND ("table0"."dueDate" = $2 OR "table0"."dueDate" = $3) ORDER BY "table0"."id" ASC LIMIT 100 ) AS "subq6"
2026-01-27 18:13:37.187 UTC [62] LOG:  checkpoint starting: time
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

marshall-lerner:
I was seeing this for gte, gt, lte, lt. IsNull was working fine for me

PR #3101, opened 2026-02-02
Populate DbType for PostgreSql

marshall-lerner:
## Why make this change?
This PR is to address #3094, which is a bug that happens because the default type conversion for C# to Npgsql is to convert a string to a TEXT field, but this doesn't work when trying to compare a DATE type for PostgreSql databases.

## What is this change?

- My change passes the column name of the Postgres column attempting to be filtered through to the filtering context, and then we override `PopulateDbTypeForParameter` to pass the DBType value through to Npgsql, much like how this is done for MsSql. This _hopefully_ means that Npgsql knows to convert the C# string into something that works with the DATE (and every other) column type.

## How was this tested?
(Currently untested)
- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

With query:
```
query {
  todos(filter: { dueDate: { gte: "2024-01-01T00:00:00.000Z" } }) {
    items {
      dueDate
    }
  }
}
```
Before: 
```
SELECT * FROM public."Todos"
WHERE "Todos"."dueDate" >= '2025-01-01T00:00:00Z'::text;
```

(Hopefully After):
```
SELECT * FROM public."Todos"
WHERE "Todos"."dueDate" >= '2025-01-01T00:00:00Z';
```

### DIFF
--- a/src/Core/Models/GraphQLFilterParsers.cs
+++ b/src/Core/Models/GraphQLFilterParsers.cs
@@ -227,6 +227,7 @@ public class GQLFilterParser
 
                             cosmosQueryStructure.DatabaseObject.Name = sourceName;
                             cosmosQueryStructure.SourceAlias = sourceAlias;
+                            cosmosQueryStructure.EntityName = entityName;
                         }
                     }
                 }
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,251 @@
+# Data API builder (DAB) - Copilot Instructions
+
+## Project Overview
+
+Data API builder (DAB) is an open-source, no-code tool that creates secure, full-featured REST, GraphQL endpoints for databases. It's a CRUD data API engine that runs in a container—on Azure, any other cloud, or on-premises. It also supports creation of DML and Custom MCP tools to build a SQL MCP Server backed by a SQL database.
+
+### Key Technologies
+- **Language**: C# / .NET
+- **.NET Version**: .NET 8.0 (see `global.json`)
+- **Supported Databases**: Azure SQL, SQL Server, SQLDW, Cosmos DB, PostgreSQL, MySQL
+- **API Types**: REST, GraphQL, MCP
+- **Deployment**: Cross-platform (Azure, AWS, GCP, on-premises)
+
+## Project Structure
+
+```
+data-api-builder/
+├── src/
+│   ├── Auth/                        # Authentication logic
+│   ├── Cli/                         # Command-line interface (dab CLI)
+│   ├── Cli.Tests/                   # CLI tests
+│   ├── Config/                      # Configuration handling
+│   ├── Core/                        # Core engine components
+│   ├── Service/                     # Main DAB service/runtime
+│   ├── Service.GraphQLBuilder/      # GraphQL schema builder
+│   ├── Service.Tests/               # Integration tests
+│   └── Azure.DataApiBuilder.sln     # Main solution file
+├── config-generators/               # Config file generation helpers
+├── docs/                            # Documentation
+├── samples/                         # Sample configurations and projects
+├── schemas/                         # JSON schemas for config validation
+├── scripts/                         # Build and utility scripts
+└── templates/                       # Project templates
+```
+
+## Building and Testing
+
+### Prerequisites
+- .NET 8.0 SDK or later
+- Database server for testing (SQL Server, PostgreSQL, MySQL, or Cosmos DB)
+
+### Building the Project
+
+```bash
+# Build the entire solution
+dotnet build src/Azure.DataApiBuilder.sln
+
+# Clean and rebuild
+dotnet clean src/Azure.DataApiBuilder.sln
+dotnet build src/Azure.DataApiBuilder.sln
+```
+
+### Running Tests
+
+DAB uses integration tests that require database instances with proper schemas.
+
+**SQL-based tests:**
+```bash
+# MsSql tests
+dotnet test --filter "TestCategory=MsSql"
+
+# PostgreSQL tests
+dotnet test --filter "TestCategory=PostgreSql"
+
+# MySQL tests
+dotnet test --filter "TestCategory=MySql"
+```
+
+**CosmosDB tests:**
+```bash
+dotnet test --filter "TestCategory=CosmosDb_NoSql"
+```
+
+**Test Configuration:**
+- Test database schemas are in `src/Service.Tests/DatabaseSchema-<engine>.sql`
+- Config files are `src/Service.Tests/dab-config.<engine>.json`
+- Connection strings should use `@env('variable_name')` syntax - never commit connection strings
+
+### Running Locally
+
+1. Open the solution: `src/Azure.DataApiBuilder.sln`
+2. Copy a config file from `src/Service.Tests/dab-config.<engine>.json` to `src/Service/`
+3. Update connection string (use environment variables)
+4. Set `Azure.DataApiBuilder.Service` as startup project
+5. Select debug profile: `MsSql`, `PostgreSql`, `CosmosDb_NoSql`, or `MySql`
+6. Build and run
+
+## Code Style and Conventions
+
+### Formatting
+- **Tool**: `dotnet format` (enforced in CI)
+- **Indentation**: 4 spaces for C# code, 2 spaces for YAML/JSON
+- **Line endings**: LF (Unix-style)
+- **Character encoding**: UTF-8
+- **Trailing whitespace**: Removed
+- **Final newline**: Required
+- Refer to `.\src\.editorconfig` for additional formatting conventions.
+
+### Running Code Formatter
+
+```bash
+# Format all files
+dotnet format src/Azure.DataApiBuilder.sln
+
+# Verify formatting (CI check)
+dotnet format src/Azure.DataApiBuilder.sln --verify-no-changes
+```
+
+### C# Conventions
+- **Usings**: Sort system directives first, no separation between groups
+- **Type preferences**: Use language keywords (`int`, `string`) over BCL types (`Int32`, `String`)
+- **Naming**: Follow standard .NET naming conventions
+- **`this.` qualifier**: Not used unless necessary
+
+### SQL Query Formatting
+When adding or modifying generated SQL queries in tests:
+- **PostgreSQL**: Use https://sqlformat.org/ (remove unnecessary double quotes)
+- **SQL Server**: Use https://poorsql.com/ (enable "trailing commas", indent string: `\s\s\s\s`)
+- **MySQL**: Use https://poorsql.com/ (same as SQL Server, max line width: 100)
+
+## Testing Guidelines
+
+### Test Organization
+- Integration tests validate the engine's query generation and database operations
+- Tests are organized by database type using TestCategory attributes
+- Each database type has its own config file and schema
+
+### Adding New Tests
+- Work within the existing database schema (SQL) or GraphQL schema (CosmosDB)
+- Add tests to the appropriate test class
+- Use base class methods and helpers for engine operations
+- Format any generated SQL queries using the specified formatters
+- Do not commit connection strings to the repository
+
+### Test Database Setup
+1. Create database using the appropriate server
+2. Run the schema script: `src/Service.Tests/DatabaseSchema-<engine>.sql`
+3. Set connection string in config using `@env()` syntax
+4. Run tests for that specific database type
+
+## Configuration Files
+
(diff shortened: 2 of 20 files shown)


marshall-lerner:
@copilot, please rereview the PR

JerryNixon:
### Recommendations

1. **Fix the method signature**:
```csharp
// Change from static to override
public override void PopulateDbTypeForParameter(KeyValuePair<string, DbConnectionParam> parameterEntry, DbParameter parameter)
{
    if (parameterEntry.Value is not null && parameterEntry.Value.DbType is not null)
    {
        parameter.DbType = (DbType)parameterEntry.Value.DbType;
    }
}
```

2. **Verify PrepareDbCommand calls it**: Ensure your `PrepareDbCommand` override in PostgreSqlExecutor properly calls `PopulateDbTypeForParameter` for each parameter.

3. **Update comments only after testing**: The comments in `IQueryExecutor.cs` and `QueryExecutor.cs` should only mention PostgreSQL once the implementation is confirmed working.

4. **Add tests**: This PR needs integration tests to validate:
   - DATE type filtering works
   - TIMESTAMP filtering works  
   - Other type conversions aren't broken

5. **Testing checklist**: Update the PR description to check off the testing boxes once tests are added.

### 🔍 **What to Test**

Based on issue #3094, ensure these queries work:
```graphql
query {
  todos(filter: { dueDate: { gte: "2024-01-01T00:00:00.000Z" } }) {
    items { dueDate }
  }
}
```

The generated SQL should **not** have `::text` casting:
```sql
-- Should generate:
WHERE "Todos"."dueDate" >= $1  -- with DbType.Date

-- NOT:
WHERE "Todos"."dueDate" >= $1::text
```

### Summary

This PR is on the right track but **needs the method signature fix** before it will work. The Copilot review caught a critical issue that must be addressed. Once fixed and tested, this should resolve the PostgreSQL date filtering problem.

src/Core/Resolvers/PostgreSqlExecutor.cs
```diff
@@ -146,6 +147,52 @@ private static bool ShouldManagedIdentityAccessBeAttempted(NpgsqlConnectionStrin
...
+        /// </summary>
+        /// <param name="parameterEntry">The parameter entry containing the value and optional DbType.</param>
+        /// <param name="parameter">The DbParameter to populate.</param>
+        public override void PopulateDbTypeForParameter(KeyValuePair<string, DbConnectionParam> parameterEntry, DbParameter parameter)
+        {
+            if (parameterEntry.Value is not null && parameterEntry.Value.DbType is not null)
+            {
+                parameter.DbType = (DbType)parameterEntry.Value.DbType;
+            }
+        }
```
rgujjul: **Must — Correctness**: The `PopulateDbTypeForParameter` override correctly sets `parameter.DbType` from `parameterEntry.Value.DbType`, but the upstream data flow may not populate `ColumnDefinition.DbType` for PostgreSQL `date` and `timestamp` columns.

The base `PopulateColumnDefinitionWithHasDefaultAndDbType` in `SqlMetadataProvider.cs:1791` calls `TypeHelper.GetDbTypeFromSystemType(columnDefinition.SystemType)`. However, `TypeHelper._systemTypeToDbTypeMap` at `TypeHelper.cs:28-49` explicitly **excludes** `typeof(DateTime)` and `typeof(DateOnly)` — the CLR types that Npgsql maps from PostgreSQL `date` and `timestamp`. The map has a comment explaining the exclusion is due to HotChocolate conflicts.

MsSql solved this via the `MsSqlMetadataProvider.PopulateColumnDefinitionWithHasDefaultAndDbType` override (`MsSqlMetadataProvider.cs:95-135`), which resolves `DbType` from the SQL `DATA_TYPE` column string when `SystemType` is `DateTime`/`DateTimeOffset`. PostgreSQL has no equivalent override.

If `ColumnDefinition.DbType` is `null` → `DbConnectionParam.DbType` is `null` → `PopulateDbTypeForParameter` skips (null check at line 190) → Npgsql defaults to `text` → the bug persists.

```suggestion
        // Option A (recommended — follows MsSql precedent): Add a
        // PopulateColumnDefinitionWithHasDefaultAndDbType override to
        // PostgreSqlMetadataProvider that maps PostgreSQL DATA_TYPE strings
        // to DbType:
        //   "date"                        → DbType.Date
        //   "timestamp without time zone" → DbType.DateTime
        //   "timestamp with time zone"    → DbType.DateTimeOffset
        //   "time without time zone"      → DbType.Time
        //
        // See MsSqlMetadataProvider.cs:123-132 for the reference pattern.

        // Option B: Run the integration tests against a real PostgreSQL instance
        // to verify. If they pass, then there's an alternative DbType population
        // path not traced in this review (e.g., a different schema introspection
        // flow for PostgreSQL).
```

src/Core/Resolvers/PostgreSqlExecutor.cs
```diff
@@ -146,6 +147,52 @@ private static bool ShouldManagedIdentityAccessBeAttempted(NpgsqlConnectionStrin
...
+                    NpgsqlParameter parameter = cmd.CreateParameter();
+                    parameter.ParameterName = parameterEntry.Key;
+                    parameter.Value = parameterEntry.Value.Value ?? DBNull.Value;
+                    PopulateDbTypeForParameter(parameterEntry, parameter);
+                    cmd.Parameters.Add(parameter);
+                }
+            }
+
+            return cmd;
+        }
```
rgujjul: **Should — Maintainability**: The `PrepareDbCommand` override is structurally identical to the base class implementation at `QueryExecutor.cs:336-364`. Since `PopulateDbTypeForParameter` is declared as `public virtual` in the base and this PR uses a proper `public override`, virtual dispatch from the base `PrepareDbCommand` will correctly invoke the PostgreSQL override without needing to duplicate `PrepareDbCommand`.

With this PR, there are now 3 copies of `PrepareDbCommand` (base, MsSql, PostgreSQL). Any future changes to parameter-creation logic — logging, telemetry, error handling, security patches — must be independently applied to all copies.

The MsSql case is different: `MsSqlQueryExecutor` uses `public static` method hiding for `PopulateDbTypeForParameter` (accepting `SqlParameter` instead of `DbParameter`), so it genuinely requires its own `PrepareDbCommand` to call the static method. PostgreSQL's `override` pattern does not have this constraint.

```suggestion
        // Remove the PrepareDbCommand override entirely.
        // Keep only PopulateDbTypeForParameter — the base class
        // PrepareDbCommand will call it via virtual dispatch.
        //
        // If the NpgsqlCommand covariant return type is needed by a
        // caller, add a comment explaining the retention.
```

src/Core/Models/GraphQLFilterParsers.cs
```diff
@@ -778,13 +780,13 @@ private static PredicateOperand GenerateRightOperand(
...
                 .Where(inValue => inValue is not null)
-                .Select(inValue => processLiterals(inValue!, null))
+                .Select(inValue => processLiterals(inValue!, columnName))
                 .ToList();
 
             return new PredicateOperand("(" + string.Join(", ", encodedParams) + ")");
         }
 
-        return new PredicateOperand(processLiteral ? processLiterals(value, null) : value.ToString());
+        return new PredicateOperand(processLiteral ? processLiterals(value, columnName) : value.ToString());
```
rgujjul: **Should — Correctness**: The change from `processLiterals(value, null)` to `processLiterals(value, columnName)` is in shared infrastructure invoked by **all** SQL database types. Before this PR, filter parameters for all databases were created without `DbType`/`SqlDbType`. After this PR:

- **MsSql**: Filter parameters now carry `DbType` AND `SqlDbType` from column metadata. `MsSqlQueryExecutor.PopulateDbTypeForParameter` will now set both on `SqlParameter` — a behavioral change to MsSql that is neither documented nor tested in this PR.
- **MySQL**: `DbConnectionParam` carries unused `DbType` data (base `PopulateDbTypeForParameter` is a no-op). No functional change but the data model changes.
- **CosmosDB**: `CosmosQueryStructure` overrides `MakeDbConnectionParam` entirely — unaffected.

The MsSql behavioral change is likely beneficial (correct types on filter parameters), but "likely safe" ≠ "tested."

Consider running existing MsSql and MySQL integration test suites (especially filter operations on date/time columns) to validate no regressions. Also consider adding a brief comment noting the cross-database scope of the `columnName` forwarding.

src/Service.Tests/SqlTests/GraphQLSupportedTypesTests/PostgreSqlGQLSupportedTypesTests.cs
```diff
@@ -142,6 +141,152 @@ private static string ProperlyFormatTypeTableColumn(string columnName)
...
+            DisplayName = "DATE filter with gt operator")]
+        [DataRow(DATE_TYPE, "gte", "'1999-01-08'", "\"1999-01-08\"", ">=",
+            DisplayName = "DATE filter with gte operator")]
+        [DataRow(DATE_TYPE, "lt", "'9999-12-31'", "\"9999-12-31\"", "<",
+            DisplayName = "DATE filter with lt operator")]
+        [DataRow(DATE_TYPE, "lte", "'2000-06-15'", "\"2000-06-15\"", "<=",
+            DisplayName = "DATE filter with lte operator")]
+        [DataRow(DATE_TYPE, "neq", "'1753-01-01'", "\"1753-01-01\"", "!=",
+            DisplayName = "DATE filter with neq operator")]
+        public async Task PostgreSQL_DateTypeFilterAndOrderBy(
```
rgujjul: **Should — Testing**: The PR modifies the `IN` code path at `GraphQLFilterParsers.cs:783` (`.Select(inValue => processLiterals(inValue!, columnName))`), but the new integration tests only cover scalar operators (`eq`, `neq`, `gt`, `gte`, `lt`, `lte`). The `IN` path creates multiple parameters in a LINQ `.Select()` chain — a distinct code branch that could exhibit the same `date >= text` error if `DbType` is not correctly propagated for each element.

The MsSql equivalent test (`MSSQL_real_graphql_in_filter_expectedValues`) includes `[DataRow(DATETIME_TYPE, ...)]`.

Consider adding `[DataRow(DATE_TYPE, ...)]` and `[DataRow(DATETIME_TYPE, ...)]` entries to the `PGSQL_real_graphql_in_filter_expectedValues` test method.

src/Core/Resolvers/IQueryExecutor.cs
```diff
@@ -165,7 +165,7 @@ public Dictionary<string, object> GetResultProperties(
         public string GetSessionParamsQuery(HttpContext? httpContext, IDictionary<string, DbConnectionParam> parameters, string dataSourceName);
 
         /// <summary>
-        /// Helper method to populate DbType for parameter. Currently DbTypes for parameters are only populated for MsSql.
+        /// Helper method to populate DbType for parameter. Currently DbTypes for parameters are only populated for MsSql and PostgreSql.
```
rgujjul: **Could — Maintainability**: The updated comment now reads: "Currently DbTypes for parameters are only populated for MsSql and PostgreSql." This hardcoded list will become stale as other databases (e.g., MySQL) add their own overrides.

```suggestion
        /// Helper method to populate DbType for parameter. Override in database-specific executors to enable proper parameter typing.
```

src/Core/Resolvers/PostgreSqlExecutor.cs
```diff
@@ -146,6 +147,52 @@ private static bool ShouldManagedIdentityAccessBeAttempted(NpgsqlConnectionStrin
...
+
+        /// <summary>
+        /// Populates the DbType for a PostgreSQL parameter when available.
+        /// This ensures proper type handling for date/time and other types,
+        /// preventing errors like "operator does not exist: date >= text".
+        /// </summary>
+        /// <param name="parameterEntry">The parameter entry containing the value and optional DbType.</param>
+        /// <param name="parameter">The DbParameter to populate.</param>
+        public override void PopulateDbTypeForParameter(KeyValuePair<string, DbConnectionParam> parameterEntry, DbParameter parameter)
```
rgujjul: **Could — Maintainability**: Good choice using proper `public override` for `PopulateDbTypeForParameter` instead of MsSql's `public static` method hiding pattern — this is the architecturally cleaner approach and leverages virtual dispatch correctly. The MsSql pattern exists because `SqlParameter.SqlDbType` requires the concrete type, making a true `override` impractical without casting.

As a minor note: MySQL currently has no `PopulateDbTypeForParameter` override. If MySQL's connector exhibits similar default-to-`text` behavior as Npgsql, the same pattern could be applied there.

PR #3112, opened 2026-02-06
Fix Serialization Side‑Effects by Using Cloned SourceDefinition Objects

Alekhya-Polavarapu:
## Why make this change?
When serializing Database Object details, the original object is being mutated during the process. While this hasn’t caused issues yet, modifying the source object in-place can introduce subtle bugs later if the same object is reused after serialization. Creating a safe, isolated copy avoids unintended side effects.

## What is this change?
This update introduces cloning of the SourceDefinition object and its derived types before serialization.
The flow is as follows:

1. Check whether any columns in the SourceDefinition require escaping.
2. If escaping is needed, a new cloned object is created.
3. A new dictionary is built with escaped column names.
4. All other properties are copied into the cloned object based on the specific object type.
5. The cloned object is then used for serialization, ensuring the original remains untouched.

This approach may introduce some performance overhead, particularly when the sourceDef object is large. An alternative would be to use reflection; however, for derived types, such as ViewDefinition and StoredProcDefinition, the `Columns` property do not expose setters . As a result, reflection cannot provide a consistent or uniform cloning solution across all object types.

## How was this tested?
- [x] Unit Tests


src/Core/Services/MetadataProviders/Converters/DatabaseObjectConverter.cs
```diff
@@ -77,7 +77,8 @@ public override void Write(Utf8JsonWriter writer, DatabaseObject value, JsonSeri
                 // Only escape columns for properties whose type(derived type) is SourceDefinition.
                 if (IsSourceDefinitionOrDerivedClassProperty(prop) && propVal is SourceDefinition sourceDef)
                 {
-                    EscapeDollaredColumns(sourceDef);
+                    // Check if we need to escape any column names
+                    propVal = GetSourceDefinitionWithEscapedColumns(sourceDef);
```
souvikghosh04, line 81: looking at the cloning approach, it seems to be a shallow copy. this should be fine for now but for awareness should be added in notes/comments.

souvikghosh04 (commented):
It will be good to mention about the cloning approach- shallow copy/deep copy so people are aware.

Aniruddh25 (approved):
LGTM after resolving comments from Copilot

Aniruddh25:
No longer needed. To avoid performance overhead, we will not be cloning.

PR #3113, opened 2026-02-07
Add internal text embedding system

robertopc1:
## Why make this change?

Internal DAB system for text embedding/vectorization to support future parameter substitution and Redis semantic search features.

## What is this change?

Configuration (runtime.embeddings)

enabled: Master toggle (default: true)
provider: azure-openai | openai
base-url, api-key, model: Provider connection (supports @env())
api-version, dimensions, timeout-ms: Optional tuning
endpoint.enabled/path/roles: Optional REST endpoint at configured path (default: /embed)
health.enabled/threshold-ms/test-text/expected-dimensions: Health check config
Core Components

IEmbeddingService with TryEmbedAsync() pattern - returns result objects, no exceptions
EmbeddingService - HTTP client with FusionCache L1 (24h TTL, SHA256 hash keys with provider/model included)
EmbeddingsOptionsConverterFactory - Custom JSON deserializer with env var replacement
EmbeddingTelemetryHelper - OpenTelemetry metrics/spans for latency, cache hits, dimensions
EmbeddingController - REST endpoint for /embed with role-based authorization
Health Check Implementation ✅

HealthCheckHelper.UpdateEmbeddingsHealthCheckResultsAsync() - Executes test embedding with threshold validation
Validates response time against health.threshold-ms
Validates dimensions against health.expected-dimensions if specified
Reports Healthy/Unhealthy status in comprehensive health check report
ConfigurationDetails includes Embeddings and EmbeddingsEndpoint status
REST Endpoint Implementation ✅

EmbeddingController serves POST requests at configured path (default: /embed)
Accepts plain text input and returns comma-separated float values (plain text)
Role-based authorization using X-MS-API-ROLE header
In development mode, defaults to anonymous access
In production mode, requires explicit role configuration
Validation & Safety

Constructor validates Azure OpenAI requires model/deployment name
Constructor validates required fields (BaseUrl, ApiKey)
Cache keys include provider and model to prevent collisions
Validates non-empty embedding arrays from API responses
Telemetry Integration

TryEmbedAsync and TryEmbedBatchAsync instrumented with activity spans and metrics
Cache hit/miss tracking in batch operations
API call duration and error tracking
Integration Points

Health check report includes embeddings status in comprehensive checks
Hot reload via EMBEDDINGS_CONFIG_CHANGED event
Startup logging when embeddings configured
CLI: dab configure --runtime.embeddings.*
Code Organization

Azure.DataApiBuilder.Config.ObjectModel.Embeddings - Config models
Azure.DataApiBuilder.Core.Services.Embeddings - Service, telemetry, interface
Azure.DataApiBuilder.Service.Controllers - EmbeddingController

## How was this tested?

- [ ] Integration Tests
- [x ] Unit Tests

## Sample Request(s)

{
  "runtime": {
    "embeddings": {
      "enabled": true,
      "provider": "azure-openai",
      "base-url": "@env('EMBEDDINGS_ENDPOINT')",
      "api-key": "@env('EMBEDDINGS_API_KEY')",
      "model": "text-embedding-ada-002",
      "endpoint": {
        "enabled": true,
        "path": "/embed",
        "roles": ["authenticated"]
      },
      "health": {
        "enabled": true,
        "threshold-ms": 5000,
        "test-text": "health check"
      }
    }
  }
}
Embed Endpoint:

curl -X POST http://localhost:5000/embed \
  -H "Content-Type: text/plain" \
  -H "X-MS-API-ROLE: authenticated" \
  -d "Hello, world!"
  
  Response:

0.123456,0.234567,0.345678,...


src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        if (result.Embedding is null || result.Embedding.Length == 0)
+        {
+            _logger.LogError("Embedding request returned empty result.");
+            return StatusCode((int)HttpStatusCode.InternalServerError, "Failed to generate embedding.");
+        }
+
+        // Return embedding as comma-separated float values (plain text)
+        string embeddingText = string.Join(",", result.Embedding);
+        return Content(embeddingText, MediaTypeNames.Text.Plain);
```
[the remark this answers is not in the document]
robertopc1: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/data-api-builder/pull/3113#discussion_r2776507377)

src/Service/Startup.cs
```diff
@@ -260,7 +262,13 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<GQLFilterParser>();
             services.AddSingleton<RequestValidator>();
             services.AddSingleton<RestService>();
-            services.AddSingleton<HealthCheckHelper>();
+            services.AddSingleton<HealthCheckHelper>(sp =>
+            {
+                ILogger<HealthCheckHelper> logger = sp.GetRequiredService<ILogger<HealthCheckHelper>>();
+                HttpUtilities httpUtility = sp.GetRequiredService<HttpUtilities>();
+                IEmbeddingService? embeddingService = sp.GetService<IEmbeddingService>();
```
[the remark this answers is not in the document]
robertopc1: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/data-api-builder/pull/3113#discussion_r2776507410)

src/Config/Converters/EmbeddingsOptionsConverterFactory.cs
```diff
@@ -0,0 +1,303 @@
...
+                }
+
+                string? propertyName = reader.GetString()?.ToLowerInvariant();
+                reader.Read();
+
+                switch (propertyName)
+                {
+                    case "enabled":
+                        enabled = reader.GetBoolean();
+                        break;
```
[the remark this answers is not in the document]
robertopc1: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/data-api-builder/pull/3113#discussion_r2776507394)

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /// <summary>
+    /// POST endpoint for generating embeddings.
+    /// Accepts plain text body and returns embedding vector as comma-separated floats.
+    /// </summary>
+    /// <param name="route">The route path after the "embed" prefix.</param>
+    /// <returns>Plain text embedding vector or error response.</returns>
+    [HttpPost]
+    [Route("embed/{*route}")]
```
JerryNixon: The route `embed/{*route}` tells ASP.NET “send any request that starts with /embed/ here, and capture everything after it,” but the code then does a strict check that Request.Path must exactly equal endpointOptions.EffectivePath, which means most /embed/... requests will be rejected unless they match one exact configured path; this makes the {*route} catch-all effectively useless and makes configuration easy to get wrong because the real routing rules are split between the attribute and a manual runtime check, pick one approach: either make routing match the configured path using endpoint routing (and remove the manual path equality check), or keep a fixed /embed/... route and use the {*route} value consistently instead of comparing full paths.

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        if (result.Embedding is null || result.Embedding.Length == 0)
+        {
+            _logger.LogError("Embedding request returned empty result.");
+            return StatusCode((int)HttpStatusCode.InternalServerError, "Failed to generate embedding.");
+        }
+
+        // Return embedding as comma-separated float values (plain text)
+        string embeddingText = string.Join(",", result.Embedding.Select(f => f.ToString("G", CultureInfo.InvariantCulture)));
+        return Content(embeddingText, MediaTypeNames.Text.Plain);
```
JerryNixon: Returning the embedding as a comma-separated text/plain string forces every client to implement brittle parsing (split + float parsing + edge cases) and makes it hard to extend the response later (dimensions, model, cache hit, etc.) without inventing a new format, so it would be more robust to return JSON like` { "embedding": [0.1, 0.2, ...], "dimensions": 1536 }` (and only keep text/plain as an optional format if you truly need it).

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -27,20 +29,24 @@ public class HealthCheckHelper
         // Dependencies
         private ILogger<HealthCheckHelper> _logger;
         private HttpUtilities _httpUtility;
+        private IEmbeddingService? _embeddingService;
```
JerryNixon, line 32: HealthCheckHelper only runs embeddings-specific health logic if it actually receives an IEmbeddingService, but because that dependency is optional (= null) it’s easy for the helper to be constructed in a way that never passes it (e.g., created manually rather than by DI), which would silently skip the embeddings health check; to avoid “health check code that never runs,” make sure HealthCheckHelper is created by the DI container so IEmbeddingService gets injected when available (or explicitly resolve it from IServiceProvider).

schemas/dab.draft.schema.json
```diff
@@ -655,6 +655,145 @@
...
+        },
+        "embeddings": {
+          "type": "object",
+          "description": "Configuration for text embedding/vectorization service. Supports OpenAI and Azure OpenAI providers.",
+          "additionalProperties": false,
+          "properties": {
+            "enabled": {
+              "type": "boolean",
+              "description": "Whether the embedding service is enabled. Defaults to true.",
+              "default": true
```
JerryNixon: This may be a change in what we originally discussed but embedding enabled needs to default to false to minimize the surface area exposed. With constraints that if this is true then the other settings are required will make this default necessary. Anyway, default should be `false`.

JerryNixon (approved):
A lot of good work here!

schemas/dab.draft.schema.json
robertopc1: Yes! Thank you for pointing it out

src/Service/HealthCheck/HealthCheckHelper.cs
robertopc1: Yes I'm including this with the changes I'm adding. I'm also adding some tests.

souvikghosh04:
closed as there is another PR- https://github.com/Azure/data-api-builder/pull/3441

Issue #3052, opened 2026-01-13
Create entities in-memory from autoentities


RubenCerna2079:
In this step we need to create the entities and save them in-memory inside of DAB. It is important to note that the entities created through auto-generation will not be serialized into the `entities` section in the config file.

PR #3129, opened 2026-02-13
Create Entities in-memory from Autoentities

RubenCerna2079:
## Why make this change?
- #3052
We need to generate all the entities from the autoentities properties. In order to do this we need to use the query that was previously created and, add the newly generated entities into the runtime so the user can use them.

## What is this change?
- `MsSqlMetadataProvider.cs`: Finish creating the `GenerateAutoentitiesIntoEntities` function so that it uses the query to receive all of the tables and turn them into entities inside the runtimeConfig.
- `RuntimeConfig.cs`: Create new function that adds the new entities to the runtimeConfig. And also change the runtimeConfig to allow for the `entities` property to be missing if the user decides to use the `autoentities` property.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

### DIFF
--- a/src/Config/DatabasePrimitives/DatabaseObject.cs
+++ b/src/Config/DatabasePrimitives/DatabaseObject.cs
@@ -43,13 +43,15 @@ public abstract class DatabaseObject
     public bool Equals(DatabaseObject? other)
     {
         return other is not null &&
-               SchemaName.Equals(other.SchemaName) &&
-               Name.Equals(other.Name);
+               string.Equals(SchemaName, other.SchemaName, StringComparison.OrdinalIgnoreCase) &&
+               string.Equals(Name, other.Name, StringComparison.OrdinalIgnoreCase);
     }
 
     public override int GetHashCode()
     {
-        return HashCode.Combine(SchemaName, Name);
+        return HashCode.Combine(
+            SchemaName is null ? 0 : StringComparer.OrdinalIgnoreCase.GetHashCode(SchemaName),
+            Name is null ? 0 : StringComparer.OrdinalIgnoreCase.GetHashCode(Name));
     }
 
     /// <summary>
--- a/src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
+++ b/src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
@@ -438,6 +438,115 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
         }
 
         /// <summary>
+        /// Test method to verify that many-to-many relationships work correctly when the linking object
+        /// is in a custom schema (not dbo). This test validates that schema names are correctly compared
+        /// using case-insensitive comparison, which is important for SQL Server where schema names are
+        /// case-insensitive.
+        /// </summary>
+        [DataRow("mySchema.TEST_SOURCE_LINK", "mySchema", "TEST_SOURCE_LINK", DisplayName = "Linking object with custom schema")]
+        [DataRow("MYSCHEMA.TEST_SOURCE_LINK", "MYSCHEMA", "TEST_SOURCE_LINK", DisplayName = "Linking object with uppercase custom schema")]
+        [DataRow("myschema.test_source_link", "myschema", "test_source_link", DisplayName = "Linking object with lowercase schema and table")]
+        [DataTestMethod]
+        public void TestRelationshipWithLinkingObjectInCustomSchema(
+            string linkingObject,
+            string expectedSchema,
+            string expectedTable
+        )
+        {
+            // Creating an EntityMap with two sample entities
+            Dictionary<string, Entity> entityMap = GetSampleEntityMap(
+                sourceEntity: "SampleEntity1",
+                targetEntity: "SampleEntity2",
+                sourceFields: new string[] { "sourceField" },
+                targetFields: new string[] { "targetField" },
+                linkingObject: linkingObject,
+                linkingSourceFields: new string[] { "linkingSourceField" },
+                linkingTargetFields: new string[] { "linkingTargetField" }
+            );
+
+            RuntimeConfig runtimeConfig = new(
+                Schema: "UnitTestSchema",
+                DataSource: new DataSource(DatabaseType: DatabaseType.MSSQL, ConnectionString: "", Options: null),
+                Runtime: new(
+                    Rest: new(),
+                    GraphQL: new(),
+                    Mcp: new(),
+                    Host: new(null, null)
+                ),
+                Entities: new(entityMap)
+            );
+
+            // Mocking EntityToDatabaseObject - entities are in the custom schema as well
+            MockFileSystem fileSystem = new();
+            FileSystemRuntimeConfigLoader loader = new(fileSystem);
+            RuntimeConfigProvider provider = new(loader) { IsLateConfigured = true };
+            RuntimeConfigValidator configValidator = new(provider, fileSystem, new Mock<ILogger<RuntimeConfigValidator>>().Object);
+            Mock<ISqlMetadataProvider> _sqlMetadataProvider = new();
+
+            Dictionary<string, DatabaseObject> mockDictionaryForEntityDatabaseObject = new()
+            {
+                {
+                    "SampleEntity1",
+                    new DatabaseTable(expectedSchema, "TEST_SOURCE1")
+                },
+                {
+                    "SampleEntity2",
+                    new DatabaseTable(expectedSchema, "TEST_SOURCE2")
+                }
+            };
+
+            _sqlMetadataProvider.Setup(x => x.EntityToDatabaseObject).Returns(mockDictionaryForEntityDatabaseObject);
+
+            // To mock the schema name and dbObjectName for linkingObject
+            _sqlMetadataProvider.Setup(x =>
+                x.ParseSchemaAndDbTableName(linkingObject)).Returns((expectedSchema, expectedTable));
+
+            string discard;
+            _sqlMetadataProvider.Setup(x => x.TryGetExposedColumnName(It.IsAny<string>(), It.IsAny<string>(), out discard)).Returns(true);
+
+            Mock<IMetadataProviderFactory> _metadataProviderFactory = new();
+            _metadataProviderFactory.Setup(x => x.GetMetadataProvider(It.IsAny<string>())).Returns(_sqlMetadataProvider.Object);
+
+            // Mock ForeignKeyPair to be defined in DB with the custom schema
+            // The schema comparison should be case-insensitive
+            // Use concrete DatabaseTable instances with differing casing so that
+            // Moq relies on DatabaseTable.Equals for argument matching.
+            // Linking table uses lowercase to ensure case-insensitive comparison is working.
+            DatabaseTable expectedLinkingTable = new(expectedSchema.ToLowerInvariant(), expectedTable.ToLowerInvariant());
+            DatabaseTable expectedSource1Table = new(expectedSchema.ToUpperInvariant(), "TEST_SOURCE1");
+            DatabaseTable expectedSource2Table = new(expectedSchema.ToUpperInvariant(), "TEST_SOURCE2");
+
+            _sqlMetadataProvider.Setup(x =>
+                x.VerifyForeignKeyExistsInDB(
+                    expectedLinkingTable,
+                    expectedSource1Table
+                )).Returns(true);
+
+            _sqlMetadataProvider.Setup(x =>
+                x.VerifyForeignKeyExistsInDB(
+                    expectedLinkingTable,
+                    expectedSource2Table
+                )).Returns(true);
+
+            // Validation should pass with custom schema
+            configValidator.ValidateRelationships(runtimeConfig, _metadataProviderFactory.Object);
+
+            // Verify that VerifyForeignKeyExistsInDB is never called with 'dbo' schema,
+            // guarding against the original dbo-fallback regression.
+            _sqlMetadataProvider.Verify(x =>
+                x.VerifyForeignKeyExistsInDB(
+                    It.Is<DatabaseTable>(t => string.Equals(t.SchemaName, "dbo", StringComparison.OrdinalIgnoreCase)),
+                    It.IsAny<DatabaseTable>()
+                ), Times.Never);
+
+            _sqlMetadataProvider.Verify(x =>
+                x.VerifyForeignKeyExistsInDB(
+                    It.IsAny<DatabaseTable>(),
+                    It.Is<DatabaseTable>(t => string.Equals(t.SchemaName, "dbo", StringComparison.OrdinalIgnoreCase))
+                ), Times.Never);
+        }
+
+        /// <summary>
         /// Test method to check that an exception is thrown when the relationship is one-many
         /// or many-one (determined by the linking object being null), while both SourceFields
         /// and TargetFields are null in the config, and the foreignKey pair between source and target


src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -287,25 +292,28 @@ public RuntimeConfig(
         };
 
         _entityNameToDataSourceName = new Dictionary<string, string>();
-        if (Entities is null)
+        if ((Entities is null || Entities.Entities.Count == 0) && Autoentities is null)
         {
             throw new DataApiBuilderException(
-                message: "entities is a mandatory property in DAB Config",
+                message: "Configuration file should contain either at least the Entities or Autoentities property",
```
Aniruddh25:
```suggestion
                message: "Configuration file should contain either at least the entities or autoentities property",
```

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -294,7 +294,86 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+                }
+            }
+
+            if (addedEntities == 0)
+            {
+                _logger.LogWarning("No new entities were generated from the autoentity patterns defined in the configuration.");
+            }
+            else
+            {
+                _runtimeConfigProvider.AddNewEntitiesToConfig(entities);
```
Aniruddh25: Dont we have access to the RuntimeConfigLoader here?

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -411,4 +411,13 @@ private static RuntimeConfig HandleCosmosNoSqlConfiguration(string? schema, Runt
 
         return runtimeConfig;
     }
+
+    public void AddNewEntitiesToConfig(Dictionary<string, Entity> entities)
```
Aniruddh25: Change the function name to "AddMergedEntitiesToConfig" since the entities passed in here, are all the entities - original + entities found from autoentities.

src/Service/Startup.cs
```diff
@@ -1140,11 +1140,13 @@ private async Task<bool> PerformOnConfigChangeAsync(IApplicationBuilder app)
                 // Now that the configuration has been set, perform validation of the runtime config
                 // itself.
 
+                // TODO: Add this check at the end of generating the new entities and skip this one only if it is 
                 runtimeConfigValidator.ValidateConfigProperties();
 
                 if (runtimeConfig.IsDevelopmentMode())
                 {
                     // Running only in developer mode to ensure fast and smooth startup in production.
+                    // TODO: Add this check at the end of generating the new entities and skip this one only if it is
```
Aniruddh25: Is there a task tracking this move for validation?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5231,6 +5231,148 @@ public async Task TestGraphQLIntrospectionQueriesAreNotImpactedByDepthLimit()
...
+                ),
+                Entities: new(useEntities ? entityMap : new Dictionary<string, Entity>()),
+                Autoentities: new RuntimeAutoentities(autoentityMap)
+            );
+
+            File.WriteAllText(CUSTOM_CONFIG_FILENAME, configuration.ToJson());
+
+            string[] args = new[] { $"--ConfigFileName={CUSTOM_CONFIG_FILENAME}" };
+
+            using (TestServer server = new(Program.CreateWebHostBuilder(args)))
```
Aniruddh25, line 5389: Can you also retrieve the serviceProvider -> RuntimeConfigProvider and the RuntimeConfig to check the entities member has 3 entities with useentity and 1 when not using previous entities?

src/Service/Startup.cs
RubenCerna2079: Yes, https://github.com/Azure/data-api-builder/issues/3131

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -294,7 +294,86 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
         /// <inheritdoc/>
         protected override async Task GenerateAutoentitiesIntoEntities()
         {
-            await Task.CompletedTask;
+            int addedEntities = 0;
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            Dictionary<string, Entity> entities = (Dictionary<string, Entity>)_entities;
```
JerryNixon: This is not so much a technical problem as a code behavior problem. This line breaks the contract of `IReadOnlyDictionary`. For some reason, a developer typed this as read-only. Why? Presumably to protect it from mutation. But that developer made a mistake by assigning it via `ToDictionary()`.

So, your cast works today because _entities is assigned via .ToDictionary() in the base class, but the declared type is IReadOnlyDictionary. If someone later CORRECTS it to ReadOnlyDictionary, this line will fail with InvalidCastException at runtime with no compile-time hint.

That's not all. _entities only has entities for this datasource (it is filtered in the SqlMetadataProvider constructor). So when you pass it to AddNewEntitiesToConfig that code will replace the full set with a subset, dropping entities from other datasources. 🫨

**Suggestion**: The easiest is to build a local dictionary from runtimeConfig.Entities.Entities (the full set), merge auto-generated entities into it, and pass that to AddNewEntitiesToConfig. That avoids both problems.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -294,7 +294,86 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+
+            foreach ((string autoentityName, Autoentity autoentity) in runtimeConfig.Autoentities.AutoEntities)
+            {
+                JsonArray? resultArray = await QueryAutoentitiesAsync(autoentity);
+                if (resultArray is null)
+                {
+                    continue;
+                }
+
+                foreach (JsonObject resultObject in resultArray!)
```
JerryNixon: The `!` is unnecessary here. You already guard for null on line 308 with continue, so the compiler knows resultArray is not null at this point. Remove the ! to keep things clean. I know these are easy to just keep everywhere, but it actually can introduce hard to debug problems later.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -294,7 +294,86 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+                {
+                    // Extract the entity name, schema, and database object name from the query result.
+                    // The SQL query returns these values with placeholders already replaced.
+                    string entityName = resultObject["entity_name"]!.ToString();
+                    string schemaName = resultObject["schema"]!.ToString();
+                    string objectName = resultObject["object"]!.ToString();
```
JerryNixon: If the SQL result has a row where one of these keys is missing, the `!` will cause a NullReferenceException before reaching your IsNullOrWhiteSpace guard below. Use ?.ToString() and let the null check catch it:

```csharp
string? entityName = resultObject["entity_name"]?.ToString();
```

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -294,7 +294,86 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+            {
+                _logger.LogWarning("No new entities were generated from the autoentity patterns defined in the configuration.");
+            }
+            else
+            {
+                _runtimeConfigProvider.AddNewEntitiesToConfig(entities);
+            }
         }
 
         public async Task<JsonArray?> QueryAutoentitiesAsync(Autoentity autoentity)
```
JerryNixon, line 384: AutoentityPatterns.Exclude is string[]? in the constructor. If someone passes null, string.Join throws ArgumentNullException. Coalesce it: autoentity.Patterns.Exclude ?? Array.Empty<string>().

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -411,4 +411,13 @@ private static RuntimeConfig HandleCosmosNoSqlConfiguration(string? schema, Runt
...
+    {
+        RuntimeConfig newRuntimeConfig = _configLoader.RuntimeConfig! with
+        {
+            Entities = new(entities)
+        };
```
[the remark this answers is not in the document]
JerryNixon: Copilot is right on this one. This replaces the entire entity collection. Combined with entities being sourced from _entities (which is per-datasource), this drops entities from other datasources. Merge into the existing set instead. Something like this:

```csharp
var merged = new Dictionary<string, Entity>(_configLoader.RuntimeConfig!.Entities.Entities);
foreach (var kvp in entities) merged[kvp.Key] = kvp.Value;
```

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -287,25 +292,28 @@ public RuntimeConfig(
         };
 
         _entityNameToDataSourceName = new Dictionary<string, string>();
-        if (Entities is null)
+        if ((Entities is null || Entities.Entities.Count == 0) && Autoentities is null)
```
JerryNixon: On line 276 you already set `this.Entities = Entities ?? new RuntimeEntities(...)`, so Entities is null can never be true here. The null branch is dead code. SImplify to `if (Entities.Entities.Count == 0 && Autoentities is null)`

src/Service/Startup.cs
JerryNixon: What's the mystery? "If it is..." What?! I suggest you finish the sentence and also link to the task here. It's okay to have a TODO in a PR. Just always think that you might hand this off to another developer, so don't use too much shorthand.

src/Service.Tests/Configuration/ConfigurationTests.cs
JerryNixon: Per @Aniruddh25's feedback: after the server starts, resolve RuntimeConfigProvider from the service provider and assert the entity count. Expect 3 when useEntities=true (2 manual + 1 auto-generated) and 1 when useEntities=false. This proves the autoentity pipeline actually ran, rather than relying only on REST/GraphQL responses which could pass for the wrong reasons.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -39,7 +39,7 @@ public abstract class SqlMetadataProvider<ConnectionT, DataAdapterT, CommandT> :
         private readonly DatabaseType _databaseType;
 
         // Represents the entities exposed in the runtime config.
-        private IReadOnlyDictionary<string, Entity> _entities;
+        protected IReadOnlyDictionary<string, Entity> _entities;
```
JerryNixon: This was changed from private to protected so the subclass can cast and mutate it. If you take the approach of building a local dictionary from runtimeConfig.Entities instead of mutating _entities directly, you can revert this back to private and preserve encapsulation.

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: The `Entities is null` is referencing the entities object that is passed into the constructor, not the one that is saved inside the constructor. But I agree that this can be simplified.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
RubenCerna2079: Not directly since the RuntimeConfigLoader is a private variable inside the Provider

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
RubenCerna2079: It can be null in the constructor but the final variable will never be null. You can see in the constructor that if `Exclude` is null then it will create it as an empty variable.
AutoentityPatterns.cs
```
if (Exclude is not null)
{
    this.Exclude = Exclude;
    UserProvidedExcludeOptions = true;
}
else
{
    this.Exclude = [];
}
```

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: I take it back, we actually need to check if both the Entities configuration that was brought is null and if the count is 0 because there are cases where the Entities configuration is not null but it is empty. Which was causing some tests to fail.

src/Core/Configurations/RuntimeConfigProvider.cs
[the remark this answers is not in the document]
RubenCerna2079: I was able to confirm that each child configs in a multi data source scenario has its own `_entities` which is what will be updated with the new changes I am implementing.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -292,10 +292,10 @@ public RuntimeConfig(
         };
 
         _entityNameToDataSourceName = new Dictionary<string, string>();
-        if ((Entities is null || Entities.Entities.Count == 0) && Autoentities is null)
+        if (Entities is null && this.Entities.Entities.Count == 0 && Autoentities is null)
```
Aniruddh25: why did we modify the condition from II to && 

Should be `Entities.null || Entities.Count == 0`

Also, there shouldnt be Entities.Entities.Count right? why is Entitites twice?

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5231,6 +5232,148 @@ public async Task TestGraphQLIntrospectionQueriesAreNotImpactedByDepthLimit()
...
+            using (HttpClient client = server.CreateClient())
+            {
+                // Act
+                HttpRequestMessage restRequest = new(HttpMethod.Get, "/api/publishers");
```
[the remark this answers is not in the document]
Aniruddh25: how was this resolved, consider adding a `using` directive, so this gets disposed off

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: This change was made so that Autoentities works with multiple data-sources. Since if the parent config file doesn't have any entities or autoentities it will throw an error if we use `Entities.null || Entities.Entities.Count`.
We use Entities twice because the first one is the `RuntimeEntities` object which it contains an `Entities` dictionary inside of it, we are trying to see if the dictionary has any elements inside of it.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -287,25 +304,38 @@ public RuntimeConfig(
         };
 
         _entityNameToDataSourceName = new Dictionary<string, string>();
-        if (Entities is null)
+        if (Entities is null && this.Entities.Entities.Count == 0 &&
```
Aniruddh25, line 307: This condition will throw NullException

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -287,25 +304,38 @@ public RuntimeConfig(
...
                 statusCode: HttpStatusCode.UnprocessableEntity,
                 subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError);
         }
 
-        foreach (KeyValuePair<string, Entity> entity in Entities)
+        if (Entities is not null)
+        {
+            foreach (KeyValuePair<string, Entity> entity in Entities)
+            {
+                _entityNameToDataSourceName.TryAdd(entity.Key, this.DefaultDataSourceName);
```
Aniruddh25, line 320: Why are we always adding to the default data source name here? Shouldnt we add it to the data source name found from the Datasource files? If there is no name found there, we should use default data source name.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -245,6 +247,21 @@ public bool TryGetEntityNameFromPath(string entityPathName, [NotNullWhen(true)]
         return _entityPathNameToEntityName.TryGetValue(entityPathName, out entityName);
     }
 
+    public bool TryAddEntityNameToDataSourceName(string entityName)
+    {
+        return _entityNameToDataSourceName.TryAdd(entityName, this.DefaultDataSourceName);
+    }
+
+    public bool TryAddEntityNameToDataSourceName(string entityName, string autoEntityDefinition)
```
Aniruddh25: nit: rename the function to inform we are adding entity name from AutoentityDefinition

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: I am not following why this would throw a NullException

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -451,6 +486,17 @@ public DataSource GetDataSourceFromEntityName(string entityName)
...
     }
 
+    /// <summary>
+    /// Gets datasourceName from AutoentityNameToDatasourceName dictionary.
+    /// </summary>
+    /// <param name="autoentityName">autoentityName</param>
+    /// <returns>DataSourceName</returns>
+    public string GetDataSourceNameFromAutoentityName(string autoentityName)
+    {
+        CheckAutoentityNamePresent(autoentityName);
```
Aniruddh25: instead of 2 functions, why cant we use `TryGet` function? if false, you can throw exception.

src/Config/ObjectModel/RuntimeConfig.cs
RubenCerna2079: We add the DefaultDataSourceName because it is the guid that is created for each of the data sources. So in order to differentiate between different data sources we use this since each data source creates their own RuntimeConfig, which in turn has its unique DefaultDataSourceName. If anything, the name might make it a bit confusing since there is no other data source name that can be used.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -292,9 +292,93 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
         }
 
         /// <inheritdoc/>
-        protected override async Task GenerateAutoentitiesIntoEntities()
+        protected override async Task GenerateAutoentitiesIntoEntities(Dictionary<string, Autoentity>? autoentities)
         {
-            await Task.CompletedTask;
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            Dictionary<string, Entity> entities = new();
```
Aniruddh25: define entities where its needed. Since autoentities could still be null, defining here is pre-mature if we are just going to return.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -292,9 +292,93 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+                    {
+                        throw new DataApiBuilderException(
+                            message: $"Entity with name '{entityName}' already exists. Cannot create new entity from autoentity pattern with definition-name '{autoentityName}'.",
+                            statusCode: HttpStatusCode.BadRequest,
+                            subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
+                    }
+
+                    if (runtimeConfig.IsRestEnabled)
+                    {
+                        _logger.LogInformation("[{entity}] REST path: {globalRestPath}/{entityRestPath}", entityName, runtimeConfig.RestPath, entityName);
```
Aniruddh25, line 365: Why is only Rest information being logged?

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -292,9 +292,93 @@ private bool TryResolveDbType(string sqlDbTypeName, out DbType dbType)
...
+            Dictionary<string, Entity> entities = new();
+            if (autoentities is null)
+            {
+                return;
+            }
+
+            foreach ((string autoentityName, Autoentity autoentity) in autoentities)
+            {
+                int addedEntities = 0;
+                JsonArray? resultArray = await QueryAutoentitiesAsync(autoentity);
```
Aniruddh25, line 307: when querying, are we asking the right datasource to query from? How is that information being passed to QueryAutoentitiesAsync?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -307,9 +310,10 @@ public string GetEntityName(string graphQLType)
         public async Task InitializeAsync()
         {
             System.Diagnostics.Stopwatch timer = System.Diagnostics.Stopwatch.StartNew();
+
             if (GetDatabaseType() == DatabaseType.MSSQL)
             {
-                await GenerateAutoentitiesIntoEntities();
+                await GenerateAutoentitiesIntoEntities(new Dictionary<string, Autoentity>(Autoentities));
```
Aniruddh25: why is the need to create new dictionary of `Autoentities` here? cant you simply pass the Readonly dictionary?

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
```diff
@@ -101,4 +101,34 @@ public async Task FailLoadMultiDataSourceConfigDuplicateEntities(string configPa
...
+        // Convert the modified JSON object back to a JSON string
+        string resultJson = baseJsonObject.ToString();
+
+        IFileSystem fs = new MockFileSystem(new Dictionary<string, MockFileData>() { { "dab-config.json", new MockFileData(resultJson) } });
+
+        FileSystemRuntimeConfigLoader loader = new(fs);
+
+        Assert.IsTrue(loader.TryLoadConfig("dab-config.json", out RuntimeConfig runtimeConfig), "Should successfully load config");
+        Assert.IsTrue(runtimeConfig.SqlDataSourceUsed, "Should have Sql data source");
+        Assert.AreEqual(expectedEntities, runtimeConfig.Entities.Entities.Count, "Default datasource should be of root file database type.");
```
Aniruddh25: The error message here seems to be not accurate.

src/Service.Tests/Multidab-config.MsSql.json
Aniruddh25, line 1522: Can you also test by naming the child data sources explicitly? I dont know if we provide the datasource name in the sample files or rely on the default one.

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
Aniruddh25: does this `runtimeConfig.Entities.Entities.Count` include entities matching from all the data source files - including the ones that matched from autoentities ?

Aniruddh25 (changes requested):
Waiting on addressing some questions..

src/Config/ObjectModel/RuntimeConfig.cs
Aniruddh25: please rename `Entities` to `entities` since its a function argument

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
RubenCerna2079: Only the Rest information is being logged to mimic the behavior that is already present in the `SqlMetadataProvider` which also outputs this message for all the already existing entities.

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
RubenCerna2079: Yes, it includes all of the entities from all data source files including from autoentities

src/Service.Tests/Multidab-config.MsSql.json
Aniruddh25: there is no explicit data source name property

Aniruddh25 (approved):
LGTM after comments resolved and pipelines succeed.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
RubenCerna2079: Yes, we are asking the right datasource to query from. It does this by running the `SqlMetadataProvider` for each of the datasources that exist inside of DAB, so if the user has autoentities in that specific datasource then it will query it.

Issue #2663, opened 2025-04-22
[Bug]: REST PATCH does not work with Identity keys


JerryNixon:
The URL requires an Id, which is incompatible with new records without a key.

JerryNixon:
Still a good idea

Issue #2992, opened 2025-11-26
[Bug]: Fix failing MS SQL Integration tests in the pipeline


anushakolan:
### What happened?

The MS SQL Integration tests in the pipeline are failing randomly with 503 Service Unavailable exceptions. 
Ex: https://github.com/Azure/data-api-builder/pull/2943/checks?check_run_id=56413554625

The count of failures and the tests that are failing seem to be random. We need to understand what is causing the flakiness and fix it.

### Version

Latest pipeline version

### What database are you using?

Azure SQL

### Relevant log output

```Text
Failures: 

Class Initialization method Azure.DataApiBuilder.Service.Tests.UnitTests.ODataASTVisitorUnitTests.SetupAsync threw exception. Microsoft.Data.SqlClient.SqlException: Connection Timeout Expired.  The timeout period elapsed while attempting to consume the pre-login handshake acknowledgement.  This could be because the pre-login handshake failed or the server was unable to respond back in time.  The duration spent while attempting to connect to this server was - [Pre-Login] initialization=5179; handshake=19; .


Assert.AreEqual failed. Expected:<BadRequest>. Actual:<ServiceUnavailable>.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

aaronburtle:
Another example: https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=15514&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=f2696cc9-a35b-5158-31dc-9689759fe573

aaronburtle:
We extend this issue to the other intermittent failures in the MSSQL pipeline seen for example here:

https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=18815&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=2f6b9d04-a34c-50be-eed8-94ef82d4279a

`  Failed Autoentities fail due to rest path [1 s]
  Error Message:
   TestCleanup method Azure.DataApiBuilder.Service.Tests.Configuration.ConfigurationTests.CleanupAfterEachTest threw exception. System.IO.IOException: The process cannot access the file 'D:\a\1\s\src\out\tests\net8.0\custom-config.json' because it is being used by another process..
  Stack Trace:
[...]

PR #3150, opened 2026-02-21
Add keyless REST PUT/PATCH support for entities with auto-generated primary keys

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2663

One test change to help the pipeline pass relates to https://github.com/Azure/data-api-builder/issues/2992

## Add keyless REST PUT/PATCH support for entities with auto-generated primary keys

We now support keyless PUT/PATCH requests so long as the keys that are missing are auto-generated by the entity in question. If the URL is keyless, and the request body is keyless, then all of the keys needed for the request must be auto-generated. If the request body contains keys, then we allow the request only if the request body contains all non auto-generated keys. So for example, if the entity has a composite key, then it would be allowed to omit any auto-generated keys from that composite key, but the non auto-generated keys must be included. This is handled with new validation logic, which will now account for keyless PUT/PATCH, simply validating that any non auto-generated keys are included.

* In `OpenApiDocumentor`, keyless `PUT` and `PATCH` operations are now documented for entities with auto-generated primary keys on the base entity path. These use the `_NoAutoPK` request body schema and only advertise `201 Created` responses. A missing `AddQueryParameters` helper method was also added.

* Stored procedure entities are unaffected.
* Mutation engine is modified to detect the truly keyless operation in case of Upsert/UpsertIncremental to then trigger the Insert workflow instead otherwise the SQL statement will be incorrectly generated.

## How was this tested?

We modify existing tests to match the new behavior, and then add a test for keyless PUT/PATCH.

## Sample Request(s)

```
PUT /api/Book

{
    "title": "My New Book",
    "publisher_id": 1234
}

PATCH /api/Book
{
    "title": "Another New Book",
    "publisher_id": 5678
}
```


src/Core/Services/RestService.cs
```diff
@@ -70,6 +70,16 @@ RequestValidator requestValidator
             ISqlMetadataProvider sqlMetadataProvider = _sqlMetadataProviderFactory.GetMetadataProvider(dataSourceName);
             DatabaseObject dbObject = sqlMetadataProvider.EntityToDatabaseObject[entityName];
 
+            // When a PUT or PATCH request arrives without a primary key in the route,
+            // convert it to an Insert operation. This supports entities with identity/auto-generated
+            // keys where the caller doesn't know the key value before creation.
+            if (string.IsNullOrEmpty(primaryKeyRoute) &&
+                (operationType is EntityActionOperation.Upsert ||
+                 operationType is EntityActionOperation.UpsertIncremental))
+            {
```
JerryNixon, line 79: The conversion to Insert is not gated by “auto-generated PK”. Should it?

aaronburtle: Good question, this is OK because our `RequestValidator` will handle the validation side of things and we keep a clean separation of responsibility.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -43,7 +43,9 @@ public class OpenApiDocumentor : IOpenApiDocumentor
         private const string GETONE_DESCRIPTION = "Returns an entity.";
         private const string POST_DESCRIPTION = "Create entity.";
         private const string PUT_DESCRIPTION = "Replace or create entity.";
+        private const string PUT_KEYLESS_DESCRIPTION = "Create entity (keyless). For entities with auto-generated primary keys, creates a new record without requiring the key in the URL.";
         private const string PATCH_DESCRIPTION = "Update or create entity.";
+        private const string PATCH_KEYLESS_DESCRIPTION = "Create entity (keyless). For entities with auto-generated primary keys, creates a new record without requiring the key in the URL.";
```
Aniruddh25: the constant value is same as above `PUT_KEYLESS_DESCRIPTION`. Use a single constant

src/Core/Services/RestService.cs
Aniruddh25: This is indeed a good question. @aaronburtle, I checked with Copilot, and it says "So: autogenerated PKs do not bypass the URL [primaryKeyRoute] requirement for update/upsert/delete. For insert, a URL PK is disallowed regardless, in [RequestValidator.cs:224-232].

   `public void ValidatePrimaryKey(RestRequestContext context)` see this function.

You need to fix the exception that we throw there, in cases of Patch/Put to not check for existence of primaryKey in the route if they are autogenerated.

aaronburtle: > This is indeed a good question. @aaronburtle, I checked with Copilot, and it says "So: autogenerated PKs do not bypass the URL [primaryKeyRoute] requirement for update/upsert/delete. For insert, a URL PK is disallowed regardless, in [RequestValidator.cs:224-232].
> 
> `public void ValidatePrimaryKey(RestRequestContext context)` see this function.
> 
> You need to fix the exception that we throw there, in cases of Patch/Put to not check for existence of primaryKey in the route if they are autogenerated.

We should be OK, because we are changing the Patch/Put to an insert operation already in `RestService`

`            if (string.IsNullOrEmpty(primaryKeyRoute) &&
                (operationType is EntityActionOperation.Upsert ||
                 operationType is EntityActionOperation.UpsertIncremental))
            {
                operationType = EntityActionOperation.Insert;
            }`
            
 And then when we do the validation on lines 224-232 of `RequestValidator` we expect there to be no Primary Key, so keyless Put Patch should be fine, since we've converted to insert and then have the right behavior
 
 `                case EntityActionOperation.Insert:
                    if (!isPrimaryKeyRouteEmpty)
                    {
                            message: PRIMARY_KEY_INVALID_USAGE_ERR_MESSAGE,
                            statusCode: HttpStatusCode.BadRequest,
                            subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
                    }`

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
```diff
@@ -120,9 +120,18 @@ public async Task TestAzureLogAnalyticsFlushServiceSucceed(string message, LogLe
 
         _ = Task.Run(() => flusherService.StartAsync(tokenSource.Token));
 
-        await Task.Delay(2000);
+        // Poll until the log appears (the flusher service needs time to dequeue and upload)
+        int maxWaitMs = 10000;
+        int pollIntervalMs = 100;
```
Aniruddh25, line 125: why start with such a low `pollIntervalMs` if the original delay was 2000ms? Perhaps start with that.

src/Core/Services/RestService.cs
Aniruddh25: looks like the conversion to Insert operation happens before we do the Request Validation. In that case, we should be ok with validation. But, then we should be doing the conversion to Insert only when the PKs are autogenerated. Otherwise, we should NOT convert into Insert and continue to throw the exception.

aaronburtle: I think we want this behavior because they could include the PKs in the body of the request, in which case we should still be able to do the insert. The logic would be to convert to Insert even for non auto-gen, then the Insert validation logic applies, and if it is non-auto-gen it must include the keys in the body, and only if those keys are missing from the body as well, we would throw.

aaronburtle: I know this is effectively going beyond the scope of this underlying issue, however, we get this additional functionality for free just by letting the request validator handle the insert op, so seems worth it.

aaronburtle: So after speaking offline we have identified the correct behavior we want, will provide it here for clarity.

In a truly keyless PUT/PATCH, we will convert to insert op, that means no key in URL, no key in body. The Insert validation will just work in this case. It will succeed if we are auto-gen, and fail if we are not. This will require identifying when we have keys in the body when the URL is missing keys.

For PUT/PATCH where the URL is keyless, but we have the keys in the body, we want to use Upsert semantics, so we will not convert to an insert. This will require new validation logic.

aaronburtle: Actually now that we have pushed the logic into the validator side of things, we will not convert any ops but just have the validator handle all of the logic here basically. No op conversion needed just correctly validating when an upsert is doable with auto gen keys that were not provided

src/Core/Services/RestService.cs
```diff
@@ -92,12 +89,6 @@ RequestValidator requestValidator
             QueryString? query = GetHttpContext().Request.QueryString;
             string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString();
 
-            string requestBody = string.Empty;
```
Aniruddh25, line 85: Why do you need to move this up? I dont see any new usage of requestBody between this position and the one above.

src/Core/Services/RequestValidator.cs
```diff
@@ -348,8 +348,12 @@ public void ValidateInsertRequestContext(InsertRequestContext insertRequestCtx)
         /// and vice versa.
         /// </summary>
         /// <param name="upsertRequestCtx">Upsert Request context containing the request body.</param>
+        /// <param name="primaryKeyInUrl">When true the primary key was provided in the URL route
+        /// and PK columns in the body are skipped (original behaviour). When false the primary key
+        /// is expected in the request body, so non-auto-generated PK columns must be present and
+        /// the full composite key (if applicable) must be supplied.</param>
         /// <exception cref="DataApiBuilderException"></exception>
-        public void ValidateUpsertRequestContext(UpsertRequestContext upsertRequestCtx)
+        public void ValidateUpsertRequestContext(UpsertRequestContext upsertRequestCtx, bool primaryKeyInUrl = true)
```
Aniruddh25: nit: the variable name should be `isPrimaryKeyInUrl` since its a boolean.

src/Service.Tests/Configuration/Telemetry/AzureLogAnalyticsTests.cs
aaronburtle: We talked about this offline but just documenting here. This is in case the log appears early we can move on, and then in the case that it doesnt appear early we will continue polling until it does with a timeout of 10s.

src/Core/Services/RestService.cs
aaronburtle: this didnt need to move, putting it back, think co-pilot refactored this for some reason I just didn't notice.

Aniruddh25 (approved):

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -541,6 +541,84 @@ await queryExecutor.ExecuteQueryAsync(
...
+                            {
+                                throw _dabExceptionWithTransactionErrorMessage;
+                            }
+
+                            if (isReadPermissionConfiguredForRole && !isDatabasePolicyDefinedForReadAction)
+                            {
+                                IEnumerable<string> allowedExposedColumns = _authorizationResolver.GetAllowedExposedColumns(context.EntityName, roleName, EntityActionOperation.Read);
+                                foreach (string columnInResponse in insertResultRow.Columns.Keys)
+                                {
+                                    if (!allowedExposedColumns.Contains(columnInResponse))
```
Aniruddh25, line 600: This piece is common between what we do for Insert below. Can this be moved to a single function that handles Insert Transaction - which can be called here and below?

You would need to put Update/UpdateIncremental that is below in a function too

Aniruddh25: Resolving this for now, but this should be fixed to avoid duplicate code.

PR #3151, opened 2026-02-21
Add On-Behalf-Of (OBO) user-delegated authentication for SQL

### LINKED ISSUE #3122: [OBO] JWT Validation, Identity Extraction, and Cache‑Key Construction
- Validate inbound JWT:
1.   signature
2.   issuer
3.   audience (must match DAB)
4.   expiry

- Extract identity claims:
1. use oid when present
2. fallback to sub when oid is missing (guest/B2B)
3. extract tid

- Reject request if neither oid nor sub exists → 401
- Compute authorization‑context hash from permission‑affecting claims
- Construct cache key exactly as defined:
`(oid OR sub) + tid + authorization‑context hash`

- Enforce design‑mandated validation rules:
1. database-type must be mssql
2. database-audience required
3. token-cache-duration-minutes ∈ [1,59]
4. runtime.caching.enabled must be false

Acceptance Criteria / Tests
- missing oid+sub → 401
- invalid config fails deterministically

anushakolan:
### Why make this change?

Closes #3122

Data API Builder currently supports managed identity authentication to Azure SQL, where the DAB service itself authenticates using its own identity. However, this approach doesn't support user-level authorization scenarios where:

- Row-Level Security (RLS) policies need to filter data based on the actual end user's identity
- Database audit logs should reflect the actual user who performed the operation
- Fine-grained permissions need to be enforced per-user at the database level

This PR introduces On-Behalf-Of (OBO) user-delegated authentication for SQL Server, enabling DAB to exchange the incoming user's access token for a downstream SQL token using the OAuth 2.0 OBO flow. This allows the actual user's identity to flow through to the database, enabling true user-level authorization.

### What is this change?

**Overview**

Implements the OAuth 2.0 On-Behalf-Of (OBO) flow for SQL Server authentication, allowing DAB to:

1. Accept a user's access token from the incoming request
2. Exchange it for a SQL Server-scoped token using MSAL
3. Use that token to authenticate to SQL Server as the actual user

**Configuration Schema**

Added new `user-delegated-auth` configuration section under data-source:

```json
{
  "data-source": {
    "database-type": "mssql",
    "connection-string": "@env('SQL_CONNECTION_STRING')",
    "user-delegated-auth": {
      "enabled": true,
      "provider": "EntraId",
      "database-audience": "https://database.windows.net"
    }
  }
}
```

**New Components**

- `IOboTokenProvider` - Interface for OBO token exchange
- `OboSqlTokenProvider` - Implementation that performs the OBO token exchange using MSAL with FusionCache-based token caching
- `IMsalClientWrapper` / `MsalClientWrapper` - Wrapper around MSAL's ConfidentialClientApplication for testability
- Schema updates - Added `user-delegated-auth` to `dab.draft.schema.json`

**Key Implementation Details**

1. **Token Caching**: Uses FusionCache (L1 in-memory cache) to avoid repeated OBO token exchanges for the same user. Cache keys are derived from user identity claims (oid/sub + tenant ID). Tokens are cached with a 5-minute buffer before expiration.
2. **Scope**: Requests token for `https://database.windows.net/.default` scope
3. **Connection String**: When OBO is configured, the `AccessToken` property is set on `SqlConnection` with the exchanged token
4. **Validation**:
   - OBO is only supported for `mssql` database type
   - `database-audience` is required when OBO is enabled
   - Required environment variables: `DAB_OBO_CLIENTID`, `DAB_OBO_CLIENTSECRET`, `DAB_OBO_TENANTID`

#### Unit Tests

Added comprehensive unit tests in `OboSqlTokenProviderUnitTests.cs`:

1. `GetAccessTokenAsync_ReturnsToken_WhenOboSucceeds` - Happy path - successful token exchange
2. `GetAccessTokenAsync_ThrowsException_WhenOboFails` - Error handling when MSAL fails
3. `GetAccessTokenAsync_UsesCorrectScope` - Verifies SQL scope is requested
4. `GetAccessTokenAsync_PassesUserAssertionCorrectly` - Verifies user token is passed correctly
5. `GetAccessTokenAsync_HandlesNullResult` - Null result handling
6. `GetAccessTokenAsync_HandlesEmptyAccessToken` - Empty token handling

Added validation tests in `ConfigValidationUnitTests.cs`:

1. `ValidateUserDelegatedAuthConfig_ValidConfig_NoErrors` - Valid OBO config passes validation
2. `ValidateUserDelegatedAuthConfig_InvalidDatabaseAudience_ReturnsError` - Missing/empty database-audience fails when enabled
3. `ValidateUserDelegatedAuthConfig_NonMsSql_ReturnsError` - OBO only supported for mssql
4. `ValidateUserDelegatedAuthConfig_NullOptions_NoErrors` - No OBO config is valid
5. `ValidateUserDelegatedAuthConfig_DisabledWithMissingFields_NoErrors` - Disabled OBO doesn't require all fields

#### Manual End-to-End Testing (Azure)

**Test Environment Setup**

Deployed DAB to Azure Container Apps with OBO authentication enabled, connected to an Azure SQL Database configured with Row-Level Security (RLS). The RLS policy filters data based on the authenticated user's email address.

**Test Scenarios Validated**

1. **User Identity Propagation**: Verified that the actual user's identity flows through to the database. When User A calls the API, they only see their own data (1 row out of 6 total rows in the table).

2. **Multi-User Isolation**: Tested with 3 different users - each user only sees rows where their email matches the `UserEmail` column, confirming RLS works correctly with OBO tokens.

3. **Authentication Enforcement**:
   - Requests without a token receive 403 Forbidden
   - Requests with invalid/expired tokens receive 401 Unauthorized

4. **Authorization Enforcement**: Users with valid tokens but no SQL database permissions receive a 500 error with login failed message (SQL Error 18456).

5. **Token Caching**: Confirmed FusionCache token caching is working via Information-level logging:
   - First request logs: `OBO token cache MISS for subject {oid}. Acquiring new token from Azure AD.`
   - Subsequent requests log: `OBO token cache HIT for subject {oid}.`
   - Each unique user creates a separate cache entry keyed by their subject ID.

6. **Database Audit Trail**: Verified that SQL Server audit logs show the actual user's identity (e.g., `user@example.com`) rather than the managed identity, enabling proper audit compliance.

**Comparison with Managed Identity**

Deployed a separate instance using managed identity (without OBO) to compare behavior:

| Scenario | Result |
|----------|--------|
| Managed Identity + RLS | Returns 0 rows (MI identity doesn't match any user) |
| Managed Identity + RLS disabled | Returns all 6 rows |
| OBO + RLS | Returns only the authenticated user's row(s) |

This validates that OBO correctly enables per-user data isolation at the database level.

### DIFF
--- a/src/Core/Models/RestRequestContexts/RestRequestContext.cs
+++ b/src/Core/Models/RestRequestContexts/RestRequestContext.cs
@@ -78,6 +78,12 @@ public abstract class RestRequestContext
     public NameValueCollection ParsedQueryString { get; set; } = new();
 
     /// <summary>
+    /// Raw query string from the HTTP request (URL-encoded).
+    /// Used to preserve encoding for special characters in query parameters.
+    /// </summary>
+    public string RawQueryString { get; set; } = string.Empty;
+
+    /// <summary>
     /// String holds information needed for pagination.
     /// Based on request this property may or may not be populated.
     /// </summary>
--- a/src/Core/Parsers/RequestParser.cs
+++ b/src/Core/Parsers/RequestParser.cs
@@ -113,14 +113,32 @@ namespace Azure.DataApiBuilder.Core.Parsers
                         context.FieldsToBeReturned = context.ParsedQueryString[key]!.Split(",").ToList();
                         break;
                     case FILTER_URL:
-                        // save the AST that represents the filter for the query
-                        // ?$filter=<filter clause using microsoft api guidelines>
-                        string filterQueryString = $"?{FILTER_URL}={context.ParsedQueryString[key]}";
-                        context.FilterClauseInUrl = sqlMetadataProvider.GetODataParser().GetFilterClause(filterQueryString, $"{context.EntityName}.{context.DatabaseObject.FullName}");
+                        // Use raw (URL-encoded) filter value to preserve special characters like &
+                        string? rawFilterValue = ExtractRawQueryParameter(context.RawQueryString, FILTER_URL);
+                        // If key exists in ParsedQueryString but not in RawQueryString, something is wrong
+                        if (rawFilterValue is null)
+                        {
+                            throw new DataApiBuilderException(
+                                message: $"Unable to extract {FILTER_URL} parameter from query string.",
+                                statusCode: HttpStatusCode.BadRequest,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                        }
+
+                        context.FilterClauseInUrl = sqlMetadataProvider.GetODataParser().GetFilterClause($"?{FILTER_URL}={rawFilterValue}", $"{context.EntityName}.{context.DatabaseObject.FullName}");
                         break;
                     case SORT_URL:
-                        string sortQueryString = $"?{SORT_URL}={context.ParsedQueryString[key]}";
-                        (context.OrderByClauseInUrl, context.OrderByClauseOfBackingColumns) = GenerateOrderByLists(context, sqlMetadataProvider, sortQueryString);
+                        // Use raw (URL-encoded) orderby value to preserve special characters
+                        string? rawSortValue = ExtractRawQueryParameter(context.RawQueryString, SORT_URL);
+                        // If key exists in ParsedQueryString but not in RawQueryString, something is wrong
+                        if (rawSortValue is null)
+                        {
+                            throw new DataApiBuilderException(
+                                message: $"Unable to extract {SORT_URL} parameter from query string.",
+                                statusCode: HttpStatusCode.BadRequest,
+                                subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                        }
+
+                        (context.OrderByClauseInUrl, context.OrderByClauseOfBackingColumns) = GenerateOrderByLists(context, sqlMetadataProvider, $"?{SORT_URL}={rawSortValue}");
                         break;
                     case AFTER_URL:
                         context.After = context.ParsedQueryString[key];
@@ -282,6 +300,39 @@ namespace Azure.DataApiBuilder.Core.Parsers
         private static bool IsNull(string value)
         {
             return string.IsNullOrWhiteSpace(value) || string.Equals(value, "null", StringComparison.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// Extracts the raw (URL-encoded) value of a query parameter from a query string.
+        /// Preserves special characters like & in filter values (e.g., %26 stays as %26).
+        /// 
+        /// IMPORTANT: This method assumes the input queryString is a raw, URL-encoded query string
+        /// where special characters in parameter values are encoded (e.g., & is %26, space is %20).
+        /// It splits on unencoded '&' characters which are parameter separators in the URL standard.
+        /// If the queryString has already been decoded, this method will not work correctly.
+        /// </summary>
+        /// <param name="queryString">Raw URL-encoded query string (e.g., "?$filter=title%20eq%20%27A%26B%27")</param>
+        /// <param name="parameterName">The parameter name to extract (e.g., "$filter")</param>
+        /// <returns>The raw encoded value of the parameter, or null if not found</returns>
+        internal static string? ExtractRawQueryParameter(string queryString, string parameterName)
+        {
+            if (string.IsNullOrWhiteSpace(queryString))
+            {
+                return null;
+            }
+
+            // Split on '&' which are parameter separators in properly URL-encoded query strings.
+            // Any '&' characters within parameter values will be encoded as %26.
+            foreach (string param in queryString.TrimStart('?').Split('&'))
+            {
+                int idx = param.IndexOf('=');
+                if (idx >= 0 && param.Substring(0, idx).Equals(parameterName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return idx < param.Length - 1 ? param.Substring(idx + 1) : string.Empty;
+                }
+            }
+
+            return null;
         }
     }
 }
--- a/src/Core/Services/RestService.cs
+++ b/src/Core/Services/RestService.cs
@@ -174,6 +174,7 @@ namespace Azure.DataApiBuilder.Core.Services
 
                 if (!string.IsNullOrWhiteSpace(queryString))
                 {
+                    context.RawQueryString = queryString;
                     context.ParsedQueryString = HttpUtility.ParseQueryString(queryString);
                     RequestParser.ParseQueryString(context, sqlMetadataProvider);
                 }
@@ -277,6 +278,7 @@ namespace Azure.DataApiBuilder.Core.Services
                     // So, $filter will be treated as any other parameter (inevitably will raise a Bad Request)
                     if (!string.IsNullOrWhiteSpace(queryString))
                     {
+                        context.RawQueryString = queryString;
                         context.ParsedQueryString = HttpUtility.ParseQueryString(queryString);
                     }
 
--- a/src/Service.Tests/SqlTests/RestApiTests/Find/DwSqlFindApiTests.cs
+++ b/src/Service.Tests/SqlTests/RestApiTests/Find/DwSqlFindApiTests.cs
@@ -222,6 +222,12 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.RestApiTests.Find
                 $"FOR JSON PATH, INCLUDE_NULL_VALUES"
             },
             {
+                "FindTestWithFilterContainingSpecialCharacters",
+                $"SELECT * FROM { _integrationTableName } " +
+                $"WHERE title = 'SOME%CONN' " +
+                $"FOR JSON PATH, INCLUDE_NULL_VALUES"
+            },
+            {
                 "FindTestWithPrimaryKeyContainingForeignKey",
                 $"SELECT [id], [content] FROM reviews " +
                 $"WHERE id = 567 AND book_id = 1 FOR JSON PATH, INCLUDE_NULL_VALUES, WITHOUT_ARRAY_WRAPPER"
--- a/src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
+++ b/src/Service.Tests/SqlTests/RestApiTests/Find/FindApiTestBase.cs
@@ -694,6 +694,23 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.RestApiTests.Find
         }
 
         /// <summary>
+        /// Tests the REST Api for Find operation with a filter containing special characters
+        /// that need to be URL-encoded. Uses existing book with '%' character (SOME%CONN).
+        /// This validates that the fix for the double-decoding issue is working correctly.
+        /// </summary>
+        [TestMethod]
+        public async Task FindTestWithFilterContainingSpecialCharacters()
+        {
+            // Testing with SOME%CONN - the %25 is URL-encoded %
+            await SetupAndRunRestApiTest(
+                primaryKeyRoute: string.Empty,
+                queryString: "?$filter=title%20eq%20%27SOME%25CONN%27",
+                entityNameOrPath: _integrationEntityName,
+                sqlQuery: GetQuery(nameof(FindTestWithFilterContainingSpecialCharacters))
+            );
+        }
+
+        /// <summary>
         /// Tests the REST Api for Find operation where we compare one field
         /// to the bool returned from another comparison.
         /// </summary>
--- a/src/Service.Tests/SqlTests/RestApiTests/Find/MsSqlFindApiTests.cs
+++ b/src/Service.Tests/SqlTests/RestApiTests/Find/MsSqlFindApiTests.cs
@@ -229,6 +229,12 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.RestApiTests.Find
                 $"FOR JSON PATH, INCLUDE_NULL_VALUES"
             },
             {
+                "FindTestWithFilterContainingSpecialCharacters",
+                $"SELECT * FROM { _integrationTableName } " +
+                $"WHERE title = 'SOME%CONN' " +
+                $"FOR JSON PATH, INCLUDE_NULL_VALUES"
+            },
+            {
                 "FindTestWithPrimaryKeyContainingForeignKey",
                 $"SELECT [id], [content] FROM reviews " +
                 $"WHERE id = 567 AND book_id = 1 FOR JSON PATH, INCLUDE_NULL_VALUES, WITHOUT_ARRAY_WRAPPER"
--- a/src/Service.Tests/SqlTests/RestApiTests/Find/MySqlFindApiTests.cs
+++ b/src/Service.Tests/SqlTests/RestApiTests/Find/MySqlFindApiTests.cs
@@ -398,6 +398,18 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.RestApiTests.Find
                 "
             },
             {
+                "FindTestWithFilterContainingSpecialCharacters",
+                @"
+                    SELECT JSON_ARRAYAGG(JSON_OBJECT('id', id, 'title', title, 'publisher_id', publisher_id)) AS data
+                    FROM (
+                        SELECT *
+                        FROM " + _integrationTableName + @"
+                        WHERE title = 'SOME%CONN'
+                        ORDER BY id asc
+                    ) AS subq
+                "
+            },
+            {
                 "FindTestWithFilterQueryStringBoolResultFilter",
                 @"
                     SELECT JSON_ARRAYAGG(JSON_OBJECT('id', id, 'title', title, 'publisher_id', publisher_id)) AS data
--- a/src/Service.Tests/SqlTests/RestApiTests/Find/PostgreSqlFindApiTests.cs
+++ b/src/Service.Tests/SqlTests/RestApiTests/Find/PostgreSqlFindApiTests.cs
@@ -412,6 +412,17 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.RestApiTests.Find
                   ) AS subq"
             },
             {
+                "FindTestWithFilterContainingSpecialCharacters",
+                @"
+                  SELECT json_agg(to_jsonb(subq)) AS data
+                  FROM (
+                      SELECT *
+                      FROM " + _integrationTableName + @"
+                      WHERE title = 'SOME%CONN'
+                      ORDER BY id asc
+                  ) AS subq"
+            },
+            {
                 "FindTestWithPrimaryKeyContainingForeignKey",
                 @"
                     SELECT to_jsonb(subq) AS data
--- a/src/Service.Tests/UnitTests/RequestParserUnitTests.cs
+++ b/src/Service.Tests/UnitTests/RequestParserUnitTests.cs
@@ -1,0 +1,80 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.DataApiBuilder.Core.Parsers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Azure.DataApiBuilder.Service.Tests.UnitTests
+{
+    /// <summary>
+    /// Test class for RequestParser utility methods.
+    /// Specifically tests the ExtractRawQueryParameter method which preserves
+    /// URL encoding for special characters in query parameters.
+    /// </summary>
+    [TestClass]
+    public class RequestParserUnitTests
+    {
+        /// <summary>
+        /// Tests that ExtractRawQueryParameter correctly extracts URL-encoded
+        /// parameter values, preserving special characters like ampersand (&).
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("?$filter=region%20eq%20%27filter%20%26%20test%27", "$filter", "region%20eq%20%27filter%20%26%20test%27", DisplayName = "Extract filter with encoded ampersand (&)")]
+        [DataRow("?$filter=title%20eq%20%27A%20%26%20B%27&$select=id", "$filter", "title%20eq%20%27A%20%26%20B%27", DisplayName = "Extract filter with ampersand and other params")]
+        [DataRow("?$select=id&$filter=name%20eq%20%27test%27", "$filter", "name%20eq%20%27test%27", DisplayName = "Extract filter when not first parameter")]
+        [DataRow("?$orderby=name%20asc", "$orderby", "name%20asc", DisplayName = "Extract orderby parameter")]
+        [DataRow("?param1=value1&param2=value%26with%26ampersands", "param2", "value%26with%26ampersands", DisplayName = "Extract parameter with multiple ampersands")]
+        [DataRow("$filter=title%20eq%20%27test%27", "$filter", "title%20eq%20%27test%27", DisplayName = "Extract without leading question mark")]
+        [DataRow("?$filter=", "$filter", "", DisplayName = "Extract empty filter value")]
+        [DataRow("?$filter=name%20eq%20%27test%3D123%27", "$filter", "name%20eq%20%27test%3D123%27", DisplayName = "Extract filter with encoded equals sign (=)")]
+        [DataRow("?$filter=url%20eq%20%27http%3A%2F%2Fexample.com%3Fkey%3Dvalue%27", "$filter", "url%20eq%20%27http%3A%2F%2Fexample.com%3Fkey%3Dvalue%27", DisplayName = "Extract filter with encoded URL (: / ?)")]
+        [DataRow("?$filter=text%20eq%20%27A%2BB%27", "$filter", "text%20eq%20%27A%2BB%27", DisplayName = "Extract filter with encoded plus sign (+)")]
+        [DataRow("?$filter=value%20eq%20%2750%25%27", "$filter", "value%20eq%20%2750%25%27", DisplayName = "Extract filter with encoded percent sign (%)")]
+        [DataRow("?$filter=tag%20eq%20%27%23hashtag%27", "$filter", "tag%20eq%20%27%23hashtag%27", DisplayName = "Extract filter with encoded hash (#)")]
+        [DataRow("?$filter=expr%20eq%20%27a%3Cb%3Ed%27", "$filter", "expr%20eq%20%27a%3Cb%3Ed%27", DisplayName = "Extract filter with encoded less-than and greater-than (< >)")]
+        public void ExtractRawQueryParameter_PreservesEncoding(string queryString, string parameterName, string expectedValue)
+        {
+            // Call the internal method directly (no reflection needed)
+            string? result = RequestParser.ExtractRawQueryParameter(queryString, parameterName);
+
+            Assert.AreEqual(expectedValue, result,
+                $"Expected '{expectedValue}' but got '{result}' for parameter '{parameterName}' in query '{queryString}'");
+        }
+
+        /// <summary>
+        /// Tests that ExtractRawQueryParameter returns null when parameter is not found.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("?$filter=test", "$orderby", DisplayName = "Parameter not in query string")]
+        [DataRow("", "$filter", DisplayName = "Empty query string")]
+        [DataRow(null, "$filter", DisplayName = "Null query string")]
+        [DataRow("?otherParam=value", "$filter", DisplayName = "Different parameter")]
+        public void ExtractRawQueryParameter_ReturnsNull_WhenParameterNotFound(string? queryString, string parameterName)
+        {
+            // Call the internal method directly (no reflection needed)
+            string? result = RequestParser.ExtractRawQueryParameter(queryString, parameterName);
+
+            Assert.IsNull(result,
+                $"Expected null but got '{result}' for parameter '{parameterName}' in query '{queryString}'");
+        }
+
+        /// <summary>
+        /// Tests that ExtractRawQueryParameter handles edge cases correctly:
+        /// - Duplicate parameters (returns first occurrence)
+        /// - Case-insensitive parameter name matching
+        /// - Malformed query strings with unencoded ampersands
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("?$filter=value&$filter=anothervalue", "$filter", "value", DisplayName = "Multiple same parameters - returns first")]
+        [DataRow("?$FILTER=value", "$filter", "value", DisplayName = "Case insensitive parameter matching")]
+        [DataRow("?param=value1&value2", "param", "value1", DisplayName = "Value with unencoded ampersand after parameter")]
+        public void ExtractRawQueryParameter_HandlesEdgeCases(string queryString, string parameterName, string expectedValue)
+        {
+            // Call the internal method directly (no reflection needed)
+            string? result = RequestParser.ExtractRawQueryParameter(queryString, parameterName);
+
+            Assert.AreEqual(expectedValue, result,
+                $"Expected '{expectedValue}' but got '{result}' for parameter '{parameterName}' in query '{queryString}'");
+        }
+    }
+}


src/Service/Startup.cs
```diff
@@ -252,6 +254,50 @@ public void ConfigureServices(IServiceCollection services)
...
+                            !string.IsNullOrEmpty(ds.UserDelegatedAuth.DatabaseAudience))
+                        {
+                            string? clientId = Environment.GetEnvironmentVariable(UserDelegatedAuthOptions.AZURE_CLIENT_ID_ENV_VAR);
+                            string? tenantId = Environment.GetEnvironmentVariable(UserDelegatedAuthOptions.AZURE_TENANT_ID_ENV_VAR);
+                            string? clientSecret = Environment.GetEnvironmentVariable(UserDelegatedAuthOptions.AZURE_CLIENT_SECRET_ENV_VAR);
+
+                            if (string.IsNullOrEmpty(clientId) || string.IsNullOrEmpty(tenantId) || string.IsNullOrEmpty(clientSecret))
+                            {
+                                // Validation will catch this and throw appropriate error
+                                break;
```
souvikghosh04: This seems a bit odd to me as we are passing the resbonsibility to something at a later stage or relying on the caller to validate. Can this be handled in a better way?

src/Core/Resolvers/OboSqlTokenProvider.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        // Acquire new token via OBO
+        try
+        {
+            string[] scopes = [$"{databaseAudience.TrimEnd('/')}/.default"];
+
+            AuthenticationResult result = await _msalClient.AcquireTokenOnBehalfOfAsync(
+                scopes,
+                incomingJwtAssertion,
```
souvikghosh04: Is there a way to validate the `incomingJwtAssertion` before calling acquiring Token or is it guranteed to be valid? I mean the audience, issuer, expiry etc. should be validated before attempting the token?

src/Service/Startup.cs
```diff
@@ -252,6 +254,50 @@ public void ConfigureServices(IServiceCollection services)
...
+                            }
```
souvikghosh04: I think this loop breaks on the first data source with OBO enabled. If DAB supports multiple databases each with OBO configured (possibly with different database-audience values or different scopes), this won't work correctly. The current design assumes a single OBO-configured data source. may be should document this as limitation/consideration or have a fix for this?

src/Core/Resolvers/OboSqlTokenProvider.cs
```diff
@@ -0,0 +1,240 @@
...
+                cancellationToken);
+
+            CachedToken newCachedToken = new(
+                AccessToken: result.AccessToken,
+                ExpiresOn: result.ExpiresOn,
+                CachedAt: DateTimeOffset.UtcNow);
+
+            _tokenCache[cacheKey] = newCachedToken;
```
souvikghosh04: How are we cleaning or evicting the expired tokens? this will keep getting added and added?

src/Service/Startup.cs
anushakolan: Updated the code, - the `IsOboConfigured()` method now handles validation upfront by checking environment variables first (fast path), and only registers the service if everything is properly configured.

src/Core/Resolvers/OboSqlTokenProvider.cs
anushakolan: The `incomingJwtAssertion` is guaranteed to be valid by the time it reaches this code:

ASP.NET Core JWT Bearer middleware validates the token upstream (issuer, audience, expiry, signature) before the request ever reaches DAB. Invalid tokens are rejected with 401 before any controller/resolver code runs.

MSAL validates the assertion as part of the OBO exchange with Azure AD. If there's any issue (token expired during request processing, wrong audience, etc.), MSAL throws `MsalServiceException` which we catch and convert to `DataApiBuilderException` with 401 Unauthorized.

Adding duplicate validation here would be redundant and could introduce drift if validation rules change between our code and the middleware/MSAL.

src/Service/Startup.cs
anushakolan: Refactored the code.
The current design actually supports multiple databases with different database-audience values. Here's how it works:

1. A single `IOboTokenProvider` is registered using one Azure AD app registration (same client ID, tenant ID, secret)
2. Each data source specifies its own database-audience in the config
3. When a query executes, `MsSqlQueryExecutor` passes the specific data source's audience to `GetAccessTokenOnBehalfOfAsync()`
4. MSAL can acquire tokens for different resource servers using the same client - the audience/scope determines which resource the token is for
The old code with the break statement was refactored - `IsOboConfigured()` now just checks if ANY data source has OBO enabled. The actual audience selection happens at query execution time, not at DI registration.

src/Core/Resolvers/OboSqlTokenProvider.cs
anushakolan: Good catch! I've added a periodic cleanup mechanism:

1. `CleanupExpiredTokensIfNeeded()` runs every 100 cache operations
2. Removes tokens where `ExpiresOn` has passed
3. Uses thread-safe Interlocked.Increment for the counter
4. Logs when tokens are removed for observability

src/Service/Startup.cs
```diff
@@ -252,6 +254,50 @@ public void ConfigureServices(IServiceCollection services)
...
+                                .Build();
+
+                            IMsalClientWrapper msalWrapper = new MsalClientWrapper(msalClient);
+                            ILogger<OboSqlTokenProvider> oboLogger = serviceProvider.GetRequiredService<ILogger<OboSqlTokenProvider>>();
+                            return new OboSqlTokenProvider(msalWrapper, oboLogger);
+                        }
+                    }
+                }
+
+                return null!;
```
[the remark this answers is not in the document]
JerryNixon: Better pattern is to only register the service when enabled (or register IOboTokenProvider? / inject IServiceProvider and GetService, or use TryAdd + an implementation that cleanly no-ops).

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -175,13 +201,42 @@ public override async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection
...
+                    {
+                        sqlConn.AccessToken = oboToken;
+                        return;
+                    }
+
+                    // OBO is enabled but we couldn't get a token (e.g., missing Bearer token in request)
+                    // This is an error during request processing - we must not fall back to managed identity
+                    throw new DataApiBuilderException(
+                        message: "User-delegated authentication is enabled but no valid user context is available.",
+                        statusCode: HttpStatusCode.Unauthorized,
```
JerryNixon, line 227: When OBO is enabled and we’re in request context, you throw Unauthorized if an OBO token can’t be acquired (ex: missing Authorization header). That’s fine, but the error message is a hardcoded string instead of using the centralized DataApiBuilderException constants you added (and it’s not super actionable for callers). Consider reusing a constant + making the message clearly say what’s missing (e.g., “Missing Bearer token” vs “no valid user context”). Also: you’re using the null-forgiving operator on DatabaseAudience!; if config validation didn’t run for some reason, this could become a null-ref instead of a clean config error.

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
JerryNixon, line 1: In the config validation tests, a bunch of tests set environment variables but don’t always clean them up unless they’re inside a try/finally. Some of the earlier “string runtimeConfigString” tests don’t set env vars (fine), but the ones that do set env vars should consistently clean up, otherwise test order can cause flaky behavior (one test’s env vars can accidentally make another test pass). I’d make sure every test that touches env vars uses try/finally cleanup (or a shared helper).

src/Core/Resolvers/OboSqlTokenProvider.cs
souvikghosh04: This however seems bit redundant to me. DAB has Redis support now. we should consider leveraging Redis and fallback to in-memory. Having no Redis support for this scenario seems a bit off. may be we should discuss this separately?

souvikghosh04: additionally, the statement in the PR description is slightly confusing with regards to this. Dictionary is not MSAL's built-in token cache though.

Image

src/Config/ObjectModel/DataSource.cs
```diff
@@ -111,3 +125,59 @@ public record CosmosDbNoSQLDataSourceOptions(string? Database, string? Container
...
+    /// </summary>
+    public const int DEFAULT_TOKEN_CACHE_DURATION_MINUTES = 45;
+
+    /// <summary>
+    /// Environment variable name for OBO App Registration client ID.
+    /// Uses DAB-specific prefix to avoid conflict with AZURE_CLIENT_ID which is
+    /// interpreted by DefaultAzureCredential/ManagedIdentityCredential as a
+    /// User-Assigned Managed Identity ID.
+    /// </summary>
+    public const string AZURE_CLIENT_ID_ENV_VAR = "DAB_OBO_CLIENT_ID";
```
souvikghosh04: nit: All the newly added CONSTs are having prefix `AZURE_` but it should be `DAB_OBO_` for consitency.

src/Service.Tests/Authentication/OboSqlTokenProviderUnitTests.cs
```diff
@@ -0,0 +1,533 @@
...
+            Assert.IsNull(result);
+        }
+
+        [TestMethod]
+        public async Task GetAccessTokenOnBehalfOfAsync_MissingOidAndSub_ThrowsUnauthorized()
+        {
+            // Arrange
+            Mock<IMsalClientWrapper> msalMock = new();
+            Mock<ILogger<OboSqlTokenProvider>> loggerMock = new();
+            OboSqlTokenProvider provider = new(msalMock.Object, loggerMock.Object);
```
souvikghosh04: these are repeated. use `TestInit()` and `[TestInitialize]` to move these and other common initializers/arrangers out of the tests. something like `[TestInitialize]
public void TestInit()
{_msalMock = new Mock<IMsalClientWrapper>();
    _loggerMock = new Mock<ILogger<OboSqlTokenProvider>>();
    _provider = new OboSqlTokenProvider(_msalMock.Object, _loggerMock.Object);}`

src/Service.Tests/Authentication/OboSqlTokenProviderUnitTests.cs
```diff
@@ -0,0 +1,533 @@
...
+                    incomingJwtAssertion: TEST_INCOMING_JWT,
+                    databaseAudience: TEST_DATABASE_AUDIENCE);
+            });
+
+            Assert.AreEqual(HttpStatusCode.Unauthorized, ex.StatusCode);
+            Assert.AreEqual(DataApiBuilderException.SubStatusCodes.OboAuthenticationFailure, ex.SubStatusCode);
+        }
+
+        [TestMethod]
+        public async Task GetAccessTokenOnBehalfOfAsync_VerifiesCorrectScopeFormat()
```
souvikghosh04: this test and the below TrimsTrailingSlashFromAudience seems identical so consider merging/refactoring them

src/Service.Tests/Authentication/OboSqlTokenProviderUnitTests.cs
```diff
@@ -0,0 +1,533 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#nullable enable
```
souvikghosh04: what's this needed for?

src/Service.Tests/Authentication/OboSqlTokenProviderUnitTests.cs
```diff
@@ -0,0 +1,533 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Identity.Client;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using AuthenticationOptions = Azure.DataApiBuilder.Config.ObjectModel.AuthenticationOptions;
+
+namespace Azure.DataApiBuilder.Service.Tests.Authentication
+{
+    [TestClass]
+    public class OboSqlTokenProviderUnitTests
```
souvikghosh04, line 26: overall feedback on this particular file- add summary docs and comments, refactor and reuse, merge similar tests

souvikghosh04 (changes requested):
added additional review comments which needs some changes and discussion. also, the PR description can be shortened. we don't have to mention all tests and functions, instead just categorize or summarize them.

src/Config/ObjectModel/DataSource.cs
```diff
@@ -111,3 +125,59 @@ public record CosmosDbNoSQLDataSourceOptions(string? Database, string? Container
...
+/// either via Managed Identity or credentials supplied in the connection string. All requests execute
+/// under the same database identity regardless of which user made the API call.
+/// 
+/// When OBO IS enabled: DAB exchanges the calling user's JWT for a database access token using the
+/// On-Behalf-Of flow. This allows DAB to connect to the database as the actual user, enabling
+/// Row-Level Security (RLS) filtering based on user identity.
+/// 
+/// OBO requires an Azure AD App Registration (separate from the DAB service's Managed Identity).
+/// Required environment variables for the OBO App Registration:
+/// - DAB_OBO_CLIENT_ID: The Application (client) ID of the App Registration used for OBO token exchange
```
Aniruddh25: Who sets these environment variables? and how do they propagate to DAB?

src/Config/DataApiBuilderException.cs
```diff
@@ -20,6 +20,12 @@ public class DataApiBuilderException : Exception
...
     public const string GRAPHQL_GROUPBY_FIELD_AUTHZ_FAILURE = "Access forbidden to field '{0}' referenced in the groupBy argument.";
     public const string GRAPHQL_AGGREGATION_FIELD_AUTHZ_FAILURE = "Access forbidden to field '{0}' referenced in the aggregation function '{1}'.";
+    public const string OBO_IDENTITY_CLAIMS_MISSING = "User-delegated authentication failed: Neither 'oid' nor 'sub' claim found in the access token.";
+    public const string OBO_TENANT_CLAIM_MISSING = "User-delegated authentication failed: 'tid' (tenant id) claim not found in the access token.";
+    public const string OBO_TOKEN_ACQUISITION_FAILED = "User-delegated authentication failed: Unable to acquire database access token on behalf of the user.";
+    public const string OBO_MISSING_BEARER_TOKEN = "User-delegated authentication failed: Missing or invalid 'Authorization: Bearer <token>' header in the request.";
+    public const string OBO_NO_USER_CONTEXT = "User-delegated authentication failed: No valid user context available. Ensure the request includes a valid Bearer token.";
```
Aniruddh25: how is this message different than for OBO_MISSING_BEARER_TOKEN?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -49,6 +49,19 @@ public class RuntimeConfigValidator : IConfigValidator
...
+        "User-delegated authentication is only supported when data-source.database-type is 'mssql'.";
+
+    public const string USER_DELEGATED_AUTH_MISSING_AUDIENCE_ERR_MSG =
+        "data-source.user-delegated-auth.database-audience must be set when user-delegated-auth is configured.";
+
+    public const string USER_DELEGATED_AUTH_CACHING_ERR_MSG =
+        "runtime.cache.enabled must be false when user-delegated-auth is configured.";
+
+    public const string USER_DELEGATED_AUTH_MISSING_CREDENTIALS_ERR_MSG =
+        "User-delegated authentication requires DAB_OBO_CLIENT_ID, DAB_OBO_TENANT_ID, and DAB_OBO_CLIENT_SECRET environment variables.";
```
Aniruddh25, line 63: IIRC, this wasn't mentioned in the specs

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -119,6 +132,68 @@ public void ValidateDataSourceInConfig(
...
+            // Skip validation if user-delegated-auth is not configured or not enabled
+            if (dataSource.UserDelegatedAuth is null || !dataSource.UserDelegatedAuth.Enabled)
+            {
+                continue;
+            }
+
+            if (dataSource.DatabaseType != DatabaseType.MSSQL)
+            {
+                HandleOrRecordException(new DataApiBuilderException(
+                    message: USER_DELEGATED_AUTH_DATABASE_TYPE_ERR_MSG,
```
Aniruddh25, line 162: Did you create a separate tracking issue for other dbs to handle OBO?

src/Core/Resolvers/OboSqlTokenProvider.cs
anushakolan: Updated the PR description.

Aniruddh25 (changes requested):
Leaving comments so far

src/Core/Configurations/RuntimeConfigValidator.cs
anushakolan: https://github.com/Azure/data-api-builder/issues/3159

src/Service.Tests/Authentication/OboSqlTokenProviderUnitTests.cs
anushakolan: Removed

src/Config/ObjectModel/DataSource.cs
anushakolan: **Who sets them:** The operator/administrator deploying DAB sets these environment variables in the hosting environment before starting the DAB service.

**How they propagate to DAB:** DAB reads these values at startup using `Environment.GetEnvironmentVariable(`). Specifically:

`RuntimeConfigValidator.ValidateUserDelegatedAuthConfig()` validates they are present when OBO is enabled
 `Startup.cs` reads them to initialize the MSAL `ConfidentialClientApplication` for OBO token exchange

This follows the same pattern as other environment variables in DAB (e.g., ASPNETCORE_URLS).

src/Config/DataApiBuilderException.cs
anushakolan: GetOboAccessTokenAsync() called
  → No Bearer header? Log OBO_MISSING_BEARER_TOKEN warning, return null
  → Has Bearer header? Try OBO exchange, return token or null

If GetOboAccessTokenAsync() returns null AND OBO is enabled:
  → Throw OBO_NO_USER_CONTEXT error (we can't proceed without user identity)

Merged them instead.

src/Core/Configurations/RuntimeConfigValidator.cs
anushakolan: You're right this wasn't explicitly in the original spec. I have updated the design spec to mention these env variables. 

- OBO requires a "confidential client" (App Registration) to exchange user tokens - Azure AD won't perform the exchange without client credentials
- Environment variables (not config file) because client secrets shouldn't be in files that may be committed to source control
- Validated at startup (fail-fast) because without it, DAB would start fine but fail on first user request with a cryptic MSAL error - the startup validation gives a clear, actionable error message.

src/Core/Resolvers/OboSqlTokenProvider.cs
anushakolan: Regarding using Redis cache there are 2 issues:
1. It is an explicit requirement to have in memory caching.
2. It will add additional round trip time to store the token in Redis. 

@JerryNixon Need your input on this? Do you think we should go with Redis cache here?

anushakolan: @Aniruddh25 FYI

anushakolan: From what Ani mentioned, we should not need to use Redis cache for OBO. Two separate features - one doesn’t depend on the other. @souvikghosh04 I think we should not go with OBO considering the issues here. Closing the discussion here, but if you see further concerns, please reopen the conversation.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -156,11 +171,22 @@ private void ConfigureMsSqlQueryExecutor()
                 MsSqlOptions? msSqlOptions = dataSource.GetTypedOptions<MsSqlOptions>();
                 _dataSourceToSessionContextUsage[dataSourceName] = msSqlOptions is null ? false : msSqlOptions.SetSessionContext;
                 _dataSourceAccessTokenUsage[dataSourceName] = ShouldManagedIdentityAccessBeAttempted(builder);
+
+                // Track user-delegated authentication settings
+                if (dataSource.IsUserDelegatedAuthEnabled && dataSource.UserDelegatedAuth is not null)
```
Aniruddh25: why is the second check needed?  `dataSource.UserDelegatedAuth is not null` - this is already covered in the first condition, no?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -175,13 +201,51 @@ public override async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection
...
+            // Check if user-delegated authentication is enabled for this data source
+            if (_dataSourceUserDelegatedAuth.TryGetValue(dataSourceName, out UserDelegatedAuthOptions? userDelegatedAuth))
+            {
+                // Check if we're in an HTTP request context (not startup/metadata phase)
+                bool isInRequestContext = HttpContextAccessor?.HttpContext is not null;
+
+                if (isInRequestContext)
+                {
+                    // Validate database-audience is configured before attempting OBO
+                    if (string.IsNullOrWhiteSpace(userDelegatedAuth.DatabaseAudience))
```
Aniruddh25: this validation should be done at startup, no need to check again. this is the core execution logic, need to keep it simple as possible - so that its quick.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -175,13 +201,51 @@ public override async Task SetManagedIdentityAccessTokenIfAnyAsync(DbConnection
...
+
+                    // OBO is enabled but we couldn't get a token (e.g., missing Bearer token in request)
+                    // This is an error during request processing - we must not fall back to managed identity
+                    throw new DataApiBuilderException(
+                        message: DataApiBuilderException.OBO_MISSING_USER_CONTEXT,
+                        statusCode: HttpStatusCode.Unauthorized,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.OboAuthenticationFailure);
+                }
+
+                // At startup/metadata phase (no HTTP context) - fall through to use Managed Identity
```
Aniruddh25: do this comment is incorrect, we don't necessarily always fall back to Managed identity. We fallback to use the connection string from the config - which could be Managed Identity or whatever DAB is configured with - could be SQL user/password too. 
We do recommend managed identity but thats not what this code will fall through to always.

src/Core/Resolvers/OboSqlTokenProvider.cs
```diff
@@ -0,0 +1,299 @@
...
+                statusCode: HttpStatusCode.Unauthorized,
+                subStatusCode: DataApiBuilderException.SubStatusCodes.OboAuthenticationFailure);
+        }
+
+        string authContextHash = ComputeAuthorizationContextHash(principal);
+        string cacheKey = BuildCacheKey(subjectId, tenantId, authContextHash);
+
+        // Check cache for valid token
+        // Refresh if: token older than cache duration OR token expires within early refresh buffer
+        if (_tokenCache.TryGetValue(cacheKey, out CachedToken? cached) && !ShouldRefresh(cached))
```
Aniruddh25: Although Redis Cache is not needed, we could use the Fusion InMemory Cache - we already use that for storing results of query. It handles the refresh and expiry logic in a better way.

src/Core/Resolvers/OboSqlTokenProvider.cs
```diff
@@ -0,0 +1,299 @@
...
+
+            CachedToken newCachedToken = new(
+                AccessToken: result.AccessToken,
+                ExpiresOn: result.ExpiresOn,
+                CachedAt: DateTimeOffset.UtcNow);
+
+            _tokenCache[cacheKey] = newCachedToken;
+
+            // Periodically clean up expired tokens to prevent unbounded memory growth
+            CleanupExpiredTokensIfNeeded();
```
Aniruddh25: If we use Fusion Cache, this periodic cleanup is automatically taken care of for us, because when we add the cachedToken in the in-memory cache, we set the expiration time.

src/Core/Resolvers/OboSqlTokenProvider.cs
```diff
@@ -0,0 +1,299 @@
...
+        {
+            return oid;
+        }
+
+        return principal.FindFirst("sub")?.Value;
+    }
+
+    /// <summary>
+    /// Builds a canonical representation of permission-affecting claims (roles and scopes)
+    /// and computes a SHA-256 hash for use in the cache key.
```
Aniruddh25: We should use SHA512 or SHA 384.

src/Core/Resolvers/OboSqlTokenProvider.cs
```diff
@@ -0,0 +1,299 @@
...
+    /// With a cache duration of N minutes and early refresh of M minutes,
+    /// the token is refreshed at (N - M) minutes after caching.
+    /// This ensures proactive refresh before the cache duration expires.
+    /// </summary>
+    private bool ShouldRefresh(CachedToken cached)
+    {
+        DateTimeOffset now = DateTimeOffset.UtcNow;
+
+        // Refresh at: cachedAt + (cacheDuration - earlyRefreshBuffer)
+        // Example: 45 min cache, 5 min buffer → refresh after 40 min
```
Aniruddh25: How were these values of 40 min, 5min buffer = 45 min decided?

src/Service/Startup.cs
```diff
@@ -252,6 +254,36 @@ public void ConfigureServices(IServiceCollection services)
...
+
+                    string authority = $"https://login.microsoftonline.com/{tenantId}";
+
+                    IConfidentialClientApplication msalClient = ConfidentialClientApplicationBuilder
+                        .Create(clientId)
+                        .WithAuthority(authority)
+                        .WithClientSecret(clientSecret)
+                        .Build();
+
+                    IMsalClientWrapper msalWrapper = new MsalClientWrapper(msalClient);
```
Aniruddh25: Shouldnt the MsalClientWrapper be added to the list of service collection? And let dependency injection provide it to OboSqlTokenProvider?

src/Service/Startup.cs
```diff
@@ -1241,6 +1273,43 @@ private static bool IsUIEnabled(RuntimeConfig? runtimeConfig, IWebHostEnvironmen
...
+            RuntimeConfig? config = _configProvider?.TryGetConfig(out RuntimeConfig? c) == true ? c : null;
+            if (config is null)
+            {
+                return false;
+            }
+
+            foreach (DataSource ds in config.ListAllDataSources())
+            {
+                if (ds.IsUserDelegatedAuthEnabled &&
+                    ds.UserDelegatedAuth is not null &&
```
Aniruddh25: ds.UserDelegatedAuth is not null is a redundant condition after checking IsUserDelegatedAuthEnabled because the former is checked in the latter's definition.

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -420,7 +420,7 @@ public void TestInfoMessageHandlerIsAdded()
             DbExceptionParser dbExceptionParser = new MsSqlDbExceptionParser(provider);
             EventHandler handler = null;
             Mock<MsSqlQueryExecutor> queryExecutor
-                = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object, handler);
+                = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object, handler, null);
```
Aniruddh25: nit: name the argument which is passed here as null, at all constructors in this file.

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -2216,6 +2216,561 @@ public void ValidateUniqueRestPathsForEntitiesInConfig(
...
+            Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ConfigValidationError, ex.SubStatusCode);
+        }
+
+        /// <summary>
+        /// Test to validate that user-delegated-auth with empty database-audience throws an error.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("", DisplayName = "Empty string audience should fail")]
+        [DataRow("   ", DisplayName = "Whitespace audience should fail")]
+        public void ValidateUserDelegatedAuth_EmptyDatabaseAudience_ThrowsError(string audience)
```
Aniruddh25: Arent empty and missing database audience similar tests?

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -2216,6 +2216,561 @@ public void ValidateUniqueRestPathsForEntitiesInConfig(
...
+            Assert.AreEqual(RuntimeConfigValidator.USER_DELEGATED_AUTH_MISSING_AUDIENCE_ERR_MSG, ex.Message);
+            Assert.AreEqual(HttpStatusCode.ServiceUnavailable, ex.StatusCode);
+            Assert.AreEqual(DataApiBuilderException.SubStatusCodes.ConfigValidationError, ex.SubStatusCode);
+        }
+
+        /// <summary>
+        /// Test to validate that user-delegated-auth with caching enabled throws an error.
+        /// </summary>
+        [TestMethod]
+        public void ValidateUserDelegatedAuth_CachingEnabled_ThrowsError()
```
Aniruddh25: There was already a test above to ensure Caching is Disabled.

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -2216,6 +2216,561 @@ public void ValidateUniqueRestPathsForEntitiesInConfig(
...
+
+        /// <summary>
+        /// Test to validate that user-delegated-auth with missing DAB_OBO_CLIENT_ID, DAB_OBO_TENANT_ID,
+        /// or DAB_OBO_CLIENT_SECRET throws an error.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(null, "test-tenant", "test-secret", DisplayName = "Missing DAB_OBO_CLIENT_ID")]
+        [DataRow("test-client", null, "test-secret", DisplayName = "Missing DAB_OBO_TENANT_ID")]
+        [DataRow("test-client", "test-tenant", null, DisplayName = "Missing DAB_OBO_CLIENT_SECRET")]
+        [DataRow("", "test-tenant", "test-secret", DisplayName = "Empty DAB_OBO_CLIENT_ID")]
```
Aniruddh25, line 2465: Please inform @JerryNixon about the need for these env variables and get an ACK. This affects how the users will be using this feature. So, we need to ensure we are on same page.

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
```diff
@@ -2216,6 +2216,561 @@ public void ValidateUniqueRestPathsForEntitiesInConfig(
...
+        /// <summary>
+        /// Test to validate that running in Azure (simulated by IDENTITY_ENDPOINT) passes validation
+        /// when all required OBO environment variables are provided.
+        /// Note: OBO always requires DAB_OBO_CLIENT_SECRET even in Azure environments because
+        /// OBO uses a confidential client application (App Registration with secret), not Managed Identity.
+        /// </summary>
+        [TestMethod]
+        public void ValidateUserDelegatedAuth_AzureEnvironment_PassesValidation()
+        {
+            // Arrange - Simulate Azure environment (IDENTITY_ENDPOINT is set in Azure Container Apps/App Service)
```
Aniruddh25: How is this test different than above apart from setting the IDENTITY_ENDPOINT? It is testing the same codepath, how is setting IDENTITY_ENDPOINT affecting validation?

src/Service.Tests/UnitTests/ConfigValidationUnitTests.cs
anushakolan: Right, removing

Issue #3131, opened 2026-02-13
Update Validation for Autoentities


RubenCerna2079:
Currently, the entities generated from the autoentities feature are added after the validation for the RuntimeConfig is done. This needs to be changed since we need to validate that the new generated entities are validated with the entities that already exist. In order to do this, we need to move only the validation of the entities so that it occurs after the autoentities are generated. We cannot move all the validation after the autoentities are generated because some of the validation is needed for the connection to the database.

PR #3152, opened 2026-02-21
Update Validation for Autoentities

RubenCerna2079:
## Why make this change?

- #3131
- Entity validation (REST path uniqueness, GraphQL type uniqueness, permissions) ran in `Startup.cs` before autoentity generation, so conflicts between manually defined and auto-generated entities were never caught.

## What is this change?

- **Validation reordered**: `ValidatePermissionsInConfig` and `ValidateEntityConfiguration` removed from `Startup.cs`; both now execute inside `SqlMetadataProvider.InitializeAsync()` after `GenerateAutoentitiesIntoEntities()` completes, ensuring the full entity set (manual + auto-generated) is validated together.

- **New `ValidateEntityAndAutoentityConfigurations` method** on `RuntimeConfigValidator`: consolidates REST/GraphQL path uniqueness, duplicate query/mutation detection, and permissions validation — runs in development mode only, consistent with prior behavior.

- **`RuntimeConfigValidator` injected into `SqlMetadataProvider`** (and subclasses `MsSqlMetadataProvider`, `PostgreSqlMetadataProvider`, `MySqlMetadataProvider`) to enable calling the new method at the correct point in the init sequence.

- **`dab validate` flow**: `ValidateEntitiesMetadata` now passes the validator into metadata provider initialization, so `dab validate` also validates entity/autoentity configurations post-generation.

- **CosmosDB**: `ValidateEntityAndAutoentityConfigurations` added to `CosmosSqlMetadataProvider.InitializeAsync()` for consistency.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

### DIFF
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -5610,7 +5610,7 @@ type Planet @model(name:""PlanetAlias"") {
                 Entities: new(new Dictionary<string, Entity>()),
                 Autoentities: new(autoentityMap));
 
-            const string CUSTOM_CONFIG = "custom-config.json";
+            const string CUSTOM_CONFIG = "autoentities-validation-config.json";
 
             File.WriteAllText(CUSTOM_CONFIG, runtimeConfig.ToJson());
             IFileSystem fileSystem = new FileSystem();


anushakolan (approved):
LGTM!

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -310,9 +314,26 @@ public string GetEntityName(string graphQLType)
         public async Task InitializeAsync()
         {
             System.Diagnostics.Stopwatch timer = System.Diagnostics.Stopwatch.StartNew();
+
             if (GetDatabaseType() == DatabaseType.MSSQL)
             {
-                await GenerateAutoentitiesIntoEntities(Autoentities);
+                await GenerateAutoentitiesIntoEntities(new Dictionary<string, Autoentity>(Autoentities));
```
Aniruddh25: Did you create separate tasks for PgSql, MySql to generate autoentities into entities?

Aniruddh25: Code for validating connection to database is below:

                    await ValidateDatabaseConnection();

But generate autoentities to entities queries the database before this. The order should be reversed.

RubenCerna2079: Just created them, they will be set in the backlog for now:
- PgSql: https://github.com/Azure/data-api-builder/issues/3172
- MySql: https://github.com/Azure/data-api-builder/issues/3173

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -38,14 +39,17 @@ public abstract class SqlMetadataProvider<ConnectionT, DataAdapterT, CommandT> :
...
         protected Dictionary<string, Entity> _linkingEntities = new();
 
         protected readonly string _dataSourceName;
 
+        // Represents the entities exposed in the runtime config.
+        private IReadOnlyDictionary<string, Entity> Entities => new ReadOnlyDictionary<string, Entity>(_runtimeConfigProvider.GetConfig().Entities.Where(x => string.Equals(_runtimeConfigProvider.GetConfig().GetDataSourceNameFromEntityName(x.Key), _dataSourceName, StringComparison.OrdinalIgnoreCase)).ToDictionary(x => x.Key, x => x.Value));
+
+        // Represents the autoentities exposed in the runtime config.
+        private IReadOnlyDictionary<string, Autoentity> Autoentities => new ReadOnlyDictionary<string, Autoentity>(_runtimeConfigProvider.GetConfig().Autoentities.Where(x => string.Equals(_runtimeConfigProvider.GetConfig().GetDataSourceNameFromAutoentityName(x.Key), _dataSourceName, StringComparison.OrdinalIgnoreCase)).ToDictionary(x => x.Key, x => x.Value));
+
```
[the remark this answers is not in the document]
RubenCerna2079: The purpose of this is to refresh it in order to update the entities if there are any new ones.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -307,11 +314,17 @@ public string GetEntityName(string graphQLType)
...
             if (GetDatabaseType() == DatabaseType.MSSQL)
             {
-                await GenerateAutoentitiesIntoEntities();
+                await GenerateAutoentitiesIntoEntities(new Dictionary<string, Autoentity>(Autoentities));
             }
 
+            // Running these graphQL validations only in development mode to ensure
+            // fast startup of engine in production mode.
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            _runtimeConfigValidator.ValidateEntityAndAutoentityConfigurations(runtimeConfig);
```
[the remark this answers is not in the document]
RubenCerna2079: It is expected that the validation will occur for each of the data sources as it needs to validate every time new entities are created.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -1594,4 +1583,26 @@ private static bool IsLoggerFilterValid(string loggerFilter)
...
+        {
+            ValidateEntityConfiguration(runtimeConfig);
+
+            if (runtimeConfig.IsGraphQLEnabled)
+            {
+                ValidateEntitiesDoNotGenerateDuplicateQueriesOrMutation(runtimeConfig.DataSource.DatabaseType, runtimeConfig.Entities);
+            }
+
+            // Running only in developer mode to ensure fast and smooth startup in production.
+            ValidatePermissionsInConfig(runtimeConfig);
```
anushakolan, line 1612: Why is this function needed, it seems to be a new addition.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5434,6 +5445,115 @@ public async Task TestAutoentitiesAreGeneratedIntoEntities(bool useEntities, int
...
+            RuntimeConfigValidator configValidator = new(configProvider, fileSystem, loggerValidator.Object);
+
+            LoggerFactory loggerFactory = new();
+
+            try
+            {
+                await configValidator.ValidateEntitiesMetadata(configuration, loggerFactory);
+                Assert.Fail("It is expected for DAB to fail due to entities not containing unique parameters.");
+            }
+            catch (Exception ex)
```
anushakolan: Shouldn't we catch the specific DAB Exception instead of the generic Exception, because we can also get AssertFailedException leading to wrong exception type.

anushakolan: This test relies on Assert.Fail inside a try block as a guard. When the expected validation exception is not thrown, Assert.Fail throws an AssertFailedException, which is then caught by the general catch (Exception) block. That leads to comparing the Assert.Fail message instead of the product exception message, which is confusing and hides the real bug (“expected exception was not thrown”).
Suggest using Assert.ThrowsExceptionAsync instead. It makes the intent explicit, avoids catching AssertFailedException, and produces a clear failure when no exception is thrown.

anushakolan (approved):
Have some minor suggestion, but rest LGTM!

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
```diff
@@ -76,6 +76,8 @@ public CosmosSqlMetadataProvider(RuntimeConfigProvider runtimeConfigProvider, IF
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
             }
 
+            runtimeConfigValidator.ValidateEntityAndAutoentityConfigurations(runtimeConfig);
```
Aniruddh25, line 79: Was the validation being executed even for CosmosDB previously?

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -332,6 +331,18 @@ public async Task InitializeAsync()
                 }
             }
 
+            if (GetDatabaseType() == DatabaseType.MSSQL)
```
Aniruddh25, line 334: Will we error out for other databases if `autoentities` section is present in the config?

We shouldn't...

src/Core/Configurations/RuntimeConfigValidator.cs
RubenCerna2079: It was already there, it specifically validates if the permissions for each entity are valid.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
RubenCerna2079: It doesn't error out if other databases if `autoentities` is present, we just don't execute it

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5434,6 +5445,115 @@ public async Task TestAutoentitiesAreGeneratedIntoEntities(bool useEntities, int
...
+            RuntimeConfigProvider configProvider = new(configLoader);
+
+            Mock<ILogger<RuntimeConfigValidator>> loggerValidator = new();
+            RuntimeConfigValidator configValidator = new(configProvider, fileSystem, loggerValidator.Object);
+
+            LoggerFactory loggerFactory = new();
+
+            try
+            {
+                await configValidator.ValidateEntitiesMetadata(configuration, loggerFactory);
```
Aniruddh25: if ValidateEntitiesMetadata is also necessary after the auto entity configuration is populated, I dont see this happening in the repositioned validations in this function: `ValidateEntityAndAutoentityConfigurations`

Will ValidateEntitiesMetadata be actually invoked when starting up or only while dab validate command is executed - because I see this called only from `TryValidateConfig` function.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -96,18 +96,6 @@ public void ValidateConfigProperties()
         ValidateLoggerFilters(runtimeConfig);
         ValidateAzureLogAnalyticsAuth(runtimeConfig);
         ValidateFileSinkPath(runtimeConfig);
-
-        // Running these graphQL validations only in development mode to ensure
```
Aniruddh25, line 100: Removing this from here, are we loosing these validations when running `dab validate`. 

I think we should keep call them in the dab validate flow as well inside TryValidateConfig similar to how ValidateEntitiesMetadata is called there.

src/Service.Tests/Configuration/ConfigurationTests.cs
Aniruddh25: We should add ValidateEntitiesMetadata to ValidateEntityAndAutoentityConfigurations as well and ensure the latter is only called at startup, not in dab validate.

Aniruddh25 (changes requested):
ValidateEntitiesMetadata needs to be called as well.

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: I will add the DAB exception, but it is technically not necessary as we still check for the message inside the catch

RubenCerna2079: The test is working properly since we are still checking inside of the catch if the message is what we expect

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -318,6 +318,19 @@ public async Task<bool> TryValidateConfig(
 
         JsonSchemaValidationResult validationResult = await ValidateConfigSchema(runtimeConfig, configFilePath, loggerFactory);
         ValidateConfigProperties();
+
+        // Running these graphQL validations only in development mode to ensureExpand commentComment on line L100
```
Aniruddh25: >         // Running these graphQL validations only in development mode to ensureExpand commentComment on line L100

delete this `Expand commentComment on line L100`

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -318,6 +318,19 @@ public async Task<bool> TryValidateConfig(
...
+        // fast startup of engine in production mode.
+        if (runtimeConfig.IsDevelopmentMode())
+        {
+            ValidateEntityConfiguration(runtimeConfig);
```
Aniruddh25: Given that this is being called in dab validate as well, can we add a test to ensure autoentities section is validated

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
Aniruddh25, line 325: ValidateEntityAndAutoentityConfigurations should only happen after the ValidateDBConnection is done.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -319,6 +315,17 @@ public string GetEntityName(string graphQLType)
         public async Task InitializeAsync()
         {
             System.Diagnostics.Stopwatch timer = System.Diagnostics.Stopwatch.StartNew();
+            if (GetDatabaseType() == DatabaseType.MSSQL)
+            {
+                await GenerateAutoentitiesIntoEntities(Autoentities);
+            }
+
+            // Running these graphQL validations only in development mode to ensure
```
Aniruddh25: is this a GraphQL validation? The comment is misleading.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -319,6 +315,17 @@ public string GetEntityName(string graphQLType)
...
+            // fast startup of engine in production mode.
+            RuntimeConfig runtimeConfig = _runtimeConfigProvider.GetConfig();
+            _runtimeConfigValidator.ValidateEntityAndAutoentityConfigurations(runtimeConfig);
```
Aniruddh25: Is ValidateEntityAndAutoentityConfigurations needed for other databases too? Asking because, it is outside the check for MSSQL

RubenCerna2079: Yes, it is necessary for other databases as well

src/Core/Configurations/RuntimeConfigValidator.cs
RubenCerna2079: Test was added

src/Core/Services/MetadataProviders/CosmosSqlMetadataProvider.cs
RubenCerna2079: Yes the validation is also executed for CosmosDB previously, will add this section back

Aniruddh25 (approved):
Synced offline, `dab validate` is now doing additional validation of permissions by calling `ValidatePermissionsInConfig` through the new ValidateEntityAndAutoentityConfigurations

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -96,18 +96,6 @@ public void ValidateConfigProperties()
         ValidateLoggerFilters(runtimeConfig);
         ValidateAzureLogAnalyticsAuth(runtimeConfig);
         ValidateFileSinkPath(runtimeConfig);
```
Aniruddh25, line 98: The Description section on "What is Changed" should be complete

Aniruddh25 (approved):

Aniruddh25:
@copilot, describe what this change is and edit the PR description to add that in the `What is this change?` section.

PR #3153, opened 2026-02-24
Added per user connection pooling for OBO.

anushakolan:
## Why make this change?
- Implements per-user connection pooling for On-Behalf-Of (OBO) authentication mode.
- When OBO is enabled, each user gets their own isolated SQL connection pool, ensuring that connections with user-specific access tokens are not shared across users.
- This is critical for row-level security scenarios where the database uses the user's identity for authorization decisions.

## What is this change?

- **Automatic per-user pooling for OBO**: When `user-delegated-auth` is enabled, the connection pool is automatically partitioned by user identity. No additional configuration is required.
- **Pool isolation via Application Name**: Each user's pool is identified by modifying the SQL Server `Application Name` connection string property to include a user-specific hash: `{baseAppName}|obo:{hash}`
- **Hash generation**: The hash is derived from the user's JWT claims (`iss` + `oid` or `sub`) using SHA512, then encoded as URL-safe Base64.
- **Startup/metadata compatibility**: When no user context is present (e.g., during startup metadata introspection), the base connection string is used without the OBO suffix.

### Key implementation details:
- `_dataSourceBaseAppName` dictionary stores the base Application Name for OBO-enabled data sources
- `GetUserPoolKeyHash()` extracts user claims and generates deterministic hash
- `GetConnectionStringForCurrentUser()` appends user hash to create pool-isolated connection strings

## How was this tested?

- [x] Integration Tests - E2E testing with Azure SQL and multiple users verified different pool hashes
- [x] Unit Tests - 3 new comprehensive unit tests added:
  - `TestOboWithUserClaims_ConnectionStringHasUserSpecificAppName` - Verifies connection string contains `|obo:` with hash when user claims present
  - `TestObo_DifferentUsersGetDifferentPoolHashes` - Verifies different users get different Application Names
  - `TestOboNoUserContext_UsesBaseConnectionString` - Verifies startup scenario uses base connection string without OBO suffix

### Manual E2E Testing

**Environment Setup:**
- Azure SQL Database with OBO authentication enabled
- DAB deployed to Azure Container Apps with Entra ID authentication
- Two test users with different Entra ID accounts

**Test Scenario 1: Single User Connection Pool Isolation**
1. Authenticated as User A and made multiple REST requests
2. Verified via Azure SQL `sys.dm_exec_sessions` that all connections had the same Application Name with user-specific hash
3. Confirmed connection pooling was working (connection reuse observed)

**Test Scenario 2: Multi-User Pool Isolation**
1. Authenticated as User A, made requests → observed Application Name: `DAB,dab_oss_2.0.0|obo:W8L-UzlymZJcRbqAoTRBBcDDvHBjKoEP1lo94PGpcp0`
2. Authenticated as User B, made requests → observed Application Name: `DAB,dab_oss_2.0.0|obo:Xk9mPqRsT2vWxYz1AbCdEfGhIjKlMnOpQrStUvWx`
3. Verified both users had **different** hashes, confirming pool isolation
4. Re-authenticated as User A → verified **same** hash as before, confirming deterministic hashing

**Test Scenario 3: Startup Without User Context**
1. Restarted DAB service
2. Verified startup metadata introspection succeeded (no user context required)
3. Confirmed base Application Name used during startup: `DAB,dab_oss_2.0.0` (no `|obo:` suffix)

**Verification Query (run on Azure SQL):**
```sql
SELECT 
    session_id,
    login_name,
    program_name AS application_name,
    login_time
FROM sys.dm_exec_sessions
WHERE program_name LIKE '%dab%'
ORDER BY login_time DESC;```

### DIFF
--- a/src/Core/Resolvers/MsSqlQueryExecutor.cs
+++ b/src/Core/Resolvers/MsSqlQueryExecutor.cs
@@ -275,11 +275,11 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         /// <summary>
         /// Hashes the pool key using SHA256 truncated to 16 bytes for a compact, URL-safe identifier.
         /// Uses SHA256 (SHA-2 family) with 128-bit truncation per Microsoft security requirements.
-        /// This produces a 16 character hash that fits well within SQL Server's 128-char Application Name limit
-        /// while providing sufficient collision resistance.
+        /// This produces a ~22 character hash (16 bytes Base64-encoded) that fits well within SQL Server's
+        /// 128-char Application Name limit while providing sufficient collision resistance.
         /// </summary>
         /// <param name="key">The pool key to hash (format: iss|oid or iss|sub).</param>
-        /// <returns>A URL-safe Base64-encoded hash of the key (16 characters).</returns>
+        /// <returns>A URL-safe Base64-encoded hash of the key (~22 characters).</returns>
         private static string HashPoolKey(string key)
         {
             byte[] fullHash = System.Security.Cryptography.SHA256.HashData(
--- a/src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
+++ b/src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
@@ -757,51 +757,33 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
         }
 
         /// <summary>
-        /// Test that pooling remains enabled regardless of whether user-delegated-auth is configured.
-        /// When OBO is enabled, per-user pooling is automatic. When disabled, pooling stays as configured.
+        /// Test that the Pooling property from the connection string is never modified by DAB,
+        /// regardless of whether OBO is enabled or disabled. If Pooling=true, it stays true.
+        /// If Pooling=false, it stays false. DAB respects the user's explicit configuration.
         /// </summary>
         [DataTestMethod, TestCategory(TestCategory.MSSQL)]
-        [DataRow(true, DisplayName = "OBO enabled - per-user pooling is automatic")]
-        [DataRow(false, DisplayName = "OBO disabled - pooling not modified")]
-        public void TestPoolingBehaviorWithAndWithoutUserDelegatedAuth(bool enableUserDelegatedAuth)
+        [DataRow(true, true, DisplayName = "OBO enabled, Pooling=true stays true")]
+        [DataRow(true, false, DisplayName = "OBO enabled, Pooling=false stays false")]
+        [DataRow(false, true, DisplayName = "OBO disabled, Pooling=true stays true")]
+        [DataRow(false, false, DisplayName = "OBO disabled, Pooling=false stays false")]
+        public void TestPoolingPropertyIsNeverModified(bool enableObo, bool poolingValue)
         {
-            // Arrange & Act
+            // Arrange
             Mock<IHttpContextAccessor> httpContextAccessor = new();
+            string connectionString = $"Server=localhost;Database=test;Pooling={poolingValue};";
+
+            // Act
             (MsSqlQueryExecutor queryExecutor, RuntimeConfigProvider provider) = CreateQueryExecutorForPoolingTest(
-                connectionString: "Server=localhost;Database=test;Pooling=true;",
-                enableObo: enableUserDelegatedAuth,
+                connectionString: connectionString,
+                enableObo: enableObo,
                 httpContextAccessor: httpContextAccessor);
 
-            SqlConnectionStringBuilder connBuilder = new(queryExecutor.ConnectionStringBuilders[provider.GetConfig().DefaultDataSourceName].ConnectionString);
-
-            // Assert - pooling should be enabled in both cases
-            string expectedMessage = enableUserDelegatedAuth
-                ? "Pooling should be enabled when user-delegated-auth is configured (per-user pooling is automatic)"
-                : "Pooling should remain as configured when user-delegated-auth is not used";
-            Assert.IsTrue(connBuilder.Pooling, expectedMessage);
-        }
-
-        /// <summary>
-        /// Test that when OBO is enabled, the user's pooling setting from the connection string is respected.
-        /// DAB does not override the user's choice - if they set Pooling=false, it stays false.
-        /// </summary>
-        [TestMethod, TestCategory(TestCategory.MSSQL)]
-        public void TestOboEnabled_RespectsUserPoolingSetting()
-        {
-            // Arrange - connection string explicitly disables pooling
-            Mock<IHttpContextAccessor> httpContextAccessor = new();
-            (MsSqlQueryExecutor queryExecutor, RuntimeConfigProvider provider) = CreateQueryExecutorForPoolingTest(
-                connectionString: "Server=localhost;Database=test;Pooling=false;",
-                enableObo: true,
-                httpContextAccessor: httpContextAccessor);
-
-            // Act - check the stored connection string builder
             SqlConnectionStringBuilder connBuilder = new(
                 queryExecutor.ConnectionStringBuilders[provider.GetConfig().DefaultDataSourceName].ConnectionString);
 
-            // Assert - pooling setting should be respected from user's connection string
-            Assert.IsFalse(connBuilder.Pooling,
-                "User's Pooling=false setting should be respected, not overridden by OBO");
+            // Assert - Pooling property should be unchanged from the original connection string
+            Assert.AreEqual(poolingValue, connBuilder.Pooling,
+                $"Pooling={poolingValue} should remain unchanged when OBO is {(enableObo ? "enabled" : "disabled")}");
         }
 
         /// <summary>
@@ -825,17 +807,64 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
             SqlConnectionStringBuilder connBuilder = new(conn.ConnectionString);
 
             // Assert - Application Name should have hash prefix followed by the base name
-            // Format: {hash}|{user-custom-appname} where hash is ~22 chars
+            // Format: {hash}|{user-custom-appname}
+            // Hash is 16 bytes truncated SHA256, Base64-encoded to ~22 chars (16 bytes * 4/3 = 21.3)
             // Hash is placed first to ensure it's never truncated if app name exceeds 128 chars
             Assert.IsTrue(connBuilder.ApplicationName.Contains("|"),
                 $"Application Name should contain '|' separator but was '{connBuilder.ApplicationName}'");
             Assert.IsTrue(connBuilder.ApplicationName.Contains("TestApp"),
                 $"Application Name should contain 'TestApp' but was '{connBuilder.ApplicationName}'");
             // Hash should be at the start (before the | separator)
+            // 16 bytes Base64-encoded (without padding) = ~22 characters
             string hashPart = connBuilder.ApplicationName.Split('|')[0];
             Assert.IsTrue(hashPart.Length >= 20 && hashPart.Length <= 25,
-                $"Hash prefix should be ~22 chars but was {hashPart.Length} chars: '{hashPart}'");
+                $"Hash prefix should be ~22 chars (16 bytes Base64) but was {hashPart.Length} chars: '{hashPart}'");
             Assert.IsTrue(connBuilder.Pooling, "Pooling should be enabled");
+        }
+
+        /// <summary>
+        /// Test that when the base Application Name + hash prefix exceeds 128 characters,
+        /// the base app name is truncated (not the hash) to fit within SQL Server's limit.
+        /// This verifies the hash-first format ensures pool isolation even with long app names.
+        /// </summary>
+        [TestMethod, TestCategory(TestCategory.MSSQL)]
+        public void TestOboWithLongAppName_TruncatesToFitWithinLimit()
+        {
+            // Arrange - Create an Application Name that would exceed 128 chars when hash is added
+            // Hash prefix is ~22 chars + "|" = 23 chars, so base app name of 120 chars would exceed limit
+            string longAppName = new('A', 120); // 120 chars, plus 23 for hash = 143 total
+
+            Mock<IHttpContextAccessor> httpContextAccessor = CreateHttpContextAccessorWithClaims(
+                issuer: "https://login.microsoftonline.com/tenant-id/v2.0",
+                objectId: "user-object-id-12345");
+
+            (MsSqlQueryExecutor queryExecutor, RuntimeConfigProvider provider) = CreateQueryExecutorForPoolingTest(
+                connectionString: $"Server=localhost;Database=test;Application Name={longAppName};",
+                enableObo: true,
+                httpContextAccessor: httpContextAccessor);
+
+            // Act
+            SqlConnection conn = queryExecutor.CreateConnection(provider.GetConfig().DefaultDataSourceName);
+            SqlConnectionStringBuilder connBuilder = new(conn.ConnectionString);
+
+            // Assert - Application Name should be truncated to 128 chars max
+            Assert.IsTrue(connBuilder.ApplicationName.Length <= 128,
+                $"Application Name should be <= 128 chars but was {connBuilder.ApplicationName.Length} chars");
+
+            // Hash should still be at the start and complete (not truncated)
+            string[] parts = connBuilder.ApplicationName.Split('|');
+            Assert.AreEqual(2, parts.Length, "Application Name should have exactly one '|' separator");
+
+            string hashPart = parts[0];
+            Assert.IsTrue(hashPart.Length >= 20 && hashPart.Length <= 25,
+                $"Hash prefix should be ~22 chars (16 bytes Base64) but was {hashPart.Length} chars: '{hashPart}'");
+
+            // The base app name should be truncated, not the hash
+            string truncatedAppName = parts[1];
+            Assert.IsTrue(truncatedAppName.Length < longAppName.Length,
+                $"Base app name should be truncated from {longAppName.Length} chars but was {truncatedAppName.Length} chars");
+            Assert.IsTrue(truncatedAppName.All(c => c == 'A'),
+                "Truncated app name should contain only the original characters (no corruption)");
         }
 
         /// <summary>
@@ -900,7 +929,7 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
             Assert.IsTrue(connBuilder.ApplicationName.StartsWith("BaseApp"),
                 $"Without user context, Application Name should start with 'BaseApp' but was '{connBuilder.ApplicationName}'");
             // When no user context, the app name should NOT have the hash prefix pattern
-            // (hash prefix would be ~22 chars followed by |)
+            // (hash prefix is 16 bytes Base64-encoded = ~22 chars, followed by |)
             string[] parts = connBuilder.ApplicationName.Split('|');
             bool hasHashPrefix = parts.Length > 1 && parts[0].Length >= 20 && parts[0].Length <= 25;
             Assert.IsFalse(hasHashPrefix,

### CODE REVIEW

File: src/Core/Resolvers/MsSqlQueryExecutor.cs [hunk 1]
```diff
@@ -150,6 +160,97 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+
+            // Per-user pooling is automatic when OBO is enabled.
+            // _dataSourceBaseAppName is only populated for data sources with user-delegated-auth enabled.
+            if (!_dataSourceBaseAppName.TryGetValue(dataSourceName, out string? baseAppName))
+            {
+                // OBO not enabled for this data source, use the standard connection string
+                return baseConnectionString;
+            }
+
+            // Extract user pool key from current HTTP context using iss:sub
```
JerryNixon: update the comment to match the real behavior so the next reader doesn’t get confused. the comment says “iss:sub”, but GetUserPoolKeyHash actually prefers oid then falls back to sub.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +160,97 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+        {
+            if (HttpContextAccessor?.HttpContext?.User is null)
+            {
+                return null;
+            }
+
+            ClaimsPrincipal user = HttpContextAccessor.HttpContext.User;
+
+            // Extract issuer claim - required for tenant isolation
+            string? iss = user.FindFirst("iss")?.Value;
```
JerryNixon, line 245: Some identity setups may not emit a literal "iss" claim in the principal (even if the token had an issuer). You already do good fallbacks for oid/sub; you could do similar for issuer if there’s an established claim type you use elsewhere in the repo. add a fallback source for issuer (if you can), or at least mention in a comment that "iss" must be present for pooling to activate.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +160,97 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+            // Prefer oid (stable GUID), fall back to sub for guest/B2B users
+            string? userKey = user.FindFirst("oid")?.Value
+                ?? user.FindFirst("http://schemas.microsoft.com/identity/claims/objectidentifier")?.Value
+                ?? user.FindFirst("sub")?.Value
+                ?? user.FindFirst(ClaimTypes.NameIdentifier)?.Value;
+
+            if (string.IsNullOrEmpty(iss) || string.IsNullOrEmpty(userKey))
+            {
+                // Cannot create a pool key without both claims
+                QueryExecutorLogger.LogDebug("Cannot create per-user pool key: missing iss or user identifier claim.");
```
JerryNixon: maybe include datasource name (still safe) to help debugging. Remember to use structured logs

JerryNixon (dismissed):

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -177,9 +278,11 @@ private void ConfigureMsSqlQueryExecutor()
                 {
                     _dataSourceUserDelegatedAuth[dataSourceName] = dataSource.UserDelegatedAuth;
 
-                    // Disable connection pooling for OBO connections since each connection
-                    // uses a user-specific token and cannot be shared across users
-                    builder.Pooling = false;
+                    // Per-user pooling: Keep pooling enabled but store the base Application Name.
+                    // At connection time, we'll append the user's iss:sub hash to create isolated pools per user.
+                    // This is automatic for OBO to prevent connection exhaustion while ensuring pool isolation.
+                    _dataSourceBaseAppName[dataSourceName] = builder.ApplicationName ?? "DAB";
```
Aniruddh25: We should reuse the Application name we generate for all dab connections. I think it is dab_oss_<version#>. I think ProductInfo class generates that app name. 

By the time we reach this codepath is builder.ApplicationName already populated always?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +160,97 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+
+        /// <summary>
+        /// Hashes the pool key using SHA256 to create a compact, URL-safe identifier.
+        /// This keeps the Application Name reasonably short while ensuring uniqueness.
+        /// </summary>
+        /// <param name="key">The pool key to hash (format: iss|oid or iss|sub).</param>
+        /// <returns>A URL-safe Base64-encoded hash of the key.</returns>
+        private static string HashPoolKey(string key)
+        {
+            using var sha = System.Security.Cryptography.SHA256.Create();
```
Aniruddh25: use SHA384 or SHA512

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +160,97 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+                // No user context (startup/metadata phase), use base connection string
+                return baseConnectionString;
+            }
+
+            // Create a user-specific connection string by modifying the Application Name
+            // This creates a separate connection pool for each unique iss:sub combination
+            // Format: {baseAppName}|obo:{hash} as per design spec
+            SqlConnectionStringBuilder userBuilder = new(baseConnectionString)
+            {
+                ApplicationName = $"{baseAppName}|obo:{poolKeyHash}",
```
Aniruddh25: instead of overwriting the application name completely, we should append to the existing value. Customers could have set some application name value already

src/Core/Resolvers/MsSqlQueryExecutor.cs
anushakolan: I will instead fail safe that issuer must be present otherwise it would be risky as there could be connection pool contamination between users.

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -669,6 +669,283 @@ public void ValidateStreamingLogicForEmptyCellsAsync()
...
+                   GraphQL: new(),
+                   Mcp: new(),
+                   Host: new(null, null)
+               ),
+               Entities: new(new Dictionary<string, Entity>()));
+
+            MockFileSystem fileSystem = new();
+            fileSystem.AddFile(FileSystemRuntimeConfigLoader.DEFAULT_CONFIG_FILE_NAME, new MockFileData(mockConfig.ToJson()));
+            FileSystemRuntimeConfigLoader loader = new(fileSystem);
+            RuntimeConfigProvider provider = new(loader) { IsLateConfigured = true };
```
Aniruddh25: why IsLateConfigured = true? This test can be combined with the previous test, with a DataRow for userdelegatedAuth = true or false

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -671,6 +671,283 @@ public void ValidateStreamingLogicForEmptyCellsAsync()
...
+
+            // No HttpContext (simulates startup/metadata phase)
+            Mock<IHttpContextAccessor> httpContextAccessor = new();
+            httpContextAccessor.Setup(x => x.HttpContext).Returns(value: null);
+
+            Mock<ILogger<QueryExecutor<SqlConnection>>> queryExecutorLogger = new();
+            DbExceptionParser dbExceptionParser = new MsSqlDbExceptionParser(provider);
+
+            // Act
+            MsSqlQueryExecutor queryExecutor = new(provider, dbExceptionParser, queryExecutorLogger.Object, httpContextAccessor.Object);
```
Aniruddh25: The Arrange/Act sections of all these tests is similar, except for a difference in ConnectionString, can we have a common helper method to set things up?

src/Core/Resolvers/MsSqlQueryExecutor.cs
JerryNixon: Yes, always!

anushakolan: @Aniruddh25 I did not get this, we are actually indeed appending the value, not overwriting it. 

In `Startup.cs`, 

`_dataSourceBaseAppName[dataSourceName] = builder.ApplicationName ?? "DAB";`

`// Line 176 - RETRIEVE the saved original:
if (!_dataSourceBaseAppName.TryGetValue(dataSourceName, out string? baseAppName))                                                 
// baseAppName = "MyCustomApp" (customer's original value)

// Line 208 - APPEND to it:
ApplicationName = $"{baseAppName}|obo:{poolKeyHash}"
// Result = "MyCustomApp|obo:abc123"
`

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
anushakolan: It is not needed, removing the usage.

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -671,6 +671,212 @@ public void ValidateStreamingLogicForEmptyCellsAsync()
...
+
+            // Assert - without user context, should use base Application Name (no |obo: suffix)
+            // Note: The actual format includes version suffix, e.g., "BaseApp,dab_oss_2.0.0"
+            Assert.IsTrue(connBuilder.ApplicationName.StartsWith("BaseApp"),
+                $"Without user context, Application Name should start with 'BaseApp' but was '{connBuilder.ApplicationName}'");
+            Assert.IsFalse(connBuilder.ApplicationName.Contains("|obo:"),
+                $"Without user context, Application Name should not contain '|obo:' but was '{connBuilder.ApplicationName}'");
+        }
+
+        #endregion
```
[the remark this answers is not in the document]
anushakolan: @copilot add a new test case to address this, don't create a new pull request but just add a new commit to this PR.

anushakolan:
@copilot code review[agent] cherry-pick 94bd1480ada78c7d40efc2cfc668972d0a0e472e commit onto my current PR.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +161,127 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+
+            // Per-user pooling is automatic when OBO is enabled.
+            // _dataSourceBaseAppName is only populated for data sources with user-delegated-auth enabled.
+            if (!_dataSourceBaseAppName.TryGetValue(dataSourceName, out string? baseAppName))
+            {
+                // OBO not enabled for this data source, use the standard connection string
+                return baseConnectionString;
+            }
+
+            // Extract user pool key from current HTTP context (prefers oid, falls back to sub)
```
RubenCerna2079, line 183: What does `oid` and `sub` mean?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +161,127 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+            // Extract issuer claim - required for tenant isolation and connection pool security.
+            // The "iss" claim must be present along with a user identifier (oid/sub) for per-user pooling.
+            // Callers are responsible for enforcing fail-safe behavior when claims are missing.
+            string? iss = user.FindFirst("iss")?.Value;
+
+            // Prefer oid (stable GUID), fall back to sub for guest/B2B users
+            string? userKey = user.FindFirst("oid")?.Value
+                ?? user.FindFirst("http://schemas.microsoft.com/identity/claims/objectidentifier")?.Value
+                ?? user.FindFirst("sub")?.Value
+                ?? user.FindFirst(ClaimTypes.NameIdentifier)?.Value;
```
RubenCerna2079, line 258: Could you explain what is the difference between these 3?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +161,127 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+        /// Uses iss|(oid||sub) to ensure each unique user identity gets its own connection pool.
+        /// Prefers 'oid' (stable GUID) but falls back to 'sub' for guest/B2B users.
+        /// </summary>
+        /// <param name="dataSourceName">The data source name for logging purposes.</param>
+        /// <returns>A URL-safe Base64-encoded hash, or null if no user context is available.</returns>
+        private string? GetUserPoolKeyHash(string dataSourceName)
+        {
+            if (HttpContextAccessor?.HttpContext?.User is null)
+            {
+                return null;
```
RubenCerna2079, line 237: I think we should also return a debug log in here, similar to how it is done below

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +161,127 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+            }
+
+            // Create the pool key as iss|userKey and hash it to keep connection string small
+            string poolKey = $"{iss}|{userKey}";
+            return HashPoolKey(poolKey);
+        }
+
+        /// <summary>
+        /// Hashes the pool key using SHA512 to create a compact, URL-safe identifier.
```
RubenCerna2079: @Aniruddh25 why do we use SHA512 and not SHA 256?

RubenCerna2079 (approved):
LGTM! Will approve assuming comments from myself and copilot will be resolved

src/Core/Resolvers/MsSqlQueryExecutor.cs
anushakolan: oid (object ID): A stable GUID uniquely identifying the user within the Entra ID tenant
sub (subject): A unique identifier for the user, used as fallback for guest/B2B users

src/Core/Resolvers/MsSqlQueryExecutor.cs
anushakolan: Added as a code comment.

src/Core/Resolvers/MsSqlQueryExecutor.cs
anushakolan: https://liquid.microsoft.com/Web/Object/Read/MS.Security/Requirements/Microsoft.Security.Cryptography.10021?support=1&support_category=Current+page+feedback&support_requestoption=Feedback

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +161,136 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+        }
+
+        /// <summary>
+        /// Hashes the pool key using SHA256 truncated to 16 bytes for a compact, URL-safe identifier.
+        /// Uses SHA256 (SHA-2 family) with 128-bit truncation per Microsoft security requirements.
+        /// This produces a ~22 character hash that fits well within SQL Server's 128-char Application Name limit
+        /// while providing sufficient collision resistance.
+        /// </summary>
+        /// <param name="key">The pool key to hash (format: iss|oid or iss|sub).</param>
+        /// <returns>A URL-safe Base64-encoded hash of the key (~22 characters).</returns>
```
Aniruddh25, line 282:
```suggestion
        /// <returns>A URL-safe Base64-encoded hash of the key (16 characters).</returns>
```

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -150,6 +161,136 @@ public override SqlConnection CreateConnection(string dataSourceName)
...
+
+            // Create the pool key as iss|userKey and hash it to keep connection string small
+            string poolKey = $"{iss}|{userKey}";
+            return HashPoolKey(poolKey);
+        }
+
+        /// <summary>
+        /// Hashes the pool key using SHA256 truncated to 16 bytes for a compact, URL-safe identifier.
+        /// Uses SHA256 (SHA-2 family) with 128-bit truncation per Microsoft security requirements.
+        /// This produces a ~22 character hash that fits well within SQL Server's 128-char Application Name limit
```
Aniruddh25:
```suggestion
        /// This produces a 16 character hash that fits well within SQL Server's 128-char Application Name limit
```

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -671,6 +671,319 @@ public void ValidateStreamingLogicForEmptyCellsAsync()
...
+        /// </summary>
+        [DataTestMethod, TestCategory(TestCategory.MSSQL)]
+        [DataRow(true, DisplayName = "OBO enabled - per-user pooling is automatic")]
+        [DataRow(false, DisplayName = "OBO disabled - pooling not modified")]
+        public void TestPoolingBehaviorWithAndWithoutUserDelegatedAuth(bool enableUserDelegatedAuth)
+        {
+            // Arrange & Act
+            Mock<IHttpContextAccessor> httpContextAccessor = new();
+            (MsSqlQueryExecutor queryExecutor, RuntimeConfigProvider provider) = CreateQueryExecutorForPoolingTest(
+                connectionString: "Server=localhost;Database=test;Pooling=true;",
```
Aniruddh25: This test needs to be fixed to assert the Pooling property set in the connection string is never modified. If it was true, it stays true. If it was false, it will stay false irrespective of whether OBO is enabled or disabled. 

When OBO is enabled, we are inherently doing the pooling, even when Pooling is set to true. So the Pooling property value may still continue to be true.

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -671,6 +671,319 @@ public void ValidateStreamingLogicForEmptyCellsAsync()
...
+            // Format: {hash}|{user-custom-appname} where hash is ~22 chars
+            // Hash is placed first to ensure it's never truncated if app name exceeds 128 chars
+            Assert.IsTrue(connBuilder.ApplicationName.Contains("|"),
+                $"Application Name should contain '|' separator but was '{connBuilder.ApplicationName}'");
+            Assert.IsTrue(connBuilder.ApplicationName.Contains("TestApp"),
+                $"Application Name should contain 'TestApp' but was '{connBuilder.ApplicationName}'");
+            // Hash should be at the start (before the | separator)
+            string hashPart = connBuilder.ApplicationName.Split('|')[0];
+            Assert.IsTrue(hashPart.Length >= 20 && hashPart.Length <= 25,
+                $"Hash prefix should be ~22 chars but was {hashPart.Length} chars: '{hashPart}'");
```
Aniruddh25: ISnt hash prefix 16 characters now?

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -671,6 +671,319 @@ public void ValidateStreamingLogicForEmptyCellsAsync()
...
+            Assert.IsFalse(connBuilder.Pooling,
+                "User's Pooling=false setting should be respected, not overridden by OBO");
+        }
+
+        /// <summary>
+        /// Test that when OBO is enabled and user claims are present, CreateConnection returns
+        /// a connection string with a user-specific Application Name containing the pool hash.
+        /// </summary>
+        [TestMethod, TestCategory(TestCategory.MSSQL)]
+        public void TestOboWithUserClaims_ConnectionStringHasUserSpecificAppName()
```
Aniruddh25, line 794: Should add a test where the specific app name and hash prefix end up crossing the 128 char limit - to exercise that codepath is handled gracefully to truncate upto 128.

Aniruddh25 (changes requested):

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
anushakolan: The hash is 16 bytes but Base64 encoding expands it:

16 bytes → Base64 → 22 characters (without padding)
Formula: ceil(16 * 4 / 3) = 22 chars

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -671,6 +671,348 @@ public void ValidateStreamingLogicForEmptyCellsAsync()
...
+
+            SqlConnection conn = queryExecutor.CreateConnection(provider.GetConfig().DefaultDataSourceName);
+            SqlConnectionStringBuilder connBuilder = new(conn.ConnectionString);
+
+            // Assert - without user context, should use base Application Name (no hash prefix)
+            // Note: The actual format includes version suffix, e.g., "BaseApp,dab_oss_2.0.0"
+            Assert.IsTrue(connBuilder.ApplicationName.StartsWith("BaseApp"),
+                $"Without user context, Application Name should start with 'BaseApp' but was '{connBuilder.ApplicationName}'");
+            // When no user context, the app name should NOT have the hash prefix pattern
+            // (hash prefix is 16 bytes Base64-encoded = ~22 chars, followed by |)
```
Aniruddh25, line 932: how does a 16 byte hash have a length > 20 for the hashPart? Wouldnt it be 16 chars itself?

anushakolan: From my understanding and copilot inputs, Bytes and characters are different. Raw bytes (0-255) can't all be printed, so we Base64 encode them. Base64 converts 3 bytes into 4 printable characters. So, 16 bytes becomes 16 ÷ 3 × 4 ≈ 22 characters.

Issue #3149, opened 2026-02-19
Change default for Entity.cache.enabled and Entity.cache.level to inherit the Runtime.Cache values.


aaronburtle:
Currently we have Enity.cache.enabled utilizing a default value of false. We want this to simply align with the Runtime.cache.enabled, and same for cache.level from the Runtime. So if those values are not included in the entity, they will simply come from the global, Runtime value.

PR #3155, opened 2026-02-24
Inherit default values for EntityCacheOption.Enabled and EntityCacheOption.Level from RuntimeCacheOptions

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3149

## What is this change?

Instead of having static default values for the `EntityCacheOptions.Enabled` and `EntityCacheOptions.Level`, we inherit these default values from the `RuntimeCacheOptions`.

## How was this tested?

We add a test that validates the new behavior in `CachingConfigProcessingTests`

### DIFF
--- a/src/Cli.Tests/ConfigGeneratorTests.cs
+++ b/src/Cli.Tests/ConfigGeneratorTests.cs
@@ -176,6 +176,14 @@ public class ConfigGeneratorTests
                     ""provider"": ""AppService""
                         },
                   ""mode"": ""production""
+                                    },
+                                ""telemetry"": {
+                                    ""open-telemetry"": {
+                                        ""enabled"": true,
+                                        ""endpoint"": ""@env('OTEL_EXPORTER_OTLP_ENDPOINT')"",
+                                        ""headers"": ""@env('OTEL_EXPORTER_OTLP_HEADERS')"",
+                                        ""service-name"": ""@env('OTEL_SERVICE_NAME')""
+                                    }
                   }
               },
               ""autoentities"": {},
--- a/src/Cli.Tests/EndToEndTests.cs
+++ b/src/Cli.Tests/EndToEndTests.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Azure.DataApiBuilder.Config.Converters;
 using Azure.DataApiBuilder.Product;
 using Cli.Constants;
 using Microsoft.Data.SqlClient;
@@ -116,7 +117,7 @@ public class EndToEndTests
         string[] args = { "init", "-c", TEST_RUNTIME_CONFIG_FILE, "--connection-string", SAMPLE_TEST_CONN_STRING, "--database-type", "mssql", "--rest.path", "/rest-api", "--rest.enabled", "false", "--graphql.path", "/graphql-api" };
         Program.Execute(args, _cliLogger!, _fileSystem!, _runtimeConfigLoader!);
 
-        DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true);
+        DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true, envFailureMode: EnvironmentVariableReplacementFailureMode.Ignore);
         Assert.IsTrue(_runtimeConfigLoader!.TryLoadConfig(
             TEST_RUNTIME_CONFIG_FILE,
             out RuntimeConfig? runtimeConfig,
@@ -196,7 +197,7 @@ public class EndToEndTests
 
         Program.Execute(args.ToArray(), _cliLogger!, _fileSystem!, _runtimeConfigLoader!);
 
-        DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true);
+        DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true, envFailureMode: EnvironmentVariableReplacementFailureMode.Ignore);
         Assert.IsTrue(_runtimeConfigLoader!.TryLoadConfig(
             TEST_RUNTIME_CONFIG_FILE,
             out RuntimeConfig? runtimeConfig,
@@ -273,7 +274,7 @@ public class EndToEndTests
         // Perform assertions on various properties.
         Assert.IsNotNull(runtimeConfig);
         Assert.IsNotNull(runtimeConfig.Runtime);
-        Assert.IsNull(runtimeConfig.Runtime.Telemetry);
+        Assert.IsNotNull(runtimeConfig.Runtime.Telemetry);
 
         string[] addTelemetryArgs;
         if (appInsightsEnabled is null)
--- a/src/Cli.Tests/EnvironmentTests.cs
+++ b/src/Cli.Tests/EnvironmentTests.cs
@@ -165,8 +165,7 @@ public class EnvironmentTests
         string? output = await process.StandardError.ReadLineAsync();
         Assert.AreEqual("Deserialization of the configuration file failed during a post-processing step.", output);
         output = await process.StandardError.ReadToEndAsync();
-        StringAssert.Contains(output, "Environmental Variable, "
-            + expectedEnvVarName + ", not found.", StringComparison.Ordinal);
+        StringAssert.Contains(output, "A valid Connection String should be provided.", StringComparison.Ordinal);
         process.Kill();
     }
 
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestAddingStoredProcedureWithRestMethodsAndGraphQLOperations.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestAddingStoredProcedureWithRestMethodsAndGraphQLOperations.verified.txt
@@ -47,6 +47,14 @@
       Authentication: {
         Provider: AppService
       }
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
   Entities: [
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceAsStoredProcedure.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceAsStoredProcedure.verified.txt
@@ -47,6 +47,14 @@
       Authentication: {
         Provider: AppService
       }
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
   Entities: [
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceWithDefaultType.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithSourceWithDefaultType.verified.txt
@@ -47,6 +47,14 @@
       Authentication: {
         Provider: AppService
       }
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
   Entities: [
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithoutIEnumerables.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestConfigGeneratedAfterAddingEntityWithoutIEnumerables.verified.txt
@@ -48,6 +48,14 @@
         Provider: AppService
       },
       Mode: Production
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
   Entities: [
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestInitForCosmosDBNoSql.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestInitForCosmosDBNoSql.verified.txt
@@ -53,6 +53,14 @@
         Provider: AppService
       },
       Mode: Production
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
   Entities: []
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestUpdatingStoredProcedureWithRestMethods.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestUpdatingStoredProcedureWithRestMethods.verified.txt
@@ -47,6 +47,14 @@
       Authentication: {
         Provider: AppService
       }
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
   Entities: [
--- a/src/Cli.Tests/Snapshots/EndToEndTests.TestUpdatingStoredProcedureWithRestMethodsAndGraphQLOperations.verified.txt
+++ b/src/Cli.Tests/Snapshots/EndToEndTests.TestUpdatingStoredProcedureWithRestMethodsAndGraphQLOperations.verified.txt
@@ -47,6 +47,14 @@
       Authentication: {
         Provider: AppService
       }
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
   Entities: [
--- a/src/Cli.Tests/Snapshots/InitTests.CosmosDbNoSqlDatabase.verified.txt
+++ b/src/Cli.Tests/Snapshots/InitTests.CosmosDbNoSqlDatabase.verified.txt
@@ -49,6 +49,14 @@
         Provider: AppService
       },
       Mode: Production
+    },
+    Telemetry: {
+      OpenTelemetry: {
+        Enabled: true,
+        Endpoint: @env('OTEL_EXPORTER_OTLP_ENDPOINT'),
+        Headers: @env('OTEL_EXPORTER_OTLP_HEADERS'),
+        ServiceName: @env('OTEL_SERVICE_NAME')
+      }
     }
   },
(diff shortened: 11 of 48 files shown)


src/Config/ObjectModel/EntityCacheOptions.cs
```diff
@@ -30,26 +30,29 @@ public record EntityCacheOptions
...
     /// The cache levels to use for a cache entry.
     /// </summary>
     [JsonPropertyName("level")]
-    public EntityCacheLevel? Level { get; init; } = null;
+    public EntityCacheLevel? Level { get; init; }
 
     [JsonConstructor]
     public EntityCacheOptions(bool? Enabled = null, int? TtlSeconds = null, EntityCacheLevel? Level = null)
     {
-        // TODO: shouldn't we apply the same "UserProvidedXyz" logic to Enabled, too?
```
Aniruddh25, line 52: Dont we need the "UserProvidedxyz" logic to Enabled?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -592,36 +593,79 @@ public virtual EntityCacheLevel GetEntityCacheEntryLevel(string entityName)
...
+    /// The level is inferred from the runtime cache Level2 configuration:
+    /// if Level2 is enabled, the level is L1L2; otherwise L1.
+    /// If runtime cache is not configured, the default cache level is used.
+    /// </summary>
+    /// <returns>Cache level that a cache entry should be stored in.</returns>
+    public virtual EntityCacheLevel GlobalCacheEntryLevel()
+    {
+        return Runtime?.Cache is not null
+            ? Runtime.Cache.InferredLevel
+            : EntityCacheOptions.DEFAULT_LEVEL;
```
Aniruddh25: The Default cache level when level 2 Cache section is NOT specified in runtime is L1 (as is evident from the InferredCacheLevel). But for EntityCacheOptions.DEFAULT_LEVEL it seems to be L1L2. Is that accurate and what we want?

src/Config/Converters/EntityCacheOptionsConverterFactory.cs
```diff
@@ -55,7 +55,9 @@ public EntityCacheOptionsConverter(DeserializationVariableReplacementSettings? r
         {
             if (reader.TokenType is JsonTokenType.StartObject)
             {
-                bool? enabled = false;
+                // Default to null (unset) so that an empty cache object ("cache": {})
```
Aniruddh25, line 58: Fix PR description. The following should be deleted - its remnant from the template. 

> Sample Request(s)
> Example REST and/or GraphQL request to demonstrate modifications
> Example of CLI usage to demonstrate modifications

src/Config/ObjectModel/EntityCacheOptions.cs
aaronburtle: I don't think so, because in the case that it is not user provided the value will be null, that let's us distinguish between if its user set or not.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: This is misleading because for us to get to the default level here, caching would be turned off at the runtime level. I will update so it isn't confusing.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -592,36 +593,79 @@ public virtual EntityCacheLevel GetEntityCacheEntryLevel(string entityName)
...
+    /// If the entity does not set Enabled (Enabled is null), the runtime cache enabled value is inherited.
+    /// Using Enabled.HasValue instead of a separate UserProvided flag ensures correctness
+    /// regardless of whether the object was created via JsonConstructor or with-expression.
     /// </summary>
-    /// <returns>Whether cache operations should proceed.</returns>
-    public virtual bool CanUseCache()
+    /// <param name="entityName">Name of the entity to check cache configuration.</param>
+    /// <returns>Whether caching is enabled for the entity.</returns>
+    /// <exception cref="DataApiBuilderException">Raised when an invalid entity name is provided.</exception>
+    public virtual bool IsEntityCachingEnabled(string entityName)
```
Aniruddh25: What happened to the function IsCachingEnabled defined in Entity.cs? Are we deleting that? If not, where would that be called now?

If we are deleting, this function should be defined in Entity.cs itself or simply modify the existing IsCachingEnabled function. We do not need to introduce this new function.

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
```diff
@@ -346,6 +346,115 @@ public void DefaultTtlNotWrittenToSerializedJsonConfigFile(string cacheConfig)
...
+    /// <summary>
+    /// Validates that GlobalCacheEntryLevel infers the cache level from the runtime cache Level2 configuration.
+    /// When Level2 is enabled, the global level is L1L2; when Level2 is absent or disabled, the global level is L1.
+    /// </summary>
+    /// <param name="globalCacheConfig">Global cache configuration JSON fragment.</param>
+    /// <param name="expectedLevel">Expected inferred cache level.</param>
+    [DataRow(@",""cache"": { ""enabled"": true }", EntityCacheLevel.L1, DisplayName = "Global cache enabled, no Level2: inferred level is L1.")]
+    [DataRow(@",""cache"": { ""enabled"": true, ""level-2"": { ""enabled"": true } }", EntityCacheLevel.L1L2, DisplayName = "Global cache enabled, Level2 enabled: inferred level is L1L2.")]
+    [DataRow(@",""cache"": { ""enabled"": true, ""level-2"": { ""enabled"": false } }", EntityCacheLevel.L1, DisplayName = "Global cache enabled, Level2 disabled: inferred level is L1.")]
+    [DataRow(@"", EntityCacheLevel.L1L2, DisplayName = "No global cache: default level is L1L2.")]
```
Aniruddh25: default level when no global cache should be L1. Similar to the case above when Global cache is enabled but L2 is not specified, L1 is preferred.

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
```diff
@@ -346,6 +346,115 @@ public void DefaultTtlNotWrittenToSerializedJsonConfigFile(string cacheConfig)
...
+
+    /// <summary>
+    /// Validates that GetEntityCacheEntryLevel returns the entity-level explicit value when set,
+    /// and falls back to the runtime-inferred level (from Level2) when the entity doesn't set it.
+    /// </summary>
+    /// <param name="globalCacheConfig">Global cache configuration JSON fragment.</param>
+    /// <param name="entityCacheConfig">Entity cache configuration JSON fragment.</param>
+    /// <param name="expectedLevel">Expected cache level returned by GetEntityCacheEntryLevel.</param>
+    [DataRow(@",""cache"": { ""enabled"": true, ""level-2"": { ""enabled"": true } }", @",""cache"": { ""enabled"": true }", EntityCacheLevel.L1L2, DisplayName = "Level2 enabled, entity has no level: inherits L1L2 from runtime.")]
+    [DataRow(@",""cache"": { ""enabled"": true }", @",""cache"": { ""enabled"": true }", EntityCacheLevel.L1, DisplayName = "No Level2, entity has no level: inherits L1 from runtime.")]
```
Aniruddh25: This DataRow  

>    [DataRow(@",""cache"": { ""enabled"": true }", @",""cache"": { ""enabled"": true }" 


has the same effect as the one in the above test:

>     [DataRow(@",""cache"": { ""enabled"": true }", EntityCacheLevel.L1, DisplayName = "Global cache enabled, no Level2: inferred level is L1.")]

Please go through each DataRow in this test and above to ensure we only have unique cases. The two test methods can be combined into a single method with multiple DataRows to check for level.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Good point, I added this with the added inheritance logic, but we can just improve the function from entity.cs. Originally I used this function in the RuntimeConfig because we dont have access to that class from entity.cs, but instead we will handle the logic at deserialization so we can keep this function working.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -492,12 +551,22 @@ public virtual int GetEntityCacheEntryTtl(string entityName)
                 subStatusCode: DataApiBuilderException.SubStatusCodes.EntityNotFound);
         }
 
-        if (entityConfig.Cache is not null && entityConfig.Cache.UserProvidedTtlOptions)
+        if (!entityConfig.IsCachingEnabled)
```
RubenCerna2079: Why are we reverting back to the previous changes?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -568,29 +643,41 @@ public virtual bool CanUseCache()
...
+
+        foreach (KeyValuePair<string, Entity> kvp in entities)
+        {
+            Entity entity = kvp.Value;
+
+            // If entity has no cache config at all, and global is enabled, create one inheriting enabled.
+            // If entity has cache config but Enabled is null, inherit the global value.
+            if (entity.Cache is null && globalCacheEnabled)
+            {
+                entity = entity with { Cache = new EntityCacheOptions(Enabled: true) };
```
RubenCerna2079: This will not be accidentally deserialized into the config right?

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: This is changed in the other cache related PR; https://github.com/Azure/data-api-builder/pull/3156/changes

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Good catch, adding the logic so we know when this is user defined so we won't write back and pollute the config.

src/Config/Converters/EntityCacheOptionsConverterFactory.cs
```diff
@@ -127,16 +129,25 @@ public EntityCacheOptionsConverter(DeserializationVariableReplacementSettings? r
         /// </summary>
         public override void Write(Utf8JsonWriter writer, EntityCacheOptions value, JsonSerializerOptions options)
         {
+            // If the cache object was not provided by the user (e.g., synthesized by
```
Aniruddh25: Why making this change, we typically dont write nulls to config. And any way null are ignored as specified by WhenWritingNull

aaronburtle: was a hack to avoid writing, but after our call I update to use the user provided flags instead.

src/Cli.Tests/Snapshots/AddEntityTests.AddEntityWithCachingEnabled.verified.txt
```diff
@@ -51,8 +51,7 @@
           Enabled: true,
           TtlSeconds: 1,
           Level: L1L2,
-          UserProvidedLevelOptions: false,
-          UserProvidedCacheOptions: true
```
Aniruddh25: Are you sure           UserProvidedCacheOptions isnt needed?

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -661,16 +667,28 @@ private static RuntimeEntities ResolveEntityCacheInheritance(RuntimeEntities ent
...
-            // If entity has no cache config at all, and global is enabled, create one inheriting enabled.
-            // If entity has cache config but Enabled is null, inherit the global value.
-            if (entity.Cache is null && globalCacheEnabled)
+            // If entity has no cache config at all, and global is enabled,
+            // set InheritedCachingEnabled so IsCachingEnabled returns true
+            // without synthesizing a fake EntityCacheOptions that would pollute serialized config.
+            if (entity.Cache is null && isGlobalCacheEnabled)
             {
-                entity = entity with { Cache = new EntityCacheOptions(Enabled: true) { UserProvidedCacheOptions = false } };
+                entity = entity with { InheritedCachingEnabled = true };
```
Aniruddh25: the value of `InheritedCachingEnabled` here should be `isGlobalCacheEnabled`, otherwise it is always going to return true even when global cache may not be enabled.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -404,6 +406,10 @@ public RuntimeConfig(string Schema, DataSource DataSource, RuntimeOptions Runtim
         this.AzureKeyVault = AzureKeyVault;
 
         SetupDataSourcesUsed();
+
+        // Resolve entity cache inheritance: if an entity's Cache.Enabled is null,
+        // inherit the global runtime cache enabled setting.
+        this.Entities = ResolveEntityCacheInheritance(this.Entities, this.Runtime);
```
Aniruddh25: Why do you need to resolve the inheritance and store it in the entities.enabled property? This is doing unnecessary overhead each time the RuntimeConfig is initialized. 

Can you not lookup the global enabled property when making the decision for a particular entity if its caching is enabled or not? If it returns enabled returns null, you then check the global runtime config cache enabled value and make the decision based on that.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -404,6 +406,10 @@ public RuntimeConfig(string Schema, DataSource DataSource, RuntimeOptions Runtim
...
     }
```
Aniruddh25, line 406: This is not taking into account setting the property for autoentities either. So, best would be to look up at the decision making point. 

Check all existing references of Entity.IsCachingEnabled

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -588,14 +597,41 @@ public virtual EntityCacheLevel GetEntityCacheEntryLevel(string entityName)
...
         }
-        else
-        {
-            return EntityCacheLevel.L1L2;
-        }
+
+        // GlobalCacheEntryLevel() returns null when runtime cache is not configured.
+        // Callers guard with IsCachingEnabled, so null is not expected here,
+        // but we default to L1 defensively.
+        return GlobalCacheEntryLevel() ?? EntityCacheLevel.L1;
```
Aniruddh25: Why is there a discrepancy between the default cache level. EntityCacheOptions defines the DEFAULT LEVEL to be EntityCacheLevel.L1L2. But, here we are returning L1. We should be consistent.

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
```diff
@@ -346,6 +346,137 @@ public void DefaultTtlNotWrittenToSerializedJsonConfigFile(string cacheConfig)
...
+    /// Validates that GlobalCacheEntryLevel infers the cache level from the runtime cache Level2 configuration.
+    /// When Level2 is enabled, the global level is L1L2; when Level2 is absent or disabled, the global level is L1.
+    /// </summary>
+    /// <param name="globalCacheConfig">Global cache configuration JSON fragment.</param>
+    /// <param name="expectedLevel">Expected inferred cache level.</param>
+    [DataRow(@",""cache"": { ""enabled"": true }", EntityCacheLevel.L1, DisplayName = "Global cache enabled, no Level2: inferred level is L1.")]
+    [DataRow(@",""cache"": { ""enabled"": true, ""level-2"": { ""enabled"": true } }", EntityCacheLevel.L1L2, DisplayName = "Global cache enabled, Level2 enabled: inferred level is L1L2.")]
+    [DataRow(@",""cache"": { ""enabled"": true, ""level-2"": { ""enabled"": false } }", EntityCacheLevel.L1, DisplayName = "Global cache enabled, Level2 disabled: inferred level is L1.")]
+    [DataTestMethod]
+    public void GlobalCacheEntryLevel_InfersFromLevel2Config(
```
Aniruddh25, line 401: There doesnt seem to be any test checking if entity inherits the right caching level when its value is absent in the config

src/Cli.Tests/Snapshots/AddEntityTests.AddEntityWithCachingEnabled.verified.txt
aaronburtle: should be reverted now

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Good catch, this was actually gated behind isGlobalCacheEnabled being true, but its clearer to still use that instead of true.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Ya, that approach makes sense, I changed it to be a lazy resolution, so this is checked during access time and resolved then.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Yup, changed to do the look up at access time, which should handle this case as well.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Since the default was L1L2 before this change, will use that value.

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
aaronburtle: adding some tests for this.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -888,4 +808,46 @@ public LogLevel GetConfiguredLogLevel(string loggerFilter = "")
...
+    /// Otherwise, inherits the global runtime cache enabled setting.
+    /// </summary>
+    /// <param name="entity">The entity to check cache configuration.</param>
+    /// <returns>Whether caching is enabled for the entity.</returns>
+    private bool IsEntityCachingEnabled(Entity entity)
+    {
+        // If entity has an explicit cache config with user-provided enabled value, use it.
+        if (entity.Cache is not null && entity.Cache.UserProvidedEnabledOptions)
+        {
+            return entity.Cache.Enabled!.Value;
```
Aniruddh25: cant we use `entity.IsCachingEnabled` function here instead of directly accessing the Cache.Enabled! value? 

Where else is IsCachingEnabled used?

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
```diff
@@ -579,4 +580,124 @@ public void InheritedEntityCacheNotWrittenToSerializedJsonConfigFile(string glob
...
+    /// had caching disabled, which caused 400 errors for valid requests when the global cache was enabled.
+    /// These methods are now pure accessors that always return a value regardless of cache enablement.
+    /// The level falls back to GlobalCacheEntryLevel() which infers from Level2 configuration;
+    /// when Level2 is absent, the inferred level is L1.
+    /// </summary>
+    /// <param name="globalCacheConfig">Global cache configuration JSON fragment.</param>
+    /// <param name="entityCacheConfig">Entity cache configuration JSON fragment.</param>
+    /// <param name="expectedTtl">Expected TTL returned by GetEntityCacheEntryTtl.</param>
+    /// <param name="expectedLevel">Expected cache level returned by GetEntityCacheEntryLevel.</param>
+    [DataRow(@",""cache"": { ""enabled"": true, ""ttl-seconds"": 10 }", @",""cache"": { ""enabled"": false }", 10, EntityCacheLevel.L1, DisplayName = "Global cache enabled with custom TTL, entity cache disabled: entity returns global TTL and global inferred level.")]
```
Aniruddh25, line 670: if global default level is L1L2. How is this expecting L1?

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
```diff
@@ -579,4 +580,124 @@ public void InheritedEntityCacheNotWrittenToSerializedJsonConfigFile(string glob
...
+    [DataRow(@",""cache"": { ""enabled"": true }", @",""cache"": { ""enabled"": false }", 5, EntityCacheLevel.L1, DisplayName = "Global cache enabled with default TTL, entity cache disabled: entity returns default TTL and global inferred level.")]
```
Aniruddh25, line 671: global inferred level expected should be L1L2. Check the specs if it should be L1 or L2. Whatever it is the DEFAULT_LEVEL value should be same in Entity.cs/RuntimeConfig.cs

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Ya we should be using it here, otherwise its just used in tests.

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
aaronburtle: I think the confusion here is coming from the original spec when we implemented L2 cache.

At the Runtime level, we had `Runtime.Cache.Level-2.Enabled` set to `false` by default, but for Entities we had `Entity.Cache.Level` set to `L1L2` by default.

So the default behavior at the runtime level was really level 1, inferred from level 2 being disabled by default, but if you had caching enabled for an entity, L1L2 was the default. This still works though because L1L2 at the entity level with L1 at the runtime level just means you get L1. The entity by default doesn't discriminate between the two, we can simply infer which one to use based on the runtime and the entity. So if runtime is l2 enabled and entity is L1L2 then we can connect to the backplane, otherwise, we just use L1.

Now that we are inheriting the level from the runtime I suppose we can change the way this works a bit and use the runtime to align entities to either L1 when level-2.enabled=false at the runtime level, or L1L2 otherwise. This is a bit different than the behavior we had before but I think it will work because the only time we can be in an L2 scenario is when the runtime section has level-2.enabled as true, and in that case we will infer L1L2 as the entity cache level. Then if there is no runtime cache section, or if level-2.enabled is false, we will have L1 as the default behavior.

image

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
aaronburtle: The global default is level-2.enabled=false, but the default value for entities has always been L1L2. We can tweak this a bit so that the default value for entities is instead inferred from the runtime level.

aaronburtle: This is actually how we have it currently, lll make sure the tests are reflecting that.

Aniruddh25 (approved):

Issue #3148, opened 2026-02-19
[Bug]: Retrieving entity cache ttl when entity cache enabled is set to false throws an exception, should just return null


aaronburtle:
### What happened?

DAB was started using a config that defined an L2 cache in the Runtime section of the config. An entity without a caching section, or with cache set to false and additional debugging code was added to see the various properties during request time, a REST request was sent for that entity to dab, and dab returned bad request instead of correctly executing the request without cache, due to the exception related to getting entity ttl.

### Version

2.0

### What database are you using?

Azure SQL

### Which API approach are you accessing DAB through?

REST

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3156, opened 2026-02-24
Return default values from GetEntityCacheEntryTtl and GetEntityCacheEntryLevel when entity cache enabled is false

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3148

Test change to help the pipeline pass relates to https://github.com/Azure/data-api-builder/issues/2992

## What is this change?

Within the code paths for executing a REST request with the RuntimeCacheOptions enabled, we would attempt to get the entity cache ttl, however, if the entity had its cache options not enabled this would result in an exception, leading to a bad request returned to the user. However, the code path was already safeguarded by checking if we have entity cache enabled before executing a cache-based response, and so we did not need the exception when getting the entity cache ttl, and can safely return the default value.

## How was this tested?

A simple regression test is added that validates the new behavior.

### DIFF
--- a/src/Core/Resolvers/SqlQueryEngine.cs
+++ b/src/Core/Resolvers/SqlQueryEngine.cs
@@ -26,7 +26,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 {
     //<summary>
     // SqlQueryEngine to execute queries against Sql like databases.
-    //<summary>
+    //</summary>
     public class SqlQueryEngine : IQueryEngine
     {
         private readonly IMetadataProviderFactory _sqlMetadataProviderFactory;
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -5894,18 +5894,26 @@ type Planet @model(name:""PlanetAlias"") {
             HttpStatusCode responseCode = HttpStatusCode.ServiceUnavailable;
             while (retryCount < RETRY_COUNT)
             {
-                // Minimal MCP request (list tools) – valid JSON-RPC request
+                // Minimal MCP request (initialize) - valid JSON-RPC request.
+                // Using 'initialize' because 'tools/list' requires an active session
+                // in the MCP Streamable HTTP transport (ModelContextProtocol 1.0.0).
                 object payload = new
                 {
                     jsonrpc = "2.0",
                     id = 1,
-                    method = "tools/list"
+                    method = "initialize",
+                    @params = new
+                    {
+                        protocolVersion = "2025-03-26",
+                        capabilities = new { },
+                        clientInfo = new { name = "dab-test", version = "1.0.0" }
+                    }
                 };
                 HttpRequestMessage mcpRequest = new(HttpMethod.Post, mcp.Path)
                 {
                     Content = JsonContent.Create(payload)
                 };
-                mcpRequest.Headers.Add("Accept", "*/*");
+                mcpRequest.Headers.Add("Accept", "application/json, text/event-stream");
 
                 HttpResponseMessage mcpResponse = await httpClient.SendAsync(mcpRequest);
                 responseCode = mcpResponse.StatusCode;
--- a/.pipelines/cosmos-pipelines.yml
+++ b/.pipelines/cosmos-pipelines.yml
@@ -49,8 +49,6 @@ steps:
 
 - task: NuGetAuthenticate@1
   displayName: 'NuGet Authenticate'
-  inputs:
-    nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
 
 # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
 # Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
--- a/.pipelines/dwsql-pipelines.yml
+++ b/.pipelines/dwsql-pipelines.yml
@@ -42,8 +42,6 @@ jobs:
   steps:
   - task: NuGetAuthenticate@1
     displayName: 'NuGet Authenticate'
-    inputs:
-      nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
 
   # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
   # Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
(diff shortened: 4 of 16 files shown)


JerryNixon (approved):
Removing the old behavior

```csharp
if (!entityConfig.IsCachingEnabled)
{
        message: $"{entityName} does not have caching enabled.",
        statusCode: HttpStatusCode.BadRequest,
        subStatusCode: DataApiBuilderException.SubStatusCodes.NotSupported);
}
```

Now means all the internal calls need new gates to check if `IsCachingEnabled`. I can't tell if every code path added this, but it's important to note that these methods no longer throw and this behavior is drastically different.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -529,10 +526,7 @@ public virtual EntityCacheLevel GetEntityCacheEntryLevel(string entityName)
 
         if (!entityConfig.IsCachingEnabled)
         {
-            throw new DataApiBuilderException(
-                message: $"{entityName} does not have caching enabled.",
-                statusCode: HttpStatusCode.BadRequest,
-                subStatusCode: DataApiBuilderException.SubStatusCodes.NotSupported);
+            return EntityCacheLevel.L1L2;
         }
```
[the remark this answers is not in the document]
RubenCerna2079: I also agree with copilot on this

RubenCerna2079 (approved):
LGTM! Just need to solve the missing comments as well as address @JerryNixon concerns

aaronburtle:
> Removing the old behavior
> 
> ```cs
> if (!entityConfig.IsCachingEnabled)
> {
>     throw new DataApiBuilderException(
>         message: $"{entityName} does not have caching enabled.",
>         statusCode: HttpStatusCode.BadRequest,
>         subStatusCode: DataApiBuilderException.SubStatusCodes.NotSupported);
> }
> ```
> 
> Now means all the internal calls need new gates to check if `IsCachingEnabled`. I can't tell if every code path added this, but it's important to note that these methods no longer throw and this behavior is drastically different.

These code paths are already guarded, making this throw redundant and effectively a landmine if you ever call this in a scenario where entity cache enabled is set to false.

JerryNixon:
## Important

 - Unit tests should assert exact TTL value (not just non-null).
 - Add/confirm at least one integration-style test where entity cache is disabled but entity TTL exists, expecting global TTL.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -3225,7 +3225,7 @@ public async Task ValidateInheritanceOfReadPermissionFromAnonymous()
         /// <param name="query">GraphQL query/mutation text</param>
         /// <param name="queryName">GraphQL query/mutation name</param>
         /// <param name="authToken">Auth token for the graphQL request</param>
-        private static async void ValidateMutationSucceededAtDbLayer(TestServer server, HttpClient client, string query, string queryName, string authToken, string clientRoleHeader)
+        private static async Task ValidateMutationSucceededAtDbLayer(TestServer server, HttpClient client, string query, string queryName, string authToken, string clientRoleHeader)
```
Aniruddh25, line 3228: How is this change related to the entity ttl?

If its unrelated, please create a separate PR.

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
```diff
@@ -416,4 +416,47 @@ private static string GetRawConfigJson(string globalCacheConfig, string entityCa
...
+    /// Regression test: Validates that when global runtime cache is enabled but entity cache is disabled,
+    /// GetEntityCacheEntryTtl and GetEntityCacheEntryLevel return sensible defaults instead of throwing.
+    /// Previously, these methods threw a DataApiBuilderException (BadRequest/NotSupported) when the entity
+    /// had caching disabled, which caused 400 errors for valid requests when the global cache was enabled.
+    /// </summary>
+    /// <param name="globalCacheConfig">Global cache configuration JSON fragment.</param>
+    /// <param name="entityCacheConfig">Entity cache configuration JSON fragment.</param>
+    /// <param name="expectedTtl">Expected TTL returned by GetEntityCacheEntryTtl.</param>
+    /// <param name="expectedLevel">Expected cache level returned by GetEntityCacheEntryLevel.</param>
+    [DataRow(@",""cache"": { ""enabled"": true, ""ttl-seconds"": 10 }", @",""cache"": { ""enabled"": false }", 10, EntityCacheLevel.L1L2, DisplayName = "Global cache enabled with custom TTL, entity cache disabled: entity falls back to global TTL.")]
```
Aniruddh25: return null if entity cache is explicitly disabled.

src/Service.Tests/Caching/CachingConfigProcessingTests.cs
```diff
@@ -416,4 +416,47 @@ private static string GetRawConfigJson(string globalCacheConfig, string entityCa
...
+    [DataRow(@",""cache"": { ""enabled"": true }", @",""cache"": { ""enabled"": false }", DEFAULT_CACHE_TTL_SECONDS, EntityCacheLevel.L1L2, DisplayName = "Global cache enabled with default TTL, entity cache disabled: entity falls back to default TTL.")]
+    [DataRow(@",""cache"": { ""enabled"": true, ""ttl-seconds"": 10 }", @"", 10, EntityCacheLevel.L1L2, DisplayName = "Global cache enabled with custom TTL, entity cache omitted: entity falls back to global TTL.")]
```
Aniruddh25: In case, entity cache is omitted, it should fall back to global. So, this is accurate. Only when it is explicitly disabled, like in the above 2 DataRow, since the entity cache is disabled, it should return null.

Aniruddh25 (changes requested):
Need to return `null` ttl when the cache is disabled at entity level

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: These were flakey tests that were failing in this PR's runs in the pipeline, so I included the changes here to speed up the pipeline tests passing.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -26,7 +26,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 {
     //<summary>
     // SqlQueryEngine to execute queries against Sql like databases.
-    //</summary>
+    //<summary>
```
Aniruddh25: this is an accidental change.

Aniruddh25:
```suggestion
    //</summary>
```

PR #3176, opened 2026-02-27
Revert "Hot Chocolate 16 Upgrade (#2822)"

RubenCerna2079:
## Why make this change?
- We need to create the new stable release for version 1.7, this cannot be currently done since the branch currently has only part of the upgrades needed for Hot Chocolate 16 as it is still in release candidate mode.

## What is this change?
We got rid of all the changes that were made related to Hot Chocolate 16 back into version 15.
- Related PR: #2822
- This reverts commit 271cbf4cc3a1dcdc1f915e091bee6a2d64b8eb1e.

## How was this tested?
This is a reversion, only the current tests that exist need to pass.


souvikghosh04 (approved):
LGTM

PR #3192, opened 2026-03-04
Improved OBO logging and added OBO observability to SESSION_CONTEXT

### LINKED ISSUE #3125: [OBO] OBO Logging & Correlation
Add required observability for OBO without logging PII.

 1. Log:
    1. OBO success/failure
    2. token cache hit/miss
    3. user identifier (oid OR sub)
    4. tenant ID
  
2. Propagate OpenTelemetry correlation ID
3. Set correlation values in SQL SESSION_CONTEXT when enabled
4. Ensure SESSION_CONTEXT is not used for authorization

anushakolan:
## Why make this change? ##
Closes #3125 
1. Adds OBO (On-Behalf-Of) observability improvements for debugging and auditing user-delegated authentication flows.
2. Enables end-to-end tracing from DAB application logs to SQL Server queries using OpenTelemetry correlation IDs.
3. Provides SQL-side visibility into which user and tenant executed a query when using user-delegated auth.

## What is this change? ##
1. Enhanced OBO Token Logging (OboSqlTokenProvider.cs)

- Added structured logging with EventType field for easier log filtering and analysis
- Included traceId in all OBO-related log messages to enable correlation with distributed traces
- Event types added: OboValidationFailed, OboTokenCacheMiss, OboTokenAcquired, OboTokenCacheHit

2. SESSION_CONTEXT Observability Values (MsSqlQueryExecutor.cs)
 Added the following values to SQL Server SESSION_CONTEXT when set-session-context is enabled:

- dab.trace_id - OpenTelemetry TraceId (set when Activity is present)
- dab.span_id - OpenTelemetry SpanId (set when Activity is present)
- dab.auth_type - Auth type indicator, value: obo (set when user-delegated auth is enabled)
- dab.user_id - User identifier from oid claim, falls back to sub claim (set when user-delegated auth is enabled)
- dab.tenant_id - Azure AD tenant ID from tid claim (set when user-delegated auth is enabled)
image

### Unit Tests ###
1. GetSessionParamsQuery_IncludesAllObservabilityValues_WhenActivityAndOboEnabled
2. GetSessionParamsQuery_ExcludesCorrelationIds_WhenNoActivity

### Manual E2E Testing ###

- Deployed to Azure Container App with Azure SQL Database
- Verified OBO token logging in container logs with traceId correlation
- Verified SESSION_CONTEXT values via stored procedure (dbo.GetSessionContext)
- Confirmed correlation between container logs and SQL SESSION_CONTEXT values
image

### DIFF
--- a/.pipelines/mssql-pipelines.yml
+++ b/.pipelines/mssql-pipelines.yml
@@ -1,9 +1,10 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# MsSql Integration Testing Pipeline config is split into two jobs:
-# 1) LinuxTests -> Run SQL Server 2019 in Linux Docker Image
-# 2) WindowsTests -> Run LocalDB preinstalled on machine
+# MsSql Integration Testing Pipeline config is split into parallel jobs:
+# 1) linux (disabled) -> Run SQL Server 2019 in Linux Docker Image
+# 2) windows_combined -> GraphQL, REST, Unit, HotReload, OpenApi, Auth, Telemetry, Caching on LocalDB
+# 3) windows_configuration -> Configuration tests on LocalDB (with schema init)
 
 trigger:
   batch: true
@@ -151,7 +152,19 @@ jobs:
       summaryFileLocation: '$(Agent.TempDirectory)/**/*cobertura.xml'
 
 
-- job: windows
+# MsSql Integration Testing is split into two parallel jobs (~20 min each):
+#
+# 1) windows_combined -> GraphQL, HotReload, REST, Unit, OpenApi, Auth,
+#    Telemetry, and Caching tests.
+#    SqlTestBase-inheriting tests (GraphQL, REST) create the DB schema;
+#    the remaining tests find it already in place.
+#
+# 2) windows_configuration -> Pure ConfigurationTests.
+#    No SqlTestBase tests in this bucket, so initDbSchema creates the
+#    schema via SqlClient before tests run.
+
+- job: windows_combined
+  displayName: 'Windows - Combined Integration Tests'
   pool:
     vmImage: 'windows-latest'
   variables:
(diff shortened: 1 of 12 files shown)


src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -520,6 +521,49 @@ public override string GetSessionParamsQuery(HttpContext? httpContext, IDictiona
                 sessionMapQuery = sessionMapQuery.Append(statementToSetReadOnlyParam);
             }
 
+            // Add OpenTelemetry correlation values for observability.
+            // These allow correlating database queries with distributed traces.
+            Activity? currentActivity = Activity.Current;
```
Aniruddh25: Who instantiates the Current activity? and when?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -520,6 +521,49 @@ public override string GetSessionParamsQuery(HttpContext? httpContext, IDictiona
...
+            if (currentActivity is not null)
```
Aniruddh25: Isnt OBO independent of whether the set-session-context property is set to true or not in the config? 

If yes, seems like the traces will get added only when that property is set to true as well - since thats when the `GetSessionParamsQuery` function could be called IIRC.

We need to insert this session context even if the claims in the for loop above are not added as long as OBO is enabled.

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -1013,6 +1013,206 @@ private static Mock<IHttpContextAccessor> CreateHttpContextAccessorWithAuthentic
...
+            // Arrange
+            TestHelper.SetupDatabaseEnvironment(TestCategory.MSSQL);
+
+            // Create runtime config with user-delegated-auth enabled and set-session-context
+            RuntimeConfig runtimeConfig = new(
+                Schema: "UnitTestSchema",
+                DataSource: new DataSource(
+                    DatabaseType: DatabaseType.MSSQL,
+                    ConnectionString: "Server=localhost;Database=TestDb;",
+                    Options: new Dictionary<string, object> { { "set-session-context", true } })
```
Aniruddh25, line 1034: we should test this when set-session-context is false and still expect OBO activity

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -1013,6 +1013,206 @@ private static Mock<IHttpContextAccessor> CreateHttpContextAccessorWithAuthentic
...
+        [TestMethod, TestCategory(TestCategory.MSSQL)]
+        public void GetSessionParamsQuery_ExcludesCorrelationIds_WhenNoActivity()
+        {
+            // Arrange
+            TestHelper.SetupDatabaseEnvironment(TestCategory.MSSQL);
+            RuntimeConfig runtimeConfig = new(
+                Schema: "UnitTestSchema",
+                DataSource: new DataSource(
+                    DatabaseType: DatabaseType.MSSQL,
+                    ConnectionString: "Server=localhost;Database=TestDb;",
```
Aniruddh25, line 1159: OBO doesnt seem to have been enabled here in this test.

src/Core/Resolvers/MsSqlQueryExecutor.cs
anushakolan: Activity.Current is not created in this SQL executor. It is created earlier in the request pipeline when DAB starts a request/query activity through its telemetry instrumentation. By the time session context values are prepared, that activity is the current one and flows through async execution context. It can still be null in non-request paths such as startup, metadata discovery, or background flows, which is why the null check is necessary.

E.g: `using Activity? activity = !isIntrospectionQuery ?
            TelemetryTracesHelper.DABActivitySource.StartActivity($"{method} /{route}") : null;`

src/Core/Resolvers/MsSqlQueryExecutor.cs
anushakolan: `GetSessionParamsQuery` call is not gated on set-session-context property so we shouldn't have an issue there.
However, I did the following changes to address your concern,
1. User claims are only added when set-session-context is true.
2. OBO observability values (dab.trace_id, dab.span_id, dab.auth_type, dab.user_id, dab.tenant_id) are added when OBO is enabled, regardless of set-session-context

src/Core/Resolvers/OboSqlTokenProvider.cs
```diff
@@ -66,21 +67,30 @@ public OboSqlTokenProvider(
     {
         if (principal is null)
         {
-            _logger.LogWarning("Cannot acquire OBO token: ClaimsPrincipal is null.");
+            _logger.LogWarning(
+                "{EventType}: Cannot acquire OBO token - ClaimsPrincipal is null (traceId: {TraceId}).",
```
aaronburtle, line 71: Should we use an `enum` for the `EventType` instead of just providing a string?

aaronburtle:
DAB currently has its own correlation system, which adds a correlation ID to the DAB logs. So if a user were to get say a 500 error on a REST call they would have in the DAB logs a correlation id associated with the error. The dab logs would also show the Obo Token being acquired with a separate traceid. But in the SQL Server session context, there will only be the trace id, not the dab correlation id.

So in order to connect the dab error log to the SQL audit, the user would have to go through some sort of APM tool to match the requests by timestamp, or theyd need to know that both IDs belong to the same http request.

If the dab.correlation_id was also included in the SESSION_CONTEXT, then the user could directly join the SQL audit log to the DAB error log without needing to use any other tools. This might be particularly useful if there are different people or teams looking at the different logs.

Not a blocker, but something I think we should consider.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -498,26 +499,85 @@ public override string GetSessionParamsQuery(HttpContext? httpContext, IDictiona
...
-                // Append statement to set read only param value - can be set only once for a connection.
-                string statementToSetReadOnlyParam = "EXEC sp_set_session_context " + $"'{claimType}', " + paramName + ", @read_only = 0;";
-                sessionMapQuery = sessionMapQuery.Append(statementToSetReadOnlyParam);
+                // Add OpenTelemetry correlation values for observability.
+                // These allow correlating database queries with distributed traces.
+                Activity? currentActivity = Activity.Current;
+                if (currentActivity is not null)
+                {
+                    string traceIdParamName = $"{SESSION_PARAM_NAME}{counter.Next()}";
+                    parameters.Add(traceIdParamName, new(currentActivity.TraceId.ToString()));
```
aaronburtle, line 551: Because of our retry policy in query executor, a transient fault may end up trying to re-add params that already exist in this dictionary, which would error out. This is an already existing issue however, and I dont think Ive seen this show up yet, but figured it is worth flagging, and i think can be fixed by changing from .add to using an indexer; parameters[traceIdParamName]=foo

aaronburtle (approved):
Looks good! Couple comments, but no blockers.

PR #3196, opened 2026-03-05
fix(mcp): allow create_record for views and added tests to verify the same.

### LINKED ISSUE #3194: [Bug]:  Create records tool not enabled for views
### What happened?

create_record is hard‑blocked for views, with the following error message:

`"The create_record tool is only available for tables."`

- create_record should be allowed for views
- The operation maps to SQL INSERT
- If the INSERT fails (non‑updatable view, constraints, triggers, etc.), the database should return the error, just like it does for tables
- Views are the required workaround for unsupported SQL data types, especially vector data types
DAB does not support vector columns directly, so users must:
     1. Create a view that omits vector columns
     2. Perform DML operations against that view

- Add automated tests to verify this scenario works fine for all tools.

### Version

1.7

Azure SQL

Local (including CLI)

_No response_

### Relevant log output

```Text

```

### Code of Conduct

anushakolan:
## Why make this change?
Closes #3194 - MCP `create_record` tool incorrectly blocks views with error `"The create_record tool is only available for tables."`
1. Views are the required workaround for unsupported SQL data types (e.g., vector columns). Users could create views that omit unsupported columns and perform DML operations against those views.
2. This bug prevents INSERT operations via MCP on view entities, breaking a critical workflow for vector database scenarios.

## What is this change?

1. Modified `CreateRecordTool.cs` to allow both tables and views to pass source type validation
2. Changed `else` block (which caught views) to`else if (EntitySourceType.StoredProcedure)` so only stored procedures are blocked
3. Views now fall through to the mutation engine, which already supports INSERT on updateable views
4. Updated error message to `"The create_record tool is only available for tables and views."`
5. Added 8 unit tests validating all DML tools (CreateRecord, ReadRecords, UpdateRecord, DeleteRecord) work with both Table and View source types

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests
        1. `DmlTool_AllowsTablesAndViews` - 8 DataRow test cases verifying no `InvalidCreateTarget` error for views
        2. Existing REST integration tests `InsertOneInViewTest` already validate view INSERT via same mutation engine

Manually tested via MCP Inspector, to verify `create_record` calls succeeds on a view.

### DIFF
--- a/.pipelines/cosmos-pipelines.yml
+++ b/.pipelines/cosmos-pipelines.yml
@@ -49,8 +49,6 @@ steps:
 
 - task: NuGetAuthenticate@1
   displayName: 'NuGet Authenticate'
-  inputs:
-    nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
 
 # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
 # Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
--- a/.pipelines/dwsql-pipelines.yml
+++ b/.pipelines/dwsql-pipelines.yml
@@ -42,8 +42,6 @@ jobs:
   steps:
   - task: NuGetAuthenticate@1
     displayName: 'NuGet Authenticate'
-    inputs:
-      nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
 
   # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
   # Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
@@ -172,8 +170,6 @@ jobs:
 
   - task: NuGetAuthenticate@1
     displayName: 'NuGet Authenticate'
-    inputs:
-      nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
   
   # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
   # Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
--- a/.pipelines/mssql-pipelines.yml
+++ b/.pipelines/mssql-pipelines.yml
@@ -43,8 +43,6 @@ jobs:
   steps:
   - task: NuGetAuthenticate@1
     displayName: 'NuGet Authenticate'
-    inputs:
-      nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
   
   # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
   # Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
@@ -176,8 +174,6 @@ jobs:
 
   - task: NuGetAuthenticate@1
     displayName: 'NuGet Authenticate'
-    inputs:
-      nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
   
   # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
   # Per Microsoft Learn Docs, "Selecting the .NET SDK version is independent from 
--- a/.pipelines/mysql-pipelines.yml
+++ b/.pipelines/mysql-pipelines.yml
@@ -41,8 +41,6 @@ jobs:
   steps:
   - task: NuGetAuthenticate@1
     displayName: 'NuGet Authenticate'
-    inputs:
-      nuGetServiceConnections: 'EngThriveNugetFeedAccessForSqlDab'
 
   # The .NET CLI commands in proceeding tasks use the .NET SDK version specified ("selected") here.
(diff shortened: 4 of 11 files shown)


aaronburtle (approved):
Looks good!

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
```diff
@@ -144,12 +144,12 @@ public async Task<CallToolResult> ExecuteAsync(
                         return McpResponseBuilder.BuildErrorResult(toolName, "ValidationFailed", $"Request validation failed: {ex.Message}", logger);
                     }
                 }
-                else
+                else if (dbObject.SourceType is EntitySourceType.StoredProcedure)
```
Aniruddh25: Did you check what kind of validation other delete and update tools have?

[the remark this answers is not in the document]
Aniruddh25: @anushakolan, this comment is valid

Aniruddh25: Why restrict stored procedures? What if the entity level dml config enables create tool for a stored procedure? And what if the stored procedure definition has inserts in it.

I know we have EXECUTE tool for stored procedures, but do we restrict a stored proc from other DML tools?

src/Service.Tests/Mcp/EntityLevelDmlToolConfigurationTests.cs
```diff
@@ -188,6 +188,61 @@ public async Task DynamicCustomTool_RespectsCustomToolDisabled()
...
+
+            // Act
+            CallToolResult result = await tool.ExecuteAsync(arguments, serviceProvider, CancellationToken.None);
+
+            // Assert - Should NOT be a source type blocking error (InvalidCreateTarget or InvalidEntity)
+            // Other errors like missing metadata are acceptable since we're testing source type validation
+            if (result.IsError == true)
+            {
+                JsonElement content = await RunToolAsync(tool, arguments, serviceProvider);
+
```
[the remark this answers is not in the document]
Aniruddh25: this is also valid

src/Service.Tests/Mcp/EntityLevelDmlToolConfigurationTests.cs
```diff
@@ -188,6 +188,61 @@ public async Task DynamicCustomTool_RespectsCustomToolDisabled()
...
+        [DataRow("DeleteRecord", "Table", "Book", "{\"entity\": \"Book\", \"keys\": {\"id\": 1}}", DisplayName = "DeleteRecord allows Table")]
+        [DataRow("DeleteRecord", "View", "BookView", "{\"entity\": \"BookView\", \"keys\": {\"id\": 1}}", DisplayName = "DeleteRecord allows View")]
+        public async Task DmlTool_AllowsTablesAndViews(string toolType, string sourceType, string entityName, string jsonArguments)
+        {
+            // Arrange
+            RuntimeConfig config = sourceType == "View"
+                ? CreateConfigWithViewEntity()
+                : CreateConfigWithDmlToolEnabledEntity();
+            IServiceProvider serviceProvider = CreateServiceProvider(config);
+            IMcpTool tool = CreateTool(toolType);
```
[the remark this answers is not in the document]
Aniruddh25: fix this

Aniruddh25 (changes requested):
Check for stored proc and fix copilot comments

src/Azure.DataApiBuilder.Mcp/BuiltInTools/CreateRecordTool.cs
anushakolan: create_record uses `InsertRequestContext` -> mutation engine generates `INSERT INTO table VALUES (...)`. 
Stored procedures require `StoredProcedureRequestContext` -> generates `EXEC proc @param=value`.
 They're incompatible request types.

I have updated the error message to convey more clarity.

anushakolan: After my changes, all 4 DML tools now have the same validation pattern:

if (dbObject.SourceType != EntitySourceType.Table && dbObject.SourceType != EntitySourceType.View)
{
    return McpResponseBuilder.BuildErrorResult(toolName, "InvalidEntity", 
        $"Entity '{entityName}' is not a table or view. For stored procedures, use the execute_entity tool instead.", 
        logger);
}

anushakolan: DML tools expect row-based operations on tables/views - they need column names, primary keys, and field values to build standard SQL statements like INSERT, UPDATE, DELETE, and SELECT.

Stored procedures work completely differently - they take named parameters and execute custom logic defined in the procedure. The mutation engine has separate code paths for these two scenarios.

If we let a stored procedure slip through to the table/view code path, it would fail with a confusing database error because column names don't exist on procedures. The upfront check gives users a clear message directing them to the correct tool.

PR #3199, opened 2026-03-06
[MCP] aggregate_records DML tool

### LINKED ISSUE #3178: [Enh]: add `aggregate_records` DML tool to MCP server
## What?

Allow models to answer: "How many products are there?" and "What is our most expensive product?"

## Why?

These are among the most common information discovery questions, a primary model use case.

## How?

Introduce a new tool: `aggregate_records` that reuses native GraphQL aggregation capabilities in DAB.

### Schema

```json
{
  "type": "object",
  "properties": {
    "entity": {
      "type": "string",
      "description": "Entity name with READ permission.",
      "required": true
    },
    "function": {
      "type": "string",
      "enum": ["count", "avg", "sum", "min", "max"],
      "description": "Aggregation function to apply.",
      "required": true
    },
    "field": {
      "type": "string",
      "description": "Field to aggregate. Use '*' for count.",
      "required": true
    },
    "distinct": {
      "type": "boolean",
      "description": "Apply DISTINCT before aggregating.",
      "default": false
    },
    "filter": {
      "type": "string",
      "description": "OData filter applied before aggregating (WHERE). Example: 'unitPrice lt 10'",
      "default": ""
    },
    "groupby": {
      "type": "array",
      "items": { "type": "string" },
      "description": "Fields to group by, e.g., ['category', 'region']. Grouped field values are included in the response.",
      "default": []
    },
    "orderby": {
      "type": "string",
      "enum": ["asc", "desc"],
      "description": "Sort aggregated results by the computed value. Only applies with groupby.",
      "default": "desc"
    },
    "having": {
      "type": "object",
      "description": "Filter applied after aggregating on the result (HAVING). Operators are AND-ed together.",
      "properties": {
        "eq":  { "type": "number", "description": "Aggregated value equals." },
        "neq": { "type": "number", "description": "Aggregated value not equals." },
        "gt":  { "type": "number", "description": "Aggregated value greater than." },
        "gte": { "type": "number", "description": "Aggregated value greater than or equal." },
        "lt":  { "type": "number", "description": "Aggregated value less than." },
        "lte": { "type": "number", "description": "Aggregated value less than or equal." },
        "in":  {
          "type": "array",
          "items": { "type": "number" },
          "description": "Aggregated value is in the given list."
        }
      }
    }
  },
  "required": ["entity", "function", "field"]
}
```

### Response Alias Convention

The aggregated value in the response is always aliased as `{function}_{field}`. For `count` with `"*"`, the alias is `count`.

### Examples

**Q1: "How many products are there?"**

```json
{
  "entity": "Product",
  "function": "count",
  "field": "*"
}
```

```sql
SELECT COUNT(*) AS count
FROM Product;
```

Example output:

| count |
| ----- |
| 77    |

---

**Q2: "What is the average price of products under $10?"**

```json
{
  "entity": "Product",
  "function": "avg",
  "field": "unitPrice",
  "filter": "unitPrice lt 10"
}
```

```sql
SELECT AVG(unitPrice) AS avg_unitPrice
FROM Product
WHERE unitPrice < 10;
```

Example output:

| avg_unitPrice |
| ------------- |
| 6.74          |

---

**Q3: "Which categories have more than 20 products?"**

```json
{
  "entity": "Product",
  "function": "count",
  "field": "*",
  "groupby": ["categoryName"],
  "having": {
    "gt": 20
  }
}
```

```sql
SELECT categoryName, COUNT(*) AS count
FROM Product
GROUP BY categoryName
HAVING COUNT(*) > 20;
```

Example output:

| categoryName | count |
| ------------ | ----- |
| Beverages    | 24    |
| Condiments   | 22    |

---

**Q4: "For discontinued products, which categories have a total revenue between $500 and $10,000?"**

```json
{
  "entity": "Product",
  "function": "sum",
  "field": "unitPrice",
  "filter": "discontinued eq true",
  "groupby": ["categoryName"],
  "having": {
    "gte": 500,
    "lte": 10000
  }
}
```

```sql
SELECT categoryName, SUM(unitPrice) AS sum_unitPrice
FROM Product
WHERE discontinued = 1
GROUP BY categoryName
HAVING SUM(unitPrice) >= 500
   AND SUM(unitPrice) <= 10000;
```

Example output:

| categoryName | sum_unitPrice |
| ------------ | ------------- |
| Seafood      | 1834.50       |
| Produce      | 742.00        |

---

**Q5: "How many distinct suppliers do we have?"**

```json
{
  "entity": "Product",
  "function": "count",
  "field": "supplierId",
  "distinct": true
}
```

```sql
SELECT COUNT(DISTINCT supplierId) AS count_supplierId
FROM Product;
```

Example output:

| count_supplierId |
| ---------------- |
| 29               |

---

**Q6: "Which categories have exactly 5 or 10 products?"**

```json
{
  "entity": "Product",
  "function": "count",
  "field": "*",
  "groupby": ["categoryName"],
  "having": {
    "in": [5, 10]
  }
}
```

```sql
SELECT categoryName, COUNT(*) AS count
FROM Product
GROUP BY categoryName
HAVING COUNT(*) IN (5, 10);
```

Example output:

| categoryName | count |
| ------------ | ----- |
| Grains       | 5     |
| Produce      | 5     |

---

**Q7: "What is the average distinct unit price per category, for categories averaging over $25?"**

```json
{
  "entity": "Product",
  "function": "avg",
  "field": "unitPrice",
  "distinct": true,
  "groupby": ["categoryName"],
  "having": {
    "gt": 25
  }
}
```

```sql
SELECT categoryName, AVG(DISTINCT unitPrice) AS avg_unitPrice
FROM Product
GROUP BY categoryName
HAVING AVG(DISTINCT unitPrice) > 25;
```

Example output:

| categoryName | avg_unitPrice |
| ------------ | ------------- |
| Meat/Poultry | 54.01         |
| Beverages    | 32.50         |

---

**Q8: "Which categories have the most products?"**

```json
{
  "entity": "Product",
  "function": "count",
  "field": "*",
  "groupby": ["categoryName"],
  "orderby": "desc"
}
```

```sql
SELECT categoryName, COUNT(*) AS count
FROM Product
GROUP BY categoryName
ORDER BY COUNT(*) DESC;
```

Example output:

| categoryName | count |
| ------------ | ----- |
| Confections  | 13    |
| Beverages    | 12    |
| Condiments   | 12    |
| Seafood      | 12    |

---

**Q9: "What are the cheapest categories by average price?"**

```json
{
  "entity": "Product",
  "function": "avg",
  "field": "unitPrice",
  "groupby": ["categoryName"],
  "orderby": "asc"
}
```

```sql
SELECT categoryName, AVG(unitPrice) AS avg_unitPrice
FROM Product
GROUP BY categoryName
ORDER BY AVG(unitPrice) ASC;
```

Example output:

| categoryName   | avg_unitPrice |
| -------------- | ------------- |
| Grains/Cereals | 20.25         |
| Condiments     | 23.06         |
| Produce        | 32.37         |

---

**Q10: "For categories with over $500 revenue from discontinued products, which has the highest total?"**

```json
{
  "entity": "Product",
  "function": "sum",
  "field": "unitPrice",
  "filter": "discontinued eq true",
  "groupby": ["categoryName"],
  "having": {
    "gt": 500
  },
  "orderby": "desc"
}
```

```sql
SELECT categoryName, SUM(unitPrice) AS sum_unitPrice
FROM Product
WHERE discontinued = 1
GROUP BY categoryName
HAVING SUM(unitPrice) > 500
ORDER BY SUM(unitPrice) DESC;
```

Example output:

| categoryName | sum_unitPrice |
| ------------ | ------------- |
| Seafood      | 1834.50       |
| Meat/Poultry | 1062.50       |
| Produce      | 742.00        |

souvikghosh04:
## Why make this change?

- Closes #3178

- Adds `aggregate_records` as a new DML tool to the MCP server, enabling models to answer common aggregation questions like "How many products are there?" and "What is our most expensive product?"
- Continuation of work from #3179 (stale PR with resolved review comments).

## What is this change?

- New `aggregate_records` DML tool (`AggregateRecordsTool.cs`) that generates SQL-level aggregation queries (`COUNT`, `AVG`, `SUM`, `MIN`, `MAX`) with support for `DISTINCT`, OData `$filter` (WHERE), `GROUP BY`, `HAVING` operators (`eq`, `neq`, `gt`, `gte`, `lt`, `lte`, `in`), `ORDER BY` (asc/desc), and cursor-based pagination (`first`/`after`) — all per the spec in #3178.
query-timeout configuration for MCP runtime options — allows setting a per-query timeout (1–600 seconds, default 30s) via `McpRuntimeOptions.QueryTimeout`, validated at startup.
- Config & CLI plumbing: `DmlToolsConfig` gains `AggregateRecords`/`UserProvidedAggregateRecords`; `McpRuntimeOptionsConverterFactory` handles query-timeout serialization; `ConfigureOptions.cs` updated for CLI configure support; JSON schema updated.
- Telemetry: Aggregation operations emit OpenTelemetry traces with structured error codes.
- Updated 37 CLI + 4 Service.Tests snapshot files to reflect new default properties.

## How was this tested?

- [x] Integration Tests — `AggregateRecordsToolTests.cs` (all 13 spec examples + edge cases), `McpQueryTimeoutTests.cs`, `EntityLevelDmlToolConfigurationTests.cs`, `McpToolRegistryTests.cs`
- [x] Unit Tests — `AggregateRecordsToolTests.cs` (unit), `McpTelemetryTests.cs`, `RequestParserUnitTests.cs`, `SqlQueryExecutorUnitTests.cs`

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
@@ -661,6 +661,11 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
             IHttpContextAccessor httpContextAccessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
             HttpContext? httpContext = httpContextAccessor.HttpContext;
 
+            if (httpContext is null)
+            {
+                return (null, McpErrorHelpers.PermissionDenied(toolName, entityName, "read", "No active HTTP request context.", logger));
+            }
+
             if (!McpAuthorizationHelper.ValidateRoleContext(httpContext, authResolver, out string roleCtxError))
             {
                 return (null, McpErrorHelpers.PermissionDenied(toolName, entityName, "read", roleCtxError, logger));
@@ -782,15 +787,22 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
             // Add groupby columns as LabelledColumns and GroupByMetadata.Fields
             foreach (string groupbyField in args.Groupby)
             {
-                sqlMetadataProvider.TryGetBackingColumn(entityName, groupbyField, out string? backingGroupbyColumn);
+                if (!sqlMetadataProvider.TryGetBackingColumn(entityName, groupbyField, out string? backingGroupbyColumn) || string.IsNullOrEmpty(backingGroupbyColumn))
+                {
+                    throw new DataApiBuilderException(
+                        message: $"GroupBy field '{groupbyField}' is not a valid field for entity '{entityName}'.",
+                        statusCode: System.Net.HttpStatusCode.BadRequest,
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                }
+
                 structure.Columns.Add(new LabelledColumn(
-                    dbObject.SchemaName, dbObject.Name, backingGroupbyColumn!, groupbyField, structure.SourceAlias));
-                structure.GroupByMetadata.Fields[backingGroupbyColumn!] = new Column(
-                    dbObject.SchemaName, dbObject.Name, backingGroupbyColumn!, structure.SourceAlias);
+                    dbObject.SchemaName, dbObject.Name, backingGroupbyColumn, groupbyField, structure.SourceAlias));
+                structure.GroupByMetadata.Fields[backingGroupbyColumn] = new Column(
+                    dbObject.SchemaName, dbObject.Name, backingGroupbyColumn, structure.SourceAlias);
             }
 
             // Build aggregation column using engine's AggregationColumn type.
-            AggregationType aggregationType = Enum.Parse<AggregationType>(args.Function);
+            AggregationType aggregationType = Enum.Parse<AggregationType>(args.Function, ignoreCase: true);
             AggregationColumn aggregationColumn = new(
                 dbObject.SchemaName, dbObject.Name, backingField, aggregationType, alias, args.Distinct, structure.SourceAlias);
 
--- a/src/Azure.DataApiBuilder.Mcp/Utils/McpTelemetryErrorCodes.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Utils/McpTelemetryErrorCodes.cs
@@ -41,6 +41,6 @@ namespace Azure.DataApiBuilder.Mcp.Utils
         /// <summary>
         /// Operation timed out error code.
         /// </summary>
-        public const string TIMEOUT = "Timeout";
+        public const string OPERATION_TIMEOUT = "OperationTimeout";
     }
 }
--- a/src/Azure.DataApiBuilder.Mcp/Utils/McpTelemetryHelper.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Utils/McpTelemetryHelper.cs
@@ -188,7 +188,7 @@ namespace Azure.DataApiBuilder.Mcp.Utils
             return ex switch
             {
                 OperationCanceledException => McpTelemetryErrorCodes.OPERATION_CANCELLED,
-                TimeoutException => McpTelemetryErrorCodes.TIMEOUT,
+                TimeoutException => McpTelemetryErrorCodes.OPERATION_TIMEOUT,
                 DataApiBuilderException dabEx when dabEx.SubStatusCode == DataApiBuilderException.SubStatusCodes.AuthenticationChallenge
                     => McpTelemetryErrorCodes.AUTHENTICATION_FAILED,
                 DataApiBuilderException dabEx when dabEx.SubStatusCode == DataApiBuilderException.SubStatusCodes.AuthorizationCheckFailed


src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,783 @@
...
+                if (isCountStar && distinct)
+                {
+                    return McpResponseBuilder.BuildErrorResult(toolName, "InvalidArguments",
+                        "Cannot use distinct=true with field='*'. DISTINCT requires a specific field name. Use a field name instead of '*' to count distinct values.", logger);
+                }
+
+                string? filter = root.TryGetProperty("filter", out JsonElement filterElement) ? filterElement.GetString() : null;
+                bool userProvidedOrderby = root.TryGetProperty("orderby", out JsonElement orderbyElement) && !string.IsNullOrWhiteSpace(orderbyElement.GetString());
+                string orderby = userProvidedOrderby ? (orderbyElement.GetString() ?? "desc") : "desc";
+
```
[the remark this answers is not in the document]
souvikghosh04: Fixed

src/Config/Converters/McpRuntimeOptionsConverterFactory.cs
```diff
@@ -150,6 +159,12 @@ public override void Write(Utf8JsonWriter writer, McpRuntimeOptions value, JsonS
                 JsonSerializer.Serialize(writer, value.Description, options);
             }
 
+            // Write query-timeout if it's user provided
+            if (value?.UserProvidedQueryTimeout is true && value.QueryTimeout.HasValue)
+            {
+                writer.WriteNumber("query-timeout", value.QueryTimeout.Value);
+            }
```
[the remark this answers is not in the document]
souvikghosh04: Fixed

src/Service.Tests/Mcp/AggregateRecordsToolTests.cs
```diff
@@ -0,0 +1,1089 @@
...
+            Assert.IsTrue(expectedMessage.Contains("database did not respond"), "Timeout message must explain the database didn't respond.");
+
+            // Verify message mentions large datasets
+            Assert.IsTrue(expectedMessage.Contains("large datasets"), "Timeout message must mention large datasets as a possible cause.");
+
+            // Verify message provides actionable remediation steps
+            Assert.IsTrue(expectedMessage.Contains("filter"), "Timeout message must suggest using a filter.");
+            Assert.IsTrue(expectedMessage.Contains("groupby"), "Timeout message must suggest reducing groupby fields.");
+            Assert.IsTrue(expectedMessage.Contains("first"), "Timeout message must suggest using pagination with first.");
+        }
```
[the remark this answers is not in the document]
souvikghosh04: tests have been refactored now

schemas/dab.draft.schema.json
```diff
@@ -275,6 +275,13 @@
               "description": "Allow enabling/disabling MCP requests for all entities.",
               "default": true
             },
+            "query-timeout": {
```
Aniruddh25: `query-timeout` is not called out in the spec, is it necessary?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,1070 @@
...
+                int fetchCount = args.First.Value + 1;
+                string offsetParam = BaseQueryStructure.GetEncodedParamName(structure.Counter.Next());
+                structure.Parameters.Add(offsetParam, new DbConnectionParam(offset));
+                string limitParam = BaseQueryStructure.GetEncodedParamName(structure.Counter.Next());
+                structure.Parameters.Add(limitParam, new DbConnectionParam(fetchCount));
+
+                string paginationClause = $" OFFSET {offsetParam} ROWS FETCH NEXT {limitParam} ROWS ONLY";
+
+                // Remove TOP N from the SELECT clause (TOP conflicts with OFFSET/FETCH)
+                sql = Regex.Replace(sql, @"SELECT TOP \d+", "SELECT");
```
Aniruddh25, line 917: `sql` should be a StringBuilder type instead of `string`

src/Service.Tests/Mcp/AggregateRecordsToolTests.cs
```diff
@@ -0,0 +1,629 @@
...
+            AssertSchemaProperty(properties, "entity", "string");
+            AssertSchemaProperty(properties, "function", "string");
+            AssertSchemaProperty(properties, "field", "string");
+            AssertSchemaProperty(properties, "distinct", "boolean");
+            AssertSchemaProperty(properties, "filter", "string");
+            AssertSchemaProperty(properties, "groupby", "array");
+            AssertSchemaProperty(properties, "orderby", "string");
+            AssertSchemaProperty(properties, "having", "object");
+            AssertSchemaProperty(properties, "first", "integer");
+            AssertSchemaProperty(properties, "after", "string");
```
Aniruddh25, line 77: What if the input args has an additional property?

src/Service/Utilities/McpStdioHelper.cs
```diff
@@ -78,15 +78,8 @@ public static bool RunMcpStdioHost(IHost host)
...
-            Mcp.Core.McpToolRegistry registry =
-                host.Services.GetRequiredService<Mcp.Core.McpToolRegistry>();
-            IEnumerable<Mcp.Model.IMcpTool> tools =
-                host.Services.GetServices<Mcp.Model.IMcpTool>();
-
-            foreach (Mcp.Model.IMcpTool tool in tools)
-            {
-                registry.RegisterTool(tool);
-            }
+            // Tools are already registered by McpToolRegistryInitializer (IHostedService)
```
Aniruddh25: Are you sure? And this change is unrelated to aggregate_records tool

src/Service.Tests/Mcp/McpQueryTimeoutTests.cs
```diff
@@ -0,0 +1,396 @@
...
+    /// - Custom value overrides default
+    /// - Timeout wrapping applies to all MCP tools via ExecuteWithTelemetryAsync
+    /// - Hot reload: changing config value updates behavior without restart
+    /// - Timeout surfaces as TimeoutException, not generic cancellation
+    /// - Telemetry maps timeout to TIMEOUT error code
+    /// </summary>
+    [TestClass]
+    public class McpQueryTimeoutTests
+    {
+        #region Default Value Tests
```
Aniruddh25: All the Default Value Tests are so trivial, no need to add them

src/Service.Tests/UnitTests/AggregateRecordsToolTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        #endregion
+
+        #region Error message builder tests
+
+        [DataTestMethod]
+        [DataRow("Product", DisplayName = "Product entity")]
+        [DataRow("LargeProductCatalog", DisplayName = "LargeProductCatalog entity")]
+        public void BuildTimeoutErrorMessage_ContainsExpectedContent(string entityName)
```
Aniruddh25: Seems like the AggregateRecordsToolTests under Mcp folder already cover some of these tests.

src/Service.Tests/UnitTests/AggregateRecordsToolTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        [DataTestMethod]
+        [DataRow("abc", 0, DisplayName = "non-numeric base64 → 0")]
+        [DataRow("-5", 0, DisplayName = "negative offset → 0")]
+        [DataRow("0", 0, DisplayName = "zero offset → 0")]
+        [DataRow("3", 3, DisplayName = "offset 3 round-trip")]
+        [DataRow("5", 5, DisplayName = "offset 5 round-trip")]
+        [DataRow("1000", 1000, DisplayName = "large offset round-trip")]
+        public void DecodeCursorOffset_Base64EncodedValue_ReturnsExpectedOffset(string rawValue, int expectedOffset)
```
Aniruddh25: I think the DecodeCursorOffset is definitely already tested under the Mcp Folder, can you please double check?

schemas/dab.draft.schema.json
souvikghosh04: `query-timeout` isn't in the `aggregate_records` spec specifically but it was added as an MCP runtime option that applies to all tools via `McpTelemetryHelper.ExecuteWithTelemetryAsync`. Aggregation queries can be expensive, so having a configurable timeout is an important safety net. It defaults to 30 seconds when not configured.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
souvikghosh04: The method receives sql from `queryBuilder.Build()` (which returns `string`) and performs at most 3 operations: one `Regex.Replace`, one `IndexOf` and one `Insert`. For this scale (~200 char SQL string, 2–3 mutations), `StringBuilder` adds allocation overhead without meaningful benefit. The engine's own query builders also return and manipulate strings rather than `StringBuilders` for post-processing. The method is called once per tool invocation, not in a loop, so it should be fine.

src/Service.Tests/Mcp/AggregateRecordsToolTests.cs
souvikghosh04: Extra properties are silently ignored. the parsing code only looks for known properties via `root.TryGetProperty("entity", ...)` etc. This matches the MCP protocol convention where models may send additional metadata that tools don't consume. None of the other DML tools (`read_records`, `create_record`, `update_record`, `delete_record`) reject unknown properties either. Adding `"additionalProperties": false` to the schema would be a breaking change if any MCP client sends extra fields.

src/Service/Utilities/McpStdioHelper.cs
souvikghosh04: Removed from this PR.

src/Service.Tests/UnitTests/AggregateRecordsToolTests.cs
souvikghosh04: Removed the overlapping ComputeAlias, DecodeCursorOffset, and error message builder tests from the UnitTests file. Kept only two test groups with unique edge cases not covered in the Mcp tests

src/Service.Tests/UnitTests/AggregateRecordsToolTests.cs
souvikghosh04: Confirmed. the Mcp tests cover null, empty, whitespace, invalid base64, non-numeric, zero, and positive offsets. Removed all the overlapping DecodeCursorOffset tests from the UnitTests file. Kept only the negative offset edge case ("-5" to 0) which isn't covered in the Mcp tests.

Aniruddh25 (approved):

schemas/dab.draft.schema.json
```diff
@@ -315,6 +315,34 @@
...
+                          "additionalProperties": false,
+                          "properties": {
+                            "enabled": {
+                              "type": "boolean",
+                              "description": "Enable/disable the aggregate-records tool.",
+                              "default": true
+                            },
+                            "query-timeout": {
+                              "type": "integer",
+                              "description": "Execution timeout in seconds for aggregate queries. Range: 1-600.",
```
anushakolan, line 337: nit: curious, how did we decide on this range?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,1075 @@
...
+    /// <summary>
+    /// Tool to aggregate records from a table/view entity configured in DAB.
+    /// Supports count, avg, sum, min, max with optional distinct, filter, groupby, having, orderby.
+    /// </summary>
+    public class AggregateRecordsTool : IMcpTool
+    {
+        public ToolType ToolType { get; } = ToolType.BuiltIn;
+
+        private static readonly HashSet<string> _validFunctions = new(StringComparer.OrdinalIgnoreCase) { "count", "avg", "sum", "min", "max" };
+        private static readonly HashSet<string> _validHavingOperators = new(StringComparer.OrdinalIgnoreCase) { "eq", "neq", "gt", "gte", "lt", "lte", "in" };
```
anushakolan, line 42: Are we not supporting `between`, `not in`?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,1075 @@
...
+                SqlQueryStructure structure = new(
+                    authCtx!.RequestContext, sqlMetadataProvider, authResolver, runtimeConfigProvider, gQLFilterParser, authCtx.HttpContext);
+
+                string? backingField = ResolveBackingField(args, entityName, sqlMetadataProvider, toolName, out CallToolResult? pkError, logger);
+                if (pkError != null)
+                {
+                    return pkError;
+                }
+
+                string alias = ComputeAlias(args.Function, args.Field);
```
anushakolan, line 252: Why do we need alias, it seems to be a keyword as well, what is the importance of it in this context?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,1075 @@
...
+            List<string> selectFields = new(args.Groupby);
+            if (!args.IsCountStar && !selectFields.Contains(args.Field, StringComparer.OrdinalIgnoreCase))
+            {
+                selectFields.Add(args.Field);
+            }
+
+            // Build and validate FindRequestContext
+            RequestValidator requestValidator = new(serviceProvider.GetRequiredService<IMetadataProviderFactory>(), runtimeConfigProvider);
+            FindRequestContext context = new(entityName, dbObject, true);
+            httpContext!.Request.Method = "GET";
```
anushakolan, line 698: Is this safe to explicitly mutate the `httpContext.Request.Method`.

anushakolan, line 698: `if (httpContext is null)
{
    return (null, McpErrorHelpers.PermissionDenied(toolName, entityName, "read",
        "No active HTTP request context.", logger));
}
`

We should first see if httpContext is not null or throw an exception when it is null, rather than asusming that it would not be null.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,1075 @@
...
+            string entityName,
+            ISqlMetadataProvider sqlMetadataProvider)
+        {
+            // Clear default columns from FindRequestContext
+            structure.Columns.Clear();
+
+            // Add groupby columns as LabelledColumns and GroupByMetadata.Fields
+            foreach (string groupbyField in args.Groupby)
+            {
+                sqlMetadataProvider.TryGetBackingColumn(entityName, groupbyField, out string? backingGroupbyColumn);
```
anushakolan: `if (!sqlMetadataProvider.TryGetBackingColumn(entityName, groupbyField, out var backingGroupbyColumn)
    || string.IsNullOrEmpty(backingGroupbyColumn))
{
        $"GroupBy field '{groupbyField}' is not a valid field for entity '{entityName}'.");
}`

we should add this check to make sure that if `TryGetBackingColumn` fails, we catch it.

schemas/dab.draft.schema.json
```diff
@@ -315,6 +315,34 @@
...
+                          "type": "object",
+                          "description": "Aggregate records tool configuration",
+                          "additionalProperties": false,
+                          "properties": {
+                            "enabled": {
+                              "type": "boolean",
+                              "description": "Enable/disable the aggregate-records tool.",
+                              "default": true
+                            },
+                            "query-timeout": {
```
anushakolan, line 335: nit: rename this to `queryTimeout`, matching other properties in this file

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,1075 @@
...
+                structure.Columns.Add(new LabelledColumn(
+                    dbObject.SchemaName, dbObject.Name, backingGroupbyColumn!, groupbyField, structure.SourceAlias));
+                structure.GroupByMetadata.Fields[backingGroupbyColumn!] = new Column(
+                    dbObject.SchemaName, dbObject.Name, backingGroupbyColumn!, structure.SourceAlias);
+            }
+
+            // Build aggregation column using engine's AggregationColumn type.
+            AggregationType aggregationType = Enum.Parse<AggregationType>(args.Function);
```
anushakolan: We could instead do, `AggregationType aggregationType = Enum.Parse<AggregationType>(args.Function, ignoreCase: true);` to be safe of case sensitive bugs.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
```diff
@@ -0,0 +1,1075 @@
...
+        /// exist in the entity's metadata. This early validation lets the model discover typos immediately.
+        /// </summary>
+        private static CallToolResult? ValidateFieldsExist(
+            AggregateArguments args,
+            string entityName,
+            ISqlMetadataProvider sqlMetadataProvider,
+            string toolName,
+            ILogger? logger)
+        {
+            if (!args.IsCountStar && !sqlMetadataProvider.TryGetBackingColumn(entityName, args.Field, out _))
```
anushakolan, line 625: Your schema says “avg/sum/min/max require a numeric field” but the implementation only checks field existence, not type.

src/Azure.DataApiBuilder.Mcp/Utils/McpTelemetryErrorCodes.cs
```diff
@@ -37,5 +37,10 @@ internal static class McpTelemetryErrorCodes
         /// Operation cancelled error code.
         /// </summary>
         public const string OPERATION_CANCELLED = "OperationCancelled";
+
+        /// <summary>
+        /// Operation timed out error code.
+        /// </summary>
+        public const string TIMEOUT = "Timeout";
```
anushakolan: nit: to be consistent with the others in the file we could rename this to `OPERATION_TIMEOUT`.

schemas/dab.draft.schema.json
Aniruddh25: this should stay `query-timeout` since its a dab config file property where we follow kebab casing. 

for reference see existing dab config properties like `dml-tools`.  This file is the schema so some properties like `additionalProperties` are as per the json config spec.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
Aniruddh25: no, wasn't in the spec, let's not add it now, we can add it for 2.1 if needed.

schemas/dab.draft.schema.json
souvikghosh04: The 30-second default follows the common `CommandTimeout` default in ADO.NET / SqlCommand. The 600 second (10-minute) upper bound was chosen as a practical ceiling - aggregate queries over very large tables can legitimately take several minutes but allowing anything beyond 10 minutes risks holding resources too long in a shared MCP context. The minimum of 1 second prevents accidental zero/negative timeouts. Happy to adjust if there's a different convention we should follow.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
souvikghosh04: The alias is the column label used in the SQL AS clause and as the JSON property name in the response (e.g., SELECT COUNT(*) AS count, SELECT AVG(price) AS avg_price). Without it, the raw SQL engine assigns an unnamed or engine-specific column name (e.g., (No column name) in SQL Server), making the JSON response unreliable for the LLM to parse. It's not a SQL keyword here - it's a variable that holds the computed label string like "count" or "avg_price".

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
souvikghosh04: Fixed. Added an explicit if (`httpContext` is null) check at the top of `AuthorizeRequestAsync` that returns a PermissionDenied error before any further access.

souvikghosh04: This follows the same pattern used by the existing ReadRecordsTool. MCP tools run over http/stdio transport rather than actual HTTP requests, so the HttpContext is the ambient context from the middleware pipeline. The RequestValidator.ValidateRequestContext and AuthorizationService.AuthorizeAsync calls downstream check Request.Method to determine the operation type, so we set it to "GET" to match a read/find operation. Since MCP tool invocations are scoped per-request and don't share the HttpContext across concurrent operations, mutating it here is safe.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
souvikghosh04: Fixed - the BuildAggregationStructure method now checks the return of TryGetBackingColumn and throws a DataApiBuilderException if the field is invalid. Note that ValidateFieldsExist already validates all groupby fields, so this is a defensive guard in case the code path ever changes.

src/Azure.DataApiBuilder.Mcp/Utils/McpTelemetryErrorCodes.cs
souvikghosh04: Fixed - renamed to OPERATION_TIMEOUT with value "OperationTimeout" to be consistent with OPERATION_CANCELLED.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
souvikghosh04: Currently, DAB's ISqlMetadataProvider doesn't expose a direct API to query column data types for a given entity field. Adding type checking would require either extending the metadata provider interface or querying SourceDefinition.Columns dictionary for the column's Type property. For count/count(*), any field type is valid so no check is needed. For avg/sum, the database engine itself will reject non-numeric columns with a clear SQL error (e.g., "Operand data type varchar is invalid for avg operator"), which we catch and surface to the MCP client.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
souvikghosh04: Fixed - updated to Enum.Parse<AggregationType>(args.Function, ignoreCase: true).

schemas/dab.draft.schema.json
anushakolan: Okay, this looks good. Only thing was if 600 is too long, as 10 minutes might be frustrating to the user, but if you think large tables could take up to 10 minutes, then fine. Thanks for clarifying.

Issue #3210, opened 2026-03-10
[Bug]: NugetCommand@2 causes timeoute for Nuget Package Restore job


aaronburtle:
### What happened?

In the pipeline the NugetCommand@2 task has a hard coded timeout which can cause the restore nuget package step to fail. This should be replaced with:

  - task: DotNetCoreCLI@2
    displayName: Restore NuGet packages
    inputs:
      command: restore
      projects: '$(solution)'
      feedsToUse: config
      nugetConfigPath: Nuget.config

### Version

1.7

### What database are you using?

Azure SQL

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3211, opened 2026-03-10
Use DotNetCoreCLI@2 for Nuget Restore to avoid 120 second timeout, and add RuntimeIdentifiers="" to avoid needing packages not in our feed

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3210

Closes https://github.com/Azure/data-api-builder/issues/3212

## What is this change?

We replace the `NuGetCommand@2` task with `DotNetCoreCLI@2 ` to avoid the hard coded 120 second timeout in the former which is causing pipeline failures.

We add `restoreArguments: '/p:RuntimeIdentifiers=""'` to the pipelines for build and test. Since these pipelines only build and test (no publish), runtime packs aren't needed. The argument overrides `RuntimeIdentifiers` to empty during restore, skipping those downloads.

## How was this tested?

Pipeline ran.

### DIFF
--- a/.pipelines/cosmos-pipelines.yml
+++ b/.pipelines/cosmos-pipelines.yml
@@ -69,6 +69,7 @@ steps:
     projects: '$(solution)'
     feedsToUse: config
     nugetConfigPath: Nuget.config
+    restoreArguments: '/p:RuntimeIdentifiers=""'
 
 - task: DotNetCoreCLI@2
   displayName: Build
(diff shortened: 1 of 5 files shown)


aaronburtle:
> LGTM, but please confirm if there are any security concerns to using `DotNetCoreCLI@2 `.

there should not be, we were already using this in the Linux jobs, so this aligns our Windows and Linux jobs to use the same tasks.

PR #3243, opened 2026-03-12
Fix IOException in MSSQL pipeline by disposing config file watchers

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/2992

Several configuration tests create a TestServer that sets up a `ConfigFileWatcher` to support hot-reload. When the file watcher detects a change, its callback reads the file, holding a read handle. Neither `ConfigFileWatcher` nor `FileSystemRuntimeConfigLoader` implemented `IDisposable`, so when the TestServer is disposed, the DI container could not clean up the file watcher. The watcher remained active, and if its callback was mid-read when `CleanupAfterEachTest` called `File.Delete`, the delete failed with an `IOException`

Hot reload of the data source in the tests were also creating failures via a race condition from a GraphQL query.

## What is this change?

`ConfigFileWatcher` now implements `IDisposable`. It stops raising events, unsubscribes the Changed handler, and disposes the underlying `IFileSystemWatcher`.

`FileSystemRuntimeConfigLoader` now implements `IDisposable`. It disposes the owned `ConfigFileWatcher`. As a DI singleton, it is now automatically disposed when the TestServer shuts down.

`CleanupAfterEachTest` retries with exponential backoff. As a safety net, `File.Delete` is retried up to 3 times on `IOException` (2s, 4s, 8s delays), matching the existing retry pattern used elsewhere. 

`Startup` changed to have the container create the service so that it will automatically dispose.

`using` keyword used on remaining test servers for disposal.

`ConfigurationHotReloadTests` now has a retry mechanism on a `GraphQL` query that was causing a race condition leading to failures.

## How was this tested?

Run against the pipeline. Several pipeline runs were initiated to add certainty that the fix is working, since this issue is intermittent.

https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=18957&view=results

https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=18958&view=results

https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=18952&view=results

https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=18959&view=results

https://dev.azure.com/sqldab/Data%20API%20builder%20Dependency%20Packages/_build/results?buildId=18960&view=results

### DIFF
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -1987,7 +1987,9 @@ type Moon {
             Assert.IsFalse(result.IsValid);
             Assert.IsFalse(EnumerableUtilities.IsNullOrEmpty(result.ValidationErrors));
             Assert.AreEqual(1, result.ErrorCount);
-            Assert.IsTrue(result.ErrorMessage.Contains("Total schema validation errors: 1\n> Required properties are missing from object: entities."));
+            // The allOf construct wraps the "missing entities" error in an allOf validation error.
+            // Verify the top-level error count and that the validation correctly identifies the config as invalid.
+            Assert.IsTrue(result.ErrorMessage.Contains("Total schema validation errors: 1\n>"));
         }
 
         /// <summary>
--- a/dab-config.json
+++ b/dab-config.json
@@ -28,7 +28,7 @@
         "allow-credentials": false
       },
       "authentication": {
-        "provider": "Simulator"
+        "provider": "Unauthenticated"
       },
       "mode": "development"
     }
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -135,7 +135,7 @@
                 "additionalProperties": false,
                 "properties": {
                   "set-session-context": {
-                    "type": "boolean",
+                    "$ref": "#/$defs/boolean-or-string",
                     "description": "Enable sending data to MsSql using session context"
                   }
                 }
@@ -169,6 +169,7 @@
     "data-source-files": {
       "type": "array",
       "description": "Names of runtime configuration files referencing additional databases.",
+      "minItems": 1,
       "default": null
     },
     "runtime": {
@@ -178,6 +179,7 @@
       "properties": {
         "pagination": {
           "type": "object",
+          "description": "Pagination settings for REST and GraphQL API responses.",
           "properties": {
             "max-page-size": {
               "type": [ "integer", "null" ],
@@ -192,7 +194,7 @@
               "minimum": 1
             },
             "next-link-relative": {
-              "type": "boolean",
+              "$ref": "#/$defs/boolean-or-string",
               "default": false,
               "description": "When true, nextLink in paginated results will use a relative URL."
             }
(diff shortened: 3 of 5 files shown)


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -710,9 +710,25 @@ type Moon {
         [TestCleanup]
         public void CleanupAfterEachTest()
```
souvikghosh04, line 711: The true auto-dispose (as mentioned in PR description) of `FileSystemRuntimeConfigLoader` when TestServer shuts down can be done with `using` block. tests in `ConfigurationTests.cs` create `TestServer` without using. If the server isn't explicitly disposed, the DI container's `IDisposable` cleanup may never run and the file watcher could be undisposed. retry in `CleanupAfterEachTest` covers this as a safety net but the more correct fix would be ensuring those TestServer instances are disposed. This could be a follow-up PR if not possible in this PR.

souvikghosh04 (approved):
Approved with a follow-up comment

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -710,9 +710,25 @@ type Moon {
...
+                int retryCount = 0;
+                const int maxRetries = 3;
+                while (true)
+                {
+                    try
+                    {
+                        File.Delete(CUSTOM_CONFIG_FILENAME);
+                        break;
+                    }
+                    catch (IOException) when (retryCount < maxRetries)
```
souvikghosh04: nit: may be log the exception message and trace

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: Agreed, the 20 or so TestServer instances that don't have `using` won't trigger the auto-dispose, Ill follow up on this in a future PR to keep the scope of this one from growing, and the CleanupAfterEachTest should cover us in the meantime.

Aniruddh25: @aaronburtle, 
If we dont add `using` in this same change, the PR's title is not serving its purpose. It is not disposing the file watchers. It is only implementing the dispose function. 

I think all those FileSystemRuntimeConfigLoader should be actually disposed as well in this same PR.

aaronburtle: Will add this in, and I think this will require changing how to register the config file watcher as well so that we will actually dispose properly. Will add that as well.

aaronburtle: Here is the documentation regarding automatic disposal, we may need to double check that we have the correct pattern for other services:

```
// Register example service in IServiceCollection.
builder.Services.AddSingleton(new ExampleService());
In the preceding code:

The ExampleService instance is not created by the service container.
The framework does not dispose of the services automatically.
The developer is responsible for disposing the services.
```

https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection/guidelines#services-not-created-by-the-service-container

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -666,13 +666,33 @@ await WaitForConditionAsync(
...
-            GQL_QUERY_NAME,
-            GQL_QUERY);
+
+        // Retry GraphQL request because metadata re-initialization happens asynchronously
+        // after the "Validated hot-reloaded configuration file" message. The metadata provider
+        // factory clears and re-initializes providers on the hot-reload thread, so requests
+        // arriving before that completes will fail with "Initialization of metadata incomplete."
+        JsonElement reloadGQLContents = default;
+        bool querySucceeded = false;
+        for (int attempt = 1; attempt <= 10; attempt++)
```
Aniruddh25, line 676: Please mention this retry attempt added for GraphQL too over here in the description.

Issue #3244, opened 2026-03-12
Fix MSSQL Integration Test timeout


aaronburtle:
We have many tests in our MSSQL Integration Tests pipeline, as well as considerable setup overhead. This leads to this pipeline sometimes taking longer than 60m to complete. But we have a hard limit of 60m to run a given pipeline. Therefore, we need a way to prevent these timeouts, either by speeding up the pipeline, or splitting it into smaller pieces.

PR #3245, opened 2026-03-12
Split MSSQL integration tests into parallel pipeline jobs and remove MSSQL category from generic unit tests

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3244

## What is this change?

This PR makes two changes to improve MSSQL integration test pipeline performance:

1. Removed TestCategory(TestCategory.MSSQL) from 5 unit test files that are not MSSQL-specific. These tests do not require a database and should run in the unit test pipeline instead:

      -  RequestValidatorUnitTests.cs
      -  EntitySourceNamesParserUnitTests.cs
      -  RequestContextUnitTests.cs
      -  RestServiceUnitTests.cs
      -  RuntimeConfigLoaderJsonDeserializerTests.cs

3. Split the single Windows MSSQL integration test job into 2 parallel jobs to reduce execution time:

      - Windows - Combined Integration Tests: GraphQL, REST, Unit, HotReload, OpenApi, Authorization, Telemetry, and Caching tests (1,246 tests). SqlTestBase-inheriting tests (GraphQL, REST) naturally create the DB schema for the co-located tests.
      
      - Window - Configuration Tests: Pure ConfigurationTests (207 tests). Since this job has no SqlTestBase tests, a System.Data.SqlClient-based schema initialization step creates the DB schema before test execution.

A new reusable step template `mssql-test-steps.yml` avoids duplicating the ~130 lines of shared setup (NuGet auth, .NET SDK, LocalDB install/start, build, test, coverage publish) across the 2 jobs. It includes an optional `initDbSchema` parameter for jobs that need explicit schema creation.

## How was this tested?

Run against the pipeline.

### DIFF
--- a/src/Config/ConfigFileWatcher.cs
+++ b/src/Config/ConfigFileWatcher.cs
@@ -20,8 +20,10 @@ namespace Azure.DataApiBuilder.Config;
 /// <seealso cref="https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.onchanged#remarks"/>
 /// <seealso cref="https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.notifyfilter"/>
 /// <seealso cref="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens#:~:text=exponential%20back%2Doff.-,Utilities/Utilities.cs%3A,-C%23"/>
-public class ConfigFileWatcher
+public class ConfigFileWatcher : IDisposable
 {
+    private bool _disposed;
+
     /// <summary>
     /// Watches a specific file for modifications and alerts
     /// this class when a change is detected.
@@ -118,6 +120,27 @@ public class ConfigFileWatcher
             // Need to remove the dependencies in startup on the RuntimeConfigProvider
             // before we can have an ILogger here.
             Console.WriteLine("Unable to hot reload configuration file due to " + ex.Message);
+        }
+    }
+
+    /// <summary>
+    /// Disposes the file watcher and unsubscribes from events to release
+    /// file handles and prevent further file change notifications.
+    /// </summary>
+    public void Dispose()
+    {
+        if (_disposed)
+        {
+            return;
+        }
+
+        _disposed = true;
+
+        if (_fileWatcher is not null)
+        {
+            _fileWatcher.EnableRaisingEvents = false;
+            _fileWatcher.Changed -= OnConfigFileChange;
+            _fileWatcher.Dispose();
         }
     }
 }
--- a/src/Config/FileSystemRuntimeConfigLoader.cs
+++ b/src/Config/FileSystemRuntimeConfigLoader.cs
@@ -30,8 +30,9 @@ namespace Azure.DataApiBuilder.Config;
 /// which allows for mocking of the file system in tests, providing a way to run the test
 /// in isolation of other tests or the actual file system.
 /// </remarks>
-public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader
+public class FileSystemRuntimeConfigLoader : RuntimeConfigLoader, IDisposable
 {
+    private bool _disposed;
     /// <summary>
     /// This stores either the default config name e.g. dab-config.json
     /// or user provided config file which could be a relative file path,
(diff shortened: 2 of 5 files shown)


.pipelines/templates/mssql-test-steps.yml
```diff
@@ -0,0 +1,161 @@
...
+    type: string
+    default: ''
+  - name: initDbSchema
+    type: boolean
+    default: false
+
+steps:
+- task: CmdLine@2
+  displayName: 'Set flag to publish received files when previous step fails'
+  condition: failed()
```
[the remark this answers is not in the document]
aaronburtle: this was a pre existing issue, that is fixed now.

.pipelines/templates/mssql-test-steps.yml
```diff
@@ -0,0 +1,161 @@
...
+
+parameters:
+  - name: testFilter
+    type: string
+  - name: testDisplayName
+    type: string
+    default: 'MsSql Integration Tests'
+  - name: testTimeout
+    type: number
+    default: 0
```
Aniruddh25, line 15: What does default 0 time out mean?

.pipelines/mssql-pipelines.yml
```diff
@@ -167,117 +180,32 @@ jobs:
...
-      command: test
-      arguments: '--filter "TestCategory=MsSql&FullyQualifiedName!~ConfigurationHotReloadTests" --no-build --configuration $(buildConfiguration) --collect "XPlat Code coverage"'
-      projects: '**/*Tests/*.csproj'
-
-
-  - task: DotNetCoreCLI@2
-    displayName: 'Hot-Reload Tests'
-    inputs:
-      command: test
-      arguments: '--filter "TestCategory=MsSql&FullyQualifiedName~ConfigurationHotReloadTests" --no-build --configuration $(buildConfiguration) --collect "XPlat Code coverage" --logger "console;verbosity=detailed"'
```
Aniruddh25, line 257: Dont we still need to have HotReload tests as a separate task?

.pipelines/templates/mssql-test-steps.yml
```diff
@@ -0,0 +1,161 @@
...
+  displayName: '${{ parameters.testDisplayName }}'
+  inputs:
+    command: test
+    arguments: '--filter "${{ parameters.testFilter }}" --no-build --configuration $(buildConfiguration) --collect "XPlat Code coverage" ${{ parameters.additionalTestArgs }}'
+    projects: '**/*Tests/*.csproj'
+  ${{ if ne(parameters.testTimeout, 0) }}:
+    timeoutInMinutes: ${{ parameters.testTimeout }}
+
+- task: CmdLine@2
+  displayName: 'Set flag to publish received files when tests fail'
```
Aniruddh25, line 134: We should also set this flag at the beginning of the template to maintain backwards compatibility if the previous step fails.

.pipelines/templates/mssql-test-steps.yml
aaronburtle: Default 0 leaves the timeout at the job level. This section gives us the option to set a task level timeout if we want, but currently this doesn't change anything.

.pipelines/mssql-pipelines.yml
aaronburtle: The changes made to eliminate the race conditions, and with mstest running these tests sequentially, it should now be safe to keep the hotreloads as part of this same task. If we run into failures again they can be split into a separate task, but it should be safe.

.pipelines/templates/mssql-test-steps.yml
aaronburtle: This was actually dead code on main so far as I can tell. When this is included as the first step, it checks if Agent.JobStatus is Failed, which will never happen on the first step because there is nothing that could have failed yet.

https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#failed

aaronburtle:
> Do we need to use a different tag for the TestCategory other than `MSSQL` for the different tests?

MSSQL is still OK, this indicates that the test relies on an MSSQL backend being setup, and will be easiest to maintain.

Issue #3209, opened 2026-03-10
Code duplication in Upsert and Insert operations


aaronburtle:
We recently modified the validation logic to allow for keyless PUT/PATCH in the case that the keys which are missing are auto-generated, and this has introduced some code duplication between the upsert and insert operations which can be factored out.

PR #3287, opened 2026-03-16
Eliminate code duplication between Upsert and Insert operations 

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3209

## What is this change?

Allow Upserts coming from REST requests to fall through to the Insert logic when auto-generated primary keys are missing. This eliminates code duplication between these two ops in the `SqlMutationEngine`

Add a test to cover the case when we have a PUT/PATCH with keys provided in the request body.

## How was this tested?

New test was added to ensure complete coverage over the scenarios introduced with the keyless PUT/PATCH change.

## Sample Request(s)

PUT that updates existing row (expects 200 OK):

```
PUT https://localhost:5001/api/magazine
Content-Type: application/json

{
    "id": 1,
    "title": "Updated Vogue",
    "issue_number": 9999
}
```

PUT that inserts new row (expects 201 Created):

```
PUT https://localhost:5001/api/magazine
Content-Type: application/json

{
    "id": 5001,
    "title": "Brand New Magazine",
    "issue_number": 42
}
```

PATCH that updates existing row (expects 200 OK):

```
PATCH https://localhost:5001/api/magazine
Content-Type: application/json

{
    "id": 1,
    "title": "Updated Vogue"
}
```

PATCH that inserts new row (expects 201 Created):

```
PATCH https://localhost:5001/api/magazine
Content-Type: application/json

{
    "id": 5001,
    "title": "Brand New Magazine"
}
```

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
@@ -88,8 +88,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                         ""orderby"": {
                             ""type"": ""string"",
                             ""enum"": [""asc"", ""desc""],
-                            ""description"": ""Sort grouped results by the aggregated value. Requires groupby."",
-                            ""default"": ""desc""
+                            ""description"": ""Sort direction for grouped results by the aggregated value. Only applies when groupby is provided; ignored otherwise.""
                         },
                         ""having"": {
                             ""type"": ""object"",
@@ -394,23 +393,10 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
             // Parse filter
             string? filter = root.TryGetProperty("filter", out JsonElement filterElement) ? filterElement.GetString() : null;
 
-            // Parse orderby
+            // Parse orderby (validation deferred until after groupby is known;
+            // if groupby is absent, orderby is silently ignored per #3279)
             bool userProvidedOrderby = root.TryGetProperty("orderby", out JsonElement orderbyElement) && !string.IsNullOrWhiteSpace(orderbyElement.GetString());
             string orderby = "desc";
-            if (userProvidedOrderby)
-            {
-                string normalizedOrderby = (orderbyElement.GetString() ?? string.Empty).Trim().ToLowerInvariant();
-                if (normalizedOrderby != "asc" && normalizedOrderby != "desc")
-                {
-                    return McpResponseBuilder.BuildErrorResult(
-                        toolName,
-                        "InvalidArguments",
-                        $"Argument 'orderby' must be either 'asc' or 'desc' when provided. Got: '{orderbyElement.GetString()}'.",
-                        logger);
-                }
-
-                orderby = normalizedOrderby;
-            }
 
             // Parse first
             int? first = null;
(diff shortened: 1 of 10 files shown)


src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -827,15 +804,18 @@ await PerformMutationOperation(
...
 
-                        if (context.OperationType is EntityActionOperation.Insert)
+                        if (effectiveOperationType is EntityActionOperation.Insert)
                         {
                             // Location Header is made up of the Base URL of the request and the primary key of the item created.
                             // For POST requests, the primary key info would not be available in the URL and needs to be appended. So, the primary key of the newly created item
                             // which is stored in the primaryKeyRoute is used to construct the Location Header.
+                            // Note: context.OperationType (not effectiveOperationType) is passed intentionally so that
+                            // ConstructCreatedResultResponse can distinguish a keyless upsert from a true Insert when
+                            // deciding whether to populate the Location header.
```
Aniruddh25: are we not populating the Location Header in the case of KeyLess upsert? Why not? 
How will the client get the location of the newly created PK?

Aniruddh25 (changes requested):
I think LocationHeader needs to be populated.

src/Core/Resolvers/SqlMutationEngine.cs
aaronburtle: Good catch, originally Upserts always had the PK in the url so populating location header information wasn't needed, now that upserts can fall through to an insert operation with auto-gen keys, that information needs to be returned in the header so the client knows the location. This is now fixed by using the effective operation type like we did to remove the code duplication.

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -542,86 +541,59 @@ await queryExecutor.ExecuteQueryAsync(
...
 
+                    // When an upsert still has no primary key values after checking the body,
+                    // it degenerates to a pure INSERT. Fall through to the shared insert/update
+                    // handling so the mutation engine generates a correct INSERT statement instead
+                    // of an UPDATE with an empty WHERE clause (WHERE 1 = 1) that would match every row.
+                    bool isKeylessUpsert = (context.OperationType is EntityActionOperation.Upsert || context.OperationType is EntityActionOperation.UpsertIncremental)
+                        && context.PrimaryKeyValuePairs.Count == 0;
+
+                    if (!isKeylessUpsert && (context.OperationType is EntityActionOperation.Upsert || context.OperationType is EntityActionOperation.UpsertIncremental))
+                    {
```
RubenCerna2079, line 596: I don't understand the purpose of the `isKeylessUpsert` variable, since in the `if` statement it needs to be false but the context.OperationType that is found inside of it needs to be true. Seems counterintuitive because of that.

RubenCerna2079: I think I see what it is being used on but this is not the correct place to set it on, I think it would be better to use it on the if statement in line 550-551

aaronburtle: this won't work on line 550-551 because `isKeylessUpsert` is only able to be computer after the body to primary key promotion code runs, that block mutates `PrimaryKeyValuesPairs.Count`. Though I agree that this if statement can be a bit confusing, but basically what it is saying is that this is an upsert AND it has keys.

PR #3297, opened 2026-03-18
Update autoentities when hot-reload succeeds

RubenCerna2079:
## Why make this change?
- #3068 
- #3304
Currently the inclusion of the autoentities feature causes hot-reload to fail when it should succeed.

## What is this change?
Now the autoentities are updated when the hot-reload succeeds by ensuring that the generated entities from the autoentities property are removed if the initialization to connect to a database is done specifically for validation purposes. Since currently, hot-reload first validates before running the actual initialization, which would cause the autoentities to fail by trying to add the generated entities that already existed.
- `Entity.cs` & `MsSqlMetadataProvider.cs`: Added new parameter that shows if an entity was generated through the autoentities property or just a regular entity.
- `RuntimeConfigProvider.cs`, `RuntimeConfig.cs` & `SqlMetadataProvider.cs`: Added function that removes the generated autoentities from the RuntimeConfig object and also removes the relation between the autoentity with the data source.
- `ConfigurationHotReloadTests.cs`: Added test to ensure autoentities works with hot-reload.
- `dab.draft.schema.json`: Fixed small bug that caused validation to fail when using the `autoentities.mcp.dml-tools`. If the validation fails it also causes hot-reload to fail.

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests

### DIFF
--- a/src/Core/Configurations/RuntimeConfigValidator.cs
+++ b/src/Core/Configurations/RuntimeConfigValidator.cs
@@ -971,7 +971,6 @@ public class RuntimeConfigValidator : IConfigValidator
         foreach ((string entityName, Entity entity) in runtimeConfig.Entities)
         {
             HashSet<EntityActionOperation> totalSupportedOperationsFromAllRoles = new();
-
             foreach (EntityPermission permissionSetting in entity.Permissions)
             {
                 string roleName = permissionSetting.Role;


src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -426,4 +426,23 @@ public void AddMergedEntitiesToConfig(Dictionary<string, Entity> newEntities)
...
+        Dictionary<string, Entity> entities = new(_configLoader.RuntimeConfig!.Entities);
+        foreach ((string name, Entity entity) in entities)
+        {
+            if (entity.IsAutoentity)
+            {
+                entities.Remove(name);
+                _configLoader.RuntimeConfig!.RemoveGeneratedAutoentityNameFromDataSourceName(name);
+            }
+        }
+
```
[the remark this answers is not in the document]
RubenCerna2079: It does not throw any error when we remove the autoentities from the collection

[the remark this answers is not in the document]
souvikghosh04: I think Copilot has a valid point. we should not modify the same collection in the loop. the suggestion from Copilot is a fail-safe approach. collect the keys to remove first and remove them separately.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -971,6 +971,16 @@ public void ValidatePermissionsInConfig(RuntimeConfig runtimeConfig)
         foreach ((string entityName, Entity entity) in runtimeConfig.Entities)
         {
             HashSet<EntityActionOperation> totalSupportedOperationsFromAllRoles = new();
+
+            if (entity.Permissions.Length == 0)
+            {
+                HandleOrRecordException(new DataApiBuilderException(
+                        message: $"Entity: {entityName} has no permissions defined.",
+                        statusCode: HttpStatusCode.ServiceUnavailable,
```
souvikghosh04: should we let DAB to be entirely unavailble for a single entity having no permissions or let it pass for others and log in the missing permission entities as warnings/errors and have DAB available?

RubenCerna2079: I just saw how DAB actually reacts to missing permissions and it just doesn't allow the user to access them at all, will delete section this to keep the same functionality that was before

src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -180,6 +181,30 @@ private static void GenerateConfigFile(
...
+                          ""include"": [ ""%book%"" ],
+                          ""name"": """ + autoentityName + @"""
+                        },
+                        ""template"": {
+                          ""rest"": {
+                              ""enabled"": true
+                            }
+                          }
+                        },
+                        ""permissions"": [
```
aaronburtle, line 196: Looks like `permissions` is at the same level as `BooksAutoentities` here, should be a child property.

RubenCerna2079: Thanks for catching that!

PR #3299, opened 2026-03-19
Fix logs for autoentities

### LINKED ISSUE #3263: [Bug]: auto-entities logging is extremely noisy, should be `debug`
## What?

`Debug` oriented logging should be `debug` type and not `info`. 
 - This allows for reasonable filtering.
 - Current output is insanely noisy.

### Rules of thumb
 - Information logs record events a human operator should care about in normal production runs. 
 - Debug logs record details useful only when diagnosing behavior, like variable/settings values..

Image

### LINKED ISSUE #3264: [Bug]: auto entity error messages are inconsistent
## What?

Inconsistent way to refer to autoentity definition name. 

 - Can be confusing or disorienting to the developer or tool reading the logs.
 - Makes it hard to be consistent with the documentation. Let's just standardize.

### Configuration
```json
  "autoentities": {
    "nixon": {
      "patterns": {
         [snip]
```

## Messages

> [!IMPORTANT]
> Scan through all the error messages, I only found these two

### Rules of thumb:

 - Use the correct property name `autoentity` -> `autoentities`
 - Call it a *definition* and put it in single quotes: `autoentities definition 'nixon'`.

### When there is a name clash.

Current

> Cannot create new entity from autoentity pattern with definition-name 'nixon'.

Better

> Cannot create new entity from autoentities definition 'nixon'.

### When no entities are found.

Current

Image

> No new entities were generated from the autoentity nixon defined in the configuration.

Better

> No new entities were generated from the autoentities definition 'nixon'.

### LINKED ISSUE #3265: [Bug]: autoentities logging in `dab validate` is incorrect.
## What

Output messages are inconsistent or unclear based on current values. 

> [!NOTE]
> These are clearly `debug` and not `information`.

### Configuration

```json
  "autoentities": {
    "nixon": {
      "patterns": {
        "include": [
          "%.%"
        ]
      },
   [snip]
```

### Output

```
Information: Autoentities include pattern: %.%
Information: Autoentities exclude pattern:
Information: Autoentities name pattern: {object}
```

 - Use the correct property name `Autoentities` -> `autoentities`.
 - Use the definition name `autoentities definition 'nixon' include pattern: %.%`
 - Handle null/empty/default, for example:
    - `autoentities definition 'nixon' include pattern: [default: all]`
    - `autoentities definition 'nixon' exclude pattern: [default: none]`
    - `autoentities definition 'nixon' name pattern: [default: {object}]`

Image

RubenCerna2079:
## Why make this change?
- Closes #3263
- Closes #3264 
- Closes #3265
Currently some of the logs for autoentities are not completely clear or they are too noisy. Which is why the messages need to be changed.

## What is this change?
This changes the log messages so they are more clear by adding `' '` to the definition names from the autoentities, and stating that they are definition names beforehand. We also change the name from `Autoentities` to `autoentities` in the logs so that they are all consistent.
Lastly we also change some of the logs from `Information` to `Debug`.

## How was this tested?
Tested locally since all the changes are related to the messages written in the logs.

### DIFF
--- a/src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
+++ b/src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
@@ -349,7 +349,8 @@ namespace Azure.DataApiBuilder.Core.Services
                         Health: autoentity.Template.Health,
                         Fields: null,
                         Relationships: null,
-                        Mappings: new());
+                        Mappings: new(),
+                        IsAutoentity: true);
 
                     // Add the generated entity to the linking entities dictionary.
                     // This allows the entity to be processed later during metadata population.
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -486,7 +486,40 @@
             "ttl-seconds": {
               "type": "integer",
               "description": "Time to live in seconds",
-              "default": 5
+              "default": 5,
+              "minimum": 1
+            },
+            "level-2": {
+              "type": "object",
+              "description": "Configuration for the level 2 (distributed) cache and backplane.",
+              "additionalProperties": false,
+              "properties": {
+                "enabled": {
+                  "$ref": "#/$defs/boolean-or-string",
+                  "description": "Enable or disable the level 2 distributed cache.",
+                  "default": false
+                },
+                "provider": {
+                  "type": "string",
+                  "description": "The provider for the L2 cache. Currently only 'redis' is supported."
+                },
+                "connection-string": {
+                  "type": "string",
+                  "description": "The connection string for the level 2 cache provider."
+                },
+                "partition": {
+                  "type": "string",
+                  "description": "The prefix to use for cache keys in level 2 and backplane, useful in a shared environment to avoid collisions."
+                }
+              },
+              "if": {
+                "properties": {
+                  "enabled": { "const": true }
+                }
+              },
+              "then": {
+                "required": ["connection-string"]
+              }
             }
           }
         },
@@ -819,15 +852,23 @@
               "additionalProperties": false,
               "properties": {
                 "mcp": {
-                  "type": "object",
-                  "description": "MCP endpoint configuration",
-                  "additionalProperties": false,
-                  "properties": {
-                    "dml-tools": {
+                  "oneOf": [
+                    {
                       "$ref": "#/$defs/boolean-or-string",
-                      "description": "Enable/disable all DML tools with default settings."
+                      "description": "Boolean shorthand: true enables dml-tools only (custom-tool remains false), false disables all MCP functionality."
+                    },
+                    {
+                      "type": "object",
+                      "description": "MCP endpoint configuration",
+                      "additionalProperties": false,
+                      "properties": {
+                        "dml-tools": {
+                          "$ref": "#/$defs/boolean-or-string",
+                          "description": "Enable/disable all DML tools with default settings."
+                        }
+                      }
                     }
-                  }
+                  ]
                 },
                 "rest": {
                   "type": "object",
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
@@ -45,13 +45,24 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
         {
             Name = "aggregate_records",
             Description = "Computes aggregations (count, avg, sum, min, max) on entity data. "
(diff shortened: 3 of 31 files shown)


src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -375,14 +375,14 @@ protected override async Task GenerateAutoentitiesIntoEntities(IReadOnlyDictiona
 
                 if (addedEntities == 0)
                 {
-                    _logger.LogWarning("No new entities were generated from the autoentity {autoentityName} defined in the configuration.", autoentityName);
+                    _logger.LogWarning("No new entities were generated from the autoentities definition '{autoentityName}'.", autoentityName);
                 }
             }
 
             _runtimeConfigProvider.AddMergedEntitiesToConfig(entities);
         }
```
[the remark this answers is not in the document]
RubenCerna2079: Added the missing parameter to the test

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -316,7 +316,7 @@ protected override async Task GenerateAutoentitiesIntoEntities(IReadOnlyDictiona
                     if (resultObject is null)
                     {
                         throw new DataApiBuilderException(
-                            message: $"Cannot create new entity from autoentity pattern due to an internal error.",
+                            message: $"Cannot create new entity from autoentities definition '{autoentityName}' due to an internal error.",
```
Aniruddh25, line 319: As per the bug https://github.com/Azure/data-api-builder/issues/3264 , the message should be:

Cannot create new entity from autoentities member 'autoentityName' due to an internal error

whereas the other bug says it should be `definition`.
Clarify that there is a discrepancy in expectation. I do think `definition` is the right term to use.

Aniruddh25 (approved):

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
RubenCerna2079: Already discussed with Jerry and said it is fine as long as all the messages are standardized, which in this case it is they use `autoentities definition`

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -329,7 +329,7 @@ protected override async Task GenerateAutoentitiesIntoEntities(IReadOnlyDictiona
 
                     if (string.IsNullOrWhiteSpace(entityName) || string.IsNullOrWhiteSpace(objectName) || string.IsNullOrWhiteSpace(schemaName))
                     {
-                        _logger.LogError("Skipping autoentity generation: entity_name or object is null or empty for autoentity pattern '{AutoentityName}'.", autoentityName);
+                        _logger.LogError("Skipping autoentity generation: 'entity_name', 'object', or 'schema' is null or empty for autoentities definition '{AutoentityName}'.", autoentityName);
```
aaronburtle: nit: `{AutoentityName}` here is PascalCase, but below in the other logs you use `{autoentityName}` which is camelCase. Should just use one style everywhere, if we do this elsewhere can align.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -395,10 +395,10 @@ protected override async Task GenerateAutoentitiesIntoEntities(IReadOnlyDictiona
...
             };
 
-            _logger.LogInformation("Query for Autoentities is being executed with the following parameters.");
-            _logger.LogInformation($"Autoentities include pattern: {include}");
-            _logger.LogInformation($"Autoentities exclude pattern: {exclude}");
-            _logger.LogInformation($"Autoentities name pattern: {namePattern}");
+            _logger.LogDebug("Query for autoentities is being executed with the following parameters.");
+            _logger.LogDebug($"The autoentities definition '{autoentityName}' include pattern: {include}");
+            _logger.LogDebug($"The autoentities definition '{autoentityName}' exclude pattern: {exclude}");
+            _logger.LogDebug($"The autoentities definition '{autoentityName}' name pattern: {namePattern}");
```
aaronburtle: I think co-pilot was right to call out the string interpolations `$"..."`, template placeholders produce properties that can be queried via Application Insights, etc, and youre already using placeholders in the other log calls in this PR. Should align.

RubenCerna2079: I wasn't aware that the placeholders produced properties that can be queried, that might be something that could be brought to the team so that we can standardize this.

PR #3303, opened 2026-03-23
[MCP] Fix for inconsistent MCP enabled check

### LINKED ISSUE #3284: [Bug]: `mcp.enabled` defaults are not working
## What?

When `mcp.enabled` is NOT in the config, but we use the default value instead, MCP does **not** work.

> [!IMPORTANT]
> Config properties with default values should not be required to be in the config to work properly.

### `/mcp` output:

```json
{
  "error": {
    "code": "GlobalMcpEndpointDisabled", // this is the error
    "message": "Route mcp was not found.",
    "status": 404
  }
}
```

### `/health` output:

```json
{
  "status": "Healthy",
  "version": "2.0.0",
  "app-name": "dab_oss_2.0.0",
  "timestamp": "2026-03-15T21:04:32.5518362Z",
  "currentRole": "anonymous",
  "configuration": {
    "rest": true,
    "graphql": true,
    "mcp": true, // it says true but it does not work!
    "caching": false,
    "telemetry": false,
    "mode": "Development"
  },
  "checks": [ [snip] ]
}
```
## Workaround

Include `mcp.enabled=true` in the config file.

souvikghosh04:
## Why make this change?
- Closes #3284

MCP endpoint returns 404 when the mcp config block is omitted, despite mcp.enabled defaulting to true across schema, CLI, and model layers.

## What is this change?

`McpEndpointRouteBuilderExtensions.TryGetMcpOptions`: Falls back to new `McpRuntimeOptions()` when `Runtime.Mcp` is `null`, so the `/mcp` endpoint is registered by default.
`RestService.GetRouteAfterPathBase`: Guards the `GlobalMcpEndpointDisabled` exception with an `IsMcpEnabled` check so MCP path requests are not unconditionally rejected.

## How was this tested?
- [ ] Integration Tests
- [x] Unit Tests

## Sample Request(s)

/mcp
/health

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
+++ b/src/Azure.DataApiBuilder.Mcp/BuiltInTools/AggregateRecordsTool.cs
@@ -88,8 +88,7 @@ namespace Azure.DataApiBuilder.Mcp.BuiltInTools
                         ""orderby"": {
                             ""type"": ""string"",
                             ""enum"": [""asc"", ""desc""],
-                            ""description"": ""Sort grouped results by the aggregated value. Requires groupby."",
-                            ""default"": ""desc""
+                            ""description"": ""Sort direction for grouped results by the aggregated value. Only applies when groupby is provided; ignored otherwise.""
                         },
                         ""having"": {
                             ""type"": ""object"",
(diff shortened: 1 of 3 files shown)


aaronburtle:
`Cli.ConfigGenerator` seems to have a similar bug in `IsConfigValid`

`bool mcpEnabled = config.Runtime?.Mcp?.Enabled == true;
`

souvikghosh04:
> `Cli.ConfigGenerator` seems to have a similar bug in `IsConfigValid`
> 
> `bool mcpEnabled = config.Runtime?.Mcp?.Enabled == true; `

Added a fix.

src/Service.Tests/Configuration/McpRuntimeOptionsSerializationTests.cs
```diff
@@ -191,6 +191,95 @@ public void TestBackwardCompatibilityDeserializationWithoutDescriptionField()
             Assert.IsNull(deserializedConfig.Runtime.Mcp.Description, "Description should be null when not present in JSON");
         }
 
+        /// <summary>
+        /// When the mcp and runtime config blocks are omitted entirely, IsMcpEnabled should default to true.
+        /// This validates the fix for GitHub issue #3284.
```
RubenCerna2079: Should we keep this part of the comment?

Issue #3261, opened 2026-03-14
[Bug]: When OpenAPI is filtered by missing role, error message is unclear.


JerryNixon:
## What?

The new role filter feature works, but when filtering by a role that is not present in the configuration the resulting `404` is difficult to debug because it has not informational message.

### Current output

```json
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.5",
  "title": "Not Found", 
  "status": 404,
  "traceId": "00-5d730e51839724219df7c2951c9990e3-5288a74e511a1dcb-00"
}
```

This generic `404` error is not communicative. Introduce a message.

### New output

```json
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.5",
  "message": "Role `Jerry` is not present in the configuration.",  // new
  "title": "Not Found", 
  "status": 404,
  "traceId": "00-5d730e51839724219df7c2951c9990e3-5288a74e511a1dcb-00"
}
```

<img width="727" height="197" alt="Image" src="https://github.com/user-attachments/assets/a014e11a-0bd3-4587-a986-169a7a29fcfe" />

Aniruddh25:
PR https://github.com/Azure/data-api-builder/pull/2998 made this change. However, the error message was probably not clear in the specification which is why copilot did not implement it accurately.

aaronburtle:
## RCA
In the RestController class (around line 255), when the /openapi/{role} route is matched, it calls TryGetDocumentForRole on the OpenApiDocumentor. That method checks whether the role exists in any entity's permissions. If the role is not found, it returns false and the controller simply returns NotFound() with no message body.

The relevant code in the RestController is straightforward: the if block on line 255 checks TryGetDocumentForRole, and on line 261 the else path is just "return NotFound();" with no additional context. ASP.NET Core's NotFound() produces the generic RFC 9110 problem details response with just a status code and title.

This is a missing feature in the original implementation rather than a regression. The TryGetDocumentForRole method already knows whether the role does not exist vs the document generation failed, but that information is not communicated back to the controller or to the caller.


## Version
**v2.0.0-rc**
The /openapi/{role} feature was introduced in PR #2998 ("Filter OpenAPI REST methods and fields based on role permissions"), first included in v2.0.0-rc.

aaronburtle:
As per the RFC for Problem Details (https://www.rfc-editor.org/rfc/rfc9457#section-3.1) we should use the `detail` field rather than `message`

PR #3305, opened 2026-03-23
Add descriptive message when OpenAPI role filter returns 404

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3261

## What is this change?

In the `RestController`, add descriptive messaging for `NotFound()` when role is missing. Add a test case to cover this new behavior.

## How was this tested?

Test case added.

Manually verified the endpoint provided the correct response.

image

## Sample Request(s)

Can manually verify by going to this endpoint without a "foo" role defined.

`https://localhost:5001/rest/openapi/foo`

### DIFF
--- a/src/Config/ObjectModel/RuntimeConfig.cs
+++ b/src/Config/ObjectModel/RuntimeConfig.cs
@@ -6,6 +6,7 @@ using System.IO.Abstractions;
 using System.Net;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using Azure.DataApiBuilder.Config.Converters;
 using Azure.DataApiBuilder.Service.Exceptions;
 using Microsoft.Extensions.Logging;
 
@@ -355,8 +356,9 @@ public record RuntimeConfig
 
             foreach (string dataSourceFile in DataSourceFiles.SourceFiles)
             {
-                // Use default replacement settings for environment variable replacement
-                DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true);
+                // Use Ignore mode so missing env vars are left as literal @env() strings,
+                // consistent with how the parent config is loaded in TryLoadKnownConfig.
+                DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true, envFailureMode: EnvironmentVariableReplacementFailureMode.Ignore);
 
                 if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config, replacementSettings: replacementSettings))
                 {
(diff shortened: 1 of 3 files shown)


src/Service/Controllers/RestController.cs
```diff
@@ -257,7 +257,16 @@ private async Task<IActionResult> HandleOperation(
                         return Content(roleDocument, MediaTypeNames.Application.Json);
                     }
 
-                    return NotFound();
+                    ProblemDetails problemDetails = new()
```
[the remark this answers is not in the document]
aaronburtle: stale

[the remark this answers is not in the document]
aaronburtle: added test case

src/Service/Controllers/RestController.cs
Aniruddh25: return the reason even in this case where we reject path separators in role, you can modify the same test with the rolename as a DataRow.

aaronburtle: Added message, and updated test.

src/Service.Tests/OpenApiDocumentor/MissingRoleNotFoundTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            using HttpClient client = server.CreateClient();
+
+            string missingRole = roleName;
+            HttpResponseMessage response = await client.GetAsync($"/api/openapi/{Uri.EscapeDataString(missingRole)}");
+
+            Assert.AreEqual(HttpStatusCode.NotFound, response.StatusCode, "Expected 404 for a role not in the configuration.");
+
+            string responseBody = await response.Content.ReadAsStringAsync();
+            JsonDocument doc = JsonDocument.Parse(responseBody);
+            JsonElement root = doc.RootElement;
```
souvikghosh04: JsonDoc should be idsposed safely

src/Service.Tests/OpenApiDocumentor/MissingRoleNotFoundTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            RuntimeConfig testConfig = config with
+            {
+                Runtime = config.Runtime with
+                {
+                    Host = config.Runtime?.Host with { Mode = HostMode.Development }
+                },
+                Entities = new RuntimeEntities(new Dictionary<string, Entity> { { "book", entity } })
+            };
+
+            File.WriteAllText(CONFIG_FILE, testConfig.ToJson());
```
souvikghosh04, line 63: should this file created should be deleted safely?

src/Service.Tests/OpenApiDocumentor/MissingRoleNotFoundTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            string[] args = new[] { $"--ConfigFileName={CONFIG_FILE}" };
+
+            using TestServer server = new(Program.CreateWebHostBuilder(args));
+            using HttpClient client = server.CreateClient();
+
+            string missingRole = roleName;
```
souvikghosh04: why use a separate variable? just use `roleName` may be?

Aniruddh25 (approved):

souvikghosh04 (approved):

PR #3306, opened 2026-03-23
Update config validation logic for entities

### LINKED ISSUE #3267: [Bug]: `dab validate` should at least `warn` when zero entities are in the config.
## What?

`dab validate` returns "Config is valid". when there are zero entities.

 - entities can be added by `autoentities`
 - entities can be added by `entities`

But when `autoentities` is not configured and `entities` is empty, `validate` should `warn`:

> [WARN] No entities are defined in this configuration.

> [!IMPORTANT]
> This is NOT an error because it could be the top-level configuration in a multi-config environment. 

Image

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3267

## What is this change?

Alters the validation logic in the following way.

Is top-level config with data-source-files? (we call this a `Root` config file)
├── YES
│   ├── Has datasource? → ValidateEntityPresence (same rules as non-root)
│   ├── No datasource but has entities/autoentities? → ERROR
│   └── No datasource, no entities → VALID (children provide everything)
│   └── For each child → ValidateNonRootConfig(child, filename)
│
└── NO (standalone or child config)
    ├── No datasource? → ERROR: "data source is required"
    └── Has datasource → ValidateEntityPresence

Note: A top-level config file without any children data-source files is NOT considered a root. And an intermediary config file, ie: is a child, that also has child configs is NOT a root. Only a top-level config with children configs is a Root.

#### ValidateEntityPresence
Count resolved autoentities from AutoentityResolutionCounts
total = manual entities + resolved autoentities

total == 0? → ERROR: "No entities found"
total > 0 but autoentities discovered nothing? → WARN: "Autoentities configured but none discovered"

No double messaging. If total is 0, only the error is recorded, not the warning.

### Truth table — top-level config

Variables (`1` = present / non-empty, `0` = absent / empty):
- **DSF** — `data-source-files` present
- **DS** — `data-source` present
- **E** — manual `entities` count > 0
- **AE** — `autoentities` count > 0 (presence, *not* resolved count)

Path is determined by `IsRootConfig = (DSF == 1) && !IsChildConfig`.

| # | DSF | DS | E | AE | AE resolved | Path | Expected | Test |
|---|:---:|:--:|:-:|:--:|:-----------:|------|----------|------|
| 1 | 0 | 0 | 0 | 0 | — | Non-root | **Error**: "data source is required" | `TestNonRootWithNoDataSourceProducesError` |
| 2 | 0 | 0 | 0 | 1 | — | Non-root | **Error**: "data source is required" | _covered by #1 — DS check fires first_ |
| 3 | 0 | 0 | 1 | 0 | — | Non-root | **Error**: "data source is required" | _covered by #1_ |
| 4 | 0 | 0 | 1 | 1 | — | Non-root | **Error**: "data source is required" | _covered by #1_ |
| 5 | 0 | 1 | 0 | 0 | — | Non-root | **Error**: "No entities found" | `TestNonRootWithDataSourceAndNoEntitiesProducesError` |
| 6a | 0 | 1 | 0 | 1 | 0 | Non-root | **Error**: "No entities found" | `TestNonRootWithDataSourceAndAutoentitiesResolvingZeroProducesError` |
| 6b | 0 | 1 | 0 | 1 | >0 | Non-root | **Valid** | `TestNonRootWithDataSourceAndAutoentitiesResolvingEntitiesIsValid` |
| 7 | 0 | 1 | 1 | 0 | — | Non-root | **Valid** | `TestNonRootWithDataSourceAndEntitiesIsValid` |
| 8a | 0 | 1 | 1 | 1 | 0 | Non-root | **Valid** + **Warn** | `TestNonRootWithEntitiesAndAutoentitiesResolvingZeroLogsWarning` |
| 8b | 0 | 1 | 1 | 1 | >0 | Non-root | **Valid** | _covered by #7 / #6b combined_ |
| 9 | 1 | 0 | 0 | 0 | — | Root | **Valid** (children carry the load) | `TestRootWithNoDataSourceAndNoEntitiesIsValid`, `TestRootConfigWithNoDataSourceAndNoEntitiesParses` |
| 10 | 1 | 0 | 0 | 1 | — | Root | **Error**: "must not define entities or autoentities" | `TestRootWithNoDataSourceButAutoentitiesProducesError` |
| 11 | 1 | 0 | 1 | 0 | — | Root | **Error**: "must not define entities" | `TestRootWithNoDataSourceButEntitiesProducesError` |
| 12 | 1 | 0 | 1 | 1 | — | Root | **Error** | _covered by #11_ |
| 13 | 1 | 1 | 0 | 0 | — | Root (with own DS) | **Error**: "No entities found" | `TestRootWithDataSourceAndNoEntitiesProducesError` |
| 14a | 1 | 1 | 0 | 1 | 0 | Root (with own DS) | **Error**: "No entities found" | `TestRootWithDataSourceAndAutoentitiesResolvingZeroProducesError` |
| 14b | 1 | 1 | 0 | 1 | >0 | Root (with own DS) | **Valid** | `TestRootWithDataSourceAndAutoentitiesResolvingEntitiesIsValid` |
| 15 | 1 | 1 | 1 | 0 | — | Root (with own DS) | **Valid** | `TestRootWithDataSourceAndEntitiesIsValid` |
| 16a | 1 | 1 | 1 | 1 | 0 | Root (with own DS) | **Valid** + **Warn** | `TestRootWithEntitiesAndAutoentitiesResolvingZeroLogsWarning` |
| 16b | 1 | 1 | 1 | 1 | >0 | Root (with own DS) | **Valid** | _covered by #15 / #14b combined_ |

### Truth table — child config (validated when iterating `root.ChildConfigs`)

Children are always treated as non-root regardless of their own `data-source-files`.

| # | DS | E | AE | AE resolved | Expected | Test |
|---|:--:|:-:|:--:|:-----------:|----------|------|
| C1 | 0 | 0 | 0 | — | **Error** naming the child file: "data source is required" | `TestChildWithNoDataSourceProducesNamedError` |
| C2 | 0 | * | * | — | **Error** naming the child file: "data source is required" | _covered by C1_ |
| C3 | 1 | 0 | 0 | — | **Error** naming the child file: "No entities found" | `TestChildWithDataSourceAndNoEntitiesProducesNamedError` |
| C4a | 1 | 0 | 1 | 0 | **Error** naming the child file: "No entities found" | `TestChildWithDataSourceAndAutoentitiesResolvingZeroProducesNamedError` |
| C4b | 1 | 0 | 1 | >0 | **Valid** | _covered by C5 (resolved entities behave the same as manual entities)_ |
| C5 | 1 | 1 | 0 | — | **Valid** | _implicitly via `TestRootWithDataSourceAndEntitiesIsValid` setup_ |
| C6a | 1 | 1 | 1 | 0 | **Valid** + **Warn** naming the child file | `TestChildWithEntitiesAndAutoentitiesResolvingZeroLogsNamedWarning` |
| C6b | 1 | 1 | 1 | >0 | **Valid** | _covered by C5_ |

### Other scenarios

| Scenario | Expected | Test |
|----------|----------|------|
| Connection-string error gates entity validation (no entity error fires when DB unreachable) | `IsConfigValid == false` due to connection error only | `TestValidateNonRootZeroEntitiesWithInvalidConnectionString` |
| Config with no entities parses cleanly (constructor no longer throws) and `IsConfigValid` returns false without throwing | parse OK, validate fails | `TestValidateConfigWithNoEntitiesProducesCleanError` _(modified)_ |
| Root parses successfully without a data source | parse OK, `IsRootConfig == true` | `TestRootConfigWithNoDataSourceAndNoEntitiesParses` |
| Non-root with DS and no entities parses successfully | parse OK, `IsRootConfig == false` | `TestNonRootConfigWithDataSourceAndNoEntitiesParses` |
| Autoentities present but resolve to nothing — must not crash, must not double-message with "No entities found" | no crash; only "No entities found" if total = 0 | `ValidateAutoentitiesConfiguration` _(modified to `isValidateOnly: true`)_ |

New tests:

`TestRootConfigWithNoDataSourceAndNoEntitiesParses` Root config (has data-source-files) without datasource parses OK
`TestNonRootConfigWithDataSourceAndNoEntitiesParses` Non-root config with datasource + no entities parses OK (validation catches it later)
`TestNonRootWithDataSourceAndNoEntitiesProducesError` Calls ValidateDataSourceAndEntityPresence directly, error recorded
`TestNonRootWithNoDataSourceProducesError` No datasource, error with "data source is required"
`TestNonRootWithDataSourceAndEntitiesIsValid` Datasource + entities, no errors
`TestRootWithNoDataSourceAndNoEntitiesIsValid` Root with child, no own datasource, valid
`TestRootWithNoDataSourceButEntitiesProducesError` Root with entities but no datasource, error
`TestRootWithDataSourceAndEntitiesIsValid` Root with own datasource + entities, valid
`TestChildWithDataSourceAndNoEntitiesProducesNamedError` Child with no entities, error names the child file
`TestChildWithNoDataSourceProducesNamedError` Child with no datasource, error names the child file
`TestNonRootWithDataSourceAndAutoentitiesResolvingZeroProducesError` Non-root with only autoentities that resolve to 0 
`TestNonRootWithDataSourceAndAutoentitiesResolvingEntitiesIsValid` Non-root with only autoentities resolving > 0 entities
`TestNonRootWithEntitiesAndAutoentitiesResolvingZeroLogsWarning` Non-root with manual entities + autoentities resolving 0
`TestRootWithNoDataSourceButAutoentitiesProducesError` Root with no datasource but autoentities defined
`TestRootWithDataSourceAndNoEntitiesProducesError` Root with own datasource and zero entities/autoentities
`TestRootWithDataSourceAndAutoentitiesResolvingZeroProducesError` Root with own datasource and autoentities resolving 0
`TestRootWithDataSourceAndAutoentitiesResolvingEntitiesIsValid` Root with own datasource and autoentities resolving > 0 
`TestRootWithEntitiesAndAutoentitiesResolvingZeroLogsWarning` Root with own datasource, manual entities, and autoentities resolving 0
`TestChildWithDataSourceAndAutoentitiesResolvingZeroProducesNamedError` Child with autoentities-only resolving 0
`TestChildWithEntitiesAndAutoentitiesResolvingZeroLogsNamedWarning` Child with manual entities + autoentities resolving 0

Modified tests:

`TestValidateConfigWithNoEntitiesProducesCleanError`  Replaced main's version (expected parse failure) with ours: parse succeeds, IsConfigValid returns false
`ValidateAutoentitiesConfiguration`  Changed to isValidateOnly: true, asserts no crashes instead of zero errors


src/Cli/ConfigGenerator.cs
```diff
@@ -3622,5 +3628,18 @@ private static bool ValidateFields(
...
+
+        /// <summary>
+        /// Logs a warning if no entities, autoentities, or data-source-files are configured.
+        /// </summary>
+        internal static void WarnIfNoEntitiesDefined(RuntimeConfig config)
+        {
+            if (config.Entities.Count() == 0
+                && (config.Autoentities is null || config.Autoentities.Count() == 0)
+                && config.DataSourceFiles is null)
+            {
```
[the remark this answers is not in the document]
aaronburtle: valid, adressing

src/Cli/ConfigGenerator.cs
```diff
@@ -2648,15 +2648,21 @@ public static bool IsConfigValid(ValidateOptions options, FileSystemRuntimeConfi
...
+                WarnIfNoEntitiesDefined(config);
+            }
+
             // Additional validation: warn if fields are missing and MCP is enabled
             if (isValid)
             {
-                if (runtimeConfigProvider.TryGetConfig(out RuntimeConfig? config) && config is not null)
+                if (runtimeConfigProvider.TryGetConfig(out RuntimeConfig? validatedConfig) && validatedConfig is not null)
                 {
```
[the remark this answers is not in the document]
aaronburtle: valid, addressing

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -359,4 +359,29 @@ private async Task ValidatePropertyOptionsFails(ConfigureOptions options)
...
         Assert.IsFalse(result.IsValid);
     }
+
+    /// <summary>
+    /// Validates that WarnIfNoEntitiesDefined logs a warning when the config has
+    /// an empty entities collection and no autoentities or data-source-files.
+    /// This calls the extracted method directly, avoiding any DB connection.
+    /// </summary>
+    [TestMethod]
+    public void TestValidateWarnsOnZeroEntities()
```
anushakolan: Why are we mocking instead of like creating an actual config and testing?

anushakolan (approved):
LGTM, Apart from the minor comment left.

src/Cli.Tests/ValidateConfigTests.cs
aaronburtle: its because the config validation doesnt rely on DB connections, http, file i/o, etc, so we are just testing the logic. Unit testing this is the right call here because it gives us all the coverage we need without having to rely on spinning up the engine, connecting to a DB etc.

src/Cli/ConfigGenerator.cs
```diff
@@ -2648,10 +2648,13 @@ public static bool IsConfigValid(ValidateOptions options, FileSystemRuntimeConfi
 
             bool isValid = runtimeConfigValidator.TryValidateConfig(runtimeConfigFile, LoggerFactoryForCli).Result;
 
-            // Additional validation: warn if fields are missing and MCP is enabled
-            if (isValid)
+            if (runtimeConfigProvider.TryGetConfig(out RuntimeConfig? config) && config is not null)
             {
-                if (runtimeConfigProvider.TryGetConfig(out RuntimeConfig? config) && config is not null)
+                // Run config-structure warnings regardless of validation outcome.
+                WarnIfNoEntitiesDefined(config);
```
Aniruddh25: if there are 0 entities, or autoentities, it is still a valid configuration. 

But do we still need to check whether any entity has missing fields when MCP is enabled - i.e. perform the below validation? 

We should still Warn if Unauthenticated provider is used with authenticated or custom roles

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -359,4 +359,29 @@ private async Task ValidatePropertyOptionsFails(ConfigureOptions options)
...
+    {
+        Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
+
+        Mock<ILogger<ConfigGenerator>> mockLogger = new();
+        SetLoggerForCliConfigGenerator(mockLogger.Object);
+
+        ConfigGenerator.WarnIfNoEntitiesDefined(config!);
```
Aniruddh25: Instead of directly calling the WarnIfNoEntitiesDefined function, test the scenario for which the bug is identified. 
In this case, can we call `IsConfigValid` and verify the warning is printed?

Aniruddh25 (approved):
Left few comments

src/Cli/ConfigGenerator.cs
```diff
@@ -2648,10 +2648,13 @@ public static bool IsConfigValid(ValidateOptions options, FileSystemRuntimeConfi
...
+
+                // Additional validation: warn if fields are missing and MCP is enabled
+                if (isValid)
```
Aniruddh25, line 2847: Could you please clarify with @JerryNixon, if it is a config validation error or warning in case:

1) entities/autoentities both sections are completely missing - i.e. null from the config -> we error out today on this case.
2) entities/autoentities - either is present but count is 0 - no validation error - but warning emitted as per this PR.

Aniruddh25 (changes requested):

src/Cli/ConfigGenerator.cs
aaronburtle: If there are no entities, then the for loop below iterates zero times and this is a no-op, so there wont really be any validation happening; there is nothing to validate. On the other hand, if validation fails for some other reason the user wont see any "missing fields with MCP" warnings, which may be an enhancement we want to add, but shouldn't be scoped to this PR.

aaronburtle: Similar to unauthenticated provider with non-anonymous roles. Since it is inside `if(isValid)` it only fires when everything else passes. If we move it outside then it would always run, but with 0 entities, the warning can never trigger.

JerryNixon:
## Config rules

> Level 1. Root config (top level (has a child))

###    Root rules:
    - if data-source-files is empty array, this is a non-Root.
    - if data-source-files is NOT empty array, 
        data-source+entities is optional. 

> Level 1.1. Non-Root A (including grandchildren)

###  Non-Root rules:
    - must have data source and must have entities (or found entities if auto-entities is configured)
    - if has children, same rules as above, must have ds and entities

## When to warn

```csharp
if (IsRoot) // at this point we know it has children
{
    if (has-data-source) // if you have a data source you must have entities
    {
        ValidateItHasEntities();
    }
    else if (!has-data-source && (has-any-entities || has-any-auto-entities-definitions))
    {
        Error("Entities found but no data source defined.");
    }
    else
    {
        Valid();
    }
}
else if (!IsRoot) // non-root/child
{
    ValidateItHasEntities();
}

ValidateItHasEntities()
{
    foreach (var definition in auto-entities-definitions)
    {
        if (definition has no entities)
        {
            Warning($"Auto-entities definition '{definition.Name}' but no entities found.");
        }
    }

    var entityCountFromAutoEntityDefinitions = 123; // count of all autoentities

    if (entityCountFromAutoEntityDefinitions == 0
        && entityCountFromEntities == 0)
    {
        Error("Data source defined but no entities found.");
    }

    Valid();
}
```

src/Cli/ConfigGenerator.cs
aaronburtle: after conferring with @JerryNixon, we have some more sophisticated validation logic, can be seen in the comment below, will update the PR and then ask for fresh reviews.

Aniruddh25:
```
if (data-source-files not present)
{
   error if data-source missing
}

// data-source-files are present or data-source is present:

if (entities and autoentities)
{
    error when none
}
else if (only entities)
{
    error when none
}
else if (only autoentities)
{
    error when none
}
else
{
    error unless data-source absent
    (already handled)
}
```

after offline discussion

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -269,15 +277,15 @@ public async Task<bool> Initialize(
 
         IsLateConfigured = true;
 
-        if (RuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig? runtimeConfig, replacementSettings))
+        if (RuntimeConfigLoader.TryParseConfig(jsonConfig, out RuntimeConfig? runtimeConfig, out _, replacementSettings))
         {
-            _configLoader.RuntimeConfig = runtimeConfig.DataSource.DatabaseType switch
+            _configLoader.RuntimeConfig = runtimeConfig.DataSource?.DatabaseType switch
```
Aniruddh25: Does TryParseConfig ensure DataSource will not be null? 
If it does not, and if it is null, how does this switch work?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -329,6 +329,19 @@ public async Task<bool> TryValidateConfig(
         // Any exceptions caught during this process are added to the ConfigValidationExceptions list and logged at the end of this function.
         await ValidateEntitiesMetadata(runtimeConfig, loggerFactory);
 
+        // Validate entity configuration (root vs non-root rules, entity counts) after autoentity resolution.
+        // Only run when there are no connection string errors, since autoentity resolution requires DB access.
+        if (!ConfigValidationExceptions.Any(x => x.Message.StartsWith(DataApiBuilderException.CONNECTION_STRING_ERROR_MESSAGE)))
+        {
+            // Re-read the config since autoentity resolution may have added new entities.
```
Aniruddh25, line 336: Is this simply getting a reference the the updatedConfig or re-doing deserialization?

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -511,6 +524,143 @@ public async Task ValidateEntitiesMetadata(RuntimeConfig runtimeConfig, ILoggerF
...
+        }
+        else if (hasEntities || hasAutoentities)
+        {
+            // Root without a data source but with entities/autoentities — invalid.
+            HandleOrRecordException(new DataApiBuilderException(
+                message: "Entities or autoentities are defined in the root config but no data source is configured. " +
+                    "A root config without a data source must not define entities or autoentities.",
+                statusCode: HttpStatusCode.ServiceUnavailable,
+                subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
+        }
```
Aniruddh25, line 583: This is missing the case where -

there is no datasource, no datasourcefiles, and no entities/autoentities -> should be erroring out in this case.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -5776,16 +5811,20 @@ public async Task ValidateAutoentitiesConfiguration()
...
-            catch (Exception ex)
-            {
-                Assert.Fail(ex.Message);
-            }
+            RuntimeConfigValidator configValidator = new(provider, fileSystem, loggerMock.Object, isValidateOnly: true);
+
+            bool isValid = await configValidator.TryValidateConfig(CUSTOM_CONFIG, TestHelper.ProvisionLoggerFactory());
+
+            // Validation may fail because autoentity patterns don't match any tables in the test DB,
+            // but it should not throw. The important thing is that the config is structurally valid
```
Aniruddh25: the comment says "it should not throw" and yet - there is an error below. The comment should be fixed.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: Just gets a reference, will not deserialize again. This is just needed because when the auto entities are resolved the provider replaces its runtimeconfig with the new entities merged in, so we just update the reference to the config to the new one.

src/Core/Configurations/RuntimeConfigValidator.cs
aaronburtle: This is covered separately when we first check `IsRootConfig`
    `public bool IsRootConfig => DataSourceFiles?.SourceFiles?.Any() == true && !IsChildConfig;`

Then we validate it as a non-root config, since `IsRootConfig` would be false. We check

`if (config.DataSource is null)` and if it is give an error.

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: good catch will update this comment.

src/Core/Configurations/RuntimeConfigProvider.cs
aaronburtle: It doesn't, but this is the late configured path, so there should always be a datasource. However, I will add a guard to match the same behavior as before this change ie: return false.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -73,7 +101,7 @@ Runtime.GraphQL is null ||
         (Runtime is null ||
          Runtime.Rest is null ||
          Runtime.Rest.Enabled) &&
-         DataSource.DatabaseType != DatabaseType.CosmosDB_NoSQL;
+         DataSource?.DatabaseType != DatabaseType.CosmosDB_NoSQL;
```
Aniruddh25, line 104: If DataSource is not present in case of root config, this will evaluate to false even though the child configs with a DataSource are of type SQL.

The condition should be either its null or if not null, Db type !=NoSQL

src/Cli/ConfigGenerator.cs
```diff
@@ -402,7 +402,7 @@ public static bool TryAddNewEntity(AddOptions options, RuntimeConfig initialRunt
             // Try to get the source object as string or DatabaseObjectSource for new Entity
             if (!TryCreateSourceObjectForNewEntity(
                 options,
-                initialRuntimeConfig.DataSource.DatabaseType == DatabaseType.CosmosDB_NoSQL,
+                initialRuntimeConfig.DataSource!.DatabaseType == DatabaseType.CosmosDB_NoSQL,
```
Aniruddh25, line 414: how are we sure DataSource is not null here?

What happens when a new entity is being added to a config with a null data source? Is that allowed? If not, where is the check that will prevent this?

src/Cli/ConfigGenerator.cs
```diff
@@ -715,7 +715,7 @@ private static bool TryUpdateConfiguredDataSourceOptions(
             ConfigureOptions options,
             [NotNullWhen(true)] ref RuntimeConfig runtimeConfig)
         {
-            DatabaseType dbType = runtimeConfig.DataSource.DatabaseType;
+            DatabaseType dbType = runtimeConfig.DataSource!.DatabaseType;
```
Aniruddh25, line 736: How can we be sure DataSource is null when trying to update the datasource? Does the caller perform any checks?

Same question for all null forgiveness operator usage in this file.

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -387,4 +380,584 @@ private async Task ValidatePropertyOptionsFails(ConfigureOptions options)
...
+            ""runtime"": {
+                ""rest"": { ""enabled"": true },
+                ""graphql"": { ""enabled"": true },
+                ""host"": { ""mode"": ""development"" }
+            },
+            ""data-source-files"": [""child1.json""],
+            ""entities"": {}
+        }";
+
+        // The root config should parse without error (no data-source required for root).
```
Aniruddh25, line 428: Shouldnt the test also check that IsConfigValid returns true?

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -387,4 +380,584 @@ private async Task ValidatePropertyOptionsFails(ConfigureOptions options)
...
+
+    /// <summary>
+    /// Validates that a non-root config with a data-source and no entities parses
+    /// successfully. Validation of entity presence happens during dab validate,
+    /// not during parsing.
+    /// </summary>
+    [TestMethod]
+    public void TestNonRootConfigWithDataSourceAndNoEntitiesParses()
+    {
+        Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(INITIAL_CONFIG, out RuntimeConfig? config));
```
Aniruddh25, line 442: What makes this a Non root config? Does INITIAL_CONFIG string specify that?

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -387,4 +380,584 @@ private async Task ValidatePropertyOptionsFails(ConfigureOptions options)
...
+        Assert.IsFalse(config.IsRootConfig);
+    }
+
+    /// <summary>
+    /// Non-root with datasource and zero entities → error.
+    /// </summary>
+    [TestMethod]
+    public void TestNonRootWithDataSourceAndNoEntitiesProducesError()
+    {
+        RuntimeConfig config = BuildTestConfig(hasDataSource: true, entities: new());
```
Aniruddh25, line 453: What makes it non-root?

Aniruddh25 (changes requested):
IsRestEnabled check needs to be fixed.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: When `DataSource` is not present, and therefore evaluate to `null`, we will have `null != DatabaseType.CosmosDB_NoSQL` which evaluates to `true`. and `IsRestEnabled` will be `true` (assuming Rest isnt disabled). As far as I can tell this matches the behavior youre suggesting.

src/Cli.Tests/ValidateConfigTests.cs
aaronburtle: No, and that is intentional. This test is specifically about parsing, the validation coverage is handled in `TestRootWithNoDataSourceAndNoEntitiesIsValid`.

src/Cli.Tests/ValidateConfigTests.cs
aaronburtle: It is because `INITIAL_CONFIG` has a data-source but no data-source-files. The missing data-source-files makes this a non root config by definition.

src/Cli.Tests/ValidateConfigTests.cs
aaronburtle: Same as above, no data-source-file makes it non-root.

src/Cli/ConfigGenerator.cs
aaronburtle: We can't be sure that DataSource is not null here. I should add a guard so we dont get null reference exception. A new entity is not allowed to be added to a config with a null data source. A guard will prevent that.

src/Cli/ConfigGenerator.cs
aaronburtle: same as above, will add a guard to prevent null reference exception. this will make all of the null forgiving operators safe since there will be no way to reach them with a null.

src/Cli.Tests/ValidateConfigTests.cs
Aniruddh25: Thats incorrect, a root config can also NOT have `data-source` files. The test name should be fixed. It can be a root config too.

aaronburtle: A top level config can, but for this change the term root means both top level and has children.

Aniruddh25 (approved):
Some tests fail, and some renames necessary.

PR #3317, opened 2026-03-25
Add level-2 property to runtime.cache in dab.draft.schema.json

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3314

## What is this change?

Adds the `level-2` property, which is already a part of our object model, to the `dab.draft.schema.json` file. Also adds `minimum: 1` to `runtime.cache.ttl-seconds` to align schema validation with the runtime enforcement that rejects values ≤ 0. The `level-2.provider` property is kept as a plain string (no enum, no default) to match the nullable string with `null` default in the object model.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

N/A — schema-only change with no REST/GraphQL/CLI surface.

---

📍 Connect Copilot coding agent with [Jira](https://gh.io/cca-jira-docs), [Azure Boards](https://gh.io/cca-azure-boards-docs) or [Linear](https://gh.io/cca-linear-docs) to delegate work to Copilot in one click without leaving your project management tool.

### DIFF
--- a/src/Core/Resolvers/SqlMutationEngine.cs
+++ b/src/Core/Resolvers/SqlMutationEngine.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System.Data;
 using System.Data.Common;
 using System.Net;
 using System.Text.Json;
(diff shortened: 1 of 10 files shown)


schemas/dab.draft.schema.json
```diff
@@ -487,6 +487,30 @@
...
+              "additionalProperties": false,
+              "properties": {
+                "enabled": {
+                  "$ref": "#/$defs/boolean-or-string",
+                  "description": "Enable or disable the level 2 distributed cache.",
+                  "default": false
+                },
+                "provider": {
+                  "type": "string",
+                  "description": "The provider for the L2 cache. Currently only 'redis' is supported."
```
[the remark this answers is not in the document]
aaronburtle: @copilot but this isn't an enum in our object model, it is a nullable string, and the default value is null.

schemas/dab.draft.schema.json
```diff
@@ -487,6 +487,30 @@
...
+                },
+                "connection-string": {
+                  "type": "string",
+                  "description": "The connection string for the level 2 cache provider."
+                },
```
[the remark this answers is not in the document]
aaronburtle: valid, added.

schemas/dab.draft.schema.json
```diff
@@ -486,7 +486,32 @@
             "ttl-seconds": {
               "type": "integer",
               "description": "Time to live in seconds",
-              "default": 5
+              "default": 5,
+              "minimum": 1
```
souvikghosh04, line 490: was there a default `minimum 1` earlier? I guess 1 second cache TTL doesn't make much sense. we can probably keep the minimum 5 may be or let Jerry confirm and have it updated in spec- https://github.com/Azure/data-api-builder/issues/3314

aaronburtle: i dont think it makes sense either, but do we want to error out if they try to set it to 1-4?

souvikghosh04: ideally, we should give a warning or default to 5 automatically. such less TTL might not be of any use. please check if we can only have the `default` and remove the `minimum` since `default` can be _internally_ enforced to have minimum value of 5.

Aniruddh25: No error please, it breaks backward compatibility. No need for warning either - keep it simple - should be non-zero. We cannot estimate how frequently the user table gets updated, and no need to do so

Issue #3239, opened 2026-03-12
Refactor Log Level


RubenCerna2079:
With the changes made in PR #3203 it allows us to refactor other sections of the Log Level.
- Remove all sections that pass a logger as a parameter inside of a function in the `FileSystemRuntimeConfig.cs`
- Change the way we update all of the logs and see if they can use the new Dynamic Logger
- Check which logs are necessary and change the messages of the logs to make them more clear.

PR #3318, opened 2026-03-25
Fix logs still appearing even when LogLevel is set to `none` bug

### LINKED ISSUE #3255: [Bug]: `log-level` seems like it is not working when two in config.
Image

### LINKED ISSUE #3256: [Bug]: CLI logs being outputed even after LogLevel is set to none
Image

### LINKED ISSUE #3262: [Bug]: Setting `Azure.DataApiBuilder` to `none` does not filter output.
Image

RubenCerna2079:
## Why make this change?
- Closes issue #3262 
The logger for the Startup class is not initialized properly, since this logger is special due to the nature of the Startup class it needs to be continuously updated as DAB initializes. This causes two problems:
  - Some logs appear even when LogLevel is set to some value that would impede those logs to appear.
  - Some logs don't appear at all, even when LogLevel is set to a value that should allow them to be logged.

- Closes issue #3256 & #3255
The CLI logger still outputs some logs even when the LogLevel is set to `none`. It is expected that if the LogLevel set is `none` or some other level that shouldn't output the `information` level, the logs will not appear.

## What is this change?
Important Note: These changes currently only allow us to change the LogLevel from the CLI with the `default` namespace in the config file. An task was created to solve this issue: https://github.com/Azure/data-api-builder/issues/3451

In order to solve issue #3262:
- We removed the LogBuffer from the services inside of `Startup.cs`, this is necessary since we wanted each class to have its own LogBuffer so that we are able to tell from which logger the logs are being outputted.
- Then, we also correctly initialized the `Startup` logger by changing the method that it was using to initialize the logger, it now uses `CreateLoggerFactoryForHostedAndNonHostedScenario` which checks if there are any LogLevel namespaces from the config file that can be applicable for the specific logger. It is important to note that there are multiple places where the logs are flushed in order to cover for the cases in which an exception is found and causes DAB to end abruptly, and when we there is an IsLateConfigured scenario.
- We also changed the logger for the LogBuffer in all the missing places where it creates logs before the logger is able to properly initialize to add those logs to the LogBuffer and only flush them after the loggers are initialized.

In order to solve issue #3256 & #3255: 
- We changed the CLI so that we add all the logs go to a single global LogBuffer that is created inside the `StartOptions.cs` until it is able to deserialize the RuntimeConfig and find which level to set the `LogLevel` in order to flush all the logs.
- This is something that we only want to happen when we use the `dab start` command, which is why we only make this change in the `StartOptions.cs` file, on the function `TryStartEngineWithOptions` inside of `ConfigGenerator.cs`, and a few functions from `Utils.cs` and `ConfigMerger.cs` that are used inside the `TryStartEngine` function.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

## Sample Request(s)
- dab start --LogLevel none
- dab start --LogLevel error


src/Cli/CustomLoggerProvider.cs
```diff
@@ -79,8 +91,9 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
         /// <inheritdoc/>
         public bool IsEnabled(LogLevel logLevel)
         {
-            return true;
+            return logLevel != LogLevel.None && logLevel >= _minimumLogLevel;
```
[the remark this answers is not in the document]
RubenCerna2079: Test has been added to `EndToEndTests.cs`

src/Service/Startup.cs
```diff
@@ -649,42 +648,42 @@ private void AddGraphQLService(IServiceCollection services, GraphQLRuntimeOption
             {
                 if (error.Exception is not null)
                 {
-                    _logger.LogError(exception: error.Exception, message: "A GraphQL request execution error occurred.");
+                    _logBuffer.BufferLog(LogLevel.Error, "A GraphQL request execution error occurred.", error.Exception);
```
aaronburtle: These error filters are going to execute at runtime, after the log buffer is flushed, so these runtime errors will be enqueued into a buffer that will never be flushed, silently swallowing the errors. Same for the change on line 657.

src/Service/Startup.cs
```diff
@@ -649,42 +648,42 @@ private void AddGraphQLService(IServiceCollection services, GraphQLRuntimeOption
...
                     return error.WithMessage(error.Exception.Message);
                 }
 
                 if (error.Code is not null)
                 {
-                    _logger.LogError(message: "Error code: {errorCode}\nError message: {errorMessage}", error.Code, error.Message);
+                    _logBuffer.BufferLog(LogLevel.Error, $"Error code: {error.Code}\nError message: {error.Message}");
```
aaronburtle: same as above

src/Config/LogBuffer.cs
```diff
@@ -35,9 +35,9 @@ public void FlushToLogger(ILogger? targetLogger)
         {
             lock (_flushLock)
             {
-                while (_logBuffer.TryDequeue(out (LogLevel LogLevel, string Message) entry))
+                while (_logBuffer.TryDequeue(out (LogLevel LogLevel, string Message, Exception? Exception) entry))
```
aaronburtle, line 38: This function allows for null targetLogger, but doesn't seem to check if the targetLogger is null before dequeue. Doesn't that mean if the targetLogger is null, we just lose all the errors in the buffer? Should we skip flushing until targetLogger is not null, or maybe find another work around?

src/Service/Startup.cs
```diff
@@ -705,21 +704,27 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, RuntimeC
...
+                // Configure Telemetry
+                // TODO: Issue #3239. Refactor this methods so that they are all called before creating the new logger factory.
+                ConfigureApplicationInsightsTelemetry(app, runtimeConfig, logLevelInit);
+                ConfigureOpenTelemetry(app, runtimeConfig, logLevelInit);
+                ConfigureAzureLogAnalytics(app, runtimeConfig, logLevelInit);
+                ConfigureFileSink(app, runtimeConfig, logLevelInit);
+
+                //Flush all logs that were buffered before setting the LogLevel.
+                // Important: All logs set before this point should use _logBuffer.
+                _logBuffer.FlushToLogger(_logger);
```
aaronburtle: If there is an exception raised before we flush the logger here, then the buffer never flushes and any errors are skipped from being logged. similarly if `TryGetConfig` returns false I dont think we flush the logger for the late configured path.

I think we should use a try/finally block here to ensure that we always flush the logger.

src/Service/Startup.cs
```diff
@@ -375,7 +374,7 @@ public void ConfigureServices(IServiceCollection services)
                     int port = PortResolutionHelper.ResolveInternalPort();
                     string baseUri = $"http://localhost:{port}";
                     client.BaseAddress = new Uri(baseUri);
-                    _logger.LogInformation($"Configured HealthCheck HttpClient BaseAddress as: {baseUri}");
+                    _logBuffer.BufferLog(LogLevel.Information, $"Configured HealthCheck HttpClient BaseAddress as: {baseUri}");
```
aaronburtle: Similar to the GraphQL error filter comment, this callback is only going to be invoked once the httpClient is resolved by direct injection, and by that point, the buffer will have already been flushed, meaning any log entries will sit in the buffer permanently.

src/Cli/ConfigGenerator.cs
```diff
@@ -37,6 +38,11 @@ public static void SetLoggerForCliConfigGenerator(
             _logger = configGeneratorLoggerFactory;
         }
 
+        public static void SetBufferForCliConfigGenerator(LogBuffer cliBuffer)
+        {
+            _cliBuffer = cliBuffer;
```
aaronburtle: This static method sets a static logbuffer globally, but StartOptions created a new log buffer each time it is invoked. This is probably going to work fine for now, but it creates a landmine where if we ever have parallel test execution or we invoke multiple StartOption instances for some reason we will have a race condition and logs from one test could flush to another, creating strange behavior.

src/Cli/Commands/StartOptions.cs
```diff
@@ -48,13 +49,13 @@ public StartOptions(bool verbose, LogLevel? logLevel, bool isHttpsRedirectionDis
 
         public int Handler(ILogger logger, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
```
aaronburtle, line 52: Looks like all the logging is now handled with SendLogToBufferOrLogger now, is the logger param here still used?

src/Service/Startup.cs
aaronburtle: also, after this is fixed (or if you make it into a followup issue), there will be another issue which is we still have some usage of the console to account for this, for example, errors during deserialization. We will need to adjust console usage to rely on the logger buffer once we know it will flush properly even during a halted startup.

src/Config/LogBuffer.cs
RubenCerna2079: I made it that way on purpose, but I will change it so that it throws a warning or an error that there is no logger and the logs cannot be flushed

aaronburtle: if the target logger is null and we are going to flush the buffer, should we instead print those logs to console? is there a case where we prefer those logs to be lost entirely?

RubenCerna2079: We don't want to just print the logs to console since that may cause issues for the MCP, I think it is better if we just change it so that those logs cannot be flushed unless the target logger is not null.

src/Service/Startup.cs
```diff
@@ -705,21 +704,27 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, RuntimeC
...
+                LogLevelInitializer logLevelInit = new(MinimumLogLevel, typeof(Startup).FullName, runtimeConfigProvider, _hotReloadEventHandler);
+
                 // Set LogLevel based on RuntimeConfig
                 DynamicLogLevelProvider logLevelProvider = app.ApplicationServices.GetRequiredService<DynamicLogLevelProvider>();
                 logLevelProvider.UpdateFromRuntimeConfig(runtimeConfig);
                 FileSystemRuntimeConfigLoader configLoader = app.ApplicationServices.GetRequiredService<FileSystemRuntimeConfigLoader>();
 
-                //Flush all logs that were buffered before setting the LogLevel
+                // Configure Telemetry
+                // TODO: Issue #3239. Refactor this methods so that they are all called before creating the new logger factory.
```
Aniruddh25: Is this issue still active #3239?

src/Service/Startup.cs
```diff
@@ -106,15 +106,14 @@ public class Startup(IConfiguration configuration, ILogger<Startup> logger)
         public void ConfigureServices(IServiceCollection services)
         {
             Startup.AddValidFilters();
-            services.AddSingleton(_logBuffer);
```
Aniruddh25, line 109: Dont we add the _logBuffer to service collection anymore?

RubenCerna2079: No, we now have them as separate objects because we want each class to have its own buffer so that they can only access their own logs and not output logs that belong to a different class.

src/Cli/ConfigGenerator.cs
RubenCerna2079: I decided to make a separate logbuffer for each of the different classes instead of using it globally, this should fix the race conditions that might occur. I also added the flush to the StartOptions as a way to prevent logs from being lost if it errors out.

src/Cli/Commands/StartOptions.cs
RubenCerna2079: With the new changes to how we use the logbuffer this logger is still necessary.

src/Service/Startup.cs
RubenCerna2079: Yes, the issue is still active and I am adding that TODO since it is not something that important to finish for 2.0.

src/Cli.Tests/UtilsTests.cs
```diff
@@ -255,7 +255,7 @@ public void TestMergeConfig()
 
         Environment.SetEnvironmentVariable(RUNTIME_ENVIRONMENT_VAR_NAME, "Test");
 
-        Assert.IsTrue(ConfigMerger.TryMergeConfigsIfAvailable(fileSystem, loader, new StringLogger(), out string? mergedConfig), "Failed to merge config files");
+        Assert.IsTrue(ConfigMerger.TryMergeConfigsIfAvailable(fileSystem, loader, new StringLogger(), null, out string? mergedConfig), "Failed to merge config files");
```
Aniruddh25: name the parameter that is `null` for better readability

src/Cli/Utils.cs
```diff
@@ -961,10 +964,10 @@ public static bool IsEntityProvided(string? entity, ILogger cliLogger, string co
         /// <summary>
         /// Returns ILoggerFactory with CLI custom logger provider.
         /// </summary>
-        public static ILoggerFactory GetLoggerFactoryForCli()
+        public static ILoggerFactory GetLoggerFactoryForCli(LogLevel minimumLogLevel = LogLevel.Information)
```
Aniruddh25, line 1014: PR title should be better explanatory - it is related to loglevel - title doesnt say anything about loglevel

src/Service/Startup.cs
Aniruddh25: This is an important detail - worth of going into PR description - as to why we need different LogBuffer objects -and what problem it causes if we dont

src/Cli/ConfigGenerator.cs
```diff
@@ -2555,7 +2557,7 @@ public static bool VerifyCanUpdateRelationship(RuntimeConfig runtimeConfig, stri
         /// </summary>
         public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRuntimeConfigLoader loader, IFileSystem fileSystem)
         {
-            if (!TryGetConfigForRuntimeEngine(options.Config, loader, fileSystem, out string runtimeConfigFile))
+            if (!TryGetConfigForRuntimeEngine(options.Config, loader, fileSystem, out string runtimeConfigFile, false))
```
Aniruddh25: nit: name the argument that is false

src/Cli/ConfigGenerator.cs
```diff
@@ -3622,5 +3640,23 @@ private static bool ValidateFields(
 
             return true;
         }
+
+        /// <summary>
+        /// Helper method that sends the log to the buffer if the buffer has being set up.
+        /// Else, it will send the log to the logger.
+        /// </summary>
+        /// <param name="logLevel">LogLevel of the log.</param>
+        /// <param name="message">Message that will be printed in the log.</param>
```
Aniruddh25: missing description of exception

src/Service/Telemetry/DynamicLogLevelProvider.cs
```diff
@@ -8,21 +8,43 @@ public class DynamicLogLevelProvider
         public LogLevel CurrentLogLevel { get; private set; }
         public bool IsCliOverridden { get; private set; }
 
+        /// <summary>
```
Aniruddh25, line 36: Can you provide a brief explanation in the PR description as to what the flow is and how it is fixing the problem? 

The description currently only explains what changes are made in each of the files - but lacks to connect the dots to inform how those changes fix the problem

src/Cli/ConfigGenerator.cs
```diff
@@ -2696,16 +2713,17 @@ public static bool TryGetConfigForRuntimeEngine(
...
             {
-                _logger.LogInformation("Using merged config file based on environment: {configToBeUsed}.", configToBeUsed);
+                SendLogToBufferOrLogger(_logger, CliBuffer, LogLevel.Information, $"Using merged config file based on environment: {configToBeUsed}.");
             }
 
-            if (!TryGetConfigFileBasedOnCliPrecedence(loader, configToBeUsed, out runtimeConfigFile))
+            if (!TryGetConfigFileBasedOnCliPrecedence(loader, configToBeUsed, out runtimeConfigFile, useLogger))
             {
-                _logger.LogError("Config not provided and default config file doesn't exist.");
+                SendLogToBufferOrLogger(_logger, CliBuffer, LogLevel.Error, "Config not provided and default config file doesn't exist.");
```
Aniruddh25: When are the logs sent to `ConfigGenerator.CliBuffer` flushed to logger from the call to TryGetConfigForRuntimeEngine

Aniruddh25 (changes requested):
Not clear how using different LogBuffer is helping to fix CLI logs being outputed even after LogLevel is set to None.

src/Cli/ConfigGenerator.cs
```diff
@@ -2589,23 +2591,38 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
-                _logger.LogInformation("Setting minimum LogLevel: {minimumLogLevel}.", minimumLogLevel);
+                SendLogToBufferOrLogger(_logger, CliBuffer, LogLevel.Information, $"Setting minimum LogLevel: {minimumLogLevel}.");
             }
             else
             {
                 minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
                 HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
 
-                _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
+                SendLogToBufferOrLogger(_logger, CliBuffer, LogLevel.Information, $"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.");
```
Aniruddh25: We should use the minimumLogLevel just defined here to log this message as well: Setting default minimum LogLevel.

src/Cli/ConfigGenerator.cs
```diff
@@ -3622,5 +3640,23 @@ private static bool ValidateFields(
...
+        /// Helper method that sends the log to the buffer if the buffer has being set up.
+        /// Else, it will send the log to the logger.
+        /// </summary>
+        /// <param name="logLevel">LogLevel of the log.</param>
+        /// <param name="message">Message that will be printed in the log.</param>
+        public static void SendLogToBufferOrLogger(Microsoft.Extensions.Logging.ILogger logger, LogBuffer? cliBuffer, LogLevel logLevel, string message, Exception? ex = null)
+        {
+            if (cliBuffer is not null)
+            {
+                cliBuffer.BufferLog(logLevel, message, ex);
```
Aniruddh25: Why do we need this helper method? The caller is anyway passing both the logger and the logbuffer to this function. Its not like the caller is unaware, if the CliBuffer is null or not, if its a static member that you initialize then it is always going to be valid. You can simply call the BufferLog in the caller itself.

src/Service/Startup.cs
RubenCerna2079: I updated it

src/Cli/Utils.cs
RubenCerna2079: Changed PR title

src/Cli/ConfigGenerator.cs
RubenCerna2079: They are flushed in the `Handler` function of `StartOptions`, if the `ConfigGenerator.TryStartEngineWithOptions` fails it will return as false which is then used by the `Handler` to log a new error.
If the function succeeds it will flush the logs in here: 
https://github.com/Azure/data-api-builder/pull/3318/changes#diff-e0639603f6dd5b6ae5bd8473dad9a99052ae7a827a78f035d326573ff1f0ea4bR2624

src/Cli/ConfigGenerator.cs
RubenCerna2079: We already use the minimuLogLevel inside of the message.

src/Cli/ConfigGenerator.cs
RubenCerna2079: There are some places where we use it where it is not always necessary as the loggers have already been initialized properly, so we pass the LogBuffer as a null so that it directly logs the message. I will also add this to the description to make it easier to understand.

Aniruddh25: In those other callers, where the logger directly logs - you can simply use logger.Log. There you dont need the buffer. 

This wrapper is unnecessary, we can simply either flush to buffer or do logger.log in the respective callers. 
That way, its easier to read as well - otherwise, in the various callers - the reader is unaware what is going to happen unless they look into the function definition.

Aniruddh25:
> For the CLI we first check if the LogBuffer is null, to see if we need to save the log inside the buffer or send directly through the logger, we do this since some of the functions that are used before the loggers are initialized are also used after the logger initialization, so we need a way to ensure that the logs aren't saved after we flush the logger as we are not expecting to use it anymore, which means the logs would be lost.

I do not understand this.

Aniruddh25:
If we use different LogBuffers for different classes, wouldn't the logs get out of place as we will be flushing the log buffers in a different order of collecting the logs.

Aniruddh25 (commented):
Why does the description say we want to fix logs only from `dab start`? Why not the other commands?

Aniruddh25 (approved):

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -852,30 +852,35 @@ public void TestEngineStartUpWithVerboseAndLogLevelOptions(string logLevelOption
...
+        Task engineTask = Task.Run(() => Program.Execute(args, logger, _fileSystem!, _runtimeConfigLoader!));
 
-        string? output = process.StandardOutput.ReadLine();
-        Assert.IsNull(output);
-        process.Kill();
+        // Wait for the CLI to set up the proper LogLevel.
+        await Task.Delay(TimeSpan.FromSeconds(5));
+
+        string engineStdOut = consoleOutput.ToString();
+        Assert.IsTrue(string.IsNullOrEmpty(engineStdOut), $"Expected no output at LogLevel {logLevelOption}, but got: {engineStdOut}");
```
Aniruddh25, line 914: Why is the expectation that there will not be any output at the LogLevel specified by logLevelOption? Shouldnt that be expected only when the level is `None`?

RubenCerna2079: At first, I thought that the test would only cover the logs in the CLI but since we are also running the DAB engine, I changed it so we only use the `None` LogLevel in this case.

Aniruddh25: We are missing coverage if you remove the other test cases. Is there a different case that covers those? If not, please retain the data rows and fix the test to ensure it succeeds. 

At this point, both engine and CLI should behave uniformly and honor output as per the log level option.

Aniruddh25 (changes requested):
Missing test cases.

src/Cli.Tests/EndToEndTests.cs
RubenCerna2079: I added the missing test cases so that they check there are no `info` LogLevel messages.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -848,6 +848,44 @@ public void TestEngineStartUpWithVerboseAndLogLevelOptions(string logLevelOption
...
+        Task engineTask = Task.Run(() => Program.Execute(args, logger, _fileSystem!, _runtimeConfigLoader!));
+
+        // Wait for the CLI to set up the proper LogLevel.
+        await Task.Delay(TimeSpan.FromSeconds(5));
+
+        string engineStdOut = consoleOutput.ToString();
+        Assert.IsNotNull(engineStdOut);
+        Assert.IsFalse(engineStdOut.Contains("info"), $"Expected no 'info' outputs at LogLevel {logLevelOption}, but got: {engineStdOut}");
+
+        Utils.LoggerFactoryForCli = Utils.GetLoggerFactoryForCli();
```
Aniruddh25: What is the point of setting the logger factory for CLI at the end?

Aniruddh25 (changes requested):
Why is it only for `dab start` why not for other commands?

RubenCerna2079:
> Why is it only for `dab start` why not for other commands?

That is not part of the scope from this PR. Created a new issue to track this https://github.com/Azure/data-api-builder/issues/3465

src/Cli.Tests/EndToEndTests.cs
RubenCerna2079: The test changes the minimum LogLevel for all the CLI loggers, which are static. If we do not reset it to its original minimum LogLevel it causes issues in other tests that rely on the ensuring the test works by checking it receives the correct log message.

RubenCerna2079: Also added a comment in the test explaining what it is for

Aniruddh25 (approved):

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
```diff
@@ -482,9 +492,18 @@ public async Task ChildConfigLoadFailureHaltsParentConfigLoading()
...
+                ILoggerFactory loggerFactory = LoggerFactory.Create(builder =>
+                {
+                    builder.SetMinimumLevel(LogLevel.Trace);
+                    builder.AddConsole();
+                });
+                ILogger<FileSystemRuntimeConfigLoader> logger = loggerFactory.CreateLogger<FileSystemRuntimeConfigLoader>();
 
+                loader.SetLogger(logger);
                 bool loaded = loader.TryLoadConfig("dab-config.json", out RuntimeConfig _);
+                await Task.Delay(TimeSpan.FromSeconds(2));
```
Aniruddh25: why is this delay being added? Any time a delay is expected, the test becomes flaky. 

Can you also check for the existence of the right post delay condition and retry the wait if needed? This should be the standard anytime you add a delay - because 2 seconds may work now, but not in future

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
```diff
@@ -95,7 +95,7 @@ public async Task CheckExceptionForBadConnectionStringForMsSql(string connection
             if (isInvalidConnectionBuilderString)
             {
                 sw = new();
-                Console.SetError(sw);
+                Console.SetOut(sw);
```
Aniruddh25: Why is the log flushing to std out instead of stderr?

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
RubenCerna2079: I added the delay because the logs were not being completely written inside the string writer when it checked if the expected log was there. So it was causing it to fail when everything was working fine.

src/Service.Tests/UnitTests/SqlMetadataProviderUnitTests.cs
RubenCerna2079: It is flushing that way because of how I set the logger factory, will change it

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
RubenCerna2079: Added a new function based on the delay condition and retry, since it was only found inside one test.

Aniruddh25 (approved):

Aniruddh25 (approved):

src/Config/FileSystemRuntimeConfigLoader.cs
aaronburtle, line 185: This needs to be set to use the logger buffer, or we need to track this as a follow up issue since this will go to console on startup.

aaronburtle, line 212: This will go to console on hot-reload. Lower priority than the startup path, but should still be tracked for moving to the logger.

src/Cli.Tests/EnvironmentTests.cs
```diff
@@ -162,7 +162,7 @@ public async Task FailureToStartEngineWhenEnvVarNamedWrong()
             $"-c {TEST_RUNTIME_CONFIG_FILE}"
         );
 
-        string? output = await process.StandardError.ReadLineAsync();
+        string? output = await process.StandardOutput.ReadToEndAsync();
         Assert.IsNotNull(output);
         // Clean error message on stderr with no stack trace.
```
aaronburtle, line 167: I think this will not go to stdout, not stderr, should either change this to stderr somehow or update the comment.

src/Service/Telemetry/DynamicLogLevelProvider.cs
aaronburtle, line 82: is this /// </summary> tag now orphaned? should remove I guess.

src/Service/Startup.cs
```diff
@@ -702,44 +701,69 @@ private void RefreshGraphQLSchema(IServiceCollection services)
...
-                // Config provided during runtime.
-                runtimeConfigProvider.IsLateConfigured = true;
-                runtimeConfigProvider.RuntimeConfigLoadedHandlers.Add(async (_, _) =>
-                {
-                    isRuntimeReady = await PerformOnConfigChangeAsync(app);
-                    return isRuntimeReady;
-                });
+                // Attempt one final flush in case there was any exception that caused the
+                // previous section to throw an error before flushing the logs.
+                FlushAllLogs(app);
```
aaronburtle, line 767: I think technically, this function can throw an exception, which could replace the original exception in the try block. I think we might want to wrap this in its own try catch to swallow any exceptions from FlushAllLogs, or surface alongside the original try.

src/Service/Startup.cs
```diff
@@ -702,44 +701,69 @@ private void RefreshGraphQLSchema(IServiceCollection services)
...
+                        // This section will only run if the runtime config is provided during runtime. E.g. IsLateConfigured is true.
+                        // Set LogLevel based on RuntimeConfig
+                        RuntimeConfigProvider runtimeConfigProvider = app.ApplicationServices.GetService<RuntimeConfigProvider>()!;
+                        RuntimeConfig runtimeConfig = runtimeConfigProvider.GetConfig();
+                        DynamicLogLevelProvider logLevelProvider = app.ApplicationServices.GetRequiredService<DynamicLogLevelProvider>();
+                        logLevelProvider.UpdateFromRuntimeConfig(runtimeConfig);
+
+                        //Flush all logs that were buffered before setting the LogLevel.
+                        // Important: All logs set before this point should use _logBuffer.
+                        FlushAllLogs(app);
```
aaronburtle, line 755: I dont think we need this FlushAllLogs call since it will happen against in finally block.

src/Cli/ConfigGenerator.cs
```diff
@@ -2843,16 +2862,32 @@ public static bool TryGetConfigForRuntimeEngine(
...
+            out string runtimeConfigFile,
+            LogBuffer? logBuffer = null)
         {
-            if (string.IsNullOrEmpty(configToBeUsed) && ConfigMerger.TryMergeConfigsIfAvailable(fileSystem, loader, _logger, out configToBeUsed))
+            if (string.IsNullOrEmpty(configToBeUsed) && ConfigMerger.TryMergeConfigsIfAvailable(fileSystem, loader, _logger, logBuffer, out configToBeUsed))
             {
-                _logger.LogInformation("Using merged config file based on environment: {configToBeUsed}.", configToBeUsed);
+                if (logBuffer is null)
+                {
+                    _logger.LogInformation("Using merged config file based on environment,: {configToBeUsed}.", configToBeUsed);
```
aaronburtle: I think this has a typo (environment,:), and its also creating a different message for same condition in the buffer branch.

_logger.LogInformation("Using merged config file based on environment,: {configToBeUsed}.", configToBeUsed);

 vs

logBuffer.BufferLog(LogLevel.Information, $"Merged config file based on environment is available: {configToBeUsed}.");

Can probably just use same message for consistency.

src/Service.Tests/TestHelper.cs
```diff
@@ -381,5 +395,24 @@ public static string AddPropertiesToJson(string configuration, string entityProp
...
             return configurationJson.ToString();
         }
+
+        /// <summary>
+        /// Helper function to add retry logic with delay when there is a function
+        /// that requires the to wait some time for a condition to be met.
+        /// </summary>
+        /// <param name="condition">Boolean condition that needs to be met</param>
+        /// <returns></returns>
+        public static async Task DelayTask(Func<bool> condition)
```
aaronburtle, line 405: Doesn't seem that there is any early exit here, so dont we always go through the entire retry count every time?

aaronburtle:
Something to note, and maybe we can discuss offline, is that there are still console usages that happen before we get any log buffer, in the pre-host code paths, ie: Program.cs:53, and we might want to move those to logger too at some point, but out of scope of this PR. Likewise, there are hot-reload code paths now that I think have access to a logger that at some point should make use of, but again out of scope of this PR. But we should maybe track these in an issue in backlog.

src/Service/Startup.cs
```diff
@@ -1363,5 +1389,18 @@ public static bool IsAzureLogAnalyticsAvailable(AzureLogAnalyticsOptions azureLo
...
+        /// <summary>
+        /// Helper function that sets the logger for FileSystemRuntimeConfigLoader and flushes the buffered logs to the logger.
+        /// </summary>
+        /// <param name="app">Contains all the services needed to set the logger.</param>
+        private void FlushAllLogs(IApplicationBuilder app)
+        {
+            FileSystemRuntimeConfigLoader configLoader = app.ApplicationServices.GetRequiredService<FileSystemRuntimeConfigLoader>();
+            configLoader.SetLogger(app.ApplicationServices.GetRequiredService<ILogger<FileSystemRuntimeConfigLoader>>());
+
+            _logBuffer.FlushToLogger(_logger);
```
aaronburtle: I think this will create confusing logs because of the ordering here. During startup, the RuntimeConfigProvider will call GetConfig, which will queue up logs into the loader buffer, ie: "Loading the config file from...", and then later Startup.Configure will make calls that queue up logs into the Startup Buffer, ie: "Application Insights are disabled", but then the startup buffer will flush first, so I think that means logs will show the latter before the former.


And because I dont think we have a guarantee that either of the log buffer will be written to entirely before the other I dont think simply reversing the order here will help.


The easiest way to fix this I think is to use just 1 log buffer, inject the same logBuffer into the loader's constructor and have startup use the same instance. Should give correct ordering automatically.

Another option would be to use time stamps and merge before flush, but I think thats overly complicated.

src/Config/LogBuffer.cs
```diff
@@ -10,34 +10,34 @@ namespace Azure.DataApiBuilder.Config
...
     /// </summary>
-    public class StartupLogBuffer
+    public class LogBuffer
     {
-        private readonly ConcurrentQueue<(LogLevel LogLevel, string Message)> _logBuffer;
+        private readonly ConcurrentQueue<(LogLevel LogLevel, string Message, Exception? Exception)> _logBuffer;
         private readonly object _flushLock = new();
 
-        public StartupLogBuffer()
+        public LogBuffer()
```
aaronburtle, line 18: since this is no longer used via DI, it may not need to remain public, I think setting it to internal would be fine, but your call to make.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -859,6 +847,76 @@ public void TestEngineStartUpWithVerboseAndLogLevelOptions(string logLevelOption
...
+
+        // Wait for the CLI to set up the proper LogLevel.
+        await Task.Delay(TimeSpan.FromSeconds(5));
+
+        string engineStdOut = consoleOutput.ToString();
+        Assert.IsNotNull(engineStdOut);
+        Assert.IsFalse(engineStdOut.Contains("info"), $"Expected no 'info' outputs at LogLevel {logLevelOption}, but got: {engineStdOut}");
+
+        // Reset the LoggerFactoryForCli to avoid impacting other tests.
+        Utils.LoggerFactoryForCli = Utils.GetLoggerFactoryForCli();
```
aaronburtle: The new tests reset here as you mentioned in another comment, but I think we need to handle this for existing tests as well. Otherwise, we risk flakey tests because the LoggerFactoryForCli is mutable and actually being changed as a part of testing, where as before it was write once at start, now it could change and possibly introduce order dependancy or flakiness.

aaronburtle (approved):
Approved but with some comments. Some of these can be tracked in new issues, but some will need to be completed before we merge. Please address comments before merging.

src/Service/Telemetry/DynamicLogLevelProvider.cs
RubenCerna2079: It is not orphaned, it is used to make it more readable, so I don't think it should be removed.

src/Service/Startup.cs
RubenCerna2079: We still need to call it there since there is a possibility that there are more logs buffered between the time it takes for the user to start DAB and the time they send the config file through the API.

src/Service.Tests/TestHelper.cs
RubenCerna2079: It would go through the retry count but wouldn't be delayed since the condition would not be met anymore. But I will add the condition to the while statement so it doesn't have to do the count every time.

src/Service/Startup.cs
RubenCerna2079: While I agree with you, this PR has been going for far too long, and I would consider to be out of the scope of this PR. I created an issue to track this https://github.com/Azure/data-api-builder/issues/3494

src/Config/FileSystemRuntimeConfigLoader.cs
RubenCerna2079: Created issue https://github.com/Azure/data-api-builder/issues/3493 to track this

RubenCerna2079: Created issue https://github.com/Azure/data-api-builder/issues/3493 to track this

src/Config/LogBuffer.cs
RubenCerna2079: We cannot change it as it causes DAB to fail at startup

src/Cli.Tests/EndToEndTests.cs
RubenCerna2079: Will move it to test cleanup so it is reset after every test run

src/Service/Startup.cs
RubenCerna2079: I am not following how this can throw its own exception? Is it because we try to get the logger for `FileSystemRuntimeConfiigLoader`?

src/Cli.Tests/EnvironmentTests.cs
RubenCerna2079: Found that there was a change to where we output our logs in the `CustomLoggerProvider`, will change it so that this error stops occuring.

src/Service/Startup.cs
RubenCerna2079: Unsure of what type of exception it might be the one you are referring to. So I made it assuming the cause is what I mentioned above.

Aniruddh25 (approved):

Issue #3319, opened 2026-03-26
[Bug]: Hot-reload race condition in test


aaronburtle:
### What happened?

In https://github.com/Azure/data-api-builder/pull/3297 we introduced a race condition between a rest request and server initialization after hot-reload. Need to apply the same pattern as we have on similar rest calls in the test file.

### Version

2.0

### What database are you using?

Azure SQL

### Relevant log output

```Text

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3320, opened 2026-03-26
Fix hot-reload race condition in Autoentities test

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3319

## What is this change?

Apply the same pattern for making the rest request in this test as in the rest of the test file.

## How was this tested?

Ran against pipeline.

### DIFF
--- a/src/Service.Tests/OpenApiDocumentor/MissingRoleNotFoundTests.cs
+++ b/src/Service.Tests/OpenApiDocumentor/MissingRoleNotFoundTests.cs
@@ -1,0 +1,98 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Abstractions;
+using System.Net;
+using System.Net.Http;
+using System.Text.Json;
+using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Service.Tests/Configuration/HotReload/ConfigurationHotReloadTests.cs
```diff
@@ -818,7 +818,10 @@ await WaitForConditionAsync(
           TimeSpan.FromMilliseconds(500));
 
         HttpResponseMessage failRestResult = await _testClient.GetAsync($"rest/autoentity_books");
-        HttpResponseMessage hotReloadRestResult = await _testClient.GetAsync($"rest/HotReload_books");
+
+        // After hot-reload, the engine may still be re-initializing metadata providers.
+        // Poll the REST endpoint to allow time for the engine to become fully ready.
+        using HttpResponseMessage hotReloadRestResult = await WaitForRestEndpointAsync("rest/HotReload_books", HttpStatusCode.OK);
```
RubenCerna2079, line 822: If we are waiting on this REST request to get a 200 status then it would be better to move it before the REST request that is failing, since we want to do both tests after the metadata is completely set.
This will let us know if the autoentities hot-reloaded properly

PR #3323, opened 2026-03-26
Let parent config AKV settings propagate to child configs

### LINKED ISSUE #3322: [Bug]: Parent config AKV settings do not propagate to child configs
### What happened?

When a parent config has AKV settings, and a child config does not, if the child config uses `@akv('')`, the settings needed to do the replacement will be missing, and the replacement will be unable to be performed.

### Version

2.0

Azure SQL

_No response_

_No response_

### Relevant log output

```Text

```

### Code of Conduct

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3322

When a parent config has `azure-key-vault` configured, child configs referenced via `data-source-files` were unable to resolve `@akv('...')` references because the parent's AKV options were not passed during child config deserialization.

## What is this change?

- Pass the parent config's `AzureKeyVault` options into `DeserializationVariableReplacementSettings` when loading each child config, enabling `@akv('...')` references in child configs to be resolved using the parent's Key Vault configuration.
- The `DeserializationVariableReplacementSettings` object is created once before the child config `foreach` loop and reused for all child configs, avoiding redundant Key Vault client initialization or secrets file reads per iteration.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

Regression test added (`ChildConfigResolvesAkvReferencesFromParentAkvOptions`) that loads a parent config referencing a local `.akv` secrets file and a child config with an `@akv()` connection string, validating the connection string is correctly resolved without requiring a real Azure Key Vault.

### DIFF
--- a/src/Config/ObjectModel/RuntimeConfig.cs
+++ b/src/Config/ObjectModel/RuntimeConfig.cs
@@ -354,11 +354,14 @@ public record RuntimeConfig
             // This loader is not used as a part of hot reload and therefore does not need a handler.
             FileSystemRuntimeConfigLoader loader = new(fileSystem, handler: null);
 
+            // Pass the parent's AKV options so @akv() references in child configs can
+            // be resolved using the parent's Key Vault configuration.
+            // If a child config defines its own azure-key-vault section, TryParseConfig's
+            // ExtractAzureKeyVaultOptions will detect it and override these parent options.
+            DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: this.AzureKeyVault, doReplaceEnvVar: true, doReplaceAkvVar: true, envFailureMode: EnvironmentVariableReplacementFailureMode.Ignore);
+
             foreach (string dataSourceFile in DataSourceFiles.SourceFiles)
             {
-                // Use Ignore mode so missing env vars are left as literal @env() strings,
-                // consistent with how the parent config is loaded in TryLoadKnownConfig.
-                DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: null, doReplaceEnvVar: true, doReplaceAkvVar: true, envFailureMode: EnvironmentVariableReplacementFailureMode.Ignore);
 
                 if (loader.TryLoadConfig(dataSourceFile, out RuntimeConfig? config, replacementSettings: replacementSettings))
                 {
--- a/src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
+++ b/src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
@@ -134,6 +134,227 @@ public class RuntimeConfigLoaderTests
     }
 
     /// <summary>
+    /// Validates that when a parent config has azure-key-vault options configured,
+    /// child configs can resolve @akv('...') references using the parent's AKV configuration.
+    /// Uses a local .akv file to simulate Azure Key Vault without requiring a real vault.
+    /// Regression test for https://github.com/Azure/data-api-builder/issues/3322
+    /// </summary>
+    [TestMethod]
+    public async Task ChildConfigResolvesAkvReferencesFromParentAkvOptions()
+    {
+        string akvFilePath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName() + ".akv");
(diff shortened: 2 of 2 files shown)


aaronburtle:
@copilot can you review this again please

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -353,10 +353,12 @@ public RuntimeConfig(
             // This loader is not used as a part of hot reload and therefore does not need a handler.
             FileSystemRuntimeConfigLoader loader = new(fileSystem, handler: null);
 
+            // Pass the parent's AKV options so @akv() references in child configs can
+            // be resolved using the parent's Key Vault configuration.
+            DeserializationVariableReplacementSettings replacementSettings = new(azureKeyVaultOptions: this.AzureKeyVault, doReplaceEnvVar: true, doReplaceAkvVar: true);
```
Aniruddh25: Are we sure if the child config has its own AKV options, those will override the ones passed onto by the parent?

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
```diff
@@ -131,4 +132,108 @@ public async Task CanLoadValidMultiSourceConfigWithAutoentities(string configPat
...
+            Assert.IsTrue(
+                loader.TryLoadConfig("dab-config.json", out RuntimeConfig runtimeConfig, replacementSettings: replacementSettings),
+                "Config should load successfully when child config has @akv() references resolvable via parent AKV options.");
+
+            Assert.IsTrue(runtimeConfig.Entities.ContainsKey("AkvChildEntity"), "Child config entity should be merged into the parent config.");
+
+            // Verify the child's connection string was resolved from the .akv file.
+            string childDataSourceName = runtimeConfig.GetDataSourceNameFromEntityName("AkvChildEntity");
+            DataSource childDataSource = runtimeConfig.GetDataSourceFromDataSourceName(childDataSourceName);
+            Assert.IsTrue(
```
Aniruddh25, line 223: Can we also verify in a different case that the child AKV settings override the ones passed onto by the parent?

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: Yes, the child options will take precedent

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
aaronburtle: Added this case.

src/Config/ObjectModel/RuntimeConfig.cs
```diff
@@ -1,867 +1 @@
...
...ty? entityConfig))        {            throw new DataApiBuilderException(                message: $"{entityName} is not a valid entity.",                statusCode: HttpStatusCode.BadRequest,                subStatusCode: DataApiBuilderException.SubStatusCodes.EntityNotFound);        }        return IsEntityCachingEnabled(entityConfig);    }    /// <summary>    /// Determines whether caching is enabled for a given entity, resolving inheritance lazily.    /// If the entity explicitly sets cache enabled/disabled (UserProvidedEnabledOptions is true), that value wins.    /// Otherwise, inherits the global runtime cache enabled setting.    /// </summary>    /// <param name="entity">The entity to check cache configuration.</param>    /// <returns>Whether caching is enabled for the entity.</returns>    private bool IsEntityCachingEnabled(Entity entity)    {        // If entity has an explicit cache config with user-provided enabled value, use it.        if (entity.Cache is not null && entity.Cache.UserProvidedEnabledOptions)        {            return entity.IsCachingEnabled;        }        // Otherwise, inherit from the global runtime cache setting.        return IsCachingEnabled;    }}
```
Aniruddh25: the line spacing seems to have been garbled in this file

Aniruddh25 (approved):
approved except for the spacing.

src/Config/ObjectModel/RuntimeConfig.cs
aaronburtle: should be fixed now

Aniruddh25 (approved):

Aniruddh25 (approved):

souvikghosh04 (approved):
LGTM

PR #3325, opened 2026-03-26
Omit redundant _NoAutoPK and _NoPK OpenAPI schemas when request-body-strict is false

### LINKED ISSUE #3260: [Bug]: when `"request-body-strict": false` OpenAPI returns "Entity_NoAutoPK" & "Entity_NoPK"
## What?

When DAB is not `strict` it no longer requires `NoAutoPk` or `NoPk` types in OpenAPI. 

 - This is misleading to tooling that doesn't realize these useless types are not required. 
 - This also multiplies the size of the OpenAPI document, 2x for every entity.

Image

## Configuration

```json
{
  "data-source": {
    "database-type": "mssql",
    "connection-string": "@env('MSSQL_CONNECTION_STRING')",
    "options": {
      "set-session-context": false
    }
  },
  "runtime": {
    "telemetry": {
      "log-level": {
        "Default": "information"
      }
    },
    "rest": {
      "enabled": true,
      "path": "/api",
      "request-body-strict": false // set to false
    },
    "graphql": {
      "enabled": true,
      "path": "/graphql",
      "allow-introspection": true
    },
    "mcp": {
      "enabled": true,
      "path": "/mcp"
    },
    "host": {
      "cors": {
        "origins": [],
        "allow-credentials": false
      },
      "authentication": {
        "provider": "Unauthenticated"
      },
      "mode": "development"
    }
  },
  "autoentities": {
    "default": {
      "patterns": {
        "include": [
          "%.%%"
        ]
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            {
              "action": "*"
            }
          ]
        }
      ]
    }
  },
  "entities": {}
}
```

## OpenAPI

```json
{
  "openapi": "3.0.1",
  "info": {
    "title": "Data API builder - REST Endpoint - anonymous",
    "version": "2.0.0"
  },

[snip]

  "components": {
    "schemas": {
      "Todo": {
        "type": "object",
        "properties": {
          "id": {
            "type": "number",
            "format": ""
          },
          "title": {
            "type": "string",
            "format": ""
          },
          "completed": {
            "type": "boolean",
            "format": ""
          }
        }
      },
      "Todo_NoAutoPK": { // should not exist
        "type": "object",
        "properties": {
          "title": {
            "type": "string",
            "format": ""
          },
          "completed": {
            "type": "boolean",
            "format": ""
          }
        }
      },
      "Todo_NoPK": { // should not exist
        "type": "object",
        "properties": {
          "title": {
            "type": "string",
            "format": ""
          },
          "completed": {
            "type": "boolean",
            "format": ""
          }
        }
      }
    }
  },
  "tags": [
    {
      "name": "Todo"
    }
  ]
}
```

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3260

## What is this change?

When request-body-strict is set to false, the OpenAPI document was still generating separate _NoAutoPK and _NoPK component schemas for each entity. These schemas only differ from the base entity schema by excluding primary key fields, a distinction that is meaningless when the request body already allows additional properties. This made the generated OpenAPI document unnecessarily verbose and confusing for consumers.

The `isRequestBodyStrict` flag is now threaded through `BuildPaths` and `CreateOperations` so that POST/PUT/PATCH request body schema references point to the base entity schema when strict mode is off. The `CreateComponentSchemas` method now gates `_NoAutoPK` and `_NoPK` schema generation on strict mode, and its XML documentation has been updated to accurately describe both strict and non-strict behavior.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

Updated existing unit test `RequestBodyStrict_False_OmitsRedundantSchemas` (renamed from `RequestBodyStrict_False_AllowsExtraFields`) to verify that when request-body-strict is false:
- `_NoAutoPK` and `_NoPK` component schemas are not generated.
- The base entity schema is still present.
- POST/PUT/PATCH operations reference the base entity schema (not `_NoAutoPK` or `_NoPK`).


src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -1342,7 +1347,7 @@ private Dictionary<string, OpenApiSchema> CreateComponentSchemas(RuntimeEntities
                     schemas.Add(entityName, CreateComponentSchema(entityName, fields: exposedColumnNames, metadataProvider, entities, isRequestBodySchema: false));
 
                     // Only generate request body schemas if mutation operations are available
-                    if (hasPostOperation || hasPutPatchOperation)
+                    if (isRequestBodyStrict && (hasPostOperation || hasPutPatchOperation))
                     {
                         // Create an entity's request body component schema excluding autogenerated primary keys.
```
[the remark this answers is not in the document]
aaronburtle: valid, addressing

aaronburtle:
@copilot can you please review again

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -466,7 +470,8 @@ private Dictionary<OperationType, OpenApiOperation> CreateOperations(
                     if (configuredRestOperations[OperationType.Patch])
                     {
                         OpenApiOperation patchOperation = CreateBaseOperation(description: PATCH_DESCRIPTION, tags: tags);
-                        patchOperation.RequestBody = CreateOpenApiRequestBodyPayload($"{entityName}_NoPK", requestBodyRequired);
+                        string patchSchemaRef = isRequestBodyStrict ? $"{entityName}_NoPK" : entityName;
```
anushakolan, line 473: Why do we want schemas of name `entityName`, why shouldn't we just skip creating the payload entirely when `isRequestBodyStrict` is `false`?
My point if `isRequestBodyStrict` is `false` then we will have 2 schemas of name `entityName` from both `Put` and `Post` operation right.

I am trying to understand the context better here, please correct me if I am wrong here.

Aniruddh25 (approved):

Aniruddh25:
Was quite difficult to review, but looks good. Noting my observations in case it helps for easier review. There is a difference between `isRequestBodyStrict` and `isRequestBodyRequired` flags. The former dictates whether there can be additional fields in the request body or not, the latter dictates whether the request body itself is required or not altogether. 

`CreateOperations` determines creates the OpenApiRequest payloads for all the operations and uses a reference name of the schema of the payload. It will use the simply the entire entity name if `isRequestBodyStrict` is false. `CreateComponentSchemas` will never create additional schemas suffixed with `_NoAutoPK` or `_NoPK`  in this case. 

If `isRequestBodyStrict` is true, `CreateComponentSchemas` will create those additional schemas but only either of _NoAutoPK or _NoPK will be referred in the `CreateOperations` function when creating the payload based on whether PK is in the uri component or not.

anushakolan (commented):
Waiting on the reply for my comment, will review after the comment is resolved.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
aaronburtle: We always create a single base schema per `entityName`, and what this line is doing is determining which schema name to reference. So, when `isRequestBodyStrict` is `false`, we will have references for PUT/PATCH/POST all to the same schema, not multiple schemas.

anushakolan: Thanks for clarifying.

anushakolan (approved):

PR #3402, opened 2026-03-30
Add read-only support for PostgreSQL array columns

bilby91:
## Summary
- Add read-only support for PostgreSQL array columns (`int[]`, `text[]`, `boolean[]`, `bigint[]`, etc.) which previously caused initialization failures
- Array columns are exposed as GraphQL list types (`[Int]`, `[String]`, etc.) and returned as JSON arrays in REST/MCP responses
- Array columns are marked read-only and excluded from create/update mutation inputs until write support is implemented

## Changes

**Core plumbing (database-agnostic):**
- `ColumnDefinition`: new `IsArrayType` and `ElementSystemType` properties
- `SqlMetadataProvider`: detect array types (`System.Array`) during schema introspection
- `SchemaConverter`: unwrap array element types and generate `ListTypeNode` for GraphQL fields
- `EdmModelBuilder`: represent array columns as `EdmCollectionTypeReference` in OData model
- `TypeHelper`: fallback handling for unresolved `System.Array` type

**PostgreSQL-specific:**
- `PostgreSqlMetadataProvider`: override `PopulateColumnDefinitionWithHasDefaultAndDbType` to resolve element types from `information_schema` `udt_name` (`_int4`→`int`, `_text`→`string`, etc.)

**Tests:**
- 14 new unit tests (array→ListTypeNode generation, nullability, AutoGenerated directive, byte[] exclusion)
- 3 new PostgreSQL e2e tests (query by PK, null arrays, multi-row queries)
- Test schema and config for `array_type_table` entity

## Limitations
- **Read-only**: mutations on array columns are blocked (marked as auto-generated/read-only)
- **No OData `$filter`**: filtering on array columns is not supported (no OData array operators)
- **PostgreSQL only**: other databases don't have native array types; generic plumbing is in place for future use
- **1D arrays only**: multi-dimensional arrays are not tested/supported

## Test plan
- [x] Unit tests pass (1325/1325, 0 failures)
- [x] PostgreSQL e2e tests pass (3/3 array tests, 711/711 non-preexisting PG tests)
- [x] CI pipeline validation (PostgreSQL Docker)


bilby91:
@aaronburtle Thanks for approving CI. I'll take a look at the failing tests and address them.

bilby91:
@aaronburtle I think fixes should be in place.

src/Core/Services/MetadataProviders/PostgreSqlMetadataProvider.cs
```diff
@@ -75,5 +77,71 @@ public override Type SqlToCLRType(string sqlType)
         {
             throw new NotImplementedException();
         }
+
+        /// <summary>
+        /// Maps PostgreSQL array udt_name prefixes to their CLR element types.
+        /// PostgreSQL array types in information_schema use udt_name with a leading underscore
+        /// (e.g., _int4 for int[], _text for text[]).
+        /// </summary>
+        private static readonly Dictionary<string, Type> _pgArrayUdtToElementType = new(StringComparer.OrdinalIgnoreCase)
```
aaronburtle, line 86: Should be able to add `_json`, `_jsonb`, and `_money` to this dictionary. Currently these types would work for requests, but will generate the wrong schema types

bilby91: Make sense. Will verify they work with some new tests.

aaronburtle:
Looks like we are missing tests on the REST side of things. Should add some simple tests to validate REST works with array types, eg. `GET /api/ArrayType` and `GET /api/ArrayType/id/1`

src/Service.GraphQLBuilder/Queries/InputTypeBuilder.cs
```diff
@@ -50,6 +50,12 @@ private static List<InputValueDefinitionNode> GenerateOrderByInputFieldsForBuilt
             List<InputValueDefinitionNode> inputFields = new();
             foreach (FieldDefinitionNode field in node.Fields)
             {
+                // Skip array/list type fields (e.g., PostgreSQL array columns) - they cannot be ordered.
+                if (field.Type.IsListType())
```
aaronburtle: CosmosDB has support for list types, so we can't just broadly skip them here. This needs to be scoped to SQL array columns only without breaking Cosmos list filtering.

aaronburtle: I believe this is why the CosmosDB pipeline is failing.

aaronburtle:
The PG_SQL Integration test failures are from mismatched configs, need to add the new section to `ConfigurationTests.TestReadingRuntimeConfigForPostgreSql.verified.txt`

bilby91:
I handle most of the stuff already. Need to replicate the Cosmos issue locally and get it fixed. Will update when I have that aspect ready.

bilby91:
@aaronburtle I think CI should pass now :)

aaronburtle:
Great progress on the fix @bilby91! The `IsListType() && IsBuiltInType()` approach fixed the Cosmos nested object array tests, but 2 Cosmos tests are still failing:

* `TestQueryFilterContains_WithStringArray`
* `TestQueryFilterNotContains_WithStringArray`

The issue is that `IsBuiltInType()` unwraps the list to check the inner type, so `[String]` (Cosmos tags field) returns true and gets skipped but Cosmos legitimately filters on `[String]` via `contains` or `notContains`.

I think the fix here is to use the `@autoGenerated` directive instead of `IsBuiltInType` as the discriminator. PostgreSQL array columns always carry this directive (because `IsReadOnly = true`), while Cosmos schema fields never do.

So in `InputTypeBuilder.cs`, if we replace both occurrences of:

```
if (field.Type.IsListType() && IsBuiltInType(field.Type))
{
    continue;
}
```

with

```
if (field.Type.IsListType() &&
    field.Directives.Any(d => d.Name.Value == AutoGeneratedDirectiveType.DirectiveName))
{
    continue;
}
```

I think that will solve the problem. Thanks again for working on this!

bilby91:
@aaronburtle Thanks a lot for you help and guidance during all the PR :)

I had identified the issue last night and forgot to push it. I had some issues with CosmoDb simulator crushing in my environment so I was not able to run the entire suite. I was able to confirm the relevant tests are passing.

Thanks again!

bilby91:
@aaronburtle All green :)

Thanks again.

aaronburtle:
> @aaronburtle All green :)
> 
> Thanks again.

Thanks for the quick iterations, taking another pass and will need another reviewer, hopefully will get this merged soon!

src/Core/Parsers/EdmModelBuilder.cs
```diff
@@ -125,6 +126,14 @@ SourceDefinition sourceDefinition
                             sqlMetadataProvider.TryGetExposedColumnName(entityAndDbObject.Key, column, out exposedColumnName!);
                             newEntity.AddKeys(newEntity.AddStructuralProperty(name: exposedColumnName, type, isNullable: false));
                         }
+                        else if (columnDef.IsArrayType)
+                        {
+                            // Array columns are represented as EDM collection types (e.g., Collection(Edm.Int32) for int[]).
+                            sqlMetadataProvider.TryGetExposedColumnName(entityAndDbObject.Key, column, out exposedColumnName!);
+                            EdmPrimitiveTypeReference elementTypeRef = new(EdmCoreModel.Instance.GetPrimitiveType(type), isNullable: true);
```
RubenCerna2079, line 134: Is there a reason why we want `isNullable` to be true?

src/Core/Parsers/EdmModelBuilder.cs
RubenCerna2079, line 119: I think we also need to update this comment in order to explain what happens when we are using an array type and how that differs from the way we add the entities to the edm model.

RubenCerna2079 (commented):
So I am half-way through the PR, will try to finish it by tomorrow

src/Core/Parsers/EdmModelBuilder.cs
bilby91: Updated the comment to mention the array column path.

src/Core/Parsers/EdmModelBuilder.cs
bilby91: isNullable: true is used because PostgreSQL arrays can contain NULL elements ('{1,NULL,3}'::int[] is valid). The element type reference should reflect this.

It's also consistent with how non-PK columns are handled in the same method — line 140 uses isNullable: true for regular structural properties. Only primary keys use isNullable: false.

I verified that changing to isNullable: false doesn't cause a runtime failure — DAB's OData pipeline doesn't enforce element-level nullability in collections at serialization time. So this is about correctness of the EDM model representation rather than a breaking behavior difference.

I added a test row (id=4) with NULL elements inside arrays ('{1,NULL,3}', '{hello,NULL,world}', etc.) and two new tests that query it:
 - QueryArrayColumnsWithNullElements (GraphQL)
 - GetArrayTypeWithNullElementsInsideArrays (REST)
 
Both pass with isNullable: true and also with false, confirming the pipeline handles both.

aaronburtle (approved):
LGTM! Thanks again for so diligently working on this one @bilby91 <3

bilby91:
@RubenCerna2079 Let me know if there are any pending changes that you want me to address! Looking forward to landing this feature!

RubenCerna2079:
> @RubenCerna2079 Let me know if there are any pending changes that you want me to address! Looking forward to landing this feature!

Hi @bilby91 I already finished with your PR. Everything looks good, but I first need to discuss something related to these changes with the team. Please wait a little longer.

aaronburtle (changes requested):
We handle pagination currently, by checking if the last item in the DB response is array type, and then if it is we know that is a set of pagination columns. But this was done when there were no array types in SQL. Now that we are going to support array types in SQL, need to double check that this wont break out pagination logic. Will look into this and get back to this PR.

bilby91:
@aaronburtle Happy to take a look into this. I'll add some coverage if missing.

RubenCerna2079:
> @aaronburtle Happy to take a look into this. I'll add some coverage if missing.

Thanks for offering, this is something that we can check on our own. I just created an issue related to it in case you want to check it. We will finish the review on your PR and then fix this a separate issue later on.
https://github.com/Azure/data-api-builder/issues/3462

RubenCerna2079 (approved):
LGTM! Thanks for taking the time to do this.

aaronburtle:
> @aaronburtle Happy to take a look into this. I'll add some coverage if missing.

Hi @bilby91 sorry for the delay I was out sick the last week or so, but I will be tackling the underlying issue and will check back once it is fixed, hopefully this week.

bilby91:
@aaronburtle Hope you are better!

Did you had a chance to look at this ? 

Thanks!!!

aaronburtle:
Thanks @bilby91 I am much better now. I've opened a PR that should accommodate collection types in SQL now.

Once this refactor merges in we can get your PR here accross the finish line

 https://github.com/Azure/data-api-builder/pull/3521
 
 Thanks for your patience.

bilby91:
@RubenCerna2079 Excited to see this moving forward :)

RubenCerna2079:
@bilby91 The PR that precedes this one was just merged. We should be able to merge this one soon enough.

bilby91:
@RubenCerna2079 Great! Yeah, I've been monitoring the other PR. Anxious to get this feature into our system.

Thanks

aaronburtle (approved):
Thanks for hanging in there while we tied up those loose ends. Looks great!

bilby91:
@aaronburtle @RubenCerna2079 Thanks so much for all the assistance! 

How does the release schedule look like ? New versions are released at a given cadance ?

Thanks

PR #3404, opened 2026-03-31
Fix memory leak: RuntimeConfigProvider._changeTokenRegistration never disposed

### LINKED ISSUE #3403: [Bug]: RuntimeConfigProvider._changeTokenRegistration is never disposed due to missing IDisposable interface and pre-created instance registration
### What happened?

`RuntimeConfigProvider` has a `Dispose()` method that cleans up its `_changeTokenRegistration` (a change token subscription), but this method is never called for two reasons:

Missing interface: `RuntimeConfigProvider` does not implement `IDisposable`, it only has a public `Dispose()` method. The DI container only auto-disposes services that implement `IDisposable` or `IAsyncDisposable`.

Pre-created instance registration: In `Startup.cs`, the provider is created before being registered with the container.

Even if `IDisposable` were added, the .NET DI container does not dispose pre-created instances; it only disposes instances it created itself (via type or factory registrations).

Need to implement `IDisposable` and switch to a factory registration or other means of getting the container to manage the lifecycle.

### Version

2.0

Azure SQL

_No response_

_No response_

### Relevant log output

```Text

```

### Code of Conduct

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3403

## What is this change?

`RuntimeConfigProvider` has a `Dispose()` method that cleans up its `_changeTokenRegistration` (a `ChangeToken.OnChange` subscription), but this method was never called due to two compounding issues:

* The class did not implement the `IDisposable` interface it only had a public `Dispose()` method. The DI container only auto-disposes services that implement `IDisposable` or `IAsyncDisposable`.

* It was registered as a pre-created instance via `services.AddSingleton(configProvider)`. The .NET DI container does not dispose pre-created instances; only instances it creates itself via type or factory registrations.

We also correctly register the loader in our tests using the same pattern (Loader was correctly registered in `Startup`).

## How was this tested?

Run against our test suite.


Aniruddh25 (approved):

src/Core/Configurations/RuntimeConfigProvider.cs
```diff
@@ -25,7 +25,7 @@ namespace Azure.DataApiBuilder.Core.Configurations;
 /// should not load the config directly, or maintain a reference to it, so that we can do hot-reloading by replacing
 /// the config that is available from this type.
 /// </remarks>
-public class RuntimeConfigProvider
+public class RuntimeConfigProvider : IDisposable
```
RubenCerna2079, line 28: I am just not following why we want to dispose of this object. Could you explain?

Aniruddh25: all objects need to be disposed after usage to avoid holding up memory.

RubenCerna2079 (approved):

PR #3408, opened 2026-04-02
Fix: remove .enabled suffix from individual DML tool configure option

### LINKED ISSUE #3373: [Bug]: [Critical] structurally wrong dab configure --runtime.mcp.dml-tools.<tool>.enabled
## Issue
CLI implies { "describe-entities": { "enabled": true } } but schema expects { "describe-entities": true }. Applies to all 7 individual DML tools.
> Each tool IS a direct boolean, no .enabled sub-property
### CLI
Each tool has .enabled suffix (e.g., dml-tools.describe-entities.enabled)
### Schema
runtime.mcp.dml-tools.describe-entities

souvikghosh04:
## Why make this change?
- Closes #3373 
 
dab configure individual DML tool options had an extra `.enabled` suffix, producing invalid nested config instead of the direct booleans the schema expects.

## What is this change?

Removed `.enabled` suffix from 7 individual DML tool names in ConfigureOptions.cs (e.g., `--runtime.mcp.dml-tools.describe-entities.enabled` is now `runtime.mcp.dml-tools.describe-entities`).
Bulk toggle (`--runtime.mcp.dml-tools.enabled`)

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests

## Sample Request(s)

Fixed: direct boolean, matches schema
`dab configure --runtime.mcp.dml-tools.describe-entities true`

Multiple tools
`dab configure --runtime.mcp.dml-tools.describe-entities true --runtime.mcp.dml-tools.create-record false`


src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -1081,6 +1081,96 @@ public void TestConfigureDescriptionForMcpSettings(string descriptionValue)
...
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(true, DisplayName = "Enable individual DML tool: describe-entities")]
+        [DataRow(false, DisplayName = "Disable individual DML tool: describe-entities")]
+        public void TestConfigureIndividualDmlToolForMcpSettings(bool updatedValue)
+        {
+            // Arrange
+            SetupFileSystemWithInitialConfig(INITIAL_CONFIG);
+
+            // Act: Set describe-entities via the corrected option name (no .enabled suffix)
```
Aniruddh25: The tests in this file do not explicitly check whether the `.enabled` is present or not in the option name. Rather, it is directly invoking the constructor. 

There are already be other tests that verify configuration via the constructor - like the E2E tests. please remove these tests as they are not necessary. Please do a self review of tests that are really necessary

Aniruddh25 (changes requested):

src/Cli.Tests/ConfigureOptionsTests.cs
souvikghosh04: Fixed

aaronburtle (approved):
LGTM!

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -1271,4 +1271,37 @@ public void TestUpdateDatabaseType(string dbType, bool isSuccess)
...
+        Assert.IsNotNull(runtimeConfig);
+
+        // Act: Run configure with the individual DML tool option through the full CLI parsing path.
+        string[] runtimeArgs = { "configure", "-c", TEST_RUNTIME_CONFIG_FILE, optionName, value };
+        int exitCode = Program.Execute(runtimeArgs, _cliLogger!, _fileSystem!, _runtimeConfigLoader!);
+
+        // Assert: Command succeeds and the config contains MCP DML tools section.
+        Assert.AreEqual(0, exitCode);
+        Assert.IsTrue(_runtimeConfigLoader!.TryLoadConfig(TEST_RUNTIME_CONFIG_FILE, out RuntimeConfig? updatedConfig));
+        Assert.IsNotNull(updatedConfig?.Runtime?.Mcp?.DmlTools);
```
Aniruddh25, line 1301: We should assert here that the correct "value" was indeed configured for the option name mapped to the right config property.

souvikghosh04: added actual and expected assertions for each property types

Aniruddh25 (approved):

PR #3419, opened 2026-04-08
Implemented MCP Set Log Level

### LINKED ISSUE #3274: [Bug]: when using `--mcp-stdio` failure with "logging/setlevel"
## What?

We get the error `Method not found: logging/setLevel` when our MCP Server (STDIO) starts.

### Incoming request to our MCP

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "logging/setLevel",
  "params": {
    "level": "info"
  }
}
```

### The problem

DAB already advertises the logging capability in its initialize response (see line 189), but the server's request dispatch loop has no handler for "logging/setLevel", so the client gets a JSON-RPC Method not found (-32601) error back.

https://github.com/Azure/data-api-builder/blob/9563a201cc816ab02d3a7bac32f72ea5bd8d29d2/src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs#L188-L199

### The solution

1. CLI `--LogLevel` — takes highest precedence (IsCliOverridden` flag in `DynamicLogLevelProvider`)
2. Config `runtime.host.logging.level` — second precedence
3. MCP `logging/setLevel` — lowest precedence, should only apply if CLI didn't override

Then, add a case in the switch and route it through the existing `DynamicLogLevelProvider`. 

> [!NOTE]
>  If `IsCliOverridden` is true, accept the request but silently ignore it — the client shouldn't get an error, but CLI wins.

## Why?

This is the **mcp.json** file in the `/.vscode` folder. 

```json
{
	"servers": {
		"dab-todo-test-mcp": {
			"type": "stdio",
			"command": "dab",
			"args": [
				"start",
				"--mcp-stdio",
				"role:anonymous",
				"--LogLevel",
				"none",
				"--config",
				"${workspaceFolder}/dab-config.json"
			]
		}
	},
	"inputs": []
}
```

When starting the MCP Server, the client passes in:

```json
```

But we do not support it, so we get:

```
Method not found: logging/setLevel
```

Image

### LINKED ISSUE #3275: [Bug]: When `-mcp-stdio` is used, control the output
## What?

> This is set as a bug because any developer using DAB would think this is a bug regardless of the details.

Standard output causes MCP clients to misread even `information` as a `warning`. 

> [!NOTE]
> This is important because we want to seamlessly snap into the AI workflow, not look like we are failing.

## Why?

The MCP client should be able to start DAB without looking like we have errored.

## How?

When `-mcp-stdio`, default `--LogLevel` flag to `none` but allow it to override if set explicitly. 

> [!IMPORTANT]
> We must ensure that `loglevel:none` in the command-line ACTUALLY emits ZERO CHARACTERS!

Image

anushakolan:
## Why make this change?

Closes #3274 - MCP Server returns "Method not found: logging/setLevel" error when clients send the standard MCP logging/setLevel request.

Closes #3275 - Control output in MCP stdio mode (default to `LogLevel.None`, redirect/suppress console output).

### MCP `logging/setLevel` Handler
- Added handler for `logging/setLevel` JSON-RPC method in `McpStdioServer.cs`
- Implemented `DynamicLogLevelProvider` with `ILogLevelController` interface to allow MCP to update log levels dynamically
- Added `IsCliOverridden` and `IsConfigOverridden` properties to enforce precedence rules

### Log Level Precedence System
**Precedence (highest to lowest):**
1. **CLI `--LogLevel` flag** - cannot be changed by MCP
2. **Config `runtime.telemetry.log-level`** - cannot be changed by MCP  
3. **MCP `logging/setLevel`** - only works if neither CLI nor config set a level
4.  Default (LogLevel.None for MCP stdio mode)

If CLI or config set a level, MCP requests are accepted but silently ignored (no error returned per MCP spec).

### Early Config Reading for MCP Mode
- Added `TryGetLogLevelFromConfig()` in `Program.cs` to read config file early (before host build)
- This ensures config log level is detected before Console redirect decision
- Console redirect for MCP stdio mode now respects config log level

### CLI Log Level Handling
- Added `Utils.CliLogLevel` property to track the parsed `--LogLevel` value
- CLI's `CustomLoggerProvider` now respects the `--LogLevel` value for its own logging

### Config Helpers
- Added `HasExplicitLogLevel()` helper to `RuntimeConfig` to correctly detect when config actually pins a log level
- This properly handles null values in telemetry section (null values don't count as explicit override)

## How was this tested?

- [x] Unit Tests (`DynamicLogLevelProviderTests` - 5 tests)
- [x] Manual Testing

### Manual Test 1: No override (MCP can change level)
1. Start MCP server without `--LogLevel` and without config `log-level`
2. MCP sends `logging/setLevel` with `level: info`
3. Result: Log level changes to info

### Manual Test 2: CLI override (MCP blocked)
1. Start MCP server with `--LogLevel Warning`
2. MCP sends `logging/setLevel` with `level: info`
3. Result: Log level stays at Warning, MCP request accepted silently

### Manual Test 3: Config override (MCP blocked)
1. Add `"telemetry": { "log-level": { "default": "Warning" } }` to config
2. Start MCP server without `--LogLevel`
3. MCP sends `logging/setLevel` with `level: info`
5. Result: Log level stays at Warning, MCP request accepted silently

### Manual Test 4: Config with null values (MCP can change level)
1. Add `"telemetry": { "log-level": { "default": null } }` to config
2. Start MCP server without `--LogLevel`
3. MCP sends `logging/setLevel` with `level: info`
4. Result: Log level changes to info (null values don't count as override)

## Sample Request(s)

MCP client sends:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "logging/setLevel",
  "params": {
    "level": "info"
  }
}
```

Server responds with empty result (success per MCP spec) and updates log level if no CLI/config override is active.


src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -228,6 +233,71 @@ private void HandleListTools(JsonElement? id)
...
+                WriteResult(id, new { });
+                return;
+            }
+
+            // Attempt to update the log level
+            // If CLI or Config overrode, this returns false but we still return success to the client
+            bool changed = logLevelController.UpdateFromMcp(level);
+            if (changed)
+            {
+                Console.Error.WriteLine($"[MCP DEBUG] Log level changed to: {level}");
```
Aniruddh25: Where are these debug messages output? Will those confuse the agent? 
Also, what if the LogLevel set by CLI is None. Do these debug messages still be shown? If not, how do you prevent them if you send them to Console.Error.WriteLine?

Aniruddh25: Do we really need these console messages?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -228,6 +233,71 @@ private void HandleListTools(JsonElement? id)
...
+            {
+                // Log level controller not available - still accept request per MCP spec
+                Console.Error.WriteLine("[MCP DEBUG] ILogLevelController not available, logging/setLevel ignored.");
+                WriteResult(id, new { });
+                return;
+            }
+
+            // Attempt to update the log level
+            // If CLI or Config overrode, this returns false but we still return success to the client
+            bool changed = logLevelController.UpdateFromMcp(level);
```
Aniruddh25: Even though the level is modified using the level controller, how does it take into effect in STDIO mode? HandleToolCallAsync seems to be writing to the console using WriteError function regardless of the logging level defined by the configuration/CLI/even the logging/setLevel call.

Aniruddh25 (changes requested):
Dont understand how the log level set by MCP tool call takes into effect. Can you show a working demo?

Aniruddh25 (changes requested):
>Manual Test 2: CLI override (MCP blocked)
>Start MCP server with --LogLevel None
>MCP sends logging/setLevel with level: info
>Result: Log level stays at None, debug message shows "Log level not changed (CLI override active)"

As far as I understand, there should be no debug message altogether given that LogLevel has been defined as None through CLI.
So, why does the description above say Result shows "Log level not changed"?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: No removed them

anushakolan:
> > Manual Test 2: CLI override (MCP blocked)
> > Start MCP server with --LogLevel None
> > MCP sends logging/setLevel with level: info
> > Result: Log level stays at None, debug message shows "Log level not changed (CLI override active)"
> 
> As far as I understand, there should be no debug message altogether given that LogLevel has been defined as None through CLI. So, why does the description above say Result shows "Log level not changed"?

Because it was writing directly to stderr. Those [[MCP DEBUG]] messages would still appear because they bypassed ILogger<T>filtering and wrote directly to stderr. That's why we removed them.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: The WriteError function is JSON-RPC protocol communication, not logging. When an MCP client sends an invalid request (wrong tool name, missing parameters, etc.), the server must respond with a JSON-RPC error so the client understands what went wrong. This is required by the MCP specification and must always work regardless of log level settings.

Log level only controls application logs (query execution, authentication, database operations, etc.) that go through the logging framework. Protocol responses are separate from logging - they're how the server talks to the client, and suppressing them would break the MCP communication.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
Aniruddh25: The resolution comment says - removed them, but I dont see the change pushed to remote yet. Reopening this comment so that this is not forgotten.

Please resolve a comment only "after" pushing to remote.

Aniruddh25:
> Note: This PR includes changes from https://github.com/Azure/data-api-builder/pull/3420 which has been merged into this branch.

PR #3420 has not been merged into this branch, has it?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -228,6 +233,71 @@ private void HandleListTools(JsonElement? id)
...
+        /// <summary>
+        /// Handles the "logging/setLevel" JSON-RPC method by updating the runtime log level.
+        /// </summary>
+        /// <param name="id">The request identifier extracted from the incoming JSON-RPC request.</param>
+        /// <param name="root">The root JSON element of the incoming JSON-RPC request.</param>
+        /// <remarks>
+        /// Log level precedence (highest to lowest):
+        /// 1. CLI --LogLevel flag - cannot be overridden
+        /// 2. Config runtime.telemetry.log-level - cannot be overridden by MCP
+        /// 3. MCP logging/setLevel - only works if neither CLI nor Config explicitly set a level
```
RubenCerna2079, line 261: nit: I would also add as a 4th log level precedence that the log level is set depending on whether DAB is started in `Production/Development` mode

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -228,6 +233,71 @@ private void HandleListTools(JsonElement? id)
...
+        /// 
+        /// If CLI or Config set the log level, this method accepts the request but silently ignores it.
```
RubenCerna2079, line 264: Wouldn't it be better to throw a warning that the request was ignored? I think it would be more beneficial to the user.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -228,6 +233,71 @@ private void HandleListTools(JsonElement? id)
...
+                WriteError(id, McpStdioJsonRpcErrorCodes.INVALID_PARAMS, "Missing or invalid 'level' parameter");
+                return;
+            }
+
+            // Get the ILogLevelController from service provider
+            ILogLevelController? logLevelController = _serviceProvider.GetService<ILogLevelController>();
+            if (logLevelController is null)
+            {
+                // Log level controller not available - still accept request per MCP spec
+                Console.Error.WriteLine("[MCP DEBUG] ILogLevelController not available, logging/setLevel ignored.");
```
RubenCerna2079: Just confirming, this is something that will need to be changed so it is buffered in my PR right?

RubenCerna2079: Just saw Ani's comment below, will this be removed as well?

src/Cli/ConfigGenerator.cs
```diff
@@ -2597,17 +2598,22 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
 
                 minimumLogLevel = (LogLevel)options.LogLevel;
                 _logger.LogInformation("Setting minimum LogLevel: {minimumLogLevel}.", minimumLogLevel);
+
+                // Only add --LogLevel when user explicitly specified it via CLI.
+                // This allows MCP logging/setLevel to work when no CLI override is present.
+                args.Add("--LogLevel");
```
RubenCerna2079, line 2741: I see that some of the changes in the PR #3420 are included here, I think it would be better to have it only in one, so it is not confusing.

src/Service/Program.cs
```diff
@@ -110,6 +111,7 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
                 .ConfigureServices((context, services) =>
                 {
                     services.AddSingleton(LogLevelProvider);
+                    services.AddSingleton<ILogLevelController>(LogLevelProvider);
```
RubenCerna2079, line 140: Why wouldn't we want only 1 signleton?

src/Service/Telemetry/DynamicLogLevelProvider.cs
```diff
@@ -1,17 +1,43 @@
...
 {
-    public class DynamicLogLevelProvider
+    /// <summary>
+    /// Provides dynamic log level control with support for CLI override, runtime config, and MCP.
+    /// </summary>
+    public class DynamicLogLevelProvider : ILogLevelController
     {
+        /// <summary>
+        /// Maps MCP log level strings to Microsoft.Extensions.Logging.LogLevel.
+        /// MCP levels: debug, info, notice, warning, error, critical, alert, emergency.
```
RubenCerna2079, line 16: Does MCP not have a `none` log level?

src/Service/Telemetry/DynamicLogLevelProvider.cs
```diff
@@ -20,7 +46,52 @@ public void UpdateFromRuntimeConfig(RuntimeConfig runtimeConfig)
...
             }
+
+            // If Config explicitly set the log level, accept the request but don't change anything.
+            // Config has second precedence after CLI.
+            if (IsConfigOverridden)
+            {
+                return false;
+            }
+
+            if (string.IsNullOrWhiteSpace(mcpLevel))
```
RubenCerna2079, line 84: From the way that you use this function in the `McpStdioServer.cs`, the `mcpLevel` shouldn't even have the possibility of being empty or null. I think it would be better to throw an error for this case.

src/Service/Telemetry/DynamicLogLevelProvider.cs
```diff
@@ -20,7 +46,52 @@ public void UpdateFromRuntimeConfig(RuntimeConfig runtimeConfig)
             if (!IsCliOverridden)
             {
                 CurrentLogLevel = runtimeConfig.GetConfiguredLogLevel();
+
+                // Track if config explicitly set a log level (not just using defaults)
+                IsConfigOverridden = !runtimeConfig.IsLogLevelNull();
+            }
```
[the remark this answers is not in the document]
anushakolan: Addressed. Changed to use HasExplicitLogLevel() which checks Values.Any(v => v.HasValue) - only returns true when at least one log level value is explicitly non-null.

src/Cli/ConfigGenerator.cs
```diff
@@ -2597,17 +2598,22 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
                 HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
 
                 _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
-            }
 
-            args.Add("--LogLevel");
-            args.Add(minimumLogLevel.ToString());
+                // Don't add --LogLevel arg since user didn't explicitly set it.
+                // Service will determine default log level based on config or host mode.
+            }
```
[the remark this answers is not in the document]
anushakolan: Addressed for MCP mode - we now read config early via TryGetLogLevelFromConfig() before host build (lines 213-221). For normal mode, we intentionally default to Error initially for backward compatibility; UpdateFromRuntimeConfig() adjusts later based on host mode.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Pushed the code.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Added a 4th item, but note that in MCP stdio mode the default is None (silent) when nothing else sets a level - this keeps stdout clean for JSON-RPC protocol. The "Debug for Development / Error for Production" default applies only to non-MCP mode.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: This is also removed.

src/Cli/ConfigGenerator.cs
anushakolan: I merged the PR into this PR, as there is a lot of common code and I found it quite difficult to separate them into 2 PRs.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Per MCP specification, logging/setLevel should return an empty success result (no error).

The challenge with logging a warning:

If CLI set --LogLevel None, we explicitly promised zero output - logging a warning would violate that
Direct Console.Error.WriteLine bypasses ILogger filtering, so the warning would appear even when the user explicitly disabled logging
In the follow-up notifications PR, we could potentially send a notifications/message to inform the client that its request was ignored due to CLI/Config override - but only when notifications are enabled.

src/Service/Program.cs
anushakolan: `Startup.cs` calls `UpdateFromRuntimeConfig(RuntimeConfig)` which is intentionally **not** on the `ILogLevelController` interface. The interface is kept minimal (only `UpdateFromMcp()`) to avoid coupling the MCP project to `RuntimeConfig`.

| Consumer | Needs | Registration Used |
|----------|-------|-------------------|
| `Startup.cs` | `UpdateFromRuntimeConfig()` | `DynamicLogLevelProvider` (concrete) |
| `McpStdioServer` | `UpdateFromMcp()` | `ILogLevelController` (interface) |

**Alternative: Add `UpdateFromRuntimeConfig()` to interface**
- Would require `ILogLevelController` (in Core) to depend on `RuntimeConfig` (in Config)
- Increases coupling, may cause circular dependency issues
- Interface becomes less focused

**Current approach:**
- Same singleton, two "doors" to access it - standard DI pattern
- Interface stays minimal and focused on MCP needs
- No changes to existing `Startup.cs` code
- Slight duplication (one extra line)

This is a common pattern when retrofitting an interface onto existing code for better testability and loose coupling.

src/Service/Telemetry/DynamicLogLevelProvider.cs
anushakolan: No, MCP specification doesn't define a "none" log level.

src/Service/Telemetry/DynamicLogLevelProvider.cs
anushakolan: Good observation. However, throwing an exception here would actually make things worse:

McpStdioServer already validates the level and returns INVALID_PARAMS error (the correct MCP error code) before calling UpdateFromMcp()
If UpdateFromMcp() throws an ArgumentException, it gets caught by the generic catch(Exception) in the request loop, which returns INTERNAL_ERROR - the wrong error code
The defensive check is a safety net for any future callers that might not validate. Silently returning false is consistent with how we handle other "can't change" cases (CLI override, Config override) - we just return false without failure.

The current flow is: McpStdioServer validates → returns proper error if invalid → calls UpdateFromMcp() with valid level. The null/empty check should never be triggered in practice.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -228,6 +249,85 @@ private void HandleListTools(JsonElement? id)
...
+
+            // If MCP successfully changed the log level to something other than "none",
+            // ensure Console.Error is pointing to the real stderr (not TextWriter.Null).
+            // This handles the case where MCP stdio mode started with LogLevel.None (quiet startup)
+            // and the client later enables logging via logging/setLevel.
+            bool isLoggingEnabled = !string.Equals(level, "none", StringComparison.OrdinalIgnoreCase);
+            if (updated && isLoggingEnabled)
+            {
+                RestoreStderrIfNeeded();
+            }
```
RubenCerna2079, line 309: Wouldn't we also need to have a function that disables the Stderr in case the user sets the log level to `none` again after setting it to something else?

src/Cli/CustomLoggerProvider.cs
```diff
@@ -71,10 +76,37 @@ public class CustomConsoleLogger : ILogger
...
+                {
+                    return;
+                }
+
+                if (!_logLevelToAbbreviation.TryGetValue(logLevel, out string? mcpAbbreviation))
+                {
+                    return;
+                }
+
+                Console.Error.WriteLine($"{mcpAbbreviation}: {formatter(state, exception)}");
```
RubenCerna2079, line 107: I think it would be good to add a comment explaining why we want to add all logs to the Console.Error. Just so we don't forget in the future.

src/Cli/Program.cs
```diff
@@ -29,6 +29,14 @@ public static int Main(string[] args)
             // Check if MCP stdio mode is requested - suppress CLI logging to keep stdout clean for JSON-RPC.
             Utils.IsMcpStdioMode = args.Any(a => string.Equals(a, "--mcp-stdio", StringComparison.OrdinalIgnoreCase));
 
+            // Check if user explicitly set --LogLevel (allows logs to stderr even in MCP mode)
+            int logLevelIndex = Array.FindIndex(args, a => string.Equals(a, "--LogLevel", StringComparison.OrdinalIgnoreCase));
```
Aniruddh25: Instead of searching for LogLevel argument, can we not set this when it is parsed by `ParseArguments` ?

src/Service/Program.cs
```diff
@@ -138,10 +141,19 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
                 })
                 .ConfigureLogging(logging =>
                 {
-                    // Set minimum level at the framework level - this affects all loggers.
-                    // For MCP stdio mode, Console.Out is redirected to stderr in Main(),
-                    // so any logging output goes to stderr and doesn't pollute the JSON-RPC channel.
-                    logging.SetMinimumLevel(LogLevelProvider.CurrentLogLevel);
+                    // For MCP stdio mode, we need dynamic log level control via logging/setLevel.
+                    // Set framework minimum to Trace so all logs pass through to the dynamic filter.
```
Aniruddh25, line 145: Why is the framework minimum required to be Trace explicitly for mcp stdio mode? The pass through to dynamic filter happens for non-stdio mode as well.

src/Service/Program.cs
```diff
@@ -160,61 +172,58 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
...
             {
-                // MCP stdio mode without explicit --LogLevel: default to None to keep stdout clean.
-                // This is NOT a CLI or config override, so MCP logging/setLevel can still change it.
-                logLevel = LogLevel.None;
+                // MCP stdio mode without explicit --LogLevel: check config for log level (second priority)
                 isLogLevelOverridenByCli = false;
-                isLogLevelOverridenByConfig = false;
+                logLevel = LogLevel.None; // Default if config doesn't have log level
+
+                // Find --config or --ConfigFileName argument
```
Aniruddh25: config argument may not be passed when you start dab. Then it defaults to dab-config.json 

Why are we reading this early? I know it is needed to identify if the logLevel is overriden by config but cant we delay this determination until config is read completely? Do we need to know before?

src/Service/Program.cs
```diff
@@ -229,6 +238,42 @@ private static LogLevel GetLogLevelFromCommandLineArgs(string[] args, bool runMc
...
+            logLevel = LogLevel.None;
+            try
+            {
+                if (!File.Exists(configFilePath))
+                {
+                    return false;
+                }
+
+                string configJson = File.ReadAllText(configFilePath);
+                if (RuntimeConfigLoader.TryParseConfig(configJson, out RuntimeConfig? config) && config is not null)
```
Aniruddh25, line 261: this is inefficient - because we are now parsing the config twice!!

Cant we use RuntimeConfigProvider.GetConfig? That way even if we need to parse EARLY, we can cache it and dont need to parse again.

Aniruddh25 (changes requested):
Some inefficiencies exist in parsing config multiple times.

src/Cli/CustomLoggerProvider.cs
```diff
@@ -18,8 +18,13 @@ public ILogger CreateLogger(string categoryName)
...
     public class CustomConsoleLogger : ILogger
     {
-        // Minimum LogLevel. LogLevel below this would be disabled.
-        private readonly LogLevel _minimumLogLevel = LogLevel.Information;
+        // Minimum LogLevel for CLI output.
+        // For MCP mode: use CLI's --LogLevel if specified, otherwise suppress all.
+        // For non-MCP mode: always use Information.
+        // Note: --LogLevel is meant for the ENGINE's log level, not CLI's output.
+        private static LogLevel MinimumLogLevel => Cli.Utils.IsMcpStdioMode
+            ? (Cli.Utils.IsLogLevelOverriddenByCli ? Cli.Utils.CliLogLevel : LogLevel.None)
```
Aniruddh25, line 26: why is `IsConfigOverriden`/log level from config not being used here to set the MinimumLogLevel?

src/Cli/Program.cs
anushakolan: Good suggestion! Unfortunately, we can't defer this to ParseArguments due to a timing constraint. Here's why:

Execution order in Main():

`ParseEarlyFlags(args)` - sets IsMcpStdioMode, IsLogLevelOverriddenByCli
`Utils.LoggerFactoryForCli` - creates CustomLoggerProvider
`loggerFactory.CreateLogger<Program>()` - creates cliLogger
`Execute(args, cliLogger, ...)`
`parser.ParseArguments<StartOptions, ...>(args)` - runs inside Execute

**The problem:** CustomLoggerProvider.MinimumLogLevel needs to know` IsMcpStdioMode` and `IsLogLevelOverriddenByCli` at step 2-3 to decide LogLevel in MCP and Non-MCP modes.

But `ParseArguments` runs at step 5, which requires cliLogger as a parameter. So it's a chicken-and-egg problem: logger needs the flags → ParseArguments needs the logger.

**What I've done instead:** Combined the two separate array searches (args.Any() for --mcp-stdio and Array.FindIndex() for --LogLevel) into a single pass through the args array in ParseEarlyFlags(), addressing the efficiency concern while respecting the timing constraint.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: Good point! However, we don't need to re-disable stderr when the user sets the level back to None. Here's why:

The stderr redirect to TextWriter.Null is specifically for quiet startup - it prevents noisy startup logs from appearing during MCP protocol handshake. Once the engine is running, log filtering is handled by MinimumLogLevel in the logger itself.

When user sets level to None:

MinimumLogLevel = LogLevel.None filters out all log calls
No logs reach Console.Error.WriteLine() even with stderr restored
Re-disabling stderr would add complexity without benefit since the log level filtering already prevents any output.

src/Service/Program.cs
anushakolan: In non-MCP mode, the log level is set once at startup and never changes, so `SetMinimumLevel(actualLevel)` is efficient - it blocks unwanted logs early in the pipeline.

In MCP stdio mode, the log level can change at runtime via `logging/setLevel`. The issue is that SetMinimumLevel() sets a hard floor in the logging framework - logs below that level are rejected before reaching our `AddFilter()`.

Example: If we set `SetMinimumLevel(LogLevel.Warning)` at startup, and the MCP client later requests `LogLevel.Debu`g, those Debug logs would be blocked by the framework before our dynamic filter even sees them.

By setting the framework minimum to Trace in MCP mode, we allow all logs to pass through the framework, and our `AddFilter(logLevel => LogLevelProvider.ShouldLog(logLevel))` does the actual runtime filtering based on the current dynamic level.

src/Service/Program.cs
anushakolan: In MCP stdio mode, stdout must be kept clean for JSON-RPC protocol messages. Before the host build starts, we need to decide:

LogLevel.None → redirect both stdout and stderr to TextWriter.Null (silent startup)
Any other level → redirect stdout to stderr (logs go to stderr, JSON-RPC uses raw stdout)

This Console redirection happens at lines 75-88, before CreateHostBuilder().Build():

If we wait until the config is fully loaded during host.Build(), the host building process itself produces logs that would go to stdout (before we've redirected), corrupting the MCP JSON-RPC channel.
```
if (runMcpStdio)
{
    if (initialLogLevel == LogLevel.None)
    {
        Console.SetOut(TextWriter.Null);
        Console.SetError(TextWriter.Null);
    }
    else
    {
        Console.SetOut(Console.Error);
    }
}

IHost host = CreateHostBuilder(args, runMcpStdio, mcpRole).Build();  // Host build produces logs
```

anushakolan: Regarding default dab-config.json:

Good catch! Fixed - now when --config is not specified; we fall back to `FileSystemRuntimeConfigLoader.DEFAULT_CONFIG_FILE_NAME (dab-config.json)` instead of null:

src/Service/Program.cs
anushakolan: The challenge is timing - `RuntimeConfigProvider` is a DI service created during host.Build(), but we need the log level before that for Console redirection in MCP stdio mode.

Timeline:

`StartEngine()` → need log level for Console.SetOut/SetError() (no DI yet)
`CreateHostBuilder().Build()` → RuntimeConfigProvider created via DI
`RuntimeConfigProvider.GetConfig()` → available, but Console redirect already happened
Options:

1. **Cache approach:** Store the early-parsed `RuntimeConfig` in a static field, have `FileSystemRuntimeConfigLoader` check this cache before re-parsing. This avoids double-parse but adds complexity.

2. **Accept double-parse:** The early parse only happens in MCP stdio mode, and JSON parsing is relatively fast. The trade-off is simplicity vs. a small performance cost on startup.

The early parse adds additional startup time for MCP stdio mode. However, 

1. This is a one-time cost at startup, not per-request
2. Only affects MCP stdio mode
3. Caching would require handling hot-reload edge cases (if config changes between early parse and host build, cache becomes stale)
4. Adds cross-assembly coupling between Program.cs and FileSystemRuntimeConfigLoader

Given the trade-off of an additional startup time vs. caching complexity with hot-reload risks, I recommend keeping the simpler double-parse approach.

src/Cli/CustomLoggerProvider.cs
anushakolan: `CustomLoggerProvider` is the CLI's logger (in Cli), separate from the engine's logging. It doesn't need `IsConfigOverridden` for two reasons:

1. **For dab start --mcp-stdio:** The CLI invokes the engine (Program.cs), which handles all logging setup. The engine reads config early via` TryGetLogLevelFromConfig()` and sets up` DynamicLogLevelProvider` (which has IsConfigOverridden). All log output goes through the engine's logging, not `CustomLoggerProvider`.

We verified this works - config with "log-level": { "default": "Warning" } and no --LogLevel flag showed only warn: logs, confirming config's level was honored by the engine.

2. **For other CLI commands (dab init, dab add, etc.):** These don't launch the engine and don't read runtime.telemetry.log-level - they just use `CustomLoggerProvider` with `MinimumLogLevel = Information`. Adding config parsing here would duplicate logic between CLI and engine.

In summary: `CustomLoggerProvider's` MCP mode setting (LogLevel.None when no CLI flag) is a safeguard for any early output before the engine takes over. Once the engine starts, it handles all logging with proper CLI > Config > MCP precedence.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
RubenCerna2079: Then why should we even need to have the Stderr disabled in the first place?

Aniruddh25 (approved):

src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            // Act
+            bool result = provider.UpdateFromMcp("debug");
+
+            // Assert
+            Assert.IsFalse(result);
+            Assert.AreEqual(LogLevel.Warning, provider.CurrentLogLevel);
+        }
+
+        [TestMethod]
+        public void UpdateFromMcp_InvalidLevel_ReturnsFalse()
```
RubenCerna2079: I think some of these tests are similar enough that you can create a single DataTestMethod for them

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
Aniruddh25: I believe we still need to disable it at startup because the log level filtering via MinimumLogLevel is not in effect at that point in time. 

Once the engine is up and can accept a change to log level using logging/setLevel API, this filtering is in place.

src/Service/Program.cs
```diff
@@ -62,6 +65,30 @@ public static bool StartEngine(string[] args, bool runMcpStdio, string? mcpRole)
         {
             try
             {
+                // Initialize log level EARLY, before building the host.
+                // This ensures logging filters are effective during the entire host build process.
+                // For MCP mode, we also read the config file early to check for log level override.
+                LogLevel initialLogLevel = GetLogLevelFromCommandLineArgs(args, runMcpStdio, out bool isCliOverridden, out bool isConfigOverridden);
```
Aniruddh25: nit: since this function is not getting the log level only from command line args, I would change its name to `GetLogLevelFromCommandLineArgsOrConfig`

wouldnt hold off my approval for this change though. you can fix it in the notifications PR if this is good to merge.

Aniruddh25 (approved):
Thanks for following through this and getting it merged in!

RubenCerna2079 (approved):
LGTM! This is a great change!

PR #3420, opened 2026-04-08
Set default log level to None for MCP in Stdio mode.

### LINKED ISSUE #3275: [Bug]: When `-mcp-stdio` is used, control the output

### LINKED ISSUE #3427: [Bug] --mcp-stdio mode writes startup log lines to stdout, corrupting the MCP JSON stream
## Summary

When DAB is started with `--mcp-stdio`, it writes human-readable startup log lines directly to **stdout** before the first MCP JSON message. MCP stdio transport expects only newline-delimited JSON on stdout, so any client that reads the stream immediately (Claude Code, VS Code MCP extension, etc.) receives non-JSON bytes and fails to parse the stream.

## Environment

- **DAB version:** 1.7.92
- **OS:** Windows 11 Enterprise
- **Transport:** `--mcp-stdio`
- **MCP client:** Claude Code 2.1.98

## Steps to Reproduce

```bash
dab start --config dab-config.json --mcp-stdio 2>/dev/null | head -5
```

The first bytes on stdout are:

```
Information: Microsoft.DataApiBuilder 1.7.92
Information: User provided config file: dab-config.json
Information: Setting default minimum LogLevel: Debug for Development mode.
...
{"jsonrpc":"2.0","id":0,"result":{"protocolVersion":"2025-06-18",...}}
```

Note: 0 bytes are written to stderr. All output — logs and MCP JSON alike — goes to stdout.

## Expected Behavior

In `--mcp-stdio` mode, **stdout should contain only MCP JSON messages** (newline-delimited). Startup logs should go to **stderr** so they don't corrupt the protocol stream.

## Actual Behavior

Startup log lines are interleaved with MCP JSON on stdout. Any MCP client that begins parsing stdout immediately (as required by the spec) receives non-JSON bytes and fails.

## Impact

MCP clients cannot connect to DAB via `--mcp-stdio` without a filtering proxy that strips the log noise. The `--LogLevel None` flag causes DAB to exit with code `4294967295` (does not suppress the pre-initialization lines anyway).

## Workaround

A Node.js wrapper that buffers stdout and discards lines that don't start with `{` before forwarding to the MCP client.

## Suggested Fix

In `--mcp-stdio` mode, redirect all ASP.NET Core / DAB log output to **stderr** so that **stdout carries only MCP protocol messages**.

anushakolan:
# PR Description

## Why make this change?

Closes #3275 - When `--mcp-stdio` is used, control the output

When DAB runs as an MCP server with `--mcp-stdio`, stdout must be reserved exclusively for JSON-RPC protocol messages. Any log output (even informational) causes MCP clients like Claude Desktop to interpret them as protocol violations, displaying warnings like "Discovered 7 tools" mixed with actual tool discovery results.

Related: #3274 - MCP `logging/setLevel` method not found

## What is this change?

This PR implements proper output control for MCP stdio mode:

1. **Default log level is `None`** when `--mcp-stdio` is used (no CLI or config override)
   - Ensures zero characters are emitted to stdout/stderr by default
   - No log noise interfering with MCP protocol

2. **Output redirection based on log level**:
   - `LogLevel.None`: Both stdout and stderr → `TextWriter.Null` (zero output)
   - Other levels: stdout → stderr (logs appear on stderr, JSON-RPC stays on stdout)

3. **Log level precedence**: CLI `--LogLevel` > Config `runtime.telemetry.log-level` > MCP `logging/setLevel`
   - CLI passes `--LogLevelFromConfig` flag when config specifies a log level
   - Service correctly identifies the source and blocks MCP override if appropriate

4. **Early logging suppression** in CLI before Service starts
   - Added `IsMcpStdioMode` flag to suppress CLI's own logging in MCP mode

## How was this tested?

- [x] Manual Tests
  - Configured `mcp.json` to run DAB with `--mcp-stdio` 
  - Verified zero output when no log-level specified (default `LogLevel.None`)
  - Verified config override by adding `"log-level": { "default": "Error" }` to `runtime.telemetry` - logs appeared on stderr
  - Verified CLI override (`--LogLevel`) takes precedence over config
- [x] Unit Tests
  - `DynamicLogLevelProviderTests` (5 tests) - CLI > Config > MCP precedence
  - `StartCommandTests` (39 tests) - CLI start command functionality

### CLI Usage - Default (Zero Output)
```bash
dab start --mcp-stdio --config dab-config.json
```
With no `log-level` in config, MCP mode produces zero log output.

### CLI Usage - Config Override
Config file (`dab-config.json`):
```json
{
  "runtime": {
    "telemetry": {
      "log-level": {
        "default": "Warning"
      }
    }
  }
}
```
```bash
dab start --mcp-stdio --config dab-config.json
```
Logs at Warning level appear on stderr; MCP `logging/setLevel` requests are accepted but ignored.

### CLI Usage - CLI Override (Highest Priority)
```bash
dab start --mcp-stdio --LogLevel Error --config dab-config.json
```
Logs at Error level appear on stderr; both config and MCP `logging/setLevel` are ignored.

### MCP logging/setLevel Request
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "logging/setLevel",
  "params": {
    "level": "info"
  }
}
```
Only changes log level if neither CLI nor config specified a level.

### Note
This PR is currently based on `dev/anushakolan/set-log-level` as it contains dependent changes. The base branch will be updated to main once `dev/anushakolan/set-log-level` is merged into main.

### DIFF
--- a/src/Cli/ConfigGenerator.cs
+++ b/src/Cli/ConfigGenerator.cs
@@ -2582,7 +2582,8 @@ namespace Cli
             List<string> args = new()
             { "--ConfigFileName", runtimeConfigFile };
 
-            /// Add arguments for LogLevel. Checks if LogLevel is overridden with option `--LogLevel`.
+            /// Add arguments for LogLevel. Only pass --LogLevel when user explicitly specified it,
+            /// so that MCP logging/setLevel can still adjust the level when no CLI override is present.
             /// If not provided, Default minimum LogLevel is Debug for Development mode and Error for Production mode.
             LogLevel minimumLogLevel;
             if (options.LogLevel is not null)
@@ -2597,6 +2598,11 @@ namespace Cli
 
                 minimumLogLevel = (LogLevel)options.LogLevel;
                 _logger.LogInformation("Setting minimum LogLevel: {minimumLogLevel}.", minimumLogLevel);
+
+                // Only add --LogLevel when user explicitly specified it via CLI.
+                // This allows MCP logging/setLevel to work when no CLI override is present.
+                args.Add("--LogLevel");
+                args.Add(minimumLogLevel.ToString());
             }
             else
             {
@@ -2604,10 +2610,10 @@ namespace Cli
                 HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
 
                 _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
-            }
 
-            args.Add("--LogLevel");
-            args.Add(minimumLogLevel.ToString());
+                // Don't add --LogLevel arg since user didn't explicitly set it.
+                // Service will determine default log level based on config or host mode.
+            }
 
             // This will add args to disable automatic redirects to https if specified by user
             if (options.IsHttpsRedirectionDisabled)
--- a/src/Cli/ConfigGenerator.cs
+++ b/src/Cli/ConfigGenerator.cs
@@ -123,18 +123,18 @@ namespace Cli
 
             bool isMultipleCreateEnabledForGraphQL;
 
-            // Multiple mutation operations are applicable only for MSSQL database. When the option --graphql.multiple-create.enabled is specified for other database types,
+            // Multiple mutation operations are applicable only for MSSQL database. When the option --graphql.multiple-mutations.create.enabled is specified for other database types,
             // a warning is logged.
             // When multiple mutation operations are extended for other database types, this option should be honored.
             // Tracked by issue #2001: https://github.com/Azure/data-api-builder/issues/2001.
             if (dbType is not DatabaseType.MSSQL && options.MultipleCreateOperationEnabled is not CliBool.None)
             {
-                _logger.LogWarning($"The option --graphql.multiple-create.enabled is not supported for the {dbType.ToString()} database type and will not be honored.");
+                _logger.LogWarning($"The option --graphql.multiple-mutations.create.enabled is not supported for the {dbType.ToString()} database type and will not be honored.");
             }
 
             MultipleMutationOptions? multipleMutationOptions = null;
 
-            // Multiple mutation operations are applicable only for MSSQL database. When the option --graphql.multiple-create.enabled is specified for other database types,
+            // Multiple mutation operations are applicable only for MSSQL database. When the option --graphql.multiple-mutations.create.enabled is specified for other database types,
             // it is not honored.
             if (dbType is DatabaseType.MSSQL && options.MultipleCreateOperationEnabled is not CliBool.None)
             {
@@ -347,7 +347,7 @@ namespace Cli
         /// <summary>
         /// Helper method to determine if the multiple create operation is enabled or not based on the inputs from dab init command.
         /// </summary>
-        /// <param name="multipleCreateEnabledOptionValue">Input value for --graphql.multiple-create.enabled option of the init command</param>
+        /// <param name="multipleCreateEnabledOptionValue">Input value for --graphql.multiple-mutations.create.enabled option of the init command</param>
         /// <returns>True/False</returns>
         private static bool IsMultipleCreateOperationEnabled(CliBool multipleCreateEnabledOptionValue)
         {
--- a/src/Service/Program.cs
+++ b/src/Service/Program.cs
@@ -9,6 +9,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Azure.DataApiBuilder.Config;
+using Azure.DataApiBuilder.Core.Telemetry;
 using Azure.DataApiBuilder.Service.Exceptions;
 using Azure.DataApiBuilder.Service.Telemetry;
 using Azure.DataApiBuilder.Service.Utilities;
@@ -110,6 +111,7 @@ namespace Azure.DataApiBuilder.Service
                 .ConfigureServices((context, services) =>
                 {
                     services.AddSingleton(LogLevelProvider);
+                    services.AddSingleton<ILogLevelController>(LogLevelProvider);
                 })
                 .ConfigureLogging(logging =>
                 {
--- a/config-generators/mssql-commands.txt
+++ b/config-generators/mssql-commands.txt
@@ -1,4 +1,4 @@
-init --config "dab-config.MsSql.json" --database-type mssql --set-session-context true --connection-string "Server=tcp:127.0.0.1,1433;Persist Security Info=False;User ID=sa;Password=REPLACEME;MultipleActiveResultSets=False;Connection Timeout=5;" --host-mode Development --cors-origin "http://localhost:5000" --graphql.multiple-create.enabled true --rest.request-body-strict true --auth.provider AppService
+init --config "dab-config.MsSql.json" --database-type mssql --set-session-context true --connection-string "Server=tcp:127.0.0.1,1433;Persist Security Info=False;User ID=sa;Password=REPLACEME;MultipleActiveResultSets=False;Connection Timeout=5;" --host-mode Development --cors-origin "http://localhost:5000" --graphql.multiple-mutations.create.enabled true --rest.request-body-strict true --auth.provider AppService
 add Publisher --config "dab-config.MsSql.json" --source publishers --permissions "anonymous:read"
 add Publisher_MM --config "dab-config.MsSql.json" --source publishers_mm --graphql "Publisher_MM:Publishers_MM" --permissions "anonymous:*"
 add Stock --config "dab-config.MsSql.json" --source stocks --permissions "anonymous:create,read,update,delete"
--- a/schemas/dab.draft.schema.json
+++ b/schemas/dab.draft.schema.json
@@ -280,6 +280,10 @@
               "description": "Allow enabling/disabling MCP requests for all entities.",
               "default": true
             },
+            "description": {
+              "type": "string",
+              "description": "Description of the MCP server, exposed as the 'instructions' field in the MCP initialize response to provide behavioral context to MCP clients and agents."
+            },
             "dml-tools": {
               "description": "Configuration for MCP Data Manipulation Language (DML) tools. Set to true/false to enable/disable all tools, or use an object to configure individual tools.",
               "oneOf": [
--- a/src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
@@ -6,6 +6,7 @@ using System.Text.Json;
 using Azure.DataApiBuilder.Config.ObjectModel;
 using Azure.DataApiBuilder.Core.AuthenticationHelpers.AuthenticationSimulator;
 using Azure.DataApiBuilder.Core.Configurations;
+using Azure.DataApiBuilder.Core.Telemetry;
 using Azure.DataApiBuilder.Mcp.Model;
 using Azure.DataApiBuilder.Mcp.Utils;
 using Microsoft.AspNetCore.Http;
@@ -131,6 +132,10 @@ namespace Azure.DataApiBuilder.Mcp.Core
                                 WriteResult(id, new { ok = true });
                                 break;
 
+                            case "logging/setLevel":
+                                HandleSetLogLevel(id, root);
+                                break;
+
                             case "shutdown":
                                 WriteResult(id, new { ok = true });
                                 return;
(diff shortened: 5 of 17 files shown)


src/Cli/ConfigGenerator.cs
```diff
@@ -2606,13 +2606,38 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
             }
             else
             {
-                minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
-                HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
+                // When --mcp-stdio is used without explicit --LogLevel:
+                // 1. Check if config has log-level set - use that (Config has priority 2)
+                // 2. Otherwise default to None for clean MCP stdio output
```
RubenCerna2079, line 2611: Should also add a comment that says that if we are not using --mcp-stdio it will default to Error for Production mode and Debug for Development mode

src/Cli/ConfigGenerator.cs
```diff
@@ -2606,13 +2606,38 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
+                if (options.McpStdio)
+                {
+                    // Check if config explicitly sets a log level
+                    if (!deserializedRuntimeConfig.IsLogLevelNull())
```
RubenCerna2079, line 2615: There is a chance that if we have some LogLevel being set for a specific namespace and this would cause this section to be run even if there is no `default` value being used. You would need to check specifically if we are using the `default` value inside the config file.

RubenCerna2079: Also, might be easier to include the logic to check if we need to use LogLevel none inside the `GetConfiguredLogLevel` function, since the way it works is that it first checks if we use the config file, and then defaults to the other LogLevels depending on the `mode` dab is started in.

src/Cli/ConfigGenerator.cs
```diff
@@ -2606,13 +2606,38 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
+                    {
+                        minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
+                        _logger.LogInformation("MCP stdio mode: Using config log-level: {minimumLogLevel}.", minimumLogLevel);
+                        // Pass --LogLevel to Service with special marker to indicate it's from config, not CLI.
+                        // This allows MCP logging/setLevel to be blocked by config override.
+                        args.Add("--LogLevel");
```
RubenCerna2079, line 2621: We don't want to use the `--LogLevel` flag here, it is causing a bug that I have a PR for https://github.com/Azure/data-api-builder/pull/3426

src/Cli/CustomLoggerProvider.cs
```diff
@@ -71,9 +71,16 @@ public class CustomConsoleLogger : ILogger
 
         /// <summary>
         /// Creates Log message by setting console message color based on LogLevel.
+        /// Skips logging when in MCP stdio mode to keep stdout clean for JSON-RPC protocol.
         /// </summary>
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
         {
+            // In MCP stdio mode, suppress all CLI logging to keep stdout clean for JSON-RPC.
+            if (Cli.Utils.IsMcpStdioMode)
```
RubenCerna2079, line 79: We also need to check if we are overriding the LogLevel from the CLI or the config file or else it would skip outputting the logs even when it is not expected.

src/Service/Program.cs
```diff
@@ -63,6 +64,28 @@ public static bool StartEngine(string[] args, bool runMcpStdio, string? mcpRole)
...
+                    // When LogLevel.None, redirect to null stream for ZERO output.
+                    // Otherwise redirect to stderr so logs don't pollute JSON-RPC.
+                    if (initialLogLevel == LogLevel.None)
+                    {
+                        Console.SetOut(TextWriter.Null);
+                        Console.SetError(TextWriter.Null);
+                    }
+                    else
+                    {
+                        Console.SetOut(Console.Error);
```
RubenCerna2079, line 85: Why would we want to change the Console to only output errors if the LogLevel is overriden by the CLI or the Config file? Wouldn't that defeat the purpose of allowing the LogLevel to be overriden in the first place?

src/Service/Program.cs
```diff
@@ -115,13 +138,20 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
                 })
                 .ConfigureLogging(logging =>
                 {
-                    logging.AddFilter("Microsoft", logLevel => LogLevelProvider.ShouldLog(logLevel));
-                    logging.AddFilter("Microsoft.Hosting.Lifetime", logLevel => LogLevelProvider.ShouldLog(logLevel));
+                    // Set minimum level at the framework level - this affects all loggers.
+                    // For MCP stdio mode, Console.Out is redirected to stderr in Main(),
+                    // so any logging output goes to stderr and doesn't pollute the JSON-RPC channel.
+                    logging.SetMinimumLevel(LogLevelProvider.CurrentLogLevel);
```
RubenCerna2079, line 144: Just curious, does this set the LogLevel for all the possible namespaces? Since the reason we had it the way we did is because the LogLevel was not being applied to some namespaces.

src/Service/Program.cs
```diff
@@ -63,6 +64,28 @@ public static bool StartEngine(string[] args, bool runMcpStdio, string? mcpRole)
...
+
+                // For MCP stdio mode, redirect Console.Out to keep stdout clean for JSON-RPC.
+                // MCP SDK uses Console.OpenStandardOutput() which gets the real stdout, unaffected by this redirect.
+                if (runMcpStdio)
+                {
+                    // When LogLevel.None, redirect to null stream for ZERO output.
+                    // Otherwise redirect to stderr so logs don't pollute JSON-RPC.
+                    if (initialLogLevel == LogLevel.None)
+                    {
+                        Console.SetOut(TextWriter.Null);
```
RubenCerna2079, line 80: Would this change if the LogLevel is intoduced in config file as part of a hot-reload?

src/Service/Program.cs
```diff
@@ -133,19 +163,59 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
         /// Using System.CommandLine Parser to parse args and return
         /// the correct log level. We save if there is a log level in args through
         /// the out param. For log level out of range we throw an exception.
+        /// When in MCP stdio mode without explicit --LogLevel, defaults to None without CLI override.
```
RubenCerna2079, line 166: nit:
```suggestion
        /// When in MCP stdio mode without explicit --LogLevel, defaults to None without CLI or Config override.
```

src/Service/Program.cs
```diff
@@ -133,19 +163,59 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
...
         /// </summary>
         /// <param name="args">array that may contain log level information.</param>
-        /// <param name="isLogLevelOverridenByCli">sets if log level is found in the args.</param>
+        /// <param name="runMcpStdio">whether running in MCP stdio mode.</param>
```
RubenCerna2079, line 169: nit: Capitalize the first letter of all the descriptions for the parameters.

RubenCerna2079 (changes requested):

src/Cli/ConfigGenerator.cs
```diff
@@ -2606,13 +2606,38 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
+                        // Don't add --LogLevel to args - let Service handle the default.
+                    }
+                }
+                else
+                {
+                    minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
+                    HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
 
-                _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
+                    _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
```
Aniruddh25, line 2636: if we dont use minimumLogLevel in any way over here in CLI when the --LogLevel argument is absent, why even bother to get the configured log level? See PR #3426 that attempts to fix part of this as well.

src/Service/Program.cs
```diff
@@ -63,6 +64,28 @@ public static bool StartEngine(string[] args, bool runMcpStdio, string? mcpRole)
         {
             try
             {
+                // Initialize log level EARLY, before building the host.
+                // This ensures logging filters are effective during the entire host build process.
+                LogLevel initialLogLevel = GetLogLevelFromCommandLineArgs(args, runMcpStdio, out bool isCliOverridden, out bool isConfigOverridden);
+                LogLevelProvider.SetInitialLogLevel(initialLogLevel, isCliOverridden, isConfigOverridden);
```
Aniruddh25, line 70: what is the use of `isConfigOverriden` flag?

src/Cli/ConfigGenerator.cs
```diff
@@ -2606,13 +2606,38 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
+                    // Check if config explicitly sets a log level
+                    if (!deserializedRuntimeConfig.IsLogLevelNull())
+                    {
+                        minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
+                        _logger.LogInformation("MCP stdio mode: Using config log-level: {minimumLogLevel}.", minimumLogLevel);
+                        // Pass --LogLevel to Service with special marker to indicate it's from config, not CLI.
+                        // This allows MCP logging/setLevel to be blocked by config override.
+                        args.Add("--LogLevel");
+                        args.Add(minimumLogLevel.ToString());
+                        args.Add("--LogLevelFromConfig");
```
Aniruddh25, line 2623: Why do we need to add --LogLevelFromConfig for mcp stdio mode, but not otherwise? Wouldnt the Service read from config regardless - what is special about mcp stdio mode that the service cannot read from config and determine the fate of whether to default to None or not?

If we do need provide the loglevel fromconfig for logs sent before Service can read the config, it needs to be done for non-mcp stdio mode as well.

src/Cli/ConfigGenerator.cs
anushakolan: Removed

src/Cli/ConfigGenerator.cs
anushakolan: Comment is no longer applicable

src/Cli/CustomLoggerProvider.cs
anushakolan: The CLI runs before the config file is parsed by Service, so at the CLI logging level we can only check if the user explicitly passed [--LogLevel] on the command line.

Config-based log-level override is handled later by Service's [UpdateFromRuntimeConfig()](vscode-file://vscode-app/c:/Users/anushakolan/AppData/Local/Programs/Microsoft%20VS%20Code/b6a47e94e3/resources/app/out/vs/code/electron-browser/workbench/workbench.html), which checks if the config file has an explicit log level set.

For CLI startup messages specifically:

In MCP mode without --LogLevel flag → suppressed (clean stdout)
In MCP mode with --LogLevel flag → written to stderr
In normal mode → written to stdout 

If config has a log-level but user didn't pass --LogLevel CLI flag, the CLI startup messages are suppressed in MCP mode. This is intentional to keep stdout clean for JSON-RPC. Once Service starts, it reads the config and adjusts the log level accordingly.

src/Service/Program.cs
anushakolan: This is a misunderstanding of what `Console.SetOut(Console.Error)` does.

It doesn't mean "only output errors" - it means redirect stdout to stderr. All logs (info, debug, warning, etc.) that would normally go to stdout are instead written to stderr.

In MCP stdio mode, stdout must be reserved exclusively for JSON-RPC protocol messages. By redirecting Console.Out to Console.Error (stderr), we allow all logs to be visible on stderr while keeping stdout clean for the MCP protocol.

So when user sets `--LogLevel Debug`:

Debug, Info, Warning, Error logs all appear on stderr
JSON-RPC messages go to stdout
MCP communication works correctly while logs are still visible

src/Service/Program.cs
anushakolan: No, hot-reload would not change the Console redirect. The `Console.SetOut()` call happens once at startup in `Main()` before the host is built.

During hot-reload, UpdateFromRuntimeConfig() updates the log level filtering via `DynamicLogLevelProvider.CurrentLogLevel`, but it doesn't re-execute `Console.SetOut()`. So if you start with `LogLevel.None` (which sets `Console.Out = TextWriter.Null`), then hot-reload the config to add a log level, the filtering would allow those log messages through, but they'd still go to `TextWriter.Null`.

This is acceptable because MCP `logging/setLevel` is the intended mechanism for dynamic log level changes at runtime - it updates the log level and logs go to stderr (which was already configured at startup).

src/Cli/ConfigGenerator.cs
anushakolan: This comment is no longer applicable since the code has been updated.

src/Cli/ConfigGenerator.cs
anushakolan: Yeah, you are right. We removed this anyways.

src/Service/Program.cs
anushakolan: Yes, SetMinimumLevel sets a global minimum threshold that applies to all namespaces - it's the framework-level floor that filters out any messages below that level.

anushakolan:
Changes merged into PR #3419. Closing this stacked PR.

src/Service/Program.cs
anushakolan: The `isConfigOverridden` flag tracks whether the user explicitly set a log level in their config file. This is used to enforce log level precedence: CLI takes highest priority, then config file, then MCP client requests.

When a config file specifies a log level, MCP clients cannot change it at runtime—their requests are accepted but silently ignored. This prevents unexpected behavior where an agent could override the operator's intended log settings.

Issue #3282, opened 2026-03-15
[Bug]: Server instruction not present in JSON schema.


JerryNixon:
## What?

The server description is defined in code:

https://github.com/Azure/data-api-builder/blob/9563a201cc816ab02d3a7bac32f72ea5bd8d29d2/src/Config/ObjectModel/McpRuntimeOptions.cs#L33-L37

But the server description is not defined in the JSON Schema:

https://github.com/Azure/data-api-builder/blob/9563a201cc816ab02d3a7bac32f72ea5bd8d29d2/schemas/dab.draft.schema.json#L268-L358

## Fix

This is what is missing

```json
"description": {
  "type": "string",
  "description": "Description of the MCP server, exposed as the 'instructions' field in the MCP initialize response to provide behavioral context to MCP clients and agents."
}
```

anushakolan:
Version 1.7

PR #3423, opened 2026-04-09
Added changes to set McpServerOptions.Instructions in HTTPS/SSE mode.

### LINKED ISSUE #3283: [Bug]: Server `instruction` not set properly in the engine.
## What?

The `runtime.mcp.description` config value is read into `McpRuntimeOptions.Description` and emitted as `instructions` in the stdio server's `HandleInitialize`, but the HTTP/SSE transport configured in `McpServerConfiguration.cs` never sets `McpServerOptions.Instructions`. MCP clients connecting over HTTP/SSE never receive `instructions` in the initialize response.

### Suggestion

```csharp
services.PostConfigure<McpServerOptions>(options =>
{
    options.ServerInfo = new() { Name = McpProtocolDefaults.MCP_SERVER_NAME, Version = McpProtocolDefaults.MCP_SERVER_VERSION };
    options.Capabilities ??= new();
    options.Capabilities.Tools ??= new();
    
    // Read the description from runtime config and set it as instructions
    // (This needs access to RuntimeConfigProvider, so the implementation 
    //  would need to use services.AddOptions<McpServerOptions>() with a 
    //  configure callback that resolves from the service provider)
    options.Instructions = runtimeConfig.Runtime?.Mcp?.Description;
});
```

## Also

Additionally, the `description` property is missing from the `mcp` object in `schemas/dab.draft.schema.json`, so configs using it fail schema validation since that object has `"additionalProperties": false`. #3282

anushakolan:
## Why make this change?

Closes #3283

`runtime.mcp.description` was configured, but MCP HTTP/SSE initialize did not return it as `instructions`. This made the engine behavior inconsistent with expected MCP initialization output.

Related discussion: https://github.com/Azure/data-api-builder/issues/3282

## What is this change?

- Wired runtime MCP description into MCP server options for HTTP/SSE initialization:
  - `runtime.mcp.description` -> `ServerInstructions`
  - Files updated:
    - `src/Azure.DataApiBuilder.Mcp/Core/McpServiceCollectionExtensions.cs`
    - `src/Azure.DataApiBuilder.Mcp/Core/McpServerConfiguration.cs`

- Added/updated regression coverage for initialize response instructions:
  - `src/Service.Tests/Configuration/ConfigurationTests.cs`
  - Test: `TestMcpInitializeIncludesInstructionsFromRuntimeDescription`

How the bug was simulated:
- Reverted the MCP wiring changes and called MCP `initialize`; `result.instructions` was missing.

How it was verified after fix:
- Re-applied the wiring, called MCP `initialize` again, and confirmed `result.instructions` is present.

## How was this tested?

- [x] Integration Tests
- [ ] Unit Tests

Ran:

- `dotnet test src/Service.Tests/Azure.DataApiBuilder.Service.Tests.csproj --filter "FullyQualifiedName~ConfigurationTests.TestMcpInitializeIncludesInstructionsFromRuntimeDescription" -v minimal`

## Sample Request(s)

MCP initialize request (HTTP):

```bash
curl -i -X POST http://localhost:5000/mcp \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{
    "jsonrpc":"2.0",
    "id":1,
    "method":"initialize",
    "params":{
      "protocolVersion":"2025-03-26",
      "capabilities":{},
      "clientInfo":{"name":"manual-test","version":"1.0.0"}
    }
  }'
```

Expected response snippet after fix:

```json
{
  "result": {
    "instructions": "Use SQL tools to query the database."
  }
}
```


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -2804,6 +2804,44 @@ public async Task TestGlobalFlagToEnableRestGraphQLAndMcpForHostedAndNonHostedEn
...
+                $"--ConfigFileName={CUSTOM_CONFIG}"
+            };
+
+            using TestServer server = new(Program.CreateWebHostBuilder(args));
+            using HttpClient client = server.CreateClient();
+
+            JsonElement initializeResponse = await GetMcpInitializeResponse(client, configuration.Runtime.Mcp);
+            JsonElement result = initializeResponse.GetProperty("result");
+
+            Assert.AreEqual(MCP_INSTRUCTIONS, result.GetProperty("instructions").GetString(), "MCP initialize response should include instructions from runtime.mcp.description.");
```
Aniruddh25, line 2840: Cant we modify an existing test that checks InitializeResponse to check if the instructions are sent back in the output? 

I expect there should be an existing test - otherwise we might be missing coverage.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -6284,6 +6322,83 @@ public static async Task<HttpStatusCode> GetMcpResponse(HttpClient httpClient, M
...
+                    string data = trimmed.Substring("data:".Length).Trim();
+                    if (!string.IsNullOrWhiteSpace(data) && data.StartsWith('{'))
+                    {
+                        return data;
+                    }
+                }
+            }
+
+            return string.Empty;
+        }
```
Aniruddh25, line 6402: Copilot suggestions are quite valid

Aniruddh25 (changes requested):
minimize test duplication.

src/Service.Tests/Configuration/ConfigurationTests.cs
anushakolan: No existing test validates MCP initialize response content. The only MCP test (`TestGlobalFlagToEnableRestGraphQLAndMcpForHostedAndNonHostedEnvironment`) only checks HTTP status codes and discards the response body. This was indeed missing coverage—the new test fills that gap.

Also refactored helpers per your feedback to minimize duplication.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -6278,7 +6320,85 @@ public static async Task<HttpStatusCode> GetMcpResponse(HttpClient httpClient, M
...
+                    return string.Empty;
+                }
+
+                string combinedPayload = string.Join("\n", dataLines);
+                return !string.IsNullOrWhiteSpace(combinedPayload) && combinedPayload.TrimStart().StartsWith('{')
+                    ? combinedPayload
+                    : string.Empty;
+            }
+
+            foreach (string rawLine in ssePayload.Split('\n'))
```
Aniruddh25, line 6370: Can you provide an example of an expected `ssePayload`?

why would there be multiple `data:` in the payload?
As of now, I see this being the format. If this is accurate expectation, can you please extend the description to provide this e.g.?
`data: { }\r\ndata: { }\r\n`

Aniruddh25 (approved):

src/Service.Tests/Configuration/ConfigurationTests.cs
anushakolan: `ssePayload` is the raw Server‑Sent Events (SSE) payload for a single dispatched event.

SSE allows the `data:` field to appear multiple times within the same event block (terminated by a blank line).
When multiple `data:` lines are present, EventSource concatenates them, inserting '\n' between each line’s value.

Example (one event, multi-line data):
data: { ...first line... }\r\n
data: { ...second line... }\r\n
\r\n

The client-visible `event.data` becomes: "{ ...first line... }\n{ ...second line... }"

RubenCerna2079 (approved):
LGTM

Aniruddh25 (approved):

PR #2904, opened 2025-10-13
[MCP] Added parameter description.

anushakolan:
## Why make this change?

- Expands stored procedure parameter definitions in the configuration file to support richer metadata, using parameter's attributes like, `name`, `description`, `required`, and `default` values.
- Enables MCP and other tooling to leverage parameter metadata for improved documentation, validation, and user experience.
- Supports backward compatibility by allowing both the legacy dictionary format and the new array-of-objects format for parameters.

## What is this change?

- Changes the `parameters` property in entity config from a dictionary mapping parameter names to values, to an array of parameter objects with explicit fields (`name`, `required`, `default`, `description`).
- Updates the JSON schema to support the new structure and maintain support for the old syntax.
- Refactors the internal model to handle both formats and expose richer metadata.
- Enhances CLI commands (`dab add` and `dab update`) to accept and manage parameter metadata fields: `name`, `required`, `default`, and `description`.
- Adds support for the expanded parameter object in CLI validation, including deprecation notes for the legacy format.
- Includes parameter descriptions in DAB's OpenAPI and GraphQL schema outputs for improved API documentation.
- This change also introduces smarter handling of optional parameters for stored procedures. If a parameter is not required and neither a user value nor a config default is provided, DAB will skip sending that parameter, allowing the database’s internal default to apply. This avoids duplication and drift between config and database defaults. The old behavior is fully supported for backward compatibility.

## How was this tested?

- [x] Manual Testing:
-  Used CLI commands to add, validate and update parameters with metadata.
-  Verified config parsing, CLI commands, and schema generation for both legacy and expanded parameter formats.
- Confirmed correct behavior in OpenAPI and GraphQL outputs.

## Sample Request(s)

**Before (legacy dictionary format):**
```json
{
  "entities": {
    "User": {
      "source": {
        "parameters": {
          "param1": "default-value"
        }
      }
    }
  }
}
```

**After (expanded array format):**
```json
{
  "entities": {
    "User": {
      "source": {
        "parameters": [
          {
            "name": "Id",
            "required": true,
            "default": "default-value",
            "description": "The unique identifier for the user."
          }
        ]
      }
    }
  }
}
```

**Command Line Examples:**

**Before (legacy dictionary format):**

```
[...]

PR #3425, opened 2026-04-09
Added missing stored procedure parameters in describe_entities response.

### LINKED ISSUE #3400: [Bug]: describe_entities missing stored procedure parameters
describe_entities missing stored procedure parameters

anushakolan:
## Why make this change?

Closes #3400.

`describe_entities` could return stored-procedure entities with empty or partial `parameters` metadata when parameters were not fully listed in config. MCP clients/agents then missed required inputs and `execute_entity` calls failed.

## What is this change?

`describe_entities` now uses the runtime database metadata as the source of truth for stored-procedure parameter discovery, with config acting as an overlay on top. Per the rules locked in #3400:

- **Names** come from the database. Config never overrides parameter names.
- **`required`** defaults to `true`. Config overrides per-parameter when the user explicitly supplies it. To make this round-trip cleanly through JSON deserialization, `ParameterMetadata.Required` is now `bool?` so "user omitted the key" is distinct from "user wrote `required: false`". The global `JsonSerializer` policy `DefaultIgnoreCondition.WhenWritingNull` keeps generated configs clean.
- **`default`** is config-only. T-SQL doesn't expose parameter defaults via metadata for ordinary stored procedures, so the DB-side value is unreliable.
- **`description`** is config-only. There is no DB column carrying per-parameter descriptions.
- Parameters discovered from the database that are not listed in config are still surfaced (with `required: true`, no default, no description) so agents can see and call them.

The merge itself is already performed upstream in `SqlMetadataProvider.FillSchemaForStoredProcedureAsync`, which populates `DatabaseStoredProcedure.StoredProcedureDefinition.Parameters` with config values laid over the introspected DB schema. `DescribeEntitiesTool.BuildParameterMetadataInfo` projects that already-merged structure directly, instead of carrying its own parallel merge.

If the upstream merge invariant is ever violated (i.e. an SP entity reaches `describe_entities` without populated DB metadata), the tool throws inside the per-entity `try` and the surrounding `catch` drops just that entity from the response. Returning the SP with `parameters: []` would be indistinguishable from a real zero-parameter SP and would mislead the agent into calling the SP with no arguments.

A config that declares an SP parameter the DB does not have continues to fail loudly at `InitializeAsync` (via `HandleOrRecordException`), so `describe_entities` never runs against a half-merged state in production.

## How was this tested?

- [x] Unit tests (projection logic, with the merged `ParameterDefinition` dictionary supplied directly)
- [x] Integration tests against a live SQL Server (real `MsSqlMetadataProvider`, real merge, real DB introspection)
- [x] Manual stdio MCP testing

### Unit tests — `DescribeEntitiesStoredProcedureParametersTests`

These tests pass a hand-built `Dictionary<string, ParameterDefinition>` representing the post-merge state and verify the projection without touching a database.

| Test | What it locks in |
| --- | --- |
| `MixedConfiguredAndUnconfiguredParameters_ProjectsBoth` | A parameter with config overrides and one without are projected side-by-side: the overridden param reflects config; the non-overridden one defaults to `required=true`, `default=null`, `description=""`. |
| `ExplicitRequiredFalse_IsHonored` (`WithDefault` / `WithoutDefault`) | An explicit `required=false` written onto the `ParameterDefinition` by the upstream merge is honored both with and without a `Default` value. |
| `DropsEntity_WhenStoredProcedureMetadataIsMissing` | If the upstream invariant ever regresses and an SP entity has no `DatabaseStoredProcedure`, the tool drops that entity from the response rather than returning it with `parameters=[]`. When it's the only entity, the tool returns the `NoEntitiesConfigured` error result. |

### Integration tests — `DescribeEntitiesStoredProcedureParametersMsSqlIntegrationTests` (`TestCategory.MSSQL`)

Extend `SqlTestBase`, so they bootstrap the real `MsSqlMetadataProvider` against the live test database via `dab-config.MsSql.json` and `DatabaseSchema-MsSql.sql`. These exercise the actual config-onto-DB merge that the unit tests cannot.

| Test | SP / config | What it locks in |
| --- | --- | --- |
| `StoredProcedureWithNoDbOrConfigParameters_EmitsEmptyParametersArray` | `GetBooks` → `get_books` (0 DB params, no config overrides) | A truly zero-parameter SP emits `parameters: []`. |
| `DbParameterWithoutConfigOverride_DefaultsToRequiredTrue` | `GetBook` → `get_book_by_id @id int` (no config override) | A DB-discovered parameter with no config entry comes out `required=true, default=null, description=""`. |
| `DbParametersWithConfigOverrides_MergesConfigValuesOntoDbParameters` | `InsertBook` → `insert_book @title, @publisher_id` (config sets both `required=false` with defaults) | Config-side `required`/`default` flow onto DB-discovered params end-to-end. |
| `PartialConfigOverride_MergesOverriddenParamAndDefaultsTheOther` | `update_book_title @id, @title` (config overrides `id` only) | Mixed merge against a real SP: the overridden param reflects config; the non-overridden one falls back to the merge defaults. |
| `SqlMetadataProvider_FailsAtInit_WhenConfigDeclaresParameterNotInDb` | `get_book_by_id` with a bogus extra param in config | Locks in the upstream invariant `DescribeEntitiesTool` depends on: a config that declares a parameter the DB does not have throws `DataApiBuilderException` at `InitializeAsync` (mentioning both the bogus param name and the SP name), so the tool never reaches a half-merged state in production. |

### Manual end-to-end test

Validated against a live SQL Server over the MCP stdio transport. Each test entity is backed by the same two-parameter stored-procedure shape — `(@id INT, @locale NVARCHAR(10) = 'en-US')` — and each entity declares its parameters differently in `dab-config.json` so we can verify how config and DB metadata combine. The custom-tool entity is backed by a separate two-parameter SP that takes a search title and a max-row count.

<details>
<summary>Stored procedure DDL</summary>

```sql
-- Each test entity gets its own SP with the same shape so config on one
-- entity does not pollute the shared parameter metadata of another.
CREATE PROCEDURE dbo.get_book_full    @id INT, @locale NVARCHAR(10) = 'en-US' AS ...
CREATE PROCEDURE dbo.get_book_partial @id INT, @locale NVARCHAR(10) = 'en-US' AS ...
CREATE PROCEDURE dbo.get_book_empty   @id INT, @locale NVARCHAR(10) = 'en-US' AS ...
CREATE PROCEDURE dbo.search_books     @title NVARCHAR(200), @max_results INT  AS ...
```
</details>

<details>
<summary>Entities in dab-config.json</summary>

```jsonc
"BookFull": {
  "source": {
    "object": "dbo.get_book_full",
    "type": "stored-procedure",
    "parameters": [
      { "name": "id",     "required": true,  "default": "42",    "description": "Book id from config" },
      { "name": "locale", "required": false, "default": "fr-FR", "description": "Locale from config" }
    ]
  },
  "permissions": [{ "role": "anonymous", "actions": [{ "action": "execute" }] }]
},
"BookPartial": {
  "source": {
    "object": "dbo.get_book_partial",
    "type": "stored-procedure",
    "parameters": { "id": 1 }                // old object form; no `required`, locale omitted
  },
  "permissions": [{ "role": "anonymous", "actions": [{ "action": "execute" }] }]
},
"BookEmpty": {
  "source": {
    "object": "dbo.get_book_empty",
    "type": "stored-procedure"
    // no parameters block at all
  },
  "permissions": [{ "role": "anonymous", "actions": [{ "action": "execute" }] }]
},
"SearchBooks": {
  "source": {
    "object": "dbo.search_books",
    "type": "stored-procedure",
    "parameters": [
      { "name": "title",       "required": true, "description": "Substring to search for in book title" },
      { "name": "max_results", "required": true, "description": "Maximum rows to return" }
    ]
  },
  "mcp": { "custom-tool": true },
  "permissions": [{ "role": "anonymous", "actions": [{ "action": "execute" }] }]
}
```
</details>

**T1 — `tools/call describe_entities`** (verifies the metadata returned to MCP clients)

For each entity below, the columns show what the tool reported for each parameter. The "What this verifies" column explains why this row matters.

| Entity (config shape) | Param | required | default | description | What this verifies |
| --- | --- | --- | --- | --- | --- |
| `BookFull` — config sets `required`, `default`, `description` for both params | `id` | true | `"42"` | `"Book id from config"` | Explicit per-param config flows through unchanged. |
| `BookFull` | `locale` | false | `"fr-FR"` | `"Locale from config"` | Explicit `required: false` from the user is still honored (rule 2: config overrides default). |
| `BookPartial` — config uses old object form `"parameters": { "id": 1 }` (no `required` key, `locale` not mentioned) | `id` | true | `"1"` | `""` | Old-format config supplies the default; absent `required` defaults to **true** (the fix). |
| `BookPartial` | `locale` | true | `null` | `""` | Param is in the SP but not in config — still surfaced from DB metadata, defaults to **required: true** (the fix; rule 5). |
| `BookEmpty` — no `parameters` block in config at all | `id` | true | `null` | `""` | Both params discovered purely from DB metadata, defaulting to required (the fix; rule 5). |
| `BookEmpty` | `locale` | true | `null` | `""` | Same as above. |
| `SearchBooks` — `mcp.custom-tool: true`; config marks both params `required: true` with descriptions | `title` | true | `null` | `"Substring to search for in book title"` | Entities marked as custom tools still appear in `describe_entities` with the same merge rules applied. |
| `SearchBooks` | `max_results` | true | `null` | `"Maximum rows to return"` | Same as above. |

**T2 — `tools/call execute_entity`** for `BookEmpty` (no params declared in config) with arguments `{ "entity": "BookEmpty", "parameters": { "id": 2, "locale": "fr-FR" } }` → `status: success`, returns the row for id=2 (`"Le Petit Prince"`). Confirms an agent that follows the `describe_entities` output can invoke the SP successfully even when the config declares zero parameters.

**T3 — `tools/call execute_entity`** for `BookPartial` with no `parameters` argument → `status: success`, returns the row `{ id: 1, title: "The Pragmatic Programmer", locale: "en-US" }`. The returned `id=1` matches the config-supplied default `"1"`, and `locale="en-US"` matches the T-SQL `DEFAULT 'en-US'` on `@locale`. Confirms that omitted parameters fall back to whichever default is available.

**T4 — `tools/call search_books`** (the custom-tool surface for `SearchBooks`) with `{ "title": "Pr", "max_results": 5 }` → `status: success`, returns 2 rows whose titles contain "Pr": `"The Pragmatic Programmer"` and `"Le Petit Prince"`. Confirms a stored procedure exposed as a top-level MCP custom tool can be invoked end-to-end with the new `Required` semantics.


src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -422,14 +425,53 @@ private static bool ShouldIncludeEntity(string entityName, HashSet<string>? enti
...
+            IServiceProvider serviceProvider,
+            ILogger? logger,
+            CancellationToken cancellationToken)
+        {
+            if (entity.Source.Type != EntitySourceType.StoredProcedure)
+            {
+                return null;
+            }
+
+            IMetadataProviderFactory? metadataProviderFactory = serviceProvider.GetService<IMetadataProviderFactory>();
```
Aniruddh25: where is the metadataProviderFactory used?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -461,10 +503,56 @@ private static List<object> BuildFieldMetadataInfo(List<FieldMetadata>? fields)
...
+            {
+                foreach (ParameterMetadata parameter in parameters)
+                {
+                    configParameters[parameter.Name] = parameter;
+                }
+            }
+
+            if (databaseObject is DatabaseStoredProcedure storedProcedure)
+            {
+                foreach ((string parameterName, ParameterDefinition parameterDefinition) in storedProcedure.StoredProcedureDefinition.Parameters)
```
Aniruddh25: What was the reason again to have two different data structures? `ParameterDefinition`and `ParameterMetadata`? why couldnt we have used `ParameterDefinition` for the config parameters?

Aniruddh25 (changes requested):
Want to know reasoning behind why the bug happened.

RubenCerna2079:
Hey just wanted to let you know that it seems this PR is related to a bug that I created for 2.0. Might be a good idea to check it and see if it can be fixed on this PR or as a follow up. https://github.com/Azure/data-api-builder/issues/3435

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: Not needed, removed.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: `ParameterDefinition` is runtime metadata discovered from the database. `ParameterMetadata` is user-specified config overrides. Both exist because DB metadata provides the source of truth for parameters, while config allows users to enrich with descriptions/defaults.

anushakolan:
> Hey just wanted to let you know that it seems this PR is related to a bug that I created for 2.0. Might be a good idea to check it and see if it can be fixed on this PR or as a follow up. #3435

Thanks for linking these! They share the same root cause: parameter metadata is read only from config, not from runtime DB metadata.

PR #3425 fixes `describe_entities` (AI agents can now discover parameters).

Bug #3435 needs a follow-up PR to fix `DynamicCustomTool.BuildInputSchema()` the same way - it currently only reads `Entity.Source.Parameters` from config, not from `IMetadataProviderFactory`.

The fix requires passing `IServiceProvider` through `CustomMcpToolFactory` to `DynamicCustomTool` so it can access runtime metadata for schema building.

anushakolan:
> Want to know reasoning behind why the bug happened.

Original implementation only read config (`Entity.Source.Parameters`), ignoring runtime-discovered parameters from `IMetadataProviderFactory`. The fix merges both: DB metadata as base, config as overrides. May be @souvikghosh04 can add anything I missed.

souvikghosh04:
> > Want to know reasoning behind why the bug happened.
> 
> Original implementation only read config (`Entity.Source.Parameters`), ignoring runtime-discovered parameters from `IMetadataProviderFactory`. The fix merges both: DB metadata as base, config as overrides. May be @souvikghosh04 can add anything I missed.

I think we are discussing two separate things-

1. This PR: based on what I see the PR is trying to achieve is its merging the DB metadata (a new and additional pull) and the config defined metadata. but this creates inconsistency- discovery will surface DB-discovered params that tools still reject/don't advertise (since they validate against config only).
2. Issue #3435: It is mentioned in the description that the custom tool cannot be run for stored procedures which requires parameters. This is not correct. custom tool implementation (PR https://github.com/Azure/data-api-builder/pull/3048) correctly handles dynamic parameters when they're passed. Please refer to the sample JSON requests in the mentioned PR and see if those are supported in MCP Inspector. If not, you can use Insomnia client or stdio (GHCP) to test.

so for point 1- we might not need this PR unless we want to move completely to DB based discoverability to avoid split brain or incosistent behaviour.
for point 2- seems unrelated to this and probably should be closed since custom tool already support dynamic parameters based on what is configured in config during elevation of a stored proc as custom tool.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -536,7 +538,8 @@ private static List<object> BuildParameterMetadataInfo(List<ParameterMetadata>?
                 // Preserve config-only parameters if metadata is not available for a configured name.
                 foreach (ParameterMetadata configParameter in configParameters.Values)
                 {
-                    if (!storedProcedure.StoredProcedureDefinition.Parameters.ContainsKey(configParameter.Name))
+                    if (!storedProcedureParameters.ContainsKey(configParameter.Name) &&
```
Aniruddh25: why cant the `ContainsKey` comparison itself be case insensitive instead of adding another && condition?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
Aniruddh25: I understand we need both the objects. My question is - why couldnt we have expanded `ParameterDefinition` - the same data structure to support user specified config overrides? Why did we introduce a new class `ParameterMetadata` = also, the name is confusing. 

Parameter information from DB metadata is called `ParameterDefinition` and not `ParameterMetadata`.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
Aniruddh25: On second thought, all `configParameters` should be present in the metadata retrieved from the database. In fact, config parameters should be a subset of database parameters. 

There might a validation check to ensure this.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -461,10 +496,66 @@ private static List<object> BuildFieldMetadataInfo(List<FieldMetadata>? fields)
...
+
+                if (storedProcedureParameters is null || storedProcedureParameters.Count == 0)
+                {
+                    // No runtime metadata available for this stored procedure. Fall back to config-only parameters.
+                    return BuildParameterMetadataInfo(parameters, databaseObject: null);
+                }
+
+                foreach ((string parameterName, ParameterDefinition parameterDefinition) in storedProcedureParameters)
+                {
+                    configParameters.TryGetValue(parameterName, out ParameterMetadata? configParameter);
```
Aniruddh25: PR description says - 

`Merged DB metadata with config metadata so config values override matching fields (required, default, description).`

However, this is not what is implemented.

With the current implementation, config values will never be added to the result since ALL config parameters should be discoverable in storedProcedureParameters by definition. 
But the current check `!storedProcedureParameters.ContainsKey(configParameter.Name)`  tries to identify the missing ones. 

Config metadata should only be used to override, which means after reading params from DB metadata, the config values should be used to replace the values from DB.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            CallToolResult result = await tool.ExecuteAsync(null, serviceProvider, CancellationToken.None);
+
+            Assert.IsTrue(result.IsError == false || result.IsError == null);
+
+            JsonElement entity = GetSingleEntityFromResult(result);
+            JsonElement parameters = entity.GetProperty("parameters");
+            Assert.AreEqual(2, parameters.GetArrayLength());
+
+            JsonElement idParam = parameters.EnumerateArray().Single(p => p.GetProperty("name").GetString() == "id");
+            Assert.IsFalse(idParam.GetProperty("required").GetBoolean(), "DB required metadata should be preferred when config cannot indicate whether 'required' was explicitly set.");
```
Aniruddh25: The test is titled DescribeEntities_ConfigParameterMetadataOverridesDatabaseParameterMetadata
whereas this assert is doing exactly the opposite - DB is overriding config?

Assert should be Assert.IsTrue

anushakolan:
Based on @souvikghosh04 's point, surfacing DB-discovered params in describe_entities while execute_entity and custom tools (list_tools / #3435) still validate and advertise from config-only creates a split-brain — agents will see params they can't actually pass. The team also seemed to agree on Teams that the right model is DB metadata as the source of truth, with config as overlay, applied consistently across all MCP surfaces.

That's bigger than what this PR currently does. So, I see two options:

Close this PR and open a single tracked task for "DB-first parameter discovery across MCP" covering describe_entities + execute_entity validation + custom tool list_tools (#3435) + the override semantics + test fix.
Keep this PR but expand its scope to cover all of the above in one change.
I'd lean toward option 1 since the work touches multiple surfaces and deserves its own design discussion. Thoughts?

On `ParameterMetadata` vs `ParameterDefinition`:

To answer @Aniruddh25 's question on why two classes, exist - the split is justified, just poorly named:

`ParameterDefinition` (in `Config/DatabasePrimitives/DatabaseObject.cs`) is a runtime/in-memory type. It holds DB-discovered metadata like SystemType, DbType, SqlDbType, HasConfigDefault, ConfigDefaultValue - none of which belong in the config file.

`ParameterMetadata` (in `Config/ObjectModel/`) is the config-bound type. It only holds what users write in JSON: name, description, required, default. It was introduced in PR #2904 when the config parameters shape changed from a dictionary ("p": "value") to an array of objects.

Mixing them would mean either polluting the config schema with runtime-only fields or scattering [JsonIgnore]s on a type that's also used as the config model. The separation is the cleaner design. Happy to rename `ParameterMetadata` → something like ParameterConfig (or ConfigParameter) in a small follow-up PR if that addresses the confusion.


I would like to know your comments before taking this PR forward.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: The entire fallback block has been removed. 

You're right. `SqlMetadataProvider.FillSchemaForStoredProcedureAsync` already throws `DataApiBuilderException` at startup if a config-listed parameter isn't found in DB metadata. I've removed the `!ContainsKey` fallback block — it was dead code by definition — and called out this guarantee in the new XML doc on `BuildParameterMetadataInfo`.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: Fixed this.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
anushakolan: Flipped to Assert.IsTrue and updated the message.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -179,7 +180,7 @@ public Task<CallToolResult> ExecuteAsync(
                         {
                             Dictionary<string, object?> entityInfo = nameOnly
                                 ? BuildBasicEntityInfo(entityName, entity)
-                                : BuildFullEntityInfo(entityName, entity, currentUserRole);
+                                : BuildFullEntityInfo(entityName, entity, currentUserRole, TryResolveDatabaseObject(entityName, entity, runtimeConfig, serviceProvider, logger, cancellationToken));
```
souvikghosh04: may be add a check to call this only for EntitySourceType.StoredProcedure only here itself instead of additional method calls? that also makes TryResolveDatabaseObject reusable for other entity types if needed in future by possibly moving it to outside this class to a helper?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -456,33 +491,68 @@ private static List<object> BuildFieldMetadataInfo(List<FieldMetadata>? fields)
...
+                }
+            }
+
+            return configOnly;
         }
 
+        private static Dictionary<string, object?> BuildParameterEntry(string name, ParameterMetadata? config) => new()
+        {
+            ["name"] = name,
+            ["required"] = config?.Required ?? true,
```
souvikghosh04: will this default to true only when config is null (param not in config at all)? ParameterMetadata.Required is bool (not bool?), so when user declares a param in config without specifying required, deserialization sets it to false. please check and validate this. also, update the PR description if needed.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -0,0 +1,290 @@
...
+            RuntimeConfig config = CreateRuntimeConfig(CreateStoredProcedureEntity(parameters: configParameters));
+            ServiceCollection services = new();
+            RegisterCommonServices(services, config);
+
+            DatabaseObject? dbObject = dbParameters is null
+                ? null
+                : CreateStoredProcedureObject("dbo", "get_book", dbParameters);
+            RegisterMetadataProvider(services, TEST_ENTITY_NAME, dbObject);
+
+            IServiceProvider serviceProvider = services.BuildServiceProvider();
```
souvikghosh04: this creates the RuntimeConfig with a default DataSource and entities, then mocks IMetadataProviderFactory and ISqlMetadataProvider.EntityToDatabaseObject. but McpMetadataHelper.TryResolveMetadata first calls config.GetDataSourceNameFromEntityName(entityName) (line 46 of McpMetadataHelper.cs). If this throws for the test config, the test would fail before reaching the mock. can you check and verify this?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -456,33 +491,68 @@ private static List<object> BuildFieldMetadataInfo(List<FieldMetadata>? fields)
...
+                {
+                    configByName[parameter.Name] = parameter;
+                }
+            }
+
+            IReadOnlyDictionary<string, ParameterDefinition>? dbParameters =
+                (databaseObject as DatabaseStoredProcedure)?.StoredProcedureDefinition?.Parameters;
 
-            if (parameters != null)
+            if (dbParameters is { Count: > 0 })
```
souvikghosh04: I think you should have a test coverage for this. If the metadata provider returns a DatabaseStoredProcedure with an empty Parameters dictionary (SP with no params), it falls through to the config-only path. currently, there is no test covering this edge case.

souvikghosh04 (changes requested):
I am considering this as PR 1. the next steps after this PR can be taken in separate PRs-

PR 2: ExecuteEntityTool: Validate params against StoredProcedureDefinition.Parameters (DB metadata) instead of config. Small change, closes the split-brain with describe_entities.
PR 3: DynamicCustomTool: Thread IServiceProvider through CustomMcpToolFactory, rewrite BuildInputSchema() to use DB metadata, add proper required array and type mapping.

note: both PR 2 and PR3 should be merged as follows ups. if not, the worst case is agents see params they can't yet pass to execute_entity or custom tools, which is better than the current scenario (params invisible entirely).

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -422,14 +424,47 @@ private static bool ShouldIncludeEntity(string entityName, HashSet<string>? enti
...
+        /// </summary>
+        private static DatabaseObject? TryResolveDatabaseObject(
+            string entityName,
+            Entity entity,
+            RuntimeConfig runtimeConfig,
+            IServiceProvider serviceProvider,
+            ILogger? logger,
+            CancellationToken cancellationToken)
+        {
+            if (entity.Source.Type != EntitySourceType.StoredProcedure)
```
Aniruddh25: no reason why we couldnt resolve metadata when its not a stored proc, the if should be outside in the caller. In fact, dont see the purpose of this helper function when you can directly call McpMetadataHelper.TryResolveMetadata in the caller.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
Aniruddh25, line 444: do we have similar problem of metadata for fields of the tables/views in the config? Do we need to merge those specified in the config + column  from the db object for a table/view - this can be a separate issue though

Aniruddh25 (changes requested):

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
anushakolan: Verified — the path is fine. Walking it through:

1. The test builds RuntimeConfig via the positional constructor at RuntimeConfig.cs:312-364, which (when DataSource is not null, as it is here) populates _entityNameToDataSourceName for every entity → DefaultDataSourceName (a fresh GUID).
2. TestHelper.GenerateInMemoryRuntimeConfigProvider round-trips the config through ToJson() → MockFileSystem → FileSystemRuntimeConfigLoader, and the reloaded config goes through the same constructor and re-populates the dictionary.
3. So when McpMetadataHelper.TryResolveMetadata calls config.GetDataSourceNameFromEntityName(entityName) it hits the populated dictionary and returns the default-datasource GUID; only the bool result feeds the rest of the helper, the actual name doesn't matter for the test.

Empirical confirmation: all 16 tests across DescribeEntitiesStoredProcedureParametersTests and DescribeEntitiesFilteringTests pass — including the 4 merge tests that exercise this exact path. If the lookup were throwing, the tool's outer try/catch would surface the entity as missing and the merge assertions on parameters would never run.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: Good points — you're right that the SP-only check shouldn't be baked into the helper, so I moved that check up to the caller. Now the helper itself doesn't care about the entity type and works for tables and views too.

 On removing the helper entirely: I'd lean toward keeping it. `TryResolveMetadata` returns four `out` parameters and we only need one of them (the `DatabaseObject`), so calling it directly leaves the call site with three `out _` discards just to get to the value we want. The wrapper turns that into a one-line call where the function name itself says what's happening:

 ```csharp
 DatabaseObject? db = McpMetadataHelper.TryResolveDatabaseObject(
     entityName, config, serviceProvider, out string error, ct);
 ```

 It compiles down to a direct call to `TryResolveMetadata` — same behavior, just less noise at the call site. Happy to delete it if you still feel it's not pulling its weight, just wanted to share the reasoning first.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: Filed #3590 to track the table/view field merge

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -548,7 +535,7 @@ private static List<object> BuildParameterMetadataInfo(
         private static Dictionary<string, object?> BuildParameterEntry(string name, ParameterMetadata? config) => new()
         {
             ["name"] = name,
-            ["required"] = config?.Required ?? true,
+            ["required"] = config?.Default is null,
```
Aniruddh25: This is WRONG. `config?.Required` should be used here.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -98,6 +102,61 @@ public async Task DescribeEntities_ConfigRequiredFalse_IsHonored()
...
+                {
+                    new() { Name = "id", Required = false, Description = "Book id" }
+                },
+                dbParameters: new Dictionary<string, ParameterDefinition>
+                {
+                    ["id"] = new() { Required = true }
+                });
+
+            Assert.AreEqual(1, parameters.GetArrayLength());
+            AssertParameter(parameters, name: "id", expectedRequired: true, expectedDefault: null, expectedDescription: "Book id");
```
Aniruddh25: expectedRequired: should be false.

The test is wrong.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -98,6 +102,61 @@ public async Task DescribeEntities_ConfigRequiredFalse_IsHonored()
...
+        /// A parameter listed in config without a <c>default</c> is reported as required, even when the user did
+        /// not specify <c>required</c> in the JSON (which deserializes the non-nullable bool to <c>false</c>).
+        /// </summary>
+        [TestMethod]
+        public async Task DescribeEntities_ConfigParameterWithoutDefault_IsRequired_EvenWhenRequiredOmitted()
+        {
+            JsonElement parameters = await RunDescribeAsync(
+                configParameters: new List<ParameterMetadata>
+                {
+                    new() { Name = "id", Required = false, Description = "Book id" }
```
Aniruddh25: If Required = false is specified in the config parameter, conflicts with the test name `_EvenWhenRequiredOmitted`

Please double check AI code.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
Aniruddh25: Why does Config Default govern whether the Parameter is optional or required when config has an explicit Required property?

Aniruddh25 (changes requested):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: My bad, fixed this.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -456,33 +477,73 @@ private static List<object> BuildFieldMetadataInfo(List<FieldMetadata>? fields)
...
                 {
-                    Dictionary<string, object?> paramInfo = new()
-                    {
-                        ["name"] = param.Name,
-                        ["required"] = param.Required,
-                        ["default"] = param.Default,
-                        ["description"] = param.Description ?? string.Empty
-                    };
-                    result.Add(paramInfo);
+                    configByName.TryGetValue(parameterName, out ParameterMetadata? config);
```
Aniruddh25: The merge from config to db parameter already happens here: 

https://github.com/Azure/data-api-builder/blob/fcb09b8f4b5feb2c8f47f60ed0811f03d3627c70/src/Core/Services/MetadataProviders/SqlMetadataProvider.cs#L525

This is filled before the db object resolution happens.
So, at this point parameter definition from db object already has config overridden values. We dont need to remerge config and db. We can simply BuildParameterEntry using db's ParameterDefinition. 

The tests should still pass.

Aniruddh25 (changes requested):

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -69,8 +69,8 @@ public async Task DescribeEntities_ConfigOverridesDatabaseMetadata_ForRequiredAn
                 },
                 dbParameters: new Dictionary<string, ParameterDefinition>
                 {
-                    ["id"] = new() { Required = false, Default = "1", Description = "DB description for id" },
-                    ["tenant"] = new() { Required = true, Description = "DB description for tenant" }
+                    ["id"] = new() { Required = true, Default = "42", Description = "Config description" },
```
Aniruddh25, line 48: Are the dbParameters defined here before or after the merge while testing? 

How do we ensure the upstream merge is tested?

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -0,0 +1,341 @@
...
+            Assert.AreEqual(0, parameters.GetArrayLength());
+        }
+
+        /// <summary>
+        /// When the DB reports no parameters, the tool falls back to the config parameters.
+        /// Per rule 2 of issue #3400, a config entry without an explicit Required value
+        /// still defaults to required=true.
+        /// </summary>
+        [TestMethod]
+        public async Task DescribeEntities_EmptyDbParameters_FallsBackToConfigParameters()
```
Aniruddh25: EmptyDbParameters but provided config parameters should be resulting in exception in FillSchemaForStoredProcedureAsync

This test is not exercising that code path, so invalid. if you fix describe entities to error out when config has params but db doesnt, this test will behave accurately to error out.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -456,33 +477,78 @@ private static List<object> BuildFieldMetadataInfo(List<FieldMetadata>? fields)
...
+                foreach ((string parameterName, ParameterDefinition definition) in dbParameters)
                 {
-                    Dictionary<string, object?> paramInfo = new()
+                    result.Add(BuildParameterEntry(parameterName, definition));
+                }
+
+                return result;
+            }
+
+            // No DB metadata: emit config parameters with the same per-field rules.
```
Aniruddh25: When there is no db metadata but config has parameters, we should be erroring earlier in FillSchemaForStoredProcedureAsync. Config only code path should not be hit.

souvikghosh04 (approved):
Approved as comments were addressed/resolved.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -177,9 +178,28 @@ public Task<CallToolResult> ExecuteAsync(
...
+                                    entityName,
+                                    runtimeConfig,
+                                    serviceProvider,
+                                    out string resolveError,
+                                    cancellationToken);
+
+                                if (databaseObject is null)
+                                {
+                                    logger?.LogDebug(
+                                        "Could not resolve metadata for stored procedure entity {EntityName}. Falling back to config parameters. Error: {Error}",
```
souvikghosh04: nit: wrap in a single quote `{EntityName}` like in other places

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -0,0 +1,341 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#nullable enable
```
souvikghosh04: nit: please remove `#nullable enable` before merge

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
```diff
@@ -456,33 +477,78 @@ private static List<object> BuildFieldMetadataInfo(List<FieldMetadata>? fields)
...
+        /// the entity yet), we fall back to the config-declared parameters so the response
+        /// stays useful.
         /// </summary>
-        /// <param name="parameters">A list of <see cref="ParameterMetadata"/> objects representing the parameters to process. Can be null.</param>
-        /// <returns>A list of dictionaries, each containing the parameter's name, whether it is required, its default
-        /// value, and its description. Returns an empty list if <paramref name="parameters"/> is null.</returns>
-        private static List<object> BuildParameterMetadataInfo(List<ParameterMetadata>? parameters)
+        /// <param name="configParameters">Parameters listed in the runtime config. Used only as a fallback when DB metadata is unavailable.</param>
+        /// <param name="databaseObject">DB metadata for the entity. Expected to be a <see cref="DatabaseStoredProcedure"/>; may be null.</param>
+        /// <returns>One dictionary per parameter with keys name, required, default, description.</returns>
```
souvikghosh04: this says returning a Dictionary but the method returns a List<object>. please review the comments here and in other places too

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
```diff
@@ -0,0 +1,341 @@
...
+
+            Assert.AreEqual(1, parameters.GetArrayLength());
+            AssertParameter(parameters, name: "locale", expectedRequired: false, expectedDefault: "en-US", expectedDescription: "Locale override");
+        }
+
+        /// <summary>
+        /// Explicit Required=false is honored even when there is no Default value supplied.
+        /// </summary>
+        [TestMethod]
+        public async Task DescribeEntities_ExplicitRequiredFalse_IsHonored_WhenNoDefault()
```
souvikghosh04: can this test and `DescribeEntities_ExplicitRequiredFalse_IsHonored` be merged in one as it looks similar? please check other tests too

souvikghosh04 (approved):
additional comments post previous review

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
anushakolan: Those `dbParameters` are post-merge shapes. That's intentional, `DescribeEntitiesTool` is only ever handed a `ParameterDefinition` that `SqlMetadataProvider/MsSqlMetadataProvider` has already merged. So these unit tests verify the projection contract given a merged input — they don't exercise the merge itself.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
anushakolan: You're right — that test isn't exercising a real production state. Tracing it through:

- `SqlMetadataProvider.cs:537-544` throws `DataApiBuilderException("Could not find parameter ... specified in config")` via `HandleOrRecordException`, which in non-validate mode (i.e. `dab start`) re-throws `SqlMetadataProvider.cs:65-75`.
- The exception bubbles through `PopulateObjectDefinitionForEntity` → `PopulateObjectDefinitionForEntities` → `InitializeAsync` and fails startup.
- `DescribeEntitiesTool` only runs after a successful startup, so the "DB metadata available + 0 DB params + non-empty config params" state can never reach it.

The test was riding on a dead branch in `BuildParameterMetadataInfo`: `dbParameters is { Count: > 0 }` falling back to config when `Count == 0`.

Combining this with your other comment ("config-only code path should not be hit"), the change is now:

- Dropped the `configParameters` parameter from `BuildParameterMetadataInfo` entirely — the method now reads only from `DatabaseStoredProcedure.StoredProcedureDefinition.Parameters`.
- Removed the config-only fallback block.
- On the impossible-state branch (`databaseObject` is null for an SP, or isn't a populated `DatabaseStoredProcedure`), the method throws `InvalidOperationException` instead of returning `[]`. The surrounding per-entity `try/catch` in `DescribeEntitiesTool` logs a warning and drops just that entity from the response.
- Same treatment at the caller site: if `McpMetadataHelper.TryResolveDatabaseObject` returns null for an SP entity, the caller throws (with the `resolveError` text) and the catch drops the entity.
- Why throw rather than return `[]`: returning the SP with `parameters=[]` is indistinguishable from a real zero-parameter SP, so the agent would happily call `execute_entity` with no arguments and hit a runtime failure. Dropping the entity makes the agent unable to call it at all, which is the correct outcome when we can't describe its inputs.
- Deleted `DescribeEntities_EmptyDbParameters_FallsBackToConfigParameters` and `DescribeEntities_FallsBackToConfigParameters_WhenDatabaseMetadataUnavailable` — both exercised the now-removed path. Added `DescribeEntities_DropsEntity_WhenStoredProcedureMetadataIsMissing` to lock in the new "drop, don't fabricate" behavior.

Net: describe_entities trusts the metadata provider's init invariant and fails loudly (per entity) rather than duplicating validation at the consumer or silently emitting misleading data.

### Trade-off: is dropping the entity acceptable?

When at least one other entity built successfully, the response is a normal success result with the broken SP simply absent — the agent has no in-payload signal, only the operator sees the `LogWarning`. When every entity failed to build (or the broken SP was the only one), `entityList.Count == 0` and the existing fall-through returns an MCP `IsError=true` result with code `NoEntitiesConfigured`, so the agent sees the error explicitly. That's the path the new test exercises.

Compared to the alternatives:

| Behavior | Agent impact | Operator visibility |
|---|---|---|
| Return SP with `parameters=[]`  | Agent calls SP with no args → runtime failure | `LogWarning` |
| Drop entity| Agent can't call SP (it's invisible) → no bad call, but no error either | `LogWarning` |
| Fail entire describe_entities call | Agent gets clear error but loses access to all other entities | `LogWarning` + error result |

Dropping is the middle ground: prevents the wrong action (no misleading metadata for the agent to act on) while preserving useful output (other entities still work). The cost is that the agent is unaware the dropped SP even exists — for an agent asking "is there a `GetOrders` SP?", it'll answer "no" instead of "yes but I can't tell you about it." If we later decide that's too quiet, the cheap follow-up is to surface a `failedEntities` array alongside `entities` so the agent can read the list and tell the user "exists but metadata couldn't be loaded." Happy to do that here or as a follow-up — your call.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
anushakolan: Got it — implemented. `BuildParameterMetadataInfo` no longer takes `configParameters` at all; it reads only from `DatabaseStoredProcedure.StoredProcedureDefinition.Parameters` and throws `InvalidOperationException` if that's missing for an SP entity (since, as you pointed out, `FillSchemaForStoredProcedureAsync` would have already failed startup). The surrounding per-entity try/catch then drops just that entity from the response. Tests for the deleted config-only branch are gone; added one that locks in the new drop-on-missing-metadata behavior.

src/Service.Tests/Mcp/DescribeEntitiesStoredProcedureParametersTests.cs
Aniruddh25: Dropping entity seems fine. This case wont arise anyway so no need to overengineer this.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/DescribeEntitiesTool.cs
Aniruddh25: will approve when I see the implementation changes.

Aniruddh25 (approved):

Aniruddh25 (approved):

PR #3426, opened 2026-04-10
Fix CLI overrides LogLevel value from config file

### LINKED ISSUE #3258: [Bug]: `log-level` seems to let `default` override other values.
Image

RubenCerna2079:
## Why make this change?
- Closes issue #3258
Whenever we use the `dab start` command the CLI overrides the minimum LogLevel from the configuration file. This is acceptable only if we use the `--LogLevel` flag. If we do not use that flag, the LogLevel for each namespace should be decided by the configuration file.

## What is this change?
This change moves the `args.Add` methods inside the `ConfigGenerator.cs` so that they are only applied when we use the `--LogLevel` flag. Having these arguments causes DAB to not allow any changes to the LogLevel in the loggers, since those arguments were always being added, DAB always assumed the CLI was overriding the LogLevel when it was not expected as it would use the `--LogLevel` flag to determine if it was an `IsLogLevelOverridenByCli` scenario.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests
- [x] Manual Tests
Tested with different configurations of the namespaces in the `log-level` property inside the config file.
Note: ALL these manual tests were tested by running the CLI as well as running DAB through Visual Studio.
Config file includes:
```
"log-level": {
    "Azure.DataApiBuilder.Core.Services.ISqlMetadataProvider": "Information",
    "Azure.DataApiBuilder.Core": "Debug",
    "Azure.DataApiBuilder.Config.FileSystemRuntimeConfigLoader": "Error",
    "default": "Warning"
}
```
Expected results:
- All the logs from `Azure.DataApiBuilder.Core.Services.ISqlMetadataProvider` will provide logs from log level information and above.
- All logs from `Azure.DataApiBuilder.Core` and its sub-namespaces that are not from `ISqlMetadataProvider` will provide logs from log level debug and above.
- All logs from `Azure.DataApiBuilder.Config.FileSystemRuntimeConfigLoader` will provide logs from log level error and above.
- Everything else will provide logs from log level warning and above, including CLI.

Config file includes:
```
"log-level": {
    "Azure.DataApiBuilder.Core": "Information",
    "Azure.DataApiBuilder.Config": "Debug",
}
```
Expected results:
- All logs from `Azure.DataApiBuilder.Core` and its sub-namespaces will provide logs from log level information and above.
- All logs from `Azure.DataApiBuilder.Config` and its sub-namespaces will provide logs from log level debug and above.
- Everything else will provide logs from log level debug and above with `host.mode = development`, including CLI.

Config file includes:
```
"log-level": {
    "Default: none
}
```
Expected results:
- No logs will be printed, including CLI.

## Sample Request(s)
dab start --LogLevel information

dab start (With config file default value debug)


src/Cli/ConfigGenerator.cs
```diff
@@ -2597,18 +2597,27 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
                 _logger.LogInformation("Setting minimum LogLevel: {minimumLogLevel}.", minimumLogLevel);
+                args.Add("--LogLevel");
+                args.Add(minimumLogLevel.ToString());
             }
             else
             {
                 minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
                 HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
+                bool isValueFromDefault = !deserializedRuntimeConfig.IsLogLevelNull() && deserializedRuntimeConfig.Runtime!.Telemetry!.LoggerLevel!
+                    .SingleOrDefault(kvp => kvp.Key.Equals("default", StringComparison.OrdinalIgnoreCase)).Value != null;
```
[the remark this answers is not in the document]
RubenCerna2079: We only want to allow one `default` in the config file, so it should stay as it is.

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -857,6 +858,80 @@ public void TestEngineStartUpWithVerboseAndLogLevelOptions(string logLevelOption
...
+        // Run TryStartEngineWithOptions on a background task because StartEngine blocks until the host shuts down.
+        Task engineTask = Task.Run(() =>
+            TryStartEngineWithOptions(options, _runtimeConfigLoader!, _fileSystem!));
+
+        // Wait for the engine to finish loading the config.
+        await Task.WhenAny(engineTask, Task.Delay(TimeSpan.FromSeconds(5)));
+
+        Assert.AreEqual(expectedIsOverridden, Startup.IsLogLevelOverriddenByCli);
+    }
+
```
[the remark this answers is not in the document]
RubenCerna2079: Changed it to instead only use Task.Delay

src/Cli/ConfigGenerator.cs
```diff
@@ -2597,18 +2597,27 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
 
-                _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
+                // Check if we set the minimum LogLevel from the config file or if it is based on the host mode.
```
Aniruddh25: we are not using the `hostModeType` at all to make the decision of what the level should be set as

src/Cli/ConfigGenerator.cs
```diff
@@ -2597,18 +2597,27 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
                 minimumLogLevel = (LogLevel)options.LogLevel;
                 _logger.LogInformation("Setting minimum LogLevel: {minimumLogLevel}.", minimumLogLevel);
+                args.Add("--LogLevel");
+                args.Add(minimumLogLevel.ToString());
             }
             else
             {
                 minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
                 HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
+                bool isValueFromDefault = !deserializedRuntimeConfig.IsLogLevelNull() && deserializedRuntimeConfig.Runtime!.Telemetry!.LoggerLevel!
```
Aniruddh25: how is the function GetConfiguredLogLevel getting the minimumLogLevel?

isnt it from the config file?

if it is obtained in a similar manner, `isValueFromDefault` should be a by product of the GetConfiguredLevel function.

src/Cli/ConfigGenerator.cs
```diff
@@ -2597,18 +2597,27 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
 
-                _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
+                // Check if we set the minimum LogLevel from the config file or if it is based on the host mode.
+                if (isValueFromDefault)
+                {
+                    _logger.LogInformation("Using default minimum LogLevel: {minimumLogLevel} from the 'Default' namespace in config file.", minimumLogLevel);
+                }
+                else
+                {
+                    _logger.LogInformation("Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
```
Aniruddh25: What are we doing in the else part to set the minimum log level? We do a log message here but what is setting the loglevel in the LogLevel provider?

src/Cli.Tests/EndToEndTests.cs
```diff
@@ -857,6 +858,80 @@ public void TestEngineStartUpWithVerboseAndLogLevelOptions(string logLevelOption
...
+                        ""origins"": [],
+                        ""allow-credentials"": false
+                    },
+                    ""authentication"": {
+                        ""provider"": ""Unauthenticated""
+                    }
+                },
+                ""telemetry"": {
+                    ""log-level"": {
+                        ""default"": ""Warning""
```
Aniruddh25, line 909: Have you tested whether different namespaces within DAB honor the different log levels? when the loglevel from CLI is not overriding the level? 

This config is only testing the default namespace scenario,

src/Cli/ConfigGenerator.cs
```diff
@@ -2597,18 +2597,27 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
             else
             {
                 minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
                 HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
+                bool isValueFromDefault = !deserializedRuntimeConfig.IsLogLevelNull() && deserializedRuntimeConfig.Runtime!.Telemetry!.LoggerLevel!
+                    .SingleOrDefault(kvp => kvp.Key.Equals("default", StringComparison.OrdinalIgnoreCase)).Value != null;
 
-                _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
+                // Check if we set the minimum LogLevel from the config file or if it is based on the host mode.
+                if (isValueFromDefault)
```
Aniruddh25: what is the use of figuring out `isValueFromDefault`?

Aniruddh25 (changes requested):
Needs answers to some questions, and more testing

src/Cli.Tests/EndToEndTests.cs
RubenCerna2079: I tested it locally, I will add a few different namespaces to honor the different log levels.

src/Cli/ConfigGenerator.cs
RubenCerna2079: Since we don't use it here directly, it has been removed

src/Cli/ConfigGenerator.cs
RubenCerna2079: It is not necessary to obtain this information, it has been removed

src/Cli/ConfigGenerator.cs
RubenCerna2079: We still need to set the minimum log level for the PR https://github.com/Azure/data-api-builder/pull/3318 as the log level for the CLI loggers still needs to be set based on the configuration file. Although making those changes is not in the scope of this PR.

src/Cli/ConfigGenerator.cs
RubenCerna2079: It is not necessary to figure out this value, it has been removed

anushakolan:
Apart from the unit tests, did you manually test it via CLI, if the behavior is as expected. It would be good to have the manual testing section as well in the PR description.

src/Cli/ConfigGenerator.cs
```diff
@@ -2603,18 +2603,17 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
                 minimumLogLevel = (LogLevel)options.LogLevel;
-                _logger.LogInformation("Setting minimum LogLevel: {minimumLogLevel}.", minimumLogLevel);
+                args.Add("--LogLevel");
+                args.Add(minimumLogLevel.ToString());
             }
             else
             {
+                // GetConfiguredLogLevel will only return the 'Default' LogLevel value from the config file.
+                // If it doesn't exist, it will return the default LogLevel based on the environment (Debug for Development and Error for Production).
                 minimumLogLevel = deserializedRuntimeConfig.GetConfiguredLogLevel();
```
Aniruddh25: why do we get the `minimumLogLevel` here in the else part? There should not be any else part altogether. If you need if for the other PR, keep it separate and add it in the other PR - dont see a need in this PR.

src/Cli/ConfigGenerator.cs
```diff
@@ -2603,18 +2603,17 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
-                HostMode hostModeType = deserializedRuntimeConfig.IsDevelopmentMode() ? HostMode.Development : HostMode.Production;
-
-                _logger.LogInformation($"Setting default minimum LogLevel: {minimumLogLevel} for {hostModeType} mode.", minimumLogLevel, hostModeType);
             }
 
-            args.Add("--LogLevel");
-            args.Add(minimumLogLevel.ToString());
+            _logger.LogInformation("Setting minimum LogLevel: {minimumLogLevel}.", minimumLogLevel);
```
Aniruddh25: This log information can stay only in the if section when we add the LogLevel argument

Aniruddh25 (approved):
Approving assuming the `else` part will be removed completely.

Aniruddh25 (approved):

anushakolan:
> Apart from the unit tests, did you manually test it via CLI, if the behavior is as expected. It would be good to have the manual testing section as well in the PR description.

I will re-review after this is addressed. Some cases specific to namespace logging testing is also good to add.

anushakolan (approved):

Issue #3331, opened 2026-03-27
[Enh]: Embedding Phase 3: Parameter Substitution


JerryNixon:
## What

Add `auto-embed` to stored procedure parameter configuration.

When enabled, DAB accepts a string value from the caller, sends that value to the configured embedding subsystem, receives a stringified embedding, then passes that string to the stored procedure parameter.

This removes the burden of real-time embedding from the app and database. The database still owns vector conversion, vector validation, and semantic search behavior.

## Why

Stored procedures often encapsulate search logic. When a stored procedure needs a vector, the caller or database currently has to create the embedding.

`auto-embed` moves that step into DAB.

The API contract stays simple:

```text
client sends text
DAB embeds text
stored procedure receives embedding string
database handles the rest
```

## Dependency

This feature depends on the runtime embedding subsystem introduced by PR #3104.

Assume that subsystem is merged before implementation begins.

`auto-embed` consumes the embedding subsystem. It doesn't define a provider contract, embedding endpoint, health model, retry model, or embedding response shape.

The embedding subsystem provides:

```text
runtime.embeddings configuration
internal embedding service
single text embedding support
batch text embedding support
embedding timeout handling
embedding telemetry
embedding validation through RuntimeConfigValidator
```

Implementation rule:

```text
If runtime.embeddings is valid, auto-embed can use it.
If runtime.embeddings is invalid, DAB validation or startup fails before auto-embed runs.
```

## Non-goals

This feature doesn't add semantic search to DAB.

This feature doesn't add native vector type support.

This feature doesn't validate vectors in DAB.

This feature doesn't support client-provided vectors.

This feature doesn't support table or view fields.

This feature doesn't support arrays or objects.

This feature doesn't change normal stored procedure parameter behavior.

This feature doesn't add a new embedding health check.

This feature doesn't add a new embedding timeout setting.

This feature doesn't add a new embedding concurrency setting.

## Configuration

`auto-embed` is a child property of a stored procedure parameter object.

The current schema already models stored procedure parameters as an array of named parameter objects. `name` is already required.

Example:

```json
{
  "entities": {
    "SearchProducts": {
      "source": {
        "object": "dbo.SearchProducts",
        "type": "stored-procedure",
        "parameters": [
          {
            "name": "searchText",
            "required": true,
            "auto-embed": true,
            "description": "Text to embed before procedure execution."
          }
        ]
      },
      "permissions": [
        {
          "role": "anonymous",
          "actions": [
            "execute"
[...]

PR #3441, opened 2026-04-12
Enable internal text embedding API 

ajtiwari07:
**Summary**
This PR adds configurable text chunking capabilities to the embeddings API, enabling automatic text segmentation before embedding generation. This feature supports both single-text and multi-document batch processing with runtime configuration and query parameter overrides.

**Changes**
Configuration 
Added EmbeddingsChunkingOptions.cs - Configuration model for chunking behavior
Enabled (bool) - Enable/disable chunking
SizeChars (int) - Chunk size in characters (default: 800)
OverlapChars (int) - Overlap between chunks (default: 100)
EffectiveSizeChars property ensures minimum valid chunk size
Modified EmbeddingsOptions.cs - Added Chunking property and IsChunkingEnabled helper
Removed EmbeddingsCacheOptions.cs - Simplified configuration by removing unused cache feature

**API Enhancements** 
Modified Controllers/EmbeddingController.cs
Auto-detects request type (single text vs. document array)
Implements overlapping text chunking algorithm
Supports query parameter overrides: $chunking.enabled, $chunking.size-chars, $chunking.overlap-chars
Returns multiple embeddings per document when chunking is enabled
Added Models/EmbedDocumentRequest.cs - Request model for document arrays
Added Models/EmbedDocumentResponse.cs - Response model with chunked embeddings
Schema ([schemas](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html))
Modified dab.draft.schema.json - Added chunking configuration schema with validation rules
Testing ([UnitTests](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html))
Added EmbeddingsChunkingOptionsTests.cs (13 tests) - Configuration validation
Added ChunkTextTests.cs (21 tests) - Chunking algorithm validation including edge cases
Modified EmbeddingControllerTests.cs (+18 tests) - API endpoint tests for chunking and document arrays
Total Test Coverage: 72 tests (48 existing + 24 new) - All passing## Why make this change?

**Testing**
All 72 unit tests passing
Edge cases covered: empty text, very small chunks, overlap larger than chunk size, Unicode text
Query parameter parsing validated
Backward compatibility verified

**Breaking Changes**
None - This is a backward-compatible addition. Existing single-text requests continue to work without modification.


src/Service/Controllers/EmbeddingController.cs
```diff
@@ -150,14 +256,102 @@ public async Task<IActionResult> PostAsync()
...
+        {
+            int remainingLength = text.Length - position;
+            int currentChunkSize = Math.Min(chunkSize, remainingLength);
+
+            chunks.Add(text.Substring(position, currentChunkSize));
+
+            // Move position forward by (chunkSize - overlap) to create overlapping chunks
+            position += chunkSize - overlap;
+
+            // Prevent infinite loop if overlap >= chunkSize
```
JimRoberts-MS: should this just be handled by validation in the options constructor?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -150,14 +256,102 @@ public async Task<IActionResult> PostAsync()
...
+
+        while (position < text.Length)
+        {
+            int remainingLength = text.Length - position;
+            int currentChunkSize = Math.Min(chunkSize, remainingLength);
+
+            chunks.Add(text.Substring(position, currentChunkSize));
+
+            // Move position forward by (chunkSize - overlap) to create overlapping chunks
+            position += chunkSize - overlap;
```
JimRoberts-MS: should we walk the chunk size back to a word (/space) boundary?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+    /// </summary>
+    private string[] ChunkText(string text, EmbeddingsChunkingOptions? chunkingOptions)
+    {
+        // If chunking is disabled or options are null, return text as single chunk
+        if (chunkingOptions is null || !chunkingOptions.Enabled)
+        {
+            return new[] { text };
+        }
+
+        int chunkSize = chunkingOptions.SizeChars;
```
prshri-msft: should we use this?
`int chunkSize = chunkingOptions.EffectiveSizeChars;`

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+    /// Supports query parameters to override chunking settings.
+    /// Default response is JSON: { "embedding": [...], "dimensions": N } for single text,
+    /// or [{ "key": "...", "data": [[...], [...]] }] for document arrays.
+    /// </summary>
+    /// <returns>Embedding vector(s) as JSON, or an error response.</returns>
+    [HttpPost]
+    [Route("embed")]
+    [Consumes("text/plain", "application/json")]
+    [Produces("application/json", "text/plain")]
+    public async Task<IActionResult> PostAsync()
```
prshri-msft: shall we pass `CancellationToken` ?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+            }
+            catch (JsonException)
+            {
+                // Not a JSON string, use requestBody as-is
+                _logger.LogDebug("Request body is not a JSON string, using as plain text.");
+            }
+        }
+
+        // Handle as single text (backward compatible)
+        return await ProcessSingleTextAsync(requestBody);
```
prshri-msft: a json object not adhering to document array format might fall into this code path, example: `{"foo":"bar"}`

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+                foreach (string chunk in chunks)
+                {
+                    EmbeddingResult result = await _embeddingService!.TryEmbedAsync(chunk);
+
+                    if (!result.Success || result.Embedding is null)
+                    {
+                        _logger.LogError("Failed to embed chunk for document key '{Key}': {Error}", doc.Key, result.ErrorMessage);
+                        return StatusCode(
+                            (int)HttpStatusCode.InternalServerError,
+                            $"Failed to generate embedding for document '{doc.Key}': {result.ErrorMessage}");
```
prshri-msft: what would be the contents of `ErrorMessage`? Should they be surface to the user?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+            }
+        }
+
+        return Ok(responses.ToArray());
+    }
+
+    /// <summary>
+    /// Processes a single text request and returns embedding (backward compatible).
+    /// </summary>
+    private async Task<IActionResult> ProcessSingleTextAsync(string text)
```
prshri-msft: do we need to apply chunking in here as well?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+        {
+            if (bool.TryParse(enabledValue, out bool parsedEnabled))
+            {
+                enabled = parsedEnabled;
+            }
+        }
+
+        if (Request.Query.TryGetValue("$chunking.size-chars", out StringValues sizeValue))
+        {
+            if (int.TryParse(sizeValue, out int size) && size > 0)
```
prshri-msft, line 357: do we need to surface incorrect param values to the end user or can we continue falling back to default values?

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,582 @@
...
+        StringBuilder cacheKeyBuilder = new();
+        cacheKeyBuilder.Append(CACHE_KEY_PREFIX);
+        cacheKeyBuilder.Append(KEY_DELIMITER);
+        cacheKeyBuilder.Append(_providerName);
+        cacheKeyBuilder.Append(KEY_DELIMITER);
+        cacheKeyBuilder.Append(model);
+        cacheKeyBuilder.Append(KEY_DELIMITER);
+        cacheKeyBuilder.Append(hashHex);
+
+        return cacheKeyBuilder.ToString();
```
prshri-msft, line 501: we should also include the dimension of the embedding model to the key

src/Service.Tests/UnitTests/ChunkTextTests.cs
```diff
@@ -0,0 +1,349 @@
...
+            int step = chunkSize - overlap;
+            if (step <= 0)
+            {
+                // Prevent infinite loop: if overlap >= chunkSize, move forward by at least 1
+                step = Math.Max(1, chunkSize);
+            }
+            position += step;
+        }
+
+        return chunks;
```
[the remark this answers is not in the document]
prshri-msft: +1 to this

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,582 @@
...
+    /// <summary>
+    /// Configures the HTTP client with timeout and authentication headers.
+    /// </summary>
+    private void ConfigureHttpClient()
+    {
+        _httpClient.Timeout = TimeSpan.FromMilliseconds(_options.EffectiveTimeoutMs);
+
+        if (_options.Provider == EmbeddingProviderType.AzureOpenAI)
+        {
+            _httpClient.DefaultRequestHeaders.Add("api-key", _options.ApiKey);
```
JimRoberts-MS, line 104: can we also support Entra token based auth here (or when the request is made so we don't end up with an expired token)

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,582 @@
...
+    /// <inheritdoc/>
+    public async Task<EmbeddingBatchResult> TryEmbedBatchAsync(string[] texts, CancellationToken cancellationToken = default)
+    {
+        if (!_options.Enabled)
+        {
+            _logger.LogDebug("Embedding service is disabled, skipping batch embed request");
+            return new EmbeddingBatchResult(false, null, "Embedding service is disabled.");
+        }
+
+        if (texts is null || texts.Length == 0)
```
JimRoberts-MS: looks like we might also want to check something like `texts.Any(string.IsNullOrEmpty)`

JimRoberts-MS: and/or in EmbedBatchAsync

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,582 @@
...
+        {
+            throw new ArgumentException(
+                $"Texts array exceeds max supported batch size of {MAX_BATCH_TEXT_COUNT}.",
+                nameof(texts));
+        }
+
+        // For batch, check cache for each text individually
+        string[] cacheKeys = texts.Select(CreateCacheKey).ToArray();
+        float[]?[] results = new float[texts.Length][];
+        List<int> uncachedIndices = new();
```
JimRoberts-MS, line 336: Would it be better to use a Dictionary in case there are some duplicate keys to avoid duplicate calls to the api?

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,582 @@
...
+            {
+                uncachedIndices.Add(i);
+                EmbeddingTelemetryHelper.TrackCacheMiss(_providerName);
+            }
+        }
+
+        // If all texts were cached, return immediately
+        if (uncachedIndices.Count == 0)
+        {
+            return results!;
```
JimRoberts-MS, line 362: would add a debug log here too

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,582 @@
...
+            int originalIndex = uncachedIndices[i];
+            results[originalIndex] = apiResults[i];
+
+            // Store embeddings using the configured FusionCache stack.
+            _cache.Set(
+                key: cacheKeys[originalIndex],
+                value: apiResults[i],
+                options =>
+                {
+                    options.SetDuration(TimeSpan.FromHours(DEFAULT_CACHE_TTL_HOURS));
```
JimRoberts-MS, line 423: should the ttl be reset on cache hits too?

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,582 @@
...
+        string requestJson = JsonSerializer.Serialize(requestBody, _jsonSerializerOptions);
+        using HttpContent content = new StringContent(requestJson, Encoding.UTF8, "application/json");
+
+        _logger.LogDebug("Sending embedding request to {Url} with {Count} text(s)", requestUrl, texts.Length);
+
+        using HttpResponseMessage response = await _httpClient.PostAsync(requestUrl, content, cancellationToken);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            string errorContent = await response.Content.ReadAsStringAsync(cancellationToken);
```
JimRoberts-MS, line 521: is it possible that there is sensitive info in this response?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -157,7 +170,8 @@ private async Task UpdateHealthCheckDetailsAsync(ComprehensiveHealthCheckReport
         {
             comprehensiveHealthCheckReport.Checks = new List<HealthCheckResultEntry>();
             await UpdateDataSourceHealthCheckResultsAsync(comprehensiveHealthCheckReport, runtimeConfig);
-            await UpdateEntityHealthCheckResultsAsync(comprehensiveHealthCheckReport, runtimeConfig, roleHeader, roleToken);
```
sayalikudale, line 160: can we add parameters , roleHeader, roleToken back here?

src/Config/Converters/EmbeddingsOptionsConverterFactory.cs
```diff
@@ -0,0 +1,299 @@
...
+                    case "timeout-ms":
+                        timeoutMs = JsonSerializer.Deserialize<int?>(ref reader, options);
+                        break;
+                    case "endpoint":
+                        endpoint = ReadEndpointOptions(ref reader, options);
+                        break;
+                    case "health":
+                        health = ReadHealthCheckConfig(ref reader, options);
+                        break;
+                    default:
```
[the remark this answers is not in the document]
sayalikudale: we might need to add chunking case here??

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+        {
+            _logger.LogWarning("Embedding endpoint called but embedding service is not available or disabled.");
+            return StatusCode((int)HttpStatusCode.ServiceUnavailable, "Embedding service is not available.");
+        }
+
+        // Check authorization
+        bool isDevelopmentMode = _runtimeConfigProvider.GetConfig()?.Runtime?.Host?.Mode == HostMode.Development;
+        string clientRole = GetClientRole();
+
+        if (!endpointOptions.IsRoleAllowed(clientRole, isDevelopmentMode))
```
sayalikudale, line 95: Is it possible to use the existing IAuthorizationResolver or ClientRoleHeaderAuthenticationMiddleware in this controller rather than implementing new  role-checking logic   (GetClientRole() + IsRoleAllowed())

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        // Create new options with query parameters (using defaults for unspecified values)
+        return new EmbeddingsChunkingOptions(enabled, sizeChars, overlapChars);
+    }
+
+    /// <summary>
+    /// Splits text into chunks if chunking is enabled and text exceeds chunk size.
+    /// </summary>
+    private string[] ChunkText(string text, EmbeddingsChunkingOptions? chunkingOptions)
```
sayalikudale: Should chunking be moved to EmbeddingService or a dedicated ITextChunker service?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+                    }
+
+                    embeddings.Add(result.Embedding);
+                }
+
+                responses.Add(new EmbedDocumentResponse(doc.Key, embeddings.ToArray()));
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error processing document with key '{Key}'", doc.Key);
```
sayalikudale: Do we want to expose keys in logs?

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: I don't think that is required, we are already doing a hash of the text.

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -350,5 +364,108 @@ private async Task PopulateEntityHealthAsync(ComprehensiveHealthCheckReport comp
...
+
+            string testText = healthConfig.TestText;
+            int thresholdMs = healthConfig.ThresholdMs;
+            int? expectedDimensions = healthConfig.ExpectedDimensions;
+
+            try
+            {
+                Stopwatch stopwatch = new();
+                stopwatch.Start();
+                EmbeddingResult result = await _embeddingService.TryEmbedAsync(testText);
```
JimRoberts-MS, line 394: I wonder if it would be useful to separate out errors from the cache from errors calling the underlying embedding api.. eg you might have cache hits for the testText but be unable to try getting embeddings for new strings from the service, or you might be able to get new embeddings but not able to cache them causing perf issues

src/Config/Converters/EmbeddingsOptionsConverterFactory.cs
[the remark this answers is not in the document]
prshri-msft: deserialization is causing type mismatch due to nullables. we might need to add some more handling for nullable types in here

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: the current approach is appropriate because IAuthorizationResolver is entity/operation-level authorization, not endpoint access control; using the role header constant is correct here.

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: not completely sure what would be the upside to it

src/Config/Converters/EmbeddingsOptionsConverterFactory.cs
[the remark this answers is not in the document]
ajtiwari07: Added chunking. I have addresses nullable in tests but didn't find any of them here, can you share an example?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,385 @@
...
+                // Not a JSON string, use requestBody as-is
+                _logger.LogDebug("Request body is not a JSON string, using as plain text.");
+            }
+        }
+
+        // Handle as single text (backward compatible)
+        return await ProcessSingleTextAsync(requestBody);
+    }
+
+    /// <summary>
```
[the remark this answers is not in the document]
ajtiwari07: Done

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Done

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Updated

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Added

src/Service/HealthCheck/HealthCheckHelper.cs
ajtiwari07: Added

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: It would be better not to swallow invalid parameters

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,391 @@
...
+            return BadRequest("Request body cannot be empty.");
+        }
+
+        // Generate embedding
+        EmbeddingResult result = await _embeddingService!.TryEmbedAsync(text, cancellationToken);
+
+        if (!result.Success)
+        {
+            _logger.LogError("Embedding request failed: {Error}", result.ErrorMessage);
+            return StatusCode((int)HttpStatusCode.InternalServerError, result.ErrorMessage ?? "Failed to generate embedding.");
```
sayalikudale: Do we need to return result.ErrorMessage as well here? Will status code and generic message not enough to share with customer?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,391 @@
...
+                }
+
+                responses.Add(new EmbedDocumentResponse(doc.Key, batchResult.Embeddings));
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error processing document.");
+                return StatusCode(
+                    (int)HttpStatusCode.InternalServerError,
+                    $"Error processing document: {ex.Message}");
```
sayalikudale: Same here as well. will ex.Message have the internal details which may not be okay to share with end user?

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: It would be good to return a idiomatic error message for the type of error like json parsing, invalid request body etc.

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Updated to Standardized API Error response

src/Service/Controllers/EmbeddingController.cs
JimRoberts-MS: will be done by the clients

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Entra Auth support is separate implementation (PMs were discussing from last i heard), we can handle it in that PR.

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: This comment is outdated; latest implementation returns parameter validation errors to the main method that returns standard ERROR response.

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Can you elaborate? We do a GET on each chunk

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Yes, ensure more hits don't TTL out.

src/Service/HealthCheck/HealthCheckHelper.cs
ajtiwari07: Can you revisit this again with the latest iteration?

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Done

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Done

schemas/dab.draft.schema.json
```diff
@@ -763,6 +763,164 @@
               "default": 4
             }
           }
+        },
+        "embeddings": {
```
souvikghosh04, line 774: nit- if embedings are specifically Open AI specific, should we rename it like `openai-embeddings`? or it can be extended to other providers? based on the current JSON schema, it looks very specific to Open AI and little room to add other provider in future

schemas/dab.draft.schema.json
```diff
@@ -763,6 +763,164 @@
...
+            },
+            "chunking": {
+              "type": "object",
+              "description": "Chunking configuration for text processing before embedding. Used to split large text inputs into smaller chunks.",
+              "additionalProperties": false,
+              "properties": {
+                "enabled": {
+                  "type": "boolean",
+                  "description": "Whether chunking is enabled. Defaults to false.",
+                  "default": false
```
souvikghosh04: should this be enabled (true) by default?

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -1591,6 +1592,263 @@ public void TestUpdateUserDelegatedAuthDatabaseAudience()
             Assert.AreEqual("EntraId", (string?)userDelegatedAuthSection["provider"]);
         }
 
+        /// <summary>
+        /// Tests that running "dab configure" with embeddings endpoint options on a config with existing embeddings
+        /// results in the endpoint options being added to the embeddings configuration.
+        /// </summary>
+        [TestMethod]
```
souvikghosh04, line 1694: there seems some scope to refactor the tests. move duplicate/common code inside helpers, collapse multiple similar `TestMethod` into `DataTestMethod` and use `DataRow` to pass different parameters wherever possible.

src/Cli.Tests/ConfigureOptionsTests.cs
```diff
@@ -1591,6 +1592,263 @@ public void TestUpdateUserDelegatedAuthDatabaseAudience()
...
+            string updatedConfig = _fileSystem!.File.ReadAllText(TEST_RUNTIME_CONFIG_FILE);
+            Assert.IsTrue(RuntimeConfigLoader.TryParseConfig(updatedConfig, out RuntimeConfig? updatedRuntimeConfig));
+            Assert.IsNotNull(updatedRuntimeConfig.Runtime?.Embeddings);
+            Assert.IsNotNull(updatedRuntimeConfig.Runtime.Embeddings.Endpoint);
+            Assert.IsTrue(updatedRuntimeConfig.Runtime.Embeddings.Endpoint.Enabled);
+            Assert.IsNotNull(updatedRuntimeConfig.Runtime.Embeddings.Endpoint.Roles);
+            CollectionAssert.AreEqual(new[] { "admin", "reader" }, updatedRuntimeConfig.Runtime.Embeddings.Endpoint.Roles);
+            // Verify base embeddings settings are preserved
+            Assert.AreEqual(EmbeddingProviderType.AzureOpenAI, updatedRuntimeConfig.Runtime.Embeddings.Provider);
+            Assert.AreEqual("https://myservice.openai.azure.com", updatedRuntimeConfig.Runtime.Embeddings.BaseUrl);
```
souvikghosh04: move this to constant or helper instead of repeating multiple places

schemas/dab.draft.schema.json
```diff
@@ -763,6 +763,164 @@
...
+              "description": "The model or deployment name. Required for Azure OpenAI (deployment name). For OpenAI, defaults to 'text-embedding-3-small' if not specified."
+            },
+            "api-version": {
+              "type": "string",
+              "description": "Azure API version. Only used for Azure OpenAI provider.",
+              "default": "2024-02-01"
+            },
+            "dimensions": {
+              "type": "integer",
+              "description": "Output vector dimensions. Optional, uses model default if not specified. Useful for Redis schema alignment.",
```
souvikghosh04: what is the role of Redis here?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -350,5 +364,108 @@ private async Task PopulateEntityHealthAsync(ComprehensiveHealthCheckReport comp
...
+                stopwatch.Stop();
+
+                int responseTimeMs = (int)stopwatch.ElapsedMilliseconds;
+                bool isResponseTimeWithinThreshold = responseTimeMs <= thresholdMs;
+                bool isDimensionsValid = true;
+                string? errorMessage = null;
+
+                if (!result.Success)
+                {
+                    errorMessage = result.ErrorMessage ?? "Embedding request failed.";
```
souvikghosh04, line 404: may add appropriate, exponential retries upto 3?

src/Service/HealthCheck/HealthCheckHelper.cs
```diff
@@ -350,5 +364,108 @@ private async Task PopulateEntityHealthAsync(ComprehensiveHealthCheckReport comp
...
+            try
+            {
+                Stopwatch stopwatch = new();
+                stopwatch.Start();
+                EmbeddingResult result = await _embeddingService.TryEmbedAsync(testText);
+                stopwatch.Stop();
+
+                int responseTimeMs = (int)stopwatch.ElapsedMilliseconds;
+                bool isResponseTimeWithinThreshold = responseTimeMs <= thresholdMs;
+                bool isDimensionsValid = true;
```
souvikghosh04: should be false by default?

src/Service.Tests/UnitTests/ChunkTextTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        int overlap = 2;
+
+        // Act
+        List<string> chunks = ChunkText(text, chunkSize, overlap);
+
+        // Assert
+        Assert.IsTrue(chunks.Count > 0);
+        string reconstructedStart = chunks[0];
+        Assert.IsTrue(reconstructedStart.Contains("Hello") || reconstructedStart.Contains("世"),
+            "Should preserve Unicode characters");
```
souvikghosh04: does this assert the emoji 🌍 as set in `string text = "Hello 世界 🌍 Émoji";`?

souvikghosh04 (commented):
Posting comments so far. My review is still in progress. also, waiting for existing review comments to be addressed.

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Done

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Updated to a hardcoded error message.

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Removed

schemas/dab.draft.schema.json
ajtiwari07: Ah these configurations are part of the spec defined by PMs.

schemas/dab.draft.schema.json
ajtiwari07: done

src/Service.Tests/UnitTests/EmbeddingsOptionsTests.cs
```diff
@@ -0,0 +1,385 @@
...
+    {
+        ""$schema"": ""https://github.com/Azure/data-api-builder/releases/download/vmajor.minor.patch/dab.draft.schema.json"",
+        ""data-source"": {
+            ""database-type"": ""mssql"",
+            ""connection-string"": ""Server=test;Database=test;""
+        },
+        ""runtime"": {
+            ""embeddings"": {
+                ""provider"": ""azure-openai"",
+                ""base-url"": ""https://my-openai.openai.azure.com"",
```
souvikghosh04, line 70: urls and other configurations, including repeated code should be put in a common place. assertions in below tests also has similar hardcoded values. keep a single source of truth and reference.

src/Config/ObjectModel/Embeddings/EmbeddingsHealthCheckConfig.cs
```diff
@@ -0,0 +1,112 @@
...
+/// <summary>
+/// Health check configuration for embeddings.
+/// Validates that the embedding service is responding within threshold and returning expected results.
+/// </summary>
+public record EmbeddingsHealthCheckConfig : HealthCheckConfig
+{
+    /// <summary>
+    /// Default threshold for embedding health check in milliseconds.
+    /// </summary>
+    public const int DEFAULT_THRESHOLD_MS = 5000;
```
souvikghosh04: dab schema JSON defaults to 1000, but EmbeddingsHealthCheckConfig.DEFAULT_THRESHOLD_MS = 5000. A user omitting threshold-ms will get 5000ms at runtime but schema documentation/validation says 1000. make it consistent.

schemas/dab.draft.schema.json
```diff
@@ -763,6 +763,195 @@
...
+                },
+                "required": ["provider"]
+              },
+              "then": {
+                "required": ["model"],
+                "properties": {
+                  "api-version": {
+                    "type": "string",
+                    "description": "Azure API version. Required for Azure OpenAI provider.",
+                    "default": "2024-02-01"
```
souvikghosh04: is this version different than what you mentioned above `"default": "2023-05-15"`? please keep it consistent across the code based, including C# files and tests.

souvikghosh04 (commented):
posting additional comments. there are several inconsistencies between the schema JSON and the internal C# code files, including tests. e.g. threshold ms, API version, roles etc. are few to name which differes in schema JSON and the internal C# files. I will wait for these to get addressed, including pending comments.

schemas/dab.draft.schema.json
ajtiwari07: Redis is factored in phase 2 where we will cache these embeddings in Redis as a L2.

src/Service/HealthCheck/HealthCheckHelper.cs
ajtiwari07: Validation logic verifies dimension of the embeddings returned by the API, so it doesn't really matter.

src/Service/HealthCheck/HealthCheckHelper.cs
ajtiwari07: Good point, the reason to avoid doing retry is to avoid incurring additional latencies especially when we expect embedding creation as a precursor to SQL query.

src/Service.Tests/UnitTests/ChunkTextTests.cs
ajtiwari07: Done

src/Cli.Tests/ConfigureOptionsTests.cs
ajtiwari07: DOne

src/Cli.Tests/ConfigureOptionsTests.cs
ajtiwari07: Refactored

src/Service.Tests/UnitTests/EmbeddingsOptionsTests.cs
ajtiwari07: Modified.

src/Config/ObjectModel/Embeddings/EmbeddingsHealthCheckConfig.cs
ajtiwari07: Modified

schemas/dab.draft.schema.json
ajtiwari07: Modified

ajtiwari07:
> posting additional comments. there are several inconsistencies between the schema JSON and the internal C# code files, including tests. e.g. threshold ms, API version, roles etc. are few to name which differes in schema JSON and the internal C# files. I will wait for these to get addressed, including pending comments.

I have revisited the config defaults and made them consistent across repo.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -296,6 +310,132 @@ public void ValidateFileSinkPath(RuntimeConfig runtimeConfig)
         }
     }
 
+    /// <summary>
+    /// Validates the embeddings configuration options when embeddings are configured.
+    /// Checks required fields, URL format, numeric constraints, and endpoint constraints.
+    /// </summary>
+    public void ValidateEmbeddingsOptions(RuntimeConfig runtimeConfig)
```
souvikghosh04, line 305: do these validations have the test coverage, including positive and negative scenarios?

src/Core/Services/Embeddings/EmbeddingTelemetryHelper.cs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /// <summary>
+    /// Tracks an embedding error.
+    /// </summary>
+    /// <param name="provider">The embedding provider.</param>
+    /// <param name="errorType">The type of error that occurred.</param>
+    public static void TrackError(string provider, string errorType)
+    {
+        _embeddingErrors.Add(1,
```
souvikghosh04, line 122: is this a seed value? can you pass this as a parameter and add proper description in <summary/> docs above?

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,594 @@
...
+        {
+            throw new ArgumentException("Texts array must not contain null or empty entries.", nameof(texts));
+        }
+
+        if (texts.Length > MAX_BATCH_TEXT_COUNT)
+        {
+            throw new ArgumentException(
+                $"Texts array exceeds max supported batch size of {MAX_BATCH_TEXT_COUNT}.",
+                nameof(texts));
+        }
```
souvikghosh04, line 285: can we extract abd move validations/pre-checks separately?

src/Service/HealthCheck/HealthCheckHelper.cs
souvikghosh04: if the outer conditions gets skipped or changed in future, this becomes unpredictable/inconsistent with the assumption that `isDimensionsValid` would always be checked. better keep default as false for safety.

src/Service.Tests/UnitTests/ChunkTextTests.cs
```diff
@@ -0,0 +1,325 @@
...
+using Azure.DataApiBuilder.Service.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Azure.DataApiBuilder.Service.Tests.UnitTests;
+
+/// <summary>
+/// Unit tests for the ChunkText functionality in EmbeddingController.
+/// </summary>
+[TestClass]
+public class ChunkTextTests
```
souvikghosh04, line 16: I think some of the tests are similar and repetative. for e.g., ChunkText_HandlesOverlapEqualToChunkSize, ChunkText_HandlesOverlapLargerThanChunkSize, ChunkText_HandlesEmptyString, ChunkText_HandlesSingleCharacter are almost same. these can be easily collapsed into a single test, attributed by [DataTestMethod] and pass the scenarios specific data as [DataRow]. similarly, other tests should also be refactored.

src/Service.Tests/UnitTests/EmbeddingsChunkingOptionsTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        // Assert
+        Assert.AreEqual(100000, options.SizeChars);
+        Assert.AreEqual(100000, options.EffectiveSizeChars);
+    }
+
+    /// <summary>
+    /// Tests that overlap can be larger than chunk size (edge case).
+    /// </summary>
+    [TestMethod]
+    public void Constructor_AllowsOverlapLargerThanChunkSize()
```
souvikghosh04, line 175: same comment like before. similar tests should be grouped under DataTestMethod

souvikghosh04 (commented):
posting additional comments. tests still need refactoring and not duplicated.

src/Core/Configurations/RuntimeConfigValidator.cs
ajtiwari07: Yes, UTs can be found under ConfigValidationUnitTests.cs.

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Can you elaborate a little more? Do you mean create a separate method just for validations?

src/Service.Tests/UnitTests/EmbeddingsChunkingOptionsTests.cs
ajtiwari07: Combined coherent tests

src/Service.Tests/UnitTests/ChunkTextTests.cs
ajtiwari07: Combined coherent tests.

src/Core/Services/Embeddings/EmbeddingTelemetryHelper.cs
ajtiwari07: updated documentation

src/Core/Services/Embeddings/EmbeddingService.cs
souvikghosh04: yes

souvikghosh04 (approved):
Approved based on the suggestions and discussions in the comments.

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: done

src/Core/Services/Embeddings/EmbeddingService.cs
JimRoberts-MS: The point is that if you had duplicates in the texts, because it's checking all in the cache first and then sending to the api, you'd end up sending duplicates to the api..

Like if texts = ["aaa", "aaa", "bbb"] and you have only {"bbb":...} in the cache, then you'd send "aaa" twice to the api, it's not necessarily likely anyways

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,605 @@
...
+            _logger.LogError(ex, "Failed to generate embedding for text");
+            activity?.SetEmbeddingActivityError(ex);
+            EmbeddingTelemetryHelper.TrackError(_providerName, ex.GetType().Name);
+
+            return new EmbeddingResult(false, null, "Failed to generate embedding.");
+        }
+    }
+
+    /// <inheritdoc/>
+    public async Task<EmbeddingBatchResult> TryEmbedBatchAsync(string[] texts, CancellationToken cancellationToken = default)
```
sayalikudale, line 167: nit: This has the 
 checks duplicated inline which are also present in the method ValidateEmbedBatchRequest. Can we use validation method here instead.

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Modified the logic to pass unique chunk strings embedding creation API call.

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: There is a difference in validations between these methods. I have extracted validation logic methods where ever applicable.

src/Service/HealthCheck/HealthCheckHelper.cs
JimRoberts-MS: not seeing a change? I think TryEmbedAsync (the public method) always tries the cache first, so you wouldn't actually hit the model api for 24 hours

src/Config/ObjectModel/Embeddings/EmbeddingsEndpointOptions.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// In production mode, must be explicitly configured (cannot be null).
+    /// </summary>
+    [JsonPropertyName("roles")]
+    public string[]? Roles { get; init; }
+
+    /// <summary>
+    /// Gets the effective roles.
+    /// Returns configured roles if specified, otherwise defaults to ["authenticated"].
+    /// </summary>
+    /// <param name="isDevelopmentMode">Whether the host is in development mode (kept for API compatibility).</param>
```
JimRoberts-MS: a little confused by this since it seems like we're not filling some interface, is development mode doesn't actually seem to be allowing anonymous access here?

src/Service/HealthCheck/HealthCheckHelper.cs
ajtiwari07: Underlying API issue would surface as per the TTL configured. We share a common error that gets bubbled up in case there's an underlying issue generating embeddings. From what i understand, you're proposing not to render a generic error. I am inclined to keep the implementation as is for now. We can sync, If you strongly think otherwise.

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,432 @@
...
+            ];
+            IActionResult result = await ProcessDocumentArrayAsync(documents, embeddingsOptions, effectiveChunking, cancellationToken);
+
+            // Apply text/plain format when requested, consistent with the non-chunked path.
+            // Each chunk's embedding is output as one line of comma-separated floats.
+            if (ClientAcceptsTextPlain() && result is OkObjectResult okResult && okResult.Value is EmbedDocumentResponse[] docResponses)
+            {
+                IEnumerable<string> lines = docResponses
+                    .SelectMany(d => d.Data)
+                    .Select(embedding => string.Join(",", embedding.Select(f => f.ToString("G", CultureInfo.InvariantCulture))));
```
JimRoberts-MS: G or G9?

JimRoberts-MS: perhaps there could be a serialization helper used across phases that standardizes it

ajtiwari07: Good catch, G9 seems to be the recommended format. Modified

src/Config/ObjectModel/Embeddings/EmbeddingsEndpointOptions.cs
ajtiwari07: Modified effective roles to use anonymous as a default dev mode access

schemas/dab.draft.schema.json
```diff
@@ -770,6 +770,195 @@
...
+                "enabled": {
+                  "type": "boolean",
+                  "description": "Whether caching is enabled for embeddings. Defaults to true.",
+                  "default": true
+                },
+                "level": {
+                  "type": "string",
+                  "description": "Cache level (L1 for in-memory only, L1L2 for in-memory + distributed). Defaults to L1.",
+                  "enum": ["L1", "L1L2"],
+                  "default": "L1"
```
Aniruddh25: Shouldnt the default cache be `L1L2` ? since we would like the embeddings to be stored in Redis Cache? and Redis is L2.

schemas/dab.draft.schema.json
```diff
@@ -770,6 +770,195 @@
...
+              "properties": {
+                "enabled": {
+                  "type": "boolean",
+                  "description": "Whether chunking is enabled. Defaults to true.",
+                  "default": true
+                },
+                "size-chars": {
+                  "type": "integer",
+                  "description": "The size of each chunk in characters.",
+                  "default": 800,
```
Aniruddh25, line 886: PR description says default is 1000. What is the correct default?

src/Cli/Commands/ConfigureOptions.cs
```diff
@@ -176,6 +191,23 @@ public ConfigureOptions(
...
+            RuntimeEmbeddingsProvider = runtimeEmbeddingsProvider;
+            RuntimeEmbeddingsBaseUrl = runtimeEmbeddingsBaseUrl;
+            RuntimeEmbeddingsApiKey = runtimeEmbeddingsApiKey;
+            RuntimeEmbeddingsModel = runtimeEmbeddingsModel;
+            RuntimeEmbeddingsApiVersion = runtimeEmbeddingsApiVersion;
+            RuntimeEmbeddingsDimensions = runtimeEmbeddingsDimensions;
+            RuntimeEmbeddingsTimeoutMs = runtimeEmbeddingsTimeoutMs;
+            // Embeddings Endpoint
+            RuntimeEmbeddingsEndpointEnabled = runtimeEmbeddingsEndpointEnabled;
+            RuntimeEmbeddingsEndpointRoles = runtimeEmbeddingsEndpointRoles;
```
Aniruddh25, line 209: IS EmbeddingsEndpointPath not configurable through dab configure? Why not?

Is EmbeddingsCache not configurable? same question for chunking options.

src/Cli/ConfigGenerator.cs
```diff
@@ -1850,6 +1879,163 @@ private static bool TryUpdateConfiguredFileOptions(
...
+                    Provider: (EmbeddingProviderType)provider,
+                    BaseUrl: baseUrl,
+                    ApiKey: apiKey,
+                    Enabled: enabled,
+                    Model: model,
+                    ApiVersion: apiVersion,
+                    Dimensions: dimensions,
+                    TimeoutMs: timeoutMs,
+                    Endpoint: endpointOptions,
+                    Health: healthOptions);
```
Aniruddh25: doesnt mention cache options, but the schema has cache options. I think we dont need cache options at the embeddings level. The global runtime cache options should be sufficient.

src/Config/Converters/EmbeddingsOptionsConverterFactory.cs
```diff
@@ -0,0 +1,361 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text.Json;
```
Aniruddh25, line 4: cache options are not deserialized either. Better to simply remove them from the dab.draft.schema

src/Service/Startup.cs
```diff
@@ -435,6 +438,50 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<IAuthorizationResolver, AuthorizationResolver>();
             services.AddSingleton<IOpenApiDocumentor, OpenApiDocumentor>();
 
+            // Register embedding service if configured and enabled.
+            // NOTE: IEmbeddingService is only registered when enabled to avoid constructor
+            // failures when config has empty/placeholder values for disabled embeddings.
+            // TODO: To support hot-reload for embeddings (toggling enabled on/off at runtime),
```
Aniruddh25, line 443: Is a separate issue created for to track this?

src/Service/Controllers/EmbeddingController.cs
```diff
@@ -0,0 +1,432 @@
...
+            return NotFound();
+        }
+
+        if (endpointOptions is null || !endpointOptions.Enabled)
+        {
+            return NotFound();
+        }
+
+        // Check if embedding service is available
+        if (_embeddingService is null || !_embeddingService.IsEnabled)
```
Aniruddh25: Whether embeddingService is enabled or not depends on if embeddingOptions.Enabled is set or not, so why check it here again?

src/Core/Services/Embeddings/EmbeddingService.cs
```diff
@@ -0,0 +1,686 @@
...
+/// Supports both OpenAI and Azure OpenAI providers.
+/// Caches embeddings using FusionCache L1 memory cache.
+/// L2/distributed cache is optional globally and is used by this service when configured.
+/// </summary>
+public class EmbeddingService : IEmbeddingService
+{
+    private readonly HttpClient _httpClient;
+    private readonly EmbeddingsOptions _options;
+    private readonly ILogger<EmbeddingService> _logger;
+    private readonly IFusionCache _cache;
```
Aniruddh25, line 27: Do we plan to use Redis Cache in a future PR?

Aniruddh25 (changes requested):
Main question around need for embedding cache options. Cant we reuse runtime cache options to decide what we do for embeddings, does it need to be granular?

schemas/dab.draft.schema.json
ajtiwari07: In this phase we only support L1 via Fusion cache. In phase 2 we will enable support for L2 with redis which would default to L1 or L1L2 like you have suggested.

schemas/dab.draft.schema.json
ajtiwari07: Modified the PR description, you can refer to PRD as a SOT: https://github.com/Azure/data-api-builder/issues/3331

src/Core/Services/Embeddings/EmbeddingService.cs
ajtiwari07: Yes, exactly that is part of phase 2 in the works.

src/Cli/Commands/ConfigureOptions.cs
ajtiwari07: Embedding Cache will get implemented in phase 2, removed it from dab schema

src/Cli/ConfigGenerator.cs
ajtiwari07: Remove embedding cache options

src/Service/Startup.cs
ajtiwari07: Yes, we are tracking it internally. Not a separate git issue.

src/Config/Converters/EmbeddingsOptionsConverterFactory.cs
ajtiwari07: Done

src/Service/Controllers/EmbeddingController.cs
ajtiwari07: Removed it

Aniruddh25 (approved):

src/Service/HealthCheck/HealthCheckHelper.cs
JimRoberts-MS: Certainly don't think it's a blocker. Just that it might be worth it to have 2-3 checks here instead of just 1, depending on how actionable we expect these health checks to be for end users..

Today I see 1 that every TTL (lets say 24 hours) will call the underlying API and hydrate the cache, then for the next 24 hours will never test the health of the API, rather it will test the health of the cache (possibly even just L1 cache)

But in the future would it be better to split it out

1. Check the health of the underlying API
2. Check the health of cache (maybe just L1 even)
3. (Check the health of L2 cache?)

As for any one of those failing you might be able to fix it faster while the other 2 were letting you have a degraded experience without fully taking down the embedding service

JimRoberts-MS (approved):

src/Service/HealthCheck/HealthCheckHelper.cs
ajtiwari07: Point taken, I will create a work item for it.

PR #3480, opened 2026-04-29
Upgrade Hot Chocolate to version 16.0.0

### LINKED ISSUE #3448: Upgrade DAB to latest HC -rc for version 16.0.0
We need to upgrade Hot Chocolate to latest -rc for version 16.0.0

aaronburtle:
## Why make this change?

Closes #3448

Hot Chocolate has released stable v16.0.0 and we upgrade DAB to that version.

### Update deps

Bump 6 HC packages to 16.0.0 in Directory.Packages.props:
HotChocolate, HotChocolate.AspNetCore, HotChocolate.AspNetCore.Authorization,
HotChocolate.Types.NodaTime, HotChocolate.Utilities.Introspection,
HotChocolate.Diagnostics.

### Breaking changes (GraphQL schema)

HC v16 forces / motivates two schema-type renames. Wire formats and runtime
values are unchanged for both. REST and CLI surface area is unchanged.

  Scalar:        Byte            ->  UnsignedByte
  Filter input:  ByteFilterInput ->  UnsignedByteFilterInput

  HC v16 split the legacy `ByteType` into a signed `ByteType` (sbyte,
  -128..127) and a new `UnsignedByteType` (byte, 0..255). SQL Server `tinyint`
  is unsigned, so DAB must bind to `UnsignedByteType`.

  Scalar:        ByteArray            ->  Base64String
  Filter input:  ByteArrayFilterInput ->  Base64StringFilterInput

  HC v16 marked `ByteArrayType` as [Obsolete] in favor of `Base64StringType`.
  Both serialize byte[] as a base64-encoded JSON string (identical wire
  format), but the GraphQL scalar name is now `Base64String`. DAB targets the
  new name so the generated schema does not depend on a deprecated scalar.

Inline literal filters continue to work for both renames. Clients that
declare typed variables of `ByteFilterInput` or `ByteArrayFilterInput`, or
that regenerate code from the schema (graphql-codegen, Strawberry Shake,
etc.), must update the type names.

### Refactor scalar APIs

v16 renamed scalar overrides (`ParseLiteral`/`ParseValue` -> 
`OnCoerceInputLiteral`/`OnCoerceInputValue`/`OnCoerceOutputValue`/
`OnValueToLiteral`) and call-site helpers (`ParseValue`/`ParseResult` -> 
`ValueToLiteral`). `DateTimeType.ValueToLiteral` only accepts `DateTimeOffset`,
so we convert at the boundary. HC v16 also elides trailing zero fractional
seconds on DateTime output (`...:54.000Z` -> `...:54Z`), which surfaced in
several test assertion updates.

### Refactor resolver/execution APIs

`ISelection` was removed (use concrete `Selection`); `Selection.SyntaxNode`
became a `ReadOnlySpan<FieldSelectionNode>` (`SyntaxNodes`) to support
field-merging; `TimeSpanType` was replaced by `DurationType` (ISO-8601,
parsed via `XmlConvert.ToTimeSpan`); `OperationResult.WithContextData` is
gone (set `singleResult.ContextData` directly).

### Refactor request interceptor and status middleware

`IntrospectionInterceptor` now has a no-arg constructor (schema services are
isolated from request services in v16) and implements the new four-arg
`OnCreateAsync(HttpContext, IRequestExecutor, OperationRequestBuilder, CancellationToken)`
signature; `RuntimeConfigProvider` is resolved from
`context.RequestServices` at request time.

`DetermineStatusCodeMiddleware` was updated for the unified `OperationResult`
type. Status codes are now propagated via context-data
(`ExecutionContextData.HttpStatusCode`) using an `ImmutableDictionary`
builder rather than mutating the response directly.

### Adopt lazy schema initialization (LazyInitialization = true)

HC v16 builds the schema eagerly during host startup by default. DAB
supports a "hosted" scenario where the runtime config is supplied after the
host starts (POST `/configuration`), so eager construction races with our
placeholder fallback. We opt back into the v15 default
(`options.LazyInitialization = true`) so the schema is built on the first
GraphQL request, by which time the runtime config and metadata provider are
ready.

### Drop ModifyOptions(o => o.EnableOneOf = true)

OneOf is on by default in v16.

### Refactor startup

`DateTimeType(disableFormatCheck: bool)` is obsolete (replaced by
`DateTimeOptions { ValidateInputFormat = ... }`, polarity flipped);
`WithOptions` only accepts `Action<GraphQLServerOptions>` (per-request);
`MapNitroApp().WithOptions(GraphQLToolOptions)` was removed. Nitro is now
served by the same `/graphql` endpoint and is governed by
`options.Tool.Enable` on the per-request server options.

### Harden Selection.SyntaxNodes access

HC v16's `Selection.SyntaxNodes` is a `ReadOnlySpan<FieldSelectionNode>`,
so direct `SyntaxNodes[0].Node` indexing would surface as
`IndexOutOfRangeException` at request time if the span were ever empty.
Added `Selection.RequireFieldNode()` in `SelectionExtensions.cs`
(`IsEmpty` check, return `SyntaxNodes[0].Node`, throw targeted
`DataApiBuilderException` otherwise). Adopted at all call sites in
`SqlQueryStructure`, `CosmosQueryStructure`, and `ExecutionHelper`.

### Test updates

Adopt v16 `OperationResultData` / `ResultDocument` / `ResultElement` /
`ResultProperty` traversal in `MultiSourceQueryExecutionUnitTests`. Update
`TestNoConfigReturnsServiceUnavailable` to recognize that lazy
`WithOptions` resolution surfaces "no runtime config" as
`DataApiBuilderException(ServiceUnavailable)` rather than a 503 response or
`ApplicationException`. Update DateTime literal assertions and filter input
name references in affected test files. Decouple the GraphQL-scalar name
from the test database column name in `GraphQLSupportedTypesTestsBase` and
the My/Pg test classes so the `Byte`/`UnsignedByte` and
`ByteArray`/`Base64String` renames do not require renaming
`byte_types`/`bytearray_types` columns.

## How was this tested?

Run against current test suite.


src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -126,7 +126,7 @@ public SqlQueryStructure(
                 sqlMetadataProvider,
                 authorizationResolver,
                 ctx.Selection.Field,
-                ctx.Selection.SyntaxNode,
+                ctx.Selection.SyntaxNodes[0].Node,
                 // The outermost query is where we start, so this can define
```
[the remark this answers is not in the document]
aaronburtle: valid, addressing.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -173,7 +173,7 @@ public SqlQueryStructure(
             IsMultipleCreateOperation = isMultipleCreateOperation;
 
             ObjectField schemaField = _ctx.Selection.Field;
-            FieldNode? queryField = _ctx.Selection.SyntaxNode;
+            FieldNode? queryField = _ctx.Selection.SyntaxNodes[0].Node;
```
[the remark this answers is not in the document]
aaronburtle: valid, addressing.

src/Core/Resolvers/CosmosQueryStructure.cs
```diff
@@ -117,14 +117,14 @@ private static IEnumerable<LabelledColumn> GenerateQueryColumns(SelectionSetNode
...
-            ISelection selection = _context.Selection;
+            Selection selection = _context.Selection;
             ObjectType underlyingType = selection.Field.Type.NamedType<ObjectType>();
 
             IsPaginated = QueryBuilder.IsPaginationType(underlyingType);
             OrderByColumns = new();
             if (IsPaginated)
             {
-                FieldNode? fieldNode = ExtractQueryField(selection.SyntaxNode);
+                FieldNode? fieldNode = ExtractQueryField(selection.SyntaxNodes[0].Node);
```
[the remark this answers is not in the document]
aaronburtle: valid, addressing

src/Core/Resolvers/CosmosQueryStructure.cs
```diff
@@ -139,7 +139,7 @@ private void Init(IDictionary<string, object?> queryParams)
             }
             else
             {
-                Columns.AddRange(GenerateQueryColumns(selection.SyntaxNode.SelectionSet!, _context.Operation.Document, SourceAlias));
+                Columns.AddRange(GenerateQueryColumns(selection.SyntaxNodes[0].Node.SelectionSet!, _context.Operation.Document, SourceAlias));
                 string typeName = GraphQLUtils.TryExtractGraphQLFieldModelName(underlyingType.Directives, out string? modelName) ?
```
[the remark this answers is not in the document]
aaronburtle: valid, addressing.

src/Core/Services/ExecutionHelper.cs
```diff
@@ -508,7 +509,7 @@ public static InputObjectType InputObjectTypeFromIInputField(IInputValueDefiniti
         {
             return GetParametersFromSchemaAndQueryFields(
                 context.Selection.Field,
-                context.Selection.SyntaxNode,
+                context.Selection.SyntaxNodes[0].Node,
                 context.Variables);
```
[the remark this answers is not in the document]
aaronburtle: valid, addressing.

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -156,12 +156,12 @@ private static Tuple<string, IValueNode> ConvertValueToGraphQLType(string defaul
...
-                    UUID_TYPE => new(UUID_TYPE, new UuidType().ParseValue(Guid.Parse(defaultValueFromConfig))),
+                    UUID_TYPE => new(UUID_TYPE, new UuidType().ValueToLiteral(Guid.Parse(defaultValueFromConfig))),
                     BYTE_TYPE => new(BYTE_TYPE, new IntValueNode(byte.Parse(defaultValueFromConfig))),
                     SHORT_TYPE => new(SHORT_TYPE, new IntValueNode(short.Parse(defaultValueFromConfig))),
                     INT_TYPE => new(INT_TYPE, new IntValueNode(int.Parse(defaultValueFromConfig))),
                     LONG_TYPE => new(LONG_TYPE, new IntValueNode(long.Parse(defaultValueFromConfig))),
-                    SINGLE_TYPE => new(SINGLE_TYPE, new SingleType().ParseValue(float.Parse(defaultValueFromConfig))),
+                    SINGLE_TYPE => new(SINGLE_TYPE, new SingleType().ValueToLiteral(float.Parse(defaultValueFromConfig))),
                     FLOAT_TYPE => new(FLOAT_TYPE, new FloatValueNode(double.Parse(defaultValueFromConfig))),
                     DECIMAL_TYPE => new(DECIMAL_TYPE, new FloatValueNode(decimal.Parse(defaultValueFromConfig))),
```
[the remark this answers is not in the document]
aaronburtle: pre-existing issue, but will fix.

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#nullable enable annotations
```
souvikghosh04: nit: this can be removed in checkin and before merge

src/Service.Tests/UnitTests/RequestParserUnitTests.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#nullable enable annotations
```
souvikghosh04: same here- should be removed before merge

src/Service.Tests/UnitTests/SqlQueryExecutorUnitTests.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#nullable enable annotations
```
souvikghosh04: nit: remove this

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
```diff
@@ -11,7 +11,13 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder.GraphQLTypes
...
     {
         public const string UUID_TYPE = "UUID";
-        public const string BYTE_TYPE = "Byte";
+        // HC v16 split the legacy Byte scalar into:
+        //  - ByteType         (runtime: sbyte, range -128..127)
+        //  - UnsignedByteType (runtime: byte,  range 0..255)
+        // SQL Server's tinyint maps to .NET byte (0..255), so DAB targets UnsignedByte.
+        // The GraphQL type name visible in the generated schema therefore changed from
+        // "Byte" to "UnsignedByte" with the HC v16 upgrade.
+        public const string BYTE_TYPE = "UnsignedByte";
```
souvikghosh04, line 20: I don't see this in PR description. also, the constant `BYTE_TYPE` in `SupportedTypes.cs` changes from `Byte` to `UnsignedByte`, and `StandardQueryInputs.cs` changes the filter type from `ByteFilterInput` to `UnsignedByteFilterInput`. These are user-visible GraphQL schema changes. Any existing client queries using ByteFilterInput or referencing the Byte scalar will break. The PR description says "No user-facing API changes", this contradicts the actual diff. This should be explicitly called out in the PR description as a breaking change or documented in migration guidance.

src/Core/Services/ExecutionHelper.cs
```diff
@@ -211,11 +211,21 @@ fieldValue.ValueKind is not (JsonValueKind.Undefined or JsonValueKind.Null))
...
-                        ByteArrayType => fieldValue.GetBytesFromBase64(),
+                        // HC v16 ships both ByteArrayType (legacy, GraphQL name "ByteArray", runtime byte[])
+                        // and Base64StringType (new, GraphQL name "Base64String", runtime byte[]). DAB's
+                        // generated schemas still use the GraphQL name "ByteArray", so HC binds entity
+                        // fields to ByteArrayType. We accept either type here so DAB also tolerates
+                        // schemas that bind to Base64StringType (e.g. via DefaultValueType).
+                        // CS0618: ByteArrayType is [Obsolete] in HC v16 in favor of Base64StringType,
+                        // but we still need to pattern-match it because it remains the type bound to
+                        // the GraphQL name "ByteArray" that DAB-generated schemas continue to use.
+#pragma warning disable CS0618
```
souvikghosh04: we should remove the pragma warning disables

src/Core/Services/ExecutionHelper.cs
```diff
@@ -211,11 +211,21 @@ fieldValue.ValueKind is not (JsonValueKind.Undefined or JsonValueKind.Null))
...
+                        Base64StringType or ByteArrayType => fieldValue.GetBytesFromBase64(),
+#pragma warning restore CS0618
                         BooleanType => fieldValue.GetBoolean(), // spec
                         UrlType => new Uri(fieldValue.GetString()!),
                         UuidType => fieldValue.GetGuid(),
-                        TimeSpanType => TimeSpan.Parse(fieldValue.GetString()!),
+                        DurationType => XmlConvert.ToTimeSpan(fieldValue.GetString()!),
```
souvikghosh04, line 220: existing TimeSpanType is now replaced with DurationType and TimeSpan.Parse() is replaced with XmlConvert.ToTimeSpan(). both are different. and if any existing database returns TimeSpan values in the old format, this will throw FormatException at runtime. Verify that all database providers return durations in consistent format, or add a fallback. also, add test coverages around this for different databases.

src/Service/Startup.cs
```diff
@@ -661,16 +661,66 @@ private void ConfigureResponseCompression(IServiceCollection services, RuntimeCo
...
+                // Hot Chocolate v16 builds the GraphQL schema eagerly during host startup. DAB
+                // supports a "hosted" scenario where the runtime config is supplied after the
+                // host starts (POST /configuration), so the schema cannot exist at startup. Eager
+                // initialization plus our placeholder-schema fallback also creates a race on
+                // late-config hydration: HC keeps serving the warm placeholder executor in the
+                // background while the new schema's warmup runs, so the first GraphQL request
+                // after hydration validates against the placeholder and returns BadRequest.
+                // Opt into lazy initialization (the v15 default) so the schema is constructed on
+                // first request, by which time the runtime config is loaded and the metadata
+                // provider has been initialized in PerformOnConfigChangeAsync.
```
souvikghosh04: nit: simplify this comment. this is almost a documentation. if needed updated docs

src/Service/Startup.cs
```diff
@@ -661,16 +661,66 @@ private void ConfigureResponseCompression(IServiceCollection services, RuntimeCo
...
+                .ModifyOptions(options => options.LazyInitialization = true)
                 .AddInstrumentation()
-                .AddType(new DateTimeType(disableFormatCheck: graphQLRuntimeOptions?.EnableLegacyDateTimeScalar ?? true))
+                .AddType(new DateTimeType(new DateTimeOptions { ValidateInputFormat = !(graphQLRuntimeOptions?.EnableLegacyDateTimeScalar ?? true) }))
                 .AddHttpRequestInterceptor<DefaultHttpRequestInterceptor>()
                 .ConfigureSchema((serviceProvider, schemaBuilder) =>
                 {
```
souvikghosh04, line 672: nit: the comments can be simplified?

src/Service/Startup.cs
```diff
@@ -919,23 +969,11 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, RuntimeC
 
                 endpoints
                     .MapGraphQL()
-                    .WithOptions(new GraphQLServerOptions
+                    .WithOptions(options =>
```
souvikghosh04, line 967: I see the dev mode nitro endpoint was disabled earlier using `endpoints.MapNitroApp().WithOptions()` which now seems to be removed. could you please check and elaborate?

souvikghosh04 (commented):
Posting comments so far..

src/Service.GraphQLBuilder/GraphQLTypes/SupportedTypes.cs
aaronburtle: Good call out, added to the PR description.

src/Service.Tests/Configuration/RuntimeConfigLoaderTests.cs
aaronburtle: sure, I can move this out of this PR, but there's really no reason for us to pollute the build logs with these warnings, so ive opened an issue here to track:

https://github.com/Azure/data-api-builder/issues/3543

src/Core/Services/ExecutionHelper.cs
aaronburtle: This CS0618 warning the pragma suppresses is introduced by this PR. On main, the same line is just `ByteArrayType => fieldValue.GetBytesFromBase64()` with no warning, because HC v13/v14 did not mark `ByteArrayType` as [Obsolete]. The v16 bump in this PR is what flags it obsolete, so removing the suppression would leave a new warning ringing every build that we ourselves introduced.

src/Core/Services/ExecutionHelper.cs
aaronburtle: Reasonable concern, but let me add some context. No DAB-supported column type binds to `Duration` today. SQL `time` round-trips through `TimeOnly` -> NodaTime's `LocalTimeType`, wired up in Startup.cs via BindRuntimeType<TimeOnly, LocalTimeType>() and BindScalarType<LocalTimeType>("LocalTime"). The `DurationType` arm in ExecutionHelper is a defensive fallback only, it doesn't fire for any provider DAB supports. HC v16's `DurationType` serializes `TimeSpan` to ISO-8601 as its wire format, so `XmlConvert.ToTimeSpan` is the symmetric inverse of HC's own  serializer. The two parsers only diverge if some other code path forced a `time`-formatted string into a Duration-bound field, which doesn't exist. A unit test of this arm would effectively be testing `XmlConvert.ToTimeSpan`, i.e. the .NET BCL, through a non-trivial `IResolverContext` mock. It doesn't add real surface area, and adding integration tests across SQL/PostgreSQL/MySQL providers presumes a Duration binding we don't have.

Happy to revisit if we expose a Duration-bound type, at which point both the binding and the round-trip test should land together.

src/Service/Startup.cs
aaronburtle: added a doc and simplified this.

src/Service/Startup.cs
aaronburtle: simplified

src/Service/Startup.cs
aaronburtle: Behavior is preserved, HC v16 removed the standalone `MapNitroApp()` endpoint, and the Nitro IDE is now served by `/graphql` itself, gated by `options.Tool.Enable` on the per-request `GraphQLServerOptions`. That's the new `.WithOptions(options => { options.Tool.Enable = IsUIEnabled(runtimeConfig, env); })` block at Startup.cs lines 899-905, replacing the old `.MapNitroApp().WithOptions(...)` pair. `IsUIEnabled` is the same gate as before (Development host mode + GraphQL enabled), so dev-mode UI works and prod stays off. There's a section in docs/design/HC16-upgrade.md ("WithOptions takes a delegate, and MapNitroApp is gone") with the before/after side by side.

src/Core/Services/ExecutionHelper.cs
aaronburtle: I actually decided to shift to just supporting the new data type so these suppressions are no longer needed. ByteArray is obsolete and 2.0 is the best time for us to transition to Base64String.

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -124,12 +133,70 @@ private ISchemaBuilder Parse(
...
+                        directives: new List<DirectiveNode>());
+
+                    rewritten.Add(new ObjectTypeDefinitionNode(
+                        objectType.Location,
+                        objectType.Name,
+                        objectType.Description,
+                        objectType.Directives,
+                        objectType.Interfaces,
+                        new List<FieldDefinitionNode> { placeholderField }));
+                    placeholderInjected = true;
```
Aniruddh25: can we simply break here if placeholder is Injected?

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -124,12 +133,70 @@ private ISchemaBuilder Parse(
...
+        /// any configuration that produces real query fields, so it is only visible in
+        /// otherwise-empty schemas (GraphQL globally disabled, all entities opting out,
+        /// no entities configured).
+        /// </summary>
+        private static DocumentNode EnsureQueryHasAtLeastOneField(DocumentNode queryNode, ISchemaBuilder sb)
+        {
+            ImmutableArray<IDefinitionNode>.Builder rewritten = ImmutableArray.CreateBuilder<IDefinitionNode>(queryNode.Definitions.Count);
+            bool placeholderInjected = false;
+
+            foreach (IDefinitionNode definition in queryNode.Definitions)
```
Aniruddh25: Instead of going through all definitions, and querying for the objectTypeName = Query, cant we just look it up using the name? or maybe get it from the root definition node? O(n) v/s O(1) operation.

Aniruddh25 (changes requested):
Consider optimizing Placeholder query node creation

src/Core/Services/GraphQLSchemaCreator.cs
aaronburtle: We can't simply break the way that this is structured because right now we are doing one pass that both scans for `Query` and also does the rebuild. If we simply break here we would miss rebuilding later `Query` definitions that are non empty (the else branch below).

But what we can do is change this into 2 separate phases. One will scan for empty query definitions, and then if found we build a placeholder. If none are found we just keep what we have, nothing to rebuild.

src/Core/Services/GraphQLSchemaCreator.cs
aaronburtle: Unfortunately, we cannot. HC's Definitions is just an ordered list of DefinitionNodes, it doesnt have a by-name index, and the DocumentNode is the root, so there's no separate root definition node to query. Name based lookups exist in the built schema, but that is too late for us since we're rewriting the AST before the schema is built.

However, as mentioned in the other comment, we can optimize this with a 2 phase approach, where if in the first phase we dont find any empty query nodes, we can skip paying the cost to rebuild.

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -152,49 +152,57 @@ private ISchemaBuilder Parse(
...
-                        objectType.Location,
-                        objectType.Name,
-                        objectType.Description,
-                        objectType.Directives,
-                        objectType.Interfaces,
-                        new List<FieldDefinitionNode> { placeholderField }));
-                    placeholderInjected = true;
-                }
-                else
+                if (queryNode.Definitions[i] is ObjectTypeDefinitionNode { Name.Value: "Query", Fields.Count: 0 })
```
Aniruddh25, line 167: Will this perform a reference equality check or simply equivalence check? I hope the latter, since addresses cannot be equal

src/Core/Services/GraphQLSchemaCreator.cs
```diff
@@ -152,49 +152,57 @@ private ISchemaBuilder Parse(
...
-            ImmutableArray<IDefinitionNode>.Builder rewritten = ImmutableArray.CreateBuilder<IDefinitionNode>(queryNode.Definitions.Count);
-            bool placeholderInjected = false;
-
-            foreach (IDefinitionNode definition in queryNode.Definitions)
+            // Locate the empty Query definition, if any. HotChocolate's DocumentNode exposes
+            // Definitions as an ordered IReadOnlyList with no by-name index, so this is the
+            // cheapest available lookup. Common case: Query has fields and we return unchanged
+            // without allocating a new definitions list.
+            int emptyQueryIndex = -1;
+            for (int i = 0; i < queryNode.Definitions.Count; i++)
```
Aniruddh25, line 165: Is there a test case that exercises this code?

Aniruddh25 (approved):
Approving with suggestions. Thanks for the change

souvikghosh04 (approved):
thanks for addressing the comments

src/Core/Services/GraphQLSchemaCreator.cs
aaronburtle: yes, this is equivalence check.

src/Core/Services/GraphQLSchemaCreator.cs
aaronburtle: added some tests to cover both branches and a no query edge case.

Aniruddh25 (approved):

PR #3484, opened 2026-04-29
Add MCP notifications/message for log streaming to clients

anushakolan:
# Add MCP `notifications/message` for log streaming to clients

## Why make this change?

Enables MCP clients (like MCP Inspector, Claude Desktop, VS Code Copilot) to receive real-time log output from DAB through the MCP protocol's `notifications/message` channel — the same way they already receive tool responses. No more digging through stderr or container logs to debug a tool call.

Related: #3274 · Depends on: #3419

## What is this change?

When an MCP client subscribes to logs by sending `logging/setLevel` with a level other than `none`, DAB starts forwarding log entries to the client as JSON-RPC notifications:

```json
{
  "jsonrpc": "2.0",
  "method": "notifications/message",
  "params": {
    "level": "info",
    "logger": "Azure.DataApiBuilder.Service.Startup",
    "data": "Starting Data API builder..."
  }
}
```

### Key behavior

- **Off by default.** In MCP stdio mode no log noise reaches the client until it explicitly subscribes via `logging/setLevel`.
- **Operator precedence respected.** CLI `--LogLevel` and config `runtime.telemetry.log-level` still win over client requests. The client subscription is acknowledged but the level stays at whatever the operator pinned.
- **Channel-safe.** All writes to stdout (responses + notifications) go through a single lock so JSON-RPC frames cannot interleave or corrupt the channel.
- **Lifecycle-safe.** Stdout is owned by a single component, disposed cleanly on process exit, and tolerates late writes during shutdown.

## How was this tested?

- **22 unit tests** covering the notification frame format, the on/off gate, lock-protected stdout under heavy concurrency (16 threads × 500 writes), idempotent disposal, and CLI/Config precedence under load.
- **End-to-end smoke tests** confirming: silent default, verbose stderr, CLI/Config override blocking client `setLevel`, real tool calls with notifications interleaved cleanly, and a 25-request burst including an 8.5 KB frame — 100 % of frames valid JSON, every response accounted for.
- **Manual verification** in MCP Inspector and VS Code MCP client.
  - Verified notifications appear in MCP Inspector when `logging/setLevel` is sent.
  - Verified VS Code MCP client behavior (auto-subscribes after `initialize`).

### DIFF
--- a/src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
@@ -8,6 +8,7 @@ using Azure.DataApiBuilder.Core.AuthenticationHelpers.AuthenticationSimulator;
 using Azure.DataApiBuilder.Core.Configurations;
 using Azure.DataApiBuilder.Core.Telemetry;
 using Azure.DataApiBuilder.Mcp.Model;
+using Azure.DataApiBuilder.Mcp.Telemetry;
 using Azure.DataApiBuilder.Mcp.Utils;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
@@ -296,16 +297,24 @@ namespace Azure.DataApiBuilder.Mcp.Core
 
             // Attempt to update the log level
             // If CLI or Config overrode, this returns false but we still return success to the client
-            bool updated = logLevelController.UpdateFromMcp(level);
+            logLevelController.UpdateFromMcp(level);
 
-            // If MCP successfully changed the log level to something other than "none",
-            // ensure Console.Error is pointing to the real stderr (not TextWriter.Null).
-            // This handles the case where MCP stdio mode started with LogLevel.None (quiet startup)
-            // and the client later enables logging via logging/setLevel.
+            // Determine if logging is enabled (level != "none")
+            // Note: Even if CLI/Config overrode the level, we still enable notifications
+            // when the client requests logging. They'll get logs at the overridden level.
             bool isLoggingEnabled = !string.Equals(level, "none", StringComparison.OrdinalIgnoreCase);
-            if (updated && isLoggingEnabled)
+            if (isLoggingEnabled)
             {
                 RestoreStderrIfNeeded();
+            }
+
+            // Enable or disable MCP log notifications based on the requested level
+            // When CLI/Config overrode, notifications are still enabled - client asked for logs,
+            // they just get them at the CLI/Config level instead of the requested level.
+            IMcpLogNotificationWriter? notificationWriter = _serviceProvider.GetService<IMcpLogNotificationWriter>();
+            if (notificationWriter != null)
+            {
+                notificationWriter.IsEnabled = isLoggingEnabled;
             }
 
             // Always return success (empty result object) per MCP spec
@@ -546,7 +555,7 @@ namespace Azure.DataApiBuilder.Mcp.Core
         {
             var response = new
             {
-                jsonrpc = "2.0",
+                jsonrpc = McpStdioJsonRpcErrorCodes.JSON_RPC_VERSION,
                 id = id.HasValue ? GetIdValue(id.Value) : null,
                 result = resultObject
             };
@@ -565,7 +574,7 @@ namespace Azure.DataApiBuilder.Mcp.Core
         {
             var errorObj = new
             {
-                jsonrpc = "2.0",
+                jsonrpc = McpStdioJsonRpcErrorCodes.JSON_RPC_VERSION,
                 id = id.HasValue ? GetIdValue(id.Value) : null,
                 error = new { code, message }
             };
--- a/src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
+++ b/src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
@@ -27,6 +27,7 @@ namespace Azure.DataApiBuilder.Mcp.Core
     {
         private readonly McpToolRegistry _toolRegistry;
         private readonly IServiceProvider _serviceProvider;
+        private readonly McpStdoutWriter _stdoutWriter;
         private readonly string _protocolVersion;
 
         private const int MAX_LINE_LENGTH = 1024 * 1024; // 1 MB limit for incoming JSON-RPC requests
@@ -35,6 +36,11 @@ namespace Azure.DataApiBuilder.Mcp.Core
         {
             _toolRegistry = toolRegistry ?? throw new ArgumentNullException(nameof(toolRegistry));
             _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+
+            // Resolve the shared stdout writer so JSON-RPC responses and
+            // notifications/message frames are serialized through one lock.
+            // Falls back to a fresh instance if DI didn't register one (defensive).
+            _stdoutWriter = _serviceProvider.GetService<McpStdoutWriter>() ?? new McpStdoutWriter();
 
             // Allow protocol version to be configured via IConfiguration, using centralized defaults.
             IConfiguration? configuration = _serviceProvider.GetService<IConfiguration>();
@@ -48,16 +54,14 @@ namespace Azure.DataApiBuilder.Mcp.Core
         /// <returns>A task representing the asynchronous operation.</returns>
         public async Task RunAsync(CancellationToken cancellationToken)
         {
-            // Use UTF-8 WITHOUT BOM
+            // Use UTF-8 WITHOUT BOM for stdin. Stdout is owned by McpStdoutWriter,
+            // which serializes all writes from McpStdioServer and the MCP logging
+            // pipeline so JSON-RPC frames cannot interleave at the byte level.
             UTF8Encoding utf8NoBom = new(encoderShouldEmitUTF8Identifier: false);
 
             using Stream stdin = Console.OpenStandardInput();
-            using Stream stdout = Console.OpenStandardOutput();
             using StreamReader reader = new(stdin, utf8NoBom);
-            using StreamWriter writer = new(stdout, utf8NoBom) { AutoFlush = true };
 
-            // Redirect Console.Out to use our writer
-            Console.SetOut(writer);
             while (!cancellationToken.IsCancellationRequested)
             {
                 string? line = await reader.ReadLineAsync(cancellationToken);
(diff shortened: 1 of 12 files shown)


src/Service/Program.cs
```diff
@@ -158,6 +171,12 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
 
                     // Add filter for dynamic log level changes (e.g., via MCP logging/setLevel)
                     logging.AddFilter(logLevel => LogLevelProvider.ShouldLog(logLevel));
+
+                    // For MCP stdio mode, add the MCP logger provider to send logs as notifications
+                    if (runMcpStdio)
+                    {
+                        logging.AddProvider(new McpLoggerProvider(McpNotificationWriter, LogLevelProvider.ShouldLog));
```
Aniruddh25: how do you ensure other LogLevelProviders dont get added when in stdio mode?

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLogNotificationWriter.cs
```diff
@@ -0,0 +1,140 @@
...
+        if (!_isEnabled || _writer == null)
+        {
+            return;
+        }
+
+        string mcpLevel = ConvertToMcpLevel(logLevel);
+
+        var notification = new
+        {
+            jsonrpc = "2.0",
```
Aniruddh25: define a constant that represents `2.0` protocol, so it can be changed in one shot.

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLogNotificationWriter.cs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// <summary>
+    /// Converts .NET LogLevel to MCP log level string.
+    /// </summary>
+    /// <remarks>
+    /// MCP levels: debug, info, notice, warning, error, critical, alert, emergency
+    /// </remarks>
+    private static string ConvertToMcpLevel(LogLevel logLevel)
+    {
+        return logLevel switch
```
Aniruddh25: Doesnt McpStdioServer or somewhere else we already define such conversion levels? Can we reuse that?

src/Service.Tests/UnitTests/McpLogNotificationTests.cs
```diff
@@ -0,0 +1,128 @@
...
+            ILogger logger2 = provider.CreateLogger("TestCategory");
+            ILogger logger3 = provider.CreateLogger("OtherCategory");
+
+            // Assert - same category should return same logger instance
+            Assert.AreSame(logger1, logger2);
+            Assert.AreNotSame(logger1, logger3);
+        }
+
+        [TestMethod]
+        public void McpLogger_BeginScope_ReturnsNonNullDisposable()
```
Aniruddh25: this is a trivial test can be skipped. since we dont even use the scope.

Aniruddh25 (dismissed):
Left some suggestions for code reuse.

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLogNotificationWriter.cs
anushakolan: Created a centralized `McpLogLevelConverter` class in Telemetry that provides:

`TryConvertFromMcp(string mcpLevel, out LogLevel logLevel)` - MCP string → .NET LogLevel (used by `DynamicLogLevelProvider`)
`ConvertToMcp(LogLevel logLevel)` - .NET LogLevel → MCP string (used by `McpLogNotificationWriter`)
This consolidates the bidirectional mapping that was previously duplicated:

`DynamicLogLevelProvider` had `_mcpLevelMapping` dictionary (MCP → LogLevel)
`McpLogNotificationWriter` had `ConvertToMcpLevel()` method (LogLevel → MCP)
Now both reference the single `McpLogLevelConverter` class.

src/Service.Tests/UnitTests/McpLogNotificationTests.cs
anushakolan: Agreed, removed `McpLogger_BeginScope_ReturnsNonNullDisposable` test.

src/Service/Program.cs
anushakolan: Good catch! Added `logging.ClearProviders()` before adding `McpLoggerProvider` in `stdio` mode. This removes all default providers (Console, Debug, EventSource, EventLog) so` McpLoggerProvider` is the exclusive provider. Without this, the default `ConsoleLoggerProvider` could write to `stdout` and corrupt the JSON-RPC stream.

Aniruddh25 (approved):

Aniruddh25 (approved):

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLogNotificationWriter.cs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// <summary>
+    /// Initializes the stdout writer for MCP notifications.
+    /// Uses Console.OpenStandardOutput() to get the raw stdout stream,
+    /// bypassing any Console.SetOut() redirections.
+    /// </summary>
+    private void InitializeWriter()
+    {
+        // Use the same approach as McpStdioServer - get raw stdout
+        Stream stdout = Console.OpenStandardOutput();
```
aaronburtle: This StreamWriter is on the same stdout as we use in `McpStdioServer.cs` in `RunAsync`

`using Stream stdout = Console.OpenStandardOutput();`

I think this will have concurrency/race conditions. Maybe we can factor these two uses into a helper with a lock?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -296,18 +297,26 @@ private void HandleSetLogLevel(JsonElement? id, JsonElement root)
...
-            // If MCP successfully changed the log level to something other than "none",
-            // ensure Console.Error is pointing to the real stderr (not TextWriter.Null).
-            // This handles the case where MCP stdio mode started with LogLevel.None (quiet startup)
-            // and the client later enables logging via logging/setLevel.
+            // Determine if logging is enabled (level != "none")
+            // Note: Even if CLI/Config overrode the level, we still enable notifications
+            // when the client requests logging. They'll get logs at the overridden level.
             bool isLoggingEnabled = !string.Equals(level, "none", StringComparison.OrdinalIgnoreCase);
-            if (updated && isLoggingEnabled)
+            if (isLoggingEnabled)
```
aaronburtle: without the guard from `updated`, won't this result in us resetting stderr in the case where it was initially set to `none` re-introducing noisy logs when someone set log level to none on startup?

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -296,18 +297,26 @@ private void HandleSetLogLevel(JsonElement? id, JsonElement root)
...
                 RestoreStderrIfNeeded();
             }
 
+            // Enable or disable MCP log notifications based on the requested level
+            // When CLI/Config overrode, notifications are still enabled - client asked for logs,
+            // they just get them at the CLI/Config level instead of the requested level.
+            IMcpLogNotificationWriter? notificationWriter = _serviceProvider.GetService<IMcpLogNotificationWriter>();
+            if (notificationWriter != null)
+            {
+                notificationWriter.IsEnabled = isLoggingEnabled;
```
aaronburtle, line 328: This makes the `IsEnabled` flag in `McpLogger.cs` redudant, and gives us 2 gates. Should refactor this so that we have 1 source of truth. I dont think the 2 gates always agree currently.

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLogNotificationWriter.cs
```diff
@@ -0,0 +1,122 @@
...
+        _writer = new StreamWriter(stdout, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false))
```
aaronburtle: We probably need to implement IDisposable and properly dispose of the StreamWriter.

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLogger.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        if (string.IsNullOrEmpty(message) && exception == null)
+        {
+            return;
+        }
+
+        // Include exception details if present
+        if (exception != null)
+        {
+            message = $"{message} Exception: {exception.GetType().Name}: {exception.Message}";
```
aaronburtle: Do we intentionally drop the stack trace, inner exceptions, etc here? Do MCP clients expect to see this?

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLogger.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    public McpLogger(string categoryName, IMcpLogNotificationWriter writer, Func<LogLevel, bool> levelFilter)
+    {
+        _categoryName = categoryName ?? throw new ArgumentNullException(nameof(categoryName));
+        _writer = writer ?? throw new ArgumentNullException(nameof(writer));
+        _levelFilter = levelFilter ?? throw new ArgumentNullException(nameof(levelFilter));
+    }
+
+    /// <inheritdoc />
+    public IDisposable? BeginScope<TState>(TState state) where TState : notnull
```
aaronburtle, line 51: Consider supporting external scopes ie: implement ISupportExternalScope as a future task.

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLoggerProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <param name="writer">The notification writer to use for sending log messages.</param>
+    /// <param name="levelFilter">A function to filter log levels. Returns true if the level should be logged.</param>
+    public McpLoggerProvider(IMcpLogNotificationWriter writer, Func<LogLevel, bool> levelFilter)
+    {
+        _writer = writer ?? throw new ArgumentNullException(nameof(writer));
+        _levelFilter = levelFilter ?? throw new ArgumentNullException(nameof(levelFilter));
+    }
+
+    /// <inheritdoc />
+    public ILogger CreateLogger(string categoryName)
```
aaronburtle, line 43: shouldnt we check _disposed to avoid creating a new McpLogger with a stale reference?

src/Azure.DataApiBuilder.Mcp/Telemetry/McpLoggerProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+    private readonly Func<LogLevel, bool> _levelFilter;
+    private readonly ConcurrentDictionary<string, McpLogger> _loggers = new();
+    private bool _disposed;
+
+    /// <summary>
+    /// Creates a new McpLoggerProvider.
+    /// </summary>
+    /// <param name="writer">The notification writer to use for sending log messages.</param>
+    /// <param name="levelFilter">A function to filter log levels. Returns true if the level should be logged.</param>
+    public McpLoggerProvider(IMcpLogNotificationWriter writer, Func<LogLevel, bool> levelFilter)
```
aaronburtle: this delegate is from `LogLevelProvider.ShouldLog` in `Program.cs`, but `Program.cs` has already used `LogLevelProvider.ShouldLog` in the previous `AddFilter` calls, so by the time McpLogger.Lo is reached LogLevelProvider.ShouldLog is already true. The _levelFilter parameter then on the McpLoggerProvider/McpLogger is then calling the same delegate with the same shared state and gets the same answer.

Id recommend dropping this delegate and updating IsEnabled to something like

`public bool IsEnabled(LogLevel logLevel) => _writer.IsEnabled && logLevel != LogLevel.None;`

aaronburtle (approved):
Looks good once comments are resolved.

Aniruddh25 (approved):

PR #3508, opened 2026-05-05
Fix MCP CLI logging: apply colors and respect config log-level

### LINKED ISSUE #3509: [Bug] MCP stdio mode: CLI logs lack colors and are suppressed when log level is set via config
## Summary
Two bugs in the CLI logger when running DAB in MCP stdio mode:

1. **Missing colors** — When `--LogLevel <level>` is passed, CLI startup logs appear on stderr without color formatting, while engine logs are properly colored.
2. **Logs suppressed** — When the runtime config file specifies `runtime.telemetry.log-level` and no `--LogLevel` CLI flag is provided, CLI startup logs are completely suppressed.

### Bug 1 — Missing colors
```powershell
dab start --mcp-stdio --LogLevel Information --config <path>\dab-config.json
```
- **Expected**: `info:` label in **green**.
- **Actual**: `info:` label in plain **white**.

### Bug 2 — CLI logs suppressed
```powershell
dab start --mcp-stdio --config <path>\dab-config.json
```
…where the config contains:
```json
"runtime": {
  "telemetry": {
    "log-level": { "default": "Information" }
  }
}
```
- **Expected**: CLI startup logs appear (version, config file, "Setting minimum LogLevel…").
- **Actual**: CLI logs are missing entirely.

## Impact
- Users relying on config-based log level lose visibility of CLI startup info in MCP mode.
- Inconsistent visual experience for CLI vs engine logs.

## Related
- PR #3419 — MCP Set Log Level
- PR #3484 — MCP notifications/message

### Version

2.0

Azure SQL

_No response_

_No response_

### Relevant log output

```Text

```

### Code of Conduct

anushakolan:
## Why make this change?

Two bugs were discovered in the CLI logging behavior when running DAB in MCP stdio mode:

1. **Missing colors**: When `start --mcp-stdio --LogLevel <level>` was used, the CLI's own startup logs (e.g., "Setting minimum LogLevel: Information.") appeared on stderr **without any colors**, while engine logs were properly colored. This created an inconsistent visual experience.
2. **Missing CLI logs with config-only log level**: When `start --mcp-stdio` was used with the `runtime.telemetry.log-level` set in the config file (and no `--LogLevel` CLI flag), CLI startup logs were **completely suppressed**, even though the user had clearly expressed an intent to see logs at that level.

## What is this change?

The CLI logger (`CustomLoggerProvider`) only knew about CLI overrides (`--LogLevel`). It never learned that the runtime config file can also explicitly request a log level. As a result:
- It defaulted to `LogLevel.None` (suppress all) whenever `--LogLevel` was absent in MCP mode, ignoring the config.
- The MCP stderr code path also skipped the foreground/background color formatting that the non-MCP path applied.

This change makes the CLI logger aware of both override sources and applies colors consistently in both stdout (non-MCP) and stderr (MCP) paths.

### Modified files:
- `src/Cli/Utils.cs` — Added `IsLogLevelOverriddenByConfig` (bool) and `ConfigLogLevel` (LogLevel) static properties.
- `src/Cli/ConfigGenerator.cs` — In `TryStartEngineWithOptions()`, sets the new properties when `RuntimeConfig.HasExplicitLogLevel()` is true.
- `src/Cli/CustomLoggerProvider.cs`:
  - Constructor: chooses minimum log level from CLI → Config → None.
  - `Log()`: suppresses only when **neither** CLI nor config set a level; applies foreground/background colors before writing the abbreviation to stderr in MCP mode.

## How was this tested?

Manually tested all combinations of MCP / non-MCP × CLI override / config override / no override:

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 1 | MCP, no flags, no config log-level | No CLI logs (clean stdout) | ✅ |
| 2 | MCP + `--LogLevel Information` | CLI logs on stderr with **green** `info:` | ✅ |
| 3 | MCP + config `log-level` only | CLI logs on stderr with **green** `info:` | ✅ Bug 2 fixed |
| 4 | MCP + `--LogLevel Warning` + config `log-level: Information` | CLI wins → no `info:` shown | ✅ |
| 5 | Non-MCP mode | Unchanged behavior | ✅ |
| 6 | MCP + `--LogLevel Warning` (StaticWebApps config) | **Yellow** `warn:` shown | ✅ Bug 1 fixed |

## Note

This is a follow-up bug fix on top of PR #3419 (MCP Set Log Level) and PR #3484 (MCP notifications/message). It targets `main`.


Aniruddh25 (approved):

src/Cli/CustomLoggerProvider.cs
```diff
@@ -116,7 +121,15 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
-                Console.Error.WriteLine($"{mcpAbbreviation}: {formatter(state, exception)}");
+                // Apply colors so the abbreviation matches the visual style of engine logs.
+                ConsoleColor mcpOriginalForeGroundColor = Console.ForegroundColor;
+                ConsoleColor mcpOriginalBackGroundColor = Console.BackgroundColor;
+                Console.ForegroundColor = _logLevelToForeGroundConsoleColorMap.GetValueOrDefault(logLevel, ConsoleColor.White);
+                Console.BackgroundColor = _logLevelToBackGroundConsoleColorMap.GetValueOrDefault(logLevel, ConsoleColor.Black);
+                Console.Error.Write($"{mcpAbbreviation}:");
+                Console.ForegroundColor = mcpOriginalForeGroundColor;
+                Console.BackgroundColor = mcpOriginalBackGroundColor;
+                Console.Error.WriteLine($" {formatter(state, exception)}");
```
souvikghosh04, line 141: should the original foreground/background color be restored to default after the specific scenario exits? also, what if there is an unknown exception and should the colour be restored back to default?

src/Cli.Tests/CustomLoggerTests.cs
```diff
@@ -4,78 +4,179 @@
...
+    /// using <see cref="Cli.Utils.ConfigLogLevel"/>.
+    /// </summary>
+    [TestMethod]
+    public void Mcp_ConfigOverride_WritesToStderrAndHonorsConfigLevel()
+    {
+        Cli.Utils.IsMcpStdioMode = true;
+        Cli.Utils.IsLogLevelOverriddenByConfig = true;
+        Cli.Utils.ConfigLogLevel = LogLevel.Information;
+
+        (string stdout, string stderr) = CaptureConsole(() =>
```
souvikghosh04, line 146: in these tests, does CaptureConsole() capture the colour info or is there a way to add some tests for them?

souvikghosh04 (approved):
Approved with some comments and suggestions.

src/Cli/CustomLoggerProvider.cs
anushakolan: Wrapped both color-mutating blocks (the new MCP-stdio path and the existing non-MCP path) in try/finally so Console.ForegroundColor and Console.BackgroundColor are always restored to their original values, even if Console.Error.Write / writer.Write throws (e.g. IOException on a broken pipe). The formatter(state, exception) call is intentionally kept outside the try/finally since colors have already been restored by then and we don't want a formatter exception to be swallowed by a finally that resets state we no longer care about.

I extended the same fix to the older non-MCP block as well for consistency — the original-color-leak risk exists there too, and that's actually the path where users see colored output in an interactive terminal (e.g. green info: from dab validate).

src/Cli.Tests/CustomLoggerTests.cs
anushakolan: No, `CaptureConsole(`) cannot capture color information.

It redirects `Console.Out / Console.Error` to `StringWriter`, which only buffers text — it discards `Console.ForegroundColor / Console.BackgroundColo`r state entirely. And we can't fall back to reading the colors back via the property getter either: when `stdout` is redirected (which it always is under MSTest / dotnet test), `Console.ForegroundColor` getter returns Gray regardless of what the setter wrote, so the test process is fundamentally blind to color behavior.

anushakolan:
> Approved with some comments and suggestions.

Addressed the comments.

Aniruddh25 (approved):

PR #3520, opened 2026-05-06
Add non-root variant to Dockerfile for dual image publishing

### LINKED ISSUE #3514: Add USER to the Dockerfile
Scanners will complain when we dont have a USER set in the Dockerfile, despite the container already having a user `app`. Add `app` as USER in Dockerfile to make scanners happy.

Link to discussion: https://github.com/Azure/data-api-builder/discussions/3481

There are some issues however with adding a user to the Dockerfile. Mainly, this could be a breaking change.

| Breakage | Avoidable in the image? |
|---|---|
| 1. Bind to port < 1024 | **Partially** (may trigger scanner but we can document how user may fix) |
| 2. Host volumes with mismatched UIDs | **No** host-side concern we can document |
| 3. Writing relative paths under `/App` | **Yes** `chown -R` |
| 4. `docker exec` lands as `app`, not root | **No** |
| 5. Downstream `RUN`/`COPY` in a derived Dockerfile | **No** (derived image must `USER 0` first) |
| 6. K8s `runAsUser` admission rules | **Positive change**, technically a policy break but will run automatically on hardened|

We cannot make #4 and #5 go away without un-doing the scanner fix. This provides for 2 options.

#### 1. Single image, minimize blast radius
What it actually changes for the typical user (docker run -p 5000:5000 ...): nothing observable. Default port works, config reads work, file-sink works.

What we tighten in the Dockerfile to remove all the avoidable breakages:

chown -R app:app /App so any feature writing relative paths under /App keeps working (eliminates #3 entirely, including any future code that writes there).
Keep the default port at 5000 (already > 1024) so #1 only bites users who explicitly override ASPNETCORE_URLS.
Document #1, #4, #5 in release notes as "if you do X, do Y."

#### 2. Two image variants, zero breakage for existing users
Publish two tags from the same code:

`mcr.microsoft.com/azure-databases/data-api-builder:<version>` unchanged, still runs as root. Existing users see no diff.
`mcr.microsoft.com/azure-databases/data-api-builder:<version>-nonroot` runs as app, satisfies scanners.
Users with scanner alerts switch tag. Everyone else is untouched. Next major (v2.x) flips the default to non-root.

This is what nginx, redis, postgres, and the .NET team itself have all done historically when introducing non-root images. It's the gold-standard "no regressions" pattern.

The downside: it requires a small change in the publish pipeline. Adding a second docker build invocation with a different Dockerfile (or --target) and push a second tag.

aaronburtle:
﻿## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3514

The published runtime image (`mcr.microsoft.com/azure-databases/data-api-builder`) runs as root because the Dockerfile never issues a `USER` instruction. Image scanners like Checkmarx One read the image's `Config.User` field and flag any final stage that's empty or `root`. Users who have to satisfy those scanners are blocked from adopting DAB unless they override `securityContext.runAsUser` in their pod spec.

DAB is just an ASP.NET Core process and does not need root privileges, so the fix is to declare a non-root user explicitly. The published `mcr.microsoft.com/dotnet/aspnet:10.0-azurelinux3.0` base image already ships with a non-root user (UID/GID 1654, exposed via the `APP_UID` env var), so no `useradd` layer is needed.

## What is this change?

Restructure the Dockerfile into a multi-target build that publishes **two** runtime variants from the same source:

| Target | Tag (proposed) | Runs as | Backwards compatible? | Scanner clean? |
|---|---|---|---|---|
| `runtime` (default, last stage) | `:<version>` | `root` | identical to today's image | No (unchanged) |
| `runtime-nonroot` (opt-in via `--target`) | `:<version>-nonroot` | UID 1654 (`$APP_UID`) | n/a new variant | Yes |

The root variant is the **last stage in the Dockerfile**, so a plain `docker build .` with no `--target` argument still produces the existing root-running image. No existing user sees a behavior change.

#### Safeguards on the non-root variant

To minimize the chance of runtime breakage when users adopt the non-root tag:

- **`chown $APP_UID:$APP_UID /App/logs`** (non-recursive) so the documented default file-sink path (`runtime.telemetry.file` → `logs/dab-log.txt`, relative to `WORKDIR /App`) is writable by the non-root user. Ownership of the published assemblies under `/App` is intentionally left unchanged — a recursive `chown -R` would duplicate every assembly layer and roughly double the non-root image size for no runtime benefit, since DAB only needs write access to `/App/logs`. Only ownership of `/App/logs` is changed, not file modes.
- **Pre-create `/App/logs`** so the documented default file-sink path works with no extra volume or permission setup.
- **Numeric `USER $APP_UID`** (rather than `USER app`) per .NET container guidance — a numeric UID is friendlier to image scanners and to Kubernetes `runAsNonRoot`/`runAsUser` checks, which cannot resolve a username to a UID at admission time.
- **Default port stays at `5000`**, which is above 1024, so binding works without `CAP_NET_BIND_SERVICE`. Users overriding `ASPNETCORE_URLS` to a privileged port (<1024) must add `--cap-add=NET_BIND_SERVICE` or front DAB with a reverse proxy.
- **OCI labels** on the non-root variant for clarity in registries/tooling.

#### Known caveats for consumers of the non-root variant

These are inherent to running any non-root container and cannot be fully eliminated in the image. They should be called out in release notes:

- **Host bind-mounts** (config, logs, certs, etc.) must be readable and writable, if DAB needs to write them, by UID 1654 on the host. Either `chown -R 1654:1654 /host/path` or, in Kubernetes, set `securityContext.fsGroup: 1654`.
- **`docker exec`** defaults to UID 1654. Use `docker exec --user 0` for administrative actions inside a running container.
- **Downstream Dockerfiles** (`FROM <this image>`) that need to install packages or write outside `/App` should add `USER 0` before those instructions, then restore `USER $APP_UID` at the end.

## Manual testing performed

Both variants were built locally and verified end-to-end:

| Check | `:<version>` (default) | `:<version>-nonroot` |
|---|---|---|
| `docker build` succeeds | yes | yes |
| `Config.User` (what scanners read) | empty (= root) | `app` |
| Runtime UID inside container | `0(root)` | `1654(app)` |
| `/App` ownership | `root:root` | `app:app` (recursive) |
| `/App/logs` pre-created | n/a | yes |
| DAB process starts | yes | yes |
| File-sink telemetry writes a log file | n/a | yes (`dab-log20260519.txt` written to `/App/logs/` by UID 1654) |
| Behavior matches previously published image | yes (byte-for-byte equivalent) | n/a (new variant) |

#### Follow-up work (separate PR)

The publish pipeline needs to be updated to build and push the second tag. That change lives outside this repo and will be done after this PR merges.


Aniruddh25 (changes requested):
Copilot suggestion is valid.

Dockerfile
```diff
@@ -14,4 +14,13 @@ COPY --from=build /out /App
...
 WORKDIR /App
 ENV ASPNETCORE_URLS=http://+:5000
+
+# Run as the non-root "app" user (UID/GID 64198) that ships with the
+# mcr.microsoft.com/dotnet/aspnet base image. DAB is just an ASP.NET Core
+# process and does not require root privileges. Declaring USER explicitly
+# sets the image's Config.User field so image scanners (e.g. Checkmarx One)
+# that require a non-root user in the final stage are satisfied.
+# Port 5000 is above 1024 so a non-root user can bind to it without CAP_NET_BIND_SERVICE.
+USER app
```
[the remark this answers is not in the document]
aaronburtle: Addressed with new approach.

Dockerfile
```diff
@@ -7,11 +7,75 @@ WORKDIR /src
...
+#     `securityContext.fsGroup: 1654`.
+#   * `docker exec` defaults to the `app` user. Use `docker exec --user 0`
+#     for administrative actions inside a running container.
+#   * Downstream Dockerfiles (FROM <this image>) that need to install
+#     packages or write outside /App should add `USER 0` before those
+#     instructions, then restore `USER app` at the end.
+# ---------------------------------------------------------------------------
+FROM runtime-base AS runtime-nonroot
+
+RUN mkdir -p /App/logs && chown -R app:app /App
```
souvikghosh04: chown -R app:app /App doubles the non-root image size. This creates duplicate layer over the published .NET assemblies. DAB only needs write access to /App/logs, not to the assemblies

souvikghosh04: either drop the recursion: RUN mkdir -p /App/logs && chown app:app /App/logs or push the copy into the non-root stage as COPY --chown=app:app --from=build /out /App (requires moving the COPY out of the shared runtime-base).

souvikghosh04 (changes requested):

Dockerfile
```diff
@@ -7,11 +7,75 @@ WORKDIR /src
...
+# explicitly sets the image's Config.User field so image scanners
+# (e.g. Checkmarx One) that require a non-root user in the final stage are
+# satisfied.
+#
+# Safeguards applied to minimize the chance of runtime breakage:
+#   * Pre-create /App/logs and `chown app:app /App/logs` (non-recursive) so
+#     the documented default file-sink path ("logs/dab-log.txt", relative to
+#     WORKDIR /App) is writable. Ownership of the published assemblies under
+#     /App is intentionally left unchanged - a recursive chown would
+#     duplicate every assembly layer and roughly double the image size for
```
Aniruddh25, line 46: The comment here is contradictory to the description in the PR which says that we are doing:  `chown -R app:app /App` recursive. 

However, not doing recursive seems the right thing to do. The PR description should be fixed.

Dockerfile
```diff
@@ -7,11 +7,75 @@ WORKDIR /src
...
+#   docker build -t <repo>:<version> .                       # no --target needed
+#   docker build --target runtime -t <repo>:<version> .
+#
+# This is the LAST stage in the file, so a plain `docker build` with no
+# --target argument produces this image. Keeping the root-running variant
+# as the default preserves backwards compatibility with the previously
+# published image - existing users see no behavior change. The non-root
+# variant is opt-in via `--target runtime-nonroot`.
+# ---------------------------------------------------------------------------
+FROM runtime-base AS runtime
```
Aniruddh25, line 84: Followup separate PR in internal repo would need changes to both the pipeline yml as well as the docker file used in one branch. The Dockerfile that is changed here is only shown as a sample for building the image by customers

Aniruddh25, line 84: If the FROM statement as runtime, is the last in the Dockerfile, would the ENTRYPOINT mentioned on line 25 still stay the same? 

Did you actually test this to build and start dab in the container?

Dockerfile
```diff
@@ -7,11 +7,75 @@ WORKDIR /src
...
+# ---------------------------------------------------------------------------
+# Default (root-running) variant. Build with either:
+#   docker build -t <repo>:<version> .                       # no --target needed
+#   docker build --target runtime -t <repo>:<version> .
+#
+# This is the LAST stage in the file, so a plain `docker build` with no
+# --target argument produces this image. Keeping the root-running variant
+# as the default preserves backwards compatibility with the previously
+# published image - existing users see no behavior change. The non-root
+# variant is opt-in via `--target runtime-nonroot`.
```
Aniruddh25, line 82: Was the requirement for this PR approved by any PM in any of our scrum syncs?

Aniruddh25 (changes requested):
Need some clarifications.

Dockerfile
aaronburtle: Yes, and I did test but not since I updated the command to non recursive, will validate everything still functions.

Dockerfile
aaronburtle: I believe @JerryNixon was present when this was discussed but we didn't have a formal design sync.

Dockerfile
aaronburtle: ive added an issue to track this follwup PR in the internal repo: https://github.com/Azure/data-api-builder/issues/3713

Aniruddh25 (approved):

souvikghosh04 (approved):
LGTM. thanks for addressing the comments. let's have this tested properly.

PR #3521, opened 2026-05-06
Refactor how we build pagination response to support collections in SQL

### LINKED ISSUE #3462: Possible bug related to new Array data types with Pagination
We handle pagination currently, by checking if the last item in the DB response is array type, and then if it is we know that is a set of pagination columns. But this was done when there were no array types in SQL. Now that we are going to support array types in SQL, need to double check that this won't break out pagination logic.

aaronburtle:
## Why make this change?

Closes https://github.com/Azure/data-api-builder/issues/3462

## What is this change?

Refactors how SQL Find responses signal pagination so it no longer relies on the JSON shape of the result list.

Previously, `SqlResponseHelpers.FormatFindResult` appended pagination metadata to the row collection as an **in-band sentinel** a single-element JSON array wrapping `{ "nextLink": ... }` (REST) or `{ "after": ... }` (MCP) and `OkResponse` recovered it by checking whether the last element of the result list was a `JsonValueKind.Array`. This worked only because SQL backends never produced array-shaped row values.

That assumption breaks with SQL Server's `JSON` type, `JSON_ARRAY()`, `vector`, and other collection-shaped column values a real data row could carry an array-shaped tail value and be misclassified as a pagination envelope.

This change carries pagination metadata **out-of-band**:

- `FormatFindResult` builds the response envelope directly as a typed projection (`{ value, nextLink }` for REST, `{ value, after }` for MCP) using a `bool hasNext` local instead of encoding the decision into the row collection.
- `OkResponse(JsonElement, bool?)` is reduced to a single-responsibility envelope wrapper. **The signature is preserved** (`isMcpRequest` is now optional and accepted-but-unused) so existing callers of `Microsoft.DataApiBuilder.Core` continue to compile and link unchanged. The shape-detection branch is removed from the body.
- `GetConsolidatedNextLinkForPagination` is preserved with its original signature and `JsonElement` return type for source compatibility. Internally, `FormatFindResult` now constructs the next-page URL inline (it needs the URL as a `string` to put into the typed response envelope), avoiding a serialize-then-parse round-trip just to satisfy the old wrapper shape.
- `DetermineExtraFieldsInResponse` gains a `JsonValueKind.Object` guard so future array/scalar/collection-typed row shapes return an empty extra-fields set rather than throwing on `EnumerateObject()`.
- Drive-by cleanup: removed a dead `!` null-forgiving operator, standardized `JsonValueKind` checks on `is`, refreshed stale comments and XML docs, removed dead commented-out code.

**Wire format is unchanged** for both REST and MCP. **Public API surface is byte-identical to `main`** no signatures added, removed, or renamed; no `[Obsolete]` warnings emitted to downstream consumers. The new gate:

```csharp
bool hasNext = isCollection && SqlPaginationUtil.HasNext(...)
```

is logically equivalent to the original gate.

## How was this tested?

Added seven focused unit tests in `SqlResponseHelpersUnitTests.cs` pinning the envelope shape across every branch:

1. Empty result
2. Single object
3. Collection without next page
4. Collection with next page (REST → `nextLink`)
5. Collection with next page (MCP → `after`)
6. Round-trip of a row containing an array-valued column (forward-looking coverage for SQL Server JSON/vector column shapes)
7. **Regression guard** for the actual pre-refactor failure mode: a result whose last top-level element is itself a JSON array the exact trigger the old in-band sentinel detection used is now correctly preserved as ordinary data instead of being unpacked as a pagination envelope.

All seven pass. `dotnet format --verify-no-changes` is clean across the touched files.


aaronburtle:
@copilot can you review this PR please?

src/Core/Resolvers/SqlPaginationUtil.cs
```diff
@@ -659,17 +656,9 @@ public static JsonElement GetConsolidatedNextLinkForPagination(string baseUri, s
...
                 : uriBuilder.Uri.AbsoluteUri; // returns full URL
-
-            // Return serialized JSON object
-            string jsonString = JsonSerializer.Serialize(new[]
-            {
-                new { nextLink = nextLinkValue }
-            });
-
-            return JsonSerializer.Deserialize<JsonElement>(jsonString);
         }
```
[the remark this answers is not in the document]
aaronburtle: Valid, will revert to maintain signatures and return types.

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -200,60 +201,26 @@ private static JsonElement RemoveExtraFieldsInResponseWithSingleItem(JsonElement
...
-                    @nextLink = nextLinkAfter["nextLink"]
-                });
-            }
-
-            // no pagination, do not need nextLink
             return new OkObjectResult(new
             {
-                value = resultEnumerated
+                value = rows
             });
```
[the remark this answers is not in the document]
aaronburtle: Valid, will revert to maintain signatures

src/Service.Tests/UnitTests/SqlResponseHelpersUnitTests.cs
```diff
@@ -0,0 +1,292 @@
...
+        /// <see cref="SqlResponseHelpers.OkResponse"/> would have misclassified this shape.
+        /// </summary>
+        [TestMethod]
+        public void FormatFindResult_RowWithArrayColumn_NotMisclassifiedAsPagination()
+        {
+            // Two rows with array-valued "tags" column. first=5 so HasNext=false.
+            JsonElement input = ParseJson(@"[
+                { ""id"": 1, ""tags"": [ ""sci-fi"", ""classic"" ] },
+                { ""id"": 2, ""tags"": [ ""fantasy"" ] }
+            ]");
```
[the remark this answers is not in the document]
aaronburtle: Valid, fixing.

aaronburtle:
@copilot I addressed your comments, can you do another review please?

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -200,60 +205,34 @@ private static JsonElement RemoveExtraFieldsInResponseWithSingleItem(JsonElement
...
         /// <returns>Correctly formatted OkObjectResult.</returns>
         public static OkObjectResult OkResponse(JsonElement jsonResult, bool? isMcpRequest = null)
         {
-            // For consistency we return all values as type Array
-            if (jsonResult.ValueKind != JsonValueKind.Array)
-            {
-                string jsonString = $"[{JsonSerializer.Serialize(jsonResult)}]";
-                jsonResult = JsonSerializer.Deserialize<JsonElement>(jsonString);
-            }
+            _ = isMcpRequest; // intentionally unused; kept for source compatibility.
```
souvikghosh04: this seems like a dead code and should be refactored, including the method params and callers

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -41,49 +41,48 @@ public static OkObjectResult FormatFindResult(
...
             }
 
-            List<JsonElement> rootEnumerated = findOperationResponse.EnumerateArray().ToList();
+            // Paginated path.
+            List<JsonElement> rows = findOperationResponse.EnumerateArray().ToList();
 
             // More records exist than requested, we know this by requesting 1 extra record,
             // that extra record is removed here.
-            rootEnumerated.RemoveAt(rootEnumerated.Count - 1);
+            rows.RemoveAt(rows.Count - 1);
```
souvikghosh04, line 79: may be add a check if rows null or 0?

souvikghosh04 (approved):
Approved with minor suggesstions

src/Core/Resolvers/SqlResponseHelpers.cs
aaronburtle: I initially did refactor this out, but the problem is that this is a public method in a public class, so any external callers that were using this would break from such a refactor. We don't really know if there are any external callers however, so we could try changing it and just hope no one is using.

src/Core/Resolvers/SqlResponseHelpers.cs
aaronburtle: Don't nee that guard here because `HasNext` guarantees row count of at least 1, we could assert to confirm the invariant, but it isn't really needed.

src/Core/Resolvers/SqlResponseHelpers.cs
aaronburtle: After checking with Fabric/GraphQL team, it seems that no external callers are going to be using this, so ive removed. This is the right time to make this change.

src/Core/Resolvers/SqlResponseHelpers.cs
```diff
@@ -94,40 +93,38 @@ public static OkObjectResult FormatFindResult(
...
 
-            // Create an 'after' object if the request comes from MCP endpoint.
+            // MCP endpoint: { value: [...], after: "<cursor>" }
             if (isMcpRequest is true)
             {
-                string jsonString = JsonSerializer.Serialize(new[]
+                return new OkObjectResult(new
                 {
-                    new { after = after }
+                    value = rows,
```
RubenCerna2079, line 104: Just curious, is there a reason we want to add the new value object? Or was that expected to be there from the start?

src/Service.Tests/UnitTests/SqlResponseHelpersUnitTests.cs
```diff
@@ -0,0 +1,340 @@
...
+            Assert.IsFalse(envelope.TryGetProperty("after", out _),
+                "REST paginated response must NOT carry an 'after' field.");
+        }
+
+        /// <summary>
+        /// MCP: a collection with more rows than requested produces the
+        /// <c>{ "value": [ ... ], "after": "..." }</c> envelope and trims the +1 probe row.
+        /// </summary>
+        [TestMethod]
+        public void FormatFindResult_CollectionWithNextPage_Mcp_ReturnsAfterEnvelope()
```
RubenCerna2079, line 161: Couldn't this and the REST test be refactored together, since they use most of the same things?

src/Service.Tests/UnitTests/SqlResponseHelpersUnitTests.cs
```diff
@@ -0,0 +1,340 @@
...
+        /// <summary>
+        /// Pins that rows containing array-valued columns (e.g. a SQL Server JSON array, vector,
+        /// or other collection-typed column) round-trip through the response pipeline unchanged.
+        /// This is forward-looking coverage for query shapes enabled by SQL Server's JSON/vector
+        /// types: the array values live inside object-shaped rows, so this case did not actually
+        /// trigger the pre-refactor shape sentinel — but it documents the supported shape and
+        /// guards against future regressions in extra-field stripping or envelope construction.
+        /// </summary>
+        [TestMethod]
+        public void FormatFindResult_RowWithArrayColumn_RoundTripsUnchanged()
```
RubenCerna2079, line 196: I think we should also add a datarow to test if we get a nextLink even when we use array values.

src/Core/Resolvers/SqlResponseHelpers.cs
aaronburtle: Ya, this is expected, we are just refactoring to make this shape explicit in the pagination branch.

src/Service.Tests/UnitTests/SqlResponseHelpersUnitTests.cs
aaronburtle: it could be but this would require adding a flag and having branching logic in the test, which is harder to read.

src/Service.Tests/UnitTests/SqlResponseHelpersUnitTests.cs
aaronburtle: added

src/Service.Tests/UnitTests/SqlResponseHelpersUnitTests.cs
```diff
@@ -218,6 +221,58 @@ public void FormatFindResult_RowWithArrayColumn_RoundTripsUnchanged()
...
+            Assert.AreEqual(JsonValueKind.Array, tags.ValueKind, "Array-typed column must remain an array.");
+            Assert.AreEqual(2, tags.GetArrayLength());
+            Assert.AreEqual("sci-fi", tags[0].GetString());
+            Assert.AreEqual("classic", tags[1].GetString());
+
+            Assert.IsTrue(envelope.TryGetProperty("nextLink", out JsonElement nextLink),
+                "REST paginated response with array-valued columns must still carry a 'nextLink' field.");
+            Assert.IsTrue(nextLink.GetString()!.Contains("$after="), "nextLink should encode the $after cursor.");
+            Assert.IsFalse(envelope.TryGetProperty("after", out _),
+                "REST paginated response must NOT carry an 'after' field.");
```
RubenCerna2079, line 276: Can we also add to the test a second `SqlResponseHelpers.FormatFindResult` that uses the `nextLink` and asserts we get the correct values `id: 2, tags: [fantasy]` in the response?

aaronburtle: ya, makes sense, adding

RubenCerna2079 (approved):
LGTM!

PR #3544, opened 2026-05-12
Allow MCP agent to override CLI and Config log levels (Agent > CLI > Config)

### LINKED ISSUE #3533: [Bug]: Log-level precedence should be Agent > CLI > Config (MCP `logging/setLevel` ignored when CLI/Config set)
## Description

Today, the runtime log-level precedence is **CLI > Config > Agent**. When either `--LogLevel` is passed on the CLI or `runtime.telemetry.log-level` is set in config, MCP `logging/setLevel` requests are silently rejected and the agent has no way to change the level at runtime.

Per the team's discussion locked on 5/1 (Jerry, confirmed by Aniruddh), the intended precedence is **Agent > CLI > Config**:

1. If nothing is set anywhere → defaults (Production = Error, Development = Debug; `--mcp-stdio` = None).
2. Config sets the default level.
3. CLI overrides Config.
4. Agent (`logging/setLevel`) overrides CLI and Config.

## Expected

- With `dab start --LogLevel Error` running, an MCP `logging/setLevel debug` request takes effect and subsequent logs include Debug entries.
- With `runtime.telemetry.log-level: error` in config, an MCP `logging/setLevel debug` request takes effect.
- CLI continues to override Config when the agent doesn't speak up.

## Actual

- `DynamicLogLevelProvider.UpdateFromMcp` returns `false` when `IsCliOverridden || IsConfigOverridden`, so the level never changes.
- Unit tests `"CLI override blocks MCP change"` and `"Config override blocks MCP change"` lock this behavior.

## Repro

1. `dab start --LogLevel Error --mcp-stdio` (or any host with `--LogLevel`).
2. Send MCP request:
   ```json
   {"jsonrpc":"2.0","method":"logging/setLevel","params":{"level":"debug"},"id":1}
   ```
3. Trigger any operation that would emit Debug logs.
4. Observe: no Debug entries appear; the agent's request was silently ignored.

## Proposed fix

1. Remove the CLI / Config early-rejects in `DynamicLogLevelProvider.UpdateFromMcp`.
2. Add `IsAgentOverridden` and have `UpdateFromRuntimeConfig` skip when it's set (so hot-reload doesn't clobber the agent).
3. Emit an `Information` log line when the agent successfully changes the level, so operators can see the override.
4. Update existing tests to reflect Agent > CLI > Config; add a hot-reload-after-agent test and an agent-beats-CLI test.
5. Manual: `dab start --mcp-stdio --LogLevel Error`, send `logging/setLevel debug`, verify logs flow at Debug.

## Out of scope

- CLI > Config rule (unchanged — CLI still beats Config when the agent doesn't speak up).
- Default levels (Production = Error, Development = Debug, `--mcp-stdio` = None — all unchanged).
- Agent "release" of override (no team ask yet — call as a follow-up if reviewers raise it).

anushakolan:
## Why make this change?

Closes #3533— *Log-level precedence should be Agent > CLI > Config; MCP `logging/setLevel` is silently ignored when CLI/Config set a level.*

The runtime today implements **CLI > Config > Agent**: once `--LogLevel` is passed or `runtime.telemetry.log-level` is set, the agent's request silently no-ops. There is no way for an agent to dial verbosity up or down at runtime, which defeats the purpose of MCP `logging/setLevel`.

The team locked the corrected precedence on 5/1 : **Agent > CLI > Config > defaults**. The agent must always be able to override; CLI still beats Config when the agent doesn't speak up; defaults (Production = Error, Development = Debug, `--mcp-stdio` = None) are unchanged.

## What is this change?

[`DynamicLogLevelProvider`](src/Service/Telemetry/DynamicLogLevelProvider.cs) is the single chokepoint for all log-level updates. The fix:

- `UpdateFromMcp` no longer rejects when CLI or Config has set a level. It updates the level, flips a new `IsAgentOverriding` flag, and emits an `Information` audit log ("Log level updated to {Level} via MCP logging/setLevel (agent override).").
- `UpdateFromRuntimeConfig` is guarded so a config hot-reload won't overwrite an Agent- or CLI-set level.
- [`ILogLevelController`](src/Core/Telemetry/ILogLevelController.cs) gains `IsAgentOverriding` and updated XML docs.
- [`Startup.cs`](src/Service/Startup.cs) wires a logger on the provider once `ILoggerFactory` exists, so the audit log reaches the standard pipeline.
- [`McpStdioServer.HandleSetLevel`](src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs) enables the MCP notification writer **before** calling `UpdateFromMcp` so the audit log surfaces to the agent.

CLI > Config, default levels, and the `--mcp-stdio` default of `None` are unchanged.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests
- [x] Manual end-to-end test against live SQL Server with the MCP Inspector

### Unit tests

Added 1 new test and updated existing ones in [`DynamicLogLevelProviderTests`](src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs) to lock Agent > CLI > Config under both single-call and concurrent paths. All 9 tests pass.

### Manual end-to-end test

Started the engine with `--mcp-stdio --LogLevel Error` and drove it from the MCP Inspector against a live SQL Server.

| # | What | How | Result |
| --- | --- | --- | --- |
| 1 | Agent sets level to debug | Send `logging/setLevel { level: "debug" }` | Success response ✅ |
| 2 | Audit log surfaces | Watch MCP notifications | Info notification with override message received ✅ |
| 3 | Debug logs flow despite `--LogLevel Error` | Trigger activity (e.g. `tools/list`) | Debug notifications received ✅ |

Also spot-checked HTTP mode: Config-only, CLI-over-Config, and Config-alone all behave as before.

### MCP `logging/setLevel` from the agent

Request:

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "logging/setLevel",
  "params": { "level": "debug" }
}
```

Response:

```json
{ "jsonrpc": "2.0", "id": 2, "result": {} }
```

Followed by the new visibility notification:

```json
{
  "jsonrpc": "2.0",
  "method": "notifications/message",
  "params": {
    "level": "info",
    "logger": "Azure.DataApiBuilder.Service.Telemetry.DynamicLogLevelProvider",
    "data": "Log level updated to Debug via MCP logging/setLevel (agent override)."
  }
}
```

### CLI usage (precedence still works as expected)

```bash
# Agent can override even when CLI pins the level:
dab start --mcp-stdio --LogLevel Error
# → agent sends logging/setLevel debug → debug logs flow

# CLI still beats Config when no agent is involved:
dab start --LogLevel Warning --config dab-config.json   # config has log-level: Information
# → only Warning and above appear

# Config alone, no CLI, no agent — unchanged:
dab start --config dab-config.json                      # config has log-level: Information
# → info: lines visible, debug: filtered out
```

### DIFF
--- a/src/Core/Telemetry/ILogLevelController.cs
+++ b/src/Core/Telemetry/ILogLevelController.cs
@@ -11,22 +11,24 @@ namespace Azure.DataApiBuilder.Core.Telemetry
     public interface ILogLevelController
     {
         /// <summary>
-        /// Gets a value indicating whether the log level was overridden by CLI arguments.
-        /// When true, runtime-config (hot-reload) updates are ignored.
+        /// Gets a value indicating whether the CLI is the source overriding the log level
+        /// (i.e., <c>--LogLevel</c> was supplied). When true, runtime-config (hot-reload)
+        /// updates are ignored.
         /// </summary>
-        bool IsCliOverridden { get; }
+        bool IsCliOverriding { get; }
 
         /// <summary>
-        /// Gets a value indicating whether the log level was explicitly set in the config file.
+        /// Gets a value indicating whether the runtime config is the source overriding the log
+        /// level (i.e., <c>runtime.telemetry.log-level</c> was explicitly set).
         /// </summary>
-        bool IsConfigOverridden { get; }
+        bool IsConfigOverriding { get; }
 
         /// <summary>
-        /// Gets a value indicating whether the log level has been overridden by an MCP
-        /// <c>logging/setLevel</c> request from the agent. When true, runtime-config (hot-reload)
-        /// updates are ignored so the agent's choice remains in effect.
+        /// Gets a value indicating whether the agent is the source overriding the log level via
+        /// an MCP <c>logging/setLevel</c> request. When true, runtime-config (hot-reload) updates
+        /// are ignored so the agent's choice remains in effect.
         /// </summary>
-        bool IsAgentOverridden { get; }
+        bool IsAgentOverriding { get; }
 
         /// <summary>
         /// Updates the log level from an MCP logging/setLevel request.
--- a/src/Core/Telemetry/ILogLevelController.cs
+++ b/src/Core/Telemetry/ILogLevelController.cs
@@ -12,26 +12,33 @@ namespace Azure.DataApiBuilder.Core.Telemetry
     {
         /// <summary>
         /// Gets a value indicating whether the log level was overridden by CLI arguments.
-        /// When true, MCP and config-based log level changes are ignored.
+        /// When true, runtime-config (hot-reload) updates are ignored.
         /// </summary>
         bool IsCliOverridden { get; }
 
         /// <summary>
         /// Gets a value indicating whether the log level was explicitly set in the config file.
-        /// When true along with IsCliOverridden being false, MCP log level changes are ignored.
         /// </summary>
         bool IsConfigOverridden { get; }
 
         /// <summary>
+        /// Gets a value indicating whether the log level has been overridden by an MCP
+        /// <c>logging/setLevel</c> request from the agent. When true, runtime-config (hot-reload)
+        /// updates are ignored so the agent's choice remains in effect.
+        /// </summary>
+        bool IsAgentOverridden { get; }
+
+        /// <summary>
         /// Updates the log level from an MCP logging/setLevel request.
         /// The MCP level string is mapped to the appropriate LogLevel.
-        /// Log level precedence (highest to lowest):
-        /// 1. CLI --LogLevel flag (IsCliOverridden = true)
-        /// 2. Config runtime.telemetry.log-level (IsConfigOverridden = true)
-        /// 3. MCP logging/setLevel (only works if neither CLI nor Config set a level)
+        /// Log-level precedence (highest to lowest):
+        /// 1. Agent (MCP <c>logging/setLevel</c>) — always wins.
+        /// 2. CLI <c>--LogLevel</c> flag.
+        /// 3. Config <c>runtime.telemetry.log-level</c>.
+        /// 4. Defaults.
         /// </summary>
         /// <param name="mcpLevel">The MCP log level string (e.g., "debug", "info", "warning", "error").</param>
-        /// <returns>True if the level was changed; false if CLI or Config override prevented the change.</returns>
+        /// <returns>True if the level was changed; false if the input was an unrecognized level.</returns>
         bool UpdateFromMcp(string mcpLevel);
     }
 }
--- a/src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs
+++ b/src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs
@@ -3,7 +3,9 @@
 
 #nullable enable
 
+using System.Collections.Generic;
 using System.Threading.Tasks;
+using Azure.DataApiBuilder.Config.ObjectModel;
 using Azure.DataApiBuilder.Service.Telemetry;
 using Microsoft.Extensions.Logging;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -12,16 +14,17 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
 {
     /// <summary>
     /// Unit tests for the DynamicLogLevelProvider class.
-    /// Tests the MCP logging/setLevel support.
+    /// Locks the runtime log-level precedence: Agent (MCP) > CLI > Config > defaults.
     /// </summary>
     [TestClass]
     public class DynamicLogLevelProviderTests
     {
         [DataTestMethod]
         [DataRow(LogLevel.Error, false, false, "debug", true, LogLevel.Debug, DisplayName = "Valid level change succeeds")]
-        [DataRow(LogLevel.Error, true, false, "debug", false, LogLevel.Error, DisplayName = "CLI override blocks MCP change")]
-        [DataRow(LogLevel.Warning, false, true, "debug", false, LogLevel.Warning, DisplayName = "Config override blocks MCP change")]
-        [DataRow(LogLevel.Error, false, false, "invalid", false, LogLevel.Error, DisplayName = "Invalid level returns false")]
+        [DataRow(LogLevel.Error, true, false, "debug", true, LogLevel.Debug, DisplayName = "Agent overrides CLI")]
+        [DataRow(LogLevel.Warning, false, true, "debug", true, LogLevel.Debug, DisplayName = "Agent overrides Config")]
+        [DataRow(LogLevel.Error, true, true, "debug", true, LogLevel.Debug, DisplayName = "Agent overrides CLI + Config")]
+        [DataRow(LogLevel.Error, false, false, "invalid", false, LogLevel.Error, DisplayName = "Invalid level returns false and leaves level untouched")]
         public void UpdateFromMcp_ReturnsExpectedResult(
             LogLevel initialLevel,
             bool isCliOverridden,
@@ -40,6 +43,10 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
             // Assert
             Assert.AreEqual(expectedResult, result);
             Assert.AreEqual(expectedFinalLevel, provider.CurrentLogLevel);
+
+            // Successful agent updates must flip IsAgentOverridden so hot-reloads of Config
+            // don't clobber the agent's choice.
+            Assert.AreEqual(expectedResult, provider.IsAgentOverridden);
         }
 
         [TestMethod]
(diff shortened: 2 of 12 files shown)


src/Core/Telemetry/ILogLevelController.cs
```diff
@@ -12,26 +12,33 @@ public interface ILogLevelController
...
-        /// When true along with IsCliOverridden being false, MCP log level changes are ignored.
         /// </summary>
         bool IsConfigOverridden { get; }
 
+        /// <summary>
+        /// Gets a value indicating whether the log level has been overridden by an MCP
+        /// <c>logging/setLevel</c> request from the agent. When true, runtime-config (hot-reload)
+        /// updates are ignored so the agent's choice remains in effect.
+        /// </summary>
+        bool IsAgentOverridden { get; }
```
RubenCerna2079: nit: I think it would be better to call this variable `IsAgentOverriding`. As right now it can imply that it is being overridden by something else.
Also applies for CLI and Config variables.

src/Service/Telemetry/DynamicLogLevelProvider.cs
```diff
@@ -6,88 +6,116 @@ namespace Azure.DataApiBuilder.Service.Telemetry
...
+                CurrentLogLevel = logLevel;
+                IsCliOverridden = isCliOverridden;
+                IsConfigOverridden = isConfigOverridden;
+            }
         }
 
         /// <summary>
-        /// Updates the current log level based on the runtime configuration, unless it was overridden by the CLI.
+        /// Updates the current log level from a runtime-config (hot-reload) change.
+        /// Skipped when the CLI or the agent has already overridden, so neither is clobbered.
```
RubenCerna2079: That is a fancy word I have never heard of 😄

src/Service/Telemetry/DynamicLogLevelProvider.cs
```diff
@@ -6,88 +6,116 @@ namespace Azure.DataApiBuilder.Service.Telemetry
...
-            // Unknown level - don't change, but don't fail either
-            return false;
+            // Surface the override so operators can see the agent moved the level.
+            // Logged outside the lock so logger sinks can't deadlock with state mutations.
+            //
+            // Emit at max(Information, newLevel) so the audit line is never filtered by the
+            // freshly-applied level. Without this, an agent that raised the floor to Warning
+            // / Error / Critical would never see its own confirmation message because the
+            // Information line would fall below the new minimum.
+            LogLevel auditLevel = logLevel > LogLevel.Information ? logLevel : LogLevel.Information;
```
RubenCerna2079: I don't understand why we always want to log this message for the agent. Wouldn't it be better to just log an error/critical message for when it fails and leave this at the information level?

src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs
RubenCerna2079: nit: Same issue as mentioned above, think it is better to have the variables say `Overriding`

src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs
```diff
@@ -113,27 +120,128 @@ public void UpdateFromMcp_UnderConcurrency_StaysConsistent()
...
+
+        /// <summary>
+        /// Race regression: a runtime-config hot-reload must never overwrite an agent-set
+        /// level even when both calls execute concurrently. Without a shared lock, a
+        /// <see cref="DynamicLogLevelProvider.UpdateFromRuntimeConfig"/> caller could pass
+        /// the <c>IsAgentOverridden</c> guard before <see cref="DynamicLogLevelProvider.UpdateFromMcp"/>
+        /// flips the flag, then write the config-pinned level on top of the agent's choice.
+        /// </summary>
+        [TestMethod]
+        public void UpdateFromMcp_BeatsConcurrentConfigHotReload()
```
RubenCerna2079: How is this test fundamentally different from `UpdateFromRuntimeConfig_RespectsAgentOverride`? They both change the log level with the agent and then try to update the log level with hot-reload and expect the agent to succeed. I think one of the two tests should be deleted.

RubenCerna2079: I also think we should do a test to ensure the hot-reload doesn't change the log level when the `isCliOverridden` is set to true

src/Service/Telemetry/DynamicLogLevelProvider.cs
```diff
@@ -6,88 +6,112 @@ namespace Azure.DataApiBuilder.Service.Telemetry
...
-            {
-                return false;
-            }
-
-            if (McpLogLevelConverter.TryConvertFromMcp(mcpLevel, out LogLevel logLevel))
+            lock (_stateLock)
             {
                 CurrentLogLevel = logLevel;
-                return true;
+                IsAgentOverriding = true;
```
souvikghosh04, line 102: shouldn't this be reset back to previous/default state after the completion of this and caller? also, consider having a test case coverage on the reset scenario.

src/Service/Program.cs
```diff
@@ -212,7 +212,7 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
                     // LogLevelProvider was already initialized in StartEngine before CreateHostBuilder.
                     // Use the already-set values to avoid re-parsing args.
                     Startup.MinimumLogLevel = LogLevelProvider.CurrentLogLevel;
-                    Startup.IsLogLevelOverriddenByCli = LogLevelProvider.IsCliOverridden;
+                    Startup.IsLogLevelOverriddenByCli = LogLevelProvider.IsCliOverriding;
```
souvikghosh04: nit: `IsLogLevelOverriddenByCli` seems to have been renamed to `isCliOverriding`. consider, renaming this Startup property for inconsistency. similar to how its done in `GetLogLevelFromCommandLineArgsOrConfig()`

src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs
```diff
@@ -12,42 +14,47 @@ namespace Azure.DataApiBuilder.Service.Tests.UnitTests
...
         [DataTestMethod]
         [DataRow(LogLevel.Error, false, false, "debug", true, LogLevel.Debug, DisplayName = "Valid level change succeeds")]
-        [DataRow(LogLevel.Error, true, false, "debug", false, LogLevel.Error, DisplayName = "CLI override blocks MCP change")]
-        [DataRow(LogLevel.Warning, false, true, "debug", false, LogLevel.Warning, DisplayName = "Config override blocks MCP change")]
-        [DataRow(LogLevel.Error, false, false, "invalid", false, LogLevel.Error, DisplayName = "Invalid level returns false")]
+        [DataRow(LogLevel.Error, true, false, "debug", true, LogLevel.Debug, DisplayName = "Agent overrides CLI")]
+        [DataRow(LogLevel.Warning, false, true, "debug", true, LogLevel.Debug, DisplayName = "Agent overrides Config")]
+        [DataRow(LogLevel.Error, true, true, "debug", true, LogLevel.Debug, DisplayName = "Agent overrides CLI + Config")]
+        [DataRow(LogLevel.Error, false, false, "invalid", false, LogLevel.Error, DisplayName = "Invalid level returns false and leaves level untouched")]
         public void UpdateFromMcp_ReturnsExpectedResult(
```
souvikghosh04, line 28: may be add a test where agent first set "debug" and then "none" (and vice a versa). this might catch the other comment/issue about resetting the `IsAgentOverriding` I mentioned above. you can add a separate test outside this test method if needed.

souvikghosh04 (approved):
Posted the comments which needs some changes. Approving this assuming those gets addressed.

src/Service/Telemetry/DynamicLogLevelProvider.cs
anushakolan: The stickiness is intentional and is the mechanism that enforces Agent > Config: once the agent sets a level, UpdateFromRuntimeConfig short-circuits on IsAgentOverriding || IsCliOverriding so a subsequent config hot-reload can't overwrite the agent's choice (covered by UpdateFromRuntimeConfig_RespectsAgentOverride). Resetting the flag at the end of UpdateFromMcp would reintroduce the bug this PR is fixing — the very next hot-reload would overwrite the agent's level.

The MCP protocol doesn't have a "release control" method either — the agent can only set a new level, not hand control back. And in --mcp-stdio mode the session lifetime is the process lifetime, so there's no natural reset point.

If you had a specific scenario in mind where reset is the desired behavior — e.g., interpreting logging/setLevel "none" as "operator regains control" instead of just "stop sending me notifications" — happy to discuss that as a follow-up since it's a design question and out of scope for this PR.

src/Service/Program.cs
anushakolan: Renamed, all of them.

src/Service.Tests/UnitTests/DynamicLogLevelProviderTests.cs
anushakolan: Good idea — added` [UpdateFromMcp_SuccessiveCalls_OverwriteAndKeepAgentOverriding]` which exercises Debug → None → Debug. It confirms each agent call overwrites the previous level (including transitions to/from LogLevel.None) and that IsAgentOverriding stays true throughout.

Worth noting: the test passes as-is because the stickiness is intentional (per the comment-6 thread). If IsAgentOverriding were reset between calls, the assertion after the second UpdateFromMcp("none") call wouldn't fail — but the test then makes the contract explicit so future readers (and any future refactor) can't quietly undo it.

RubenCerna2079 (approved):

PR #3591, opened 2026-05-15
Filter disabled tools from tools/list discovery

### LINKED ISSUE #3398: [Bug]: dml-tools boolean disable format is completely ignored
There is no working way to selectively disable individual DML tools in the MCP server. The only granularity that works is entity-level "mcp": false or "mcp": {"dml-tools": false}.

> "dml-tools": false does not disable DML tools.

## Configuration

```json
"dml-tools": {
  "delete-record": false,
  "create-record": false
}
```
### Problem

Setting a tool to false (boolean) has zero effect. All tools remain in tools/list and are callable. The dab configure CLI generates this exact format, but the server ignores it.

### Deeper problem

```json
"dml-tools": {
  "describe-entities": {"enabled": true},
  "read-records": {"enabled": true},
  "delete-record": {"enabled": false}  // ← this causes ALL tools to disappear
}
```

souvikghosh04:
## Summary

- Closes #3398

Internally, all tools respects the enabled/disabled config flag. However, discovery was still made. Disabled tools were still returned in `tools/list` MCP responses. This fix filters them out at the listing level.

## Changes

- Added `IsEnabled(RuntimeConfig)` to `IMcpTool` interface
- Implemented per-tool config checks in all 7 built-in tools + `DynamicCustomTool` (always enabled)
- Added `GetEnabledTools(RuntimeConfig)` to `McpToolRegistry`
- Updated HTTP and stdio `tools/list` handlers to use `GetEnabledTools` with fail-fast (no silent fallback to unfiltered list)
- Added unit tests for filtering logic and per-tool `IsEnabled` behavior

### Behavior

**Before:** `tools/list` returned all registered tools regardless of `dml-tools` config flags. Setting `"delete-record": false` or `"create-record": false` had zero effect — disabled tools still appeared in the tool list and were callable.

**After:** `tools/list` only returns tools where `IsEnabled(RuntimeConfig)` is true. Disabled DML tools no longer appear in the response. Both HTTP (Streamable) and stdio transports are covered.


src/Azure.DataApiBuilder.Mcp/Core/McpToolRegistry.cs
Aniruddh25: Is this function used anywhere anymore?

src/Service.Tests/Mcp/McpToolRegistryTests.cs
```diff
@@ -293,6 +294,206 @@ public void RegisterTool_WithLeadingTrailingWhitespace_DetectsDuplicate()
...
+        {
+            // Arrange
+            McpToolRegistry registry = new();
+            registry.RegisterTool(new MockMcpTool("enabled_tool", ToolType.BuiltIn, isEnabledFunc: _ => true));
+            registry.RegisterTool(new MockMcpTool("disabled_tool", ToolType.BuiltIn, isEnabledFunc: _ => false));
+
+            // Act
+            List<Tool> allTools = registry.GetAllTools().ToList();
+
+            // Assert - GetAllTools should not filter
```
Aniruddh25: If we dont need GetAllTools function, we can remove it and the test associated with it.

Aniruddh25 (approved):

Aniruddh25 (approved):

aaronburtle (approved):
LGTM!

PR #3596, opened 2026-05-15
Phase 3: Stored Procedure Parameter Substitution (`auto-embed: true`)

### LINKED ISSUE #3331: [Enh]: Embedding Phase 3: Parameter Substitution

prshri-msft:
Implements [#3331](https://github.com/Azure/data-api-builder/issues/3331). Closes #3331.

## What this PR adds

A new `auto-embed: true` flag for stored-procedure parameters. DAB acts as a **string-substitution layer**: takes plain text from the caller, calls `IEmbeddingService` for the vector, serializes it to a JSON array string (`"[0.012,-0.045,...]"`), and substitutes it as the parameter value. The sproc receives a string-compatible parameter (NVARCHAR/VARCHAR) and owns any `CAST(... AS VECTOR(N))` itself.

Single-call semantic search: client sends free text → DAB embeds → DB matches.

## Example

```jsonc
// entities: { ... }
"SearchProducts": {
  "source": {
    "type": "stored-procedure",
    "object": "dbo.SearchProducts",
    "parameters": [
      { "name": "query_text", "auto-embed": true },
      { "name": "top_k", "default": "5" }
    ]
  },
  "rest": { "enabled": true, "methods": ["post"] },
  "permissions": [{ "role": "anonymous", "actions": [{ "action": "execute" }] }]
}
```

Schema accepts `boolean-or-string`, so `"auto-embed": "@env(ENABLE_EMBED)"` is also valid.

```sql
CREATE PROCEDURE dbo.SearchProducts
    @query_text NVARCHAR(MAX),
    @top_k INT = 5
AS
BEGIN
    DECLARE @v VECTOR(1536) = CAST(@query_text AS VECTOR(1536));
    SELECT TOP(@top_k) Id, Name, VECTOR_DISTANCE('cosine', embedding, @v) AS Distance
    FROM dbo.Products
    ORDER BY Distance ASC;
END
```

```text
POST /api/SearchProducts
Content-Type: application/json

{ "query_text": "wireless headphones", "top_k": 3 }
```

## Implementation

- **Config**: `auto-embed` on `ParameterMetadata`; `boolean-or-string` in schema.
- **Validation**: 2 startup rules — stored-procedure entities only; `runtime.embeddings` must be configured. Per-MSSQL string-compatibility check rejects non-string sproc params (VECTOR, INT, …) at startup.
- **DI**: `IEmbeddingService` is non-optional. `NullEmbeddingService.Instance` registered when embeddings are not configured.
- **Substitution**: `ParameterEmbeddingHelper.SubstituteEmbedParametersAsync` runs **Collect → Batch → Substitute**. Defaults are honored; null/empty/whitespace → `string.Empty` (no API call); non-string runtime values → 400. All auto-embed params for one entity batch into a single `TryEmbedBatchAsync` call.
- **CLI**: `--parameters.auto-embed` on `dab add` / `dab update`; preserved across updates.
- **API metadata**: `auto-embed` surfaced in OpenAPI, GraphQL, and MCP `describe-entities`.
- **Telemetry**: Meter `DataApiBuilder.AutoEmbedSubstitution` (2 counters + 1 histogram) + Activity `AutoEmbed.Substitute`. Tags: `entity`, `sproc`, `param_names`, `embedding.provider`, `embedding.model`, `outcome`, `duration_ms`. **Never** includes input text or vector values (spec data-safety requirement).
- **HTTP status codes** (per spec § Error behavior):
  - `400` non-string input
  - `502` provider call failed or returned malformed response (empty vector / wrong count)
  - `503` embedding subsystem unavailable
  - `500` reserved for DAB-internal bugs
- **Vector format**: `G9` + `InvariantCulture` for round-trip precision.

## Tests

54+ new unit tests, all passing in CI:

- `ParameterEmbeddingHelperTests` (34) — full substitution helper + telemetry coverage
- `ConfigValidationUnitTests` (10) — startup rules
- `SqlMetadataProviderUnitTests` (6) — string-compatibility check
- `ConfigurationTests`, `DescribeEntitiesStoredProcedureParametersTests`, CLI tests — config round-trip + MCP metadata + CLI parsing

```bash
dotnet test src/Service.Tests --filter "FullyQualifiedName~ParameterEmbeddingHelperTests|FullyQualifiedName~ValidateAutoEmbed|FullyQualifiedName~DescribeEntitiesStoredProcedureParametersTests|FullyQualifiedName~AutoEmbedParam_Description"
dotnet test src/Cli.Tests --filter "FullyQualifiedName~AutoEmbed"
```

Manually verified end-to-end against live Azure SQL + Azure OpenAI (`text-embedding-3-small`): REST POST/GET, GraphQL, MCP execute-entity; single + multi-embed-param sprocs; positive paths + input-validation + provider-failure cases.

## Notable design decisions

- **DAB is a string substitution layer, not a type-coercion engine.** Earlier iterations rewrote sproc parameter type from VECTOR to String at metadata discovery. Per spec, the sproc declares the param as a string type and DAB never touches sproc metadata.
- **Provider-neutral.** Original PR was MSSQL-only at the validator level; per spec, no DB-type gate. Per-provider string checks live in the provider's metadata code.
- **Defaults are honored.** Original PR rejected `auto-embed` + `default`; per spec, defaults are allowed and embedded when the caller omits the param.
- **Renamed `embed` → `auto-embed`** to clarify DAB does the embedding on the caller's behalf.
- **`NullEmbeddingService` null-object** moves the "is this configured?" check into `IsEnabled`; DI misconfiguration now fails at startup instead of producing `NullReferenceException` at first request.

---

Prior review history (during [#3441](https://github.com/Azure/data-api-builder/pull/3441)): [`ajtiwari07/data-api-builder#1`](https://github.com/ajtiwari07/data-api-builder/pull/1).


schemas/dab.draft.schema.json
```diff
@@ -1245,7 +1245,8 @@
...
                               "default": { "type": ["string", "number", "boolean", "null"], "description": "Default value" },
-                              "description": { "type": "string", "description": "Parameter description. Since descriptions for multiple parameters are provided as a comma-separated string, individual parameter descriptions must not contain a comma (',')." }
+                              "description": { "type": "string", "description": "Parameter description. Since descriptions for multiple parameters are provided as a comma-separated string, individual parameter descriptions must not contain a comma (',')." },
+                              "embed": { "type": "boolean", "description": "When true, the parameter text is automatically converted to an embedding vector via the configured embedding service before being passed to the stored procedure. Requires runtime.embeddings to be configured. Only valid on stored-procedure entities. The target stored procedure parameter must be declared as VECTOR(N) — DAB cannot detect non-VECTOR misconfigurations at startup due to SQL Server metadata limitations.", "default": false }
```
JerryNixon: Per our discussion, change "embed" to "auto-embed".

JerryNixon:
## Spec mismatches

### Acceptance criteria

| Issue                            | Current PR behavior                                       | Required change                                                                                                           |
| -------------------------------- | --------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
| Property name                    | Uses `embed`                                              | Rename the property to `auto-embed`                                                                                       |
| `@env()` and `@akv()` support    | Schema is plain boolean; runtime is `bool`                | Change the schema to `boolean-or-string` and resolve `@env()` / `@akv()` at config load                                   |
| Type rule                        | Requires `VECTOR(N)` / `Byte[]`; rejects strings          | Validate that the target parameter is string-compatible, such as `nvarchar` or `varchar`; reject `VECTOR`                 |
| Provider scope                   | Hard-fails non-MSSQL at startup                           | Remove the MSSQL gate. Treat embedding as a provider-neutral string-substitution step                                     |
| Default values                   | Rejects `embed: true` plus `default` at startup           | Allow defaults and embed them when the caller omits the parameter                                                         |
| Empty, whitespace, or null input | Returns `400`                                             | Skip embedding and pass `""` to the stored procedure                                                                      |
| Missing parameter name           | Typo in config is silently ignored                        | Validate that each `auto-embed` parameter name exists on the entity and fail startup otherwise                            |
| CLI support                      | No `--parameters.auto-embed` in `dab add` or `dab update` | Add the option to both commands                                                                                           |
| Metadata                         | Not exposed in OpenAPI, GraphQL, or MCP describe          | Surface `auto-embed` in OpenAPI schema, GraphQL SDL or description, and MCP tool description                              |
| Telemetry                        | No logs or OTEL in `ParameterEmbeddingHelper`             | Add structured logs and OTEL spans or metrics around embedding calls, including latency, success or failure, and provider |
| MCP support                      | No MCP test; verification covers REST and GraphQL only    | Add an MCP integration test and document MCP verification steps                                                           |

## Behavior bugs

| Issue                  | Current PR behavior                                                             | Required change                                                                                              |
| ---------------------- | ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------ |
| Status codes           | Provider or batch failure returns `500`                                         | Map provider errors to `502`; map unavailability or timeout to `503`                                         |
| Provider error details | Drops `batchResult.ErrorMessage`                                                | Propagate provider status, provider code, sanitized message, and correlation ID                              |
| Optional DI dependency | `IEmbeddingService? = null` allows DI misconfiguration to fail on first request | Make the dependency non-optional and register a null-object `IEmbeddingService` when embeddings are disabled |
| Hot reload             | Validator reruns, but `ApplyEmbedTypeOverride` only runs at init                | Rerun `ApplyEmbedTypeOverride` on hot reload and add a hot-reload test                                       |
| Spec-conflicting tests | Tests lock in rejected defaults and `400` for null or empty values              | Update tests to assert defaults are embedded and null, empty, or whitespace values pass `""`                 |

## Nits

| Issue                                               | Required change                                                     |
| --------------------------------------------------- | ------------------------------------------------------------------- |
| Schema description leaks implementation limitation  | Move the caveat to docs and keep the schema description user-facing |
| `using ...Embeddings;` out of sort order in 4 files | Run `dotnet format --verify-no-changes` and fix using order         |
| Entity lookup duplicated at 3 call sites            | Extract a helper method and call it from the three sites            |

prshri-msft:
> ## Spec mismatches
> ### Acceptance criteria
> Issue	Current PR behavior	Required change
> Property name	Uses `embed`	Rename the property to `auto-embed`
> `@env()` and `@akv()` support	Schema is plain boolean; runtime is `bool`	Change the schema to `boolean-or-string` and resolve `@env()` / `@akv()` at config load
> Type rule	Requires `VECTOR(N)` / `Byte[]`; rejects strings	Validate that the target parameter is string-compatible, such as `nvarchar` or `varchar`; reject `VECTOR`
> Provider scope	Hard-fails non-MSSQL at startup	Remove the MSSQL gate. Treat embedding as a provider-neutral string-substitution step
> Default values	Rejects `embed: true` plus `default` at startup	Allow defaults and embed them when the caller omits the parameter
> Empty, whitespace, or null input	Returns `400`	Skip embedding and pass `""` to the stored procedure
> Missing parameter name	Typo in config is silently ignored	Validate that each `auto-embed` parameter name exists on the entity and fail startup otherwise
> CLI support	No `--parameters.auto-embed` in `dab add` or `dab update`	Add the option to both commands
> Metadata	Not exposed in OpenAPI, GraphQL, or MCP describe	Surface `auto-embed` in OpenAPI schema, GraphQL SDL or description, and MCP tool description
> Telemetry	No logs or OTEL in `ParameterEmbeddingHelper`	Add structured logs and OTEL spans or metrics around embedding calls, including latency, success or failure, and provider
> MCP support	No MCP test; verification covers REST and GraphQL only	Add an MCP integration test and document MCP verification steps
> ## Behavior bugs
> Issue	Current PR behavior	Required change
> Status codes	Provider or batch failure returns `500`	Map provider errors to `502`; map unavailability or timeout to `503`
> Provider error details	Drops `batchResult.ErrorMessage`	Propagate provider status, provider code, sanitized message, and correlation ID
> Optional DI dependency	`IEmbeddingService? = null` allows DI misconfiguration to fail on first request	Make the dependency non-optional and register a null-object `IEmbeddingService` when embeddings are disabled
> Hot reload	Validator reruns, but `ApplyEmbedTypeOverride` only runs at init	Rerun `ApplyEmbedTypeOverride` on hot reload and add a hot-reload test
> Spec-conflicting tests	Tests lock in rejected defaults and `400` for null or empty values	Update tests to assert defaults are embedded and null, empty, or whitespace values pass `""`
> ## Nits
> Issue	Required change
> Schema description leaks implementation limitation	Move the caveat to docs and keep the schema description user-facing
> `using ...Embeddings;` out of sort order in 4 files	Run `dotnet format --verify-no-changes` and fix using order
> Entity lookup duplicated at 3 call sites	Extract a helper method and call it from the three sites

Thanks for the thorough review, @JerryNixon
I had a few clarifying questions before I push the remaining changes that you requested

 The "type rule", "provider scope", "default values", and "empty/whitespace/null input" items together represent a meaningful redesign relative to what's currently in the PR.

   1. Re: type rule (NVARCHAR/VARCHAR target instead of VECTOR) — For sprocs receiving NVARCHAR-typed embed params, do you envision sproc authors writing an explicit DECLARE @vec VECTOR(N) = CAST(@param AS VECTOR(N)) in the sproc body, or do you expect SQL Server's implicit conversion in VECTOR_DISTANCE(..., @param) to handle it transparently? 
   
   2. Re: allowing defaults + embedding them — In an earlier iteration of this PR we deliberately rejected auto-embed: true + default after analyzing the UX (a missing search query usually indicates a client bug rather than intent to fall back to a canned semantic search). Curioius to learn about the design decision behind supporting default values
   
   3. Re: empty / whitespace / null input → skip embedding, pass "" — Should null be normalized to "" for the sproc, or remain distinct (bind as DBNull)? And for non-string inputs supplied where an auto-embed param is expected (e.g., a number), do you still want a 400, or also pass through as ""?
   
   4. Re: missing parameter name validation — MsSqlMetadataProvider already throws Could not find parameter 'X' specified in config for procedure 'Y' at startup when the param name doesn't match any sproc parameter (assuming the DB is reachable). Are you referring to a dab validate CLI scenario where the DB isn't reachable, or a different gap you've seen?

src/Core/Services/Embeddings/ParameterEmbeddingHelper.cs
```diff
@@ -0,0 +1,467 @@
...
+            // failed) rather than only the generic "Failed to generate embeddings" line.
+            // Per-param specificity is lost at the batch level, so naming all involved params
+            // helps identify the request context.
+            string paramNames = string.Join(", ", embedRequests.Select(r => $"'{r.paramName}'"));
+            string providerDetail = string.IsNullOrWhiteSpace(batchResult.ErrorMessage)
+                ? string.Empty
+                : $" Provider error: {batchResult.ErrorMessage}";
+            DataApiBuilderException batchEx = new(
+                message: $"Failed to generate embeddings for parameter(s) {paramNames}.{providerDetail}",
+                statusCode: HttpStatusCode.InternalServerError,
```
sayalikudale: Can we preserve batchResult.ErrorMessage here instead of hardcoding a generic 500?

src/Core/Resolvers/Factories/MutationEngineFactory.cs
```diff
@@ -29,18 +31,12 @@ public class MutationEngineFactory : IMutationEngineFactory
         private readonly IHttpContextAccessor _httpContextAccessor;
         private readonly IAuthorizationResolver _authorizationResolver;
         private readonly GQLFilterParser _gQLFilterParser;
+        private readonly IEmbeddingService _embeddingService;
+        private readonly ILogger<IMutationEngine> _logger;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="MutationEngineFactory"/> class.
         /// </summary>
```
souvikghosh04, line 39: nit- why is the header summary of params removed?

src/Core/Resolvers/SqlMutationEngine.cs
```diff
@@ -347,6 +355,17 @@ private static bool IsPointMutation(IMiddlewareContext context)
             ISqlMetadataProvider sqlMetadataProvider = _sqlMetadataProviderFactory.GetMetadataProvider(dataSourceName);
             IQueryBuilder queryBuilder = _queryManagerFactory.GetQueryBuilder(sqlMetadataProvider.GetDatabaseType());
             IQueryExecutor queryExecutor = _queryManagerFactory.GetQueryExecutor(sqlMetadataProvider.GetDatabaseType());
+
+            // Phase 3: substitute auto-embed:true parameters with embedding vectors (REST mutation path).
```
souvikghosh04: nit- I think you can remove 'Phase 3' from comments

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -140,6 +144,16 @@ await ExecuteAsync(structure, dataSourceName, isMultipleCreateOperation: true),
             ISqlMetadataProvider sqlMetadataProvider = _sqlMetadataProviderFactory.GetMetadataProvider(dataSourceName);
             if (sqlMetadataProvider.GraphQLStoredProcedureExposedNameToEntityNameMap.TryGetValue(context.Selection.Field.Name, out string? entityName))
             {
+                // Phase 3: substitute auto-embed:true parameters with embedding vectors (GraphQL path).
```
souvikghosh04: nit- remove Phase 3

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -231,6 +235,35 @@ protected override async Task FillSchemaForStoredProcedureAsync(
             GraphQLStoredProcedureExposedNameToEntityNameMap.TryAdd(GenerateStoredProcedureGraphQLFieldName(entityName, procedureEntity), entityName);
         }
 
+        /// <summary>
+        /// Per spec #3331, auto-embed parameters must be string-compatible (nvarchar, varchar,
```
souvikghosh04: nit- please remove all references of internal specs and other unrelated references

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -231,6 +235,35 @@ protected override async Task FillSchemaForStoredProcedureAsync(
...
+            }
+
+            if (parameterDefinition.SystemType != typeof(string))
+            {
+                throw new DataApiBuilderException(
+                    message: $"Procedure '{schemaName}.{storedProcedureName}': " +
+                             $"parameter '{parameterName}' has 'auto-embed: true' in config but is " +
+                             $"declared as '{parameterDefinition.SystemType.Name}' in the stored procedure. " +
+                             $"Auto-embed parameters must be string-compatible (nvarchar, varchar, etc.).",
+                    statusCode: HttpStatusCode.ServiceUnavailable,
```
souvikghosh04: auto embed validation error should not be classified as ServiceUnavailable. BadRequest is more appropriate

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
```diff
@@ -1415,10 +1437,15 @@ private static OpenApiSchema CreateSpRequestComponentSchema(Dictionary<string, P
                 ParameterDefinition def = kvp.Value;
                 string typeMetadata = TypeHelper.GetJsonDataTypeFromSystemType(def.SystemType).ToString().ToLower();
 
+                bool isAutoEmbed = configParams?.FirstOrDefault(p => p.Name == parameter)?.AutoEmbed ?? false;
+                string? description = isAutoEmbed
+                    ? (def.Description ?? string.Empty) + " (auto-embed: DAB converts this value to an embedding before execution)"
```
souvikghosh04: this seems to be hard-coded and duplicated in multiple places. for e.g.g in `AddStoredProcedureInputParameters()`. please refactor those.

src/Service/Startup.cs
```diff
@@ -477,9 +485,14 @@ public void ConfigureServices(IServiceCollection services)
...
                 else
                 {
+                    services.AddSingleton<IEmbeddingService>(NullEmbeddingService.Instance);
                     _logger.LogInformation("Embeddings service is configured but disabled.");
                 }
             }
+            else
+            {
+                services.AddSingleton<IEmbeddingService>(NullEmbeddingService.Instance);
+            }
```
souvikghosh04: do we really need nested checks and logs for embedding configured and enabled/disabled? update the logs and messages accordingly if needed

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
```diff
@@ -67,13 +67,20 @@ public static FieldDefinitionNode GenerateStoredProcedureSchema(
                         }
                     }
 
+                    // Build description, appending auto-embed indicator when configured.
+                    string baseDescription = definition.Description
+                        ?? $"parameters for {name.Value} stored-procedure";
+                    bool isAutoEmbed = entity.Source.Parameters
+                        ?.FirstOrDefault(p => p.Name == param)?.AutoEmbed ?? false;
+                    string fullDescription = isAutoEmbed
+                        ? $"{baseDescription} (auto-embed: DAB converts this value to an embedding before execution)"
```
souvikghosh04: nit- same here, its repeated

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -231,6 +235,35 @@ protected override async Task FillSchemaForStoredProcedureAsync(
             GraphQLStoredProcedureExposedNameToEntityNameMap.TryAdd(GenerateStoredProcedureGraphQLFieldName(entityName, procedureEntity), entityName);
         }
 
+        /// <summary>
+        /// Per spec #3331, auto-embed parameters must be string-compatible (nvarchar, varchar,
+        /// etc.) so DAB can pass the embedding JSON as a plain string. The stored procedure
+        /// is responsible for any CAST to VECTOR. Rejects non-string types at startup.
+        /// </summary>
+        internal static void ValidateAutoEmbedStringCompatibility(
```
souvikghosh04, line 259: since the embedding service touches the startup and other providers, like, PG, MSSQL etc. might fail. do we have validations at startup where it fails/warns if the provider is not MSSQL?

src/Cli.Tests/Snapshots/UpdateEntityTests.TestConversionOfSourceObject_a70c086a74142c82.verified.txt
```diff
@@ -1,4 +1,4 @@
-{
+﻿{
```
souvikghosh04: it shows a diff in line 1 but I don't see a change. this could be due to change in format. please make sure the .txt files (this and other snapshot files) are encoded in UTF-8

souvikghosh04 (changes requested):
mostly refactorings, minor suggestions and nits. also, check the startup validation

src/Core/Services/Embeddings/ParameterEmbeddingHelper.cs
```diff
@@ -0,0 +1,481 @@
...
+    /// <param name="logger">Optional logger for structured per-operation diagnostics. Engines pass their own ILogger; tests may omit.</param>
+    public static Task SubstituteEmbedParametersAsync(
+        IDictionary<string, object?> resolvedParams,
+        RuntimeConfig runtimeConfig,
+        string entityName,
+        IEmbeddingService embeddingService,
+        CancellationToken cancellationToken,
+        ILogger? logger = null)
+    {
+        Entity entity = runtimeConfig.Entities[entityName];
```
ajtiwari07: We should handle absent keys in lookup.

src/Core/Resolvers/Factories/MutationEngineFactory.cs
prshri-msft: Addressed in Stage 3.24: the constructor XML documentation now includes every parameter, including `embeddingService` and `logger`.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
prshri-msft: Addressed in Stage 3.30: startup validation resolves each entity's data source and rejects auto-embed unless the database type is MSSQL or DWSQL. Config validation tests cover the unsupported-provider path.

src/Core/Services/Embeddings/ParameterEmbeddingHelper.cs
prshri-msft: Addressed in 3e354869: the entity lookup now uses `TryGetValue`; a missing entity returns the canonical 404 `EntityNotFound` response. Added `MissingEntity_ThrowsEntityNotFound` to cover the contract.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
prshri-msft: Correction: this thread was addressed in Stage 3.24 by removing internal spec and unrelated references from production code and comments.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
prshri-msft: Addressed in Stage 3.30: startup validation resolves each entity's data source and rejects auto-embed unless the database type is MSSQL or DWSQL. Config validation tests cover the unsupported-provider path.

src/Config/ObjectModel/ParameterMetadata.cs
```diff
@@ -24,5 +24,19 @@ public class ParameterMetadata
...
+        /// When true, the parameter value (text) is automatically embedded via the
+        /// EmbeddingService and the resulting vector is passed to the stored procedure.
+        /// Only valid on stored-procedure entities when runtime.embeddings is configured.
+        ///
+        /// IMPORTANT: The target stored procedure parameter must be declared as VECTOR(N).
+        /// SQL Server's metadata system reports VECTOR(N) and varbinary indistinguishably,
+        /// so DAB cannot detect this misconfiguration at startup. If embed:true is applied
+        /// to a non-VECTOR parameter (e.g., NVARCHAR or VARBINARY), the request will fail
+        /// at runtime with a SQL error or return semantically incorrect results.
+        /// It is the developer's responsibility to ensure the sproc parameter is VECTOR(N).
```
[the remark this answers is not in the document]
prshri-msft: Addressed in Stage 3.23. The XML documentation now requires a string-compatible parameter and explains that type validation is performed by the database metadata provider; the obsolete “cannot detect” wording was removed.

schemas/dab.draft.schema.json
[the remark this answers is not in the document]
prshri-msft: Addressed in Stage 3.23. The schema now states the string-compatible parameter requirement and stored-procedure CAST responsibility; the inaccurate “cannot detect non-VECTOR” wording was removed.

src/Core/Azure.DataApiBuilder.Core.csproj
```diff
@@ -58,6 +58,10 @@
     <ProjectReference Include="..\Service.GraphQLBuilder\Azure.DataApiBuilder.Service.GraphQLBuilder.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Azure.DataApiBuilder.Service.Tests" />
+  </ItemGroup>
+
   <ItemGroup>
```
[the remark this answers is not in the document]
prshri-msft: Addressed. The duplicate project-level friend-assembly declaration was removed; the single assembly attribute remains in `SqlMetadataProvider.cs`.

schemas/dab.draft.schema.json
prshri-msft: Addressed in Stage 3.11. The configuration property and schema now use `auto-embed`.

src/Core/Resolvers/SqlMutationEngine.cs
prshri-msft: Addressed in Stage 3.24. Phase/spec references were removed from production comments.

src/Core/Resolvers/SqlQueryEngine.cs
prshri-msft: Addressed in Stage 3.24. Phase/spec references were removed from production comments.

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
prshri-msft: Addressed in Stage 3.25. Auto-embed validation failures now return `BadRequest` rather than `ServiceUnavailable`.

src/Core/Services/OpenAPI/OpenApiDocumentor.cs
prshri-msft: Addressed in Stage 3.26. The indicator is centralized in `AutoEmbedDescription` and reused across metadata surfaces.

src/Service/Startup.cs
prshri-msft: Addressed in Stage 3.28. Embedding-service registration now uses one flattened condition and consolidated logging.

src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
prshri-msft: Addressed in Stage 3.26. The indicator is centralized in `AutoEmbedDescription` and reused across metadata surfaces.

src/Cli.Tests/Snapshots/UpdateEntityTests.TestConversionOfSourceObject_a70c086a74142c82.verified.txt
prshri-msft: Addressed in Stage 3.27. The BOM was removed; the snapshot is UTF-8 without BOM and now begins with `{`.

src/Core/Configurations/RuntimeConfigValidator.cs
```diff
@@ -421,7 +422,154 @@ public void ValidateEmbeddingsOptions(RuntimeConfig runtimeConfig)
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ConfigValidationError));
             }
         }
+    }
+
+    /// <summary>
+    /// Validates that parameters with embed=true are only used on stored-procedure entities
+    /// and that runtime.embeddings is configured when embed parameters are present.
```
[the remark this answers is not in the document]
prshri-msft: Updated in a27220d8. The XML summary now matches current behavior: auto-embed is limited to stored-procedure entities on MSSQL/DWSQL, requires enabled embedding configuration, and permits parameter defaults.

src/Core/Services/Embeddings/ParameterEmbeddingHelper.cs
```diff
@@ -0,0 +1,279 @@
...
+        EmbeddingBatchResult batchResult = await embeddingService.TryEmbedBatchAsync(texts, cancellationToken);
+
+        if (!batchResult.Success || batchResult.Embeddings is null)
+        {
+            // Batch failure: we lose per-param error specificity here, but the
+            // batch result's ErrorMessage typically explains the underlying issue.
+            // Naming all involved params helps the user identify the request context.
+            string paramNames = string.Join(", ", embedRequests.Select(r => $"'{r.paramName}'"));
+            throw new DataApiBuilderException(
+                message: $"Failed to generate embeddings for parameter(s) {paramNames}.",
```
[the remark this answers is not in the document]
prshri-msft: Addressed in Stage 3.18. Batch failures preserve `batchResult.ErrorMessage` when present, append it to the 502 `BadGateway` response, and retain the provider detail in substitution telemetry and logging.

src/Core/Services/Embeddings/ParameterEmbeddingHelper.cs
prshri-msft: Addressed in Stage 3.18. Batch failures preserve `batchResult.ErrorMessage` when present, append it to the 502 `BadGateway` response, and retain the provider detail in substitution telemetry and logging.

PR #3600, opened 2026-05-19
ExecuteEntityTool: DB metadata based parameter support

### LINKED ISSUE #3588: Parameter behaviour in MCP tool: ExecuteEntityTool
- follow up to https://github.com/Azure/data-api-builder/pull/3425

Validate params against StoredProcedureDefinition.Parameters (DB metadata) instead of config. Small change, closes the split-brain with describe_entities.

souvikghosh04:
## Why make this change?

- Closes #3588

`ExecuteEntityTool` validated parameters against config-side `Source.Parameters` only, rejecting DB-discovered parameters that had no config entry. Defaults were applied in an all-or-nothing fashion (only when the user supplied zero parameters).

## What is this change?

- Parameter validation now uses `StoredProcedureDefinition.Parameters` (the merged DB metadata) as source of truth, accepting any parameter the DB declares.
- Default application iterates `ParameterDefinition.HasConfigDefault`/`ConfigDefaultValue` per-missing-parameter, so partial user input gets remaining defaults filled in. This is an intentional behavioral change from the previous all-or-nothing guard.
- Safe-cast `dbObject` to `DatabaseStoredProcedure` with graceful error instead of unconditional cast.

## How was this tested?

- [x] Unit Tests

| Test | Scenario |
|------|----------|
| `AcceptsDbDiscoveredParam` | Param in DB metadata but not config - accepted |
| `RejectsInvalidParamName` (x2) | Param not in DB metadata - error |
| `AcceptsMultipleValidParams` | Multiple valid params pass validation |
| `RejectsRequest_WhenAnyParamInvalid` | Mix of valid + invalid params - rejected |
| `AppliesConfigDefaults` | Missing params filled from `HasConfigDefault`/`ConfigDefaultValue` |
| `UserParamsOverride` | User-supplied values take precedence over config defaults |
| `DoesNotInjectWithoutDefault` | Param without `HasConfigDefault` is not injected |
| `ZeroParamSP` | SP with no parameters succeeds with no input |
| `RejectsNonStoredProcedureEntity` | Non-SP entity types return error |
| `ReturnsError_WhenEntityNotFound` | Non-existent entity returns EntityNotFound |

- [x] Integration Tests

| Test | Scenario |
|------|----------|
| Success (x4 DataRows) | GetBook(id), InsertBook(null), InsertBook(override), GetBooks(zero-param) |
| `GetBookById_ReturnsRecord` | Validates returned data matches expected |
| `InsertBookWithDefaults_ExecutesSuccessfully` | Config defaults produce valid insert |
| Rejection (x2 DataRows) | Invalid param names rejected with clear error |

- [x] Manual Testing

| Category | Result |
|----------|--------|
| Valid DB-discovered params | Pass |
| Invalid/unknown/extra params rejected | Pass |
| Config defaults / partial params / override | Pass |
| Non-SP entity type gating | Pass |
| Successful SP execution (zero/single/multi-param) | Pass |
| Non-existent entity | Pass |
| SQL injection (parameterized) | Pass |
| Type mismatch | Pass |
| Permission denied | Pass |
| No results for valid param | Pass |

## Sample Request(s)

NA


src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -162,13 +162,20 @@ public async Task<CallToolResult> ExecuteAsync(
...
+                }
+
+                StoredProcedureDefinition spDefinition = storedProcedure.StoredProcedureDefinition;
+                if (parameters != null)
                 {
-                    // Validate all provided parameters exist in metadata
                     foreach (KeyValuePair<string, object?> param in parameters)
                     {
-                        if (!entityConfig.Source.Parameters.Any(p => p.Name == param.Key))
+                        if (!spDefinition.Parameters.ContainsKey(param.Key))
```
anushakolan, line 182: SQL Server parameter names are case-insensitive.
Will this case-insensitive check cause any bugs?

Aniruddh25: case sensitivity depends on DB collation. Are we checking if DB collation is case sensitive or not?

src/Service.Tests/Mcp/ExecuteEntityToolTests.cs
```diff
@@ -0,0 +1,447 @@
...
+        [DataRow("nonexistent", DisplayName = "Completely unknown param")]
+        [DataRow("ID", DisplayName = "Case-sensitive mismatch")]
+        public async Task ExecuteEntity_RejectsInvalidParamName(string invalidParamName)
+        {
+            Dictionary<string, ParameterDefinition> dbParams = new()
+            {
+                ["id"] = new()
+            };
+
+            CallToolResult result = await ExecuteWithMockedEngineAsync(
```
Aniruddh25, line 83: Do these unit tests cover additional scenarios than the integration tests?

Aniruddh25 (approved):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: Added a code comment clarifying this intentional behavior. We also have a unit test ([DataRow("ID", DisplayName = "Case-sensitive mismatch")]) validating it.

src/Service.Tests/Mcp/ExecuteEntityToolTests.cs
souvikghosh04: Yes unit tests cover scenarios that integration tests can't easily test:

Case-sensitive mismatch ("ID" vs "id") — no SP in the DB schema has this exact setup
Multi-param with one invalid ("id" valid + "bogus" invalid) — tests first-invalid-rejection logic
Non-SP entity rejection — exercises the dbObject is not DatabaseStoredProcedure guard (Table entity)
Context capture — unit tests inspect the exact ResolvedParameters dictionary passed to the query engine (integration tests can only observe JSON output)
Speed — run in <1s without DB; CI-safe without MsSql filter

RubenCerna2079 (approved):
LGTM

PR #3601, opened 2026-05-19
DynamicCustomTool: DB metadata based parameter support

### LINKED ISSUE #3589: Parameter behaviour in MCP tool: DynamicCustomTool
- follow up to https://github.com/Azure/data-api-builder/pull/3425

Thread IServiceProvider through CustomMcpToolFactory, rewrite BuildInputSchema() to use DB metadata, add proper required array and type mapping.

souvikghosh04:
## Why make this change?

- Closes #3589

DynamicCustomTool (per-entity custom MCP tools) used `entityConfig.Source.Parameters` / `ParameterMetadata` for parameter validation and defaults. This meant:
- Parameters discovered only via DB metadata (not listed in config) were rejected.
- Defaults were applied all-or-nothing instead of per-missing-parameter.

## What is this change?

- **Validate against DB metadata**: After authorization, safe-cast `dbObject` to `DatabaseStoredProcedure` and validate user-supplied params against `spDefinition.Parameters` (source of truth populated by `FillSchemaForStoredProcedureAsync`).
- **Per-param config defaults**: Iterate `spDefinition.Parameters`; for each missing param with `HasConfigDefault == true`, inject `ConfigDefaultValue`. User-supplied params always take precedence.
- **Removed old logic**: Replaced `entityConfig.Source.Parameters` / `ParameterMetadata` loop.

## How was this tested?

- [x] Unit Tests

| Test | Scenario |
|------|----------|
| `AcceptsDbDiscoveredParam` | Param in DB metadata but not config - accepted |
| `RejectsInvalidParamName` (x2) | Param not in DB metadata - error |
| `AppliesConfigDefaults` | Missing params filled from `HasConfigDefault`/`ConfigDefaultValue` |
| `UserParamsOverride` | User-supplied values take precedence over config defaults |
| `DoesNotInjectWithoutDefault` | Param without `HasConfigDefault` is not injected |
| `ZeroParamSP` | SP with no parameters succeeds with no input |

- [x] Integration Tests

| Test | Scenario |
|------|----------|
| Success (x4 DataRows) | GetBook(id), InsertBook(null), InsertBook(override), GetBooks(zero-param) |
| `GetBookById_ReturnsRecord` | Validates returned data matches expected |
| Rejection (x2 DataRows) | Invalid param names rejected with clear error |

- [x] Manual Testing

| Category | Result |
|----------|--------|
| Valid params / no results | Pass |
| Type mismatch - error | Pass |
| Zero-param SPs | Pass |
| Config defaults / partial params | Pass |
| Multi-param update | Pass |
| SQL injection (parameterized) | Pass |
| Boundary values (-1, 0, MAX_INT) | Pass |
| Empty/long/unicode strings | Pass |
| Read-after-write, FK errors, cross-entity | Pass |

## Sample Request(s)

NA


src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -201,15 +208,14 @@ public async Task<CallToolResult> ExecuteAsync(
                     }
                 }
 
-                // Then, add default parameters from configuration (only if not already provided by user)
-                if ((parameters == null || parameters.Count == 0) && entityConfig.Source.Parameters != null)
+                // Apply config-declared defaults from the merged ParameterDefinitions.
+                // This covers all parameters (including DB-discovered ones with config defaults)
+                // and applies them when the user didn't supply a value.
+                foreach ((string paramName, ParameterDefinition paramDef) in spDefinition.Parameters)
```
Aniruddh25: can spDefinition.Parameters ever be null? if yes, we need to handle the possible NRE here

src/Service.Tests/Mcp/DynamicCustomToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,210 @@
...
+            JsonElement root = doc.RootElement;
+            Assert.AreEqual(entityName, root.GetProperty("entity").GetString());
+            Assert.AreEqual("Stored procedure executed successfully", root.GetProperty("message").GetString());
+        }
+
+        /// <summary>
+        /// Verify GetBook with id=1 returns a matching record through DynamicCustomTool.
+        /// </summary>
+        [TestMethod]
+        public async Task DynamicCustomTool_GetBookById_ReturnsMatchingRecord()
```
Aniruddh25, line 86: How is this test different than DynamicCustomTool_SuccessfulExecution data row 1?

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
```diff
@@ -203,15 +210,14 @@ public async Task<CallToolResult> ExecuteAsync(
...
                 {
-                    foreach (ParameterMetadata param in entityConfig.Source.Parameters)
+                    if (!context.FieldValuePairsInBody.ContainsKey(paramName) && paramDef.HasConfigDefault)
```
Aniruddh25: is it going to be case sensitive comparison or insensitive for the param name?

it depends on the database collation whether to treat parameters case sensitive or insensitive. What is our specification when we merged the stored proc + config parameters?

src/Service.Tests/Mcp/DynamicCustomToolTests.cs
```diff
@@ -200,6 +217,293 @@ public void GetToolMetadata_UsesDefaultParameterDescription_WhenNotProvided()
...
+            Assert.AreEqual("default_tenant", capturedContext.ResolvedParameters["tenant"]);
+        }
+
+        /// <summary>
+        /// Zero-param SP with no user params passes empty parameters.
+        /// </summary>
+        [TestMethod]
+        public async Task ExecuteAsync_ZeroParamSP_PassesEmptyParams()
+        {
+            Dictionary<string, ParameterDefinition> dbParams = new();
```
Aniruddh25, line 342: If we have the Integration tests in DynamicCustomToolMssqlIntegrationTests, what additional coverage do these unit tests provide?

src/Service.Tests/Mcp/ExecuteEntityToolTests.cs
```diff
@@ -0,0 +1,447 @@
...
+        private const string SP_SOURCE_OBJECT = "get_book";
+
+        #region Parameter Validation Tests
+
+        /// <summary>
+        /// A parameter that exists in DB metadata (StoredProcedureDefinition.Parameters)
+        /// is accepted even if it has no config-side entry.
+        /// </summary>
+        [TestMethod]
+        public async Task ExecuteEntity_AcceptsDbDiscoveredParam_NotInConfig()
```
Aniruddh25, line 54: Why do we have ExecuteEntityTool tests in this PR?

Aniruddh25 (approved):
Approving with suggestions on fixing the tests, and few questions.

anushakolan:
The snapshot file needs to be updated; I am approving this PR to prevent any blockers from my side.

anushakolan (approved):

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: yes, StoredProcedureDefinition.Parameters has a public setter and is initialized as = new(), but could theoretically be null if reassigned. Added is not null guards around both the validation and defaults loops in both tools.

src/Azure.DataApiBuilder.Mcp/BuiltInTools/ExecuteEntityTool.cs
souvikghosh04: yes, case-sensitive. this is consistent with the existing REST/GraphQL stored procedure execution path, which uses the same StoredProcedureDefinition.Parameters dictionary populated by FillSchemaForStoredProcedureAsync. The parameter keys are stored exactly as the DB metadata returns them. Added a code comment clarifying this is intentional. We also have a unit test ([DataRow("ID", DisplayName = "Case-sensitive mismatch")]) that explicitly validates this behavior.

src/Service.Tests/Mcp/DynamicCustomToolMsSqlIntegrationTests.cs
souvikghosh04: Data row 1 ("GetBook", "{\"id\": 1}") validates the response structure only (entity name, message string). GetBookById additionally validates the actual returned data content, it parses into the value array and asserts records[0].id == 1. This tests that the parameter value is correctly passed through to the stored procedure and the right record comes back.

src/Service.Tests/Mcp/DynamicCustomToolTests.cs
souvikghosh04: certain isolated parts are tested including some edge cases

src/Service.Tests/Mcp/ExecuteEntityToolTests.cs
souvikghosh04: This branch was created on top of the ExecuteEntityTool fix (PR #3600) since DynamicCustomTool follows the same pattern. Because #3600 hasn't merged to main yet, GitHub shows the ExecuteEntityTool files in this diff as well.

Once #3600 merges, I'll rebase this branch onto main and those files will drop out of the diff automatically, avoiding a force-push or rewrite.

souvikghosh04:
waiting for https://github.com/Azure/data-api-builder/pull/3600 to merge as this was branched out from that with related fixes.

src/Azure.DataApiBuilder.Mcp/Core/DynamicCustomTool.cs
```diff
@@ -170,7 +170,28 @@ public async Task<CallToolResult> ExecuteAsync(
...
-                // 6) Build request payload
+                // 6) Validate parameters against DB metadata (StoredProcedureDefinition.Parameters),
+                // which is the source of truth for parameter names.
+                // Note: Comparison is case-sensitive (default Dictionary<string,...> comparer),
+                // consistent with the existing REST/GraphQL SP execution path.
+                if (dbObject is not DatabaseStoredProcedure storedProcedure)
+                {
+                    return McpResponseBuilder.BuildErrorResult(toolName, "InvalidEntity", $"Entity '{EntityName}' is not a stored procedure.", logger);
+                }
+
```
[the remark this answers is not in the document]
souvikghosh04: this is a known design limitation and not a bug to fix in this PR:

BuildInputSchema() runs at tool registration time using config-side ParameterMetadata (what's available at startup)
Runtime validation uses DB metadata (populated later by FillSchemaForStoredProcedureAsync)
The schema advertised to MCP clients is a subset — it shows configured parameters. DB-discovered params without config entries are still accepted at runtime, which is the more permissive direction (clients won't be blocked by a strict schema)
Fixing this properly requires lazy schema generation or a post-initialization schema refresh, which is a separate feature.
created https://github.com/Azure/data-api-builder/issues/3615 to track this.

Aniruddh25 (approved):

aaronburtle (approved):

PR #3617, opened 2026-05-20
Fix notice-generation script

### LINKED ISSUE #3607: Bug: Fix notice-generation script
Currently the notice-generation script is not replacing the `Microsoft 2023` sections as expected since the logic currently replaces it but doesn't save those changes. We want to fix this to ensure the notice files are generated as we expected.

RubenCerna2079:
## Why make this change?

- Solves issue #3607
Currently the notice file script has a bug that causes it to not replace some of the strings correctly as the values in the `content` variable that are replaced are overwritten by other changes, so they are not saved.

## What is this change?

Changed the way the values are changed inside the `content` value by using regex. Currently the only values that need to be deleted are for the years 2023 and 2024. This might change in the future.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests
- [x] Manual Testing
Tested this manually by running the script locally.

## Sample Request(s)

N/A


aaronburtle (approved):
Thanks for fixing this, looks good!

Aniruddh25 (approved):

PR #3646, opened 2026-06-02
Mark required stored-procedure parameters as NON_NULL in GraphQL

### LINKED ISSUE #3501: [Bug]: Required params not marked in GraphQL
## Expected

Required params typed as `String! (NON_NULL)`

## Actual

All params always nullable `String`

anushakolan:
## Why make this change?

Closes #3501.

The GraphQL schema generated for stored-procedure entities emitted every input argument as nullable (`String`), so introspection couldn't distinguish required parameters from optional ones. Clients calling those mutations had no way to know which arguments were mandatory until SQL Server failed the call at execution time.

## What is this change?

In `GenerateStoredProcedureSchema`, each parameter's GraphQL type is now wrapped in `NonNullTypeNode` when the parameter is required, otherwise it stays a nullable `NamedTypeNode`. A parameter is treated as required unless the runtime config explicitly sets `required: false` for it. Names, value types, and the existing `defaultValue` handling are unchanged.

A required parameter with a config-supplied default still emits `NON_NULL` and keeps the default — that's valid GraphQL and matches client expectations.

The default-to-required behavior is intentional: T-SQL nullability (`is_nullable`) describes whether the parameter's type accepts NULL, not whether the caller must supply it. Defaulting to required produces a schema that's correct at execute time; users can opt out per-parameter via config when a procedure declares true defaults.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests
- [x] Manual end-to-end test against a live SQL Server

### Unit tests

Added to `StoredProcedureBuilderTests`:

- `StoredProcedure_RequiredFlag_ProducesNonNullType` (3 data rows): `required: true` → `NonNullTypeNode`, `required: false` → nullable `NamedTypeNode`, parameter omitted from config → defaults to `NonNullTypeNode`.
- `StoredProcedure_RequiredWithDefault_KeepsDefaultValue`: required parameter with a config default emits `NonNullTypeNode` and preserves the default value.

### Manual end-to-end test

Three GraphQL mutations were configured against the existing `dbo.insert_book(@title varchar(max), @publisher_id int)` stored procedure. Each entity exercised a different config shape, and `__schema` introspection confirmed the input argument kinds:

| GraphQL field | Config shape | `title` | `publisher_id` |
| --- | --- | --- | --- |
| `executeInsertBookPartialConfig` | only `title` listed (`required: true`) | NON_NULL ✅ | NON_NULL ✅ (DB-only → defaults to required) |
| `executeInsertBookExplicitNotRequired` | `title` `required: false`, `publisher_id` `required: true` | nullable ✅ | NON_NULL ✅ |
| `executeInsertBookNoConfigParams` | no `parameters` block | NON_NULL ✅ | NON_NULL ✅ (silent → defaults to required) |

All scenarios produced the expected GraphQL types.

## Sample request

GraphQL introspection used to validate the schema:

```graphql
{
  __schema {
    mutationType {
      fields {
        name
        args {
          name
          type { kind name ofType { kind name } }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
+++ b/src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
@@ -866,9 +866,7 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.GraphQLQueryTests
         /// <summary>
         /// Checks failure on providing arguments with no default in runtimeconfig.
         /// In this test, there is no default value for the argument 'id' in runtimeconfig, nor is it specified in the query.
-        /// Stored procedure expects id argument to be provided.
-        /// This test validates the "Development Mode" error message which denotes the
-        /// specific missing parameter and stored procedure name.
+        /// GraphQL validation should reject the request because required argument 'id' is missing.
         /// </summary>
         [TestMethod]
         public async Task TestStoredProcedureQueryWithNoDefaultInConfig()
@@ -883,7 +881,7 @@ namespace Azure.DataApiBuilder.Service.Tests.SqlTests.GraphQLQueryTests
             JsonElement result = await ExecuteGraphQLRequestAsync(graphQLQuery, graphQLQueryName, isAuthenticated: false);
             SqlTestHelper.TestForErrorInGraphQLResponse(
                 response: result.ToString(),
-                message: "Procedure or function 'get_publisher_by_id' expects parameter '@id', which was not supplied.");
+                message: "The argument `id` is required.");
         }
 
         /// <summary>
--- a/src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
+++ b/src/Service.GraphQLBuilder/GraphQLStoredProcedureBuilder.cs
@@ -25,6 +25,12 @@ namespace Azure.DataApiBuilder.Service.GraphQLBuilder
         /// It uses the parameters to build the arguments and returns a list
         /// of the StoredProcedure GraphQL object.
         /// </summary>
+        /// <remarks>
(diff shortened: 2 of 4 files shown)


Aniruddh25 (approved):

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
```diff
@@ -866,9 +866,7 @@ public async Task TestStoredProcedureQueryWithResultsContainingNull()
         /// <summary>
         /// Checks failure on providing arguments with no default in runtimeconfig.
         /// In this test, there is no default value for the argument 'id' in runtimeconfig, nor is it specified in the query.
-        /// Stored procedure expects id argument to be provided.
-        /// This test validates the "Development Mode" error message which denotes the
-        /// specific missing parameter and stored procedure name.
+        /// GraphQL validation should reject the request because required argument 'id' is missing.
         /// </summary>
         [TestMethod]
         public async Task TestStoredProcedureQueryWithNoDefaultInConfig()
```
souvikghosh04, line 872: do we need to add tests for PostgresSQL and MySQL too?

souvikghosh04 (approved):
LGTM. please check if test coverage needed for PostgreSQL and MySQL

src/Service.Tests/SqlTests/GraphQLQueryTests/DwSqlGraphQLQueryTests.cs
anushakolan: Good question. For this PR I only updated MSSQL and DW SQL because those are the suites with the `executeGetPublisher` stored-procedure fixture and the ones failing in CI. PostgreSQL/MySQL don’t currently have an equivalent SP test fixture for this case, so there wasn’t a direct test to update here.

PR #3647, opened 2026-06-02
Handle stdin EOF gracefully in MCP stdio mode (#3428)

### LINKED ISSUE #3428: [Bug] --mcp-stdio mode crashes (exit 4294967295) when stdin receives EOF
## Summary

When DAB is running in `--mcp-stdio` mode and stdin is closed (EOF), DAB exits with code **4294967295** (0xFFFFFFFF / -1) instead of shutting down gracefully. This causes MCP health checks to always report the server as disconnected, because health-check tools (e.g. `claude mcp list`) intentionally close stdin after the `initialize` handshake.

## Environment

- **DAB version:** 1.7.92
- **OS:** Windows 11 Enterprise
- **Transport:** `--mcp-stdio`
- **MCP client:** Claude Code 2.1.98

## Steps to Reproduce

```bash
# Send initialize, then immediately close stdin
echo '{"jsonrpc":"2.0","method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"test","version":"1"}},"id":1}' \
  | dab start --config dab-config.json --mcp-stdio
echo "Exit code: $?"
```

After DAB finishes startup and sends the `initialize` response, stdin reaches EOF (because `echo` exits). DAB exits with code **4294967295** instead of 0.

## Expected Behavior

DAB should handle stdin EOF gracefully:
1. Stop accepting new requests.
2. Finish any in-progress requests.
3. Exit with code **0**.

## Actual Behavior

DAB exits with code **4294967295** (unhandled exception / abnormal termination) when stdin closes.

## Impact

MCP health-check tools close stdin after receiving the `initialize` response (standard probe behavior). DAB's crash on EOF causes every health check to report `✗ Failed to connect`, even when the server is fully functional and responds correctly to the `initialize` request.

In a real session (where stdin stays open), this crash does not occur — but it breaks all automated connectivity checks and monitoring.

## Suggested Fix

In `--mcp-stdio` mode, add a graceful stdin-EOF handler: catch the `ObjectDisposedException` / `IOException` on the stdin read loop and exit cleanly with code 0 rather than propagating an unhandled exception.

anushakolan:
## Why make this change?

When `dab start --mcp-stdio` was launched by a client (such as MCP Inspector or VS Code) and the client closed the stdin pipe (EOF), the process crashed with exit code `4294967295` instead of exiting cleanly. This caused noisy error signals in any tool or host that expected a clean `0` exit on shutdown.

## What is this change?

Modified `McpStdioServer.RunAsync` to handle stdin EOF as a normal shutdown signal:

- Replaced direct `Console.OpenStandardInput()` + `StreamReader` with `Console.In` (testable, injectable)
- Added explicit `null` check on the return value of `ReadLineAsync` — `null` means EOF, which now triggers a clean `return` instead of an unhandled exception
- Blank lines continue to be skipped as before

### Unit Tests
Added `McpStdioServerRunAsyncTests`:
- `RunAsync_EofOnStdin_ExitsGracefullyWithoutOutput` — feeds empty stdin, verifies clean exit with no stdout output
- `RunAsync_BlankLineThenShutdown_IgnoresBlankLineAndHandlesShutdown` — feeds blank line + shutdown request, verifies single valid response and clean exit

### Manual Testing
- Reproduced the exact reported scenario: piped an `initialize` request then closed stdin — verified `EXIT_CODE=0`
  ```
  echo '{"jsonrpc":"2.0","method":"initialize",...,"id":1}' | dab start --config dab-config.json --mcp-stdio
  echo "Exit code: $?"   # was 4294967295, now 0
  ```
- Verified `EXIT_CODE=0` for normal `shutdown` request path
- Verified successful MCP tool calls (`read_records`, `describe_entities`) continue to work after the change


src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -61,17 +61,20 @@ public McpStdioServer(McpToolRegistry toolRegistry, IServiceProvider serviceProv
...
         public async Task RunAsync(CancellationToken cancellationToken)
         {
-            // Use UTF-8 WITHOUT BOM for stdin. Stdout is owned by McpStdoutWriter,
-            // which serializes all writes from McpStdioServer and the MCP logging
-            // pipeline so JSON-RPC frames cannot interleave at the byte level.
-            UTF8Encoding utf8NoBom = new(encoderShouldEmitUTF8Identifier: false);
-
-            using Stream stdin = Console.OpenStandardInput();
-            using StreamReader reader = new(stdin, utf8NoBom);
+            // Read through Console.In so tests can inject stdin and the process
```
aaronburtle: nit: leaks test concerns into production code. from a production standpoint, the rationale here I think is that we want to honor the configured Console.In ie:

// Read via Console.In so the loop honors the configured Console.InputEncoding
// (set to UTF-8 by Program.Main for stdio mode) and shares the SyncTextReader.

src/Service.Tests/UnitTests/McpStdioServerRunAsyncTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        [TestMethod]
+        public async Task RunAsync_EofOnStdin_ExitsGracefullyWithoutOutput()
+        {
+            (McpStdioServer server, StringWriter stdoutCapture) = CreateServerWithCapturedOutput();
+            TextReader originalIn = Console.In;
+
+            try
+            {
+                // Empty input immediately yields EOF (ReadLineAsync returns null).
+                Console.SetIn(new StringReader(string.Empty));
```
aaronburtle: nit: Console.SetIn is process-global, if we ever parallelize this test it could break other tests that read Console.In. Can instead extract the read source, and then tests can pass a StringReader directly.

src/Service.Tests/UnitTests/McpStdioServerRunAsyncTests.cs
```diff
@@ -0,0 +1,87 @@
...
+
+                await server.RunAsync(CancellationToken.None);
+
+                string[] lines = stdoutCapture
+                    .ToString()
+                    .Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries);
+
+                Assert.AreEqual(1, lines.Length,
+                    "Expected a single response line for shutdown request.");
+                StringAssert.Contains(lines[0], "\"id\":1");
```
aaronburtle: nit: parsing this line as JSON and then asserting on jsonrpc/id/result.ok would align more closely with the other mcp tests while being a bit more robust.

aaronburtle (approved):
Looks good, just a couple nits.

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
```diff
@@ -61,17 +61,20 @@ public McpStdioServer(McpToolRegistry toolRegistry, IServiceProvider serviceProv
...
+            // still follows the configured console input encoding in stdio mode.
+            TextReader reader = Console.In;
```
Aniruddh25: Do we no longer need utf8NoBom for stdin? Why was it needed before and what changed?

Aniruddh25 (approved):
Approving with some questions to avoid regression around UTFNoBOM

src/Azure.DataApiBuilder.Mcp/Core/McpStdioServer.cs
anushakolan: We needed that explicit UTF-8 reader before because we were directly opening the stdin byte stream and creating our own text reader, so encoding had to be specified there. After switching to the standard console reader path, input decoding now follows the console input encoding configured at startup for stdio mode, so the separate UTF-8 reader setup is no longer necessary.

aaronburtle (approved):

PR #3649, opened 2026-06-04
Comprehensive MCP tool integration tests and refactoring

### LINKED ISSUE #2922: [MCP] Add integration tests for MCP tools
Add new tests for MCP. Tests should have coverage on MCP related config and runtime. MCP tools added should also have good test coverage.

souvikghosh04:
## Why make this change?

Add comprehensive integration test coverage for MCP built-in tools (ReadRecords, CreateRecord, UpdateRecord, DeleteRecord, AggregateRecords) and refactor existing tests to eliminate code duplication.

## What is this change?

- New `McpToolTestBase` shared base class with reusable service provider builders and assertion helpers
- New integration tests for 5 MCP DML tools covering success paths and error cases
- Refactored `ExecuteEntityToolMsSqlIntegrationTests` and `DynamicCustomToolMsSqlIntegrationTests` to inherit from base class
- Used `[DataTestMethod]`/`[DataRow]` to collapse similar test cases

## How was this tested?

- [x] Integration Tests
- [ ] Unit Tests

## Sample Request(s)

N/A - test-only change.


src/Service.Tests/Mcp/CreateRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            DatabaseEngine = TestCategory.MSSQL;
+            await InitializeTestFixture();
+        }
+
+        /// <summary>
+        /// Creates a new book record with valid data and verifies success.
+        /// </summary>
+        [TestMethod]
+        public async Task CreateRecord_ValidData_ReturnsSuccess()
```
anushakolan, line 34: `CreateRecord_ValidData_ReturnsSuccess` and `CreateRecord_ReturnsCreatedData` insert records into the shared MSSQL test database but do not clean them up. Since other tests in this PR assert counts/order/filter behavior over the same Book table, these inserts make the suite non-idempotent and potentially order-dependent. Can we either wrap each mutation test in cleanup logic, use a per-test transaction/rollback pattern, or create records with a unique marker and delete them in finally?

src/Service.Tests/Mcp/CreateRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            CallToolResult result = await ExecuteCreateAsync("Book", data);
+
+            AssertSuccess(result, "CreateRecord should succeed.");
+
+            JsonElement root = ParseResultRoot(result);
+
+            Assert.IsTrue(root.TryGetProperty("result", out JsonElement resultElement),
+                "Response should contain 'result' property.");
+
+            if (resultElement.ValueKind == JsonValueKind.Object && resultElement.TryGetProperty("value", out JsonElement valueArray))
```
anushakolan: A few tests make assertions only inside conditional blocks and otherwise pass silently. For example, if the response does not contain the expected `result.value` shape, `CreateRecord_ReturnsCreatedData` would still pass. Can we make the response shape required by using `Assert.IsTrue/TryGetProperty` before accessing it or add an `Assert.Fail` in the else branch?

anushakolan: Better pattern would be, 

```
Assert.IsTrue(root.TryGetProperty("result", out JsonElement resultElement),
    "Response should contain 'result'.");

Assert.AreEqual(JsonValueKind.Object, resultElement.ValueKind,
    "Create response result should be an object.");

Assert.IsTrue(resultElement.TryGetProperty("value", out JsonElement valueArray),
    "Create response result should contain 'value'.");

Assert.AreEqual(JsonValueKind.Array, valueArray.ValueKind);
Assert.IsTrue(valueArray.GetArrayLength() > 0);```

File: src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs [hunk 3]
```diff
@@ -0,0 +1,315 @@
...
+    public class AggregateRecordsToolMsSqlIntegrationTests : McpToolTestBase
+    {
+        [ClassInitialize]
+        public static async Task SetupAsync(TestContext context)
+        {
+            DatabaseEngine = TestCategory.MSSQL;
+            await InitializeTestFixture();
+        }
+
+        #region COUNT Tests
```
anushakolan, line 30: The aggregate tests currently assert lower bounds such as `count >= 21` and `count >= 10`. Because the PR also adds create/update/delete tests that mutate the same Book table, these assertions can pass even when filtering/counting is incorrect. Can we isolate the test data or assert exact expected values from a controlled fixture?

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        /// <summary>
+        /// Validates that numeric aggregation functions (sum, avg, min, max) succeed on publisher_id.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("sum", "sum_publisher_id", DisplayName = "SUM of publisher_id")]
+        [DataRow("avg", "avg_publisher_id", DisplayName = "AVG of publisher_id")]
+        [DataRow("min", "min_publisher_id", DisplayName = "MIN of publisher_id")]
+        [DataRow("max", "max_publisher_id", DisplayName = "MAX of publisher_id")]
+        public async Task Aggregate_NumericFunction_ReturnsResult(string function, string expectedAlias)
```
anushakolan: `Aggregate_NumericFunction_ReturnsResult` accepts `expectedAlias` but does not assert it. The test currently only checks for non-empty content, so it could pass even if the aggregate function returned the wrong field or response shape. Can we parse the response and assert that `expectedAlias` exists and contains a numeric value?

src/Service.Tests/Mcp/DeleteRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,180 @@
...
+
+            AssertSuccess(result, "DeleteRecord should succeed for existing record.");
+            Assert.IsFalse(string.IsNullOrWhiteSpace(GetFirstTextContent(result)));
+        }
+
+        /// <summary>
+        /// Attempts to delete a record with a non-existent key, expecting an error.
+        /// </summary>
+        [TestMethod]
+        public async Task DeleteRecord_NonExistentKey_ReturnsError()
```
anushakolan, line 54: Tests use hardcoded `id = 99999` as a non-existent key. This is fragile, especially because some create tests currently leave inserted records behind. Can we generate a guaranteed-missing key dynamically, or create/delete a record and then use that deleted id for the not-found path?

src/Service.Tests/Mcp/McpToolTestBase.cs
```diff
@@ -0,0 +1,225 @@
...
+            if (expectedSubstring != null)
+            {
+                StringAssert.Contains(GetFirstTextContent(result), expectedSubstring);
+            }
+        }
+
+        /// <summary>
+        /// Parses the first text content of a result into a JsonElement root.
+        /// </summary>
+        protected static JsonElement ParseResultRoot(CallToolResult result)
```
anushakolan, line 218: `ParseResultRoot` returns `JsonDocument.Parse(content).RootElement` without disposing or cloning the JsonDocument. To avoid retaining unmanaged buffers and to make the returned JsonElement independent of the document lifetime, can we return `RootElement.Clone()` from a using block?

```
using JsonDocument document = JsonDocument.Parse(content);
return document.RootElement.Clone();
```

src/Service.Tests/Mcp/UpdateRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,222 @@
...
+            if (resultElement.ValueKind == JsonValueKind.Array && resultElement.GetArrayLength() > 0)
+            {
+                return resultElement[0].GetProperty("id").GetInt32();
+            }
+
+            Assert.Fail("Could not extract created book ID from response.");
+            return -1;
+        }
+
+        private static async Task DeleteBook(int id)
```
anushakolan, line 186: The cleanup helper `DeleteBook` ignores the delete result. If cleanup fails, the test will still pass while leaving test data behind. Can we assert success or at least fail the test with the returned content when cleanup does not complete?

```
CallToolResult deleteResult = await ExecuteToolAsync(deleteTool, serviceProvider, args);
AssertSuccess(deleteResult, $"Cleanup failed for Book id {id}.");
```

anushakolan:
Thanks, Souvik for this broad test coverage. I have added some comments that needs to be addressed. Also, the CosmosDB and MSSQL integration tests are failing, please take a look at that as well.

src/Service.Tests/Mcp/DeleteRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,180 @@
...
+            CreateRecordTool createTool = new();
+
+            var args = new Dictionary<string, object?>
+            {
+                { "entity", "Book" },
+                { "data", new Dictionary<string, object> { { "title", title }, { "publisher_id", 1234 } } }
+            };
+
+            CallToolResult createResult = await ExecuteToolAsync(createTool, serviceProvider, args);
+            Assert.IsTrue(createResult.IsError != true, $"Setup: Failed to create book for deletion test. {GetFirstTextContent(createResult)}");
```
anushakolan: Nit: we consistently use `result.IsError != true`; consider `Assert.IsFalse(result.IsError)` for readability.

anushakolan (changes requested):

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
+// Copyright (c) Microsoft Corporation.
```
anushakolan, line 1: Several tests accept multiple response shapes. Because the MCP tool response contract is stable, it would be better for these integration tests to assert the expected shape strictly. Otherwise, regressions in response formatting could go unnoticed.

anushakolan, line 1: `CreateBookForUpdate`, `CreateBookForDeletion`, and response-id extraction are very similar. Since this PR introduces `McpToolTestBase`, we could centralize these helpers to reduce drift and make future MCP mutation tests easier to add.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+// Licensed under the MIT License.
```
anushakolan, line 2: We’ve made good progress centralizing DI wiring into `McpToolTestBase`, but a few tests still construct or diverge from those helpers. This creates a risk of drift and inconsistent behavior across tests. Can we ensure all MCP tool tests exclusively use the base class builders unless there’s a strong reason not to?

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+
```
anushakolan, line 3: nit: Minor: some test names (e.g., Aggregate_NumericFunction_ReturnsResult) are a bit generic and don’t capture the specific scenario or expected behavior. Since this suite is meant as integration coverage, slightly more descriptive names (Given/When/Then style) would make it easier to debug failures.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+using System;
```
anushakolan, line 4: We repeat response-shape parsing logic (result/value handling, array vs object cases) across multiple tests. Since this is MCP-specific response behavior, it might be worth extracting helpers like `GetResultArray(result)` or `GetFirstRecord(result)` into `McpToolTestBase` to reduce duplication and centralize contract handling.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+using System.Collections.Generic;
```
anushakolan, line 5: Coverage is strong overall, but a few additional negative scenarios could improve robustness:
- invalid filter/orderby syntax in ReadRecords
- invalid field names or type mismatches in Create/Update
- invalid groupby fields or combinations in Aggregate

These would help validate request validation paths more thoroughly.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+using System.Text.Json;
```
anushakolan, line 6: Suggestion: We have strong isolated coverage per tool, and one delete→read validation. It might be valuable to add a single end-to-end lifecycle test (Create → Read → Update → Read → Delete → Read) to validate consistency across MCP tools as a system.
``

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: The >= assertions are intentional for a shared integration DB where mutation tests may add records. Using exact values would make tests brittle to execution order. The MIN/MAX tests already assert exact expected values (1234/2345) since seed data bounds are stable.

src/Service.Tests/Mcp/DeleteRecordToolMsSqlIntegrationTests.cs
souvikghosh04: In practice, auto-increment IDs in the test DB won't reach 99999 during a test run, so this is stable. Creating/deleting a record just to get a 'known-deleted' ID adds complexity for minimal benefit here.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: Aggregate tests now strictly assert JsonValueKind.Array and property existence. The GroupBy non-paginated test still handles both shapes since the tool may or may not paginate based on result count.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: All mutation tests now use from the base class. Only the direct tool.ExecuteAsync(null, ...) calls in NoArguments tests bypass helpers intentionally (testing null-input edge cases).

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: Renamed Aggregate_NumericFunction_ReturnsResult to [Aggregate_NumericFunction_ReturnsExpectedAlias] for clarity. Other names follow the Tool_Scenario_ExpectedBehavior convention which is sufficient for debugging.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: Added ExtractCreatedBookId() to base class. For read results, the shape is consistent (result.value array), so inline GetProperty calls are clear enough without an abstraction

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: Currently all tools have integration tests coverage which was fundamental purpose of the PR. We can add scenario based tests later based on the area or feature worked on in future.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: Currently all tools have integration tests coverage which was fundamental purpose of the PR. We can add scenario based tests later based on the area or feature worked on in future.

anushakolan (approved):
Thanks for addressing all the comments, Souvik!! This looks good to me.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,324 @@
...
+            Assert.IsTrue(resultArray.GetArrayLength() > 0, "Result array should not be empty.");
+            Assert.AreEqual(1234, resultArray[0].GetProperty("min_publisher_id").GetInt32(),
+                "MIN publisher_id should be 1234 (from seed data).");
+        }
+
+        /// <summary>
+        /// Validates that MAX returns the expected maximum value.
+        /// </summary>
+        [TestMethod]
+        public async Task Aggregate_Max_ReturnsExpectedMaxValue()
```
RubenCerna2079, line 151: I think we should also add SUM and AVG tests and assert that the expected value is correct.

src/Service.Tests/Mcp/CreateRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,147 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using ModelContextProtocol.Protocol;
+
+namespace Azure.DataApiBuilder.Service.Tests.Mcp
+{
+    /// <summary>
+    /// Integration tests for CreateRecordTool against a real MsSql database.
+    /// </summary>
+    [TestClass, TestCategory(TestCategory.MSSQL)]
+    public class CreateRecordToolMsSqlIntegrationTests : McpToolTestBase
```
RubenCerna2079, line 19: Will we add test for other databases at some point? If we are planning to do so we should add tasks for that.

src/Service.Tests/Mcp/CreateRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                { "publisher_id", 1234 }
+            };
+
+            CallToolResult result = await ExecuteCreateAsync("Book", data);
+
+            AssertError(result, message: "CreateRecord should fail when missing required NOT NULL fields.");
+        }
+
+        private static async Task<CallToolResult> ExecuteCreateAsync(string entity, Dictionary<string, object> data)
```
RubenCerna2079, line 143: nit: I think it would be good to add regions similar to how it is done in AggregateRecordsToolMsSqlIntegrationTests.cs

src/Service.Tests/Mcp/DeleteRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            IServiceProvider serviceProvider = BuildMutationServiceProvider();
+            DeleteRecordTool tool = new();
+
+            var args = new Dictionary<string, object?>
+            {
+                { "entity", entity },
+                { "keys", keys }
+            };
+
+            return await ExecuteToolAsync(tool, serviceProvider, args);
```
RubenCerna2079, line 168: nit: Same comment as above, would be better to use regions.

src/Service.Tests/Mcp/DynamicCustomToolMsSqlIntegrationTests.cs
```diff
@@ -115,10 +101,8 @@ public async Task DynamicCustomTool_InvalidParamName_ReturnsError(string entityN
             Dictionary<string, object> parameters = new() { { paramName, paramValue } };
             CallToolResult result = await ExecuteCustomToolAsync(entityName, parameters);
 
-            Assert.IsTrue(result.IsError == true,
+            AssertError(result, paramName,
                 $"Custom tool should reject parameter '{paramName}' not in DB metadata for '{entityName}'.");
-            string content = GetFirstTextContent(result);
-            StringAssert.Contains(content, paramName);
```
RubenCerna2079, line 121: Why would we not want to continue ensuring that the actual data is correct.

src/Service.Tests/Mcp/AggregateRecordsToolMsSqlIntegrationTests.cs
souvikghosh04: these are covered in different tests
image

src/Service.Tests/Mcp/CreateRecordToolMsSqlIntegrationTests.cs
souvikghosh04: Currently MCP is mainly focused around MSSQL and haven't been worked on for other databases. Although, some functionalities may work for other databases, but we haven't published it officially. In future when we have full support, we will introduce those tests.

src/Service.Tests/Mcp/DynamicCustomToolMsSqlIntegrationTests.cs
souvikghosh04: AssertError with the expectedSubstring parameter already calls StringAssert.Contains on the response content, so the assertion is still being made.

src/Service.Tests/Mcp/ReadRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,189 @@
...
+                Assert.IsTrue(currentId <= previousId, $"Records should be in descending order. Got {currentId} after {previousId}.");
+                previousId = currentId;
+            }
+        }
+
+        /// <summary>
+        /// Reads records with first parameter to limit page size.
+        /// </summary>
+        [TestMethod]
+        public async Task ReadRecords_WithFirst_ReturnsLimitedResults()
```
RubenCerna2079, line 112: I think we should also add a test with the after link, and ensure it gets the correct records. We can run using `first` parameter and then use the after link it gives to run the next `ExecuteReadAsync`.

src/Service.Tests/Mcp/ReadRecordsToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// </summary>
+        [TestMethod]
+        public async Task ReadRecords_InvalidEntity_ReturnsError()
+        {
+            CallToolResult result = await ExecuteReadAsync("NonExistentEntity");
+
+            AssertError(result, "NonExistentEntity");
+        }
+
+        private static async Task<CallToolResult> ExecuteReadAsync(
```
RubenCerna2079, line 193: nit: I think it would be good to add regions to separate the tests from the helper methods

src/Service.Tests/Mcp/UpdateRecordToolMsSqlIntegrationTests.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            string argsJson = JsonSerializer.Serialize(args);
+            using JsonDocument arguments = JsonDocument.Parse(argsJson);
+
+            CallToolResult result = await tool.ExecuteAsync(arguments, serviceProvider, CancellationToken.None);
+
+            AssertError(result);
+        }
+
+        private static async Task<CallToolResult> ExecuteUpdateAsync(
```
RubenCerna2079, line 163: nit: Same as previous comment

RubenCerna2079 (approved):
LGTM! Approving assuming comments will be resolved.

RubenCerna2079 (approved):

PR #3651, opened 2026-06-06
Remove excess logs in dab validate command

### LINKED ISSUE #3266: [Bug]: `dab validate` outputs too much
## What?

Dab validate should only show problems: Warnings or Errors. 

 - Never show `information`.
 - The exception to this is "Config is valid" at the end.

> Apologies to @RubenCerna2079 who tried to explain this to me earlier.

Image

RubenCerna2079:
## Why make this change?

- Solves issue #3266

## What is this change?

Remove excessive logs when the user uses `dab validate`. 
This includes: 
- Removing unnecessary one-liner information logs
- Reducing warning logs so that only more general warnings appear (e.g. logs related to entities missing `field` descriptions when using MCP)
- Only showing certain relevant information logs when the user uses `dab start`. All of this is done in order to ensure that the logs in `dab validate` are not cluttered. (e.g. logs related to REST paths for entities when enabled)
- Removed logs assertion in tests that used it, as those logs were removed.

## How was this tested?

- [ ] Integration Tests
- [ ] Unit Tests
- [x] Manual Tests
Tested manually by running `dab validate` and seeing that the logs are as expected, and ran `dab start` to ensure previous logs didn't change.

## Sample Request(s)

Previous `dab validate` logs:
image

New `dab validate` logs:
image


src/Cli/ConfigGenerator.cs
```diff
@@ -3155,8 +3156,6 @@ public static bool IsConfigValid(ValidateOptions options, FileSystemRuntimeConfi
                 return false;
             }
 
-            _logger.LogInformation("Validating config file: {runtimeConfigFile}", runtimeConfigFile);
-
             RuntimeConfigProvider runtimeConfigProvider = new(loader);
```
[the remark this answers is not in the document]
RubenCerna2079: This is not necessary, the only place where we want to know if the config is valid or invalid is once `dab validate` completes.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -123,16 +123,19 @@ public SqlMetadataProvider(
...
-                    _logger.LogInformation(message: "REST calls are disabled for the entity: {entity}", entityName);
+                    if (runtimeConfig.IsRestEnabled)
+                    {
+                        string restPath = entityMetatdata.Rest?.Path ?? entityName;
+                        _logger.LogInformation("[{entity}] REST path: {globalRestPath}/{entityRestPath}", entityName, runtimeConfig.RestPath, restPath);
+                    }
+                    else
+                    {
+                        _logger.LogInformation(message: "REST calls are disabled for the entity: {entity}", entityName);
+                    }
```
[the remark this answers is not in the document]
RubenCerna2079: These logs will only occur in `dab start` the purpose of the PR is to specifically keep logging to a minimum in `dab validate`

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -362,13 +368,16 @@ protected override async Task GenerateAutoentitiesIntoEntities(IReadOnlyDictiona
...
-                    else
-                    {
-                        _logger.LogInformation(message: "REST calls are disabled for the entity: {entity}", entityName);
+                        if (runtimeConfig.IsRestEnabled)
+                        {
+                            _logger.LogInformation("[{entity}] REST path: {globalRestPath}/{entityRestPath}", entityName, runtimeConfig.RestPath, entityName);
+                        }
+                        else
+                        {
+                            _logger.LogInformation(message: "REST calls are disabled for the entity: {entity}", entityName);
```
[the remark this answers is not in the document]
RubenCerna2079: These logs will only occur in `dab start` the purpose of the PR is to specifically keep logging to a minimum in `dab validate`

src/Service.Tests/Configuration/ConfigurationTests.cs
aaronburtle: >/// It also verifies that the expected log message is logged.

I believe this was removed, so comment is now stale. Similar comment for `TestBasicConfigSchemaWithNoOptionalFieldsIsValid`

src/Core/Services/MetadataProviders/MsSqlMetadataProvider.cs
```diff
@@ -362,13 +368,16 @@ protected override async Task GenerateAutoentitiesIntoEntities(IReadOnlyDictiona
                             subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorInInitialization);
                     }
 
-                    if (runtimeConfig.IsRestEnabled)
+                    if (!_isValidateOnly)
```
aaronburtle: This block seems to be duplicated in `SqlMetadataProvider`, can factor out to make sure they maintain behavior and eliminate code duplication.

aaronburtle:
This PR removes assertions but doesnt add in any new ones, could consider adding in a negative case that asserts the suppressed logs are absent when isValidateOnly is true to help prevent any regression.

aaronburtle (approved):
Looks good, just a couple comments.

RubenCerna2079:
> This PR removes assertions but doesnt add in any new ones, could consider adding in a negative case that asserts the suppressed logs are absent when isValidateOnly is true to help prevent any regression.

Added new test to ensure suppressed logs are absent when isValidateOnly is true

src/Cli/Commands/ValidateOptions.cs
```diff
@@ -38,7 +38,7 @@ public int Handler(ILogger logger, FileSystemRuntimeConfigLoader loader, IFileSy
             }
             else
             {
-                logger.LogError("Config is invalid.");
+                logger.LogInformation("Config is invalid.");
```
souvikghosh04, line 41: if users were relying previously over Error signals and now they have to rely on Info signals, we have to ensure its documented or else we might get this as an issue/discussion from users in future. can we double check with Jerry if this could be an issue?

src/Cli/ConfigGenerator.cs
```diff
@@ -3196,14 +3194,21 @@ public static bool IsConfigValid(ValidateOptions options, FileSystemRuntimeConfi
...
-                                _logger.LogWarning($"Entity '{entity.Key}' is missing 'fields' definition while MCP is enabled. " +
-                                    "It's recommended to define fields explicitly to ensure optimal performance with MCP.");
+                                missingFields = true;
+                                _logger.LogDebug($"Entity '{entity.Key}' is missing 'fields' definition while MCP is enabled.");
                             }
                         }
+
+                        if (missingFields)
+                        {
+                            _logger.LogWarning($"One or more entities are missing `fields` definition while MCP is enabled. " +
```
souvikghosh04: nit: remove the `$` as the entire string is just constant.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
```diff
@@ -1019,6 +1008,30 @@ protected virtual void PopulateMetadataForLinkingObject(
...
         }
 
+        /// <summary>
+        /// Helper method that logs the REST paths for all entities and shows if the REST calls are enabled/disabled for any entity.
+        /// </summary>
+        /// <param name="runtimeConfig"></param>
+        /// <param name="entities"></param>
+        protected void LogRestPathsForEntities(RuntimeConfig runtimeConfig, IReadOnlyDictionary<string, Entity> entities)
+        {
+            if (!_isValidateOnly)
```
souvikghosh04, line 1018: validation should be moved outside to callers. this is a helper method specifically for Logging as the name suggests.

src/Cli.Tests/ValidateConfigTests.cs
```diff
@@ -405,6 +402,37 @@ public void TestValidateNonRootZeroEntitiesWithInvalidConnectionString()
...
 
+    /// <summary>
+    /// Tests that logs related to suppressed messages do not appear.
+    /// </summary>
+    [TestMethod]
+    public async Task TestValidateSuppressedLogsDoNotAppear()
+    {
+        // Arrange
+        StringWriter writer = new();
+        Console.SetOut(writer);
```
souvikghosh04, line 423: The test calls Console.SetOut(writer) but never restores the original Console.Out. If tests later in the same fixture also write to console (or the test class doesn't have a [TestCleanup] that resets it), the writer will leak. Adding [TestCleanup] would probably be more cleaner and efficient.

souvikghosh04 (commented):
the downgrading of LogError to LogInformation can have some impact on users. we should check this.

src/Cli/Commands/ValidateOptions.cs
RubenCerna2079: This was already discussed with @JerryNixon, it is expected that the error message comes from the actual error that is happening, while letting the user know if the config is valid or invalid should only be an info message.

src/Core/Services/MetadataProviders/SqlMetadataProvider.cs
RubenCerna2079: I think it is better to keep it where it is. This method will only be used in this section as it is protexted which has access to the `_isValidationOnly` and if we need to use it somewhere new at some point in the future it will still need to check the validation in order to not add unnecessary logs.

souvikghosh04 (approved):
Approving, as the comments have been addressed

Issue #3257, opened 2026-03-14
[Bug]: `--LogLevel` should be `--loglevel` (and case insensitive) like all the other CLI flags


JerryNixon:
<img width="904" height="540" alt="Image" src="https://github.com/user-attachments/assets/c7eddb14-7515-48f1-a92b-c3d9bcae57b7" />

RubenCerna2079:
I don't see any flags that are case insensitive

PR #3653, opened 2026-06-08
Change CLI flag --LogLevel to --log-level

RubenCerna2079:
## Why make this change?

- This solves issue #3257

## What is this change?

We changed the `dab start` `--LogLevel` flag to `--log-level` in order to have it follow the conventions we already have. We also allow for the CLI to still use the previous `--LogLevel` flag but add a warning sign to ensure the user knows it is deprecated. Lastly, also changed all of the comments that mention the previous `--LogLevel` and updated it to use the new `--log-level`.
Added new `isLogLevelLegacy ` value that is passed into the `DynamicLogLevelProvider.cs` which would then be used to add the log in case the deprecated flag is used in an Aspire scenario. In which the flag is passed to the dll directly and not through CLI

## How was this tested?

- [ ] Integration Tests
- [X] Unit Tests
Changed the tests that were for `--LogLevel` so they now use `--log-level` and added a few extra cases to ensure the `--LogLevel` flag still works

## Sample Request(s)
dab start --log-level information
dab start --LogLevel warning


src/Cli/ConfigGenerator.cs
```diff
@@ -3079,19 +3079,30 @@ public static bool TryStartEngineWithOptions(StartOptions options, FileSystemRun
...
             if (options.LogLevel is not null)
             {
-                if (options.LogLevel is < LogLevel.Trace or > LogLevel.None)
+                logLevel = options.LogLevel;
+            }
+            else if (options.LogLevelLegacy is not null)
+            {
+                options.CliBuffer.BufferLog(LogLevel.Warning, $"--LogLevel is deprecated, please use --log-level instead.");
+                logLevel = options.LogLevelLegacy;
+            }
```
souvikghosh04, line 3091: do we really need this branch logic? since the new naming convention will be part of new release and documented anyways. instead, the unsupported format should ideally error out instead of a warning or silent fallback.

src/Service/Program.cs
```diff
@@ -237,19 +237,21 @@ private static LogLevel GetLogLevelFromCommandLineArgsOrConfig(string[] args, bo
             LogLevel logLevel;
             isConfigOverriding = false;
 
-            // Check if --LogLevel was explicitly specified via CLI (case-insensitive parsing)
-            int logLevelIndex = Array.FindIndex(args, a => string.Equals(a, "--LogLevel", StringComparison.OrdinalIgnoreCase));
+            // Check if --log-level or --LogLevel was explicitly specified via CLI (case-insensitive parsing)
+            int logLevelIndex = Array.FindIndex(args, a =>
+                string.Equals(a, "--log-level", StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(a, "--LogLevel", StringComparison.OrdinalIgnoreCase));
```
souvikghosh04: The deprecation warning only fires through the dab start's ConfigGenerator path. When the service host is invoked directly (e.g., dotnet Azure.DataApiBuilder.Service.dll --LogLevel Information, which is what containers and Aspire scenarios do), the legacy flag works but the user is never told it's deprecated. Either emit the same warning here, or document explicitly that the service host will keep both spellings indefinitely. Additionally, I also have commented about do we really need to handle deprecated scenario since the unsupported format ideally should error out and document in new release.

src/Cli/Commands/StartOptions.cs
```diff
@@ -21,11 +21,12 @@ public class StartOptions : Options
 
         public LogBuffer CliBuffer { get; }
 
-        public StartOptions(bool verbose, LogLevel? logLevel, bool isHttpsRedirectionDisabled, bool mcpStdio, string? mcpRole, string config)
+        public StartOptions(bool verbose, LogLevel? logLevel, LogLevel? logLevelLegacy, bool isHttpsRedirectionDisabled, bool mcpStdio, string? mcpRole, string config)
```
souvikghosh04: New parameters should be added at the end to avoid any existing callers breaking the sequenece of intended params passed in earlier format.

souvikghosh04 (commented):
Posting comments so far. rethink if we really need the legacy flag or should error out on unsupported format as we generally do and document these.

src/Cli/ConfigGenerator.cs
RubenCerna2079: I think it should throw a warning right now since we are not sure if there are any users that are currently using the old flag. We need to give them time to adapt to the new flag before it is completely deprecated.

src/Service/Program.cs
RubenCerna2079: Had to add new logic to ensure that the warning log also appears through the situation you showed, this is because this section cannot log things as the loggers are still not even created.

src/Cli/ConfigGenerator.cs
souvikghosh04: supporting multiple flag options for same functionality would be confusing and erroneous in future. The docs are essentially the single source of truth and users must follow that. If you really want to keep this for now, please create a task to clean this up sometime later and update the docs. also, its only recently this was introduced and this might not be used much. so it shouldn't be a problem to deprecate this.

souvikghosh04 (approved):
Looks good. Just make sure to cleanup the deprecated flag as supporting both would create confusion.

JerryNixon (approved):

aaronburtle:
`docs/testing-guide/ai-foundry-integration.md` is still using the deprecated flag:

`--command-line "dotnet Azure.DataApiBuilder.Service.dll --ConfigFileName $DAB_CONFIG_PATH --LogLevel Debug"`

src/Service/Program.cs
```diff
@@ -222,34 +222,46 @@ public static IHostBuilder CreateHostBuilder(string[] args, bool runMcpStdio, st
...
+            foreach (string argument in args)
+            {
+                if (string.Equals(argument, "--LogLevel", StringComparison.OrdinalIgnoreCase))
+                {
+                    isLogLevelLegacy = true;
+                    break;
+                }
+            }
+
+            int logLevelIndex = Array.FindIndex(args, a =>
```
aaronburtle: nit: this scan is redundant with the loop above

src/Cli/Commands/StartOptions.cs
aaronburtle: nit: think this comment is stale now

aaronburtle (approved):
Looks good overall just a couple comments.

src/Cli/ConfigGenerator.cs
RubenCerna2079: New task was created for this to be done in the future. https://github.com/Azure/data-api-builder/issues/3679

aaronburtle (approved):
re-approval

Issue #3332, opened 2026-03-27
[Enh]: Embedding Phase 4: Table semantic search


JerryNixon:
## Phase 4: Semantic search over Redis vector indexes

This phase adds semantic search to DAB read operations by using Redis vector search to find candidate database keys, then using DAB’s existing query pipeline to return database records. The sample issue you provided uses the same pattern of introducing a focused config surface, showing JSON shape, and describing the runtime behavior. ([GitHub][1])

### Goal

Enable callers to pass natural language into REST, GraphQL, and eventually MCP read operations.

DAB will:

1. Embed the semantic search value.
2. Query a configured Redis vector index.
3. Read database key values from Redis results.
4. Collapse duplicate keys.
5. Query the database using the original request plus key narrowing.
6. Return normal DAB results with an optional semantic distance field.

Redis population, chunking, backfill, embedding jobs, and index maintenance are out of scope.

## Entity configuration

Semantic search is configured per explicit entity.

```json
{
  "entities": {
    "Product": {
      "source": {
        "type": "table",
        "object": "dbo.Product"
      },
      "semantic-search": {
        "enabled": true,
        "redis-index-name": "idx:product-semantic",
        "redis-index-type": "hash",
        "redis-index-multiplier": 2,
        "similarity-threshold": 0.8,
        "input-description": "Searches product descriptions using natural language.",
        "output-description": "Semantic distance score returned by semantic search."
      }
    }
  }
}
```

### Properties

| Property                 |              Type |                                                Default |   Min |   Max | Required                      |
| ------------------------ | ----------------: | -----------------------------------------------------: | ----: | ----: | ----------------------------- |
| `enabled`                | boolean-or-string |                                                `false` |   n/a |   n/a | no                            |
| `redis-index-name`       |            string |                                                   none |   n/a |   n/a | only when `enabled` is `true` |
| `redis-index-type`       |       string enum |                                                 `hash` |   n/a |   n/a | no                            |
| `redis-index-multiplier` |           integer |                                                    `2` |   `1` |  `10` | no                            |
| `similarity-threshold`   |            number |                                                  `0.8` | `0.0` | `1.0` | no                            |
| `input-description`      |            string |     `Natural language value used for semantic search.` |   n/a |   n/a | no                            |
| `output-description`     |            string | `Semantic distance score returned by semantic search.` |   n/a |   n/a | no                            |

## JSON schema addition

Add `semantic-search` to explicit `entities.*.properties`.

```json
[...]

PR #3663, opened 2026-06-11
Redis-Backed Semantic Search (Phase 4)

ajtiwari07:
**Summary**
This PR adds end-to-end semantic (vector) search support to DAB, enabling entities to be queried using natural language. Results are retrieved from a Redis vector index via FT.SEARCH KNN and then narrowed in the SQL database to return fully enriched records, with results ordered by semantic distance.

Git Issue: https://github.com/Azure/data-api-builder/issues/3332

**What's New**
Semantic Search Pipeline
[RedisSemanticSearchService](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — new service that takes the user's search text, generates an embedding (or accepts a raw float array), issues a Redis FT.SEARCH KNN query against a configured index, and returns primary-key candidates with cosine distances
ApplySemanticCandidates() in SqlQueryStructure — translates Redis candidates into a SQL [WHERE (pk1=… AND pk2=…) OR (…)](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) predicate, so only the semantically relevant rows are fetched from the database
TryApplySemanticNarrowingAsync() in SqlQueryEngine — orchestrates the full flow: call Redis → deduplicate candidates by PK signature → apply SQL predicate → attach distance scores → re-order results by distance (unless user specified explicit ordering)
SemanticDistanceByPrimaryKeySignature map on SqlQueryStructure — carries Redis distances through to the final JSON response, injected as _distance / distance fields for REST and GraphQL respectively

**Configuration**
New entity.semantic_search config block per entity: enabled, redis-index-name, [redis-index-type](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (hash/[json](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html)), redis-index-multiplier, [similarity-threshold](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
runtime.cache.level-2.connection-string is reused as the Redis connection for both caching and vector search
runtime.embeddings block: provider, endpoint, [api-key](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html), model, api-version, dimensions, timeout-ms
CLI support: --semantic-search.enabled, --semantic-search.redis-index-name, [--semantic-search.redis-index-type](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html), --semantic-search.redis-index-multiplier, --semantic-search.similarity-threshold
Schema updated in [dab.draft.schema.json](vscode-file://vscode-app/c:/Users/ajtiwari/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html)


src/Cli/Commands/ConfigureOptions.cs
```diff
@@ -61,6 +61,8 @@ public ConfigureOptions(
             int? runtimePaginationMaxPageSize = null,
             int? runtimePaginationDefaultPageSize = null,
             bool? runtimePaginationNextLinkRelative = null,
+            string? runtimeCacheLevel2Provider = null,
```
ajtiwari07, line 64: Added these configurations since they were missing from CLI.

ajtiwari07 (commented):
Fixed failing validation compile issue in tests: removed invalid Dictionary.AsReadOnly() calls in SemanticSearchTextFlowTests. Commit 3cbca260 has been pushed to this PR branch.

ajtiwari07 (commented):
Update: commit 3cbca260 pushed with CI compile fix in SemanticSearchTextFlowTests (remove invalid Dictionary.AsReadOnly() usage). Re-run should pick up this change.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -92,6 +93,80 @@ public class SqlQueryStructure : BaseSqlQueryStructure
...
+                Predicates.Add(Predicate.MakeFalsePredicate());
+                return;
+            }
+
+            Predicate? combinedOr = null;
+
+            foreach (SemanticSearchCandidate candidate in candidates)
+            {
+                Predicate? andPredicate = null;
+                foreach ((string columnName, object? value) in candidate.ColumnValues)
```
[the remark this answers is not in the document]
ajtiwari07: Cannot make this assumption on behalf of the clients, we will query on all columns present in the document from vector search.

src/Core/Models/SemanticSearchCandidate.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Azure.DataApiBuilder.Core.Models;
+
+/// <summary>
+/// Represents one semantic candidate with SQL column values extracted from
+/// the semantic document and primary key values used for dedupe/output mapping.
+/// </summary>
+public record SemanticSearchCandidate(
+	IReadOnlyDictionary<string, object?> PrimaryKeyValues,
+	IReadOnlyDictionary<string, object?> ColumnValues,
+	double Distance);
```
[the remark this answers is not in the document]
ajtiwari07: Done

src/Core/Services/RestService.cs
```diff
@@ -235,6 +236,12 @@ private async Task<IActionResult> DispatchQuery(RestRequestContext context, Data
 
             if (context is FindRequestContext findRequestContext)
             {
+                if (findRequestContext.IncludeSemanticDistanceInResponse
+                    && !findRequestContext.FieldsToBeReturned.Contains(SemanticSearchConstants.REST_DISTANCE_FIELD))
+                {
+                    findRequestContext.FieldsToBeReturned.Add(SemanticSearchConstants.REST_DISTANCE_FIELD);
+                }
```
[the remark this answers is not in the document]
ajtiwari07: Done

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,451 @@
...
+        RuntimeConfigProvider runtimeConfigProvider,
+        IMetadataProviderFactory metadataProviderFactory,
+        IEmbeddingService? embeddingService,
+        ILogger<RedisSemanticSearchService> logger)
+    {
+        _runtimeConfigProvider = runtimeConfigProvider;
+        _metadataProviderFactory = metadataProviderFactory;
+        _embeddingService = embeddingService;
+        _logger = logger;
+    }
```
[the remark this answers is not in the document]
ajtiwari07: Done

schemas/dab.draft.schema.json
```diff
@@ -1487,6 +1488,67 @@
                 }
               }
             },
+            "semantic-search": {
+              "type": "object",
+              "description": "Semantic search configuration for this entity.",
+              "additionalProperties": false,
+              "properties": {
+                "enabled": {
+                  "$ref": "#/$defs/boolean-or-string",
```
JerryNixon, line 1497: semantic-search.enabled does not honor boolean-or-string. The schema allows enabled as boolean-or-string, but EntitySemanticSearchOptions.Enabled is a bool. @env() / @akv() configs will not round-trip/resolve correctly. Use the existing DAB boolean-or-string pattern and add env/AKV tests.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,495 @@
...
+        EmbeddingResult result = await _embeddingService.TryEmbedAsync(semanticSearchValue);
+        if (!result.Success || result.Embedding is null)
+        {
+            return [];
+        }
+
+        return result.Embedding;
+    }
+
+    private static bool TryParseVectorText(string text, out float[]? vector)
```
JerryNixon, line 149: client-provided vectors bypass the embedding subsystem. TryParseVectorText() accepts raw JSON float arrays from the caller. The specs say callers provide natural language and DAB embeds it; #3331 explicitly excludes client-provided vectors. Remove raw vector input support unless I am missing something. There is no need for this added complexity as users will not use it like this, almost 99.9999% sure of this.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    private static byte[] ToRedisVectorBytes(float[] vector)
+    {
+        byte[] bytes = new byte[vector.Length * sizeof(float)];
+        Buffer.BlockCopy(vector, 0, bytes, 0, bytes.Length);
+        return bytes;
+    }
+
+    private async Task<float[]> GetEmbeddingAsync(string semanticSearchValue)
```
JerryNixon, line 128: embedding failure silently returns empty results.
GetEmbeddingAsync() returns [] when embedding is disabled/fails, and the caller returns a successful empty response. Spec says embedding failure must fail the request and must not continue as if Redis returned no matches. Throw a sanitized DAB error instead.

src/Core/Resolvers/Sql Query Structures/SqlQueryStructure.cs
```diff
@@ -92,6 +93,80 @@ public class SqlQueryStructure : BaseSqlQueryStructure
...
+        /// <summary>
+        /// Primary-key signature to semantic distance map.
+        /// </summary>
+        public Dictionary<string, double> SemanticDistanceByPrimaryKeySignature { get; }
+
+        /// <summary>
+        /// Applies a semantic narrowing predicate of the form:
+        /// (col1 = ... AND col2 = ...) OR (...)
+        /// </summary>
+        public void ApplySemanticCandidates(IReadOnlyList<SemanticSearchCandidate> candidates)
```
JerryNixon, line 115: SQL narrowing uses all Redis document columns, not primary keys. ApplySemanticCandidates() iterates candidate.ColumnValues, which can include arbitrary non-PK Redis fields. The spec says Redis results provide database key values and DAB builds a key predicate. Narrow only on PrimaryKeyValues. We're wanting simple, intuitive behavior unless there is a technicalreason I am missing here.

src/Core/Models/RestRequestContexts/RestRequestContext.cs
```diff
@@ -95,6 +95,31 @@ protected RestRequestContext(string entityName, DatabaseObject dbo)
...
+
+    /// <summary>
+    /// Optional semantic threshold override provided by the caller.
+    /// </summary>
+    public double? SemanticThreshold { get; set; }
+
+    /// <summary>
+    /// True when semantic distance should be included in REST output.
+    /// </summary>
+    public bool IncludeSemanticDistanceInResponse { get; set; }
```
JerryNixon, line 112: REST $select behavior violates the spec.
context.IncludeSemanticDistanceInResponse is set but never used by SqlQueryEngine; REST responses add semantic_distance whenever semantic search runs. Spec says $select=id,description must not return semantic_distance unless explicitly selected. It's okay to omit this, if for no other reason, so the user can expect the returning structure. Right?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -189,7 +210,230 @@ await ExecuteAsync(structure, dataSourceName, isMultipleCreateOperation: true),
...
+                    statusCode: System.Net.HttpStatusCode.BadRequest,
+                    subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+            }
+
+            double effectiveThreshold = restContext?.SemanticThreshold
+                ?? (graphQLParameters is not null && graphQLParameters.TryGetValue(QueryBuilder.SEMANTIC_THRESHOLD_ARGUMENT_NAME, out object? gqlThresholdObj) && gqlThresholdObj is double gqlThresholdVal
+                    ? gqlThresholdVal
+                    : entity.SemanticSearch.SimilarityThreshold);
+
+            int first = restContext?.First
```
JerryNixon, line 298: pagination is still active during semantic search.
Spec says pagination continuation is unsupported and first is semantic top, not a second page. This PR still flows through normal DAB pagination/limits and may emit paginated connection behavior. Disable continuation semantics and treat first only as the requested semantic result count.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,495 @@
...
+
+            RedisResult rawResult = await db.ExecuteAsync(
+                "FT.SEARCH",
+                options.RedisIndexName!,
+                $"*=>[KNN {top} @{vectorFieldName} $vec AS {VECTOR_SCORE_FIELD}]",
+                "PARAMS",
+                "2",
+                "vec",
+                vectorBytes,
+                "SORTBY",
```
JerryNixon, line 101: top semantic rows can be dropped before in-memory sorting. SQL still uses the normal _limit while sorting by semantic score happens after SQL execution. If SQL returns only the first rows by PK/order, better semantic candidates can be omitted. Fetch the full deduped candidate key set needed for ranking, then sort/truncate.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -189,7 +210,230 @@ await ExecuteAsync(structure, dataSourceName, isMultipleCreateOperation: true),
...
+            if (shouldReturnEmpty)
+            {
+                return emptySemanticResponse;
+            }
+
+            JsonDocument? response = await ExecuteAsync(structure, dataSourceName);
+            return ApplySemanticDistanceAndOrderingIfNeeded(response, structure, dataSourceName, includeRestField: true, includeGraphQlField: false);
+        }
+
+        private async Task<(bool shouldReturnEmpty, JsonDocument? emptyResponse)> TryApplySemanticNarrowingAsync(
```
JerryNixon, line 229: GraphQL empty semantic result shape is wrong for connections. TryApplySemanticNarrowingAsync() returns JsonDocument.Parse("[]") before connection shaping. For paginated GraphQL fields the expected shape is a connection with items: [], not a raw array. Return through the normal pagination formatter.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,495 @@
...
+
+        if (similarity > 1.0)
+        {
+            similarity = 1.0;
+        }
+
+        return similarity;
+    }
+
+    private static string NormalizeFieldName(string field)
```
JerryNixon, line 479: Spec requires Redis records to include real database primary key column names exactly, including case. NormalizeFieldName() strips JSON paths and accepts case-insensitive matches, which can map the wrong field. Enforce exact PK names and fail/skip invalid records deterministically.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,495 @@
...
+using Azure.Identity;
+using Microsoft.Extensions.Logging;
+using StackExchange.Redis;
+
+namespace Azure.DataApiBuilder.Service.Services.SemanticSearch;
+
+/// <summary>
+/// Resolves semantic candidates by:
+/// 1) generating/retrieving an embedding vector for semantic_search input,
+/// 2) performing a Redis FT.SEARCH KNN query,
```
JerryNixon, line 28: The spec says Redis applies the threshold. This PR runs KNN then filters in DAB. Push threshold into the Redis query or explicitly justify/spec-change the behavior.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
JerryNixon, line 149: TryParseVectorText() parses caller input as a float array before embedding. Besides violating spec, this creates avoidable CPU/memory pressure. Remove vector parsing.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,495 @@
...
+
+                primaryKeys[pk] = pkValue;
+            }
+
+            if (!hasAllPrimaryKeys)
+            {
+                continue;
+            }
+
+            results.Add(new SemanticSearchCandidate(primaryKeys, sqlColumns, similarity));
```
JerryNixon, line 376: Using Redis document columns beyond PK lets external index contents influence SQL predicates on arbitrary columns. That is brittle and can interact badly with policy/RLS. Treat Redis as candidate-key source only; DAB/database authorization remains authoritative.

JerryNixon (commented):
- keyless views are not rejected at startup. Spec requires startup failure when semantic search is enabled on a view without resolvable primary keys. Current validation checks table/view type but not resolvable PK metadata. Add metadata validation before runtime.

 - reserved-name validation misses real database columns. Reserved name checks only inspect configured fields/mappings. If the backing table has a real exposed column named semantic_distance and no explicit field config, startup may pass. Validate against resolved metadata/exposed field names.

 - Embedding/Redis failures should produce useful sanitized errors with correlation/provider details where available. Current embedding failure becomes empty results, Redis failures collapse into a generic BadRequest, and there is no provider status/error propagation.

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -189,7 +211,234 @@ await ExecuteAsync(structure, dataSourceName, isMultipleCreateOperation: true),
...
+                        subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+                }
+
+                return (false, null);
+            }
+
+            if (string.IsNullOrWhiteSpace(semanticSearchText))
+            {
+                return (false, null);
+            }
```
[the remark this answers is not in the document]
souvikghosh04: this seems like a valid point. also, the validation check on GraphQL part below seems inconsistent compared to REST.

src/Service.GraphQLBuilder/Queries/QueryBuilder.cs
```diff
@@ -191,21 +194,29 @@ public static FieldDefinitionNode GenerateGetAllQuery(
                 location: null,
                 new NameNode(GenerateListQueryName(name.Value, entity)),
                 new StringValueNode($"Get a list of all the {GetDefinedSingularName(name.Value, entity)} items from the database"),
-                QueryArgumentsForField(filterInputName, orderByInputName),
+                QueryArgumentsForField(filterInputName, orderByInputName, includeSemanticArguments: entity.SemanticSearch?.Enabled ?? false),
```
souvikghosh04, line 197: the execution seems to be happening in SqlQueryEngine.cs only so are we guarding it against datasources/databases which are not MS SQL?

src/Core/Resolvers/SqlQueryEngine.cs
```diff
@@ -79,16 +84,32 @@ public SqlQueryEngine(
...
             if (structure.PaginationMetadata.IsPaginated)
             {
+                JsonDocument? queryResult = await ExecuteAsync(structure, dataSourceName);
+                JsonDocument? processedResult = ApplySemanticDistanceAndOrderingIfNeeded(queryResult, structure, dataSourceName, includeRestField: false, includeGraphQlField: true);
+
                 return new Tuple<JsonDocument?, IMetadata?>(
-                    SqlPaginationUtil.CreatePaginationConnectionFromJsonDocument(await ExecuteAsync(structure, dataSourceName), structure.PaginationMetadata, structure.GroupByMetadata),
+                    SqlPaginationUtil.CreatePaginationConnectionFromJsonDocument(processedResult, structure.PaginationMetadata, structure.GroupByMetadata),
                     structure.PaginationMetadata);
             }
```
souvikghosh04, line 105: When semantic search yields zero candidates for a paginated GraphQL query, the code returns JsonDocument.Parse("[]") directly and skips SqlPaginationUtil.CreatePaginationConnectionFromJsonDocument(0). Paginated GraphQL collections expect a connection shape (e.g. items, hasNextPage, endCursor), so the response will not match the schema.

src/Service/Services/SemanticSearch/RedisSemanticSearchService.cs
```diff
@@ -0,0 +1,495 @@
...
+            return ParseCandidates(rawResult, options.RedisIndexType, sourceColumns, primaryKeyColumns, similarityThreshold);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "Semantic search query failed for entity {entityName} and index {indexName}.", entityName, options.RedisIndexName);
+            throw new DataApiBuilderException(
+                message: $"Semantic search index '{options.RedisIndexName}' for entity '{entityName}' was not found or could not be queried.",
+                statusCode: System.Net.HttpStatusCode.BadRequest,
+                subStatusCode: DataApiBuilderException.SubStatusCodes.BadRequest);
+        }
```
souvikghosh04, line 118: every error is classified as 400: Bad Request? what about 500, 503, 403, 404 etc?

souvikghosh04 (changes requested):
posting comments so far

PR #3664, opened 2026-06-12
Fix failure on REST requests with $filter and $orderby parameters

### LINKED ISSUE #3662: [Bug]: Regression Unable to extract $filter parameter from query string.
### What happened?

In DAB 2.0.8 the `$filter` does not work anymore, but works perfectly fine in 1.7.93.

`GET .../?$filter=numb eq '12345' and age eq '31'`

results in 

```
{
  "error": {
    "code": "BadRequest",
    "message": "Unable to extract $filter parameter from query string.",
    "status": 400
  }
}
```

### Version

2.0.8

Azure SQL

AppService

REST

### Relevant log output

```Text

```

### Code of Conduct

RubenCerna2079:
## Why make this change?

- Fix issue #3662

This issue is caused by a regression in PR https://github.com/Azure/data-api-builder/pull/3080
The regression is that we start using the raw encoded URL and we assume that the `$` character is not encoded when using the `ExtractRawQueryParameter` function, this is half true, since depending on where the URL comes from it will encode it or it will treat it as a special character and leave it as it is.
This means the function looks for the `$filter` or `$orderby` when the URL has the possibility to change them to `%24filter` or `%24orderby` which causes the failure.

## What is this change?

This change fixes a bug that caused using `$filter` or `$orderby` in REST. This is done by encoding the `parameterName` so that it has the same format as the raw URL, it will only change it if it is unable to find the original `$` character in the raw query.

## How was this tested?

- [ ] Integration Tests
- [x] Unit Tests
Added end to end test to ensure that REST requests that use `$filter` and `$orderby` succeed for both cases where a user uses the special character `$` or its URL encoded format `%24`.

## Sample Request(s)
http://localhost:5000/api/Book?$orderby=id desc
http://localhost:5000/api/Book?$filter=publisher_id 1234

### DIFF
--- a/src/Core/Services/RestService.cs
+++ b/src/Core/Services/RestService.cs
@@ -71,7 +71,8 @@ namespace Azure.DataApiBuilder.Core.Services
             DatabaseObject dbObject = sqlMetadataProvider.EntityToDatabaseObject[entityName];
 
             QueryString? query = GetHttpContext().Request.QueryString;
-            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString().Replace("%24", "$");    //Add replacement in order to ensure '$' sign is present as expected
+            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString()
+                .Replace("?%24", "?$").Replace("&%24", "&$");   //Add replacement in order to ensure '$' sign is present as expected only for the case where it is after a '?' or '&' signs.
 
             // Read the request body early so it can be used for downstream processing.
             string requestBody = string.Empty;
--- a/src/Core/Services/RestService.cs
+++ b/src/Core/Services/RestService.cs
@@ -71,8 +71,7 @@ namespace Azure.DataApiBuilder.Core.Services
             DatabaseObject dbObject = sqlMetadataProvider.EntityToDatabaseObject[entityName];
 
             QueryString? query = GetHttpContext().Request.QueryString;
-            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString()
-                .Replace("?%24", "?$").Replace("&%24", "&$");   //Add replacement in order to ensure '$' sign is present as expected only for the case where it is after a '?' or '&' signs.
+            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString();
 
             // Read the request body early so it can be used for downstream processing.
             string requestBody = string.Empty;
--- a/src/Core/Services/RestService.cs
+++ b/src/Core/Services/RestService.cs
@@ -71,7 +71,7 @@ namespace Azure.DataApiBuilder.Core.Services
             DatabaseObject dbObject = sqlMetadataProvider.EntityToDatabaseObject[entityName];
 
             QueryString? query = GetHttpContext().Request.QueryString;
-            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString();
+            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString().Replace("%24", "$");    //Add replacement in order to ensure '$' sign is present as expected
 
             // Read the request body early so it can be used for downstream processing.
             string requestBody = string.Empty;
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -4228,9 +4228,11 @@ type Planet @model(name:""PlanetAlias"") {
         /// End to end test that validates that REST requests with OData query
         /// options $filter and $orderby succeed to ensure no regression can occur.
         /// </summary>
-        [TestMethod]
+        [DataTestMethod]
+        [DataRow("/api/Book?$orderby=id desc&$filter=publisher_id eq 1234", DisplayName = "Regular REST uri")]
+        [DataRow("/api/Book?%24orderby=id%20desc&%24filter=publisher_id%20eq%201234", DisplayName = "Regular REST uri")]
         [TestCategory(TestCategory.MSSQL)]
-        public async Task TestForRestRequestsWithFilterAndOrderbyParameters()
+        public async Task TestForRestRequestsWithFilterAndOrderbyParameters(string restUri)
         {
             // The configuration file is constructed by merging hard-coded JSON strings to simulate the scenario where users manually edit the
             // configuration file (instead of using CLI).
--- a/src/Service.Tests/Configuration/ConfigurationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationTests.cs
@@ -4229,8 +4229,8 @@ type Planet @model(name:""PlanetAlias"") {
         /// options $filter and $orderby succeed to ensure no regression can occur.
         /// </summary>
         [DataTestMethod]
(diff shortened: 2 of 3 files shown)


src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4224,6 +4224,42 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+            {
+                    $"--ConfigFileName={configFileName}"
+            };
+
+            using (TestServer server = new(Program.CreateWebHostBuilder(args)))
+            using (HttpClient client = server.CreateClient())
+            {
+                // Act
+                RuntimeConfigProvider configProvider = server.Services.GetService<RuntimeConfigProvider>();
+                using HttpRequestMessage restRequest = new(HttpMethod.Get, "/api/Book?$orderby=id desc&$filter=publisher_id eq 1234");
```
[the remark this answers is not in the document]
RubenCerna2079: It is expected that the user uses the literal '$', this requires no change

src/Core/Services/RestService.cs
```diff
@@ -71,7 +71,7 @@ RequestValidator requestValidator
             DatabaseObject dbObject = sqlMetadataProvider.EntityToDatabaseObject[entityName];
 
             QueryString? query = GetHttpContext().Request.QueryString;
-            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString();
+            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString().Replace("%24", "$");    //Add replacement in order to ensure '$' sign is present as expected
```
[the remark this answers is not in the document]
RubenCerna2079: Comment is valid, the replace method now takes into consideration that the '$' sign happens after '?' and '&'.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4224,6 +4224,42 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+            string[] args = new[]
+            {
+                    $"--ConfigFileName={configFileName}"
+            };
+
+            using (TestServer server = new(Program.CreateWebHostBuilder(args)))
+            using (HttpClient client = server.CreateClient())
+            {
+                // Act
+                RuntimeConfigProvider configProvider = server.Services.GetService<RuntimeConfigProvider>();
```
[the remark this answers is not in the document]
RubenCerna2079: Removed unnecessary variable

anushakolan (approved):

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4224,6 +4224,41 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+            string[] args = new[]
+            {
+                    $"--ConfigFileName={configFileName}"
+            };
+
+            using (TestServer server = new(Program.CreateWebHostBuilder(args)))
+            using (HttpClient client = server.CreateClient())
+            {
+                // Act
+                using HttpRequestMessage restRequest = new(HttpMethod.Get, "/api/Book?$orderby=id desc&$filter=publisher_id eq 1234");
```
souvikghosh04: I don't think this explicitly tests the `%24`. the httpclient would internally normalize this, so this can pass with or without the fix. you might need to explicitly pass the encoded values in URI `"/api/Book?%24orderby=id%20desc&%24filter=publisher_id%20eq%201234")` and then do proper assertions.

src/Core/Services/RestService.cs
```diff
@@ -71,7 +71,8 @@ RequestValidator requestValidator
             DatabaseObject dbObject = sqlMetadataProvider.EntityToDatabaseObject[entityName];
 
             QueryString? query = GetHttpContext().Request.QueryString;
-            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString();
+            string queryString = query is null ? string.Empty : GetHttpContext().Request.QueryString.ToString()
+                .Replace("?%24", "?$").Replace("&%24", "&$");   //Add replacement in order to ensure '$' sign is present as expected only for the case where it is after a '?' or '&' signs.
```
souvikghosh04: shouldn't this be parsed using already available libraries instead of manually? we have something like this below `                    context.RawQueryString = queryString;
                    context.ParsedQueryString = HttpUtility.ParseQueryString(queryString);` or using `Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery()`

souvikghosh04: this should be solved/fixed at root level to ensure downstream usage of the querystring is already valid.

souvikghosh04: the trailing comment on RestService.cs:75 says what the code does but not why $ arrives URL-encoded in the first place, nor what PR #3080 changed that broke it. future readers will look at the Replace chain and have no idea whether they can remove it.

src/Service.Tests/Configuration/ConfigurationTests.cs
```diff
@@ -4224,6 +4224,41 @@ public async Task OpenApi_InteractiveSwaggerUI(
...
+                using HttpResponseMessage restResponse = await client.SendAsync(restRequest);
+
+                // Assert - Verify REST response
+                Assert.AreEqual(HttpStatusCode.OK, restResponse.StatusCode, "REST request to auto-generated entity should succeed");
```
souvikghosh04, line 4260: this only checks HttpStatusCode.OK. a regression where $filter is silently dropped would still return 200 with an unfiltered list. can we also assert that the response body contains the rows you expect (publisher_id == 1234 only) and in the order requested.

souvikghosh04 (changes requested):
needs some changes on parsing logic and additional tests assesrtion

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: No, the whole point of the test is to make an end-to-end scenario with values that a user would use. Adding the encoded values directly would defeat that purpose and would not help us catch any possible regressions in the future. DAB should be able to encode these values by itself and not throw any errors.

souvikghosh04: the intent is fair but partially justifiable. this test doesn't hit the fix's code path. HttpClient won't percent-encode $ (it's a legal sub-delimiter per RFC 3986), so Request.QueryString never contains %24 here, the Replace calls never fire. you can verify by reverting the fix and confirming the test still passes.

DAB is the server. it doesn't encode. it decodes what arrives on the wire. The regression is specifically when a client sends %24filter. the test must send the encoded form.

Suggestion: keep both as [DataRow] inputs: the current URI as a smoke test, plus a %24-encoded variant as the actual regression guard.

RubenCerna2079: Added new data row as you asked.

src/Core/Services/RestService.cs
RubenCerna2079: Had copilot address this issue, it changed the format of how we solve the issue.

src/Service.Tests/Configuration/ConfigurationTests.cs
RubenCerna2079: Added missing section of test

souvikghosh04 (approved):
LGTM. thanks for addressing the comments.

souvikghosh04 (approved):

PR #3669, opened 2026-06-16
Restrict POST /configuration endpoint to loopback

souvikghosh04:
## Summary

Restrict the late-configured `POST /configuration` endpoint so it is only callable from the loopback interface, with an optional bootstrap-token check on top.

## Why

`POST /configuration` is used in hosted mode (e.g. Azure Static Web Apps) to deliver the runtime config — including the database connection string — to an uninitialized DAB instance. It currently runs before the authentication middleware and has no other access control, which is more permissive than required for what is effectively a host-to-runtime bootstrap channel.

## Change

In `src/Service/Startup.cs`, the existing middleware that gates `POST /configuration` now also checks `IsConfigurationRequestAuthorized(HttpContext)` and returns `403 Forbidden` if it fails. The check enforces:

1. The request must originate from a loopback address. Null `RemoteIpAddress` (in-process callers such as `TestServer`) is treated as loopback.
2. If the `DAB_CONFIG_AUTH_TOKEN` environment variable is set, the request must include a matching `X-DAB-CONFIG-AUTH` header (fixed-time comparison).

The rest of the middleware ordering is unchanged.

## Backward compatibility

- Azure Static Web Apps platform config injector (posts over loopback inside the container) keeps working unchanged.
- Existing in-process `TestServer` tests keep working unchanged.
- The bootstrap token is opt-in — when the env var is unset, behavior on loopback matches today's behavior.

## Tests

- [x] Unit Tests

### DIFF
--- a/src/Service.Tests/Configuration/ConfigurationEndpointAuthorizationTests.cs
+++ b/src/Service.Tests/Configuration/ConfigurationEndpointAuthorizationTests.cs
@@ -104,25 +104,43 @@ public class ConfigurationEndpointAuthorizationTests
         HttpStatusCode expectedStatus)
     {
         // Arrange
+        // Other tests in the assembly (e.g. TestLoadingLocalCosmosSettings) set
+        // ASPNETCORE_ENVIRONMENT / DAB_ENVIRONMENT without cleanup. Those env vars cause
+        // FileSystemRuntimeConfigLoader to find an environment-specific dab-config.*.json on
+        // disk and auto-initialize the runtime, which makes POST /configuration return
+        // 409 Conflict before our security middleware can be exercised. Snapshot and clear
+        // them so each test starts from an uninitialized runtime.
+        string originalAspNetEnv = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
+        string originalDabEnv = Environment.GetEnvironmentVariable("DAB_ENVIRONMENT");
+        Environment.SetEnvironmentVariable("ASPNETCORE_ENVIRONMENT", null);
+        Environment.SetEnvironmentVariable("DAB_ENVIRONMENT", null);
         Environment.SetEnvironmentVariable(Startup.CONFIG_AUTH_TOKEN_ENV_VAR, configuredToken);
 
-        using TestServer server = new(Program.CreateWebHostFromInMemoryUpdatableConfBuilder(Array.Empty<string>()));
-        using HttpClient httpClient = server.CreateClient();
+        try
+        {
+            using TestServer server = new(Program.CreateWebHostFromInMemoryUpdatableConfBuilder(Array.Empty<string>()));
+            using HttpClient httpClient = server.CreateClient();
 
-        HttpRequestMessage request = new(HttpMethod.Post, configurationEndpoint)
-        {
-            Content = BuildPostContent(configurationEndpoint),
-        };
-        if (providedHeader is not null)
-        {
-            request.Headers.Add(Startup.CONFIG_AUTH_HEADER, providedHeader);
+            HttpRequestMessage request = new(HttpMethod.Post, configurationEndpoint)
+            {
+                Content = BuildPostContent(configurationEndpoint),
+            };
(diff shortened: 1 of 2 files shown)


src/Service.Tests/Configuration/ConfigurationEndpointAuthorizationTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        {
+            Environment.SetEnvironmentVariable("ASPNETCORE_ENVIRONMENT", originalAspNetEnv);
+            Environment.SetEnvironmentVariable("DAB_ENVIRONMENT", originalDabEnv);
+        }
+    }
+
+    [TestCleanup]
+    public void Cleanup()
+    {
+        Environment.SetEnvironmentVariable(Startup.CONFIG_AUTH_TOKEN_ENV_VAR, null);
```
anushakolan: We restore ASPNETCORE_ENVIRONMENT and DAB_ENVIRONMENT in the finally block.

However, DAB_CONFIG_AUTH_TOKEN is set but not restored. TestCleanup resets it to null instead of the original value. This could affect other tests if they rely on a pre-existing value and may introduce order-dependent behavior.

Can we snapshot and restore it in finally like the other env vars to keep tests fully isolated?

anushakolan (commented):
Just a minor comment to address; otherwise LGTM.

It would be helpful to create a tracking task and link it to this PR, along with relevant context. This makes it easier to understand the background and motivations for the change in the future.

anushakolan (approved):

src/Service.Tests/Configuration/ConfigurationEndpointAuthorizationTests.cs
```diff
@@ -0,0 +1,185 @@
...
+    [DataRow("10.0.0.1", null, null, false, DisplayName = "Private IPv4 (10/8), no token => deny")]
+    [DataRow("172.17.0.1", null, null, false, DisplayName = "Docker bridge IP, no token => deny")]
+    [DataRow("203.0.113.50", null, null, false, DisplayName = "Public IPv4, no token => deny")]
+    // --- Bootstrap token configured ---
+    [DataRow("127.0.0.1", "secret", "secret", true, DisplayName = "Loopback + correct token => allow")]
+    [DataRow("127.0.0.1", "secret", "wrong", false, DisplayName = "Loopback + wrong token => deny")]
+    [DataRow("127.0.0.1", "secret", null, false, DisplayName = "Loopback + missing header => deny")]
+    [DataRow("192.168.1.50", "secret", "secret", false, DisplayName = "Non-loopback + correct token => still deny")]
+    [DataRow(null, "secret", null, false, DisplayName = "In-process + missing header => deny")]
+    [DataRow(null, "secret", "secret", true, DisplayName = "In-process + correct token => allow")]
```
RubenCerna2079, line 94: Shouldn't we also add private and public IP addresses with the correct and the wrong tokens to ensure it works as expected with those? Or is that already covered with the loopback tests?

RubenCerna2079 (approved):
LGTM! Approving with the expectation that the comment from Anusha gets resolved.

src/Service.Tests/Configuration/ConfigurationEndpointAuthorizationTests.cs
souvikghosh04: Centralized snapshot/restore of DAB_CONFIG_AUTH_TOKEN, ASPNETCORE_ENVIRONMENT, and DAB_ENVIRONMENT into [TestInitialize]/[TestCleanup]. Original values are now preserved verbatim. Also simplified EndToEnd_PostConfiguration_StatusCodeMatrix by removing its inline snapshot/finally block (no longer needed).

src/Service.Tests/Configuration/ConfigurationEndpointAuthorizationTests.cs
souvikghosh04: Added new ones to cover 4 such scenarios.

RubenCerna2079 (approved):

PR #3677, opened 2026-06-22
Add Support for Vector Data Type in SQL for REST

### LINKED ISSUE #3654: Foundation for adding support for Vector Data Type

RubenCerna2079:
## Why make this change?

- Solves issue #3654 

## What is this change?

These changes add the base to support the Vector data type for REST, GraphQL, and MCP. This PR focuses mainly on the REST endpoint, ensuring that it is able to both read and write to the vector columns through it. Lastly, it also adds support to ensure that OpenAPI is able to show columns with the Vector data type as `array: [numeric]`.

The following files were changed to show that DAB is able to support the Vector data type and also change the type internally to system types it can work with. Such as the `float` type for `DbType`, the `Numeric` type for `JsonDataType`, and `SINGLE_TYPE` for GraphQL.
- SqlTypeConstants
- TypeHelper.cs
- SqlSchemaConverter.cs

The following file was changed to allow DAB to save the vector columns as an array type, which is necessary for GraphQL to be able to know if the column of an entity is a simple type or an array. 
- MsSqlMetadataProvider.cs

The following files were changed to parse the information from the user to a format that can be used by DAB to create the queries for SQL. 
- BaseSqlQueryStructure.cs: Parses the array information by creating a new function that checks for the simple type of the array (e.g If it is an array of floats it will use the type float) and parse each value inside the array.
- MsSqlQueryExecutor.cs: Parses the values given by the user and adds them to an SqlVector object which allows for an easier way to query the values.

The following file was changed to allow DAB's OpenAPI to support array types by checking if the main type is `array` and then taking the subtype and creating the appropriate item with it.
- OpenApiDocumentor.cs

Note: All of these changes are only applicable to Vectors of Float32, support for Vectors of Float16 will have to wait as support for it in .NET is still missing.

## How was this tested?

- [x] Integration Tests
- [ ] Unit Tests

Added tests that ensure the following scenarios on REST:
- Read regular vector 
- Read vector with max number of dimensions possible
- Read vector with null value
- Read vector with `$first` argument
- Read vector with `$after` argument
- Read vector with `$filter` argument - fails
- Read vector with `$orderby` argument - fails
- Insert regular vector 
- Insert vector with null value
- Insert vector with number values in scientific format
- Insert vector with number values in string format
- Insert vector with more dimensions than allowed - fails
- Insert vector with non-valid values - fails
- Update vector with new values

## Sample Request(s)
Note: Currently the values will be outputted in scientific notation, this will be changed in the future https://github.com/Azure/data-api-builder/issues/3680

GET https://localhost:5001/api/dbo_normalvector/
image

POST https://localhost:5001/api/dbo_normalvector/
```
{
	"ProductID": 2100,
	"Embedding": [
		1.15202468461012559910005,
		0.00005,
		5.111111111
	]
}
```
image

PATCH https://localhost:5001/api/dbo_normalvector/ (Adding new row)
```
{
	"ProductID": 3100,
	"Embedding": [
		1.15202468461012559910005,
		0.00005,
		5.111111111
	]
}
```
image

PATCH https://localhost:5001/api/dbo_normalvector/ (Update existing row)
```
{
	"ProductID": 2100,
	"Embedding": [
		25,
		1.2,
		510.35
	]
}
```
image

PUT https://localhost:5001/api/dbo_normalvector/ (Adding new row)
```
{
	"ProductID": 4100,
	"Embedding": [
		0,
		1,
		37
	]
}
```
image

PUT https://localhost:5001/api/dbo_normalvector/ (Edit existing row)
```
{
	"ProductID": 2100,
	"Embedding": [
		0.220224962,
		0.66666667,
		37
	]
}
```
image

https://localhost:5001/swagger
image
image

### DIFF
--- a/src/Core/Resolvers/MsSqlQueryExecutor.cs
+++ b/src/Core/Resolvers/MsSqlQueryExecutor.cs
@@ -18,6 +18,7 @@ using Azure.DataApiBuilder.Service.Exceptions;
 using Azure.Identity;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Data.SqlClient;
+using Microsoft.Data.SqlTypes;
 using Microsoft.Extensions.Logging;
 
 namespace Azure.DataApiBuilder.Core.Resolvers
@@ -693,6 +694,19 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                         && parameterEntry.Value?.Length is not null)
                     {
                         parameter.Size = parameterEntry.Value.Length.Value;
+                    }
+
+                    // if sqldbtype is vector then set the value as an SqlVector object
+                    if (parameter.SqlDbType is SqlDbType.Vector)
+                    {
+                        List<float> values = new();
+                        foreach (float val in (Array)parameter.Value)
+                        {
+                            values.Add(val);
+                        }
+
+                        SqlVector<float> value = new(values.ToArray());
+                        parameter.Value = value;
                     }
 
                     cmd.Parameters.Add(parameter);
--- a/src/Core/Resolvers/QueryExecutor.cs
+++ b/src/Core/Resolvers/QueryExecutor.cs
@@ -13,6 +13,7 @@ using Azure.DataApiBuilder.Core.Configurations;
 using Azure.DataApiBuilder.Core.Models;
 using Azure.DataApiBuilder.Service.Exceptions;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Data.SqlTypes;
 using Microsoft.Extensions.Logging;
 using Polly;
 using Polly.Retry;
@@ -502,7 +503,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                             {
                                 if (!ConfigProvider.GetConfig().MaxResponseSizeLogicEnabled())
                                 {
-                                    dbResultSetRow.Columns.Add(columnName, dbDataReader[columnName]);
+                                    dbResultSetRow.Columns.Add(columnName, GetColumnInformation(dbDataReader, columnName));
                                 }
                                 else
                                 {
@@ -883,6 +884,22 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                     statusCode: HttpStatusCode.RequestEntityTooLarge,
                     subStatusCode: DataApiBuilderException.SubStatusCodes.ErrorProcessingData);
             }
+        }
+
+        /// <summary>
+        /// Helper function to get column information from the DbDataReader and handle special cases like SqlVector<float>.
+        /// </summary>
+        /// <param name="dbDataReader"></param>
+        /// <param name="columnName"></param>
+        /// <returns></returns>
+        private static object GetColumnInformation(DbDataReader dbDataReader, string columnName)
+        {
+            if (dbDataReader[columnName] is SqlVector<float> columnValue)
+            {
+                return columnValue.Memory;
+            }
+
+            return dbDataReader[columnName];
         }
 
         internal virtual void AddDbExecutionTimeToMiddlewareContext(long time)
--- a/src/Core/Resolvers/QueryExecutor.cs
+++ b/src/Core/Resolvers/QueryExecutor.cs
@@ -894,12 +894,8 @@ namespace Azure.DataApiBuilder.Core.Resolvers
         /// <returns></returns>
         private static object GetColumnInformation(DbDataReader dbDataReader, string columnName)
         {
-            if (dbDataReader[columnName] is SqlVector<float> columnValue)
-            {
-                return columnValue.Memory;
-            }
-
-            return dbDataReader[columnName];
+            object value = dbDataReader[columnName];
+            return value is SqlVector<float> vector ? vector.Memory : value;
         }
 
         internal virtual void AddDbExecutionTimeToMiddlewareContext(long time)
--- a/src/Core/Resolvers/QueryExecutor.cs
+++ b/src/Core/Resolvers/QueryExecutor.cs
@@ -555,7 +555,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                             {
                                 if (!ConfigProvider.GetConfig().MaxResponseSizeLogicEnabled())
                                 {
-                                    dbResultSetRow.Columns.Add(columnName, dbDataReader[columnName]);
+                                    dbResultSetRow.Columns.Add(columnName, GetColumnInformation(dbDataReader, columnName));
                                 }
                                 else
                                 {
@@ -823,7 +823,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
             {
                 dataRead = columnSize;
                 ValidateSize(availableBytes, dataRead);
-                dbResultSetRow.Columns.Add(columnName, dbDataReader[columnName]);
+                dbResultSetRow.Columns.Add(columnName, GetColumnInformation(dbDataReader, columnName));
             }
 
             return dataRead;
--- a/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
+++ b/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
@@ -479,21 +479,22 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                     throw new NotSupportedException($"{systemType.Name} is not supported");
             }
 
-            using JsonDocument arg = JsonDocument.Parse(param);
-            if (arg.RootElement.ValueKind != JsonValueKind.Array)
+            try
             {
-                throw new FormatException($"Expected an array for {systemType.Name} but got {arg.RootElement.ValueKind}");
-            }
+                List<object> list = new();
+                object[] values = JsonSerializer.Deserialize<object[]>(param) ?? Array.Empty<object>();
+                for (int i = 0; i < values.Length; i++)
+                {
+                    string stringValue = values[i]?.ToString() ?? string.Empty;
+                    values[i] = ParseParamAsSystemType(stringValue, typeOfArray);
+                }
 
-            List<object> list = new();
-            object[] values = JsonSerializer.Deserialize<object[]>(param) ?? Array.Empty<object>();
-            foreach (object value in values)
+                return values;
+            }
+            catch
             {
-                string stringValue = value.ToString() ?? string.Empty;
-                list.Add(ParseParamAsSystemType(stringValue, typeOfArray));
+                throw new FormatException($"Expected an array for {systemType.Name} but got an unexpected value");
             }
-
-            return list.ToArray();
         }
 
         /// <summary>
--- a/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
+++ b/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
@@ -481,7 +481,6 @@ namespace Azure.DataApiBuilder.Core.Resolvers
 
             try
             {
-                List<object> list = new();
                 object[] values = JsonSerializer.Deserialize<object[]>(param) ?? Array.Empty<object>();
                 for (int i = 0; i < values.Length; i++)
                 {
--- a/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
+++ b/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
@@ -473,7 +473,7 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                 case "Single[]":
                     typeOfArray = typeof(Single);
                     break;
-                    
+
                 default:
                     throw new NotSupportedException($"{systemType.Name} is not supported");
             }
--- a/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
+++ b/src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
@@ -453,9 +453,28 @@ namespace Azure.DataApiBuilder.Core.Resolvers
                 "Guid" => Guid.Parse(param),
                 "TimeOnly" => TimeOnly.Parse(param),
                 "TimeSpan" => TimeOnly.Parse(param),
-                "Single[]" => JsonSerializer.Deserialize<float[]>(param) ?? Array.Empty<float>(),
+                "Single[]" => ParseArrayIntoSystemType(param, systemType),
                 _ => throw new NotSupportedException($"{systemType.Name} is not supported")
             };
+        }
+
+        private static object ParseArrayIntoSystemType(string param, Type systemType)
+        {
+            switch (systemType.Name)
+            {
+                case "Single[]":
+                    List<object> list = new();
+                    object[] values = JsonSerializer.Deserialize<object[]>(param) ?? Array.Empty<object>();
+                    foreach (object value in values)
+                    {
+                        string stringValue = value.ToString() ?? string.Empty;
+                        list.Add(ParseParamAsSystemType(stringValue, typeof(Single)));
+                    }
+
+                    return list.ToArray();
+                default:
+                    throw new NotSupportedException($"{systemType.Name} is not supported");
+            }
         }
 
         /// <summary>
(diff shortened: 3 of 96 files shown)


src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -452,10 +453,30 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
                 "Guid" => Guid.Parse(param),
                 "TimeOnly" => TimeOnly.Parse(param),
                 "TimeSpan" => TimeOnly.Parse(param),
+                "Single[]" => ParseArrayIntoSystemType(param, systemType),
                 _ => throw new NotSupportedException($"{systemType.Name} is not supported")
             };
         }
 
+        private static object ParseArrayIntoSystemType(string param, Type systemType)
```
RubenCerna2079, line 469: With this pattern we can also write to other array values such as the ones found in PostgreSQL. Let me know if you think this is viable.

src/Service.Tests/DatabaseSchema-MsSql.sql
```diff
@@ -232,6 +233,12 @@ CREATE TABLE type_table(
     uuid_types uniqueidentifier DEFAULT newid()
 );
 
+CREATE TABLE vector_type_table(
+    id int IDENTITY(5001, 1) PRIMARY KEY,
+    vector_data vector(3),
+    vector_data_max vector(1998)
+);
```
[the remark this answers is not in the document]
RubenCerna2079: This will be addressed

src/Service.Tests/SqlTests/RestApiTests/MsSqlRestVectorTypesTests.cs
```diff
@@ -0,0 +1,286 @@
...
+        /// since vector(N) stores 32-bit floats which may not round-trip exactly through JSON.
+        /// </summary>
+        private const double VECTOR_COMPONENT_DELTA = 0.0001;
+
+        [ClassInitialize]
+        public static async Task SetupAsync(TestContext context)
+        {
+            DatabaseEngine = TestCategory.MSSQL;
+            await InitializeTestFixture();
+        }
```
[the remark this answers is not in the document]
RubenCerna2079: Same issue as comment above, will fix this.

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -695,6 +696,19 @@ public override SqlCommand PrepareDbCommand(
...
+                    {
+                        List<float> values = new();
+                        foreach (float val in (Array)parameter.Value)
+                        {
+                            values.Add(val);
+                        }
+
+                        SqlVector<float> value = new(values.ToArray());
+                        parameter.Value = value;
+                    }
```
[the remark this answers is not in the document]
RubenCerna2079: Issue is irrelevant, when using Null values we don't pass through that part of the code. Also, tests confirm we can write Null values without any issue.

src/Service.Tests/SqlTests/RestApiTests/MsSqlRestVectorTypesTests.cs
```diff
@@ -0,0 +1,286 @@
...
+        [DataTestMethod]
+        [DataRow("{ \"vector_data\": [0.125, 0.25, 0.5] }", new[] { 0.125f, 0.25f, 0.5f }, true, DisplayName = "Insert valid vector")]
+        [DataRow("{ \"vector_data\": null }", null, true, DisplayName = "Insert valid null vector")]
+        [DataRow("{ \"vector_data\": [5e-1, 2.5e-1, 7.5e-1] }", new[] { 0.5f, 0.25f, 0.75f }, true, DisplayName = "Insert valid vector with scientific notation")]
+        [DataRow("{ \"vector_data\": [\"0.5\", \"0.25\", \"0.75\"] }", new[] { 0.5f, 0.25f, 0.75f }, true, DisplayName = "Insert valid vector with numbers as string values")]
+        [DataRow("{ \"vector_data\": [1.25, 2.25, 3.25, 4.25] }", null, false, DisplayName = "Insert invalid vector with more dimensions than allowed")]
+        [DataRow("{ \"vector_data\": [\"not\", \"a\", \"number\"] }", null, false, DisplayName = "Insert invalid vector with invalid values")]
+        public async Task InsertVectorType(string requestBody, float[] expectedValue, bool expectedSuccess)
+        {
+            HttpResponseMessage postResponse = await HttpClient.PostAsync(
```
Aniruddh25, line 106: In the description, the `value` returned for the vector type column contains properties like `IsNull`, `Length` and the actual vector elements are present in the property `Memory`. Why are these additional properties being returned?

src/Service.Tests/SqlTests/RestApiTests/MsSqlRestVectorTypesTests.cs
```diff
@@ -0,0 +1,286 @@
...
+
+            JsonElement readBack = await GetRecordByIdAsync(4);
+            AssertVectorEquals(readBack.GetProperty("vector_data"), expected);
+        }
+
+        /// <summary>
+        /// PATCH /api/VectorType/id/5 - Verify that an existing record's vector value is updated.
+        /// </summary>
+        [TestMethod]
+        public async Task PatchVectorType_Update()
```
Aniruddh25, line 172: Is there a test for DELETE? Do we support the DELETE verb?

RubenCerna2079: We do not have a specific test for DELETE, but in all of the tests related to INSERT we use the DELETE method and assert that the operation was successful.

src/Service.Tests/SqlTests/RestApiTests/MsSqlRestVectorTypesTests.cs
RubenCerna2079: That is the information that it returns because we use it as an SqlVector object, I just added some parsing to ensure that it is returned as an array

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
```diff
@@ -577,6 +577,7 @@ public static string GetGraphQLTypeFromSystemType(Type type)
                 "Byte[]" => BYTEARRAY_TYPE,
                 "TimeOnly" => LOCALTIME_TYPE,
                 "TimeSpan" => LOCALTIME_TYPE,
+                "SqlVector`1" => SINGLE_TYPE,
```
Aniruddh25: What is the ``1` for?

src/Core/Resolvers/MsSqlQueryExecutor.cs
```diff
@@ -695,6 +696,19 @@ public override SqlCommand PrepareDbCommand(
                         parameter.Size = parameterEntry.Value.Length.Value;
                     }
 
+                    // if sqldbtype is vector then set the value as an SqlVector object
+                    if (parameter.SqlDbType is SqlDbType.Vector)
+                    {
+                        List<float> values = new();
+                        foreach (float val in (Array)parameter.Value)
```
aaronburtle, line 703: If the parameter.Value here is null (from parameter.Value = parameterEntry.Value?.Value ?? DBNull.Value), then doesn't this (`(Array)parameter.Value`) lead to an InvalidCastException?

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -885,6 +886,22 @@ private void ValidateSize(long availableSizeBytes, long sizeToBeReadBytes)
             }
         }
 
+        /// <summary>
+        /// Helper function to get column information from the DbDataReader and handle special cases like SqlVector<float>.
+        /// </summary>
+        /// <param name="dbDataReader"></param>
+        /// <param name="columnName"></param>
+        /// <returns></returns>
+        private static object GetColumnInformation(DbDataReader dbDataReader, string columnName)
```
aaronburtle, line 895: It seems that this helper, which converts SqlVector<float> to .Memory so its serializes as a JSON array is only called from the non-max response branch in the async ExtractRestulSetFromDbDataReaderAsync function

                                ```if (!ConfigProvider.GetConfig().MaxResponseSizeLogicEnabled())
                                {
                                    dbResultSetRow.Columns.Add(columnName, GetColumnInformation(dbDataReader, columnName));
                                }```

I think this means that there are a number of paths that end up returning the raw SqlVector<float> type. So PUT/POST/PATCH along with the max response size path for GET will all have the wrong type.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
aaronburtle: I don't think this code is reachable, but I think if it was it would be wrong. So by the time we get to this code, the metadata provider has already rewritten SystemType to Single[], and GetGraphQLTypeFromSystemTYpe will unwrap the array to the element type first, so the "Single" case is hit and we never reach this type.

But if we were to reach this here, I think mapping a vector to a scalar (SINGLE_TYPE) rather than a list ([FLOAT]) is incorrect.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -452,10 +453,49 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
...
 
+        /// <summary>
+        /// Takes the array of the parameter we are going to parse and converts each element to the specified system type.
+        /// </summary>
+        /// <param name="param"></param>
+        /// <param name="systemType"></param>
+        /// <returns></returns>
+        /// <exception cref="NotSupportedException"></exception>
+        /// <exception cref="FormatException"></exception>
+        private static object ParseArrayIntoSystemType(string param, Type systemType)
```
aaronburtle, line 469: I think this flow of this function can be improved. The payload is parsed twice, once in JsonDocument.Parse(param) on 482, in order to check the ValueKind, and then again via JsonSerializer.Deserialize<object[]>(param). It then builds a List by parsing yet again. So 3 passes and several allocations all for one vector.

I think it should be reasonable to parse once and produce float[] directly, which can then be consumed by the executor.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -452,10 +453,49 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
...
+            if (arg.RootElement.ValueKind != JsonValueKind.Array)
+            {
+                throw new FormatException($"Expected an array for {systemType.Name} but got {arg.RootElement.ValueKind}");
+            }
+
+            List<object> list = new();
+            object[] values = JsonSerializer.Deserialize<object[]>(param) ?? Array.Empty<object>();
+            foreach (object value in values)
+            {
+                string stringValue = value.ToString() ?? string.Empty;
```
aaronburtle: if `value` here is null I think this will throw a NullReferenceException, the `??` won't help because the .ToString() will already be invoked on the null already.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -452,10 +453,49 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
...
+            {
+                case "Single[]":
+                    typeOfArray = typeof(Single);
+                    break;
+
+                default:
+                    throw new NotSupportedException($"{systemType.Name} is not supported");
+            }
+
+            using JsonDocument arg = JsonDocument.Parse(param);
```
aaronburtle: I think if this Parse function throws a JSON exception this will be missed by GetParamAsSystemType and surface as a 500.

src/Service.Tests/SqlTests/RestApiTests/MsSqlRestVectorTypesTests.cs
```diff
@@ -0,0 +1,286 @@
...
+            {
+                Assert.IsFalse(postResponse.IsSuccessStatusCode, "Expected that inserting vector should fail.");
+            }
+        }
+
+        /// <summary>
+        /// PUT /api/VectorType/id/4 - Verify that an existing record's vector value is replaced (full update).
+        /// </summary>
+        [TestMethod]
+        public async Task PutVectorType_Update()
```
aaronburtle, line 133: nit: this test mutates data without restoring

Aniruddh25 (approved):
A few clarifying questions remain. But otherwise looks good.

src/Core/Resolvers/QueryExecutor.cs
RubenCerna2079: For the GET response, this is not an issue as we just receive the information in the form of a string not in the SqlVector object.
I will give an answer once I test the other REST requests.

src/Service.GraphQLBuilder/Sql/SchemaConverter.cs
RubenCerna2079: That is how the type appears inside .NET. Although it is not necessary to have it here.

RubenCerna2079: You are half right in the part that this code is not reachable. Due to how we set the metadata of vector rows in `MsSqlMetadataProvider.cs` it already sets its `IsArrayType` as true and its `ElementSystemType` as "Single". This causes it to go to line 554 in this file and uses the `ElementSystemType`. This only sets the base type, this is not where we set it as an array.

src/Core/Resolvers/MsSqlQueryExecutor.cs
RubenCerna2079: Copilot also asked this question, this is not an issue since DAB passes the `Null` values as `NVarChar` type so it would not even go through this section.
https://github.com/Azure/data-api-builder/pull/3677#discussion_r3470806507

src/Core/Resolvers/QueryExecutor.cs
RubenCerna2079: Will add the new helper function in the other sections that also get the column information when there is a max response size

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
RubenCerna2079: I made it better by only having to parse via JsonSerializer.Deserialize<object[]>(param) and then parse again each of the objects inside the array individually into the same object[]. I think this is the best we can do, since we want to ensure that all the values inside the array are parsed properly.

src/Core/Resolvers/Sql Query Structures/BaseSqlQueryStructure.cs
```diff
@@ -452,10 +453,50 @@ protected static object ParseParamAsSystemType(string param, Type systemType)
...
+                    typeOfArray = typeof(Single);
+                    break;
+
+                default:
+                    throw new NotSupportedException($"{systemType.Name} is not supported");
+            }
+
+            try
+            {
+                List<object> list = new();
```
aaronburtle: nit: this list appears to be unused.

src/Core/Resolvers/QueryExecutor.cs
```diff
@@ -885,6 +886,22 @@ private void ValidateSize(long availableSizeBytes, long sizeToBeReadBytes)
...
+        {
+            if (dbDataReader[columnName] is SqlVector<float> columnValue)
+            {
+                return columnValue.Memory;
+            }
+
+            return dbDataReader[columnName];
```
aaronburtle: There are 2 reads here, which ordinarily would not be a problem, however, if the config is set with MaxResponseSizeLogicEnabled, then the reader's command behavior is set to Sequential Access. That can cause problems with a double read.

easy fix though, something like this should do it

```
private static object GetColumnInformation(DbDataReader dbDataReader, string columnName)
{
    object value = dbDataReader[columnName];
    return value is SqlVector<float> vector ? vector.Memory : value;
}
```

aaronburtle (approved):
Just a couple final comments, but once they are addressed, should be good to go.

Aniruddh25 (approved):

aaronburtle (approved):
re-approving for policy

